PR #307, opened 2026-03-01
fix(attendance): recover longrun async commit after poll timeout

zensgit:
## Summary\n- treat async commit poll timeout/transient timeout as retryable in perf runner\n- add idempotency-based recovery path that replays commit-async without commitToken\n- continue polling recovered job with configurable grace window\n\n## Why\nLongrun scenarios (100k/500k commit) intermittently fail from transient 5xx on /attendance/import/jobs/:id and timeout before backend finishes. This change recovers by reloading the existing async job via idempotency key before failing.\n\n## Validation\n- node --check scripts/ops/attendance-import-perf.mjs\n- inspected failed GA artifacts: 22536868864-r3 / 22537963463-r3 / 22539314446-r3\n


zensgit:
Follow-up pushed:\n- serialize longrun matrix () to reduce async worker contention\n- docs updated with latest evidence and remaining P1 status\n\nCurrent latest non-drill run on this branch: #22555028596 -> rows100k-commit still times out (), P1 issue #157 remains open.

PR #556, opened 2026-03-27
fix(ci): harden storage compose exec

zensgit:
## Summary
- replace the eval-based backend exec wrapper in `attendance-check-storage.sh`
- branch explicitly on `docker compose` vs `docker-compose`
- add design / verification MD for the production storage health failure

## Why
`Attendance Remote Storage Health (Prod)` currently fails after falling back to backend exec with:

- `unknown shorthand flag: 'f' in -f`

That is a compose transport bug, not a real storage threshold failure.

## Verify
- `git diff --check`
- `bash -n scripts/ops/attendance-check-storage.sh`
- `rg -n "docker compose -f|docker-compose -f|unsupported compose command" scripts/ops/attendance-check-storage.sh`
- Claude Code boundary review => `SAFE_MINIMAL_UNBLOCK`

### DIFF
--- a/docs/development/attendance-remote-storage-compose-exec-fix-20260327.md
+++ b/docs/development/attendance-remote-storage-compose-exec-fix-20260327.md
@@ -1,0 +1,45 @@
+# Attendance Remote Storage Compose Exec Fix
+
+## Context
+
+After the deploy chain was fully recovered, a separate production workflow still failed:
+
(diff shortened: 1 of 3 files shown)


zensgit:
Follow-up note: this slice only fixes the deterministic compose transport bug in storage fallback exec. Preflight/Metrics summary behavior is being triaged separately and is intentionally not mixed into this PR.

PR #786, opened 2026-04-11
feat(after-sales): add runtime admin surface

zensgit:
## Summary
- add plugin-local runtime admin APIs for after-sales automations and refundAmount field policies
- add the corresponding AfterSalesView runtime admin UI and test coverage
- sync the after-sales manifest workflow list with the real default automation set
- add design and verification docs for the runtime-admin slice

## Verification
- `pnpm --filter @metasheet/core-backend test:unit`
- `DATABASE_URL=postgresql://metasheet:metasheet123@localhost:5432/metasheet_v2 pnpm --filter @metasheet/core-backend exec vitest --config vitest.integration.config.ts run tests/integration/after-sales-plugin.install.test.ts --reporter=dot`
- `pnpm --filter @metasheet/core-backend build`
- `pnpm --filter @metasheet/web exec vitest run tests/AfterSalesView.spec.ts tests/AfterSalesView.follow-ups.spec.ts tests/AfterSalesView.customers.spec.ts tests/AfterSalesView.installed-assets.spec.ts tests/AfterSalesView.service-records.spec.ts --watch=false`
- `pnpm --filter @metasheet/web build`

## Docs
- `docs/after-sales-runtime-admin-v1-design-20260411.md`
- `docs/after-sales-runtime-admin-v1-verification-20260411.md`

PR #788, opened 2026-04-11
fix(after-sales): surface runtime admin refresh failures

zensgit:
## Summary
- fix AfterSalesView runtime-admin save flows so post-save refresh failures surface as errors
- cover both failure paths with focused frontend tests

## Context
- follow-up to merged PR #786
- #786 landed the runtime-admin surface itself
- this PR only keeps the save UX honest when the follow-up refresh fails after a successful write

## Verification
- `pnpm --filter @metasheet/web exec vitest run tests/AfterSalesView.spec.ts --watch=false`

## Scope
- `apps/web/src/views/AfterSalesView.vue`
- `apps/web/tests/AfterSalesView.spec.ts`

### DIFF
--- a/apps/web/src/views/AfterSalesView.vue
+++ b/apps/web/src/views/AfterSalesView.vue
@@ -4132,7 +4132,10 @@ async function requestTicketRefund(ticket: TicketViewModel) {
   }
 }
 
-async function loadFieldPoliciesForCurrentState(state: CurrentResponse): Promise<void> {
+async function loadFieldPoliciesForCurrentState(
+  state: CurrentResponse,
+  options: { strict?: boolean } = {},
+): Promise<void> {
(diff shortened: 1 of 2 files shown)


zensgit (commented):
First-pass review: no blocking findings.

What I checked:
- `AfterSalesView.vue` now treats post-save refresh as part of the save success path instead of swallowing refresh failures
- the strict refresh path is only used from the two runtime-admin save handlers, so initial page load and non-admin 403 hiding semantics stay unchanged
- focused frontend coverage now exercises both failure cases

Local verification:
- `pnpm --filter @metasheet/web exec vitest run tests/AfterSalesView.spec.ts --watch=false`
- result: 29 tests passed

This looks like the right follow-up to merged PR #786. Once checks are green, the remaining blocker is approval/merge policy only.

PR #789, opened 2026-04-11
ci(after-sales): enforce integration test gate

zensgit:
## Summary
- add a dedicated `test:integration:after-sales` script for the real-DB after-sales integration file
- add a strict `after-sales integration` job to `plugin-tests.yml`
- keep the broader integration sweep non-blocking, but make this one critical after-sales path fail PR CI when it regresses

## Why
`after-sales-plugin.install.test.ts` is now carrying important real-DB regression coverage for installer, runtime-admin, approval bridge, and reinstall semantics. Before this PR, the generic integration step in `plugin-tests.yml` was still `|| true`, so failures in that file could be silently ignored by CI.

This PR gives us one explicit job with `DATABASE_URL` + migrations + real Postgres that must pass.

## Verification
- `DATABASE_URL=postgresql://metasheet:metasheet123@localhost:5432/metasheet_v2 pnpm --filter @metasheet/core-backend test:integration:after-sales`
  - passed locally: `24 tests` in `after-sales-plugin.install.test.ts`
- `git diff --check -- .github/workflows/plugin-tests.yml packages/core-backend/package.json`
  - passed

Related: #715

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -42,7 +42,7 @@ jobs:
       - name: Use empty npmrc for pnpm
         run: |
           echo "" > /tmp/empty-npmrc
-          echo "NPM_CONFIG_USERCONFIG=/tmp/empty-npmrc" >> $GITHUB_ENV
+          echo "NPM_CONFIG_USERCONFIG=/tmp/empty-npmrc" >> "$GITHUB_ENV"
 
       - name: Get pnpm store directory
         id: pnpm-cache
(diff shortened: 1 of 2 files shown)


zensgit (commented):
First-pass review: no blocking findings.

What I checked:
- the workflow change adds a dedicated `after-sales integration` job instead of relying on the broad integration sweep that can silently no-op without `DATABASE_URL`
- the new package script scopes execution to `tests/integration/after-sales-plugin.install.test.ts`
- the workflow still keeps the broader test matrix intact while making the real Postgres after-sales path visible in CI

Scope is appropriately narrow:
- `.github/workflows/plugin-tests.yml`
- `packages/core-backend/package.json`

Given the stated goal of turning the after-sales DB-backed integration path into an explicit CI signal, this change is directionally correct. Once checks are green, the remaining blocker is approval/merge policy only.

PR #808, opened 2026-04-11
feat(multitable): people-search endpoint + test mock sync

zensgit:
## Summary
- Add `GET /api/multitable/people-search?q=&baseId=&limit=` — non-admin user search via the system People sheet, supporting the @mention composer in comments
- Fix `capabilityOrigin` missing from grid/workbench mocks in `multitable-workbench-manager-flow` test (was causing test failure)
- Add `docs/multitable-view-compat-matrix.md` documenting which views use legacy bridge vs multitable API

## Test plan
- [x] `multitable-workbench-manager-flow.spec.ts` passes
- [ ] Manual: `GET /api/multitable/people-search?q=alice&baseId=<id>` returns matching people records


zensgit:
Review note: this PR is close, but I don't think `/people-search` is ready to merge as-is.

Main issue:

- `GET /api/multitable/people-search` does not enforce auth / readable-sheet capability checks before loading record summaries from the system people sheet. In the same file, existing read paths like `/records-summary` and `/fields/:fieldId/link-options` both call `resolveSheetReadableCapabilities(...)` before returning data. This new route currently goes straight from `baseId -> people sheet -> loadRecordSummaries(...)`, which looks like a permission boundary bypass.

Suggested fix:

- After resolving `peopleSheetId`, run the same `resolveSheetReadableCapabilities(req, pool.query.bind(pool), peopleSheetId)` gate and require authenticated read access before returning summaries.

Minor doc issue:

- `docs/multitable-view-compat-matrix.md` says "All 7 views" but the table only lists 6 views.

Once the ACL check is aligned with the existing summary/link-option routes, I'd be comfortable re-reviewing it.

zensgit:
Fixed both issues in 47ed100:

1. **ACL gate added**: `/people-search` now calls `resolveSheetReadableCapabilities(req, query, peopleSheetId)` and checks `capabilities.canRead` before returning summaries — aligned with `/records-summary` and `/fields/:fieldId/link-options` patterns.

2. **Doc fix**: Changed "All 7 views" to "All 6 view types (Grid, Form, Kanban, Gallery, Calendar, Timeline) plus the Record Drawer" — the Drawer is a sub-component, not a standalone view type.

PR #811, opened 2026-04-11
feat(after-sales): refresh ticket edit flow onto current main

zensgit:
## Summary
- refresh the stale ticket edit flow onto current `main`
- preserve current refund field-policy guards and existing object update/error behavior while landing the edit path
- preserve `low` ticket priority through the edit UI and route seam, and lock the communication registration with `updateTicket`

## Verification
- `pnpm --dir /Users/huazhou/Downloads/Github/metasheet2/.worktrees/after-sales-next-major-slice-refresh --filter @metasheet/web exec vitest run tests/AfterSalesView.spec.ts --reporter=dot`
- `pnpm --dir /Users/huazhou/Downloads/Github/metasheet2/.worktrees/after-sales-next-major-slice-refresh --filter @metasheet/core-backend exec vitest run tests/unit/after-sales-plugin-routes.test.ts --reporter=dot`
- `pnpm --dir /Users/huazhou/Downloads/Github/metasheet2/.worktrees/after-sales-next-major-slice-refresh --filter @metasheet/core-backend run test:integration:after-sales`

## Notes
- this supersedes the stale local-only `feature/after-sales-next-major-slice` by re-landing its ticket edit flow on top of current `main`
- frontend verification still emits the known Vite websocket bind warning in this sandbox (`EPERM 0.0.0.0:24678`), but the suite passes

PR #813, opened 2026-04-11
fix(ci): unblock plugin system tests baseline

zensgit:
## Summary
- bypass `rbacGuard` in `approvals-routes` unit tests so the suite continues to exercise route-specific approval error handling
- align approval assignment upserts with the new active-assignment unique index introduced on `main`
- unblock baseline `Plugin System Tests` failures that are currently also blocking after-sales PR #811

## Verification
- `pnpm --dir /Users/huazhou/Downloads/Github/metasheet2/.worktrees/main-approvals-ci-check --filter @metasheet/core-backend exec vitest run tests/unit/after-sales-approval-bridge.test.ts --reporter=dot`
- confirmed current `main` is already red on the same `Plugin System Tests` workflow and identical `after-sales integration` failures

## Notes
- `tests/unit/approvals-routes.test.ts` cannot be executed reliably inside this sandbox because `supertest` attempts to bind an ephemeral listener and hits `EPERM`; CI is the real verifier for that file
- this PR is intended to unblock the baseline so feature PR #811 can be evaluated on its own diff

### DIFF
--- a/packages/core-backend/src/services/AfterSalesApprovalBridgeService.ts
+++ b/packages/core-backend/src/services/AfterSalesApprovalBridgeService.ts
@@ -414,8 +414,13 @@ export class AfterSalesApprovalBridgeService {
           `INSERT INTO approval_assignments
            (instance_id, assignment_type, assignee_id, source_step, is_active, metadata)
            VALUES ($1, 'role', $2, $3, TRUE, $4::jsonb)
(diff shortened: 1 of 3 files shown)


zensgit:
Closing this fallback CI-fix PR as superseded. Current main has already recovered: the latest Plugin System Tests run for merge commit 7c1f61117 (run 24276802830) completed successfully after PR #811 merged, so this branch is no longer needed to unblock after-sales work.

PR #815, opened 2026-04-11
fix(migrations): restore approval migration ordering

zensgit:
## Summary
- rename the approval template migration so it sorts after the already-released user namespace admission migration
- preserve the migration contents; this is an ordering-only hotfix

## Why
Production databases that already executed zzzz20260411120000_create_user_namespace_admissions fail later deploys because the approval template migration currently sorts into the same timestamp slot before it.

## Verification
- confirmed the production deploy was blocked by Kysely migration ordering on 2026-04-11
- verified the corrected order is:
  - zzzz20260411120000_create_user_namespace_admissions
  - zzzz20260411120100_approval_templates_and_instance_extensions
  - zzzz20260411123000_add_created_action_to_approval_records
- production release was patched with the same ordering and the pending migrations then executed successfully


zensgit:
Superseded by #814, which is already merged into main. Closing this duplicate hotfix PR to avoid a second migration-order patch on top of the same sequence change.

zensgit:
Closing note: this PR is superseded by main. The migration-order fix was already merged through #814, and keeping this branch open would introduce a duplicate approval migration in the merge result.

PR #958, opened 2026-04-20
feat(collab): wire Yjs opt-in into grid text cells behind build flag

zensgit:
## Summary

- Wires `useYjsDocument` + `useYjsTextField` into `MetaCellEditor` for grid text cells, behind the build-time flag `VITE_ENABLE_YJS_COLLAB=true` (off by default). Closes audit PR #944 gap #4 (Yjs frontend integration).
- Adds a new composable `useYjsCellBinding` that gates the Yjs call path, enforces a 2.5s connect timeout, and falls back gracefully to the existing REST submit path on flag-off, timeout, error, or mid-session disconnect. No user-visible error and no swallowed edits on failure.
- Scope is deliberately narrow: only `string` non-date-like cells, only grid view, only cell edit (record create/delete untouched, other cell types untouched, backend untouched).

## Why this is behind a flag

The backend Yjs pipeline was validated in isolation (see `docs/operations/yjs-node-client-validation-20260420.md`), but a real two-browser concurrent edit has still not been exercised against production UX. The preflight checklist (`docs/operations/poc-preflight-checklist.md`) requires an end-to-end run with the flag on before claiming rollout-ready. This PR unblocks that validation step without committing to it.

The flag is **compile-time** (Vite env), not runtime. That lets production bundles that don't want Yjs tree-shake the whole call path out.

## How to enable it for testing

```bash
VITE_ENABLE_YJS_COLLAB=true pnpm --filter @metasheet/web build
# or for dev
VITE_ENABLE_YJS_COLLAB=true pnpm --filter @metasheet/web dev
```

Verification path after enabling:

1. Open a grid text cell, start editing.
2. Devtools Network panel: look for Socket.IO `/yjs` handshake.
3. `GET /api/admin/yjs/status` - `activeDocCount` should increment while the editor is focused and decrement on blur.
4. In a second browser editing the same field: `MetaYjsPresenceChip` should appear next to the input on both sides.

## What a reviewer should specifically look at

1. **Dual-write suppression** (`apps/web/src/multitable/components/MetaGridTable.vue`): `yjsHandledCellKey` is set when `MetaCellEditor` emits `yjs-commit`, and `confirmEdit` skips the REST `patch-cell` emit for that exact cell. If the key comparison is broken, every Yjs edit will also fire a REST patch (not corruption - idempotent - but defeats the purpose).
2. **Flag gate position** (`useYjsCellBinding.ts`): the flag check runs BEFORE `useYjsDocument()` is invoked. If anyone refactors this so the flag check sits inside `useYjsDocument`, the composable's `watch(recordId, ..., { immediate: true })` will still fire `io('/yjs')` on mount. The "flag off - no io() call" test is the regression guard.
3. **Focus / blur lifecycle**: composable ties to parent lifecycle via `onUnmounted(release)`. Because the cell editor is `v-if`'d in/out by `isEditing`, teardown runs on every cell blur. If the editor is later refactored to stay mounted, call `release` explicitly to avoid socket leaks.
4. **Character-level merge is NOT shipped here.** `useYjsTextField.setText` does `delete(0, length); insert(0, new)` per keystroke - last-write-wins, not per-character CRDT merge. That's fine for the POC but reviewers should know two simultaneous editors will see replacement, not merge. A follow-up PR should diff old/new and emit `insertAt`/`deleteRange` for real CRDT semantics.
5. **Date-like strings stay on REST** - `fieldIdRef` returns null for date-like cells, so Yjs never engages there.

Full reviewer-attention list and fallback contract: `docs/operations/yjs-frontend-opt-in-wired-20260420.md`.

## Relation to audit PR #944

This is gap #4 from that audit. Gap #1 (Dashboard) and gap #2 (Automation logs) already landed as separate PRs. Gap #3 (Field Validation Panel) is in flight on another branch.

## Test plan

- [x] `apps/web/tests/multitable-yjs-cell-binding.spec.ts` (+4 tests, all passing)
  - [x] Flag off -> `ioMock` never called, `active` stays false
  - [x] Flag set to `"1"` (not exactly `"true"`) -> `ioMock` never called
  - [x] Flag on + synced -> Y.Text drives input, `yjs:update` emitted on change
  - [x] Flag on + timeout -> disconnect + fallback, no further Yjs emits
- [x] `apps/web/tests/yjs-awareness-presence.spec.ts` still passes (no regressions in presence composable)
- [x] Grid/cell editor tests still pass (`multitable-grid-attachment-editor`, `multitable-grid-field-comment`, `multitable-grid-link-renderer`)
- [x] `vue-tsc` clean
- [x] Flag OFF behavior byte-identical to pre-PR for all other cell types (number, date, select, boolean, link, attachment) - no code paths touched

Baseline before this commit: 1757 tests, 1657 passing, 100 pre-existing failures (e.g. `platform-app-shell.spec.ts` - unrelated). With this commit: +4 new tests, all passing, zero regressions.

Follow-ups explicitly NOT in this PR:
- Real two-browser session with the flag on, captured as video + `/api/admin/yjs/status` samples.
- Character-level merge via diff-to-ops translation.
- Decision on whether to expose as runtime per-tenant feature flag.

Audit PR: #944.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/apps/web/src/multitable/components/MetaFieldManager.vue
+++ b/apps/web/src/multitable/components/MetaFieldManager.vue
@@ -173,6 +173,15 @@
           </div>
         </template>
 
+        <MetaFieldValidationPanel
+          v-if="configTarget && validationPanelVisible"
+          class="meta-field-mgr__validation"
+          :field-id="configTarget.id"
+          :field-type="validationPanelFieldType"
+          :rules="validationDraft"
+          :options="validationPanelOptions"
+          @update:rules="onValidationRulesChange"
+        />
         <div v-if="fieldConfigError" class="meta-field-mgr__error">{{ fieldConfigError }}</div>
         <div class="meta-field-mgr__config-actions">
           <button class="meta-field-mgr__btn-cancel" @click="closeConfig">Cancel</button>
@@ -208,7 +217,7 @@
 
 <script setup lang="ts">
 import { computed, onBeforeUnmount, reactive, ref, watch } from 'vue'
-import type { MetaField, MetaFieldCreateType, MetaSheet } from '../types'
+import type { FieldValidationRule, MetaField, MetaFieldCreateType, MetaSheet } from '../types'
 import {
   normalizeStringArray,
   resolveAttachmentFieldProperty,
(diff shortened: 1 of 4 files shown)


zensgit:
Wrong branch — recreating with the correct head ref (codex/wire-yjs-text-cell-20260420).

PR #1137, opened 2026-04-24
refactor(multitable): add record post-commit hooks

zensgit:
## Summary

- Replace the direct Yjs invalidator seam in multitable record writes with a generic post-commit hook seam.
- Keep `setYjsInvalidator(...)` as a compatibility shim while wiring app boot and route construction through `setPostCommitHooks(...)`.
- Preserve existing REST/default invalidation and `source === 'yjs-bridge'` skip semantics.
- Add development and verification docs under `docs/development/`.

## Verification

- `pnpm --filter @metasheet/core-backend exec vitest run tests/unit/record-write-service.test.ts tests/unit/record-service.test.ts tests/unit/yjs-rest-invalidation.test.ts --reporter=dot` -> 46/46 passed
- `pnpm --filter @metasheet/core-backend exec vitest run tests/integration/multitable-record-patch.api.test.ts --reporter=dot` -> 6/6 passed
- `pnpm --filter @metasheet/core-backend exec tsc --noEmit --pretty false` -> passed

Docs:
- `docs/development/multitable-post-commit-hooks-development-20260424.md`
- `docs/development/multitable-post-commit-hooks-verification-20260424.md`

### DIFF
--- a/docs/development/multitable-post-commit-hooks-development-20260424.md
+++ b/docs/development/multitable-post-commit-hooks-development-20260424.md
@@ -1,0 +1,57 @@
+# Multitable Post-Commit Hooks Development - 2026-04-24
+
+## Goal
+
+Replace the direct Yjs invalidator seam in multitable record writes with a generic post-commit hook seam.
+
+This keeps REST/Yjs consistency cleanup in the authoritative write path, while avoiding a permanent dependency from `RecordWriteService` and `RecordService` to Yjs-specific infrastructure.
+
+## Scope
+
+- Added `packages/core-backend/src/multitable/post-commit-hooks.ts`.
+- Updated `RecordWriteService` to accept `RecordPostCommitHook[]`.
+- Updated `RecordService` to accept `RecordPostCommitHook[]`.
+- Kept `setYjsInvalidator(...)` as a compatibility shim.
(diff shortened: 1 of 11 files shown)


zensgit:
Review update: addressed the blocking timing issue found during review. RecordWriteService post-commit hooks now run immediately after the DB transaction resolves, before lookup/rollup recompute, summary rebuild, realtime, or eventBus fan-out. Added regression coverage for hook-before-recompute ordering and for hooks still running when a later recompute helper fails. Rebased on latest main and reran local target verification: 4 related Vitest files pass (54 tests) and core-backend tsc --noEmit passes.

zensgit:
Final review note before merge: no remaining blocking findings after the timing fix. The original concern was that RecordWriteService post-commit hooks ran after lookup/rollup recompute and summary rebuild; this is now fixed so hooks run immediately after the DB transaction resolves, before any post-transaction helper awaits or notifications. Added regression tests for hook-before-recompute ordering and for hooks still running when a later recompute helper fails. Local verification passed: 4 related Vitest files, 54 tests; core-backend tsc --noEmit; git diff --check. GitHub CI is green across required checks. Remaining non-blocking follow-ups: decide separately whether no-op RecordService.patchRecord notifications should be suppressed and whether create/delete should use the same hook seam.

PR #1168, opened 2026-04-25
fix(integration): harden K3 WISE preflight boundaries

zensgit:
## Summary
- harden K3 WISE live PoC preflight input normalization for customer GATE JSON variants
- reject truthy string Submit/Audit values instead of silently accepting unsafe intent
- normalize SQL Server mode aliases and detect schema-qualified/quoted K3 core table names before write-capable PoC modes
- add design and verification docs for the G1 boundary probe

## Boundary Probe
- 15/15 boundary cases pass after patch
- covered URL trailing slash, baseUrl fallback, environment casing, Chinese production text, string Submit/Audit flags, SQL mode casing, schema/quoted K3 core table names, similar middle-table names, and plaintext credential redaction

## Verification
- node --test scripts/ops/integration-k3wise-live-poc-preflight.test.mjs
- node scripts/ops/integration-k3wise-live-poc-preflight.mjs --print-sample > /tmp/k3wise-preflight-sample.json
- node -e "JSON.parse(require('fs').readFileSync('/tmp/k3wise-preflight-sample.json','utf8')); console.log('sample json ok')"
- git diff --check -- scripts/ops/integration-k3wise-live-poc-preflight.mjs scripts/ops/integration-k3wise-live-poc-preflight.test.mjs docs/development/integration-core-k3wise-preflight-boundary-hardening-design-20260425.md docs/development/integration-core-k3wise-preflight-boundary-hardening-verification-20260425.md
- node --check scripts/ops/integration-k3wise-live-poc-preflight.mjs && node --check scripts/ops/integration-k3wise-live-poc-preflight.test.mjs

PR #1169, opened 2026-04-25
fix(integration): coerce remaining K3 WISE preflight boolean strings

zensgit:
## Summary

Follow-up to PR #1168 closing 3 same-class boolean-string bugs that the boundary hardening PR missed, plus extending numeric coercion (0/1 from spreadsheet exports).

**Scale**: +216/-4, 4 files (1 code, 1 test, 2 docs), 1 commit.

## What this PR fixes

PR #1168 introduced `normalizeSafeBoolean()` and applied it to `k3Wise.autoSubmit` and `k3Wise.autoAudit`. Reviewer audit found **the same `=== true` strict-equality pattern still in use at three additional customer-supplied boolean fields**:

| Site | Original code | Risk before this PR |
|---|---|---|
| `sqlServer.enabled === true` | line 212 | **Most severe.** Customer types `enabled: "true"` → script reads as falsy → `sqlEnabled = false` → entire SQL Server validation block (allowedTables guard, mode check, writeCoreTables guard) is silently skipped while customer believes they enabled it. |
| `sqlServer.writeCoreTables === true` | line 216 | Even with `sqlEnabled` correctly true, `writeCoreTables: "true"` would be read as falsy → core-table writes not blocked. |
| `bom.enabled === true` | line 223 | `bom.enabled: "true"` read as falsy → `bom.productId` requirement check skipped → BOM PoC runs without product scope. |

Same fix mechanism as PR #1168: route through `normalizeSafeBoolean()`, persist canonical boolean back into the returned `gate` object so downstream code reads true booleans.

## Number support extension

Customer JSON often comes from spreadsheet exports where booleans are 0/1. Extended `normalizeSafeBoolean()` to handle:
- `value === 1` → `true`
- `value === 0` → `false`
- Any other finite number → throws `\`${field} must be 0 or 1 when given as a number\``
- `NaN`/`Infinity` → throws `\`${field} must be a finite boolean, 0/1, or boolean-like string\``

## Verification

- [x] `node --test scripts/ops/integration-k3wise-live-poc-preflight.test.mjs` → **13/13 pass** (was 9/9, +4 new)
- [x] `node --check scripts/ops/integration-k3wise-live-poc-preflight.mjs` → syntax clean
- [x] `git diff --check` → no whitespace issues
- [x] All 9 existing tests from #1168 unchanged and still pass — no regression
- [x] Manual grep: no `=== true` checks remaining on customer-supplied boolean fields. Two remaining `gate.bom.enabled === true` reads (lines 415, 462) are post-normalization and now safe.

## New test cases

1. `buildPacket coerces sqlServer.enabled "true" string and still applies allowedTables guard`
2. `buildPacket coerces sqlServer.writeCoreTables "true" string`
3. `buildPacket coerces bom.enabled "true" string and enforces productId requirement`
4. `buildPacket accepts numeric 0/1 for boolean flags but rejects other numbers`

## Out of scope

- **Reviewer item #2 (UX)**: `mode: 'disabled'` while `enabled: true` produces the generic "must be readonly, middle-table, or stored-procedure" error. Clearer guidance ("set `enabled: false` instead") deferred to a separate minor PR.
- Same-class audit on `integration-k3wise-live-poc-evidence.mjs` — that script consumes pre-validated PoC output, no customer-side boolean inputs.

## Why this matters now

K3 WISE GATE email already sent to customer. Their JSON answer can land any time. Without this fix, a customer who types `"true"` instead of `true` (very common — JSON-by-hand or spreadsheet export) would unknowingly bypass safety checks. This PR closes that window before the answer arrives.

### DIFF
--- a/docs/development/integration-core-k3wise-preflight-bool-coercion-design-20260425.md
+++ b/docs/development/integration-core-k3wise-preflight-bool-coercion-design-20260425.md
@@ -1,0 +1,50 @@
+# K3 WISE Preflight Boolean Coercion Sweep · Design
+
+> Date: 2026-04-25
+> Follow-up to: PR #1168 (boundary hardening)
+> Scope: Same-class bug audit on the K3 WISE live PoC preflight script
+
+## Problem
+
+PR #1168 introduced `normalizeSafeBoolean()` to fix a class of bugs where customer JSON containing string `"true"` / `"yes"` / `"是"` would slip past `=== true` strict equality checks for `k3Wise.autoSubmit` and `k3Wise.autoAudit` — making the customer think Save-only was on while the script actually treated those flags as not-true.
+
+Reviewer audit after the merge found **the same bug pattern still exists at three additional sites** in the same script:
+
+| Site | Original code | Risk if customer types `"true"` instead of `true` |
+|---|---|---|
+| `sqlServer.enabled === true` | line 212 | `sqlEnabled = false` → **entire SQL Server validation block is skipped**, including the K3 core-table guard. Customer thinks SQL channel is enabled with safety; script skipped all validation. |
+| `sqlServer.writeCoreTables === true` | line 216 | Even with `sqlEnabled` correctly true, `"true"` here would be read as falsy → core-table writes would not be blocked. |
+| `bom.enabled === true` | line 223 | `bomEnabled = false` → `bom.productId` requirement check does not fire → BOM PoC runs without product scope. |
+
+This is the most severe of the three (#1) because it cascades — disabling the entire SQL Server validation chain when the customer believed they had enabled it.
+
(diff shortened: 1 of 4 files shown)


zensgit:
Review note before merge: no blocking findings. I reviewed the remaining K3 WISE preflight boolean coercion patch and verified the risky customer-input sites are now normalized: sqlServer.enabled, sqlServer.writeCoreTables, and bom.enabled. Local verification passed (13/13 preflight tests, syntax check, diff whitespace check), and GitHub CI is green across contracts, pr-validate, after-sales integration, Node 18, Node 20, and coverage. GitHub cannot accept an approval from the same author account, so this comment records the review outcome before admin merge.

PR #1171, opened 2026-04-25
feat(approval): add SLA breach notification channels

zensgit:
## Summary

- Closes the WP5 SLA observability loop — wires `ApprovalSlaScheduler.onBreach` to DingTalk + email-stub channels
- Channels are **env-gated**: register only when env vars configured (no per-breach failure noise on default deployments)
- Failure isolation: a channel throwing does NOT block other channels; notifier throwing does NOT block scheduler tick
- DingTalk reuses leaf helpers from `integrations/dingtalk/robot.ts`; email is a logging stub (no SMTP/SendGrid in repo and task spec forbade new deps)

## Verification

- `npx tsc --noEmit` → clean
- `approval-breach-notifier`: 8/8 ✓
- `dingtalk-breach-channel`: 6/6 ✓
- `approval-sla-scheduler`: 7/7 ✓ (regression green)
- `approval-metrics-service`: 15/15 ✓ (existing coverage green after JOIN method added)
- Aggregate: 24/24 focused (post-review) at HEAD `50179d9d7` on `origin/main@892ed2f9c`

## Risk / Rollback

- Risk: **best-effort semantics** — failed-channel breaches do NOT retry on next scheduler tick (filtered by `sla_breached=TRUE`). Explicitly documented in dev MD; persistent `breach_notified_at` column is the planned follow-up.
- Risk: in-memory FIFO `Set<instanceId>` (bounded 5000) for in-process dedupe — bounded under steady state but cannot defend against rapid leader churn until persistent column lands.
- Rollback: revert the 3 commits; scheduler reverts to no-op `onBreach` behavior. No migration to undo.

## Follow-up

- Persistent `breach_notified_at` column for cross-restart / leader-takeover dedupe (small migration + WHERE-clause filter, ~60 LoC).
- Real SMTP transport (today: logging stub) — pending dep policy decision on `nodemailer` / a managed sender.
- Live DingTalk integration test (today: mock-only); manual `curl` verification path documented in verification MD.

See `docs/development/approval-sla-breach-notify-development-20260425.md` and `*-verification-20260425.md` for full design + verification.

---

## Follow-up branch ready (post-merge action)

The `breach_notified_at` persistent-dedupe follow-up listed above is **already implemented** on a stacked branch and verified — held off opening as a stacked PR per queue-drain discipline. To be opened as a normal PR (not Draft) once this PR (#1171) merges:

- Branch: `codex/approval-sla-breach-notified-at-20260426`
- HEAD: `3ffda7a1f`
[...]

PR #1174, opened 2026-04-25
feat(approval): persist breach_notified_at + restart-resilient retry

zensgit:
## Summary

Closes the persistent-column follow-up named in PR #1171's commit message and code comment.

**Scale**: +507/-10 across 8 files (1 migration, 4 code, 2 test, 2 docs), 1 commit.

## Problem

PR #1171 ships `ApprovalBreachNotifier` with **in-memory only** dedupe. Failure mode:

1. Scheduler atomic flip flags row X as `sla_breached = TRUE` and returns it.
2. Notifier dispatches X via `channel.send` (succeeds at the wire).
3. Leader process crashes **before** `markNotified(X)` runs.
4. Restart: `notifiedIds` is empty.
5. `checkSlaBreaches`'s WHERE filters `sla_breached = FALSE` — X never re-surfaces.
6. **X is permanently un-notified-in-the-record but actually was notified.** Operations cannot answer "did we send for X?" without log archaeology, and any failed dispatch has no retry path.

PR #1171's commit message correctly noted this **does not produce duplicates** (atomic UPDATE prevents re-entry). The cost is **missed escalations stay missed** — for an SLA system, that's a real customer-impact gap.

## Fix

| Layer | Change |
|---|---|
| Migration `058_add_approval_breach_notified_at.sql` | `ALTER TABLE approval_metrics ADD COLUMN breach_notified_at TIMESTAMPTZ NULL` + partial index on `instance_id WHERE sla_breached = TRUE AND breach_notified_at IS NULL` |
| `ApprovalMetricsService.markBreachNotified(id, now)` | Guarded UPDATE — sets timestamp only if NULL, so retry calls are safe no-ops |
| `ApprovalMetricsService.findUnnotifiedBreaches(limit)` | The recovery query, ordered oldest-first, capped at 5000 |
| `ApprovalBreachNotifier.markNotified` | Now `async`; persists after in-memory mark; persist failure logs and continues (in-memory dedupe still applies, future restart reconciles) |
| `ApprovalBreachNotifier.notifyMissedBreaches()` | New: pulls `findUnnotifiedBreaches` and routes through existing `notifyBreaches` pipeline. Idempotent. |
| `packages/core-backend/src/index.ts` startup | Fires `notifyMissedBreaches()` once when SLA scheduler initializes — fire-and-forget; failure logs and continues |

## Verification

- [x] `pnpm vitest run tests/unit/approval-metrics-service.test.ts tests/unit/approval-breach-notifier.test.ts` — **35/35 pass** (was 23 before; +12 new)
  - 6 new on `approval-metrics-service`: `markBreachNotified` guards (3) + `findUnnotifiedBreaches` ordering/cap/fallback (3)
  - 6 new on `approval-breach-notifier`: persistence-on-success / no-persist-on-fail / persist-failure-non-fatal / replay-happy / replay-empty / replay-DB-failure
- [x] `git diff --check` — clean
- [x] No regression in existing 8 + 15 unit tests across the two files

## Design decisions intentionally minimal

- **Persistent-failure non-fatal**: DB UPDATE rejection logs and continues. In-memory dedupe still prevents same-process re-send; next leader's `notifyMissedBreaches` reconciles. Avoids cascading failure when DB has transient issues.
- **Startup retry fire-and-forget**: `void breachNotifier.notifyMissedBreaches().then().catch()` — scheduler init never awaits it. Best-effort recovery cannot block startup.
- **Cap at 5000**: A long-paused leader returning could trigger an unbounded query. Hard cap; operator can re-invoke for larger backlogs (manual triage scenario).
- **No backfill of pre-existing breached rows**: First post-deploy leader will trigger a one-time bounded replay burst — expected and documented in the design doc.
- **No per-channel persistence**: We record "at least one channel succeeded" via `anySent`. Per-channel ack tracking would require a join table; out of scope for this PR.

## Out of scope

- Per-channel ack tracking (DingTalk vs email separately)
- Re-notification on requester / template metadata change
- Multi-leader coordination (relies on PR #1160 leader lock)
- Pagination beyond the 5000 cap (caller can re-invoke if needed)

## Cross-references

- PR #1171 (the predecessor): merged 2026-04-25 as commit `58862b394`. Code comment at `ApprovalBreachNotifier.ts` lines 17-24 named this exact follow-up.
- Migration sequence: most recent on `origin/main` was `057_create_integration_core_tables.sql` (#1140); this PR claims `058`.

### DIFF
--- a/docs/development/approval-breach-notified-persistence-design-20260426.md
+++ b/docs/development/approval-breach-notified-persistence-design-20260426.md
@@ -1,0 +1,71 @@
+# Approval Breach Notification Persistence · Design
+
+> Date: 2026-04-26
+> Follow-up to: PR #1171 (`feat(approval): add SLA breach notification channels`)
+> Closes: the "persistent breach_notified_at column" follow-up acknowledged in PR #1171's commit message and code comment.
+
+## Problem
+
+PR #1171 ships `ApprovalBreachNotifier` with **in-memory only** dedupe — a `Set<instanceId>` of already-notified rows on the leader process. The commit message and class doc-comment both flag this as a known limitation:
+
+> `A persistent breach_notified_at column is tracked as a follow-up.`
+
+The concrete failure mode this gap leaves open:
+
+1. SLA scheduler atomic flip (`UPDATE approval_metrics SET sla_breached = TRUE WHERE sla_breached = FALSE RETURNING instance_id`) returns row X.
+2. Notifier dispatches X via channel.send (succeeds at the wire).
+3. Leader process crashes **before** `markNotified(X)` runs.
+4. Restart: in-memory `notifiedIds` is empty.
+5. `checkSlaBreaches` will never return X again (its WHERE filters `sla_breached = FALSE` and X is already TRUE).
+6. **X is now permanently un-notified-in-the-record but actually was notified**, producing an inconsistency — operations cannot answer "did we send for X?" without log archaeology.
+
+A more concerning variant: dispatch failed, leader crashed before retry, X never gets retried — the notifier had no concept of "missed breach to retry" because the in-memory mark was the only state.
+
+PR #1171's commit message correctly noted that this **does not produce duplicate notifications** (the `sla_breached = FALSE` WHERE atomically excludes already-flagged rows from re-entering the pipeline). The cost is **missed dispatches stay missed**, not duplicates — but for an SLA system, missed escalations are a real customer-impact issue.
+
+## Solution
+
+Add the persistent column the previous PR's comment promised, plus the matching machinery to use it:
+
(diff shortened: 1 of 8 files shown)


zensgit:
Closing as superseded by #1172.

#1172 has already merged into main as `f8db1821cddcaa3b7ad8eebcb47c2eb22a635799` and covers the complete persistent breach notification dedupe flow:

- Kysely migration for `breach_notified_at`
- scheduler retry-pending union dispatch
- successful notification marking
- focused tests and verification docs

This PR is an earlier parallel implementation, now conflicts with main and has failing checks. Keeping it open would duplicate the migration surface and preserve an incomplete retry model.

PR #1182, opened 2026-04-26
fix(integration): coerce K3 WISE evidence packet safety hand-edits

zensgit:
> **Closes the audit series**: this is the 4th and final named deferred item from PR #1175's design doc. After this merges, both customer-facing K3 WISE Live PoC scripts (\`preflight.mjs\` + \`evidence.mjs\`) are fully hardened against realistic edge cases.

## Summary
\`requirePacketSafety()\` enforces the live-PoC safety contract via 4 strict-equality checks. In the normal flow, preflight emits real boolean values and strict equality is safe. The hand-edit edge case: during incident response, an operator may re-run the evidence compiler against a hand-edited packet (tweaked one field for a re-test, or copy-pasted from a markdown table that serializes bools as strings). Hand-edited \`saveOnly: \"true\"\` (string) currently false-fails with a confusing error.

## Approach
Coerce all 4 safety fields (\`saveOnly\`, \`autoSubmit\`, \`autoAudit\`, \`productionWriteBlocked\`) through the existing \`normalizeSafeBoolean\` helper from PR #1175. **Predicate semantics unchanged**:
- \`saveOnly\` must be truthy (string \`\"false\"\` / numeric \`0\` / Chinese \`\"否\"\` still correctly fail the guard)
- \`autoSubmit\` / \`autoAudit\` must be falsy (string \`\"true\"\` / numeric \`1\` / \`\"是\"\` still correctly fail)
- \`productionWriteBlocked\` must be truthy

Coercion only widens the **input surface**, not the **predicate**. There is no input value that previously failed and now passes where the underlying intent was to bypass safety.

## Files changed
- \`scripts/ops/integration-k3wise-live-poc-evidence.mjs\` — \`requirePacketSafety\` uses \`normalizeSafeBoolean\` for all 4 safety fields (~15 lines, +inline comment)
- \`scripts/ops/integration-k3wise-live-poc-evidence.test.mjs\` — 8 new test cases (~89 lines)
- \`docs/development/integration-core-k3wise-evidence-packet-safety-coercion-design-20260426.md\`
- \`docs/development/integration-core-k3wise-evidence-packet-safety-coercion-verification-20260426.md\`

## Test plan
- [x] \`node --test scripts/ops/integration-k3wise-live-poc-evidence.test.mjs\` reports **31/31 pass** (was 23/23, +8 new)
- [x] Hand-edited string \`\"true\"\` / numeric \`1\` / Chinese \`\"是\"\` for safety fields are accepted
[...]

PR #1187, opened 2026-04-26
fix(integration): guard against concurrent pipeline runs and stale run leaks

zensgit:
## Problem

Two bugs in the same invariant — the `running` status of a pipeline run must be exclusive per pipeline and bounded in lifetime:

**Bug 1 — Concurrent run guard missing**: Two simultaneous `POST /pipelines/:id/run` requests both passed `createPipelineRun`. Both reads started from the same watermark baseline; both advanced the watermark to the same point; both wrote to the target ERP (idempotency blocked duplicates but the double watermark advance silently discarded work that should have been queued for the next incremental run).

**Bug 2 — Stuck-run / status leak**: A process crash between `startRun` and `finishRun`/`failRun` left a run permanently `status='running'`. Once the concurrent-run guard was added, this became a deadlock: no new runs could ever start for that pipeline, with no automatic recovery path.

### `pipelines.cjs`
- Add `PipelineConflictError` error class
- In `createPipelineRun`: query for existing `running` runs on the same `(tenant_id, workspace_id, pipeline_id)` tuple before inserting; throw `PipelineConflictError` with the blocking run's ID if one exists
- Add `abandonStaleRuns({ tenantId, workspaceId, [pipelineId], [olderThanMs] })`: selects all `running` runs, filters those whose `started_at` predates the threshold (default 4h), marks them `failed` with an `error_summary` explaining why — call this on plugin startup or before launching a new run if you suspect a crash

### `http-routes.cjs`
- `inferHttpStatus`: add `Conflict` → 409 so `PipelineConflictError` returns HTTP 409 instead of 500

## Tests added (7 new scenarios)

**`pipelines.test.cjs`** (+5):
1. concurrent run rejected with `PipelineConflictError`, includes blocking run ID in details
2. terminated run (succeeded) does not block new run
3. running run on a different pipeline does not block this pipeline
4. `abandonStaleRuns` abandons runs older than threshold, leaves fresh and other-tenant runs untouched
5. `abandonStaleRuns` with custom `olderThanMs` threshold

**`http-routes.test.cjs`** (+1 in testErrorResponseShape):
6. `PipelineConflictError` thrown by runner → HTTP 409 with error body

All 18 integration-core test files pass.

## Out of scope

- Distributed lock (advisory PG lock, Redis) — the equality-based guard is sufficient for single-node PoC; if multi-node scale-out is needed later, replace `select + insert` with `SELECT ... FOR UPDATE SKIP LOCKED` in a real DB transaction
- Scheduler-triggered auto-abandon on startup — `abandonStaleRuns` is exported; caller decides when to invoke it

## Cross-references
- Broader-surface audit: `docs/development/bool-coercion-audit-broader-surface-20260426.md` § "What this means for the M2 GATE wait"
- Live PoC evidence compiler: PR #1182 (packet safety coercion, same module)

### DIFF
--- a/docs/development/integration-core-concurrent-run-guard-design-20260426.md
+++ b/docs/development/integration-core-concurrent-run-guard-design-20260426.md
@@ -70,13 +70,46 @@ if (runningRows.length > 0) {
 **Why here and not in `runPipeline`:** `createPipelineRun` is the DB-authoritative
 gate. Checking in `runPipeline` before calling `startRun` would have a TOCTOU
 window — two callers check simultaneously, both see no running run, both insert.
-This PR places the guard at the insert point and wraps the check+insert critical
-section in an in-process `(tenantId, workspaceId, pipelineId)` lock. That closes
-the async race for the single-node PoC runtime while keeping the invariant owned
-by `createPipelineRun`, where the `disabled` pipeline check also lives. A true
-distributed lock (advisory PG lock, `SELECT ... FOR UPDATE SKIP LOCKED`, or a
-partial unique index) would close the window across multiple Node processes and
-is left for production hardening.
+This PR places the friendly guard at the insert point and wraps the check+insert
+critical section in an in-process `(tenantId, workspaceId, pipelineId)` lock.
+That closes the async race for the single-node PoC runtime while keeping the
+invariant owned by `createPipelineRun`, where the `disabled` pipeline check also
+lives.
+
+### DB-authoritative partial unique index (migration 058)
+
+The in-process lock is not enough for a multi-node deployment: two separate Node
+processes can still both snapshot "no running row" before either insert commits.
+Migration 058 makes the invariant database-authoritative:
+
+```sql
+WITH duplicate_running AS (...)
+UPDATE integration_runs
+SET status = 'failed',
+    finished_at = COALESCE(finished_at, NOW()),
+    error_summary = COALESCE(error_summary, 'abandoned: duplicate running run closed before unique guard migration')
+WHERE id IN (SELECT id FROM duplicate_running WHERE duplicate_rank > 1);
+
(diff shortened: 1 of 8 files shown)


zensgit:
Follow-up pushed: upgraded the concurrent-run guard from single-process serialization to DB-authoritative enforcement.

Changes in bb1e78252:
- Added migration 058 with a partial unique index: uniq_integration_runs_one_running_per_pipeline on (tenant_id, COALESCE(workspace_id, ''), pipeline_id) WHERE status='running'.
- Added pre-index cleanup that fails duplicate running rows before creating the unique index, so migration does not fail on environments where the old bug already produced duplicates.
- Normalized Postgres 23505 on that constraint into PipelineConflictError, preserving the existing 409-ready API shape.
- Added tests for DB unique-conflict normalization and 058 SQL structure.
- Updated design and verification docs, including a real local Postgres smoke: duplicate running rows reduced to 1 running + 1 failed, and a new duplicate insert is blocked by unique_violation.

Local validation:
- node plugins/plugin-integration-core/__tests__/pipelines.test.cjs
- node plugins/plugin-integration-core/__tests__/http-routes.test.cjs
- Real Postgres throwaway cluster smoke for migrations 057+058
- git diff --check

Note: pnpm -F plugin-integration-core test was attempted in the temporary worktree but could not run there because the worktree has no node_modules and this machine defaults to Node v24.14.1; CI Node 18/20 should run the package script normally.

zensgit:
CI follow-up for bb1e78252: all required checks are green on the fresh commit.

Passed:
- migration-replay
- e2e
- pr-validate
- test (18.x)
- test (20.x)
- contracts strict/dashboard/openapi
- telemetry-plugin
- core-backend-cache
- after-sales integration
- coverage

Skipped as expected: Strict E2E with Enhanced Gates.

PR #1188, opened 2026-04-26
fix(integration): auto-sweep stale runs before starting a new pipeline run

zensgit:
## Problem

PR #1187 added `abandonStaleRuns` to the pipeline registry and exported it — but nothing called it automatically. A process crash between `startRun` and `finishRun/failRun` would leave a run permanently `status='running'`. Once the concurrent-run guard from #1187 is active, this becomes a silent deadlock: every subsequent trigger of that pipeline returns 409 CONFLICT with no way to recover short of a manual DB update.

## Fix

`runPipeline` now calls `pipelineRegistry.abandonStaleRuns` scoped to the triggering pipeline before handing off to `runLogger.startRun`. The call is defensive (`typeof check`) — existing registries without `abandonStaleRuns` (pre-#1187 or test mocks) are unaffected.

**Order after this change:**
1. `loadPipelineContext` — validates pipeline exists and is active
2. **`abandonStaleRuns`** — marks crashed/stuck running runs as failed (auto-recovery)
3. `runLogger.startRun` → `createPipelineRun` — concurrent-run guard now sees a clean slate

An operator who triggers a run after a previous crash gets automatic recovery on the first attempt rather than a confusing 409 CONFLICT followed by a manual support ticket.

## Tests added (+2 scenarios in `pipeline-runner.test.cjs`)

| # | Scenario | What it pins |
|---|---|---|
| 17a | `abandonStaleRuns` called once per `runPipeline`, scoped to correct tenant + pipeline | Verifies the autowire actually fires |
| 17b | `runPipeline` works when registry has no `abandonStaleRuns` method | Backward-compatibility: pre-#1187 registries are unaffected |

All 18 integration-core test files pass (0 regressions).

## Note

This PR is useful standalone but its full effect is only visible once #1187 is merged (before #1187, there is no concurrent-run guard to unlock). Both PRs are independent — merging this before #1187 is safe.

### DIFF
--- a/docs/development/integration-core-stale-run-autowire-design-20260426.md
+++ b/docs/development/integration-core-stale-run-autowire-design-20260426.md
@@ -1,0 +1,82 @@
+# Integration-Core Stale-Run Autowire · Design
+
+> Date: 2026-04-26
+> PR: #1188
+> Depends on: PR #1187 (abandons the stale run; this PR wires the call)
+
+## Problem
+
+PR #1187 exported `pipelineRegistry.abandonStaleRuns` but left it uncalled. The tool
+exists; no caller invokes it automatically. Result: a process crash between `startRun`
+and `finishRun`/`failRun` leaves a run permanently `status='running'`. Once #1187's
+concurrent-run guard is active, this becomes a silent deadlock — every subsequent trigger
+of that pipeline returns HTTP 409 CONFLICT with `runningRunId` pointing at a ghost run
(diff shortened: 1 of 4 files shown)


zensgit:
Closing as superseded by #1197. #1197 keeps the same stale-run recovery wiring but makes abandonStaleRuns best-effort, so transient cleanup failures do not block the main pipeline run. #1187 is now merged and provides the DB-authoritative concurrent-run guard that #1197 builds on.

PR #1189, opened 2026-04-26
perf(integration): composite index for concurrent-run guard and stale-run queries

zensgit:
## Problem

PR #1187 added a concurrent-run guard that queries:
```sql
SELECT * FROM integration_runs
WHERE tenant_id=? AND workspace_id=? AND pipeline_id=? AND status='running'
LIMIT 1
```

`abandonStaleRuns` (also #1187) queries the same columns without `LIMIT 1`.

The existing indexes are three single-column indexes (`scope` on tenant+workspace, `pipeline` on pipeline_id, `status` on status). The planner must merge-join or nested-loop across all three, which is O(N) per pipeline trigger at scale.

## Fix

Migration 058 adds:
```sql
CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_integration_runs_tenant_pipeline_status
  ON integration_runs (tenant_id, pipeline_id, status);
```

With this index:
- The concurrent-run guard becomes a single index scan → LIMIT 1 short-circuit (O(1))
- `abandonStaleRuns` scans only `running` rows for the relevant tenant+pipeline
- `listPipelineRuns` with `pipelineId + status` filter also benefits

`CONCURRENTLY` means no table lock in production during the migration.

## No code changes

This is a migration-only PR. No application logic changed.

## CI note

The `migration-sql.test.cjs` validates only 057 (the original MVP migration). The 058 migration is tested by the CI `migration-replay` job which replays all migrations against a real Postgres instance.


zensgit:
Refreshed after #1187/#1197 merged.

Changes in eb3799f56:
- Renamed the migration from 058 to 059 because #1187 now owns 058 with the DB-authoritative running-run unique index.
- Retargeted the index from the old guard-focused shape to run-history/pipeline-scoped lookup performance:
  - idx_integration_runs_scope_pipeline_status_created_at on (tenant_id, workspace_id, pipeline_id, status, created_at DESC)
- Updated design/verification docs to clarify correctness vs performance split: #1187/058 enforces one running row; this PR only improves lookup/history performance.
- Extended migration-sql.test.cjs to validate 059.

Local validation:
- Real local Postgres smoke applying 057 + 058 + 059 and verifying both indexes exist
- git diff --check

PR #1197, opened 2026-04-26
fix(integration): call abandonStaleRuns as best-effort before each pipeline run

zensgit:
## Summary

- Wires `pipelineRegistry.abandonStaleRuns()` into `runPipeline` to recover stuck `running` rows from previous process crashes before starting a new run.
- Wraps the call in try-catch: a transient DB failure during stale-run cleanup must never block the main pipeline execution — this is a best-effort housekeeping step, not a hard prerequisite.
- **Supersedes PR #1188** which added the call without error protection. If #1188 merges before this PR, the effective behavior difference is the try-catch guard; the two PRs will conflict and this one should be used.

## Test plan

- [ ] **17a**: `abandonStaleRuns` is called once per `runPipeline` invocation with correct `tenantId`/`pipelineId`.
- [ ] **17b**: `abandonStaleRuns` throws `'DB connection lost'` → pipeline run still succeeds, `rowsRead === 1`.
- [ ] **17c**: Registry without `abandonStaleRuns` method → no `TypeError`, pipeline runs normally.
- [ ] All 18 `plugin-integration-core` test files pass (0 regressions).

### DIFF
--- a/docs/development/integration-core-stale-run-besteffort-design-20260426.md
+++ b/docs/development/integration-core-stale-run-besteffort-design-20260426.md
@@ -1,0 +1,80 @@
+# Design: Call abandonStaleRuns Best-Effort Before Pipeline Runs
+
+**PR**: #1197  
+**Date**: 2026-04-26  
+**Supersedes**: #1188  
+**Primary file**: `plugins/plugin-integration-core/lib/pipeline-runner.cjs`
+
+## Problem
+
+PR #1187 made the concurrent-run invariant real: a pipeline can have at most
+one `status='running'` row in a tenant/workspace scope. That prevents duplicate
+watermark reads and double ERP writes, but it also means a crash-stuck
+`running` row can block the next run.
+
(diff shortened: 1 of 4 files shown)


zensgit:
Refreshed after #1187 merged.

Changes now in this PR:
- Merged latest main containing #1187's DB-authoritative running-run guard.
- Kept #1197 focused on runner wiring: call pipelineRegistry.abandonStaleRuns() best-effort before startRun().
- Added design + verification docs:
  - docs/development/integration-core-stale-run-besteffort-design-20260426.md
  - docs/development/integration-core-stale-run-besteffort-verification-20260426.md

Local validation:
- node plugins/plugin-integration-core/__tests__/pipeline-runner.test.cjs
- node plugins/plugin-integration-core/__tests__/pipelines.test.cjs
- git diff --check

This supersedes #1188 because cleanup failure is intentionally non-fatal here; #1188 wired the call as a hard prerequisite.

PR #1337, opened 2026-05-06
fix(integration): reject secret-bearing K3 preflight URLs

zensgit:
## Summary
- Reject inline username/password credentials in K3 WISE and PLM endpoint URLs before generating live PoC packets
- Reject secret-like URL query keys such as token/api_key while preserving non-secret routing params
- Update K3 WISE GATE sample and fixture README so customers keep secrets in credentials fields

## Verification
- node --test scripts/ops/integration-k3wise-live-poc-preflight.test.mjs
- pnpm run verify:integration-k3wise:poc
- git diff --check origin/main..HEAD

## Docs
- docs/development/integration-k3wise-preflight-url-secret-guard-development-20260506.md
- docs/development/integration-k3wise-preflight-url-secret-guard-verification-20260506.md

### DIFF
--- a/docs/development/integration-k3wise-preflight-url-secret-guard-development-20260506.md
+++ b/docs/development/integration-k3wise-preflight-url-secret-guard-development-20260506.md
@@ -1,0 +1,48 @@
+# K3 WISE Preflight URL Secret Guard Development - 2026-05-06
+
+## Context
+
+The K3 WISE live PoC preflight packet is safe to archive only when customer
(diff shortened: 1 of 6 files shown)


zensgit:
Superseded by #1505. Current main already contains a broader K3 WISE preflight URL/free-text secret guard via assertNoSecretLikeText(); #1505 keeps the remaining customer-facing template guidance and the non-secret query-parameter compatibility assertion on top of current main.

PR #1344, opened 2026-05-06
fix(integration): hydrate saved K3 setup booleans

zensgit:
## Summary

- Normalize saved K3 WISE `autoSubmit` / `autoAudit` values when hydrating external-system config into the setup page.
- Accept backend-compatible variants such as `"true"`, `1`, and Chinese operator values like `是` / `否`.
- Default unknown or malformed saved values to unchecked so a previous system's enabled flag cannot leak into the currently selected system.

## Verification

- `pnpm --filter @metasheet/web exec vitest run --watch=false tests/k3WiseSetup.spec.ts`
- `pnpm --filter @metasheet/web type-check`
- `git diff --check`

## Docs

- `docs/development/integration-k3wise-setup-bool-hydration-development-20260506.md`
- `docs/development/integration-k3wise-setup-bool-hydration-verification-20260506.md`

### DIFF
--- a/apps/web/src/services/integration/k3WiseSetup.ts
+++ b/apps/web/src/services/integration/k3WiseSetup.ts
@@ -253,6 +253,23 @@ function parseOptionalPositiveInteger(value: string): number | undefined {
   return Number.isInteger(parsed) && parsed > 0 ? parsed : undefined
 }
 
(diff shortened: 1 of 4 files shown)


zensgit:
Superseded by #1511. Current-main refresh keeps the saved K3 setup boolean hydration fix, expands the variant test matrix, and updates verification/docs on top of current main.

PR #1451, opened 2026-05-09
feat(multitable): add grid bulk edit action

zensgit:
## Summary
- Lane C from PR #1448 — turns the existing grid selection + `patchRecords` plumbing into a user-facing bulk Set/Clear field flow.
- New `MetaBulkEditDialog.vue` (field picker filtered by new `isFieldBulkEditable` helper, MetaCellEditor for value entry, per-mode UX, error rendering).
- New bulk-bar affordances on `MetaGridTable.vue` (`Set field` + `Clear field`) gated by a `canBulkEdit` prop; existing `Delete selected` unchanged.
- New composable method `bulkPatch({ fieldId, value, recordIds })` that issues a single `patchRecords` request with `expectedVersion` per selected row and returns a forward-compatible `{ updated, failed }` shape.
- `MultitableWorkbench.vue` wires the emit, mounts the dialog, surfaces success + `VERSION_CONFLICT` errors via existing toast helpers.
- 19 new focused tests (5 bulk-bar UX + 11 dialog + 3 composable); existing `multitable-grid.spec.ts`: 35 → 38.

## Honest scope cut (called out, not buried)
**Per-record partial-success UI is scoped to a follow-up backend PR.** Backend `patchRecords` is single-transactional (`record-write-service.ts:518`) and aborts on the first `VersionConflictError` / `RecordValidationError`. The dialog already accepts a `failed[]` array in its return shape so the day backend gains partial-success, the consumer renders it without rework. Until then: a `VERSION_CONFLICT` from any selected record fails the entire batch and the dialog displays a 'modified elsewhere; reload + retry' error, satisfying *partial failures do not silently disappear* within the all-or-nothing constraint.

## Acceptance bullet → evidence
| #1448 acceptance bullet | Evidence |
|---|---|
| Bulk bar offers Set/Clear only when selected records exist | `multitable-grid-bulk-edit.spec.ts` (5 tests) |
| Field picker excludes read-only/system/derived/link/attachment/hidden fields | `multitable-bulk-edit-dialog.spec.ts` (3 picker tests inc. `fieldPermissions.readOnly` map + `canEdit=false`) |
| Value editor reuses field-specific editor behavior where practical | Dialog mounts `MetaCellEditor` for the picked field; covered indirectly by apply-with-set test |
| Apply path uses one `patchRecords` request with expected versions | Composable test *bulkPatch sends one patchRecords request with expectedVersion per selected row* |
| UI reports success / conflict counts (within backend constraints) | Dialog `error` + `resultMessage` props; `renders an error message when the parent passes a conflict / failure` |
| Partial failures do not silently disappear | Conflict propagation test + forward-compat `failed[]` shape |
| Record history / subscription side effects consistent | Backend `patchRecords` reused unchanged |
| Existing bulk delete unchanged | `multitable-grid-bulk-edit.spec.ts` asserts \`Delete selected\` still in the bar |

## K3 PoC Stage 1 Lock
- Does NOT modify \`plugins/plugin-integration-core/*\`.
- Frontend-only; no migration / OpenAPI dist / package script / env var.
- Does NOT touch DingTalk / public-form / Gantt / Hierarchy / formula / automation runtime.

## Test plan
- [x] `pnpm --filter @metasheet/web exec vue-tsc -b --noEmit` clean.
- [x] `pnpm --filter @metasheet/web exec vitest run tests/multitable-grid.spec.ts` → 38/38 (was 35; +3 new bulkPatch).
- [x] `pnpm --filter @metasheet/web exec vitest run tests/multitable-bulk-edit-dialog.spec.ts` → 11/11.
- [x] `pnpm --filter @metasheet/web exec vitest run tests/multitable-grid-bulk-edit.spec.ts` → 5/5.
- [x] `git diff --check origin/main..HEAD` clean.
- [x] Pre-existing `multitable-workbench-import-flow.spec.ts` failure (`window.localStorage.clear is not a function`) verified with `git stash` to be a happy-dom polyfill issue on `origin/main`, NOT caused by this PR.

### DIFF
--- a/apps/web/src/multitable/components/MetaBulkEditDialog.vue
+++ b/apps/web/src/multitable/components/MetaBulkEditDialog.vue
@@ -31,11 +31,17 @@
           <div v-if="mode === 'set' && selectedField" class="meta-bulk-edit__row">
             <span class="meta-bulk-edit__label">Value</span>
             <div class="meta-bulk-edit__value-wrap">
+              <!--
+                Intentionally NOT wiring @confirm to onApply: MetaCellEditor
+                fires `confirm` on select/boolean `@change` (cells/MetaCellEditor.vue:113,124),
+                which would silently submit the bulk patch the moment the
+                user opens the select dropdown or toggles a checkbox. Bulk
+                edit must require an explicit "Set value" click.
+              -->
               <MetaCellEditor
                 :field="selectedField"
                 :model-value="value"
                 @update:modelValue="value = $event"
-                @confirm="onApply"
                 @cancel="onCancel"
               />
             </div>
--- a/apps/web/src/multitable/components/MetaBulkEditDialog.vue
+++ b/apps/web/src/multitable/components/MetaBulkEditDialog.vue
@@ -1,0 +1,177 @@
+<template>
+  <Teleport to="body">
+    <div v-if="visible" class="meta-bulk-edit-overlay" @click.self="onCancel">
+      <div class="meta-bulk-edit-modal" role="dialog" :aria-label="title">
+        <div class="meta-bulk-edit__header">
+          <strong>{{ title }}</strong>
+          <button class="meta-bulk-edit__close" aria-label="Close" @click="onCancel">&times;</button>
+        </div>
+
+        <div class="meta-bulk-edit__body">
+          <p class="meta-bulk-edit__hint">{{ summary }}</p>
+
+          <label class="meta-bulk-edit__row">
+            <span class="meta-bulk-edit__label">Field</span>
+            <select
+              v-model="selectedFieldId"
+              class="meta-bulk-edit__select"
+              aria-label="Field to update"
+              @change="onFieldChange"
+            >
+              <option value="">(choose a field)</option>
+              <option v-for="field in eligibleFields" :key="field.id" :value="field.id">
+                {{ field.name }}
+              </option>
+            </select>
+          </label>
+          <p v-if="!eligibleFields.length" class="meta-bulk-edit__hint meta-bulk-edit__hint--muted">
+            No bulk-editable fields are available for the current selection.
+          </p>
+
+          <div v-if="mode === 'set' && selectedField" class="meta-bulk-edit__row">
+            <span class="meta-bulk-edit__label">Value</span>
+            <div class="meta-bulk-edit__value-wrap">
+              <MetaCellEditor
+                :field="selectedField"
+                :model-value="value"
+                @update:modelValue="value = $event"
+                @confirm="onApply"
+                @cancel="onCancel"
+              />
+            </div>
+          </div>
+          <div v-else-if="mode === 'clear' && selectedField" class="meta-bulk-edit__row">
+            <span class="meta-bulk-edit__label">Action</span>
+            <span class="meta-bulk-edit__hint">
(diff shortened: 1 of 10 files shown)


zensgit:
Blocking review for #1451. I found two issues that should be fixed before merge.

1. Bulk-edit row selection is gated by delete permission. `MultitableWorkbench.vue` only enables multi-select with `gridAllowsAnyDelete`, and `MetaGridTable.vue` disables/selects rows using `resolveRowActions(row.id).canDelete`. A user or row with `canEdit=true` and `canDelete=false` cannot select rows, so the new bulk edit action is unreachable even though `canBulkEdit` is true. Please decouple selection eligibility from delete, or make selection action-aware: bulk edit should use `canEdit`, bulk delete should use `canDelete`. Add a regression test for edit-only rows.

2. `MetaBulkEditDialog.vue` wires embedded `MetaCellEditor` `confirm` directly to `onApply`. Several editors emit `confirm` during normal value selection (`select` on change, `boolean` on change, rating on click). In a bulk edit modal this applies changes to all selected records before the user clicks the explicit `Set value` button. Please remove the editor-level auto-apply and require the modal primary button for all field types; add tests that changing select/boolean values does not submit until the primary button is clicked.

The patch response/apply path looks OK: `patchRecords` returns updated versions and `records`, and `applyPatchResult` merges them into the grid.

zensgit:
Second-pass review after e31c99a: one blocker remains.

The dialog auto-submit issue is fixed: `MetaBulkEditDialog` no longer binds editor `confirm` to `onApply`, and the added select/boolean regression tests cover it.

However, the edit-only selection bug is only partially fixed. `MultitableWorkbench.vue` now enables multi-select when `gridAllowsAnyDelete || effectiveRowActions.canEdit`, but `MetaGridTable.vue` still makes individual rows selectable only when `resolveRowActions(row.id).canDelete`:

- grouped checkbox `:disabled="!resolveRowActions(row.id).canDelete"`
- flat checkbox `:disabled="!resolveRowActions(row.id).canDelete"`
- `selectableRowIds` filters on `canDelete`
- `toggleSelectRow` returns unless `canDelete`

So a user/row with `canEdit=true` and `canDelete=false` sees multi-select enabled but cannot select rows, and bulk edit is still unreachable. Please move row selection eligibility to edit-or-delete, or make it action-aware. The regression test should mount rows with `canEdit=true, canDelete=false` and prove Set/Clear can select and emit, while Delete selected remains hidden/disabled.

zensgit:
Re-review after f62ebdf6d: no blocking findings.

The two previous blockers are addressed:

- Row selection now uses `rowAllowsAnyBulkAction(recordId)` for both checkbox disabled states, select-all eligibility, and `toggleSelectRow`, so edit-only rows are selectable.
- Bulk delete/edit emission is action-aware and filters selected IDs to deletable/editable subsets respectively.
- The dialog no longer auto-applies on embedded editor `confirm`; select/boolean regression coverage is present.

CI is green on the latest head. Non-blocking follow-up: if a selection contains zero editable rows, the Set/Clear buttons can still be visible when sheet-level `canBulkEdit` is true, but clicking them emits nothing. That is acceptable for this PR and can be folded into the planned partial-success/per-row-result UX follow-up.

PR #1648, opened 2026-05-19
feat(attendance): add bulk report-records sync

zensgit:
## Summary

Adds a bulk sync path for `attendance_report_records` while keeping the existing single-user sync compatible.

- Preserves `{ from, to, userId }` and the existing per-user writer
- Adds explicit list sync with `{ userIds }`
- Adds all-user paginated sync with `{ allUsers: true, page, pageSize }`
- Reuses the existing per-user export/sync pipeline instead of rewriting attendance aggregation
- Extends the admin UI with an all-active-users toggle and page/pageSize controls
- Adds development and verification docs for the slice

## Boundaries

- No migration
- No direct `meta_*` writes
- No change to `attendance_*` fact-source ownership
- `attendance_report_records` remains a rebuildable report layer, not a query fact source
- `pnpm install` produced tracked node_modules symlink noise in the local worktree; it was intentionally left unstaged and is not in this PR

## Test plan

- [x] `node --check plugins/plugin-attendance/index.cjs`
- [x] `pnpm --filter @metasheet/core-backend exec vitest run tests/unit/attendance-report-field-catalog.test.ts --reporter=dot`
- [x] `pnpm --filter @metasheet/core-backend exec vitest run tests/unit/attendance-report-field-catalog.test.ts tests/unit/attendance-report-field-formula-engine.test.ts --reporter=dot`
- [x] `NODE_OPTIONS=--no-experimental-webstorage pnpm --filter @metasheet/web exec vitest run tests/AttendanceReportFieldsSection.spec.ts --watch=false`
- [x] `NODE_OPTIONS=--no-experimental-webstorage pnpm --filter @metasheet/web exec vitest run tests/AttendanceReportFieldsSection.spec.ts tests/attendance-admin-regressions.spec.ts --watch=false`
- [x] `pnpm --filter @metasheet/web type-check`
- [x] `pnpm --filter @metasheet/core-backend build`
- [x] `git diff --check`

Note: frontend Vitest printed `WebSocket server error: Port is already in use`, but the target specs passed.

### DIFF
--- a/apps/web/src/views/attendance/AttendanceReportFieldsSection.vue
+++ b/apps/web/src/views/attendance/AttendanceReportFieldsSection.vue
@@ -522,8 +522,41 @@
             id="attendance-report-record-sync-user"
             v-model="recordSyncUserId"
             type="text"
-            :placeholder="tr('Required in v1', 'v1 必填')"
+            :disabled="recordSyncAllUsers"
+            :placeholder="recordSyncAllUsers ? tr('All users mode', '全员模式') : tr('Required for single user', '单员工必填')"
             data-report-record-sync-user
+          >
+        </label>
+        <label class="attendance-report-fields__filter attendance-report-fields__filter--inline" for="attendance-report-record-sync-all-users">
+          <span>{{ tr('All active users', '全部活跃员工') }}</span>
+          <input
+            id="attendance-report-record-sync-all-users"
+            v-model="recordSyncAllUsers"
+            type="checkbox"
+            data-report-record-sync-all-users
+          >
+        </label>
+        <label class="attendance-report-fields__filter" for="attendance-report-record-sync-page">
+          <span>{{ tr('Page', '页码') }}</span>
+          <input
+            id="attendance-report-record-sync-page"
+            v-model.number="recordSyncPage"
+            type="number"
+            min="1"
+            :disabled="!recordSyncAllUsers"
+            data-report-record-sync-page
+          >
+        </label>
+        <label class="attendance-report-fields__filter" for="attendance-report-record-sync-page-size">
+          <span>{{ tr('Page size', '每页数量') }}</span>
+          <input
+            id="attendance-report-record-sync-page-size"
+            v-model.number="recordSyncPageSize"
+            type="number"
+            min="1"
+            max="100"
+            :disabled="!recordSyncAllUsers"
+            data-report-record-sync-page-size
           >
         </label>
         <button
@@ -881,11 +914,21 @@ interface AttendanceReportRecordsSyncResult {
   degraded?: boolean
   reason?: string | null
   synced?: number
+  rowsSynced?: number
   patched?: number
   created?: number
   skipped?: number
   failed?: number
   duplicateRowKeys?: number
+  usersScanned?: number
+  usersSynced?: number
+  usersFailed?: number
+  totalUsers?: number
+  page?: number
+  pageSize?: number
+  hasNextPage?: boolean
+  userSelection?: string
+  failedUsers?: Array<{ userId?: string; error?: string; failedRows?: number }>
   fieldFingerprint?: string
   syncedAt?: string
   multitable?: {
@@ -965,6 +1008,9 @@ const recordSyncStatusKind = ref<'info' | 'warn' | 'error'>('info')
 const recordSyncFrom = ref(dateInputValue(-30))
 const recordSyncTo = ref(dateInputValue(0))
 const recordSyncUserId = ref('')
+const recordSyncAllUsers = ref(false)
+const recordSyncPage = ref(1)
+const recordSyncPageSize = ref(50)
 const reportRecordsSyncResult = ref<AttendanceReportRecordsSyncResult | null>(null)
 const fieldSearchTerm = ref('')
 const fieldStatusFilter = ref<ReportFieldStatusFilter>('all')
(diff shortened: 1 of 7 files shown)


zensgit:
## Review-only audit (no scope-gate per `[[attendance-multitable-report-boundary]]` governance)

Diff scope: 7 files, +651/-17 — plugin route + bulk writer, backend unit tests, frontend Vue + spec, dev/verification MD + TODO log update. Read every hunk against the 4 requested points and the 6-point boundary contract.

### 1. 批量路由参数互斥与兼容 — **PASS**

- Zod schema: `userId.optional()` + `userIds.array().max(100).optional()` + `allUsers.boolean().optional()` + `page/pageSize.coerce.optional()` — all optional, validated.
- Mutual exclusivity: `allUsers === true && explicitUserIds.length > 0` → 400 `Use either allUsers or userId/userIds, not both`. ✅
- Required: `allUsers !== true && explicitUserIds.length === 0` → 400 `userId, userIds, or allUsers is required`. ✅
- Bulk cap enforced twice (zod `.max(100)` + runtime `if length > MAX_PAGE_SIZE`); belt-and-suspenders, fine.
- Back-compat: `explicitUserIds.length === 1 && allUsers !== true && !Array.isArray(parsed.data.userIds)` routes to the **original** `syncAttendanceReportRecords` — old single-user clients get byte-identical response shape (no `usersScanned`/`totalUsers`). ✅
- Minor nuance worth a one-liner in the dev doc: `normalizeAttendanceReportRecordsSyncUserIds([userId, ...userIds])` collapses `{ userId: "u-1", userIds: ["u-1"] }` to `["u-1"]`, but the `Array.isArray(parsed.data.userIds)` guard still steers it to the bulk path (different response shape). Not a bug; just an API quirk if a caller toggles between forms. Non-blocking.

### 2. allUsers 分页源与 fallback — **PASS**

- Primary: `user_orgs JOIN users WHERE is_active=true` with `ORDER BY uo.user_id ASC LIMIT/OFFSET` — deterministic.
- Fallback: triggered **only** on `isDatabaseSchemaError(error)` (narrow gate; won't swallow real query errors). Reads `attendance_records DISTINCT user_id WHERE work_date BETWEEN $from AND $to`, same `ORDER BY user_id ASC LIMIT/OFFSET`.
- `userSelection` discriminator (`'allUsers' | 'attendanceRecordsFallback' | 'explicit'`) returned in the response — good observability for which path ran.
- Asymmetry to acknowledge (already inherent to the design, not a bug): primary includes **all active org users** regardless of attendance data; fallback includes **only users with attendance_records in the window**. New employees with no records are picked up by primary but invisible to fallback. Fine for v1 — fallback is a degraded compatibility mode; non-attendance users have nothing to sync.
- Pagination atomicity: count + page are two round-trips, not transactional. Acceptable for admin-triggered manual paging; would matter only under high-frequency concurrent membership churn.

### 3. 逐用户复用 writer 的失败汇总 — **PASS** with one accounting nuance

- `syncAttendanceReportRecordsForUsers` iterates `selection.userIds`, calls the **existing PR2 writer** unchanged per user. No re-implementation of the row-level write path. ✅
- Two failure paths captured cleanly:
  - Per-user **exception** → `usersFailed += 1`, `failed += 1`, push `{ userId, error: <msg> }`, **continue** to next user (does not abort the batch).
  - Per-row **writer failure** (`result.failed > 0`) → `failed += result.failed`, push `{ userId, failedRows: N }`.
- Degraded short-circuit: if any user's writer returns `degraded: true`, the loop exits early returning `{ ...aggregate, ...result, ...selection-metadata }` — partial aggregate stats preserved. Defensible (multitable layer being unavailable is a system-wide condition, not worth continuing).
- 🟡 **Unit-mixing in `aggregate.failed`**: it sums `result.failed` (row count) + `1` per throwing user (user count), and the final `aggregate.usersFailed = aggregate.failedUsers.length` reassignment broadens "usersFailed" to include both throwing-users *and* row-failure-users. Observers seeing `failed: 3, usersFailed: 2` can't tell whether "3" is rows or users. Cosmetic, not blocking; if you want to clean it up later, splitting into `rowsFailed` + `usersFailed` (no dimensional overlap) makes the response self-describing. Non-blocking.

### 4. 只写私有 `attendance_report_records` 且不改变事实源 — **PASS**

- New code **READ-ONLY** against `attendance_*` / `users` / `user_orgs`:
  - `loadAttendanceReportRecordsSyncUserPage`: `SELECT user_id FROM user_orgs JOIN users` (primary) or `SELECT DISTINCT user_id FROM attendance_records` (fallback). ID columns only, no payload.
- The actual write path is the existing PR2 `syncAttendanceReportRecords`, whose body is unchanged in this diff (only `rowsSynced: result.synced` derived alias added on the return shape).
- Multitable boundary preserved: writes only flow through `provisioning.ensureObject/ensureView` + `records.queryRecords/createRecord/patchRecord` to the private `attendance_report_records` object.
- New module exports (`syncAttendanceReportRecordsForUsers`, `normalize…UserIds`, `normalize…Page`, `load…UserPage`) are orchestration/helpers; **none** expose `attendance_*` write capability.

### Boundary contract (6 points from memory) — **all PASS**

1. multitable does not read/write `attendance_*` directly ✅
2. sync writes ONLY private `attendance_report_records` ✅
3. every record carries `source` / `fingerprint` / `period` / `syncedAt` — PR2 writer preserved ✅
4. rebuildable layer, not a migration of `attendance_*` ✅
5. NO new `attendance_*` migration ✅
6. tests cover: explicit-users dedupe + aggregate (`backend test #1`), idempotent second run (`second.toMatchObject({ synced: 3, skipped: 3, created: 0 })`), primary vs fallback paging (`backend test #2`), frontend single-user vs bulk body shapes (`frontend spec #1`) ✅

### Test matrix per verification MD

- Backend catalog + formula: 44 tests PASS
- Frontend section + admin regression: 32 tests PASS
- Web type-check / core-backend build / `git diff --check`: PASS
- Non-fatal Vitest WebSocket port-in-use warning noted in verification MD; consistent with prior PRs.

### Verdict

**Approve as-is.** All 4 requested review points PASS; 6-point boundary contract PASS unconditionally. The unit-mixing in `aggregate.failed` (point 3) is the only thing worth a follow-up cleanup, but it's documentation/cosmetic, not a correctness issue and does not block merge.

PR #1810, opened 2026-05-24
docs(attendance): close out PR5 production heartbeat observation window

zensgit:
## Summary

Docs-only closeout for the 24-hour read-only production-bundle fingerprint observation window opened against PR #1796 (PR5 strong-control runtime) and PR #1804 (advanced-scheduling workbench operator runbook).

**Outcome: PASS.** 4 iterations over 24 h; no post-iter-2 ALERT; no fingerprint regression; no bundle rollback. The single iter=2 ALERT was a transient probe-side flake (HTML extraction timeout), independently verified non-regressive at 02:48Z and corroborated by the two subsequent OK iterations on the forward-deployed bundle hash `DX_EUprZ`.

| File | Type |
|---|---|
| `docs/development/attendance-pr5-prod-heartbeat-closeout-20260524.md` | new closeout MD (+232 lines) |

## What the closeout records

- **Observation window** — iter=1 (baseline, bundle `CWe-UNpc`) / iter=2 (ALERT triaged transient) / iter=3 + iter=4 (OK, bundle `DX_EUprZ` from forward deploys of PR #1791 + #1802 + #1804) / HEARTBEAT_DONE 2026-05-24T17:24:01Z PASS
- **Live read-only API probe** of `GET /api/attendance/advanced-scheduling/workbench`: 5/5 cases PASS (2 positive + 3 negative), matches #1804 runbook contract
- **JWT-injected UI acceptance**: all 4 acceptance points PASS in EN + ZH (strong-control toggle visible, default off, save-time strong-control copy clear, workbench read-only with only "Reload workbench / 重载工作台" button)
- **Real login-form E2E**: `/attendance` auth guard → `POST /api/auth/login 200` → redirect → token + features persisted → `/api/auth/me` confirms admin → admin gating activates
- **Per-fingerprint persistence table**: all 5 required strings (3 PR5 + 2 PR4) non-zero across all 3 OK iterations
- **Lessons recorded** (no code change): step-5 admin assertion was over-strict in the /tmp login script; logout-after-click left the session intact (unverified finding, deferred)
- **Decision tree** for next-round candidates (C → A → B → D → E addendum), all ranked but **NOT executing** without separate explicit opt-in

## Boundaries held throughout

- ❌ No PR6 reporting / multitable snapshot work
- ❌ No scheduling write path
- ❌ No LIMIT 500 truncation fix
- ❌ No fixture writes on production or staging
- ❌ No frontend / backend runtime code change
[...]

PR #1814, opened 2026-05-25
fix(auth): clear local session on logout

zensgit:
## Summary
- harden shell logout to clear the shared local auth/session state after calling /api/auth/logout
- keep the redirect to /login even if cleanup fails
- add a regression spec for clicking Sign out and clearing auth/user/feature localStorage keys
- record the production diagnosis and verification in docs/development/auth-logout-local-session-cleanup-verification-20260525.md

## Verification
- pnpm --filter @metasheet/web exec vitest run tests/App.spec.ts --watch=false
- pnpm --filter @metasheet/web build
- git diff --check

## Boundaries
- frontend shell only
- no backend route changes
- no database or migration changes
- no production data writes

### DIFF
--- a/apps/web/src/App.vue
+++ b/apps/web/src/App.vue
@@ -77,7 +77,7 @@ import { usePlugins } from './composables/usePlugins'
 import { setMultitableApiErrorLocaleResolver } from './multitable/api/client'
 import { resolveRouteDocumentTitle } from './router/routeTitles'
 import { useFeatureFlags } from './stores/featureFlags'
-import { getApiBase } from './utils/api'
+import { clearStoredAuthState, getApiBase } from './utils/api'
 
 const route = useRoute()
 const { navItems: pluginNavItems, fetchPlugins } = usePlugins()
@@ -186,12 +186,9 @@ async function logout(): Promise<void> {
   }
   clearToken()
   try {
-    if (typeof localStorage !== 'undefined') {
-      localStorage.removeItem('metasheet_features')
-      localStorage.removeItem('metasheet_product_mode')
-    }
+    clearStoredAuthState()
   } catch {
-    // Ignore feature cache cleanup failures.
+    // Ignore local session cleanup failures; logout should still leave the view.
   }
   window.location.assign('/login')
 }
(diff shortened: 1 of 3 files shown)


zensgit:
**Independent review — APPROVE (no blocking issues).** Closes the §3.6 logout finding from the #1810 heartbeat closeout.

**Root cause match — confirmed.** Old `logout()` cleared only `metasheet_features` + `metasheet_product_mode`, leaving `auth_token`/`jwt` in localStorage → guard saw "authenticated" against a server-rejected token → blank page. New path delegates to shared `clearStoredAuthState()` which clears 9 keys (token: `auth_token`/`jwt`/`devToken`; tenant: `tenantId`/`workspaceId`; user-state: `metasheet_features`/`metasheet_product_mode`/`user_permissions`/`user_roles`). Both diagnosis-retained keys are now cleared. Using `window.location.assign('/login')` (hard nav, not router.push) is the right call — forces a clean reload that actually resolves the blank page.

**Wire-vs-fixture drift check — PASS** (my main concern). `App.spec.ts` *mocks* `clearStoredAuthState`, so its localStorage assertions test the mock, not the wire. That's fine here because the **real** key list is independently locked by `apps/web/tests/utils/api.test.ts:93`, and `App.spec.ts` correctly locks the wiring (App.vue calls `clearStoredAuthState` once + `fetch` POST `/api/auth/logout` w/ Bearer + `assign('/login')`). Both halves covered.

**CI:** 11 SUCCESS + 1 expected SKIPPED. **Boundary:** frontend shell only; no backend/DB/migration/attendance-logic change; diagnosis artifacts kept in `/tmp`.

**Non-blocking NITs (optional polish, not gates):**
1. `App.spec.ts` localStorage assertions are tautological — the mock removes the keys then asserts they're gone; verifies the mock, not the wire. Load-bearing assertions (fetch/called-once/assign) are correct. Could drop the block or spy on the real fn.
2. `api.test.ts` doesn't assert `tenantId`/`workspaceId` cleared — real fn clears them and the MD claims "tenant hints," but no test locks that pair. Add 2 assertions for full test-backing.
3. Sign-out button located by localized text `退出登录` rather than a stable `data-*`/`id` — brittle to copy/locale changes; a `data-testid` would match the `data-attendance-*` selector discipline.

**Recommendation:** approve for admin squash merge; post-merge production logout smoke is the right follow-up as planned.

zensgit:
Production logout smoke PASS on `index-BkJCQC6c.js` (2026-05-25, post-deploy): logout API 200 / localStorage `auth_token`+`jwt`+`devToken` cleared / redirect to `/login` / login form renders (no blank page). The §3.6 blank-page finding from the #1810 heartbeat closeout is resolved by this PR. Closing the loop.

PR #1819, opened 2026-05-25
docs(attendance): lock comprehensive-hours reporting snapshot boundaries

zensgit:
## What

A **boundary contract** (docs-only, single new file) for comprehensive-hours PR6. It does **not** decide product shape and does **not** implement a writer — it locks the boundaries any future PR6 implementation must hold.

## Key reframe (verified during scoping)

The "reporting / multitable snapshot" mechanism is **already shipped** (report-records sync track, dev docs 2026-05-16…19). So this is written as **"extend the existing substrate"**, not a greenfield lock. Verified line-by-line against `plugins/plugin-attendance/index.cjs` at `74b736722`:

- Snapshot objects `attendance_report_records` (`:1999`) + `attendance_report_period_summaries` (`:2125`)
- Producers `syncAttendanceReportRecords` (`:2427`) / `syncAttendanceReportPeriodSummary` (`:3509`)
- Routes `POST /report-records/sync` (`:27454`) / `/report-period-summaries/sync` (`:27552`)
- Governance fields `row_key`/`field_fingerprint`/`source_fingerprint`/`synced_at`; fingerprint excludes volatile (`:2397`); idempotency skip (`:2549`); degraded (`:2435`/`:2440`); duplicate (`:2544`); stale-null (`:2519`); resumable retry (sync-jobs migration)

**Only genuinely-new surface:** a comprehensive-hours **category** in the field catalog (`:361`) whose computed metrics ride the existing dynamic value-column path (`:2453`). No comprehensive-hours category exists today; comprehensive-hours is preview/enforcement-only (`:11687`) and does not yet flow into the snapshot.

## Boundaries locked (forward commitments for PR6 impl)

No new writer · no `attendance_*` migration · no raw `meta_*` write · rebuildable projection (never a 2nd fact source) · **attendance compute must not read the snapshot** · no multitable-as-input · no K3/Data-Factory/integration-core · no UI/layout decision.

## Reading note

Every "Hard Boundary" is trivially satisfied by this docs-only PR (one .md file). They are commitments for the **future impl PR**, where the evidence gate applies. The Test Matrix uses an **honest enforcement-mechanism column** (integration / guard / unit / review-only — review-only is labeled as such, not dressed up as a CI guard).

## Scope discipline

Single new file in `docs/development/`. No runtime/migration/meta_* change. All citations re-verified with `grep -n` (not the search-digest). Product/UI/metric-set decisions are explicit Open Questions requiring a separate opt-in.

[...]

PR #1823, opened 2026-05-25
docs(attendance): PR6 runtime orientation — reporting blocked on cap-policy persistence

zensgit:
## What

Read-only **orientation** before any PR6 implementation (docs-only, single new file, no code). It answers the four orientation questions and closes with an explicit slice-size verdict.

## Verdict: the first value-plumbing micro-slice is BLOCKED on a missing prerequisite

**Favorable half:** `attendance_report_period_summaries` already has the catalog-driven dynamic value-column path (`index.cjs:3528-3543`, value loop `:3590-3595`), so a new comprehensive-hours catalog code *can* structurally reach the period snapshot — at the correct user+period grain (matching `buildAttendanceComprehensiveHoursComparison` `:11910`). The value-column mechanism is **not** the blocker.

**Disqualifying half:** `comprehensive_hours_excess_minutes = actual − cap`, but **the cap exists only in a live request**. `normalizeAttendanceComprehensiveHoursCapMinutes` (`:12001`) reads it from request input / a request-supplied `policyDraft`; a missing cap is a hard `INVALID_CAP` error (`:12029`), not a fallback to a stored value. There is no comprehensive-hours migration and no settings/cycle/group store holding a cap. Comprehensive-hours is fully **stateless / request-time** (consistent with PR0–PR5; even PR5 strong-control mode was non-persisted). A background sync has no cap, and *every* comprehensive-hours-distinct metric depends on the cap — so there is nothing to compute.

**Why it's blocked, not just large:** the prerequisite (persisted cap policy) collides with the PR6 contract itself —

| Prerequisite | #1819 boundary hit |
|---|---|
| Persist a cap → new policy/settings table | #2 No new migration |
| Admins set the cap → policy CRUD UI | #8 No UI decision |

Orientation did its job: caught the prerequisite **before** any code.

## The four questions (answered in the doc)

- Functions that would change *once unblocked* (period value loop, a comparison call **+ cap source**, catalog seed, tests) — minimal/localized.
- What must NOT change: daily records, route, UI, migration, writer, allUsers (compute is explicit-users-only, `ALL_USERS_NOT_SUPPORTED` `:11971`).
- First metric locked: `comprehensive_hours_excess_minutes`.
- Test landing points: T1 period wire round-trip / T3 source-fingerprint inclusion / T5 no-parallel-producer guard — plus the gap that no dedicated period-summary sync test file exists yet.

## Sequencing (not decided here)

Cap-policy persistence is the real next link and a separate opt-in (it deliberately crosses #1819 #2). Once a cap is persisted, the reporting slice becomes genuinely small. This PR decides nothing beyond recording the finding.

All citations verified with `grep -n`/`sed` against `index.cjs` at `0c0a27de1`.

### DIFF
--- a/docs/development/attendance-comprehensive-hours-pr6-runtime-orientation-20260525.md
+++ b/docs/development/attendance-comprehensive-hours-pr6-runtime-orientation-20260525.md
@@ -47,13 +47,23 @@ there is no cap-free comprehensive-hours value to snapshot.
 
 ## 3. Slice-size verdict: NOT small — blocked
 
-The first micro-slice is not merely "large"; it is gated behind a feature that does not
-exist and that PR6's contract forbids building inside PR6:
+The first micro-slice is not merely "large"; it is gated behind a capability that does not
+exist today: **a persisted comprehensive-hours cap policy.** That prerequisite introduces
+**new persistent policy / admin-configuration semantics** — caps that survive beyond a
+single request, set and owned by an admin — which is precisely why it must be its own
+opt-in, separate from PR6 reporting.
 
-| Prerequisite the slice needs | #1819 boundary it collides with |
-| --- | --- |
-| Persist a cap (per user / group / cycle) → new policy or settings table | **#2 No new migration** |
-| Let admins set the cap → policy CRUD UI | **#8 No UI / no presentation decision** |
+**What the prerequisite does NOT presuppose:** it is *not* necessarily a DB migration and
+*not* necessarily a new UI. A cap-policy shape could be carried by the existing
+`system_configs` settings-JSON path (`packages/core-backend/src/services/ConfigService.ts:216`/`:253`)
+without a new migration, and an admin surface might reuse an existing settings screen. Whether
+a migration, a new table, or a new admin surface is needed is a **downstream design choice**
+for that separate opt-in — not something this orientation should pre-decide, and not
+something to assume crosses #1819 #2 (no migration) or #8 (no UI) by default.
+
+The point that matters for PR6: PR6's contract forbids *deciding* persistence/admin-config
+shape inside PR6, so the cap-policy prerequisite is out of PR6 scope regardless of how it is
+later implemented.
 
 Orientation succeeded exactly as intended: the prerequisite is caught **before** any code,
 not half-way through.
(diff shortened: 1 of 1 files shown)


zensgit:
VERDICT: BLOCK — one boundary statement needs narrowing before merge.

Finding:
- `docs/development/attendance-comprehensive-hours-pr6-runtime-orientation-20260525.md:53-56` and `:92-101` state that cap-policy persistence “deliberately crosses #1819's no-migration boundary” and frames the next design-lock as scoping “a migration + how caps are set.” The blocker that orientation found is real: the current runtime only accepts cap via request/policyDraft, and background period-summary sync has no durable cap source. But the migration conclusion is too strong. The existing attendance settings path persists JSON through `system_configs` (`saveSettings` inserts/updates the existing settings row) and `settingsSchema`/`normalizeSettings` already support adding new top-level policy shapes without a DB migration. A future cap-policy design may choose a table/migration, but it is not forced by today's code.

Why this matters:
- #1819 intentionally narrowed PR6 to no new migration/no UI/no hidden persistence. This PR should say “cap persistence is a separate opt-in because it adds a persisted policy/admin configuration surface,” not “it necessarily crosses no-migration.” Otherwise the orientation doc closes off a valid future design path (settings JSON, no migration) and may steer the next slice toward an unnecessary migration.

Suggested minimal fix:
- Line 55: replace “new policy or settings table | #2 No new migration” with something like “new persisted cap-policy source (settings JSON or table) | outside PR6; if a table/migration is chosen, it requires separate opt-in; any admin configuration/UI also requires separate opt-in.”
- Lines 94-101: replace “deliberately crosses #1819's no-migration boundary” / “scopes a migration + how caps are set” with “deliberately crosses PR6’s no-hidden-persistence/product-decision boundary; migration is an open design choice, not assumed.”

Non-blocking note:
- Line 88 says there is no dedicated period-summary sync test file; true, but `packages/core-backend/tests/unit/attendance-report-field-catalog.test.ts` already contains period-summary sync tests (`period-summary sync: upsert / skip / duplicate / subtype values / stale formula null`, plus explicit-users aggregate). The eventual impl can extend that existing target; it does not necessarily need a new test file.

Everything else I checked is aligned: period-summary is the right grain, the current cap is request-only, `ALL_USERS_NOT_SUPPORTED` is real for v1, and the PR is docs-only with `git diff --check` clean.

Note: GitHub would not let me submit this as a formal “request changes” review because it treats this as my own PR, so I am posting it as a blocking review comment instead.

zensgit:
Fixed in \`147462f21\`. Both points agreed and verified against the code before rewriting.

**Blocking — cap-persistence framing narrowed.** Removed the overclaim that cap persistence "necessarily crosses the no-migration boundary." Confirmed \`system_configs\` is an existing key/value config table with \`ConfigService\` (\`packages/core-backend/src/services/ConfigService.ts:216\`/\`:253\`), so a cap-policy shape could ride the settings-JSON path **without a new DB migration**. New framing (§3/§5): cap persistence is a separate opt-in because it introduces **new persistent policy / admin-config semantics** — not because it needs a migration. Whether a migration / new table / new admin surface is needed is a **downstream design choice** for that opt-in, not presupposed here. PR6's contract forbids *deciding* that shape inside PR6, which is the actual reason cap-policy is out of PR6 scope.

**Non-blocking — test target.** Confirmed \`attendance-report-field-catalog.test.ts:1331\` already covers period-summary sync (managed formula fields, value columns, active codes, row key, source fingerprint). §4(d) now says the impl slice can **extend this existing target** (add the comprehensive-hours value + T1 wire round-trip), no new file required.

Also corrected the §6 cross-reference wording. Ready for re-review.

zensgit:
Re-review VERDICT: PASS-with-merge-note.

The prior blocker is fixed. The document now correctly says the cap-policy prerequisite is out of PR6 because it introduces new persistent policy/admin-configuration semantics, while leaving migration-vs-settings-JSON as a downstream design choice. That matches the code: the current comprehensive-hours cap still only comes from request/policyDraft (`normalizeAttendanceComprehensiveHoursCapMinutes`), and the existing `system_configs`/ConfigService path means persistence does not necessarily require a DB migration.

I also verified the non-blocking test-target note is corrected: the doc now points future impl work at the existing `packages/core-backend/tests/unit/attendance-report-field-catalog.test.ts:1331` period-summary sync target instead of implying a new test file is required.

Rechecked:
- docs-only: one new file under `docs/development/`
- `git diff --check`: clean
- secret/home-path scan on the new doc: clean
- CI: 10/10 success

Merge note: PR is currently `MERGEABLE` but `BEHIND`, so rebase/update onto latest `main` before admin squash merge, then let the checks rerun green. No further content changes requested.

PR #1825, opened 2026-05-25
docs(attendance): design-lock comprehensive-hours cap-policy persistence

zensgit:
## What

A **design-lock** (docs-only, single new file, no code) for the comprehensive-hours cap-policy persistence that PR6 reporting is blocked on (per orientation #1823). **This design does NOT unlock implementation** — the cap-policy impl is a separately-gated opt-in (named impl PR), and PR6 is a further opt-in after it.

## V1 contract locked

**REQUIRED — resolver contract (the keystone):**
```
resolveComprehensiveHoursCap(orgId, userId, period)
  -> { capMinutes, source, fingerprintPayload } | null
```
- `period` is the real resolved period object; `type ∈ month/quarter/year/payroll_cycle/custom_range` (the literal enum, `index.cjs:11676`/`:11702`).
- `null` when unconfigured → sync **stale-nulls** the columns (inherits `:2519`), not zero/error.
- `userId` is in the V1 signature for forward-compat (per-user override slots in later), though V1 ignores it.

**RECOMMENDED source:** org default by cycle-type for `month/quarter/year`; `payroll_cycle`/`custom_range` return `null` in V1 (deferred). **Storage: reuse `system_configs` settings-JSON via `ConfigService` (`ConfigService.ts:216`/`:253`) — no migration.** `effective_key` = config-revision id, **not** an effective-date (resolves the version-vs-no-audit tension).

**Fingerprint bridge to #1819/PR6:** companion catalog value-fields (`comprehensive_hours_cap_minutes/_cap_source/_cap_effective_key`) ride the existing dynamic value-column path (`:3536-3543`/`:3595`); the existing `source_fingerprint` exclusion rule picks them up automatically (no fingerprint code change). Re-sync is **passive** (fingerprint change on next sync).

**OPTIONAL extension (separate opt-in):** attendance-group override — esp. if effective-date/audit is needed (likely a table → migration then).

**EXPLICIT deferred:** per-user override, employee category, historical audit table, effective-dating + retroactive recompute, payroll_cycle/custom_range org-default derivation.

**Stage lock:** docs-only; impl is a separately-gated opt-in.

## Test matrix

R1–R6 with an honest enforcement column (unit / integration / review-only). R4 = cap-edit→fingerprint→re-sync against real wire.

All citations grep-verified against `index.cjs` at `996bf3c73`; cycle-type literals match code (`month/quarter/year`).

### DIFF
--- a/docs/development/attendance-comprehensive-hours-cap-policy-persistence-design-20260525.md
+++ b/docs/development/attendance-comprehensive-hours-cap-policy-persistence-design-20260525.md
@@ -28,7 +28,7 @@ resolveComprehensiveHoursCap(orgId, userId, period)
 
 **Input**
 - `orgId`, `userId` — strings.
-- `period` — the already-resolved period object produced by `resolveAttendanceComprehensiveHoursPeriod` (`:11702`): `{ type, key, from, to, label }`, where `type ∈ { month, quarter, year, payroll_cycle, custom_range }` (the literal enum in `ATTENDANCE_COMPREHENSIVE_HOURS_PERIOD_TYPES` `:11676`; the spec's "monthly/quarterly/yearly" = `month/quarter/year`). `key` is the stable period identifier (`2026-03`, `2026-Q1`, `range:from:to`).
+- `period` — the already-resolved period object produced by `resolveAttendanceComprehensiveHoursPeriod` (`:11702`): `{ type, key, from, to, label }`, where `type ∈ { month, quarter, year, payroll_cycle, custom_range }` (the literal enum in `ATTENDANCE_COMPREHENSIVE_HOURS_PERIOD_TYPES` `:11676`; the spec's "monthly/quarterly/yearly" = `month/quarter/year`). `key` is the stable period identifier (`2026-03`, `2026-Q1`, `range:from:to`). Note: the PR6 producer emits `date_range`/`payroll_cycle`, not these types directly — §2.1 defines the required bridge that derives `period.type` before the resolver is called.
 
 **Output**
 - `capMinutes` — resolved cap in minutes (same unit as `buildAttendanceComprehensiveHoursComparison` expects, `:11910`).
@@ -41,7 +41,7 @@ resolveComprehensiveHoursCap(orgId, userId, period)
     comprehensive_hours_cap_effective_key: <effectiveKey>,
   }
   ```
-- **`null`** when no cap is configured for this (org, period-type). The caller (period sync) then **stale-nulls** the `comprehensive_hours_*` columns — inheriting the existing disabled-field stale-null pattern (`:2519`). Null is **not** an error and **not** zero.
+- **`null`** when no cap is configured for this (org, period-type). The caller (period sync) then **stale-nulls** the `comprehensive_hours_*` columns — inheriting the existing period-summary stale-null pattern (`:3590-3592`, the `!activeValueCodes.has(column.id) → null` branch). Null is **not** an error and **not** zero.
 
 **Why `userId` is in the V1 signature even though V1 ignores it:** forward-compat. The
 per-user override extension (§5) slots in without changing any caller. V1 resolves the same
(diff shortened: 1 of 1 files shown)


zensgit:
VERDICT: BLOCK — design direction is right, but the period-type bridge is missing.

Finding 1 (blocking): V1 cap source is keyed by comprehensive-hours period types, but the PR6 target producer currently only receives report-period sync types.
- In this doc, the resolver input is pinned to `resolveAttendanceComprehensiveHoursPeriod` output with `type ∈ { month, quarter, year, payroll_cycle, custom_range }` (`docs/development/attendance-comprehensive-hours-cap-policy-persistence-design-20260525.md:29-31`), and V1 org defaults only cover `month/quarter/year` while `payroll_cycle/custom_range` return `null` (`:52-55`).
- But the existing `POST /api/attendance/report-period-summaries/sync` schema accepts only `from/to` or `cycleId` (`plugins/plugin-attendance/index.cjs:27555-27564`), and `resolveAttendanceReportPeriodSyncPeriod` produces `periodType: 'payroll_cycle'` or `periodType: 'date_range'` (`plugins/plugin-attendance/index.cjs:3461-3498`). The current UI also only exposes `date_range` / `payroll_cycle` for period-summary sync.
- Result: if PR6 obeys #1819/#1823 and reuses the existing period-summary producer/route, the V1 resolver as written has no `month/quarter/year` signal for normal `from/to` syncs. Those rows would fall through to `custom_range`/null, stale-null the comprehensive-hours columns, and the “org default by cycle-type” would not actually unblock PR6 reporting.

Suggested minimal doc fix:
- Add an explicit “period-type bridge” decision. For example: the cap resolver can still use `{ month, quarter, year, payroll_cycle, custom_range }`, but PR6 period-summary sync must derive that type from the existing report period before calling it: exact calendar month range → `month`, exact quarter range → `quarter`, exact calendar year range → `year`, otherwise `custom_range`; `payroll_cycle` remains null in V1 unless a later cycle-template mapping is opted in. This preserves “no route change” and makes the V1 defaults usable through the existing `/report-period-summaries/sync` path.
- Or, if the intended design is to add/require a comprehensive-hours-specific period input route later, say that explicitly — but that would be a separate opt-in and should not be implied as unlocked here.

Finding 2 (small but should fix while editing): stale-null citation points at daily, not period.
- Lines `:44` and `:111` cite `:2519`, which is the daily `attendance_report_records` stale clear loop. The period-summary path has its own inactive-value stale-null at `plugins/plugin-attendance/index.cjs:3590-3592`, and the source fingerprint is built from the period logical payload at `:3598` with exclusions at `:3364-3374`.
- Please cite the period path (`:3590-3592`, optionally with daily `:2519` as analogous prior art) so future PR6 impl follows the right producer.

Non-blocking wording suggestion:
- `system_configs settings-JSON via ConfigService` is directionally fine for “no migration,” but the attendance plugin already has its own `attendance.settings` path (`SETTINGS_KEY`, `loadSettings/saveSettings`, `settingsSchema`, `normalizeSettings`) over `system_configs`. Consider wording this as “reuse existing `system_configs` settings JSON / attendance settings normalizer (or ConfigService-equivalent), no new table by default” unless you intentionally want a separate config key outside `attendance.settings`.

What passed:
- Docs-only, single new file.
- `git diff --check` clean; secret/home-path scan clean.
- CI is 10/10 green.
- Stage-lock language is good: design-lock does not unlock implementation; cap-policy impl and PR6 reporting remain separate opt-ins.

zensgit:
Fixed in \`35c782d16\`. The bridge gap was a real miss — good catch; verified against code before fixing.

**Blocking — period-type bridge added (new §2.1).** Confirmed `resolveAttendanceReportPeriodSyncPeriod` (`:3432`) emits **only** `payroll_cycle` (`:3462`) / `date_range` (`:3497`), never `month/quarter/year` — so without a bridge the org defaults are unreachable and PR6 always stale-nulls. Added the required bridge: derive `period.type` from the `date_range` window —
- exact natural month → `month`
- exact natural quarter → `quarter`
- exact natural year → `year`
- other `date_range` → `custom_range` (→ null → stale-null)
- `payroll_cycle` → stays `null` in V1; cycle-template mapping is a separate opt-in (§6)

Resolver keeps its clean `month/quarter/year/payroll_cycle/custom_range` vocabulary; the bridge sits between the producer and the resolver. Added R7 to the test matrix, and a forward-ref from §1.

**Citation fix — stale-null.** `:2519` was the daily path; corrected to the period path `:3590-3592` (the `!activeValueCodes.has(column.id) → null` branch) in the §1 null bullet and test R2.

**Storage wording broadened.** Now prefers the **existing `attendance.settings` normalizer** (`SETTINGS_KEY` `:51`, `attendance.settings.updated` event `:27801`) or a `ConfigService` settings-JSON equivalent, and explicitly forbids bypassing the normalizer with a parallel config write. Still migration-free. R6 updated.

All new citations grep-verified at `996bf3c73`. Ready for re-review.

zensgit:
Re-review VERDICT: PASS-with-merge-note.

The prior blockers are fixed:
- Period-type bridge is now explicit: existing period-summary sync emits `date_range` / `payroll_cycle`, and the design derives exact month/quarter/year from aligned `date_range` windows before calling the cap resolver. That makes org defaults reachable without changing the existing `/report-period-summaries/sync` route.
- Stale-null references now point at the period-summary path (`:3590-3592`) instead of the daily path.
- Storage wording now prefers the existing `attendance.settings` normalizer / settings JSON path and explicitly forbids a parallel config write; migration remains unnecessary for V1 unless a later opt-in pulls in group/effective-date/audit complexity.

Rechecked:
- docs-only: one new file under `docs/development/`
- `git diff --check`: clean
- secret/home-path scan on the new doc: clean
- CI: 10/10 success

Merge note: PR is `MERGEABLE` but currently `BEHIND`, so rebase/update onto latest `main` before admin squash merge, then let checks rerun green. No further content changes requested.

PR #1829, opened 2026-05-25
feat(attendance): persist + resolve comprehensive-hours cap defaults (V1)

zensgit:
## What

First **runtime** slice of comprehensive-hours cap-policy persistence, implementing the V1 contract from the design-lock (#1825). Makes the cap **persistable / resolvable / fingerprintable**. 2 files, +200.

**Explicitly NOT in this slice** (per #1825 + the agreed scope): no UI, no attendance-group override, no effective-dating, **no migration**, no change to preview behavior, and **no PR6 reporting value-plumbing** (the resolver is not wired into the period sync — that's the next separate opt-in).

## Changes

- **Settings:** `attendance.settings` gains `comprehensiveHours.capDefaults { month, quarter, year }` (minutes; positive-int-or-null), normalized via `normalizeAttendanceComprehensiveHoursSettings` + merged in `mergeSettings`. Settable through the **existing** `/api/attendance/settings` route — no new route, no UI.
- **Resolver:** `resolveAttendanceComprehensiveHoursCap(settings, orgId, userId, period) → { capMinutes, source, fingerprintPayload } | null`. V1 = org default by cycle-type (`month/quarter/year`); `payroll_cycle`/`custom_range` → `null`. `orgId`/`userId` are forward-compat (per-org / per-user override later), unused in V1 because `attendance.settings` is a single global config.
- **Period-type bridge:** `bridgeAttendanceDateRangeToComprehensiveHoursPeriod(from,to)` — the period sync emits `date_range`, so exact natural month/quarter/year windows map to those cycle-types (else `custom_range`), making the org defaults reachable through the real producer.
- **effective_key:** content hash of `capDefaults` (config-revision marker, **not** an effective-date); carried in `fingerprintPayload` so a cap edit changes the row's `source_fingerprint` → passive re-sync (PR6 wires this).

## Staging note (for the "no dead helpers" check)

The resolver / bridge / fingerprint-payload are the **locked contract units** from #1825, consumed by the named-next **PR6 value-plumbing** slice. Their current exerciser is the R1–R7 suite. This is deliberate staging, not dead code — the settings half is already live via the existing settings route. The integration halves of R2/R4 (sync stale-null, actual re-sync) require the period-sync wire and belong to PR6; they are **not** faked here.

## Tests / verification

- R1–R7 + normalizer: **19/19** in `attendance-comprehensive-hours-control.test.ts` (8 new).
- Full core-backend unit suite: **2300/2300** green (no settings-shape regression).
- `tsc` (build:cache) clean.

### DIFF
--- a/packages/core-backend/tests/integration/attendance-plugin.test.ts
+++ b/packages/core-backend/tests/integration/attendance-plugin.test.ts
@@ -3687,9 +3687,23 @@ describe('Attendance Plugin Integration', () => {
         body: JSON.stringify({
           ...originalSettings,
           minPunchIntervalMinutes: 60,
+          comprehensiveHours: { capDefaults: { month: 10560, quarter: 31680, year: 126720 } },
         }),
       })
       expect(saveSettingsRes.status).toBe(200)
+
+      // Real-wire round-trip: comprehensiveHours.capDefaults must survive the PUT zod
+      // schema + persistence and come back on GET — not be silently stripped. Guards the
+      // save-path drift class (a field present in the normalizer but missing from the
+      // route schema would persist as nothing).
+      const reloadRes = await requestJson(`${baseUrl}/api/attendance/settings`, {
+        headers: { Authorization: `Bearer ${token}` },
+      })
+      expect(reloadRes.status).toBe(200)
+      const reloaded = (reloadRes.body as {
+        data?: { comprehensiveHours?: { capDefaults?: Record<string, number | null> } }
+      } | undefined)?.data
+      expect(reloaded?.comprehensiveHours?.capDefaults).toEqual({ month: 10560, quarter: 31680, year: 126720 })
 
       const futurePunchRes = await requestJson(`${baseUrl}/api/attendance/punch`, {
         method: 'POST',
--- a/packages/core-backend/tests/unit/attendance-comprehensive-hours-control.test.ts
+++ b/packages/core-backend/tests/unit/attendance-comprehensive-hours-control.test.ts
@@ -484,3 +484,96 @@ describe('attendance comprehensive working-hours control helpers', () => {
     expect(pluginSource).not.toContain("'DELETE',\n      '/api/attendance/comprehensive-hours/preview'")
   })
 })
+
+describe('comprehensive-hours cap-policy persistence (V1)', () => {
+  const settingsWith = (capDefaults: Record<string, number | null>) => ({
+    comprehensiveHours: { capDefaults },
+  })
+
+  // R1 — resolver contract: org default by cycle-type for month/quarter/year.
+  it('R1: resolves the org default cap per cycle-type', () => {
+    const settings = settingsWith({ month: 10560, quarter: 31680, year: 126720 })
+    for (const [type, expected] of [['month', 10560], ['quarter', 31680], ['year', 126720]] as const) {
+      const resolved = helpers.resolveAttendanceComprehensiveHoursCap(settings, 'org-1', 'u-1', { type, key: 'k' })
+      expect(resolved).not.toBeNull()
+      expect(resolved.capMinutes).toBe(expected)
+      expect(resolved.source).toBe('org_default_by_cycle_type')
+    }
+  })
+
+  // R2 (resolver half) — no cap configured → null (not 0, not error). The sync-side
+  // stale-null behaviour is exercised by PR6 against the real wire.
+  it('R2: returns null when the cap is unset for the cycle-type', () => {
+    const settings = settingsWith({ month: null, quarter: null, year: null })
+    expect(helpers.resolveAttendanceComprehensiveHoursCap(settings, 'org-1', 'u-1', { type: 'month', key: 'k' })).toBeNull()
+    expect(helpers.resolveAttendanceComprehensiveHoursCap({}, 'org-1', 'u-1', { type: 'year', key: 'k' })).toBeNull()
+  })
+
+  // R3 — payroll_cycle / custom_range have no org default in V1 → null.
+  it('R3: payroll_cycle and custom_range resolve to null in V1', () => {
+    const settings = settingsWith({ month: 10560, quarter: 31680, year: 126720 })
+    expect(helpers.resolveAttendanceComprehensiveHoursCap(settings, 'org-1', 'u-1', { type: 'payroll_cycle', key: 'c' })).toBeNull()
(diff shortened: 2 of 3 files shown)


zensgit:
VERDICT: BLOCK — the runtime contract is not actually persistable through the real settings route yet.

Finding 1 (blocking): `comprehensiveHours` is normalized but not accepted by `PUT /api/attendance/settings`.
- The PR adds `DEFAULT_SETTINGS.comprehensiveHours` and `normalizeSettings(...).comprehensiveHours` (`plugins/plugin-attendance/index.cjs:93-103`, `:10096` on PR HEAD), but the route input schema still stops at `formula` and has no `comprehensiveHours` entry (`plugins/plugin-attendance/index.cjs:14138-14208`).
- The actual save route parses `req.body` through that zod object before `mergeSettings`/`saveSettings` (`plugins/plugin-attendance/index.cjs:27787-27798`). Zod object parsing strips unknown keys by default, so a request like `{ comprehensiveHours: { capDefaults: { month: 10560 } } }` is silently dropped before it reaches the normalizer.
- Net effect: the PR summary says “settable through existing /api/attendance/settings route,” but the real wire cannot persist it. This is the same class of bug as the earlier calendar-policy save-path issue: local helper tests pass while the real settings route discards the new field.

Required fix:
- Add `comprehensiveHours` to `settingsSchema`, with `capDefaults.month/quarter/year` accepted as the same intended wire types you want to support (number, or number/string if you deliberately want numeric strings).
- Add a route/merge-level unit or integration test that exercises the real `settingsSchema -> mergeSettings -> normalizeSettings` path, not only `normalizeAttendanceComprehensiveHoursSettings` directly. At minimum, prove the parsed body is not stripped and the saved/returned settings include `comprehensiveHours.capDefaults`.

Finding 2 (blocking unless schema requires full capDefaults): partial cap updates can erase sibling caps.
- `mergeSettings` currently merges `comprehensiveHours` only one level deep (`plugins/plugin-attendance/index.cjs:10143-10146`). If the route later accepts `{ comprehensiveHours: { capDefaults: { month: 600 } } }` while existing settings have `quarter/year`, the merged object replaces the whole `capDefaults`, and `normalizeSettings` turns missing `quarter/year` into `null`.
- That is risky because the settings API is otherwise partial-update shaped. Either deep-merge `capDefaults`, or make the schema require all three cap defaults whenever `capDefaults` is present and document that wire contract. I recommend deep-merge; it matches the existing partial settings route ergonomics.

What passed:
- The resolver/bridge helpers themselves are well scoped and not wired into PR6 yet.
- The date-range bridge covers month/quarter/year/custom_range correctly for the tested cases.
- No migration, no new route, no UI, no PR6 writer wiring.
- CI is green, but it does not cover the real settings save wire path for this new field yet.

zensgit:
Fixed in \`c1c63589a\`. You were right — and it was exactly the save-path drift class; my "settable through existing route" claim was false and, worse, untested. Both points fixed plus a real-wire lock.

**1. zod schema strip (the BLOCK).** `settingsSchema` (the `PUT /api/attendance/settings` validator) now accepts `comprehensiveHours.capDefaults { month, quarter, year }` (`int >= 0 | null`). Previously zod stripped the unknown field before `mergeSettings`/`saveSettings` ever saw it, so the cap never persisted.

**2. shallow merge.** `mergeSettings` now **deep-merges** `capDefaults`, so a partial update like `{ capDefaults: { month } }` no longer nulls out the existing `quarter`/`year`.

**3. Real-wire round-trip locked.** The integration test (`attendance-plugin.test.ts`, hits the route over HTTP with Postgres) now PUTs `comprehensiveHours.capDefaults` and asserts the GET returns them — proving no zod strip + actual persistence. This is the test that would have caught the original bug; it runs in CI with the DB. I added it because the R1–R7 units tested the resolver in isolation and missed the wire — the same gap you flagged.

Local guards added too (runnable without a DB): a `mergeSettings` deep-merge unit test, and a source-level assertion that `settingsSchema` includes `comprehensiveHours.capDefaults`.

**Verification:** `attendance-comprehensive-hours-control` 21/21; full core-backend unit suite **2302/2302**; `tsc` clean. The integration round-trip I could not run locally (no Postgres) — it runs in CI; flagging that honestly rather than claiming a local pass. Resolver/bridge still not wired into any PR6 producer; no migration / no new route / no UI. Ready for re-review.

zensgit:
Re-review VERDICT: PASS-with-notes

The two blocking findings from the first pass are fixed in PR HEAD c1c63589a.

What I re-checked:

- `settingsSchema` now accepts `comprehensiveHours.capDefaults` in `PUT /api/attendance/settings`, so the field is no longer silently stripped before `mergeSettings()` / `saveSettings()` (`plugins/plugin-attendance/index.cjs:14215`).
- `mergeSettings()` now deep-merges `comprehensiveHours.capDefaults`, so a partial update such as `{ month }` preserves existing `quarter` / `year` caps (`plugins/plugin-attendance/index.cjs:10143`).
- The real-wire regression lock is now present: the integration test PUTs `comprehensiveHours.capDefaults` through the actual settings route and then GETs settings back to assert the values survived the route validator + persistence path (`packages/core-backend/tests/integration/attendance-plugin.test.ts:3681`). This is the key save-path drift guard that was missing before.
- R1-R7 plus normalizer / merge / schema guards now cover the resolver, period bridge, fingerprint payload, effective-key change, and the route-schema acceptance contract (`packages/core-backend/tests/unit/attendance-comprehensive-hours-control.test.ts:493`).
- Scope stays within the agreed slice: no migration, no new route, no UI, no preview behavior change, no PR6 snapshot writer wiring. Diff is still only `plugins/plugin-attendance/index.cjs` plus the two test files.

Checks observed:

- GitHub checks: 14 PASS + 1 expected SKIP (`Strict E2E with Enhanced Gates`).
- `git diff --check origin/main...origin/pr-1829`: PASS.
- Added-lines secret/home-path scan for the touched files: no hits.

Non-blocking note:

- V1 rides the existing global `attendance.settings` storage; `resolveAttendanceComprehensiveHoursCap(settings, orgId, userId, period)` intentionally keeps `orgId` / `userId` forward-compatible but unused. The source label `org_default_by_cycle_type` is acceptable for this slice as the configured default cap source, but it should not be read as true per-org storage isolation yet. Any real per-org / group / user override remains a separate design + implementation opt-in.

Merge note:

- PR is currently `MERGEABLE` but `BEHIND`; rebase onto current `origin/main`, let checks rerun green, then admin squash is OK from this review.

PR #1843, opened 2026-05-26
docs(attendance): staging acceptance evidence for comprehensive-hours reporting V1

zensgit:
## What

Docs-only (single new file) recording the **staging acceptance** of comprehensive-hours reporting V1 (#1833 value-plumbing) against `:8082` on 2026-05-26. This is the real-stack evidence the producer-level mock tests in #1833 could not provide — moving reporting V1 from *code + closeout* to **staging-verified closeout**.

## Result: PASS

- **Gates:** staging admin JWT valid · `auth/me` admin/attendanceAdmin · `GET settings` carries `comprehensiveHours.capDefaults` (proves the staging backend was redeployed to a `main` including #1833; an earlier preflight correctly **blocked** when it predated #1829).
- **PUT cap 600** (partial body) → saved with `quarter`/`year` preserved (deep-merge on the real wire) → GET-confirmed.
- **Period sync** (single user, natural-April `date_range`) → `created=1`, **non-degraded**.
- **Snapshot row read back** (via `/api/multitable/records` + `/fields`): all **4 `comprehensive_hours_*` columns present** — `excess=0` (admin has no April attendance), `cap=600`, `source=org_default_by_cycle_type`, `effective_key=cfg:7a6c2f2969e9`.
- **Idempotency:** re-sync unchanged → `skipped=1`.
- **Cap-change re-sync:** `600→900` → `patched=1`, new `effective_key` → proves the cap is in `source_fingerprint` (pre-#1833 code would not re-sync on a cap-only change).
- **Restore:** original settings (capDefaults all-null) restored; GET confirms.

## Safety

Staging only, production untouched, no code change, no migration. Token read from a 0600 file, never printed, deleted after the run, and rotated server-side (it had been pasted in operator chat). The only writes were a reversible settings cap change (restored) and rebuildable snapshot rows.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1833, opened 2026-05-25
feat(attendance): PR6 value-plumbing — surface comprehensive_hours_excess_minutes (V1)

zensgit:
## What

Minimal PR6 runtime slice: wires the #1829 cap resolver into `syncAttendanceReportPeriodSummary`, surfacing **one** period-level metric `comprehensive_hours_excess_minutes` + 3 cap companion fields into the `attendance_report_period_summaries` snapshot. 3 files, +311/-3.

**Out of scope (unchanged):** no route, no UI, **no migration**, no new writer/producer, daily `attendance_report_records` untouched, preview behavior unchanged, no group/per-user override, no effective-dating, no `payroll_cycle` cap derivation.

## Design

- 4 **system-managed** value columns (`ATTENDANCE_COMPREHENSIVE_HOURS_PERIOD_VALUE_COLUMNS`) appended to the provisioned set (`allValueColumns`) — they ride the existing `ensureObject` + `records.createRecord/patchRecord` path. **Not** added to `valueFields`, so the field-config fingerprint is unaffected (these are system columns, not catalog config — no catalog seed, no migration).
- `buildAttendanceComprehensiveHoursPeriodSummaryValues`: **fail-closed** — only an explicit `date_range` bridging to an aligned month/quarter/year with a configured org default yields a cap; `payroll_cycle` / `custom_range` / missing `periodType` / unset cap all **stale-null** all four columns. Excess reuses `buildAttendanceComprehensiveHoursComparison` (no parallel computation).
- Values set into `logical` **after** the catalog value loop and **before** `buildAttendanceReportPeriodSummarySourceFingerprint(logical)`, so cap value/source/effective_key are hashed into `source_fingerprint` → a cap edit re-syncs the row; existing pre-PR6 rows re-sync once automatically (extra keys diverge the fingerprint).

## Tests (28/28, 7 new)

value/stale-null logic · **real-sync stale-null** · computed write · **cap-fingerprint re-sync** (capture-and-replay through the real fingerprint mechanism) + skip-when-unchanged · **pre-PR6 row migration** · **no raw INSERT/UPDATE + no `meta_`/snapshot-table SQL** (writes only via records API) · no-parallel-producer + wiring/ordering source guards. Full core-backend unit suite **2309/2309**; `tsc` clean.

### Test-level disclosure

[...]

PR #1804, opened 2026-05-24
docs(attendance): add advanced-scheduling workbench operator runbook

zensgit:
## Summary

Docs-only slice — adds an operator runbook for the existing read-only Advanced Scheduling Workbench (`GET /api/attendance/advanced-scheduling/workbench`) plus a verification record. No runtime code, no migrations, no scope expansion.

| File | Type |
|---|---|
| `docs/operations/attendance-advanced-scheduling-workbench-runbook.md` | new operator runbook (+424 lines) |
| `docs/development/attendance-advanced-scheduling-workbench-runbook-verification-20260524.md` | new verification record (+134 lines) |

## What the runbook documents

- **Entrypoint** — `Admin Center → Attendance → Scheduling → Advanced scheduling`, GET-only route, `attendance:admin` permission.
- **Workbench sections** — read-only snapshot banner, summary metrics, schedule-groups coverage table, diagnostics list, truncation / sampling vs. aggregate semantics (assignment row sampling capped at 500/kind; aggregates remain exact).
- **Five diagnostic codes** with severity, API message verbatim, trigger condition, target field (`userIds` or `scheduleGroupIds`), and operator remediation pointing at the matching admin section:
  - `schedule_group_without_members` (warning)
  - `assignment_without_schedule_group` (warning)
  - `user_multiple_schedule_groups` (warning)
  - `user_mixed_assignment_kinds` (info)
  - `scheduler_scope_unknown_schedule_group` (warning)
- **Operator triage flow** — date range → truncation banner → aggregates → diagnostics → matching admin section → re-poll.
- **Safe GET-only live-check commands** that use `AUTH_TOKEN_FILE` indirection so no JWT is ever in the doc or echoed to the terminal.

## Explicit scope boundaries (recorded in runbook §5)

- ❌ Read-only — does not edit `attendance_schedule_groups`, `attendance_schedule_group_members`, `attendance_scheduler_scopes`, `attendance_shifts`, `attendance_rotation_rules`, `attendance_assignments`, or `attendance_rotation_assignments`.
- ❌ Not a substitute for the effective-calendar resolver — day-level precedence stays with `resolveEffectiveCalendar`.
- ❌ No grid edit / copy-paste.
- ❌ No Excel import.
- ❌ No temporary shift / dispatch / shift swap.
- ❌ No K3 / Data Factory / Bridge Agent touch.
- ❌ **PR6 (reporting / multitable snapshot) remains explicitly deferred** — this PR does not advance it.

## Grounding

[...]

PR #1850, opened 2026-05-26
docs(attendance): comprehensive-hours reporting V1 staging acceptance runbook

zensgit:
## What

Docs-only operator runbook (`docs/operations/`) that operationalizes the staging acceptance proven on 2026-05-26 (#1843), so future verifications (after a new deploy, or for a data-bearing user) are push-button. Read-only-first, fail-closed, fully reversible.

## Contents

- **Prerequisites:** staging admin JWT (distinct `JWT_SECRET` from prod), deploy including #1833 (staging lane doesn't auto-mirror main), migration-alignment caveat.
- **Secret hygiene:** token kept in a 0600 file and passed via `curl -H @<header-file>` so it never appears on `argv`/process list; rotate server-side if exposed.
- **3 read-only gates**, incl. the #1829-deploy discriminator (`GET settings` must carry `comprehensiveHours.capDefaults`).
- **7 acceptance steps:** save original → PUT test cap (deep-merge) → GET confirm → single-user natural-month `date_range` sync → read snapshot back via `/api/multitable/records` + `/fields` (map `fld_*`→name) → idempotency + cap-change re-sync behavioral proof → restore.
- **Pass criteria + troubleshooting:** `degraded`=env/migration gap (not a code defect); missing `comprehensiveHours`=pre-#1829 build; `excess=0`=no-data user; `401`=expired/non-staging token; `custom_range`/`payroll_cycle`→null by design.

Route citations verified against `plugins/plugin-attendance/index.cjs`; all four cross-referenced docs exist on main.

### DIFF
--- a/docs/operations/attendance-comprehensive-hours-reporting-staging-acceptance-runbook.md
+++ b/docs/operations/attendance-comprehensive-hours-reporting-staging-acceptance-runbook.md
@@ -1,0 +1,108 @@
+# Comprehensive-Hours Reporting V1 — Staging Acceptance Runbook
+
+Operator procedure to verify comprehensive-hours **reporting V1** (value-plumbing, #1833) on
+a real stack after a staging deploy. It is staging-only, read-only-first, fail-closed, and
+fully reversible (the one settings change is restored at the end). First proven 2026-05-26
+(`docs/development/attendance-comprehensive-hours-pr6-staging-acceptance-20260526.md`).
+
+## When to run
+
+- After deploying a build that includes #1833 to a staging/test backend, to confirm the
+  cap → resolver → period sync → multitable snapshot path works end-to-end.
+- Not against production. No code changes are part of this procedure.
+
+## Prerequisites
+
+- A staging **admin** JWT minted against that environment's `JWT_SECRET` (staging uses a
+  distinct secret from prod — a prod token returns `401 Invalid token`). Store it in a
+  `0600` file; never paste it into chat or a ticket.
+- The staging backend deployed to `main` **including #1833** (the staging deploy lane does
+  not auto-mirror `main` — confirm the deploy explicitly).
+- **Migration alignment:** staging postgres may trail prod-track. If it is behind, the report
+  objects / multitable provisioning can be unavailable and the sync returns `degraded`.
+  Apply pending migrations as part of the deploy.
+
+## Secret hygiene
+
+Keep the token in files only — never on a command line (it would show in the process list /
+shell history). Build a curl header file without the token touching `argv`:
+
+```bash
+JWT=/path/to/staging-admin.jwt          # 0600
+{ printf 'Authorization: Bearer '; cat "$JWT"; } > /tmp/stg-hdr && chmod 600 /tmp/stg-hdr
+# then: curl -s -H @/tmp/stg-hdr ...
+# at the end: rm -f /tmp/stg-hdr
+```
+
+If the token is ever exposed (e.g. pasted into chat), deleting local files is not enough —
+**invalidate/rotate it server-side**.
+
+Set once: `BASE=http://<staging-host>:<port>` (e.g. the staging backend).
+
+## Gates — run read-only first; STOP if any is red
+
+| Gate | Command | Pass |
+| --- | --- | --- |
+| G1 — JWT valid | decode the JWT payload, check `exp` > now | not expired; `roles` includes `admin` |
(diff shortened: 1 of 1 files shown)


zensgit:
Review VERDICT: BLOCK — restore step is not actually reversible for non-null existing caps

The runbook is otherwise a good operationalization of the #1843 acceptance, but one safety issue needs fixing before merge.

Blocking finding:

- `docs/operations/attendance-comprehensive-hours-reporting-staging-acceptance-runbook.md:64-87` saves the original settings, but the restore command hard-codes `{month:null,quarter:null,year:null}`. That only restores the specific first acceptance run where the original caps were all null. As a reusable runbook, if staging already has `quarter`/`year` or `month` configured, this procedure would erase them while claiming it is "fully reversible". Please change the procedure to restore the original `data.comprehensiveHours.capDefaults` captured in `/tmp/stg-orig.json` (or restore the complete original settings payload through the existing settings route), and adjust step 2's expected result from "saved `{month:600, quarter:null, year:null}`" to "month becomes 600; other cycle types remain whatever the original settings had".

Suggested shape:

- Step 2: assert `month == 600` and `quarter/year == original quarter/year`.
- Step 7: build the restore body from `/tmp/stg-orig.json` rather than hard-coding nulls; then GET and compare the restored capDefaults to the saved original capDefaults.

Why this blocks: this is an ops runbook, and the safety claim is reversibility. A copy-paste operator should not accidentally clear real staging cap defaults.

Other checks look good:

- Single docs-only file under `docs/operations/`.
- Secret hygiene is aligned with the #1804 lesson: token stays in a 0600 file/header file and is not placed directly in curl argv.
- The read-only gates correctly include the #1829 deploy discriminator (`settings` must carry `comprehensiveHours.capDefaults`).
- The acceptance flow correctly verifies PUT/GET settings, natural-month period sync, four snapshot columns, idempotent skip, and cap-change patch.
- Route citations for settings and period-summary sync resolve against `plugins/plugin-attendance/index.cjs`.
- `git diff --check origin/main...origin/pr-1850`: PASS.
- Added-lines scan found no token/JWT value or private key.

Once the restore semantics are fixed, I expect this to pass.

zensgit:
Fixed in \`856a4e161\`. You're right — the hardcoded null restore would wipe pre-existing staging caps while claiming "fully reversible." Corrected:

- **Step 1:** original `capDefaults` is now the explicit source of truth (`ORIG_CAPS`); the runbook states *not* to assume null.
- **Step 2/3:** `month` becomes `600` while `quarter`/`year` **keep their original values** (deep-merge overwrites only `month`); GET-confirm asserts `month==600` **and** `quarter`/`year` unchanged from `ORIG_CAPS`.
- **Step 7:** restore body is built from `/tmp/stg-orig.json` — `PUT` `ORIG_CAPS` (capDefaults always carries all three keys, so it fully reverts steps 2/6), then GET-compare asserts restored `== ORIG_CAPS`. (Full-original-payload restore offered as the equivalent.)
- Pass criteria updated to "restores to exactly the saved `ORIG_CAPS`."

Single docs file, diff-check clean, no token in added lines. Ready for re-review.

zensgit:
Re-review VERDICT: PASS

The prior blocker is fixed in PR HEAD `856a4e16`.

What changed and checks out:

- Step 1 now makes the saved original `data.comprehensiveHours.capDefaults` (`ORIG_CAPS`) the restore source of truth and explicitly says not to assume null.
- Step 2/3 now correctly assert only `month` is overwritten to `600`, while `quarter`/`year` remain unchanged from `ORIG_CAPS`.
- Step 7 now builds the restore body from `/tmp/stg-orig.json` and verifies restored `capDefaults` equals `ORIG_CAPS`, so the runbook is reusable even when staging already has non-null caps.
- The previous hard-coded `{month:null,quarter:null,year:null}` restore is gone.

Other checks remain good:

- Single docs-only file under `docs/operations/`.
- Secret hygiene keeps the token out of curl argv via `-H @/tmp/stg-hdr`; no token/JWT value or private key in added lines.
- Read-only gates, #1829 deploy discriminator, period sync acceptance, idempotency check, cap-change re-sync proof, and troubleshooting table are all aligned with the #1843 evidence.
- `git diff --check origin/main...origin/pr-1850`: PASS.
- CI: 10/10 PASS.

One operational note, not a blocker: the runbook intentionally leaves rebuildable period-summary snapshot rows in place; the reversible state is the settings cap change. That matches the #1843 acceptance posture.

Admin squash is OK from this review.

PR #1859, opened 2026-05-26
feat(attendance): read-only rule_set preview-divergence diagnostic on scheduling workbench

zensgit:
## What

Enhances the existing `GET /api/attendance/advanced-scheduling/workbench` with **one read-only `info` diagnostic** that surfaces a known preview-semantics divergence. 3 files, +149.

A `groupId`-mode preview applies the linked attendance group's `rule_set` (`index.cjs:11452`/`:11462`), while the per-user `effective-calendar` calc-chain (`userId` branch, `:11445`) intentionally does **not**. This diagnostic flags schedule groups whose linked attendance group carries a `rule_set` so operators can see the gap. **It changes no calculation.**

## Boundaries (all held)

Read-only (one `SELECT`); no write to `attendance_*`/`meta_*`; **no migration**; no new POST/PUT/PATCH/DELETE (existing GET enhanced); auto-absence / working-hours / comprehensive-hours / effective-calendar calc-chains untouched; frontend only displays it (generic `v-for`, **no fix button**). This PR takes **no position** on whether the calc-chain should adopt the rule_set — that's a separate product/compliance decision.

## Changes

- **Builder:** new optional `ruleSetAttendanceGroupIds` input (defaults `[]` → backward compatible); flags active schedule groups whose `attendanceGroupId` is in the set; emits `schedule_group_rule_set_preview_divergence` (`info`), appended **last** (preserves existing order assertions).
- **Route:** one read-only `SELECT id FROM attendance_groups WHERE org_id=$1 AND rule_set_id IS NOT NULL`, passed to the builder.
- **Frontend:** none — renders via the existing generic loop (`AttendanceView.vue:4432`).

## Why `info` severity

It fires on every well-configured org that uses rule_sets; `warning` would be permanent noise. Matches `user_mixed_assignment_kinds` (also `info` for a by-design boundary). Message says "(by design — calc chain unchanged)".

## Tests

Workbench test file **8/8** (4 new): positive · negative (linked group without rule_set → no false positive) · backward-compat (no input / no `attendanceGroupId`) · source-guard (new query is `SELECT`; workbench handler **region** has no write verbs — scoped, since the file legitimately writes `attendance_groups` elsewhere via group sync). Full core-backend unit suite **2320/2320**; `tsc` clean. Includes a development/verification MD.

### DIFF
--- a/docs/development/attendance-advanced-scheduling-rule-set-diagnostic-verification-20260526.md
+++ b/docs/development/attendance-advanced-scheduling-rule-set-diagnostic-verification-20260526.md
@@ -25,7 +25,7 @@ by group than the actual per-user effective-calendar resolves. The diagnostic su
 ## Data signal (read-only)
 
 - `attendance_schedule_groups.attendance_group_id` → `attendance_groups.rule_set_id` (both columns already exist; no migration).
-- Route adds: `SELECT id FROM attendance_groups WHERE org_id = $1 AND rule_set_id IS NOT NULL` (query form A — simplest; `attendance_groups` is small and the builder intersects against the active schedule groups anyway). Org-scoped, read-only.
+- Route adds a **schema-safe** read-only lookup, `loadAttendanceRuleSetAttendanceGroupIds(db, orgId)` → `SELECT id FROM attendance_groups WHERE org_id = $1 AND rule_set_id IS NOT NULL` (query form A — simplest; `attendance_groups` is small and the builder intersects against the active schedule groups anyway). Org-scoped, read-only. **Because this is an optional visibility signal, the helper catches `isDatabaseSchemaError` and returns `[]`** (diagnostic absent) on an older schema lacking `attendance_groups.rule_set_id` or the table — so it never fails the whole workbench with `503 DB_NOT_READY`. Non-schema errors still propagate.
 - Builder (`buildAttendanceAdvancedSchedulingWorkbench`): new optional input `ruleSetAttendanceGroupIds` (defaults to `[]` → backward compatible). `scheduleGroupsWithRuleSetOverride` (`:7626`) = active schedule groups whose `attendanceGroupId` is in that set. If any, emit the diagnostic (`:7677`), appended **last** in the diagnostics array (preserves the existing order assertion).
 
 ## The diagnostic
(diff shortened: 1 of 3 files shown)


zensgit:
Codex review — VERDICT: BLOCK until one fallback is added

The slice direction is good, and most boundaries hold: existing GET only, read-only SELECTs, no frontend write affordance, no migration, no calc-chain change, and CI is green.

Blocking issue: the optional diagnostic now makes the whole workbench depend on `attendance_groups.rule_set_id`.

The new query is inside the workbench route `Promise.all`:

- `plugins/plugin-attendance/index.cjs:26204-26208` selects `attendance_groups.rule_set_id`
- if an older/staging schema lacks that column, Postgres raises `42703`
- `isDatabaseSchemaError()` treats `42703` as schema error (`:5392-5395`)
- the route catch returns `503 DB_NOT_READY` (`:26252-26255`)

That means a read-only, informational diagnostic can take down the existing advanced-scheduling workbench on old-schema installs. This is especially risky because this exact column already has a narrow fallback precedent: `loadAttendanceGroupByIdOrCode()` tries `rule_set_id`, then on `isDatabaseSchemaError` falls back to the pre-rule-set query and returns `ruleSetId: null` (`plugins/plugin-attendance/index.cjs:11300-11334`).

Please make this diagnostic fail closed instead of failing the route. Suggested shape:

- wrap the `SELECT id FROM attendance_groups ... rule_set_id IS NOT NULL` in a helper
- return `[]` on `isDatabaseSchemaError(error)`
- rethrow non-schema errors
- pass that empty array into `buildAttendanceAdvancedSchedulingWorkbench`
- add a unit/source test that simulates or locks the `42703` / missing-column fallback for this diagnostic query

Everything else I checked looks sound:

- The divergence evidence is real: `userId` branch does not apply the group rule_set, while `groupId` branch does via `buildAttendanceRuleWithRuleSetOverride()` (`:11445-11462`).
- `info` severity is the right choice; this is by-design visibility, not a violation.
- The frontend renders diagnostics generically, so no frontend diff is needed and no fix button is introduced.
- Added-lines scan did not find new write SQL, migrations, route writes, or secrets.
- `git diff --check` passed.
- PR checks are green: 14 pass + the expected Strict E2E skip.

zensgit:
Fixed in \`b9dfd363e\`. Sharp catch — the inline query in the workbench \`Promise.all\` would have 503'd the entire endpoint on any schema lacking \`attendance_groups.rule_set_id\`, which breaks the "optional visibility" boundary.

- Extracted \`loadAttendanceRuleSetAttendanceGroupIds(db, orgId)\`: on \`isDatabaseSchemaError\` (undefined column/table) it returns \`[]\` → the diagnostic is simply absent and the workbench keeps serving; **non-schema errors still propagate** (so a real DB fault isn't masked).
- Route calls the helper instead of the inline query.
- New test locks the fallback: undefined_column (\`42703\`) → \`[]\`; connection error (\`08006\`) → rethrow; happy path → ids.

Workbench test **9/9**; full core-backend unit suite **2321/2321**; \`tsc\` clean; verification MD updated with the schema-safe note. Ready for re-review.

zensgit:
Codex re-review — VERDICT: PASS after one tiny docs-count correction

The original blocker is fixed correctly in `b9dfd363e`.

What I verified:

- `loadAttendanceRuleSetAttendanceGroupIds(db, orgId)` now wraps the optional `rule_set_id` lookup.
- `isDatabaseSchemaError` returns `[]`, so old schemas simply omit the diagnostic instead of taking down the whole workbench.
- non-schema errors still propagate; the helper does not mask real DB faults.
- route now passes the helper result into `buildAttendanceAdvancedSchedulingWorkbench`.
- tests cover the three important paths: `42703` → `[]`, `08006` → rethrow, happy path → ids.
- read-only / no new route / no migration / no frontend write affordance / calc-chain untouched all still hold.
- CI is green: 14 pass + expected Strict E2E skip.

One remaining docs-only correction before squash:

`docs/development/attendance-advanced-scheduling-rule-set-diagnostic-verification-20260526.md` is internally inconsistent:

- line 44 says the workbench test file is `9/9; 5 new`
- line 50 also says `9/9`
- but lines 54-55 still say `8/8` and `2320/2320`

Please update the Verification run section to the current numbers (`9/9`, `2321/2321`, keeping `tsc` clean). No code changes requested beyond that.

zensgit:
Docs-count fixed in `b947cafd4` — the "Verification run" section now reads 9/9 / 2321/2321, consistent with the "Tests" header. No code change. Proceeding to rebase + admin squash.

PR #1868, opened 2026-05-26
feat(integration): add K3 Material read-only smoke

zensgit:
## Summary

Implements the narrow #1709 read-only runtime slice for K3 WISE Material/GetDetail:

- opt-in `adapter.read()` support only for `object: material` when material operations explicitly include `read`
- calls `POST /K3API/Material/GetDetail` with `{ Data: { FNumber }, GetProperty: false }`
- harvests reference objects from the returned material detail into `_k3ReferenceObjects`, including the customer-observed `FUnitGroupID: { FName }` shape
- rejects missing keys, broad filters, cursor/watermark reads, and BOM reads
- keeps Save/Submit/Audit untouched and asserts no write endpoint is called during the smoke

## Scope Boundaries

Still not included: BOM read, list/search, pagination, broad filters, master-code resolver, server-side pipeline composition, Save/Submit/Audit, or any migration/API route change.

## Verification

- `git diff --check origin/main...HEAD`
- `pnpm -F plugin-integration-core test:k3-wise-adapters`
- `pnpm -F plugin-integration-core test`
- `pnpm verify:integration-k3wise:poc`

Docs added:

- `docs/development/integration-k3wise-material-detail-read-smoke-design-20260526.md`
- `docs/development/integration-k3wise-material-detail-read-smoke-verification-20260526.md`

Fixture is redacted and contains no real K3 host/token/password/material/account values.

### DIFF
--- a/docs/development/integration-k3wise-material-detail-read-smoke-design-20260526.md
+++ b/docs/development/integration-k3wise-material-detail-read-smoke-design-20260526.md
@@ -1,0 +1,113 @@
+# K3 WISE Material/GetDetail Read-Only Smoke Design - 2026-05-26
+
+## Purpose
+
+This slice implements the smallest runtime unlock for #1709: a read-only smoke
+that calls K3 WISE `Material/GetDetail` for one known material number and
+harvests the reference objects already present in that detail payload.
+
+It is not a general source connector. It exists to prove that MetaSheet can read
+one customer-approved material detail without crossing into Save, Submit, Audit,
+BOM, pagination, list reads, broad filters, or server-side pipeline composition.
+
+## Decision
+
+The unlocked runtime surface is:
+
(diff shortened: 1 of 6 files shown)


zensgit:
Addressed the default-off review nit in the PR branch.

Added an explicit keyed negative test:

- default K3 WebAPI adapter
- `read({ object: 'material', filters: { FNumber: 'MAT-DEFAULT-OFF' } })`
- expects `UnsupportedAdapterOperationError`

This locks the important invariant that Material/GetDetail read remains off by default even when the request shape is otherwise valid. Verification doc A1 was updated to mention both default-off checks.

Local rerun after the patch:

- `git diff --check`
- `pnpm -F plugin-integration-core test:k3-wise-adapters`

PR #1867, opened 2026-05-26
test(attendance): make skipped integration suite explicit

zensgit:
## Summary

Closes #1866 first-step visibility gap.

This PR does **not** make the attendance integration suite blocking yet. It makes the current missing-DB path explicit so the check can no longer look like a green, fully executed suite:

- uses `describe.skip` when neither `ATTENDANCE_TEST_DATABASE_URL` nor `DATABASE_URL` is present
- turns setup early-returns for port/DB/table/server failures into clear setup errors
- adds a `beforeEach` guard so a started suite cannot silently pass if `baseUrl` was never initialized

## Why this shape

A real DB probe shows the suite has historical failures once it actually runs, so directly adding `DATABASE_URL` to CI and removing `|| true` would couple #1866 with unrelated test debt. This PR only fixes the misleading "green = silently returned" signal.

## Verification

- `pnpm --filter @metasheet/core-backend test:integration:attendance`
  - Passes with `78 skipped` when no DB env is provided.
- `DATABASE_URL=<local scratch db> pnpm --filter @metasheet/core-backend migrate`
  - Passed.
- `ATTENDANCE_TEST_DATABASE_URL=<local scratch db> DATABASE_URL=<local scratch db> pnpm --filter @metasheet/core-backend test:integration:attendance`
  - Expected current failure: `75 passed / 3 failed`.
  - Failures exposed by the real run:
    - `exports attendance CSV with ISO date values and timezone-consistent timestamps`
    - `exports attendance JSON when format=json is requested`
    - `effective-calendar validates strictly, enforces RBAC, and 404s missing group`

## Follow-up

After those historical failures are handled, a separate PR can wire `DATABASE_URL` into `.github/workflows/plugin-tests.yml` and remove `|| true` from `Run integration tests`.

PR #1861, opened 2026-05-26
fix(attendance): validate UUID route inputs

zensgit:
## Summary
- Split the attendance RBAC wrapper so permission lookup failures still return `Permission check failed`, while handler `HttpError` responses keep their business/validation code.
- Validate malformed UUID route params before DB access for attendance request, integration, import batch, payroll cycle, and holiday routes.
- Validate request body UUID references for `leaveTypeId`, `overtimeRuleId`, and `approvalFlowId`.

## Verification
- `node --check plugins/plugin-attendance/index.cjs`
- `pnpm --filter @metasheet/core-backend exec vitest run tests/unit/attendance-uuid-validation-routes.test.ts --reporter=dot`
- `pnpm --filter @metasheet/core-backend exec vitest --config vitest.integration.config.ts run tests/integration/attendance-plugin.test.ts -t "malformed attendance" --reporter=dot` (current local env has no attendance test DB, so the wider fixture-guarded integration suite remains skipped)
- `git diff --check -- plugins/plugin-attendance/index.cjs packages/core-backend/tests/integration/attendance-plugin.test.ts packages/core-backend/tests/unit/attendance-uuid-validation-routes.test.ts`

Fixes #1853
Fixes #1854
Fixes #1856
Fixes #1858

PR #1836, opened 2026-05-26
docs(attendance): close out comprehensive-hours reporting chain (#1819→#1833)

zensgit:
## What

Docs-only capstone (single new file) for the comprehensive-hours **reporting / snapshot** capability — the design→runtime→snapshot chain after the PR0–PR5 preview/control work.

## Why now

The capability is closed at value-plumbing V1. The remaining items (effective-dating, group/user override, payroll_cycle mapping, UI, more metrics) are explicit **product / compliance decisions**, not kernel-polish — so this records a clean edge before any of them is opted into, and prevents the boundary from being lost.

## Contents

- **Final lineage:** #1819 boundary → #1823 orientation → #1825 cap-policy design → #1829 cap-policy impl → #1833 value-plumbing.
- **Delivered capability** end-to-end (org-default cap → per-period `comprehensive_hours_excess_minutes` + cap companion fields in `attendance_report_period_summaries`, passive re-sync).
- **How to configure** the cap via the existing `PUT /api/attendance/settings` route (minutes, deep-merge, no UI).
- **Resolution + re-sync flow** (date_range bridge → org-default-by-cycle-type → fail-closed stale-null → fingerprint-driven re-sync).
- **Known limitations** (global not per-org; month/quarter/year only; payroll_cycle/custom_range null; one metric; no effective-dating; no UI; producer-level test coverage).
- **Future opt-in list** (each a separate product/compliance decision).

All `index.cjs` citations verified at `a0876ffe4`; all 5 cross-referenced docs exist on main.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1884, opened 2026-05-26
docs(attendance): acceptance smoke-harness consolidation + CI-flake classifier (P3 v1)

zensgit:
## What

P3 consolidation: one ops doc + one reusable read-only tool that **encode the judgment** from this session's acceptance work (so future verifications need less manual reasoning and avoid the specific mistakes that bit us). 6 files; mostly decision rules, cross-referencing existing runbooks rather than copying them.

## Contents

- **`docs/operations/attendance-acceptance-harness.md`** — consolidation map (links #1850 staging-acceptance runbook, workbench runbook, #1810 heartbeat, #1843 login/logout evidence, #1836 closeout) + decision rules:
  - **R1 token hygiene** — 0600 file, `curl -H @header-file` (never argv), delete ≠ invalidate (rotate server-side), staging/prod distinct `JWT_SECRET`.
  - **R2 settings save/restore** — restore from the **saved original**, never a hardcoded value (the #1850 BLOCK: hardcoded nulls wipe pre-existing caps).
  - **R3 infra-flake vs genuine** — the 2–4s-across-unrelated-suites signature + signatures table + the classifier.
  - **R4 green-by-skip 3-point check** — `DATABASE_URL` in step env? `|| true`? test-level `!baseUrl` early-return? (the #1861→#1867→#1872 arc).
  - **R5 wire/save-path drift** — a field isn't "persistable" until a route round-trip proves it (#1829 zod-strip, #1781 serialization-strip).
- **`scripts/ci/ci-flake-classify.sh`** — classify a PR's failing checks as GitHub-infra flake vs genuine using recorded incident signatures. **`gh`-only auth, no token, no prod/staging writes.** Classify-only by default; `--rerun` opt-in **only** when every failure is infra. Exit codes `0` clear / `2` all-infra / `3` genuine / `1` error (chainable into a bounded-retry loop).
- **`scripts/ci/__tests__` + `__fixtures__`** — matcher unit test over recorded incident excerpts (no live-CI dependency). PASS locally.

## Boundaries

Read-only / diagnostic. No new route, no migration, no `attendance_*`/`meta_*` write, no prod/staging mutation, no auto-anything. Scripts use only local `gh` auth.

## Deliberately excluded

The #1810 §3.6 "logout doesn't clear session" finding stays **unverified** in its closeout — not promoted into this runbook.

### DIFF
--- a/docs/operations/attendance-acceptance-harness.md
+++ b/docs/operations/attendance-acceptance-harness.md
@@ -36,7 +36,7 @@ command line.**
 
 ### R3 — GitHub-infra flake vs genuine failure
 - Many checks failing in **2–4 s** across **unrelated** suites = a shared pre-code step failure (Set up job / Checkout), almost always GitHub infra, not your change.
-- Infra signatures: `pnpm/action-setup` / `codeload.github.com` / "could not be found at the URI" / "Failed to download archive" (action-CDN); `HTTP 403` / "unable to access … github" / "RPC failed" / "expected 'packfile'" / git exit 128 (checkout/throttle).
+- Infra signatures are **context-bound** (a bare `HTTP 403` / `exit code 128` is *not* a signature — genuine auth/permission tests print those): action-CDN — `pnpm/action-setup` / `codeload.github.com` / "Download action repository" / "could not be found at the URI" / "Failed to download archive"; git checkout — "unable to access 'https://…github" (URL-bound 403) / "RPC failed; HTTP" / "/usr/bin/git' failed with exit code 128" / "fatal: expected 'packfile'".
 - Tool: `scripts/ci/ci-flake-classify.sh <PR#> [--rerun]` — classify-only by default; exit `0` clear / `2` all-infra / `3` genuine. Re-run **only** when classified all-infra; bound retries (≤3 rounds) and **never merge red CI**.
 - Fail-safe: a failure matching no infra signature is treated as **genuine** (never silently rerun).
 
--- a/docs/operations/attendance-acceptance-harness.md
+++ b/docs/operations/attendance-acceptance-harness.md
@@ -1,0 +1,66 @@
+# Attendance Acceptance Smoke Harness — Consolidation & Decision Rules
+
+One entry point for attendance acceptance/verification. It is **mostly decision rules and
+disambiguators**, not procedures — the step-by-step lives in the existing runbooks (linked
+below); this doc encodes the judgment that was learned the hard way so future acceptances need
+less manual reasoning and avoid the specific mistakes that bit us.
+
+**Auth surface:** the tools referenced here use only the local `gh` CLI or a JWT read from a
+`0600` file. **No script in this harness reads a JWT/service token or puts a secret on a
+command line.**
+
+## Consolidation map (link, don't copy — these are the source of truth for procedure)
+
+| Concern | Runbook / evidence (procedure) |
+| --- | --- |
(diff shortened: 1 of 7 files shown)


zensgit:
Codex review — BLOCKED

One blocking issue in the classifier matcher.

`scripts/ci/ci-flake-classify.sh:29` treats bare `HTTP 403` and `failed with exit code 128` as infra signatures. That is too broad for this repo: an actual product/test failure can legitimately include `HTTP 403` from an attendance/admin/auth endpoint. In that case `is_infra_log` returns true and `--rerun` would classify the job as all-infra even though the failure is genuine. I reproduced it on the PR head:

```bash
bash -lc 'source scripts/ci/ci-flake-classify.sh; if is_infra_log $'"'"'test (20.x) Run unit tests AssertionError: expected HTTP 200 but received HTTP 403 from /api/attendance/settings'"'"'; then echo classified_infra; else echo classified_genuine; fi'
# classified_infra
```

Please make the infra matcher contextual instead of matching naked status/code fragments. For example: action-download signatures should require `pnpm/action-setup` / `codeload.github.com` / `Download action repository` / `Failed to download archive`; checkout signatures should require checkout/git context such as `Checkout repository`, `/usr/bin/git`, `fatal: unable to access`, or the repo URL together with `HTTP 403` / `RPC failed` / `expected packfile` / exit 128. Also add a negative fixture for an app-level `HTTP 403` test failure so this does not regress.

Everything else I checked looks aligned with the P3 boundaries: docs-only + CI helper, no prod/staging writes, no route/migration/write path, and current CI is green. But this classifier is the executable part of the PR, so the fail-safe matcher needs to be tightened before merge.

zensgit:
Fixed in \`3211d4510\`. Real bug — agreed; the bare \`HTTP 403\` / \`exit code 128\` signatures would mark a genuine auth/permission test failure as infra and \`--rerun\` could blindly rerun it.

Every signature is now **context-bound**:
- action-download → `pnpm/action-setup` / `codeload.github.com` / `Download action repository` / `could not be found at the URI` / `Failed to download archive`
- git checkout → `unable to access 'https://…github` (URL-bound 403/auth) / `RPC failed; HTTP` / `/usr/bin/git' failed with exit code 128` (git-bound 128) / `fatal: expected 'packfile'`

So a bare app-level `HTTP 403` (e.g. an RBAC test asserting 403) no longer matches → stays **GENUINE**. Added a negative fixture `genuine-app-403.txt` reproducing your case and a test assertion. Matcher test now **4/4** (2 infra, 2 genuine incl. the app-403 negative). Runbook R3 updated to state the signatures are context-bound. Fail-safe direction preserved (unknown → genuine, never silent rerun).

zensgit:
Codex re-review — PASS.

The blocker is fixed in `3211d4510`. I verified the matcher now stays fail-safe for app-level auth/permission failures:

```bash
scripts/ci/__tests__/ci-flake-classify.test.sh
# 4/4 matcher cases pass: 2 infra, 2 genuine

# app-level HTTP 403 reproduction now classifies as genuine
# classified_genuine
```

I also spot-checked the contextual behavior:
- checkout GitHub 403 with repo/git context => infra
- app/test `HTTP 403` without git context => genuine
- `/usr/bin/git` exit 128 => infra
- generic `node script failed with exit code 128` => genuine

Current CI is green: 10/10 success. No blocking findings remain. Merge condition is just the usual branch-protection/admin squash flow; no red-CI merge needed.

PR #1886, opened 2026-05-26
docs(attendance): P3.1 — "PR CI is red" triage flow for the acceptance harness

zensgit:
## What

Adoption follow-up to #1884 (which landed the harness + classifier). Turns "the tool exists" into "here's exactly how to use it when a PR is red." **Docs-only**, single file (+33/-2).

## Adds

A copy-paste **"Triage flow: a PR's CI is red"** section to `docs/operations/attendance-acceptance-harness.md`:
1. **Classify first** — `scripts/ci/ci-flake-classify.sh <PR#>` (don't rerun/merge yet); exit-code branches `0` clear / `2` infra / `3` genuine / `1` error.
2. **If INFRA, rerun** (opt-in) — `--rerun`, bounded ≤3 rounds, stop if it persists.
3. **Never admin-merge red CI** — even red-by-infra; merge only on a genuine green settle.
4. **Optional bounded auto-loop** snippet that chains on the exit code (`3`=genuine stops, `0`=green stops).

Reinforces the fail-safe: an unmatched failure (e.g. a real `HTTP 403` auth/permission test) is **GENUINE**, never silently rerun.

## Boundaries

Docs-only. No script/CI-wiring/prod/staging change, no new route/migration. The referenced `scripts/ci/ci-flake-classify.sh` is already on main (executable, `gh`-only). Verified the doc's test command runs (matcher PASS).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1887, opened 2026-05-26
docs(attendance): design-lock payroll_cycle comprehensive-hours cap mapping

zensgit:
## What

Design-lock (docs-only, single new file +111) for how a `payroll_cycle` period resolves a comprehensive-hours cap — closing the most visible V1 gap (today it fail-closes to `null`). **Does NOT unlock implementation**; no runtime/migration/new-config/write. Impl is a separate opt-in.

## Load-bearing fact (cited, not asserted)

Payroll templates are **structurally monthly** — migration CHECKs `end_month_offset IN (0,1)` (`:63`) + `start_day`/`end_day BETWEEN 1 AND 31` (`:55`/`:59`) admit at most a one-month-boundary window; there is no quarterly/yearly cadence field. So `month` is the correct cycle-type, and it maps **cross-month pay periods (26th→25th)** correctly — where the existing date-range bridge would mis-map them to `custom_range → null`.

## V1 mapping (locked)

`payroll_cycle` → the **`month`** org-default cap, with two guards:
1. **Template present** — `template_id` is nullable (`onDelete('set null')`, `:80`); no template ⇒ no cadence claim ⇒ `null`. V1 uses **presence only** (no JOIN, no template-config read) — presence is the cadence assertion the `:63` CHECK guarantees.
2. **Span ≤ 62 days** — anomaly guard (62 = legitimate monthly max; hand-edited multi-month spans ⇒ `null`).

`source` = **`payroll_cycle_template_monthly`** (distinct from `org_default_by_cycle_type`). Carries `period.templateId` (mirrors `period.cycleId`); fingerprint payload rides the existing #1833 companion-fields path, so a cap edit re-syncs payroll-cycle rows too.

## Lift point + forward gate

Shows the exact future diff at `buildAttendanceComprehensiveHoursPeriodSummaryValues` `:12140` (fail-closed default preserved for everything else). **Forward gate:** if a future migration relaxes the `end_month_offset` CHECK, this mapping must be revisited.

## Deferred (each its own opt-in)

Quarterly/yearly cadence · per-template/per-cycle override · reading template `config` · effective-dating · template-less cycles (→ `null`). Test matrix P1–P8 included.

All citations grep-verified at `a8ca0917f`.

### DIFF
--- a/docs/development/attendance-comprehensive-hours-payroll-cycle-cap-mapping-design-20260526.md
+++ b/docs/development/attendance-comprehensive-hours-payroll-cycle-cap-mapping-design-20260526.md
@@ -1,0 +1,111 @@
+# Comprehensive-Hours `payroll_cycle` Cap-Mapping — Design Lock — 2026-05-26
+
+**Status:** design-lock (docs-only). **Does NOT unlock implementation.** No runtime change, no
+migration, no new settings/config, no `attendance_*`/`meta_*` write. It locks the contract by
+which a `payroll_cycle` period resolves a comprehensive-hours cap; the implementation is a
+separate, explicit opt-in. Verified against `plugins/plugin-attendance/index.cjs` and the
+payroll migration at `a8ca0917f`.
+
+## 1. The gap
+
+Today a `payroll_cycle` period resolves **no** comprehensive-hours cap. The value-plumbing
+fail-closes at `index.cjs:12140` (`if (!period || period.periodType !== 'date_range') return
+nullValues`) and the cap resolver (`:12093`) only keys `month`/`quarter`/`year`. So a
+period-summary row for a payroll cycle stale-nulls the `comprehensive_hours_*` columns. This is
+the most visible V1 gap (#1825/#1833 deferred it as "cycle-template mapping").
+
+## 2. Load-bearing fact — payroll templates are structurally **monthly**
+
+`attendance_payroll_templates` defines a pay window by `start_day` / `end_day` /
+`end_month_offset`, constrained by CHECKs in `migrations/zzzz20260128120000_create_attendance_rule_sets_and_payroll.ts`:
+
+```
+CHECK (start_day BETWEEN 1 AND 31)      -- :55
+CHECK (end_day   BETWEEN 1 AND 31)      -- :59
+CHECK (end_month_offset IN (0, 1))      -- :63   ← the load-bearing constraint
+```
+
+`end_month_offset ∈ {0,1}` means a template spans at most one month-boundary: a same-month
+window (offset 0) or a cross-month window (offset 1, e.g. the 26th→25th pay period). There is
+**no** quarterly/yearly cadence field. So every payroll cycle is, by schema, a **monthly**
+period. That makes `month` the correct comprehensive-hours cycle-type for a payroll cycle —
+and notably it maps cross-month pay periods (26th→25th) to `month` correctly, where the
+existing date-range bridge (`:12067`) would mis-map them to `custom_range` → null.
+
+## 3. V1 mapping (locked)
+
+A `payroll_cycle` period resolves the **`month`** org-default cap
+(`attendance.settings.comprehensiveHours.capDefaults.month`), with two guards:
+
+1. **Template presence required.** `template_id` is nullable (`onDelete('set null')`, migration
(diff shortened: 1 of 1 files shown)


zensgit:
Codex review — BLOCKED

One design-contract issue: the `templateId + span <= 62` guard is documented as if it guarantees the cycle is genuinely monthly, but the current write paths do not enforce that.

In `plugins/plugin-attendance/index.cjs`, both create and update allow a cycle to carry a `template_id` while using manually supplied `startDate` / `endDate`; they validate only parseability and `start <= end`, not that the dates match the template window:

- create: `24548-24608`, inserts `payload.templateId`, `payload.startDate`, `payload.endDate`
- update: `24784-24833`, keeps/sets `templateId` and accepts supplied dates

The migration does prove payroll **templates** are monthly-ish (`start_day/end_day` checks and `end_month_offset IN (0,1)` at `54-63`), and `template_id` is nullable/on-delete-set-null at `80`. But that does not prove every **cycle row with a template_id** was generated from that template window. A hand-edited cycle with a template and a <=62-day arbitrary range can pass the proposed guard while still not matching the template cadence. So the current wording in the design doc is too strong:

- line 43: "presence of a templateId is the cadence assertion the CHECK at :63 guarantees"
- lines 44-47: span guard says hand-edited multi-month anomalies "must not receive" a month cap and that 62 days avoids genuinely non-monthly spans

Please either:

1. Make the design exact: join/read the template row in the future impl and validate the cycle dates against the template window before mapping to `month`; add this to the lift point/test matrix, or
2. Keep the no-JOIN/presence-only V1, but explicitly document it as a coarse heuristic: `templateId + span<=62` is treated as monthly by policy even though it does not prove exact template-window generation; exact template-window validation is deferred. In that case remove/soften the "must not" / "without admitting genuinely non-monthly spans" claims and add a test-matrix row for a template-bound manual <=62d span so the compromise is intentional rather than accidental.

Everything else I checked is aligned: docs-only, no runtime/migration/write path, current `payroll_cycle` stale-null gap is real, cited cross-reference docs exist, and CI is green. The PR is also currently `BEHIND` after #1886, so it will need rebase + checks after the wording/contract fix.

zensgit:
Fixed in \`05914c269\` (rebased onto latest main — was BEHIND). Took your **Option B** (keep the no-JOIN slice, but make it honestly coarse), after verifying the route behavior:

The POST create handler (`index.cjs:24580`) uses body `startDate`/`endDate` **as-is** — the template window (`resolvePayrollWindow` `:7121`) is only used to *fill in missing* dates — and validates only `start ≤ end` (`:24636`). So `templateId + span≤62` is **necessary, not sufficient** for a monthly window. You're right; my "guarantees / must not / without admitting non-monthly" was over-strong.

- **§3** reframed as a deliberate **coarse heuristic** with an explicit *"Known imprecision"* paragraph: a templateId-bearing, hand-dated, ≤62-day cycle that doesn't match its template window **will** still get the `month` cap. Noted the small blast radius (one reporting metric — excess minutes in the snapshot — never calc-chain/enforcement) and that cycles are normally `auto_generate`-d so dates usually match.
- **§7** added the precise upgrade as a deferred opt-in: read the template + verify dates match `resolvePayrollWindow` → else `null` (removes the imprecision).
- **§8** added row **P5b** making the compromise explicit (V1 → `month`; precise upgrade → `null`).

Docs-only, single file; citations grep-verified; rebased + CI re-running. Ready for re-review.

zensgit:
Codex re-review — PASS, with merge precondition.

The design-contract blocker is resolved. The doc now explicitly frames V1 as a deliberate coarse heuristic rather than a proof: `templateId + span<=62` is necessary but not sufficient, the known imprecision is called out, the blast radius is scoped to the reporting snapshot metric, precise template-window validation is deferred as a separate opt-in, and P5b locks the hand-entered <=62d compromise in the test matrix. The previous over-strong "guarantee / must not / without admitting" language is gone.

I re-checked the underlying facts: payroll template constraints are monthly-ish, cycle rows already expose `templateId`, and current value-plumbing still fail-closes `payroll_cycle` to null. No runtime/migration/write-path change in this PR.

One live-state caveat: the PR is currently `BEHIND` again (`origin/main...origin/pr-1887` is 4/2, and GitHub reports `mergeStateStatus=BEHIND`) even though checks on the current head are green. Please rebase onto latest main and let checks re-run before admin squash. After that, merge is OK from this review.

PR #1894, opened 2026-05-27
feat(attendance): payroll_cycle comprehensive-hours cap mapping (V1 coarse)

zensgit:
Implements the **V1 coarse** `payroll_cycle → month` cap mapping locked in #1887 (design-lock). Runtime slice only — **no migration / route / UI / write path**; the precise template-window validation (design §7) is **deliberately deferred**.

## What it does (design §3/§4)
- **Producer** `resolveAttendanceReportPeriodSyncPeriod` emits `period.templateId` (mirrors `cycleId`) — presence is the coarse monthly-cadence signal.
- **Span guard** `attendancePayrollCycleWithinMonthlySpan` — valid dates, `from ≤ to`, inclusive span `≤ 62` (the conservative ceiling for a structurally-monthly `end_month_offset ∈ {0,1}` template).
- **Builder branch**: `date_range` (unchanged) · `payroll_cycle` + `templateId` + span≤62 → **`month`** org default, `source=payroll_cycle_template_monthly` · everything else → fail-closed `null`.
- **Resolver** gains an optional `sourceLabel` (default `org_default_by_cycle_type`); cap minutes + `effective_key` (=`hash(capDefaults)`) unchanged → a cap edit re-syncs payroll rows exactly like date-range rows. Source literals locked as exported consts.

## Accepted imprecision (locked by P5b)
The cycle create route uses body `startDate`/`endDate` as-is (only `start ≤ end`), so a templateId-bearing cycle with hand-entered ≤62d dates that are **not** a real monthly window **still** gets the `month` cap. Conscious coarse-heuristic trade-off — blast radius is one reporting metric, never calc/enforcement. Precise fix = §7 deferred opt-in.

## Tests — `attendance-comprehensive-hours-control.test.ts` (39 passed, 11 new)
P1 (excess+payroll source) · P2 (cross-month: bridge nulls, payroll branch resolves — proves behavior *change*) · P3 (cap unset→null) · P4 (templateId null→null) · P5 (span>62→null) · **P5b** (date-mismatched → still month, documented imprecision) · span-guard boundary · source-literal lock · P8 (date_range unchanged) · **mandatory producer wire test** (real `resolveAttendanceReportPeriodSyncPeriod`, mocked db — catches a producer dropping `templateId`, cf. #1781 `dayIndex` drift) · P7 (cap-edit re-sync on the **payroll** branch via real sync). Adjacent: `attendance-report-field-catalog` 32/32.

## Boundaries
[...]

PR #1900, opened 2026-05-27
feat(attendance): validate payroll_cycle cap mapping against template window (§7)

zensgit:
Implements the deferred **§7 precise template-window validation** from #1887, replacing the V1 **coarse** heuristic shipped in #1894. A `payroll_cycle` now maps to the `month` comprehensive-hours cap **only when its own dates match the window its template would generate** — a templateId-bearing cycle with hand-entered, mismatched dates no longer gets a cap. Runtime slice only: `plugins/plugin-attendance/index.cjs` + unit suite + verification doc. **No migration / route / UI / write path.**

## What changed vs V1 (#1894)
| | V1 coarse (#1894) | §7 precise (this) |
|---|---|---|
| Gate | `templateId` + span≤62 | `templateId` + `templateWindowMatches===true` + span≤62 |
| Date-mismatched in-span template-bound cycle | resolved `month` (documented imprecision) | **`null`** |
| Template read | none | reads `attendance_payroll_templates`, recomputes window |

## Implementation
- **`verifyAttendancePayrollCycleTemplateWindow(db, orgId, cycle)`** (new): reads the template, recomputes the expected window via `resolvePayrollWindow(mapPayrollTemplateRow(row), parseDateInput(cycle.startDate))` (same call the cycle-generation routes use), compares to the cycle's dates. **Fail-closed** (`try/catch → false`): no templateId / missing template / unparseable dates / any read error → `false`, so an old/partial schema degrades to "unverified → no cap" instead of 503-ing the period sync.
- **Producer** stamps `period.templateWindowMatches` alongside `period.templateId`.
- **Builder** payroll branch requires `templateWindowMatches === true` (strict) + `templateId` + span guard (retained as cheap defense-in-depth). `source=payroll_cycle_template_monthly` and the `date_range` path unchanged.

## Anchor convention
Anchors `resolvePayrollWindow` on the **cycle start date** — identical to the create/generate routes. A template-generated cycle round-trips to its own dates → match; a hand-entered start off the template boundary recomputes a different window → no match.

## Tests — `attendance-comprehensive-hours-control.test.ts` (44 passed, +5 over #1894)
**P5b flipped** from "still month" → **null** (the imprecision §7 removes). New: §7 gate (false/undefined/truthy-non-true → null, strict `===true`); **producer** match / mismatch / no-template / template-row-gone; **verify helper** match/mismatch/no-template/**throwing-db fail-closed**. Producer + verify-helper tests are the wire-vs-fixture guard (builder fixtures set the flag directly and can't catch a producer that mis-computes it). Adjacent `attendance-report-field-catalog` 32/32; `git diff --check` clean.

## Boundaries
One read-only `SELECT` on `attendance_payroll_templates` (fail-closed). No migration/route/UI/raw write. Still deferred: quarterly/yearly cadence (§6), per-cycle override.

Verification doc: `docs/development/attendance-comprehensive-hours-payroll-cycle-cap-precise-window-verification-20260526.md`.

## Production data effect (intentional)
Existing snapshot rows for date-mismatched templateId-bearing cycles (granted a cap by the V1 coarse rule) flip to **all-null** on their next sync: `templateWindowMatches` becomes false → cap/source change → `source_fingerprint` changes → the existing re-sync mechanism patches the row (same path the P7 tests exercise). Intended correction, not a regression; rows self-correct, no backfill.

## Producer uniqueness verified
The only path reaching the snapshot builder with a `payroll_cycle` period is `resolveAttendanceReportPeriodSyncPeriod` (both sync entrypoints use its `resolvedPeriod.period`). The job descriptor (`:2704`) stores only `{cycleId}` and re-resolves for dates; the preview path (`:12385`) uses a caller-supplied cap, not the payroll→month mapping. No inline payroll-period constructor bypasses the strict `templateWindowMatches` gate.

### DIFF
--- a/docs/development/attendance-comprehensive-hours-payroll-cycle-cap-precise-window-verification-20260526.md
+++ b/docs/development/attendance-comprehensive-hours-payroll-cycle-cap-precise-window-verification-20260526.md
@@ -1,0 +1,89 @@
+# Comprehensive-Hours `payroll_cycle` Cap-Mapping — §7 Precise Template-Window Validation — 2026-05-26
+
+Implements the deferred **§7 precise upgrade** from
+`attendance-comprehensive-hours-payroll-cycle-cap-mapping-design-20260526.md` (#1887), replacing
+the V1 **coarse** heuristic (templateId present + span≤62) shipped in the prior slice (#1894). A
+`payroll_cycle` now maps to the `month` cap **only when its own dates match the window its
+template would generate** — a templateId-bearing cycle with hand-entered, mismatched dates no
+longer receives a cap. Runtime slice only: `plugins/plugin-attendance/index.cjs` + its unit suite
++ this doc. **No migration / route / UI / write path.**
+
+## What changed vs V1 (#1894)
+
+| | V1 coarse (#1894) | §7 precise (this slice) |
+| --- | --- | --- |
+| Gate | `templateId` present **and** span≤62 | `templateId` **and** `templateWindowMatches===true` **and** span≤62 |
+| Date-mismatched, in-span, template-bound cycle | resolved `month` (documented imprecision) | **`null`** (no cap) |
+| Template read | none (no JOIN) | reads `attendance_payroll_templates`, recomputes window |
+
+## Implementation
+
+1. **`verifyAttendancePayrollCycleTemplateWindow(db, orgId, cycle)`** (new, near `resolvePayrollWindow`):
+   reads the cycle's template, recomputes the expected window via
+   `resolvePayrollWindow(mapPayrollTemplateRow(row), parseDateInput(cycle.startDate))` (the same
+   call the cycle-generation routes use), and returns `expected.startDate === cycle.startDate &&
+   expected.endDate === cycle.endDate`. **Fail-closed**: no templateId, missing template row,
+   unparseable dates, or any read/compute error → `false` (wrapped in `try/catch`), so an
+   old/partial schema degrades to "unverified → no cap" instead of 503-ing the period sync.
+2. **Producer** `resolveAttendanceReportPeriodSyncPeriod` calls the verifier and stamps
+   `period.templateWindowMatches` (a boolean) alongside the existing `period.templateId`.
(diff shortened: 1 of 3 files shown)


zensgit:
Codex review — PASS.

No blocking findings for `33c14fc3e`.

What I verified:
- The §7 gate is strict: `payroll_cycle` maps to the month cap only when `templateId` is present, `templateWindowMatches === true`, and span is <=62 days. False, missing, or truthy-non-true `templateWindowMatches` stale-nulls.
- The producer now computes and carries `period.templateWindowMatches` from the real `attendance_payroll_cycles` row plus template read; the period-summary route and sync-job path both re-resolve through `resolveAttendanceReportPeriodSyncPeriod` before calling the writer, so I did not find a bypass that feeds a synthetic payroll period into the builder.
- P5b is correctly flipped: template-bound, in-span, date-mismatched cycles now null instead of receiving the coarse V1 month cap. Matching 26th→25th and calendar-month template shapes are covered.
- Date-range behavior and `payroll_cycle_template_monthly` source label stay unchanged; cap edit re-sync still rides the existing source-fingerprint mechanism.
- Boundaries hold: no migration, no route/UI/settings expansion, no new write path, and no raw `attendance_*` / `meta_*` writes. The only new read is the intended `attendance_payroll_templates` lookup in the producer.
- The broad fail-closed template verification behavior is explicit and tested; old/missing template rows or read errors degrade to `templateWindowMatches=false` / stale-null rather than 503.

Local verification I ran:
- `git diff --check origin/main...origin/pr-1900`
- `pnpm --filter @metasheet/core-backend exec vitest run tests/unit/attendance-comprehensive-hours-control.test.ts --watch=false` → 44/44 pass
- `pnpm --filter @metasheet/core-backend exec vitest run tests/unit/attendance-report-field-catalog.test.ts --watch=false` → 32/32 pass

Merge precondition: review-time branch state is 0 behind / 1 ahead. CI is almost green; `test (20.x)` was still in progress on my last poll, while the other real checks were SUCCESS and Strict E2E was the usual conditional SKIPPED. Merge after that final check is green.

PR #1910, opened 2026-05-27
test(attendance): deterministic settle for commit-async skipped summary

zensgit:
Stabilizes the flaky integration test flagged in #1905: `attendance-plugin.test.ts` › *retains compact skipped summary for completed commit-async jobs and idempotent retry*. **Test-only**, exact assertions preserved.

## Root cause (confirmed)
The job response is a **compact** summary (`index.cjs:15246-15253`): `failedRows` is always present but transitions `0→1`, while `skippedCount`/`skippedRows` are **omitted until non-zero** — and all three are written together, a beat *after* `status` flips to `completed`. `waitForImportJobCompletion` returned on the **first** `completed`, so it intermittently read `failedRows=0` / missing `skippedRows`.

## Fix
- `waitForImportJobCompletion` gains an optional **`settle(job)`** predicate applied *after* `completed`, polling a bounded `settleAttempts` budget (default 40×250ms = 10s). If `completed` is reached but the predicate never holds, it throws with the **full job payload** — a genuine missing-summary bug fails loudly and fast, not as a flake. Inline fetch extracted to `fetchImportJob`.
- The failing test passes `settle: (job) => Array.isArray(job?.skippedRows) && job.skippedRows.length >= 1` — a **structural presence** gate (skippedRows is the field absent until the summary lands; once present, `failedRows`/`skippedCount` are final too). It does **not** pre-judge the count, so the exact `toBe(1)` / `toHaveLength(1)` assertions are unchanged.

## Scope
Test/test-helper only; no runtime/route change. Other `waitForImportJobCompletion` callers pass no `settle` → behavior unchanged. The neighboring tolerant `toBeGreaterThanOrEqual(0)` tests are untouched (separate concern). The heavier server-side fix (write the summary atomically with the status) is deliberately deferred. Local: transforms + collects 78 tests (DB-gated → skipped); **proof is CI's DB-backed run**.

Follow-up to the lead doc `attendance-plugin-commit-async-skipped-summary-flake-lead-20260526.md` (#1905).

### DIFF
--- a/packages/core-backend/tests/integration/attendance-plugin.test.ts
+++ b/packages/core-backend/tests/integration/attendance-plugin.test.ts
@@ -149,6 +149,18 @@ function delay(ms: number): Promise<void> {
   })
 }
 
+async function fetchImportJob(baseUrl: string, token: string, jobId: string): Promise<any> {
+  const jobRes = await requestJson(`${baseUrl}/api/attendance/import/jobs/${jobId}`, {
+    method: 'GET',
+    headers: {
+      Authorization: `Bearer ${token}`,
+      'Content-Type': 'application/json',
+    },
+  })
+  expect(jobRes.status).toBe(200)
+  return (jobRes.body as { data?: any } | undefined)?.data
+}
+
 async function waitForImportJobCompletion(
   baseUrl: string,
   token: string,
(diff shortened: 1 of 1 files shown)


zensgit:
Codex review — PASS, no blocking findings.

What I verified:
- Scope is exactly one test file: `packages/core-backend/tests/integration/attendance-plugin.test.ts`; no runtime/product code touched.
- The helper change keeps existing callers backward-compatible: `settle` is optional, and the other `waitForImportJobCompletion(...)` callsites retain their previous "return when completed" behavior.
- The formerly flaky test does not relax the important assertions: `failedRows`, `skippedCount`, and `skippedRows` still assert exact `1` / length `1`. The new predicate only waits for `skippedRows` presence after `status === completed`, then the strict assertions still decide pass/fail.
- Backend write shape matches the flake hypothesis: the async job can mark status `completed` before the compact summary payload is written, while `failedRows`/`skippedCount`/`skippedRows` are then produced together by the summary payload update. Waiting for summary presence is therefore the right test-side stabilization, and the heavier server-side atomic-write fix remains explicitly deferred.
- Failure mode is loud and bounded: if the job reaches `completed` but the compact summary never appears, the helper throws with the last job payload instead of silently accepting a partial state.
- `git diff --check origin/main...origin/pr-1910` is clean.

Verification note: I do not have `ATTENDANCE_TEST_DATABASE_URL` or `DATABASE_URL` locally, so I did not reproduce the DB-backed integration run here. The PR's CI did run the DB-backed integration suite green per the check rollup.

Merge precondition: the branch is currently `BEHIND` (`origin/main...origin/pr-1910` = `3 1`). Please rebase onto latest `main` and require the checks to go green again before admin-squash.

PR #1923, opened 2026-05-27
docs(attendance): design-lock payroll-cycle quarter-year cadence

zensgit:
## Summary

Docs-only design lock for the deferred quarterly/yearly `payroll_cycle` cap-mapping extension.

This locks the key decision for the eventual impl: quarter/year support must come from an explicit `attendance_payroll_templates.cadence` value (`month | quarter | year`), never from span inference. It also records the schema shape, window-generation contract, cap source labels, fingerprint/re-sync behavior, boundaries, and test matrix.

## Why

#1887/#1894/#1900 closed the monthly payroll-cycle mapping. The remaining §6 gap is quarter/year cadence, but the current template schema is monthly-only (`end_month_offset IN (0,1)`) and the generator advances by one month. This doc prevents the next runtime slice from reintroducing a coarse heuristic.

## Verification

- `git diff --cached --check` before commit
- secret/home-path scan against the staged diff: 0 matches for `Bearer`, JWT literal, private-key block, `/Users/chouhua`, token keyword
- cross-reference files verified present on `origin/main`

## Boundaries

- No runtime code
- No migration in this PR
- No route/UI/write-path change
- No per-cycle/group/user override
- No effective-dating or retroactive recompute
- Implementation remains a separate explicit opt-in

### DIFF
--- a/docs/development/attendance-comprehensive-hours-payroll-cycle-quarter-year-cadence-design-20260527.md
+++ b/docs/development/attendance-comprehensive-hours-payroll-cycle-quarter-year-cadence-design-20260527.md
@@ -1,0 +1,223 @@
+# Comprehensive-Hours `payroll_cycle` Quarter/Year Cadence - Design Lock - 2026-05-27
+
+**Status:** design-lock (docs-only). This document does **not** unlock implementation.
+The implementation would require a schema/config change on payroll templates, so it must be
+a separate explicit opt-in with its own review, migration plan, tests, and rollback note.
+
+**Scope:** extend the already-shipped payroll-cycle cap mapping so a payroll-cycle period can
+resolve `quarter` or `year` cap defaults when, and only when, the payroll template explicitly
+declares that cadence and the cycle dates match that template window.
+
+Verified against `plugins/plugin-attendance/index.cjs` and payroll migration at `d5026a04f`.
+
+## 1. Current edge
+
+The current chain is complete for monthly payroll templates:
+
+- #1887 design-lock established `payroll_cycle -> month` mapping because current payroll
+  templates are structurally monthly.
+- #1894 shipped the first coarse implementation.
+- #1900 replaced the coarse gate with precise template-window validation:
+  `templateId && templateWindowMatches === true && span <= 62d` (`plugins/plugin-attendance/index.cjs:12215-12218`).
+
+The remaining §6 gap is quarterly/yearly payroll cadence. Today the table does not model that:
+
+- `attendance_payroll_templates` has `start_day`, `end_day`, and `end_month_offset`, but no
+  cadence field (`packages/core-backend/src/db/migrations/zzzz20260128120000_create_attendance_rule_sets_and_payroll.ts:43-47`).
+- The migration constrains `end_month_offset IN (0, 1)`, which only represents same-month or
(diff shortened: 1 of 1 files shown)


zensgit:
## Independent review — Claude — verdict: **PASS** (docs-only design-lock)

Single file +219/-0, 10/10 CI green. Reviewed against `origin/main` `42fea5daf`.

### Verified
**Citations accurate** (baseline `d5026a04f` is real): §7 gate `templateWindowMatches===true`+span guard → `:12217-12218`; generate `addMonthsToDate(anchorBase, i)` → `:24841`; monthly source label `payroll_cycle_template_monthly` → `:12129` (exact); migration has no `cadence` field, `end_month_offset` default `0` (`:45`), `CHECK (end_month_offset IN (0,1))` (`:63`).

**Discipline consistent with #1887→#1894→#1900:**
- "**Explicit `attendance_payroll_templates.cadence`, never infer from span/name/metadata/offset**" — exactly the lesson #1900 paid for; the §2 rationale is load-bearing and correct.
- QY5 producer wire-vs-fixture marked mandatory (internalizes the #1781/#1900 drift class). Distinct per-cadence source labels. Passive fingerprint re-sync reuses existing fields (consistent with #1833/#1900). Fail-closed default; legacy → `month`. Boundaries explicit; impl a separate opt-in.
- Offset table (month 0\|1, quarter 2\|3, year 11\|12) is internally consistent with the stated `endDay>=startDay → cadenceMonths-1 / else cadenceMonths` rule (hand-verified calendar-quarter offset 2, rolling 26→25 offset 3).

### Should-strengthen before the impl opt-in (non-blocking for this docs PR)
**Make the removal of the `≤62d` span guard explicit.** The current monthly gate is
`templateId && templateWindowMatches===true && attendancePayrollCycleWithinMonthlySpan(from,to)` (`:12218`, span ≤62). A quarter (~90d) / year (~365d) window **fails** that guard — so an impl that generalizes the cadence path but *keeps* the span guard would silently stale-null **all** quarter/year cycles. §5's builder pseudocode correctly omits span (implying removal), but the doc never states "the `:12218` span guard must be dropped — window-match + explicit cadence supersede it." Lock it explicitly in §4/§5. (Minor: `resolvePayrollWindow`'s month-only `offset===0 && endDay<startDay → 1` auto-correct at ~`:7134` can't fire for quarter/year — harmless, but worth confirming the impl doesn't inherit it silently.)

### Bottom line
PASS — citations clean, discipline-consistent, boundaries correct. Adding the one span-guard-removal line closes the only place the doc leaves an impl-critical decision implicit. No runtime/migration here; impl stays a separate gated opt-in.

zensgit:
Applied the review note in 825d3bae7: §4 now explicitly says the existing monthly-only  guard must be replaced with cadence-aware validation for quarter/year, and §5 pseudocode includes a cadence-aware span/window guard. Rebased onto current  and force-pushed; waiting for CI to go green before merge.

zensgit:
Applied the review note in 825d3bae7: §4 now explicitly says the existing monthly-only `span <= 62d` guard must be replaced with cadence-aware validation for quarter/year, and §5 pseudocode includes a cadence-aware span/window guard. Rebased onto current `origin/main` and force-pushed; waiting for CI to go green before merge.

PR #1928, opened 2026-05-27
fix(integration): clarify K3 row-level save failures

zensgit:
## Summary

Follow-up from #1792 after the second M1 Material Save-only attempt on `d5026a04f` failed safely. The live response surfaced only the envelope-level `Successful` message even though the row-level success gate failed, which makes the dead letter misleading.

This PR changes diagnostics only:

- preserves explicit K3 row-level validation messages when present;
- when a Save response fails row-level success but only has a success-like envelope message, synthesizes a structural row-level failure summary instead of surfacing `Successful`;
- keeps the summary redacted and bounded: row count, successful row count, failed row count, success entity count, envelope status, and envelope message presence;
- leaves Save success criteria unchanged.

## Boundaries

- No third M1 Save-only approval.
- No Submit / Audit / BOM / list/search / pagination / multi-record / production.
- No customer dictionary/default population.
- No raw K3 payload, raw material code, token, authorityCode, host, or connection string exposure.

The #1792 write path remains blocked until the actual Save body preview includes the reviewed customer-profile fields.

## Verification

- `node plugins/plugin-integration-core/__tests__/k3-wise-adapters.test.cjs`
- `pnpm -F plugin-integration-core test`
- `git diff --check`
- Added-line secret-shape sweep: clean

## Notes

The full plugin test in the isolated worktree reused the main checkout dependency symlinks for local execution only; those symlinks were removed before commit.

### DIFF
--- a/docs/development/integration-k3wise-m1-save-row-diagnostic-design-20260527.md
+++ b/docs/development/integration-k3wise-m1-save-row-diagnostic-design-20260527.md
@@ -1,0 +1,40 @@
+# K3 WISE M1 Save Row Diagnostic Design - 2026-05-27
+
+## Purpose
+
+After the second M1 Material Save-only attempt, the live K3 response still failed the row-level success gate while surfacing only the envelope-level message `Successful`. That wording is misleading for the operator: the write did not pass, but the dead letter reads like a success.
+
+This change improves diagnostics only. It does not approve another Save-only attempt and does not change the Save success criteria.
+
+## Scope
+
+- Keep the existing row-level success gate unchanged.
+- When a Save response fails row-level success but the only response message is success-like, replace the dead-letter message with a structural failure summary.
+- Preserve useful K3 validation messages when they exist, such as row-level `FMessage`.
+- Keep diagnostics redacted and structural: row count, successful row count, failed row count, success entity count, envelope status, and envelope message presence.
+
+## Non-goals
+
+- No third M1 Save-only approval.
+- No Submit, Audit, BOM, list/search, pagination, multi-record, production, or direct K3 SQL work.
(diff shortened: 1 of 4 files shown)


zensgit:
Review finding — **F1 (fix before merge): Chinese negated-success is misclassified as success-like.**

`isSuccessLikeResponseMessage` uses **exact-match** for English but a **substring** regex for Chinese:

```js
return ['success','successful','succeeded','ok'].includes(normalized)  // exact
  || /成功/.test(normalized)                                            // substring
```

So `操作不成功` / `未成功` / `没成功` (= *failure*) all return `successLike: true` → a genuine Chinese row-level failure message gets **discarded** and replaced by the generic synthesized summary. That is self-defeating for this PR (whose goal is more-informative failure messages), and K3 (金蝶) is a Chinese ERP so these are plausible real messages. English `unsuccessful` is correctly `false` (exact-match), so it's a CN-substring vs EN-exact inconsistency.

**Fix (pick one):** negative lookbehind `/(?<![不未没])成功/.test(normalized)`, or `&& !/[不未没]\s*成功/.test(normalized)`.

**Test gap:** the added test only covers the no-message (envelope-only "Successful") path. Add a case where the K3 row message is `操作不成功` / `未成功` → the surfaced message must **preserve** it, not synthesize the generic one.

Otherwise the PR is good: Save criteria unchanged (only the `!saveSucceeded` message), it composes with #1912 (test locks `diagnostic.validationMessage === message`), and R-REDACT is safe (synthesized message = numeric summary counts only; real-message path still scrubbed). Full review: internal `/tmp/pr1928-review-claude-20260527.md`.

PR #1960, opened 2026-05-28
feat(data-sources): A0 persistence + A0.1 owner scoping (Lane A hardening)

zensgit:
## Summary
- A0: wire `DataSourceManager.initialize(db)` at startup — sources now persist to / load from `data_sources` (was in-memory-only, lost on restart).
- A0.1: per-source owner scoping — list/get/all `:id` ops scoped via `assertAccess` (404 on non-owner), create requires auth, PUT preserves owner. `workspace_id` stored `null` (workspace-shared = follow-up).
- Hardening from review: upsert revives soft-deleted rows + syncs `auto_connect`; create rejects duplicates before persist; new atomic `updateDataSource` (persist-then-swap) so a failed update can't destroy the source.

## Out of scope (separate gated slices)
A-RO (raw `/query` write guard), A1 (credential encryption), A4 (enum/registry/driver reconcile), Lane B (MSSQL connector), Lane C (deploy form).

## Known soft edge
`updateDataSource` tears down the old adapter before constructing the new; only matters if a future adapter's constructor throws (current ones don't). Harden via construct-before-teardown if it ever does.

## Test plan
- [x] 14 unit/route tests (ownership 404, list filter, PUT preservation, persist→reload round-trip, dup-create no-mutation, failed-update atomicity)
- [x] type-check 0 errors, eslint clean

## Heads-up
`GET /api/data-sources/health` still uses unscoped `healthCheck()` and is currently shadowed by `/:id`, so it is unreachable today. When `/health` is fixed later, owner scoping must be added at the same time.

### DIFF
--- a/docs/development/data-sources-mssql-windows-deploy-design-20260527.md
+++ b/docs/development/data-sources-mssql-windows-deploy-design-20260527.md
@@ -1,6 +1,6 @@
-# 外接适配器完善 + MSSQL 多版本支持 + Windows 部署形态 — 设计稿(待审核)
+# 外接适配器完善 + MSSQL 多版本支持 + Windows 部署形态 — 设计稿(已落仓)
 
-> 状态:**草稿 v2(已并入 2026-05-27 review findings + 裁示)**,写在 `/tmp`,待你确认后落 `docs/development/`。
+> 状态:**已落仓 v2(并入 2026-05-27 三轮 review findings + 裁示)**,提交 `7dea4c685`,分支 `codex/data-sources-lane-a-hardening-20260527`(基于 origin/main,未 push)。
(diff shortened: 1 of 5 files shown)


zensgit (commented):
Reviewed the A0/A0.1 hardening end to end. I did not find any new blocking issues in this slice.

What I verified:
- `DataSourceManager.initialize(db)` is wired at startup, so sources now persist to and reload from `data_sources` instead of being memory-only.
- Per-source owner scoping is consistently enforced across list and `:id` routes, with non-owners receiving uniform 404s and unauthenticated create rejected.
- The review-driven hardening fixes are in place: soft-deleted rows are revived on upsert, `auto_connect` stays in sync, duplicate create is rejected before persist, and update is now persist-then-swap so a failed update does not destroy the source.
- Regression coverage is solid for restart round-trips, owner preservation, duplicate-create no-mutation, and failed-update atomicity.

Approving this PR. Before merge, the branch should be rebased onto the latest `main` and revalidated so the final squash lands with fresh green checks against current head.

PR #1980, opened 2026-05-28
feat(attendance): show resolved labels for group members

zensgit:
## Summary
- Resolve labels for the currently loaded attendance-group member page through the existing attendance-scoped batch resolver.
- Render name/email above the immutable userId, while preserving userId as the only identity and write payload fact.
- Fail closed: resolver 403/404/500/network errors silently fall back to saved userIds; member loading, add, and remove flows stay usable.

## Boundaries
- Frontend-only: 2 files.
- No backend, schema, migration, route, permission, OpenAPI, or plugin changes.
- No persisted labels, no enrichment endpoint, no member pagination change, no picker semantics change.
- AttendanceRulesAndGroupsSection remains retired.

## Tests
- pnpm --filter @metasheet/web exec vitest run tests/attendance-admin-anchor-nav.spec.ts tests/attendance-admin-regressions.spec.ts --watch=false

## Review notes
- Locks current-page-only resolution by asserting the resolver body is the visible member userId list.
- Locks same-label/different-userId behavior by keeping two rows distinct and removing by userId.
- Locks resolver failure fallback by asserting no label is rendered and userId remains visible.

### DIFF
--- a/apps/web/src/views/AttendanceView.vue
+++ b/apps/web/src/views/AttendanceView.vue
@@ -3168,14 +3168,31 @@
                         <table class="attendance__table">
                           <thead>
                             <tr>
-                              <th>{{ tr('User ID', '用户 ID') }}</th>
+                              <th>{{ tr('Member', '成员') }}</th>
                               <th>{{ tr('Joined', '加入时间') }}</th>
                               <th>{{ tr('Actions', '操作') }}</th>
                             </tr>
                           </thead>
                           <tbody>
                             <tr v-for="member in attendanceGroupMembers" :key="member.id">
-                              <td>{{ member.userId }}</td>
+                              <td>
+                                <span class="attendance__group-member-identity" data-attendance-group-member-identity>
+                                  <strong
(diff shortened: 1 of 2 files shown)


zensgit:
Applied the should-fix for #1974 §5.2.

What changed:
- `attendanceGroupMemberResolverUserIds()` now pre-filters to UUID-shaped user IDs before calling `/api/attendance-admin/users/batch/resolve`.
- Non-compatible/legacy member IDs are skipped by the resolver and keep the row-level `userId` fallback, so a future strict resolver cannot turn one bad ID into a whole-page label failure.
- Updated label-path fixtures to use UUID-shaped IDs, and added a regression test that a `legacy-user-1` member does not call the resolver and still renders its `userId`.

Verification:
- `pnpm --filter @metasheet/web exec vitest run tests/attendance-admin-anchor-nav.spec.ts tests/attendance-admin-regressions.spec.ts --watch=false` -> 58/58 pass
- `pnpm --filter @metasheet/web type-check` -> pass
- `git diff --cached --check` before amend -> pass

Still frontend-only: same 2 files, no backend/schema/route/permission changes.

PR #2048, opened 2026-05-29
docs(integration): DF-T3b from_reference_table mapping-sheet resolver — design & acceptance

zensgit:
## DF-T3b — `from_reference_table` mapping-sheet resolver — design & acceptance (design-first)

Per owner: T3b is one tier riskier than T3a (it **changes the composed Save body**), so lock a short implementation design / acceptance MD **before** any runtime.

### Scope boundary (load-bearing)
Changes **how** `from_reference_table` composes a reference (per-material mapping-sheet lookup → full `{FNumber,FName}`/`{FID,FName}`, lifting #1824) in the **shared composition path**, byte-parity with Save. Does **NOT** change the write **scope** — still M1 one-record Material Save-only; no Submit/Audit/BOM/multi-record/new endpoint.

### Grounded against verified code (not memory)
- `from_reference_table` today = `http-routes.cjs:674` uniform pre-resolved scalar (= the #1824 limit).
- Shaping = `applyReferenceShape` (`k3-save-body-composer.cjs:44`, DF-T1-0 single source) — object passes through verbatim → **no new shaper**.
- Live adapter Save = schema path (`composeSchemaBody`/`projectRecordForBody`); does not consume fieldRules.
- The existing parity test is **behavioral byte-equality** but exercises the **schema path**; the `from_reference_table` (fieldRules) path is **not yet under that parity** → bringing it under is **T3b's keystone** (extend the behavioral test, don't write a new ad-hoc one).

### The nine locks
bulk read + per-run in-memory index (**no cross-run cache**) · enabled-only · blank `sourceCode` ignored · 0→unresolved · 2+ enabled-complete → **ambiguous fail-closed (never pick-first)** · missing `fNumber`/`fID`/`fName`→unresolved · **preview ≡ Save byte-parity** test · evidence = field/domain/`sourceCode`-presence/error-type only (**no values**) · no K3 write/Submit/Audit/BOM/multi-record.

### Deferred to impl (like #2036)
Exact source-adapter read call shape + the exact shared injection point. Decomposes **T3b-1** (pure resolver, latent) → **T3b-2** (wire + extend parity test, byte-parity-gated own opt-in) → **T3c**.

Design-only — no runtime, no code. Left for owner review.

### DIFF
--- a/docs/development/data-factory-df-t3b-reference-resolver-design-20260529.md
+++ b/docs/development/data-factory-df-t3b-reference-resolver-design-20260529.md
@@ -1,0 +1,110 @@
+# DF-T3b — `from_reference_table` mapping-sheet resolver — implementation design & acceptance (2026-05-29)
+
+> **Design-first, no runtime in this PR.** DF-T3b lifts the #1824 uniform-reference limit by composing
+> a **per-material** K3 reference object from a mapping sheet (DF-T3a manifest, #2043) instead of the
+> one authored sample's uniform scalar. Because it **changes what gets composed into the Save body**, it
+> is one tier riskier than the schema-only T3a — so the contract is locked here before any code.
+
+## Scope boundary (the load-bearing sentence)
+
+T3b changes **how** `from_reference_table` composes a reference — a per-material mapping-sheet lookup
+producing the full `{FNumber,FName}` / `{FID,FName}` object, vs. today's uniform pre-resolved scalar —
+in the **shared composition path**, so the no-write preview stays **byte-identical** to the adapter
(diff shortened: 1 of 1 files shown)


zensgit:
Folded in the owner's non-blocking suggestion: the T3b-2 parity acceptance now **requires a negative control** — a case where the preview resolver and the Save resolver are intentionally desynchronized and the byte-parity assertion **fails** (proving the test catches resolver divergence, not passing vacuously). Added to both the parity section and the acceptance checklist. Force-pushed (`0fb85060e`), 0-behind.

PR #2063, opened 2026-05-29
feat(integration): DF-T3b-2a resolver wire into target preview + resolved-record parity

zensgit:
## DF-T3b-2a — resolver wire into target preview + resolved-record schema parity + 3-state fail-closed

Second slice of DF-T3b (#2048 design; resolver #2055). Wires the T3b-1 pure resolver into the composed body. **Option (a) + single-representation** (owner-confirmed).

### What this PROVES (precise — option a, not b)
- **`schema-compose(resolvedRecord) ≡ adapterSave(resolvedRecord)`** byte-parity for a `from_reference_table` reference (extends `k3-save-body-composer.parity.test.cjs`, both `projectRecordForBody`).
- **The operator preview (`buildTargetPayloadPreview`) resolves and fail-closes correctly** (its own tests in `http-routes.test.cjs`).
- It does **not** establish byte-parity between `buildTargetPayloadPreview` and the Save directly — that's option (b), explicitly deferred.

### Production state: dormant until T3b-2b
The route does **not** pass indexes and the adapter Save isn't resolver-wired — both stay dormant (`from_reference_table` → unresolved sentinel) until **DF-T3b-2b** sources indexes from a live bulk-read. Parity is proven through the test harness feeding a resolved record, not a live Save.

### Key decisions
- **Single-representation**: the resolved `{FNumber|FID,FName}` rides in the record through the existing schema (`applyReferenceShape` passthrough) — **no new schema descriptor** (avoids DF-T1-0 dual-representation drift).
- **Fail-closed = sentinel, not drop**: schema-path preview has no bodyTemplate + adapter Save only throws on `findUnfilledPlaceholders`, so a dropped field is silently omitted. `'<unresolved>'` makes `findUnfilledPlaceholders` fire on **both** sides (same `K3_WISE_PRESET_PLACEHOLDER_UNFILLED`) → identical disposition for all 3 statuses (single `status !== 'resolved'` gate).
- **Shared decision fn** `resolveReferenceRuleValue` called by both the preview and the record materializer → can't diverge. Indexes are a **server-side options param** (not client-injectable).
- `from_reference_table` semantics change is safe — **zero scalar-dependent producers** (authoring uses `preserve_template`; T2a never emits it).

### Tests + negative controls
Parity two regimes (resolved → byte-equal; non-resolved → identical fail-closed disposition). Controls: **wire** (ignore index → operator test fails), **sentinel** (blank → fail-closed breaks both sides), **desync** (different indexes → parity catches divergence). Affected suites green locally; full 28-step chain runs in CI.

Verification MD: `docs/development/data-factory-df-t3b-2a-resolver-wire-parity-development-20260529.md`. Left for owner review — no auto-merge.

### DIFF
--- a/docs/development/data-factory-df-t3b-2a-resolver-wire-parity-development-20260529.md
+++ b/docs/development/data-factory-df-t3b-2a-resolver-wire-parity-development-20260529.md
@@ -1,0 +1,128 @@
+# DF-T3b-2a — resolver wire into target preview + resolved-record schema parity + three-state fail-closed — development verification (2026-05-29)
+
+> Second slice of DF-T3b ([#2048](docs/development/data-factory-df-t3b-reference-resolver-design-20260529.md)
+> design; resolver = #2055). Wires the T3b-1 pure resolver into the composed Save body and locks
+> **preview ≡ Save byte-parity** for a `from_reference_table` reference. The live mapping-sheet
+> bulk-read / fuller pipeline wiring is **DF-T3b-2b** (a separate opt-in).
+
+## Locked scope (owner-set)
+
+In:
+- **Resolver wire into the target preview** — `from_reference_table` in `buildTargetPayloadPreview`
+  (the operator/DF-T2c surface) now resolves a per-material reference via an injected mapping index
+  instead of the old uniform `raw = rule.value` scalar (the #1824 limit).
+- **Resolved-record schema parity** — a shared step materializes the resolved reference INTO the
+  record; the parity test proves it composes **byte-identically** through the schema-path preview and
+  the real adapter Save.
+- **Three-state fail-closed** — `unresolved` / `ambiguous` / `incomplete` all block both compose paths
+  via the single `status !== 'resolved'` gate.
+
+Explicitly **out** (DF-T3b-2b / later):
+- ❌ No live source-adapter bulk-read (indexes are injected; the route does **not** source them yet).
+- ❌ No fuller pipeline wiring; ❌ no K3 write / Submit / Audit / BOM / multi-record; ❌ no UI /
+  authoring / import / export; ❌ no frontend change.
+
+## Architecture decisions (grounded, owner-confirmed)
+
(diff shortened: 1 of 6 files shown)


zensgit:
**P1 (owner review) fixed: sourceField extraction divergence.** The shared *decision* fn was shared, but the sourceCode *extraction* before it was not — operator preview used `getPath` (nested), the materializer used flat `out[sourceField]`. A nested `sourceField` (e.g. `source.unitGroup`) resolved in preview but sentinel'd in the materializer → exactly the preview-green/Save-fail-closed drift, latent until T3b-2b. Fixed: `resolveReferenceRulesIntoRecord` now reads/writes via the SAME `getPath`/`setPath` semantics. Added a permanent cross-path test (`testNestedSourceFieldNoDivergence`) asserting preview ≡ materializer for a nested sourceField + a 4th negative control (revert to flat read → that test fails). So 'shared decision cannot diverge' is now genuinely closed (extraction + decision). Force-pushed `bb82baa08`, 0-behind.

zensgit:
**P2 (owner review) fixed: read-snapshot divergence on overlapping rules.** Same class as P1, one level deeper — the materializer read its source from `out` (which earlier rules rewrite), while the operator preview always reads the original `sourceRecord`. Reproduced your case (rule 1 `targetField:'source.unitGroup'` writes the path rule 2 reads → preview resolved, materializer `<unresolved>`, statuses [resolved, unresolved]). Fixed: `resolveReferenceRulesIntoRecord` now reads every `sourceField` from the **original record snapshot** and writes to an **independent deep clone** — earlier writes can't change later reads, and the input record is never mutated. **Invariant locked: read snapshot = original sourceRecord; write = materialized out.** Added a resolver-level overlap test (both resolve + input unmutated) + cross-path `testSourceTargetOverlapNoDivergence` (preview ≡ materializer) + a 5th negative control (read from mutated out → both overlap tests fail). Force-pushed `247bcaf1b`, 0-behind.

PR #2073, opened 2026-05-29
feat(integration): DF-T3b-2b live mapping-sheet bulk-read → preview-route resolve (read-only, API-reachable)

zensgit:
## DF-T3b-2b — live mapping-sheet bulk-read → preview-route resolve (read-only)

Makes the preview **route** resolve `from_reference_table` from **real** mapping sheet rows **when the request supplies `referenceMappingSources`**: bulk-read via the staging source-adapter → build `referenceMappingIndexes` → feed the DF-T3b-2a (#2063) seam. **READ-ONLY** — no pipeline-runner / `upsert` / K3 Save (real-Save wire = DF-T3b-2c).

### Reachability (precise — no overclaim)
**API-reachable, NOT operator-UI-reachable.** No shipped client sends `referenceMappingSources` yet (`IntegrationWorkbenchView.previewPayload` doesn't — UI was out of scope), so an operator clicking Preview still gets no resolution until a **separate, not-yet-scoped UI wire** sends the sources. 2b ships the backend capability + the API surface.

### What shipped
- `reference-mapping-source.cjs` (new): `bulkReadRows` loops the staging adapter's paginated `read()` until `done`, **bounded by `maxPages` and THROWS on exceed** (never silently truncate — a truncated dictionary would resolve real codes to `unresolved`); `buildReferenceMappingIndexes` per `{domain,object,template}` via the T3b-1 resolver. Per-run, no cross-run cache. **Reuses** `createMetaSheetStagingSourceAdapter` — no new sheet reader (#2036: read by `sheetId`).
- `http-routes.cjs`: `templatesPreview` wires `referenceMappingSources [{domain,systemId,object}]` → **tenant-scoped** system load → `adapterRegistry.createAdapter` → live bulk-read → preview options seam. No sources → unchanged. Unknown domain → 400. **Index never client-supplied.**

### Binding — grounded against #2036
#2036 specifies reading "the same way the staging source-adapter already reads cleansing sheets … by `sheetId`" and deferred the read-call shape to T3b. `referenceMappingSources` is that binding — not an invented config.

### wire-vs-fixture (keystone)
`reference-mapping-source.test.cjs` drives the **real** staging adapter with a mocked `recordsApi.queryRecords`: asserts paginated bulk-read ran (`queryRecords` per page, offset advances `[0,1,2]`, stops on `done`, targets the `sheetId`) + index resolves. **Negative control**: bypass the bulk-read → `queryRecords` count `0` fails. `maxPages` cap throws. Route test proves the live bulk-read through the registry + three-state fail-closed at the preview layer + unknown-domain 400.

### Carry into DF-T3b-2c (on record)
Per-row fail-closed at K3-call → clean dead-letter/provenance (not batch-abort, not auto-retry-hammered); 2c builds indexes from the **same bindings** the preview uses (config-layer parity); adapter `read` I/O failure is a separate category to handle at the Save path.

Verification MD: `docs/development/data-factory-df-t3b-2b-live-bulk-read-preview-development-20260529.md`. Left for owner review — no auto-merge.


zensgit:
**P1 + P2 (owner review) fixed pre-merge.**

**P1 — bulk-read entry point fail-closed to `metasheet:staging` only.** Without it, the preview route was an arbitrary-adapter `read()` entry point — a caller passing a K3/other external `systemId` (any adapter with `read()`) would trigger an external read, not a read-only workspace mapping-sheet read. The route now throws `INVALID_TEMPLATE_PREVIEW` **before `createAdapter`** when `system.kind !== 'metasheet:staging'`. Test asserts a non-staging system is rejected **and** `createAdapter`/`queryRecords` are **never called**.

**P2 — duplicate `referenceMappingSources` domains fail closed (400).** One sheet per domain (#2036); `Object.assign` would otherwise let the last binding silently win. Normalizer rejects a repeated domain before any adapter/read. Test asserts the 400 + no adapter created.

Both negative-controlled (disable each guard → its test fails) + a happy-path control (valid staging source still creates the adapter and reads). Force-pushed `909f87a94`, 0-behind.

PR #2130, opened 2026-05-30
feat(multitable): A6-1 persistent WorkflowJob runtime (opt-in linear job plane)

zensgit:
## What

Implements the A6-1 scout (**#2112**) after explicit owner opt-in. **First A6 capability-half runtime**: opted-in rules (`automation_rules.execution_mode='workflow_job_v1'`) persist one **C1-shaped job per action**; existing rules stay on the legacy steps path by default.

- **migration** `zzzz20260530120000`: new `multitable_automation_jobs` (+ `execution_id` index) + nullable `automation_rules.execution_mode`.
- **`ActionJobLifecycle`** (onStart/onSettled/onSkipped) threaded `execute(rule,evt,factory?)`→`executeActions`; hooks fire **OUTSIDE the inner per-action try/catch** so a job-write failure fails the EXECUTION (fail-closed), not a swallowed step.
- **`AutomationJobService`**: `lifecycleFor` (running on start; resolved/failed via C1 bridge on settle; skipped on fail-stop; `result` via `redactValue`, `error` via `redactString` — A1 invariant) + `listByExecution` (C1 views, enum-strict).
- `executeRule` builds the factory **only** for opt-in rules (single path-choice point → retries of opt-in rules also write jobs). A2 detail prefers persisted jobs, falls back to legacy steps (list unchanged, no N+1).

## ⚠️ Latency — runtime present, enable path deferred (dormant in prod)
Read path fully wired (all rule reads use `.selectAll()` → `execution_mode` reaches `executeRule` at runtime — **verified, not dead-on-arrival**). But **no write path sets it**: `createRule`/`updateRule` deliberately don't accept `execution_mode` (enable path scout-deferred). So the job plane is **dormant in production** — a rule is opt-in-able only by **direct DB / fixture**, not the API. A follow-up (its own opt-in) adds the enable writer. Intended: runtime lands + is provable; turning it on is a separate reviewable step.

## Fail-closed (opt-in durable provenance)
Stricter than the legacy log-write swallow / the `/test`+retry fail-OPEN (#2053), deliberately because A6-1 is opt-in: **(a)** onStart throw → side effect **never runs** + execution failed; **(b)** onSettled throw *after* the action → side effect **ran once** + execution failed ("must not pretend the action didn't run"). **Operator-visible (scout-accepted):** a job-write blip turns a side-effect-completed opt-in run into `failed` → an A5 retry can double-fire. Surfaced in the verification doc.

## Scope held
NO suspend/resume / branch/parallel / BPMN / approval-as-job / worker / default per-action writes — **A6-2..A6-5 stay frozen / demand-gated**. Lock: post-GATE (#1993); A6-1 governed by the run-governance demand gate, not K3; no integration-core/RBAC/K3-write touch.

## Tests
**181 unit** (executor lifecycle ordering / fail-stop skipped / **both fail-closed sub-cases** / opt-out-no-jobs-byte-identical / service path-choice / job persistence+redaction+C1 read / A2 detail prefer-jobs-vs-legacy) + **real-DB round-trip** `multitable-automation-jobs.test.ts` wired into `plugin-tests.yml` (closes the new-table wire-vs-fixture gap). `tsc` clean; eslint clean for new files (one pre-existing `prefer-const` in `parseCreateRuleInput` left as-is — core-backend has no lint script).

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -177,6 +177,7 @@ jobs:
             tests/integration/multitable-viewconfig-resave-guard.test.ts \
             tests/integration/multitable-records-list-authz.test.ts \
             tests/integration/multitable-automation-retry.test.ts \
+            tests/integration/multitable-automation-jobs.test.ts \
             --reporter=dot
 
       - name: Start core backend (background)
--- a/docs/development/multitable-automation-a6-1-workflowjob-runtime-verification-20260530.md
+++ b/docs/development/multitable-automation-a6-1-workflowjob-runtime-verification-20260530.md
@@ -1,0 +1,47 @@
+# Multitable Automation A6-1 — Persistent WorkflowJob Runtime (verification) — 2026-05-30
+
+- **Slice**: A6-1 — opt-in persistent linear WorkflowJob runtime. Implements the A6-1 scout (#2112) after explicit owner opt-in ("启动 A6-1 runtime"). First A6 capability-half runtime.
+- **Status**: ✅ implemented + verified. A6-2+ (suspend/resume, branch/parallel, BPMN, approval-as-job) remain frozen / demand-gated.
+- **Lock**: post-GATE (K3 blanket lock retired, #1993); A6-1 governed by the run-governance demand gate, not K3. Automation is multitable-kernel (no integration-core / RBAC / K3-write touch).
+
+## What shipped
+
+| File | Change |
+|---|---|
+| migration `zzzz20260530120000` | new `multitable_automation_jobs` (+ `idx…execution_id`) + nullable `automation_rules.execution_mode`. |
+| `db/types.ts` | `MultitableAutomationJobsTable` + `execution_mode` on the rules table + Database registration. |
+| `automation-executor.ts` | `AutomationRule.executionMode?`; `ActionJobLifecycle` (onStart/onSettled/onSkipped) + `ActionJobLifecycleFactory`; `execute(rule, evt, jobLifecycleFactory?)`; `executeActions(..., jobLifecycle?)` fires hooks **outside the inner per-action try/catch**. |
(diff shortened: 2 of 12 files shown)


zensgit:
## Reviewer aid — the two things most worth your attention

**1. Fail-closed semantics (the one deliberate behavior change for opt-in rules)**
The job lifecycle hooks fire OUTSIDE the inner per-action try/catch, so a job-write failure fails the **execution**, not a swallowed step. This is *stricter* than the legacy log-write swallow and the `/test`+retry fail-OPEN (#2053) — on purpose, because A6-1 is opt-in durable provenance. Two pinned sub-cases:
- (a) `onStart` throw → side effect **never runs** + execution `failed`;
- (b) `onSettled` throw *after* the action → side effect **ran once** + execution `failed` ("must not pretend the action didn't run").

**Operator-visible trade (scout-accepted):** a transient job-write blip turns a side-effect-completed opt-in run into `failed` → an A5 retry can then double-fire those side effects. Conscious, documented in the verification doc — not a surprise.

**2. Latency — runtime present, but dormant in production**
Read path is fully wired (all `automation_rules` reads use `.selectAll()`, so `execution_mode` reaches `executeRule` at runtime — verified, **not** dead-on-arrival). But `createRule`/`updateRule` deliberately do **not** accept `execution_mode` → the job plane is **dormant**: a rule is opt-in-able only by direct DB / fixture, not the API. The enable writer is a separate follow-up opt-in. So merging this does **not** turn jobs on for anyone — it lands + proves the runtime, gated-off by default.

**Boundary:** A6-2..A6-5 (suspend/resume, branch/parallel, BPMN, approval-as-job) stay frozen/demand-gated; A6-1 runtime ≠ authorization for A6-2.

181 unit (incl. both fail-closed sub-cases + opt-out byte-identity + A2 detail prefer-jobs) + real-DB round-trip wired into `plugin-tests.yml`. Holding merge for your review.

PR #2136, opened 2026-05-30
feat(attendance): admin workbench T4 — picker-backed scope targets (users + groups)

zensgit:
## What
Replaces the scheduler-scope form's **free-text target fields with real pickers**, so an admin no longer types raw UUIDs to scope a sub-admin (after read-only **T1 #2067** / create **T2 #2108** / edit+deactivate **T3 #2123**):

- **Users** → a new `AttendanceUserMultiPickerField` (search the directory via the shared `useAttendanceAdminUsers` composable, add results as **removable chips**), emits `string[]`.
- **Attendance groups + schedule groups** → checkbox multi-selects over the already-loaded `attendanceGroups` / advanced-scheduling-workbench `scheduleGroups` lists, **mirroring the existing actions checkbox-array pattern** (no new component for these).
- **Departments / roles / role tags** stay free-text (open vocabulary, no enumerable source); the form hint now describes the mixed input.

## No-silent-caps (attendance groups only — verified, not assumed)
The attendance-group loader caps at `pageSize 200` while the endpoint reports the real `COUNT(*)` (plugin-attendance `index.cjs:26237`), so a hint shows when the loaded list is shorter than the total — otherwise a group past the cap would be silently un-pickable. I **read the schedule-group source rather than assuming**: the workbench returns schedule groups **uncapped** and its `total` echoes `items.length` (`index.cjs:7851`), so the schedule-group picker **needs no hint and gets none** (no dead guard).

## Model / wire
The three picked fields move from `string` to `string[]`; submit/edit/reset spread the arrays. **All six targets still land as arrays under the same scope keys**, so the backend `.strict()` shape and the `>=1 target` guard are unchanged. **Pure frontend; no backend / enforcement touch.**

## Tests (gated `attendance-admin-regressions.spec.ts`)
- **create** picks a user + both group kinds + free-text dept/role/role-tag → asserts the full strict body round-trips (`toEqual`, distinct sentinel per key — the A2 wrong-key guard),
- **edit** prefills the group checkboxes + user chip from array targets and round-trips them on an unchanged PUT (the `string`→`string[]` hydration is the likeliest silent regression),
- a **capped** attendance-group source surfaces the hint; an **uncapped** schedule-group source carries none,
- **removing** the only user chip empties the target so the create guard refuses (covers the chip-remove path).

## Verification
`vue-tsc -b` clean · web **build** clean · gated specs green (**regressions 56/56**, **anchor-nav 30/30**). ~472 net lines. **Out of scope**: department/role/role-tag pickers (no source), enforcement.

> Not auto-merging — opened for your review first.

### DIFF
--- a/apps/web/src/views/AttendanceView.vue
+++ b/apps/web/src/views/AttendanceView.vue
@@ -1353,18 +1353,6 @@
                     <span>{{ tr('Departments', '部门') }}</span>
                     <textarea id="attendance-scheduler-scope-target-departments" v-model="schedulerScopeForm.departments" rows="1" :placeholder="tr('dept-a, dept-b', 'dept-a, dept-b')"></textarea>
                   </label>
-                  <label class="attendance__field" for="attendance-scheduler-scope-target-attendance-groups">
-                    <span>{{ tr('Attendance groups', '考勤组') }}</span>
-                    <textarea id="attendance-scheduler-scope-target-attendance-groups" v-model="schedulerScopeForm.attendanceGroupIds" rows="1" :placeholder="tr('attendance group id(s)', '考勤组 ID')"></textarea>
-                  </label>
-                  <label class="attendance__field" for="attendance-scheduler-scope-target-schedule-groups">
-                    <span>{{ tr('Schedule groups', '排班组') }}</span>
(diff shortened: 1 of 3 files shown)


zensgit:
Superseded by 6e5e3d032 ("feat(attendance): add scheduler scope target pickers"), which already landed a more unified all-six-targets chip UI on main. Closing rather than force-merging this parallel implementation. The one genuine delta here — the no-silent-caps truncation hint for the attendance-group picker (its source caps at pageSize 200 while the endpoint reports the real COUNT(*)) — will land as a small follow-up against the merged implementation. Branch kept on remote for reference.

PR #2154, opened 2026-05-30
feat(web): data-sources UI-3 non-secret edit flow

zensgit:
## Summary

Adds the next data-sources CRUD slice after #2147/#2151: editing non-secret data-source configuration from the UI.

- Adds a typed `GET /api/data-sources/:id` detail client and `PUT /api/data-sources/:id` update client
- Adds store actions for loading a sanitized detail and updating non-secret config
- Adds row-level `编辑` in `DataSourcesView`
- Reuses the create form in edit mode with `id`/`type` locked
- Hides credential inputs in edit mode and never sends `credentials` in the PUT payload

## Scope / boundaries

- Frontend only
- Updates name, connection fields, and `readOnly`
- Does not rotate credentials; backend update schema does not accept credentials yet
- Does not add query/select/schema execution UI
- Does not add connect/disconnect write buttons

## Verification

- `pnpm --filter @metasheet/web exec vitest run tests/data-sources-ui.spec.ts --watch=false`
- `pnpm --filter @metasheet/web exec vue-tsc --noEmit`
- `pnpm --filter @metasheet/web exec eslint 'src/data-sources/*.ts' src/stores/dataSources.ts src/views/DataSourcesView.vue tests/data-sources-ui.spec.ts --max-warnings=0`
- `git diff --check`

## Test notes

Adds builder/store/component coverage for the edit path. The component test mounts `DataSourcesView`, clicks `编辑`, loads a sanitized detail, verifies password inputs are not rendered, submits, and asserts the PUT payload is credential-free.

### DIFF
--- a/apps/web/src/data-sources/api.ts
+++ b/apps/web/src/data-sources/api.ts
@@ -1,10 +1,21 @@
 // Typed client for the generic external data-source connector API.
 import { apiGet, apiFetch } from '../utils/api'
-import type { CreateDataSourcePayload, DataSourceListItem, DataSourceTestResult } from './types'
+import type {
+  CreateDataSourcePayload,
+  DataSourceDetail,
+  DataSourceListItem,
+  DataSourceTestResult,
+  UpdateDataSourcePayload,
+} from './types'
 
 interface ListEnvelope {
   ok: boolean
   data?: { items?: DataSourceListItem[]; total?: number }
+}
+
+interface DetailEnvelope {
+  ok: boolean
+  data?: DataSourceDetail
 }
 
 interface TestEnvelope {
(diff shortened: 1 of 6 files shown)


zensgit:
## Review notes (from the data-sources connector backend side)

Two findings on the edit flow — **P1 I've fixed on the backend** (non-colliding), **P2 is yours** (frontend):

**[P1 — handled in #2155]** The edit payload re-sends `connection` with only the visible `{host,port,database}`, and the backend PUT was spreading `connection` **wholesale** (deep-merging only `options`/`poolConfig`). So editing a SQL Server source silently dropped `encrypt`/`trustServerCertificate`/`tlsMinVersion`/`tlsCiphers`/`legacyTls` → weakened cert validation / broken legacy TLS. **#2155** adds `connection: { ...oldConfig.connection, ...parse.data.connection }` (defense-in-depth) + a regression test, so this PR is now **safe regardless** of what the form sends. No change needed here for P1 — but seeding the update form from the loaded `detail.connection` is a nice belt (so the user *sees* the real connection).

**[P2 — please fix here]** `types.ts` supports `connection.server`, but the edit/create form reads only `form.host` and (UI-1) marks **host `required`**. An API/runbook-created MSSQL source that used `server` (not `host`) loads with a blank Host and **can't be saved as-is**. Suggest: read `server` into the form, and don't `required`-block `host` when `server` is present (accept either). Same fix likely covers create + edit.

CI/diff otherwise clean. Full review note: `/tmp/pr2154-review-claude-20260530.md`.

PR #2161, opened 2026-05-30
fix(web): data-sources UI supports connection.server (P2) — server-only MSSQL sources are editable

zensgit:
## P2 from the #2154 review — `connection.server` support

**The bug (now on main after #2154):** the edit/create form read only `form.host` and (UI-1) marked host `required`. A SQL Server source created via API/runbook with **`connection.server`** (named instance, no `host`) loaded with a blank Host and **couldn't be saved**. `types.ts` already supported `server`; the form ignored it.

### Fix
- **`CreateFormState` + both builders** (`buildCreatePayload`/`buildUpdatePayload`) carry `server` and send it when set (omit-blank discipline kept).
- **`DataSourcesView`**: a **Server** field (shown for `sqlserver`), `fillFormFromDetail` reads `connection.server`, **host is no longer hard-required**, and submit validates **"host OR server"** for SQL — so a `server`-only source is savable while an empty SQL connection is still blocked.

### Tests
`form()` helper carries `server`; +2 cases — create/update can connect by `server` (no host). **20/20 green** · `vue-tsc` 0 · eslint clean. Lock-safe: frontend only.

Pairs with **#2155** (the backend `connection` deep-merge): together, editing a SQL Server source now neither drops its TLS keys (backend) nor blocks on a `server`-only connection (frontend).

### DIFF
--- a/apps/web/src/data-sources/buildPayload.ts
+++ b/apps/web/src/data-sources/buildPayload.ts
@@ -12,6 +12,7 @@ export interface CreateFormState {
   name: string
   type: DataSourceType
   host: string
+  server: string
   port: number | '' | undefined
   database: string
   username: string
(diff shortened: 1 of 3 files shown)


zensgit:
## Fixed the P2 follow-up (server scoped to sqlserver) — push 2a6b48210

Good catch — the first cut let a stale `server` bypass Postgres's host requirement (pick SQL Server → fill Server → switch back to Postgres → submit a `server`-only Postgres source). Now `server` is **sqlserver-only**, three ways:

- **builder**: `connection.server` is sent only when `type === 'sqlserver'` (both create + update).
- **submit guard**: Postgres still requires `host`; **SQL Server** accepts `host || server`. Message is type-aware (`请填写 Host` vs `Host 或 Server 之一`).
- **`watch(form.type)`**: clears `form.server` when switching away from sqlserver, so no stale value lingers.

+1 test (`Postgres does NOT send server even if form.server is set`). **21/21 · vue-tsc 0 · eslint clean.** Full review note: `/private/tmp/pr2161-review-claude-20260530.md`.

PR #2172, opened 2026-06-01
feat(data-sources): add read-only table preview

zensgit:
## Summary\n- add a SQL-only read-only preview action on the data-sources page\n- load schema/table choices via the existing /api/data-sources/:id/schema endpoint\n- preview rows through bounded /api/data-sources/:id/select with limit=100 and render the returned columns/rows\n- keep HTTP sources out of the table preview path and do not add raw query, export, writes, joins, or query-builder UI\n\n## Verification\n- pnpm --filter @metasheet/web exec vitest run tests/data-sources-ui.spec.ts tests/data-sources-api-preview.spec.ts --watch=false\n- pnpm --filter @metasheet/web exec vue-tsc --noEmit\n\n## Guardrails\n- API client test asserts the preview path calls /select and never /query\n- UI test asserts preview calls schema + bounded select, hides preview for HTTP sources, and renders returned rows\n

### DIFF
--- a/apps/web/src/data-sources/api.ts
+++ b/apps/web/src/data-sources/api.ts
@@ -4,6 +4,9 @@ import type {
   CreateDataSourcePayload,
   DataSourceDetail,
   DataSourceListItem,
+  DataSourceSchemaInfo,
+  DataSourceSelectPayload,
+  DataSourceSelectResult,
   DataSourceTestResult,
   RotateDataSourceCredentialsPayload,
   UpdateDataSourcePayload,
@@ -22,6 +25,16 @@ interface DetailEnvelope {
 interface TestEnvelope {
   ok: boolean
   data?: DataSourceTestResult
+}
+
+interface SchemaEnvelope {
+  ok: boolean
+  data?: DataSourceSchemaInfo
+}
+
+interface SelectEnvelope {
+  ok: boolean
+  data?: DataSourceSelectResult
 }
 
 interface ErrorEnvelope {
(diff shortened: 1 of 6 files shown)


zensgit:
## Independent review — APPROVE (merge first of the pair) ✅

UI-preview slice. Reviewed the full diff + backend handlers.

**Strengths**
- **Honors the A5 boundary** — calls `/select` with `limit:100`; a test explicitly asserts it **never** calls raw `/query`. Right discipline (structured bounded read, not raw SQL).
- **Read-only / lock-safe** — only `GET /schema` + `POST /select`, both `rbacGuard('data_sources','read')` + `assertAccess`; no write/credential/RBAC surface.
- `isSqlSource` fail-safe (unknown type hides preview); `formatCell` safe for null/Date/object; `previewColumns` prefers metadata, falls back to row-key union.
- Tests cover API-client + mounted-view render (asserts `limit=100`).

**Notes (non-blocking)**
- F1 — first UI to render actual customer **row data** (by design for a preview, properly gated). Cells render verbatim; worth a one-line doc note that preview shows raw source values. No code change.
- F2 — passes qualified `schema.table` to `/select`; backend tolerates via A2 per-segment quoting. Cosmetic asymmetry vs #2173 (which splits name/schema).

**Merge**: `MERGEABLE`; BLOCKED only by pending `test (18.x)/(20.x)`. **Merge this one FIRST** — #2173 is stacked/diverged on top of it. Rebase to 0-behind, wait green, squash.

PR #2176, opened 2026-06-01
docs(multitable): design cross-sheet related echo mask

zensgit:
## Summary
- design-lock the post-F3 crossSheetRelated write-echo finding
- require per-related-sheet layer-2 ∧ layer-3 allowed-field masking; do not reuse the edited-sheet echo set
- lock mask-not-drop for field-denied related records, drop unreadable related sheets, and keep realtime/F4/F5 out of scope

## Verification
- docs-only
- git diff --check

### DIFF
--- a/docs/development/multitable-cross-sheet-related-echo-mask-design-20260601.md
+++ b/docs/development/multitable-cross-sheet-related-echo-mask-design-20260601.md
@@ -1,0 +1,126 @@
+# Multitable cross-sheet related echo mask design
+
+**Date:** 2026-06-01
+**Slice:** post-F3 new finding #1 — `crossSheetRelated` write-path echo mask.
+**Status:** DESIGN-LOCK ONLY. Implementation is a separate explicit opt-in.
+**Grounding:** `origin/main` @ `f99841073`.
+
+---
+
+## 0. Why this exists
+
+F3 (#2169) closed the same-sheet write echo: `PATCH /records/:recordId` and `POST /patch` now compute a layer-2 ∧ layer-3 readable echo set for the edited sheet and pass it into `RecordWriteService` (`univer-meta.ts:7373-7391`, `:8327-8338`). That echo set is explicitly **read-back only** (`record-write-service.ts:212-217`), not the write gate.
+
+During the F3 review we found a different echo channel:
+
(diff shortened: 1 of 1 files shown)


zensgit:
## Independent review — APPROVE (design-lock) ✅

Verified every cited code seam against `origin/main`:
- **Leak is real (headline)** — `record-write-service.ts:860` `crossSheetRelated = relatedRecords.filter(record.sheetId !== sheetId)` returned unmasked at `:989`, while same-sheet at `:858`/`:886` **is** masked via `filterRecordDataByFieldIds`. Exact asymmetry described.
- **Safety floor correct** — `computeDependentLookupRollupRecords` (`routes/univer-meta.ts:1817`) already calls `resolveReadableSheetIds` (sheet-level drop), so this is **not** all-record/no-auth; the gap is precisely **field-level within readable related sheets**. (Tightens my F3 report's looser "unmasked" framing — there IS a sheet gate.)
- **F3 read-back semantics** — `record-write-service.ts:213` literally "Echo / read-back field set ONLY (NOT the writable set)" ✓
- **Reuse helpers exist** — `resolveSheetReadableCapabilities` (`:1174`), `loadFieldPermissionScopeMap` (`:774`).

**Design strengths:** D3 per-related-sheet allow set explicitly forbids reusing the edited sheet's `readableEchoFieldIds` (sheet-scoped field ids — the cross-context trap); D5 fail-closed on missing subject (kills the empty-map fail-open pattern); R2+R4 are the tests that earn it (visible-B-field survives / independent per-sheet masks). Non-goals correctly carve out realtime broadcast (`:938-966`), F4/F5, central RBAC/K3.

**Minor (impl-time, non-blocking):** (1) once D4 masks same-sheet dependents at the producer, `RecordWriteService`'s existing same-sheet mask (`:854`) becomes a redundant-but-idempotent second pass **iff** the producer's edited-sheet composite == F3's `readableEchoFieldIds` — confirm they're the same set, keep R6 green. (2) R1 cross-sheet lookup/rollup seed is heavier than F3's formula case — budget one discovery cycle; positive control required so "empty data" can't false-pass. (3) §0 grounds at `f99841073`; re-confirm line numbers at impl time (main has advanced).

**Verdict:** technically accurate, fail-closed, rigorous matrix, no scope creep, lock-safe. Approve as design-lock — **implementation remains a separate explicit opt-in** per its Status line. Full notes: a prior-session MD at `/tmp/pr2176-review-claude-20260601.md`.

PR #2179, opened 2026-06-01
fix(multitable): mask cross-sheet related lookup/rollup echo by field_permissions (#2176)

zensgit:
Implements the **#2176 design-lock** (`multitable-cross-sheet-related-echo-mask-design-20260601.md`). Follows **F3 (#2169)** which closed the same-sheet write echo.

## The leak (verified on `origin/main`)
`computeDependentLookupRollupRecords` recomputes lookup/rollup fields for records on **other** sheets that depend on the edited record. `record-write-service.ts:860` returned that cross-sheet subset **unmasked** (same-sheet records at `:858`/`:886` *are* masked). So a user who can **write sheet A** and **read sheet B** received B's recomputed computed-field value **even when B's `field_permissions.visible=false` denied that field**.

## The fix — at the producer seam (D4)
Upgrade the single `resolveReadableSheetIds()` call into a per-related-sheet `resolveSheetReadableCapabilities()` pass that supplies **both**:
- **D2** sheet-read drop (`!capabilities.canRead → continue`)
- **D3** the layer-2 ∧ layer-3 allowed-field mask **keyed to the RELATED sheet** (not the edited sheet's field ids) — mirrors the same-sheet `readableEchoFields` composite at the POST /patch route (`:8336`)

Plus **D1** mask-not-drop (denied record → `data:{}`, keeps the "recomputed" signal) and **D5** fail-closed (no subject → empty allow set, never derive against an empty scope map → never fail OPEN).

## Verification — proven RED→GREEN on a real Postgres
`tests/integration/multitable-cross-sheet-echo-mask.test.ts` (`describeIfDatabase` — runs under the `plugin-tests.yml` DB-gated step alongside the F3 test).

| | Result |
|---|---|
| **WITHOUT fix** (revert `univer-meta.ts`) | **R1 FAILS** — `expected 11 to be undefined`: the denied rollup value `11` leaks |
| **WITH fix** | **6/6 pass** |
| **F3 same-sheet test** | **5/5 pass** (no regression) |
| `tsc --noEmit` | 0 errors |

Tests: **R1** canary (denied cross-sheet computed omitted) + **R2** visible positive control (a readable computed field on B survives — guards against over-masking B by A's ids) + **R4** independent per-sheet masks (a second related sheet C keeps its own visible field) + per-subject (ungranted user still receives the value) + **R6** same-sheet regression (the producer now double-masks same-sheet dependents — idempotent since both composites are layer-2 ∧ layer-3 for the same sheet+subject) + **R-D5** (the write route 401s a no-subject request **before** the producer runs, so the D5 branch is unreachable-by-construction belt-and-suspenders).

## Scope / safety
- Lock-safe: multitable field-masking only. **No** central RBAC/auth, **no** `plugin-integration-core`, **no** K3 areas.
- Non-goals (per design): realtime broadcast (`record-write-service.ts:938-966`, separate per-recipient design), F4 create-echo, F5 link-options.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -160,7 +160,7 @@ jobs:
           : "${DATABASE_URL:?DATABASE_URL is required for after-sales integration}"
           pnpm --filter @metasheet/core-backend exec vitest --config vitest.integration.config.ts run tests/integration/after-sales-plugin.install.test.ts --reporter=dot
 
-      - name: Run multitable real-DB integration (permission golden + view-aggregate + formula dry-run + read-path field mask + dashboard authz + automation retry provenance)
+      - name: Run multitable real-DB integration (permission golden + view-aggregate + formula dry-run + read-path field mask + write echo masks + dashboard authz + automation retry provenance)
         if: matrix.node-version == '20.x'
         env:
           DATABASE_URL: postgresql://postgres@localhost:5432/metasheet_test
@@ -180,6 +180,7 @@ jobs:
             tests/integration/multitable-records-summary-field-mask.test.ts \
             tests/integration/multitable-dashboard-chart-authz.test.ts \
             tests/integration/multitable-write-echo-field-mask.test.ts \
+            tests/integration/multitable-cross-sheet-echo-mask.test.ts \
             tests/integration/multitable-automation-retry.test.ts \
             tests/integration/multitable-automation-jobs.test.ts \
             --reporter=dot
--- a/packages/core-backend/src/routes/univer-meta.ts
+++ b/packages/core-backend/src/routes/univer-meta.ts
@@ -8356,8 +8356,10 @@ export function univerMetaRouter(): Router {
       // F3 (#2106 §3 F3): RecordWriteService uses these ONLY for the read-back echo (it masks record / related /
       // formula data + summaries at record-write-service.ts:828/854/882/914/929), so narrow them to the
       // layer-2 ∧ layer-3 readable set — a field_permissions-denied value must not be echoed. fieldById (the
-      // write gate, below) stays from ALL fields, so a write-only-no-read field remains writable. (crossSheetRelated
-      // @record-write-service.ts:985 is returned UNMASKED — a separate finding, see the verification doc.)
+      // write gate, below) stays from ALL fields, so a write-only-no-read field remains writable. (The
+      // crossSheetRelated echo @record-write-service.ts:860/989 is masked SEPARATELY — by the related sheet's
+      // own field permissions at the producer seam, computeDependentLookupRollupRecords; see #2176. These
+      // edited-sheet echo ids are deliberately NOT used for it, as the field ids are scoped to this sheet.)
       const echoFieldScopeMap = access.userId ? await loadFieldPermissionScopeMap(pool.query.bind(pool), sheetId, access.userId) : new Map()
       const echoFieldPermissions = deriveFieldPermissions(visiblePropertyFields, capabilities, { hiddenFieldIds: [], fieldScopeMap: echoFieldScopeMap })
       const readableEchoFields = visiblePropertyFields.filter((field) => echoFieldPermissions[field.id]?.visible !== false)
--- a/packages/core-backend/src/routes/univer-meta.ts
+++ b/packages/core-backend/src/routes/univer-meta.ts
@@ -1869,13 +1869,37 @@ async function computeDependentLookupRollupRecords(
   }
 
   const results: RelatedComputedRecord[] = []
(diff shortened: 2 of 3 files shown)


zensgit:
**Follow-up (push bd633961c):** added the test to the `plugin-tests.yml` multitable real-DB step's explicit file list. That step runs a named list (not the whole `tests/integration` dir), so a `describeIfDatabase` test omitted from it would silently **skip** in CI (vacuous green). Verified locally under the exact CI invocation (`vitest --config vitest.integration.config.ts run tests/integration/multitable-cross-sheet-echo-mask.test.ts`) → 6/6 pass with `DATABASE_URL` set, and proven RED (R1 leaks value `11`) when the `univer-meta.ts` fix is reverted. So the CI DB step now actually exercises the mask, not a skip.

zensgit:
## Independent review (2 adversarial agents) + review-driven hardening

Since this is my own code, I had **two fresh subagents** (neither shown my reasoning) try to break it — a security-leak lens and a test-rigor lens — then re-verified their findings against the code and fixed the actionable ones in `004a15d13`.

### Security → **CLEAN**
Attacked 6 vectors (fail-open, wrong-subject, bypass seams, computed coverage, admin, D5) — all safe, verified against code:
- Mask keyed to the RELATED sheet (per-sheet `resolveSheetReadableCapabilities` inside the loop); no cross-contamination.
- `crossSheetRelated` referenced only at `record-write-service.ts:860`/`:989`; the realtime broadcast + EventBus carry **same-sheet-only** data → no unmasked cross-sheet emitter.
- `deriveFieldPermissions` has no admin bypass → exact parity with the same-sheet path.
- **One out-of-scope backlog item:** `applyLookupRollup` reads the foreign lookup **target** field gated by sheet-read only, never field-perm (`univer-meta.ts:1707/1728/1737`). **Pre-existing, on every read path, NOT introduced/worsened here.** Worth a separate slice; doesn't block this PR.

### Tests → was GAPS-BUT-OK, now strengthened
The valuable catch: **R6's comment was wrong.** Editing `REC_B` finds no producer dependents (no `meta_links` row has `foreign_record_id=REC_B`), so the producer returns `[]` — R6 actually exercises the **RWS same-sheet path (F3's), not the producer's** `allowedFieldIds`. It also had **no positive control**. Fixed in `004a15d13`:
- R1/R2: pin `B_VISIBLE` to the computed value (proves the rollup fired — running it caught real test-state coupling across tests).
- R6: added the positive control (pinned to the **live** A_SRC) + corrected the comment.
- R3/R5/D1 left uncovered by **explicit reasoned acceptance** (the R5 gate-swap `resolveReadableSheetIds`→per-sheet `resolveSheetReadableCapabilities` is **read-equivalent** — `applyContextSheetReadGrant` reproduces `canReadWithSheetGrant`'s exact OR-clause, verified by the agent).

**Re-verified RED→GREEN on a real Postgres** after the hardening: without the fix R1 fails (`expected 11 to be undefined`); with it 6/6 pass.

Full synthesis: `/tmp/pr2179-synthesis-review-claude-20260601.md`. Net: security clean, tests now rigorous.

zensgit:
## P3 stale-comment fixed (`486b65059`)

Good catch — updated the F3 comment at the POST /patch echo composite. It still flagged \`crossSheetRelated ... returned UNMASKED — a separate finding\`, which is exactly what this PR closes. The note now reads that the cross-sheet echo is masked **separately, at the producer seam (\`computeDependentLookupRollupRecords\`) by the related sheet's own field permissions — see #2176**, so a future reader won't chase an already-fixed leak. Also corrected the stale \`:985\` line ref to \`:860/989\`.

Comment-only (verified: 0 code lines added/removed; tsc 0 errors). No runtime change.

zensgit:
## Superseded by #2178 — closing, not merging

While this PR was in review, the parallel session shipped the **same fix** as **#2178** (squash \`5377c1830\`, **MERGED to origin/main**). It closes the identical cross-sheet field-read leak at the same producer seam (\`computeDependentLookupRollupRecords\`): per-related-sheet \`resolveSheetReadableCapabilities\` + fail-closed \`!access.userId || !capabilities.canRead → continue\` + \`filterRecordDataByFieldIds(extractLookupRollupData(...))\` mask. **The leak is already fixed on main**, so this PR is now a duplicate fix and is CONFLICTING/DIRTY against it.

Per maintainer direction: **closing without merge.** The fix half is redundant.

The test depth here is stronger than #2178's (6 cases vs 3 — value-pinned R1/R2, per-subject, R-D5 no-subject fail-closed precondition, and a corrected R6 positive control). That coverage is being ported to the merged #2178 implementation as a **test-only follow-up PR** (`test(multitable): strengthen cross-sheet related echo mask coverage`), so nothing of value is lost.

Closing in favor of #2178 + the follow-up.

PR #2181, opened 2026-06-01
docs(multitable): design realtime broadcast invalidation

zensgit:
## Summary

Design-lock for the remaining realtime broadcast finding after crossSheetRelated (#2178).

It locks two hard requirements for the later implementation slice:

- `join-sheet` must become a trusted sheet-read gate: Socket.IO clients send `auth.token`, the server verifies it, and room membership/presence uses token-derived identity rather than `?userId=`.
- `sheet:op` becomes value-free invalidation: keep `kind` / `recordId(s)` / `fieldIds`, but no `recordPatches[].patch` reaches the broadcast payload. Clients refetch through the already-gated HTTP read paths.

## Notes

During grounding, the issue proved broader than value masking alone: current `join-sheet` accepts arbitrary sheet IDs and trusts a client-supplied `userId` query for presence. The design therefore treats subscription authorization and no-value payloads as one implementation slice; either half alone is incomplete.

## Verification

Docs-only. Self-checks:

- Anchors verified against `origin/main` @ `5377c1830`.
- `git diff --check` clean.
- Scope is a single new docs file; no runtime code changed.

## Non-goals

- No implementation in this PR.
- No Yjs document-content changes.
- No comment-room auth changes.
- No F4/F5 HTTP echo fixes.

### DIFF
--- a/docs/development/multitable-realtime-broadcast-invalidation-design-20260601.md
+++ b/docs/development/multitable-realtime-broadcast-invalidation-design-20260601.md
@@ -1,0 +1,152 @@
+# Multitable realtime broadcast invalidation design
+
+**Date:** 2026-06-01
+**Slice:** post-crossSheetRelated new finding #2 - realtime `sheet:op` value broadcast.
+**Status:** DESIGN-LOCK ONLY. Implementation is a separate explicit opt-in.
+**Grounding:** `origin/main` @ `5377c1830` (after #2178).
+
+---
+
+## 0. Why this exists
+
+The record read/write echo arc now gates persisted HTTP responses by the layer-2 and layer-3 field-read composite. The realtime channel is still outside that floor:
+
+- `publishMultitableSheetRealtime` publishes the payload to `spreadsheet.cell.updated` unchanged after cache invalidation (`realtime-publish.ts:40-44`).
+- `CollabService` forwards that payload to every socket in `sheet:${spreadsheetId}` as `sheet:op` (`CollabService.ts:25-33`).
+- Publishers put raw values in `recordPatches[].patch`: `RecordWriteService.patchRecords` includes user changes and formula values (`record-write-service.ts:941-966`); direct create/update paths include raw `patch` (`record-service.ts:601-614`, `:1009-1021`); form update and attachment delete paths do the same (`univer-meta.ts:7248-7260`, `:8143-8155`).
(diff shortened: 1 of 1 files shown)


zensgit:
## Independent review — APPROVE (design-lock) ✅

Verified both premises against `origin/main` — both holes are real:
- **D0 (room-membership) — REAL, arguably the higher-severity half.** `CollabService.ts:189-194`: `join-sheet` calls `socket.join(buildSheetRoom(sheetId))` **unconditionally** — no token verify, no `canRead`. The socket *can* verify a token (`:45-60`) but join-sheet doesn't use it; `userId` comes from the **spoofable** `handshake.query.userId` and drives presence. So any socket joins any sheet room, gets value-bearing `sheet:op`, and can spoof presence.
- **D1 (value channel) — REAL.** `realtime-publish.ts`: `publishMultitableSheetRealtime` does `_invalidateCache(...)` then `eventBus.publish('spreadsheet.cell.updated', payload)` unchanged — `recordPatches[].patch` carries user + formula values an in-room subscriber shouldn't see.
- **Frontend** (`useMultitableSheetRealtime.ts:249-252`): connects with `query:{userId}`, **no `auth.token`** — confirms membership rests on an unauthenticated query param.

**The strong design call:** it ties D0+D1 together — D3 keeps `fieldIds` as metadata **only because** D0 gates the room (strip values but leave arbitrary sockets in → still leaks record/field *activity*). The value-free-over-per-recipient-masking choice is the right safety call (per-recipient deferred honestly as a future perf optimization). D0 is feasible with existing pieces — `resolveSheetCapabilitiesForUser` + the Yjs adapter's `setAuthChecker` injection precedent (`index.ts:2021`) — no central RBAC touch. R-matrix is security-grade (R2 authz-not-just-authn, R3 canary-absent-from-raw-payload, R7 cache-still-fires, R8 trusted-presence) and correctly mandates a real Socket.IO server.

**Non-blocking suggestions:** (1) **lead with D0** in the impl narrative — the unauth room join + spoofable presence is at least as serious as the value leak and easy to under-read under "invalidation". (2) Consider splitting D0 (room auth, larger — socket auth + 3 frontend composables) from D1 (value strip, small) into two slices, OR keep them one PR so the channel can't be half-closed (R5 must assert no value-apply path survives). (3) Re-confirm line refs at impl (join-sheet is `:189-194`, drifted from the doc's `:169-176`).

APPROVE as design-lock; implementation a separate opt-in. Full notes: `/tmp/pr2181-review-claude-20260601.md`.

PR #2173, opened 2026-06-01
feat(data-sources): add schema browser UI

zensgit:
## Summary
- Stacked on #2172 to avoid colliding with the open read-only preview branch that touches the same UI files.
- Adds a read-only schema browser action for SQL sources: load `/schema`, choose a table/view, then load `/tables/:table?schema=...` for field details.
- Displays columns with type, nullable/PK constraint, and default value; this does not call `/select` or read business rows.
- Updates the data-sources tracker so UI-schema + UI-preview are marked complete once the stack lands.

## Verification
- `pnpm --filter @metasheet/web exec vitest run tests/data-sources-ui.spec.ts tests/data-sources-api-preview.spec.ts --watch=false` (33/33)
- `pnpm --filter @metasheet/web type-check`
- `pnpm --filter @metasheet/web exec eslint src/data-sources/api.ts src/data-sources/types.ts src/stores/dataSources.ts src/views/DataSourcesView.vue tests/data-sources-ui.spec.ts tests/data-sources-api-preview.spec.ts --max-warnings=0`
- `git diff --check`

## Notes
- This PR intentionally targets #2172's branch as its base. After #2172 merges, rebase/retarget this PR onto `main` before final squash.
- Browser/manual visual smoke not run in this turn; coverage is via jsdom UI specs + type/lint gates.

PR #2182, opened 2026-06-01
fix(data-sources): guard schema browser against stale schema responses

zensgit:
Follow-up to the #2173 post-merge review: hardens this view against async schema-browser/preview stale-response races. Read-only boundary unchanged (still `schema` / `tables/:table`, no `/query`, `data_sources:read` + `assertAccess` intact, no new credential read path) — these are UI-state-consistency fixes.

## 1. `openSchemaBrowser` (the originally-flagged one)
`openSchemaBrowser(id)` set `activeSchemaId = id`, then `await store.loadSchema(id)` without re-checking afterward. On a fast A→B switch, A's slow response arrives last, clobbers the global `schemaTable` to A's first table (not a valid option in the displayed B panel → the `<select>` collapses to empty) and fires `loadTableInfo(A, …)` while B is showing.

**Fix:** capture `requestedId`, clear `schemaTable` on open, and `if (activeSchemaId.value !== requestedId) return` after the await. Single guard suffices — the later `loadTableInfo` await writes only the keyed `tableDetails[id:table:schema]` cache and clears the *old* id's `schemaErrors`, neither read by the new panel.

## 2. `openPreview` (added per review — same class, stronger)
Same race, but worse: A's late continuation sets the global `previewTable` to A's table, then `runActivePreview()` reads `activePreviewId` (= the *new* source B) → `previewRows(B, { table: A's table })`, a **cross-source wrong query**. Fixed with the identical pattern (capture `requestedId`, clear `previewTable` on open, recheck `activePreviewId` after the await, then run preview).

## Tests (out-of-order, each proven red without its guard)
`data-sources-ui.spec.ts` adds two parallel out-of-order tests (park A's schema → switch to B fast → resolve A last):
- schema browser: B stays selected, `loadTableInfo(A,…)` never called. Red without guard → `select.value` collapses to `''`.
- preview: `previewRows` called exactly once for B, never the cross-source query. Red without guard → **2 calls** (the legit B preview + the stale `previewRows('b', {table: A})`).

Full spec **33/33**; `eslint` + `vue-tsc -b` clean.

### DIFF
--- a/apps/web/src/views/DataSourcesView.vue
+++ b/apps/web/src/views/DataSourcesView.vue
@@ -633,9 +633,16 @@ function closeSchemaBrowser(): void {
 }
 
 async function openPreview(id: string): Promise<void> {
+  const requestedId = id
   activePreviewId.value = id
+  // Clear the prior panel's selection up front so it cannot linger while the new schema loads.
+  previewTable.value = ''
   store.clearPreviewError(id)
   const schema = store.schemas[id] ?? await store.loadSchema(id)
+  // A newer preview may have opened while loadSchema was in flight (fast A→B switch). Drop this
+  // stale response so it cannot set previewTable to the wrong source's table and then run a
+  // cross-source query — runActivePreview() reads the *current* activePreviewId.
+  if (activePreviewId.value !== requestedId) return
   const firstChoice = [
     ...(schema?.tables ?? []),
     ...(schema?.views ?? []),
--- a/apps/web/src/views/DataSourcesView.vue
+++ b/apps/web/src/views/DataSourcesView.vue
@@ -605,8 +605,14 @@ async function testConnection(id: string): Promise<void> {
 }
 
 async function openSchemaBrowser(id: string): Promise<void> {
+  const requestedId = id
   activeSchemaId.value = id
+  // Clear the prior panel's selection up front so it cannot linger while the new schema loads.
+  schemaTable.value = ''
   const schema = store.schemas[id] ?? await store.loadSchema(id)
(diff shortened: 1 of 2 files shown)


zensgit:
## Independent review — APPROVE ✅

A real out-of-order race fix, not cosmetic. Verified against `origin/main`:

**The bug is real and worse than a flicker.** `runActivePreview()` reads `const id = activePreviewId.value` (`DataSourcesView.vue:644`) — the **current** panel. So on a fast A→B switch where A's slow schema response arrives last, pre-fix it sets `previewTable` to A's table and fires `previewRows('b', { table: A's-table })` — a **cross-source wrong query** (B's connection, A's table name).

**The fix is the canonical guard:** capture `const requestedId = id` before the await, then `if (activePreviewId.value !== requestedId) return` after `loadSchema` resolves; plus clear `schemaTable`/`previewTable` up-front so the prior selection can't linger. Minimal (13 lines), correctly placed after the await.

**Tests are the strong part** — both use a **parked promise** (`resolveA` held) to *deterministically force* the out-of-order race (A parked → B fast → A resolved last), then assert the no-wrong-query invariant: `previewRowsMock` called exactly once with B's table and **never** with `{ table: 'public.a_items' }`. That's how to test a race — force the ordering, don't hope for it.

**Nits (non-blocking):** the guard covers the schema-load race; a stale *table-info* (second await) is effectively dropped already since `loadActiveTableInfo` re-reads the current `selectedSchemaChoice` — worth a one-line confidence check but not a hole. Same-id refresh correctly passes the guard (not a stale case).

APPROVE — lock-safe (frontend data-sources only), ready on green. Full notes: `/tmp/pr2182-review-claude-20260601.md`.

Issue #2205, opened 2026-06-02
[C2 entity smoke] Validate read-only SQL data source bridge on entity machine


zensgit:
## Purpose

Validate the deployed C2 read-only SQL data-source bridge on the entity machine after installing release `multitable-onprem-datasource-c2-20260602-9f558d522`.

Release: https://github.com/zensgit/metasheet2/releases/tag/multitable-onprem-datasource-c2-20260602-9f558d522
Bound source commit: `9f558d522`

This test is for the read-only database-to-Data-Factory path only. It does **not** authorize K3 Save-only, Submit, Audit, BOM, multi-record writes, production writes, or external-write behavior.

## Scope

Validate this chain on the entity machine:

`/data-sources` read-only SQL connection -> Data Factory `data-source:sql-readonly` source connection -> load objects/schema as request user -> dry-run reads rows -> zero writes.

Expected security posture:

- Data Factory connection stores only `{ dataSourceId, object }`.
- Credentials remain only in `/data-sources`; they are not copied into integration connection config.
- Reads are owner-scoped and fail-closed.
- A writable database source must not be accepted by the read-only bridge.
- No K3 operation is executed.

## Test Steps

1. Confirm deployed package

   - Entity machine is upgraded with release `multitable-onprem-datasource-c2-20260602-9f558d522`.
   - Backend and web are restarted.
   - Migrations/healthcheck pass.

2. Create or select a read-only SQL data source

   In `/data-sources`:

   - Create a PostgreSQL or SQL Server connection using read-only credentials.
   - Confirm the connection test passes.
   - Use a small allowlisted/readable table or view, for example `public.items`.

3. Create a Data Factory source connection

   In Data Factory Workbench connection manager:

   - Enable advanced connectors if needed.
   - Select kind `data-source:sql-readonly`.
   - Use the structured picker to select the data source by name.
   - Fill object/table name, for example `public.items`.
   - Save the connection.

4. Load source metadata

   - Select the new bridge connection as the Data Factory source.
   - Load objects.
   - Load schema.

   Expected:

   - Object list includes the target table/view.
   - Schema loads without missing-principal failure.
   - If the logged-in user is not the data-source owner, the read fails closed without leaking credentials.

5. Run dry-run

   - Configure a simple target/staging path as needed for dry-run.
   - Run dry-run only.

   Expected:

   - `rowsRead > 0` when the table has rows.
   - `rowsWritten = 0`.
   - Preview/cleansed output shows data rows as expected.
[...]

PR #2228, opened 2026-06-03
fix(integration): Bridge Agent recovery legibility — unreachable message + stale-error retest hint (#2223)

zensgit:
## Summary

Operator-recovery slice for **#2223**: when the readonly **Bridge Agent stops**, the source system goes to `error`; after the agent restarts and objects/schema read fine, the pipeline dry-run stays **blocked by the stale `error`** and the failure reads as a generic fetch error — so an operator can't tell the source just needs a retest.

**Read-only legibility only.** No K3 Save/Submit/Audit/BOM, no C3 incremental, no external write, no RBAC/auth touch.

## What changed

The error→active transition **already existed** (`resolveTestedStatus` clears `error`→`active` on a successful test and won't auto-activate `inactive`) — this slice makes the recovery **legible** and **test-locks** it. Two messages + one UI affordance:

1. **Bridge Agent unreachable → clear message (not generic fetch failure).** `requestJson` now detects a down localhost agent and throws `BRIDGE_AGENT_UNREACHABLE`: *"Bridge Agent is not reachable at … — the readonly Bridge Agent service may not be running. Start its scheduled task / process, then retest the source connection."* Detection keys on undici's **verified** `TypeError: fetch failed` (I probed a dead localhost port — `cause.code`/`cause.errors[]` are **not** populated on every Node version, so they're secondary enrichment, not the gate). Covers **both** `testConnection` and the objects/schema read path (one site).
2. **Dry-run stale error → retest hint.** `assertActiveSystem` appends *"the source connection last failed its test; retest/reactivate it (Workbench 测试连接) then re-run"* when a source sits in `error` — and a distinct *"activate it"* hint for `inactive` (a connection test deliberately won't reactivate inactive). The `{field, systemId, status}` details (#2205 references `systemId`) are preserved.
3. **Workbench affordance.** `testSystem` reports recovery explicitly (`error`→`active` ⇒ "数据源连接已恢复，已重新激活"); the `error` status label names the retest action so the readiness panel hint is actionable.

## Maps to #2223 acceptance

- ✅ Bridge Agent down → source test fails → `status=error`, **message clear** (`BRIDGE_AGENT_UNREACHABLE`)
- ✅ Bridge Agent restored → retest succeeds → `status=active` (existing `resolveTestedStatus`, now locked)
- ✅ objects/schema pass (unchanged read path; unreachable message also clear there)
- ✅ dry-run no longer reads as broken when blocked by stale error → **retest hint** points at the fix
- ✅ inactive source retest success does **NOT** auto-activate (existing guard, now locked symmetrically)

## Tests

- bridge-agent: unreachable via the verified `fetch failed` shape + `cause.code` / dual-stack `cause.errors[]` enrichment shapes + a **negative control** (reachable-agent HTTP 500 is NOT mislabeled unreachable).
- pipeline-runner: `assertActiveSystem` 3-branch hint (error→retest / inactive→activate / other→status) + details preserved.
- http-routes: explicit **error→active** recovery lock (symmetric to the existing inactive-preservation test).
- web spec: actionable error-label lock.
- **Full plugin-integration-core suite (33 files) + web spec (17) green; web type-check clean.**

The recovery-message string branch (`已恢复…`) is a trivial `priorStatus==='error'` ternary backed by the backend error→active lock; the more complex/visible readiness-hint label is the one frontend-locked.

**Leave-for-review — not auto-merging.**

### DIFF
--- a/apps/web/src/views/IntegrationWorkbenchView.vue
+++ b/apps/web/src/views/IntegrationWorkbenchView.vue
@@ -2148,7 +2148,7 @@ async function refreshBootstrap(): Promise<void> {
 function connectionStatusLabel(system: WorkbenchExternalSystem | null): string {
   if (!system) return '未选择'
   if (system.status === 'active') return system.lastTestedAt ? '已连接' : '可用'
-  if (system.status === 'error') return system.lastError ? `异常：${system.lastError}` : '异常'
+  if (system.status === 'error') return system.lastError ? `异常：${system.lastError}（点击"测试连接"重新激活）` : '异常（点击"测试连接"重新激活）'
   return '未启用'
 }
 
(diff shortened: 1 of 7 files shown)


zensgit:
Reviewed. No blockers from my side. The slice stays within recovery legibility: unreachable Bridge Agent now gets a specific operator action, stale source status gets the retest/reactivate hint, and inactive still does not auto-activate on test. The reachable-agent HTTP 500 negative control is the important guard here. CI is green; PR is BEHIND, so merge should use the usual rebase -> fresh CI -> squash chain.

PR #2287, opened 2026-06-04
feat(data-adapters): PLM integration capability handshake client (C1)

zensgit:
## C1 — PLM integration capability handshake client (PLM-COLLAB Phase 2.5 consumer)

Starts closing the cross-repo loop for the PLM integration handshake on the MetaSheet
(consumer) side: `PLMAdapter.getIntegrationCapabilities()` queries the provider's advisory
manifest `GET /api/v1/integrations/capabilities` so the UI can degrade gracefully by
`supported` / `entitled` / `api_version` / `action_status`.

### Behavior
- **Advisory only (F3):** the manifest is a UI/degradation hint, NEVER an authorization
  decision — the PLM still enforces `is_entitled` at every feature endpoint.
- **Graceful degradation (F1):** legacy/non-yuantus mode → `unsupported-mode`; a PLM
  without the endpoint / 404 / network error → `unavailable`. The safe default for both is
  to HIDE automation entries (never assume support).
- **Caching (F2):** successful results are cached on the adapter instance; failures are NOT
  cached (retry next call). `refreshIntegrationCapabilities()` clears + re-fetches and is
  warmed fire-and-forget after a real connect.

### Scope
Adapter consumption + safe degradation + cache semantics only. No route (C2 follow-up), no
frontend UI (C3 follow-up), no pact change.

### Pact: intentionally no required interaction
C1 intentionally does not add a required pact interaction for
`/api/v1/integrations/capabilities`. The capability manifest is advisory/optional; old PLM
instances may not expose it. Consumer behavior is guarded by unit tests: success caches,
failure does not cache, unavailable/unsupported safely hides upgrade entries. Pact coverage
can be revisited only when we decide all supported PLM versions must expose this endpoint.

### Verification
- `pnpm --filter @metasheet/core-backend exec tsc --noEmit`: 0 errors.
- New `tests/unit/plm-adapter-capabilities.test.ts` (6 tests): typed manifest in yuantus mode;
  404/error → `unavailable`; non-yuantus → `unsupported-mode` (no fetch attempted); success
  caching; refresh re-fetch; no-cache-on-failure.
- Full backend unit suite: **2733 passed** (the fire-and-forget connect warm-up breaks nothing).
- ESLint clean on the changed source.

### DIFF
--- a/packages/core-backend/src/data-adapters/PLMAdapter.ts
+++ b/packages/core-backend/src/data-adapters/PLMAdapter.ts
@@ -632,6 +632,39 @@ const YUANTUS_SUPPORTED_OPERATIONS = [
   'cad_review_update',
 ]
 
+/**
+ * PLM-COLLAB Phase 2.5 (Integration Handshake) consumer types: the advisory capability
+ * manifest returned by the provider's `GET /api/v1/integrations/capabilities`. ADVISORY
+ * ONLY -- used to decide UI degradation / entry display; NEVER an authorization decision
+ * (the PLM provider still enforces is_entitled at every feature endpoint).
+ */
+export interface IntegrationFeatureCapability {
+  supported: boolean;
+  api_version: string | null;
+  entitled: boolean;
+  cache_scope?: { supported?: string; entitled?: string };
+  scenarios?: string[];
+  actions?: string[];
+  action_status?: string;
+}
+
+export interface IntegrationCapabilityManifest {
+  schema_version: string;
+  provider: string;
+  advisory: boolean;
+  features: Record<string, IntegrationFeatureCapability>;
+}
+
+/**
+ * Result of querying the manifest. The handshake degrades gracefully: a legacy/non-yuantus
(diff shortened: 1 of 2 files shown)


zensgit:
Review verdict: APPROVE — no blocking findings.

I reviewed the PR branch implementation against HTTPAdapter.query semantics and the existing PLMAdapter lifecycle. The graceful-degradation contract holds: legacy/non-yuantus never fetches and returns unsupported-mode; missing endpoint / HTTP error / malformed top-level response returns unavailable and is not cached; a valid yuantus advisory manifest is cached; refresh clears and refetches. connect() warm-up is fire-and-forget and cannot fail connect.

The advisory-only boundary is preserved: this adds no route, no frontend gating, no pact requirement, and no authorization decision. CI is green.

Residual/non-blocking: Gemini's active-promise caching suggestion is reasonable to avoid duplicate concurrent manifest calls before the first successful cache fill, but it is not a safety blocker. Future C2/C3 consumers should continue treating feature entries as advisory and safe-hide on unsupported/unentitled.

PR #2326, opened 2026-06-05
feat(data-factory): sync stock-preparation options

zensgit:
## Summary
- add C6 stock-preparation option sync: admin-only route updates select/dropdown field metadata from configured option sets
- add predefined option-action binding metadata with a backend allowlist; no SQL/JS/URL/function/payload bodies are accepted
- expose a minimal admin workbench panel for option sync and update C6 docs/verification

## Safety boundaries
- metadata-only: no PLM write, no K3 write, no MetaSheet business-row write
- plugin field-property patch goes through project namespace + object-scope checks
- evidence is values-free; tests cover duplicate/unknown option sources, redacted/secret-shaped values, unknown actions, executable keys, and patch failures

## Verification
- `pnpm --filter @metasheet/core-backend exec vitest run tests/unit/multitable-plugin-scope.test.ts --reporter=dot`
- `pnpm --filter @metasheet/web type-check`
- `pnpm --filter @metasheet/web exec vitest run tests/IntegrationWorkbenchView.spec.ts --watch=false`
- `git diff --check`


zensgit:
Addressed the two review hardening points in the PR head:

- `mergeJsonObject` now rejects prototype-pollution keys (`__proto__`, `constructor`, `prototype`) recursively; `multitable-provisioning.test.ts` locks the nested `__proto__` case and confirms Object prototype is not polluted.
- option `order` is now strict numeric input only; numeric strings like `"1"` fail closed and are covered in `stock-preparation-option-sync.test.cjs`.

Fresh CI is green: all required checks pass, Strict E2E is skipped as expected.

PR #2333, opened 2026-06-06
feat(attendance): S2-1 apply in/out merge punch policy

zensgit:
## Summary
- Exposes `punchPolicy.merge.{internalWinsOnIn, externalWinsOnOut}` through attendance settings.
- Applies the S2 in/out merge policy after real check events land on the two trusted write paths: normal `/punch` and approved `outdoor_punch`.
- Preserves record-only facts by locking the pre-write record and treating non-event first/last values as protected candidates; S2 recompute only intervenes when the corresponding key is enabled and internal/outdoor events coexist for that side.
- Keeps raw `attendance_events` immutable and continues to use `source=outdoor_approval` as the now-reserved classifier from S2-0.

## Verification
- `node --check plugins/plugin-attendance/index.cjs`
- fresh local Postgres DB migrated from scratch, then `pnpm --filter @metasheet/core-backend test:integration:attendance` → 4 files / 102 tests passed

## Scope / deferred
- Backend only. No frontend admin card in this PR.
- No staging smoke and no tracker ✅ flip here; S2-2 frontend card and S2-3 reverse/staging closeout remain separate opt-ins.

### DIFF
--- a/packages/core-backend/tests/integration/attendance-outdoor-punch.test.ts
+++ b/packages/core-backend/tests/integration/attendance-outdoor-punch.test.ts
@@ -3,6 +3,7 @@ import type { MetaSheetServer } from '../../src/index'
 import * as path from 'path'
 import net from 'net'
 import http from 'http'
+import { randomUUID } from 'crypto'
 import { Pool } from 'pg'
 
 // ② 打卡策略组 S3 外勤审批 — route-level real-DB integration (design-lock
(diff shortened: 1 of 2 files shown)


zensgit:
**REQUEST CHANGES**

The current S2-1 implementation still clobbers record-only import/override values on a later ordinary punch.

**Mechanism:** the route loads `protectedRecord`, then runs `upsertAttendanceRecord(... mode:'append' ...)` **before** `applyAttendanceInOutMergePolicy`. Append can overwrite a record-only `first_in_at`/`last_out_at`. The merge restore only runs when the same side has **both** internal and outdoor check events, so the required §5.11/§5.12 path (record-only import/correction → later ordinary punch, **no outdoor event**) is not protected.

The added protected-record test **false-passes** because it hand-inserts an `outdoor_approval` event first, creating the mixed bucket that triggers the restore branch.

**Required before merge:**
1. Add real-DB fail-first coverage for import/bulk record-only value + later ordinary punch.
2. Add real-DB fail-first coverage for missed/time-correction override + later punch.
3. Fix the append-before-merge ordering so protected record-only facts cannot be destroyed before the policy logic runs.

Issue #2343, opened 2026-06-06
[Data Factory] PLM stock-preparation duplicate-expanded-key handling


zensgit:
## Goal

Define product/runtime handling for `duplicate_expanded_key` rows discovered during PLM project BOM -> stock-preparation runs. This follows #2340, where a new-sample Apply-only gate successfully wrote clean rows while holding duplicate rows for manual confirmation.

This issue does not authorize another Apply by itself. It defines how held duplicate rows should be reviewed, resolved, and eventually applied or skipped.

## Trigger from #2340

Values-free onsite result:

- Fresh dry-run before apply: `manual_confirm_required`.
- `rowsExpanded=246`.
- `add=190`.
- `manual_confirm=28`.
- Owner explicitly approved exactly one Apply-only run with `acceptManualConfirmHold=true`.
- Apply result: `created=190`, `held=28`, `failed=0`.
- Post-Apply dry-run: `existingRows=190`, `add=0`, `skip=190`, `manual_confirm=28`.

Interpretation: clean non-duplicate rows wrote successfully and are idempotent; duplicate expanded-key rows were correctly held instead of silently picked/merged/written.

## Problem

`duplicate_expanded_key` means the current idempotency/identity shape collapses more than one expanded PLM row into the same target key. The system must not silently pick the first duplicate or merge by material code alone.

## Required decisions

1. Duplicate review UI
   - Show grouped duplicate summaries without exposing values in public issue evidence.
   - Let an authorized user inspect values in-tenant only.

2. Resolution policy
   Decide which policies are supported:
   - keep held / block;
   - skip selected duplicates;
   - merge quantities only when key business attributes match;
   - keep multiple rows by using a more granular idempotency key;
   - source correction / mapping correction.

3. Idempotency strategy
   - If keeping multiple rows, identify the stable discriminator: BOM path, parent context, source detail id, sort line, or another source-native line identity.
   - Do not use material/component code alone.

4. Apply strategy
   - Manual-confirm rows stay held until an explicit duplicate policy is chosen.
   - A future apply must rerun fresh dry-run first and use a fresh token.
   - No retry loop.

5. Evidence
   - Public evidence remains values-free: counts, statuses, conflict types, policy names, error codes only.

## Hard boundaries

This issue does not unlock:

- another Apply for the #2340 sample;
- production rollout;
- batch/multi-project mode;
- K3 Save / Submit / Audit / BOM;
- PLM/external database write;
[...]

PR #2378, opened 2026-06-08
docs(data-factory): design duplicate held-group decisions

zensgit:
## Summary

Docs-only D4 held-group decision design for #2343 after the D3 keep-multiple validation partial pass.

- records the D3 closeout as a clean partial pass: 1 group / 2 rows resolved, 27 groups / 54 rows still held with `missing_stable_discriminator`
- adds a D4 design contract for the remaining held groups: anchor strategy choice in legacy stock-preparation behavior + values-free quantity-shape/held-reason diagnostics
- locks that `merge_quantity` is not the default, unknown legacy behavior defaults to `source_correction_required`, and no additional Apply/runtime behavior is authorized

## Boundaries

- docs-only
- no route/UI/runtime/apply changes
- no idempotency-key/planner behavior change
- no K3, PLM write, external DB write, migration, or package change

## Verification

- `git diff --cached --check`

Refs #2343

### DIFF
--- a/docs/development/data-factory-plm-project-bom-stock-preparation-execution-plan-todo-20260604.md
+++ b/docs/development/data-factory-plm-project-bom-stock-preparation-execution-plan-todo-20260604.md
@@ -845,38 +845,87 @@ Acceptance locks:
 - No PLM write, external database write, K3 path, migration, package, or new
   policy execution beyond `keep_multiple_rows`.
 
-### 🟡 Duplicate-expanded-key D3 validation - #2340 sample keep-multiple closeout (PENDING - validation runbook)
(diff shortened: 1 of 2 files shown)


zensgit:
Fixed the row-vs-group count issue flagged in review.

Changes in `74f28343e`:
- pre-policy dry-run `manual_confirm` now uses row-level count `56` (28 held groups * 2 rows), so `190 + 56 = 246`;
- post-policy dry-run `manual_confirm` now uses row-level count `54`, so `2 + 190 + 54 = 246`;
- re-pull `manual_confirm` now uses row-level count `54`, so `192 + 54 = 246`;
- added an explicit note that dry-run decision counts are row-level, while `heldGroupCount` tracks groups separately.

`git diff --check` passes and the old `manual_confirm=...27/28` forms no longer appear in the touched docs. CI is rerunning on the amended commit.

PR #2382, opened 2026-06-08
docs(data-factory): add D4 held-group decision evidence template

zensgit:
## Summary
- Adds an operator-facing values-free evidence template for D4 held-group decisions after the D3 partial pass.
- Captures legacy stock-preparation behavior, quantity-shape counts, held reason counts, stable-discriminator counts, and policy recommendations without business values.
- Makes `source_correction_required` the default when legacy behavior is unknown, and keeps `merge_quantity`, `select_representative`, and `skip_selected` as separately gated future slices.

## Boundary
- Docs-only.
- Template explicitly keeps `applyAuthorizedNow:false`.

## Verification
- `git diff --cached --check`
- `git diff --check`
- Values scan: no project/component/target-id sample values in the template; only forbidden-evidence labels mention raw SQL/tokens as disallowed content.

Refs #2343

### DIFF
--- a/docs/operations/data-factory-plm-stock-preparation-d4-held-group-decision-evidence-template-20260608.md
+++ b/docs/operations/data-factory-plm-stock-preparation-d4-held-group-decision-evidence-template-20260608.md
@@ -1,0 +1,229 @@
+# Data Factory PLM Stock Preparation D4 Held-Group Decision Evidence Template - 2026-06-08
+
+## Purpose
+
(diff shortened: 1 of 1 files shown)


zensgit:
Addressed review feedback in `a1a01d700`: made the evidence template generic instead of hardcoding 27 groups, and aligned `stableDiscriminatorCounts` keys with the planner diagnostic shape (`any/sourceDetail/pathParent/sortLine`).

PR #2383, opened 2026-06-08
docs(data-factory): design source-correction duplicate holds

zensgit:
## Summary
- Adds a docs-only D4 design for `source_correction_required`, the conservative policy candidate for duplicate groups held with `missing_stable_discriminator`.
- Defines the policy as held-only: no target row write, no PLM/source write, no silent drop, no successful resolution claim.
- Locks default usage when legacy behavior is unknown, legacy rejects/reports, no stable discriminator exists, or quantity semantics are unsafe.

## Boundary
- Docs-only.

## Verification
- `git diff --cached --check`
- `git diff --check`

Refs #2343


zensgit:
Addressed review feedback in `0495676f0`: aligned evidence scope names with the existing policy scopes (`decision_only | run_only | table_scope`), avoiding new `future_*` variants.

PR #2384, opened 2026-06-08
docs(data-factory): design representative duplicate selection

zensgit:
## Summary
- Adds a docs-only D4 design for `select_representative`, the duplicate policy candidate for legacy dedup/keep-one behavior.
- Defines the policy as explicit representative selection plus values-free skipped-demand evidence.
- Locks out implicit pick-first, component-code inference, silent drop, PLM/source writes, and stale selector reuse.

## Boundary
- Docs-only.

## Verification
- `git diff --cached --check`
- `git diff --check`

Refs #2343


zensgit:
Addressed review feedback in `310fa6257`: aligned evidence scope names with existing policy scopes and flattened `quantityShapeCounts.*` keys so the evidence block stays flat/parseable.

PR #2385, opened 2026-06-08
docs(data-factory): design selected duplicate skips

zensgit:
## Summary
- Adds a docs-only D4 design for `skip_selected`, the duplicate policy candidate for explicitly ignored duplicate demand.
- Defines it as owner-reviewed selected skip with values-free skipped-demand evidence.
- Locks out default skipping, implicit skip-all-but-first, silent demand drop, stale selection reuse, PLM/source writes, and target writes for skipped rows.

## Boundary
- Docs-only.

## Verification
- `git diff --cached --check`
- `git diff --check`

Refs #2343


zensgit:
Addressed review feedback in `bd0ff5b5e`: aligned evidence scope names with the existing policy scopes (`decision_only | run_only | table_scope`).

PR #2386, opened 2026-06-08
docs(data-factory): design duplicate quantity merge

zensgit:
## Summary
- Adds a docs-only D4 design for `merge_quantity`, the duplicate policy candidate for confirmed additive demand.
- Defines sum-of-path-quantities semantics and rejects pick-one/max/min/average behavior.
- Locks that `merge_quantity` is not the default for `missing_stable_discriminator`; unknown legacy behavior still defaults to `source_correction_required`.

## Boundary
- Docs-only.

## Verification
- `git diff --cached --check`
- `git diff --check`
- Values scan: no project/component/target-id/raw-quantity sample values in the design.

Refs #2343


zensgit:
Addressed review feedback in `f47ee34d3`: aligned evidence scope names with the existing policy scopes (`decision_only | run_only | table_scope`).

Issue #2342, opened 2026-06-06
[Data Factory] PLM stock-preparation large-BOM strategy


zensgit:
## Goal

Define and implement the strategy for large PLM BOMs in the stock-preparation table action, after #2340 proved that a real owner-provided sample can exceed the current deployed row cap (`max_rows_exceeded` at `rowsExpanded=500`, `maxRows=500`).

This issue is a strategy/gate issue. It does not authorize C4 Apply by itself. It should keep #2340 from becoming a rushed cap-increase/apply gate.

## Trigger

In #2340, the second owner-provided sample dry-run showed:

- dry-run only; no Apply;
- `status=failed`;
- `canApply=false`;
- `expansion.status=failed`;
- `expansion.rowsExpanded=500`;
- `expansion.maxRows=500`;
- `expansion.readCount=1068`;
- `expansion.errorTypes=max_rows_exceeded`;
- plan was still computed for the bounded rows, but Apply correctly stayed blocked.

Values intentionally omitted: project number, PLM row values, target row values, request payload, dry-run token, source ids, sheet/view ids, field ids, target binding, connection details, credentials, and Bridge secrets.

## Important current-state note

The core expansion helper default is not inherently 500. The deployed action/config path is currently enforcing `maxRows=500` for the onsite gate. Any change must explicitly decide which limit applies and where it is configured.

## Required decisions

1. Dry-run cap strategy
   - Is 500 only a smoke-test cap, while production dry-run may use a higher reviewed cap?
   - What is the safe maximum for synchronous dry-run?
   - What evidence should show when output is bounded/truncated/fail-closed?

2. Large-BOM preview/readiness behavior
   - Should large BOM dry-run fail closed at cap, or return a bounded preview plus a separate `largeBom=true` readiness state?
   - What status/counts are safe to show values-free?

3. Apply strategy
   - Should large BOM Apply remain blocked until a batch/background/checkpointed writer exists?
   - If synchronous Apply is allowed at a higher cap, what extra owner approval is required?
   - How are partial failures, retries, and idempotency reported values-free?

4. UI/operator contract
   - The UI must not make a large BOM look like a normal one-click Apply.
   - It should show counts/status and require explicit owner review before any write.

5. Evidence restrictions
   - Continue values-free issue evidence only.
   - No project numbers, PLM row values, target row values, request payloads, dry-run tokens, source ids, sheet ids, field ids, targetBinding, raw SQL, credentials, or Bridge secrets.

## Hard boundaries

[...]

PR #2387, opened 2026-06-08
docs(data-factory): plan large-BOM background validation

zensgit:
## Summary

- add a docs-only #2342 test plan for future large-BOM C3 background expansion and C4 checkpointed apply implementation
- lock values-free evidence expectations, budget/error distinctions, durable checkpoint requirements, and negative controls
- keep the current synchronous bounded preview unchanged: no runtime, no route/UI, no package, no apply, no K3/PLM/external writes

## Verification

- git diff --check
- values-free scan for sample project/component ids, credentials, tokens, and raw payload markers

## Boundaries

This PR is implementation-prep only. It does not authorize runtime changes, background jobs, checkpoint writers, or another Apply.


zensgit:
Addressed review feedback in `9e7e25330`: clarified that D4 refers to #2343 duplicate-strategy D4 (not large-BOM C4), and clarified entity-machine as onsite hardware validation rather than a state-machine typo.

PR #2389, opened 2026-06-08
test(data-factory): add large-BOM job contract

zensgit:
## Summary

- add a latent #2342 large-BOM job contract helper for future C3 background expansion and C4 checkpointed apply
- pin background/apply status enums, values-free evidence projections, and a fail-closed authoritative-artifact gate
- wire the new focused test into plugin-integration-core's test chain

## Verification

- pnpm --filter plugin-integration-core test:stock-preparation-large-bom-jobs
- pnpm --filter plugin-integration-core test:stock-preparation-bom-expansion
- pnpm --filter plugin-integration-core test:stock-preparation-table-actions
- git diff --check
- values-free scan over the new helper/test/doc

Note: the full plugin test chain was attempted in the isolated temporary worktree, but that worktree lacks workspace node_modules and failed at the existing host-loader smoke step before the new test area (`ERR_MODULE_NOT_FOUND` for `tsx`). CI should run the full chain in the normal dependency environment.

## Boundaries

Latent helper only. No route, UI, worker, package, dry-run/apply behavior change, MetaSheet write, PLM write, external DB write, or K3 path.


zensgit:
Addressed review feedback in `cfa87865f`: `safeEvidenceToken` now fails closed on non-string provided values; `nonNegativeInteger` now accepts only safe non-negative integer numbers or numeric strings, rejecting booleans, arrays, whitespace-only strings, and other coercion shapes. Added tests for those invalid-type cases plus a numeric-string happy path.

PR #2393, opened 2026-06-09
feat(data-factory): add large-BOM expansion job skeleton

zensgit:
## Summary

Adds the #2342 C3-1/C3-2 background expansion job skeleton for PLM stock-preparation large BOMs.

This introduces:
- durable job-store gate (`context.storage.durable === true`) for large-BOM background expansion jobs;
- queued job creation with private principal + server-side action snapshot for future worker resume;
- values-free public job projection;
- start / inspect / cancel REST routes under table actions.

## Boundaries Held

- No C3 checkpointed expansion worker.
- No authoritative artifact.
- No C4 checkpointed apply.
- No source adapter read.
- No target records read/write.
- No PLM/external write, K3 write, raw SQL, or browser-supplied source/target/caps/plan/payload.
- Memory-only job storage fails closed.
- Missing authenticated principal fails closed; no system/admin/service fallback.

## Verification

Passed:

```sh
pnpm --filter plugin-integration-core test:stock-preparation-large-bom-jobs
pnpm --filter plugin-integration-core test:http-routes
pnpm --filter plugin-integration-core test:stock-preparation-table-actions
git diff --check
```

Not run to completion locally:

```sh
pnpm --filter plugin-integration-core test
```

The full plugin script stopped in this clean worktree before changed-code tests because `node --import tsx __tests__/host-loader-smoke.test.mjs` could not resolve `tsx`; pnpm also reported this worktree has no `node_modules`. CI should run this in a normal install environment.

## Notes

Verification doc: `docs/development/data-factory-plm-stock-preparation-large-bom-c3-job-skeleton-verification-20260608.md`.

PR #2394, opened 2026-06-09
feat(data-factory): add large-BOM expansion worker seam

zensgit:
## Summary

Stacked on #2393. Adds the first #2342 C3 background expansion worker seam.

This introduces `runLargeBomBackgroundExpansionJob(...)` in the large-BOM job helper:
- transitions runnable jobs to `running`;
- reads PLM through the provided source adapter using the existing app-side flat-read BOM expander;
- seals a private expansion artifact on valid completion;
- marks scale/correctness failures as non-authoritative `failed` jobs;
- exposes only values-free public evidence (`artifactRevisionPresent`, counts, status, source kind, read object/error tokens).

## Important Boundary

This is intentionally **not** the full checkpoint/frontier-resume implementation yet. It creates the worker + artifact seam for C3, but keeps explicit checkpoint resume as a later C3 slice.

Still out of scope:
- route/UI worker trigger;
- explicit frontier queue;
- chunk checkpoint resume;
- planner handoff;
- C4 checkpointed apply;
- target records read/write;
- PLM/external writes;
- K3 writes.

## Hardening

- C4 authoritative gate now requires `completed + authoritative + sealed artifact revision`.
- Completed job retry is idempotent: it returns the sealed artifact as-is, does not re-read source, and does not append duplicate artifact rows.
- Scale budget failure stays non-authoritative and exposes values-free bounded error evidence.

## Verification

Passed:

```sh
pnpm --filter plugin-integration-core test:stock-preparation-table-actions
```

Verification doc: `docs/development/data-factory-plm-stock-preparation-large-bom-c3-worker-verification-20260608.md`.

### DIFF
--- a/docs/development/data-factory-plm-stock-preparation-large-bom-c3-worker-verification-20260608.md
+++ b/docs/development/data-factory-plm-stock-preparation-large-bom-c3-worker-verification-20260608.md
@@ -1,0 +1,113 @@
+# Data Factory #2342 - C3 background expansion worker verification (2026-06-08)
+
+## Scope
+
+This stacked slice builds on the C3 job skeleton and adds the first background
+expansion worker seam:
+
+- run a queued/running/paused/failed background expansion job;
+- read the PLM source through the configured source adapter;
+- seal a private expansion artifact on success;
+- mark scale/correctness failures as non-authoritative failed jobs;
+- keep public job evidence values-free.
+
+It is intentionally not the full checkpoint/frontier resume implementation.
+It provides the execution and artifact seam that the later C3 checkpoint slice
+can split into resumable chunks.
+
+## Boundaries
+
+Still out of scope:
+
+- route/UI worker trigger;
+- explicit frontier queue;
+- chunk checkpoint resume;
+- authoritative planner handoff;
+- C4 checkpointed apply;
+- MetaSheet target reads or writes;
+- PLM/external writes;
+- K3 writes.
+
+The worker only calls `sourceAdapter.read(...)`. It never receives or uses a
+records API and therefore cannot create or patch target rows.
+
+## Worker Behavior
+
+`runLargeBomBackgroundExpansionJob(...)`:
+
+- requires durable job storage;
+- requires a source adapter with `read()`;
+- refuses cancelled/expired jobs;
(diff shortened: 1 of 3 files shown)


zensgit:
APPROVE — worker seam looks correct; I folded in the stack-comment cleanup and rebased it onto the updated #2393 head.

Verified locally after the rebase:

- `pnpm --filter plugin-integration-core test:stock-preparation-table-actions`
- `git diff --check origin/codex/df-large-bom-c3-job-skeleton-20260608..HEAD` plus worktree diff-check

What I verified:

- Worker reads through the injected source adapter using the existing app-side flat-read expansion path; every source call stays equality-filtered and never carries raw SQL/query text.
- Successful completion seals a private artifact with rows/summary + deterministic revision, and the public projection only exposes values-free counts/status/source-kind/read-object/error tokens.
- Scale/correctness failures remain non-authoritative (`failed`, no artifact, no artifactRevision), and the authoritative gate now requires `completed + authoritative + sealed artifact revision`.
- Completed-job retry is idempotent: returns the sealed artifact, does not re-read the source, and does not append duplicate artifact rows.
- Route/UI trigger, explicit checkpoint frontier resume, planner handoff, C4 apply, target records read/write, PLM/external/K3 writes all remain out of scope.

One stack hygiene fix was necessary after #2393 moved: this branch was still carrying the old “not wired to routes/workers” module comment and showed as DIRTY against the new base. I rebased onto #2393 `772b295d7`, resolved that comment to match the new worker boundary, reran the tests above, and force-pushed `4cc9629ef`.

Fresh CI is rerunning now; merge-ready once it settles green and the stack is landed in order.

zensgit:
Patched the C3 worker review finding.

- Rebased onto #2393 scoped job contract.
- Worker run now uses the scoped background job key.
- Worker catch now safely falls back to `read_failed` if `error.code` / `error.name` is not a safe evidence token, so the failed job is persisted instead of being left `running`.
- Test throws an unsafe-value-shaped error and asserts persisted failed/read_failed values-free evidence.

Local verification: `node plugins/plugin-integration-core/__tests__/stock-preparation-large-bom-jobs.test.cjs`.

PR #2396, opened 2026-06-09
feat(data-factory): add large-BOM planner handoff

zensgit:
## Summary

Stacked on #2394. Adds the read-only #2342 C3 planner handoff after a background expansion job completes.

This introduces `planLargeBomBackgroundExpansionJob(...)`:
- requires a completed authoritative expansion artifact;
- combines private expanded rows with caller-provided existing target rows;
- reuses the existing `planStockPreparationConflicts` C3 planner;
- stores a private plan artifact + revision on the job;
- exposes values-free plan evidence through the public job projection.

## Boundary

Still no route/UI, no target records API read, no target write, no PLM/external write, no K3, no C4 checkpointed apply.

The helper intentionally accepts `existingRows` as input. Future route/worker wiring must load those rows from the server-scoped configured target binding; browser-supplied target scope remains forbidden.

## Hardening

- Planner handoff refuses non-authoritative jobs through `LARGE_BOM_ARTIFACT_NOT_AUTHORITATIVE`.
- Public projection exposes `planRevisionPresent`, not the revision itself.
- Public plan evidence is the existing values-free planner evidence: counts, row counts, conflict types, duplicate diagnostics/resolution fingerprints and policy labels.

## Verification

Passed:

```sh
pnpm --filter plugin-integration-core test:stock-preparation-conflict-planner
```

Verification doc: `docs/development/data-factory-plm-stock-preparation-large-bom-c3-planner-handoff-verification-20260608.md`.

### DIFF
--- a/docs/development/data-factory-plm-stock-preparation-large-bom-c3-planner-handoff-verification-20260608.md
+++ b/docs/development/data-factory-plm-stock-preparation-large-bom-c3-planner-handoff-verification-20260608.md
@@ -1,0 +1,82 @@
+# Data Factory #2342 - C3 large-BOM planner handoff verification (2026-06-08)
+
+## Scope
+
+This stacked slice adds the read-only C3 planner handoff after a background
+expansion job completes:
+
+- load a completed authoritative expansion artifact;
+- combine its private expanded rows with caller-provided existing target rows;
+- reuse the existing C3 `planStockPreparationConflicts` planner;
+- store a private plan artifact and revision on the job;
+- expose values-free plan evidence through the public job projection.
+
+It does not add a route, UI, target records API read, target write, PLM write,
+K3 path, or C4 checkpointed apply.
+
+## Contract
+
+`planLargeBomBackgroundExpansionJob(...)` requires an authoritative artifact:
+
+- `status='completed'`;
+- `authoritative=true`;
+- sealed artifact revision present.
+
+Queued/running/failed/cancelled/expired jobs and completed jobs without a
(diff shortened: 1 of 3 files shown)


zensgit:
APPROVE — planner handoff is clean after one fail-closed hardening I folded into the branch.

Verified locally after rebasing onto the updated #2394 head:

- `pnpm --filter plugin-integration-core test:stock-preparation-conflict-planner`
- `git diff --check origin/codex/df-large-bom-c3-worker-20260608..HEAD` plus worktree diff-check

What I verified:

- `planLargeBomBackgroundExpansionJob(...)` refuses non-authoritative jobs through `LARGE_BOM_ARTIFACT_NOT_AUTHORITATIVE`; bounded/failed/running artifacts cannot reach planning.
- The helper reuses the existing C3 planner over the private sealed expansion rows plus server-provided `existingRows`; no route/UI/browser payload path is introduced in this slice.
- Public projection exposes only `planRevisionPresent` plus existing values-free planner evidence. Private plan decisions, existing target values, target row ids, sheet/field ids, project/component/source/path values stay out of public evidence.
- No target records API read/write, no C4 apply, no PLM/external/K3 write, no raw SQL.

I found one small fail-closed gap while reviewing: malformed entries inside `existingRows` were silently filtered. Since future wiring will load `existingRows` from the server-scoped target binding, silently dropping a malformed target row could make the planner think a row is missing and generate an unsafe add. I changed that to reject with `LARGE_BOM_PLAN_EXISTING_ROWS_INVALID` and added a regression test proving no `planRevision` is persisted on that path. That hardening is in `51e314f8a`.

PR is CLEAN against #2394 and pr-validate is green; merge-ready when the stack reaches it.

PR #2397, opened 2026-06-09
feat(data-factory): add large-BOM checkpoint apply writer

zensgit:
## Summary

Adds the latent C4 checkpoint apply writer service for the large-BOM path.

- creates checkpoint apply jobs only from a completed authoritative C3 plan artifact;
- requires durable storage, authenticated principal, and Data Factory write/admin permission;
- requires explicit manual-confirm acknowledgement before apply job creation when held rows are present;
- applies plan decisions in bounded chunks by reusing the existing `applyStockPreparationPlan` writer;
- persists checkpoint progress and keeps retry safe through the existing idempotent add/find-then-patch behavior;
- exposes values-free public apply-job summaries only.

## Boundary

Still not wired to any route, UI, worker scheduler, production apply button, PLM read, external DB write, K3 write, Submit/Audit/BOM, or batch rollout. This is a service-layer C4 primitive stacked on #2396.

## Verification

```text
pnpm --filter plugin-integration-core test:stock-preparation-apply-writer
pnpm --filter plugin-integration-core test:stock-preparation-conflict-planner
git diff --check
```

Verification doc: `docs/development/data-factory-plm-stock-preparation-large-bom-c4-writer-verification-20260608.md`.

## Notes

The retry test simulates a write-after-success / checkpoint-lost window by resetting the checkpoint to the same decision after the target row already exists. The rerun patches the existing row and does not create a duplicate.

### DIFF
--- a/docs/development/data-factory-plm-stock-preparation-large-bom-c4-writer-verification-20260608.md
+++ b/docs/development/data-factory-plm-stock-preparation-large-bom-c4-writer-verification-20260608.md
@@ -1,0 +1,53 @@
+# DF Large-BOM C4 Checkpoint Apply Writer Verification - 2026-06-08
+
+## Scope
+
+Implemented the latent C4 checkpoint apply writer service for large-BOM jobs.
+
+- Creates a checkpoint apply job only from a completed authoritative C3 plan artifact.
+- Requires durable job storage, authenticated principal, and Data Factory `write` or `admin` permission.
+- Requires explicit manual-confirm acknowledgement when the C3 plan contains held rows.
+- Applies plan decisions in bounded chunks by reusing the existing C4 `applyStockPreparationPlan` writer.
+- Persists checkpoint progress after each chunk; retry remains safe because add decisions are idempotent find-then-patch.
+- Public projection is values-free: counts, status, revision-presence booleans, result status tokens, error-code tokens, and field-category tokens only.
+
+## Boundary
+
+Not included in this slice:
+
+- No route or UI.
+- No browser-supplied plan, target, sheet id, field id, or payload.
+- No PLM/source read.
+- No target read outside the injected MetaSheet records API used by the existing C4 writer.
+- No external DB write, K3 write, Submit, Audit, BOM, production rollout, or batch apply enablement.
+
+## Tests
+
+Local verification:
+
+```text
+pnpm --filter plugin-integration-core test:stock-preparation-large-bom-jobs
+pnpm --filter plugin-integration-core test:stock-preparation-apply-writer
+pnpm --filter plugin-integration-core test:stock-preparation-conflict-planner
+pnpm --filter plugin-integration-core test:stock-preparation-bom-expansion
+pnpm --filter plugin-integration-core test:http-routes
+```
+
+Covered cases:
+
+- durable storage is mandatory;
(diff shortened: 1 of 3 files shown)


zensgit:
APPROVE — C4 checkpoint apply writer is sound after one preflight hardening I folded into the branch.

Verified locally after rebasing onto #2398 and after the fix:

- `pnpm --filter plugin-integration-core test:stock-preparation-apply-writer`
- `pnpm --filter plugin-integration-core test:stock-preparation-conflict-planner`
- `git diff --check origin/codex/df-large-bom-c3-route-20260608..HEAD` plus worktree diff-check

What I verified:

- Apply jobs can only be created from completed authoritative C3 plan artifacts, with durable storage, authenticated principal, and Data Factory write/admin permission.
- Held/manual-confirm rows require explicit acknowledgement before apply job creation.
- Chunk execution reuses the existing `applyStockPreparationPlan` writer, preserving find-then-patch/idempotent add behavior and human-field protections.
- Checkpoint retry after a lost checkpoint does not duplicate adds: the rerun patches the already-created row and keeps one target row.
- Public apply-job summaries are values-free: counts/status/error-code categories only; no project/component/source/target values, sheet ids, field ids, credentials, or tokens.
- Still latent: no route/UI/scheduler, no PLM read, no external DB write, no K3, no production/batch rollout.

I added one defensive preflight: `runLargeBomCheckpointApplyJobChunk` now validates `recordsApi` before marking the job `running`. Without that, a miswired route with a missing records API could leave a queued job stuck as `running` with no checkpoint/evidence. The new test asserts missing records API returns `LARGE_BOM_APPLY_RECORDS_API_UNAVAILABLE` and leaves status/checkpoint unchanged. This is in `9fea93139`.

PR is CLEAN against #2398 and pr-validate is green; merge-ready when the stack reaches it.

zensgit:
Patched the C4 checkpoint-apply concurrency finding.

- Checkpoint apply jobs are now tenant/workspace scoped too, not just background expansion jobs.
- `running` is now the in-flight chunk state: another run returns 409 `LARGE_BOM_APPLY_RUN_IN_PROGRESS` before writing.
- Completed non-terminal chunks now persist `paused`, so the next explicit run can continue from the advanced checkpoint.
- Tests lock concurrent running rejection, no duplicate create/patch, and values-free public evidence.

Local verification: `node plugins/plugin-integration-core/__tests__/stock-preparation-large-bom-jobs.test.cjs`.

zensgit:
Follow-up after Tesla re-review: strengthened the concurrency fix.

The earlier persisted-`running` guard did not cover two callers that both read `queued`/`paused` before either wrote `running`. This head now adds a module-level single-flight guard keyed by the scoped apply job key.

- Same-process concurrent `run` attempts on the same tenant/workspace/action/applyJobId now fail with `409 LARGE_BOM_APPLY_RUN_IN_PROGRESS` before the second write path starts.
- `running` remains the durable in-flight marker; completed non-terminal chunks persist `paused` for the next explicit run.
- Added a blocking recordsApi test that starts one run, holds it inside `createRecord`, then asserts a second concurrent run is rejected and no duplicate create/patch occurs.

Local verification: `node plugins/plugin-integration-core/__tests__/stock-preparation-large-bom-jobs.test.cjs`.

PR #2402, opened 2026-06-09
docs(data-factory): design #2388 source snapshot diff gate

zensgit:
## Summary

Adds the #2388 C0 design for a PLM-BOM source snapshot + snapshot-diff lifecycle gate.

This is docs-only. It designs the generic Data Factory `SourceSnapshot` / `SourceDiffResult` seam while keeping the first implementation scoped to the PLM-BOM pilot.

## Locked Scope

- Treat already-shipped behavior as invariants to verify, not rebuild:
  - PLM-missing rows mark inactive, never physical delete.
  - Human/manual fields remain protected from PLM refresh payloads.
  - Apply requires a fresh dry-run token/revision gate.
- New #2388 scope:
  - private in-tenant source snapshots;
  - latest-pull vs last-applied-snapshot diff;
  - `missing_child_bom` held behavior.
- PLM snapshot internals use `ComponentNode` + `BomEdge`.
- Node/edge identity is internal to snapshot/diff only; this PR does not migrate the flat target `idempotencyKey` scheme.
- `missing_child_bom` requires an explicit source signal; missing or ambiguous signal fails closed and must not be treated as a complete leaf.
- Full cross-source generalization waits for a second real source.

## Boundary

No runtime, route, UI, migration, package, worker, MetaSheet write, PLM write, external database write, K3 path, raw SQL, or production rollout.

## Verification

- Checked #2388 issue/comments through GitHub REST API because GraphQL intermittently returned EOF.
- Checked current helper grounding in:
  - `stock-preparation-bom-expansion.cjs`
  - `stock-preparation-conflict-planner.cjs`
  - `stock-preparation-apply-writer.cjs`
  - `stock-preparation-table-actions.cjs`
- `git diff --check`
- `git diff --check HEAD~1..HEAD`

### DIFF
--- a/docs/development/data-factory-plm-project-bom-stock-preparation-execution-plan-todo-20260604.md
+++ b/docs/development/data-factory-plm-project-bom-stock-preparation-execution-plan-todo-20260604.md
@@ -927,10 +927,53 @@ Acceptance locks:
 - Any future executable policy requires a fresh dry-run, fresh token, reviewed
   policy evidence, and explicit owner acknowledgement.
 
+### 🟡 Source snapshot diff gate C0 - PLM-BOM lifecycle design (ACTIVE - #2388)
+
+Gated on: #2388 explicit direction + no runtime opt-in.
+
+Design doc:
+`docs/development/data-factory-plm-stock-preparation-source-snapshot-diff-c0-design-20260608.md`.
+
+Scope:
+
+- Design the Data Factory `SourceSnapshot` / `SourceDiffResult` seam while
+  implementing only the PLM-BOM pilot in later slices.
+- Separate already-shipped invariants from new work:
+  `mark_inactive` for PLM-missing rows, human-field preservation, and fresh
+  dry-run token are invariants to verify, not rebuild.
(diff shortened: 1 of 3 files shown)


zensgit:
Verdict: APPROVE, with the merge-order note that this should land before the #2342 runtime stack.

I re-read the design against the current concern: #2402 does lock the load-bearing #2388 semantics, not just lifecycle prose:

- `missing_child_bom` requires an explicit source signal that a component should have children.
- childless + explicit leaf signal is a valid leaf;
- childless + explicit assembly/has-BOM signal is held as `missing_child_bom`;
- childless + no usable signal is held, not guessed as a complete leaf.
- large-BOM background expansion must not produce authoritative `complete` artifacts while incomplete-assembly semantics are unresolved.

That is exactly the semantic foundation that should precede any large-scale #2342 runtime merge. The PR is docs-only, values-free, and keeps node/edge identity internal to snapshot/diff without silently migrating the flat target idempotency key.

Merge mechanics: it is currently behind latest main, so it needs the normal rebase -> fresh CI -> squash chain before landing.

PR #2409, opened 2026-06-09
docs(attendance): lock multi-shift day design

zensgit:
## Summary
- Add a design-lock for the DingTalk benchmark item "一天多班次" / multi-shift day support.
- Backfill the attendance benchmark tracker from ⬜ to 🟡 for this item, with implementation still locked behind staged opt-ins.
- Ground the design in the current one-shift assumptions: runtime conflict checks, single-row effective schedule projections, and day-level attendance records.

## Locked decisions
- v1 is default-off, direct-shift-only, max 3 slots (`slot_index` / `slotIndex` 0..2).
- Same-slot date overlap is rejected; different-slot overlap is allowed only when shift time windows do not overlap.
- Rotation assignments remain exclusive in v1.
- Effective-calendar/planned-minute surfaces must expose `slots[]` and sum planned minutes; day-level records remain one row per user/day.
- Fixed apply, auto-shift A1, compliance caps, admin UI, and staging smoke each have explicit follow-up slices.

## Scope
Docs only. No runtime code, schema migration, frontend UI, OpenAPI, or staging changes in this PR.

## Verification
- `git diff --cached --check`
- `git diff --cached --stat`


zensgit:
Review follow-up addressed on head `102e9baef`:\n\n- N1: §3 now explicitly reconciles the `slot_index` storage delta with the tracker's hot-table migration batching rule. M0 may ship a standalone latent column only after pre-flight confirms adjacent `status` / `locked_at` / shift-constraint DDL is not ready to batch; otherwise those changes must be coordinated into one migration set.\n- N2: §4 now states that if effective-calendar `slots[]` is part of the OpenAPI contract, M2 must update the contract and generated artifacts in the same slice.\n- N3: §5 now states multi-shift is schedule-side only and does not change record-side first/last or S2 internal/outdoor merge semantics.\n\nFresh-green after rebase onto `origin/main@59c883c0`: PR is MERGEABLE / CLEAN; contracts, pr-validate, plugin tests Node 18/20, after-sales, and coverage all pass.

PR #2410, opened 2026-06-09
docs(multitable): design formula-over-lookup A-full

zensgit:
## Summary

Docs-only design-lock for the formula-over-lookup A-full slice: foreign-record target changes should propagate through related lookup/rollup fields and recompute formulas on the related records.

This builds on the shipped A-min path and create/submit/import hydration work, but explicitly stays implementation-free.

## Locked decisions

- Use the existing related-record recompute seam instead of a new derived graph engine.
- Keep A-full bounded to one hop: `foreign source field -> related lookup/rollup -> related formula`.
- Keep full hydrated related rows for formula eval; use field-masked patches only for echo.
- Gate affected computed field ids by source sheet + changed target field + linked updated record.
- Preserve the current permission/reachability boundary and avoid cross-sheet realtime fan-out in this slice.

## Out of scope

- Runtime implementation.
- Recursive derived graph / formula-to-formula.
- Dry-run hydration.
- Lookup/rollup materialization.
- Parser or `src/formula/engine.ts` changes.
- Realtime fan-out, frontend, central RBAC/auth, migrations.

## Verification

Docs-only. Local checks:

- `git diff --check`

### DIFF
--- a/docs/development/multitable-formula-over-lookup-afull-design-20260609.md
+++ b/docs/development/multitable-formula-over-lookup-afull-design-20260609.md
@@ -1,0 +1,207 @@
+# Multitable Formula-over-Lookup A-full — Design-Lock — 2026-06-09
+
+> Status: **DESIGN-LOCK (docs-only). No runtime code in this PR.**
+> Builds on `multitable-formula-over-lookup-design-20260603.md` (A-min) and
+> `multitable-formula-over-lookup-create-design-20260603.md` (create/submit/import).
+> Current recon: `origin/main` @ `db6b128ee`.
+> K3 / frozen: multitable kernel-polish only. Do **not** touch
+> `packages/core-backend/src/formula/engine.ts`, central RBAC/auth, migrations, or
+> `plugin-integration-core`.
+
+## 0. Scope in one line
+
+Support the bounded A-full case: when a **foreign record's target field** changes,
+existing linked records whose `lookup`/`rollup` values change should also recompute
+their own formula fields that reference those computed fields.
+
+This is **not** a general derived-field graph, not recursive transitive propagation,
+not lookup/rollup materialization, and not dry-run hydration. It is one synchronous
+write-path slice on top of the existing `POST /api/multitable/patch` related-record
+recompute seam.
+
+## 1. Current-state ground truth
+
+### 1.1 What is already shipped
+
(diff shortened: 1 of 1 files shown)


zensgit:
**APPROVE (design-lock).** Code-grounded against `origin/main@db6b128ee` — the anchors check out: `computeDependentLookupRollupRecords` returns a **masked echo** (`filterRecordDataByFieldIds(extractLookupRollupData(...), allowedFieldIds)`) with no formula recalc (the gap is real), `recalculateFormulaFields`'s `depends_on_field_id` gate is as described, and realtime publishes the edited sheet only. The design is sound, one-hop-bounded, reuses the existing helper, and respects the frozen lines (AF8).

**§3.2 is the load-bearing call and it's right:** two views — a full hydrated row for `recalculateRecordFromData()` (reader-agnostic, like A-min) vs. a masked echo for the HTTP response. Feeding the permission-filtered echo into eval would compute formulas against nulled inputs (wrong + cross-actor-inconsistent). Keep that strict.

Three design judgment calls worth folding a sentence each into §3.4/§2 **before** the impl PR (documented bounds, not blockers):

1. **Actor-dependent consistency (§3.4):** propagation runs on the editing actor's request, gated by *their* read access — so a foreign edit by a user who can't read the related sheet leaves its formulas stale, and the persisted value depends on who last edited. Strictly better than always-stale A-min, async-graph deferred — just confirm "best-effort, editor-gated consistency" is the accepted semantic (reports read the materialized value).
2. **Cross-record write-on-read:** A-full **persists a formula on a *related* record the actor may only be able to read** (A-min wrote the editor's own record). It's a derived-value side-effect, so read-gating is defensible — state it explicitly so it isn't read as a write-permission escalation.
3. **Bound is depth, not breadth:** one hop, but a high-fan-in foreign record recomputes formulas on *all* related records synchronously on the write path. Acknowledge the fan-in cost (breadth cap / log-and-skip, or explicitly accept at current scale).

Note (no action): A-full inherits — doesn't worsen — the pre-existing "formula-result encodes its inputs" property (AF2 gates the formula field by its own visibility; the one-hop value matches what a same-record formula would expose).

Impl-PR boundary (§7) is correctly scoped and matches the A-min lineage discipline. Independent review MD: /tmp/pr2410-review-claude-20260609.md.

PR #2523, opened 2026-06-12
fix(multitable): guard hierarchy parent field downgrades

zensgit:
## Summary

Closes the S4 residual where a field already used as a hierarchy `config.parentFieldId` could later be downgraded through `PATCH /fields/:fieldId`.

- adds a reverse guard in the field PATCH path: an explicit hierarchy parent field that is currently a single-value link cannot be changed to a multi-value link or non-link while the hierarchy view still points at it
- keeps the existing boundary: auto/first-link hierarchy mode is unchanged, and unreferenced link fields can still be converted to multi-value
- extends the existing S4 route-level test harness with field PATCH cases

## Verification

- `pnpm --filter @metasheet/core-backend exec vitest run tests/unit/multitable-hierarchy-parent-link-guard.test.ts --watch=false`
- `git diff --check`

Note: targeted ESLint on `univer-meta.ts` still reports pre-existing lint debt in that large route file (for example old no-useless-escape / no-extra-semi entries outside this patch), so I did not treat that as a new failure for this slice.

### DIFF
--- a/packages/core-backend/src/routes/univer-meta.ts
+++ b/packages/core-backend/src/routes/univer-meta.ts
@@ -4335,11 +4335,13 @@ async function validateGanttDependencyConfig(
 // link field via a generic record patch (no dedicated reparent endpoint), so a MULTI-value
 // link used as parent gets silently overwritten on every drag-to-reparent. Reject saving a
 // hierarchy view whose explicit parentFieldId is not a single-value (`limitSingleRecord`)
-// link field. An absent parentFieldId is allowed (runtime auto/first-link fallback unchanged).
-// Documented residuals (accepted): (a) PATCH /fields/:fieldId can later flip limitSingleRecord
-// or the field type without view re-validation — same pre-existing class as gantt's dependency
-// field, follow-up candidate; (b) provisioning ensureView/createView bypass route-layer view
-// validation by design (template library ships no hierarchy views today).
+// same-sheet link field. An absent parentFieldId is allowed (runtime auto/first-link fallback unchanged).
+// PATCH /fields/:fieldId carries the matching reverse guard below: once a hierarchy
+// view explicitly uses a single-value link as its parent field, that field cannot
+// be downgraded to multi-value, non-link, or a different target sheet while the
+// view still points at it.
+// Documented residual (accepted): provisioning ensureView/createView bypasses
+// route-layer view validation by design (template library ships no hierarchy views today).
 async function validateHierarchyParentLinkConfig(
   query: QueryFn,
   sheetId: string,
@@ -4347,7 +4349,7 @@ async function validateHierarchyParentLinkConfig(
   config: Record<string, unknown>,
 ): Promise<string | null> {
   if (viewType !== 'hierarchy') return null
-  const parentFieldId = typeof config.parentFieldId === 'string' ? config.parentFieldId.trim() : ''
+  const parentFieldId = normalizeHierarchyParentFieldId(config.parentFieldId)
   if (!parentFieldId) return null
 
   const fieldRes = await query(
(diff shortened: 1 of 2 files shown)


zensgit:
S4 field-downgrade guard is ready for review.

What changed:
- `PATCH /fields/:fieldId` now has the reverse guard for the S4 hierarchy invariant: if a field is currently a single-value link and an explicit hierarchy view uses it as `config.parentFieldId`, the field cannot be downgraded to multi-value or non-link until that view is changed.
- Existing boundaries stay intact: hierarchy auto/first-link mode is unchanged, unreferenced link fields can still become multi-value, and rename/still-single-value edits are allowed.

Verification:
- `pnpm --filter @metasheet/core-backend exec vitest run tests/unit/multitable-hierarchy-parent-link-guard.test.ts --watch=false` -> 10/10 pass
- `pnpm --filter @metasheet/core-backend build` -> pass
- `git diff --check` -> pass
- Independent subagent review: no findings; confirmed the guard is scoped to explicit hierarchy parent refs, does not overblock rename/unreferenced fields, and the tests are meaningful/negative-control capable.
- GitHub CI: CLEAN, all required checks green; Strict E2E skipped as usual.

Note: targeted ESLint on `univer-meta.ts` still reports pre-existing lint debt outside this patch, so I did not widen this slice into a large lint cleanup.

zensgit:
Review: APPROVE, no findings.

What I checked:
- The reverse guard runs on `PATCH /api/multitable/fields/:fieldId` after sheet field-management permission is established and after the next field shape is normalized, but before any `UPDATE meta_fields` write.
- It only blocks a referenced single-value hierarchy parent link when the next shape is no longer a single-value link.
- It allows safe cases: rename while staying single-value, unrelated/unreferenced link fields switching to multi-record, hierarchy views without explicit `parentFieldId`, and non-hierarchy views carrying stray `parentFieldId`.

Verification:
- `pnpm --filter @metasheet/core-backend exec vitest run tests/unit/multitable-hierarchy-parent-link-guard.test.ts --watch=false` → 10/10 pass
- `git diff --check` → clean

Ordering note: this is the backend safety net and should land before #2525, which is the matching UI guard.

zensgit:
Update after re-checking the subagent findings: the earlier approve comment is superseded until this patched head settles.

Both P1s are now fixed in `dd09bcf4f`:

- The reverse guard now trims persisted hierarchy `parentFieldId` values (`btrim(config->>'parentFieldId')`) before matching, so a view stored with whitespace around the field id still blocks a downgrade.
- Hierarchy parent links now have to remain same-sheet single-value links. The forward create/update guard rejects an external-sheet parent link, and the reverse field mutation guard blocks removing or retargeting the parent link's target sheet while an active hierarchy view references it.

Verification on the patched head:

- `pnpm --filter @metasheet/core-backend exec vitest run tests/unit/multitable-hierarchy-parent-link-guard.test.ts --watch=false` → 15/15 pass
- `git diff --check` → clean

CI is rerunning on the new head. I’ve asked for re-review against the two P1 closures before calling it merge-ready again.

zensgit:
Second patch on the remaining P1 from re-review: fixed the whitespace-class mismatch.

The bypass was real: forward validation uses JS `.trim()`, while PostgreSQL `btrim(text)` only trims plain spaces by default. A persisted `parentFieldId` wrapped in tabs/newlines could validate on write and then miss the reverse downgrade lookup.

Patched in `2719bb102`:

- Reverse lookup now uses `btrim(config->>'parentFieldId', E' \\t\\n\\r\\f\\v')` so spaces, tabs, newlines, CR, form feed, and vertical tab are treated consistently with the route validation intent.
- Added a route-level regression for `parentFieldId: '\\tfld_parent_single\\n'` that drives the actual PATCH /fields downgrade path and asserts no field update occurs.

Verification:

- `pnpm --filter @metasheet/core-backend exec vitest run tests/unit/multitable-hierarchy-parent-link-guard.test.ts --watch=false` → 16/16 pass
- `pnpm --filter @metasheet/core-backend build` → pass
- `git diff --check` → clean

CI is rerunning on the new head; I asked for re-review of this final P1 closure.

zensgit:
Tiny follow-up on the previous patch: I amended the SQL literal to keep the whitespace set as PostgreSQL `E''` escape sequences (`E' \\t\\n\\r\\f\\v'`) instead of relying on JS runtime control characters. Same behavior, clearer boundary for future review.

Final patched head is now `06511aa04`.

Verification repeated:

- `pnpm --filter @metasheet/core-backend exec vitest run tests/unit/multitable-hierarchy-parent-link-guard.test.ts --watch=false` → 16/16 pass
- `pnpm --filter @metasheet/core-backend build` → pass
- `git diff --check` → clean

zensgit:
Final patch for the whitespace mismatch P1 is now in `ed8d7ff94`.

I replaced the SQL-trim approach with the cleaner invariant:

- POST/PATCH `/views` now canonicalize hierarchy `config.parentFieldId` before persistence, using the same JS trim semantics as validation. New dirty values do not get saved into `meta_views.config`.
- The reverse field-mutation guard now reads candidate hierarchy view configs and compares `parentFieldId` through the same normalization helper, so legacy configs with Unicode whitespace (for example NBSP) still block downgrade/retarget.
- The route-level tests now cover:
  - PATCH view normalizes NBSP-wrapped parentFieldId before saving;
  - POST view normalizes NBSP-wrapped parentFieldId before saving;
  - legacy NBSP-wrapped persisted config still blocks PATCH /fields downgrade;
  - existing space/tab/newline persisted-config regressions remain covered.

Verification repeated on `ed8d7ff94`:

- `pnpm --filter @metasheet/core-backend exec vitest run tests/unit/multitable-hierarchy-parent-link-guard.test.ts --watch=false` → 19/19 pass
- `pnpm --filter @metasheet/core-backend build` → pass
- `git diff --check` → clean

This should close the JS `.trim()` vs PostgreSQL trim mismatch fully, including Unicode whitespace and historical dirty config.

zensgit:
Final status on `ed8d7ff94`: re-review is clean and CI has settled green.

The final fix closes the whitespace mismatch by sharing one normalization path:

- validation uses `normalizeHierarchyParentFieldId()`;
- POST/PATCH view persistence canonicalizes hierarchy `config.parentFieldId` before saving;
- the reverse field-mutation guard reads candidate hierarchy configs and compares through the same helper, so legacy dirty configs with Unicode whitespace still block downgrade/retarget.

Subagent re-review: APPROVE — remaining whitespace P1 is closed, including NBSP/historical dirty config.

CI: all checks completed; required checks green, Strict E2E skipped as expected.

Local verification on the final head:

- `pnpm --filter @metasheet/core-backend exec vitest run tests/unit/multitable-hierarchy-parent-link-guard.test.ts --watch=false` → 19/19 pass
- `pnpm --filter @metasheet/core-backend build` → pass
- `git diff --check` → clean

This PR is back to merge-ready from the hierarchy parent downgrade-guard side.

zensgit:
Rebased onto current `main` (was 98 behind on the hot `univer-meta.ts`) and verified — the rebase applied cleanly with no conflicts despite the cross-base churn. Checks against current main:

- backend `tsc` build: clean (exit 0)
- guard unit suite (`multitable-hierarchy-parent-link-guard.test.ts`): 19/19
- regression sweep — view-config integration (`multitable-view-config.api` + `viewconfig-resave-guard` + `viewconfig-filter-literal-redaction`): 28/28 (the tightened same-sheet `validateHierarchyParentLinkConfig` + the new field-PATCH reverse guard don't break existing valid view/field paths)

Review verdict stands: **APPROVE** — sound, and slightly broader than the rank-10 residual (also requires the hierarchy parent link to be same-sheet). Closes the accepted residual at the old `univer-meta.ts:4339` (field PATCH could flip `limitSingleRecord` / type / target with a hierarchy view still pointing at it). CI re-running on the rebased head.

PR #2554, opened 2026-06-12
feat(multitable): record locking storage + write-path enforcement + re-expose (rank 8)

zensgit:
Implements the full **lock_record storage contract** (design-lock `docs/development/multitable-lock-record-storage-contract-design-20260611.md`), completing the action that #2278 only hid from the rule-editor dropdown. Today `executeLockRecord` UPDATEs a non-existent `meta_records.locked` column → runtime crash; this builds the real store, the write-path enforcement, the wire round-trip, and re-exposes the action in the UI.

## The 6 design decisions, as-built (owner ratified "按推荐")
| # | Decision | As-built |
|---|---|---|
| a | Lock granularity | **Whole record** |
| b | Unlock permission (`canUnlock`) | **locker ∨ owner ∨ sheet-admin** — gates the unlock *action* + frontend action visibility |
| c | Manual vs automation lock | **Single shared column set** (`locked`/`locked_by`/`locked_at`) |
| d | Lock scope | **Block edits AND deletes; comments still allowed** (comments are a separate path — they never go through `patchRecords`/`deleteRecord`) |
| e | Admin bypass | **No silent bypass** — the write/delete guards reject even an admin's edit while locked; admin must explicitly unlock first |
| f | Unlock trigger | **Manual action + automation unlock** (`config.locked === false`) |

### Reconciling decision e with §2.2's `!canUnlock` prose
§2.2 literally writes the guard as `recordRow.locked && !canUnlock(actor)`, but `canUnlock` includes admin, while decision e + LR-T6 require an admin's edit to be **rejected** while locked. Resolution: a focused **`canEditWhileLocked(actorId, record)` = locker ∨ owner** (no admin) gates the per-mutation write/delete path, while `canUnlock` (locker ∨ owner ∨ admin) gates the explicit unlock action and the frontend action visibility. This keeps the mutation path free of per-mutation admin-bypass logic (decision e's intent) and makes LR-T6 pass.

## Migration / tier
`packages/core-backend/src/db/migrations/zzzz20260612140000_add_meta_record_locked.ts` — **zzzz tier**, sorts strictly after the current highest (`zzzz20260612130000_*`). Adds `locked boolean NOT NULL DEFAULT false`, `locked_by text NULL`, `locked_at timestamptz NULL`, and `idx_meta_records_locked_by`, mirroring the `modified_by` precedent. Clean `down()` (verified live: down→0 cols, up→3 cols, in a rolled-back txn). Existing rows default `locked=false`.

## Wire round-trip (the wire-vs-fixture trap)
Lock state rides as **TOP-LEVEL record metadata, NOT inside `data`** → it is never swept by §2a.3's `filterRecordDataByFieldIds` / `maskStoredRecordFieldIds` masking. Threaded through:
- `GET /view` (the grid loader) — **all 5 SELECT branches** + the `UniverMetaRecord` type
- `GET /records` (cursor list) — `query-service` `mapRecordRow` + `LoadedMultitableRecord` + route serialization
- `GET /records/:recordId` (drawer single-read)
- the **PATCH echo**
A server-authoritative per-row **`canUnlock` boolean** is computed for locked rows (via a creator-map scoped to locked ids) so the client never receives `created_by`. New **`POST /records/:recordId/lock`** route performs manual lock/unlock (lock requires edit rights; unlock requires `canUnlock`).

`LR-T10` proves the columns round-trip through the **real** wire (write→lock→re-read on `/view` + `/records`), and asserts the lock fields are top-level (not in `data`). The §2a.3 mask suites stay green (36 tests) — confirming lock columns don't perturb masking.

## `canUnlock` implementation
`packages/core-backend/src/multitable/record-lock.ts` — pure, zero new primitives: `canUnlock` returns true for `canManageSheetAccess`, else locker (`lockedBy===actor`) ∨ owner (`createdBy===actor`); `canEditWhileLocked` is the same minus the admin layer; `mapRecordLockState` normalizes DB columns → `{locked, lockedBy, lockedAt}` wire shape.

## Frontend (reverse #2278)
- Rule editor: `UNSUPPORTED_SELECTABLE_ACTION_TYPES` now empty → `lock_record` selectable again.
- Grid (`MetaGridTable.vue`): lock indicator (🔒) on locked rows, lock/unlock action shown per `canUnlock`/edit, locked rows made non-editable (`isEditable` returns false). Emits `toggle-lock`.
- Drawer (`MetaRecordDrawer.vue`): lock status banner + locker/time + unlock button (per `canUnlock`).
- `MultitableWorkbench.vue` wires both surfaces to `client.setRecordLock` + reload.
- i18n via existing label modules (`meta-core-labels`, `meta-record-labels`, `workbench-labels`) — no new module.

## Tests — LR-T1..T10, fail-first, RED→GREEN
RED proven first (7 backend failed / 1 passed: missing columns + `executeLockRecord` crash), then GREEN:
- **Backend (real DB, `describeIfDatabase`)** `multitable-record-lock.test.ts`: T1 edit-reject (403), T2 delete-reject (403), T3 comment-still-allowed, T4 canUnlock three layers, T5 automation lock→unlock writes/clears columns, T6 admin not silently bypassed, T7 pre-existing rows = locked false, T10 wire round-trip. **8/8 green.**
- **Migration unit** `meta-record-locked-migration.test.ts` (T8): column shapes + index + clean down(). **3/3 green.**
- **Frontend (jsdom)** `meta-grid-table-record-lock.spec.ts` (T9): indicator + action visibility follows canUnlock + locked-row read-only. **6/6 green.**

## Neighbor suites
§2a.3 mask suites (write-echo / create-echo / cross-sheet / lookup-foreign-mask ×7 / read-field-mask / search-filter-mask), FOL-1 invalidation, records-list-authz — **all green**. `record-service` + `record-write-service` + `automation-v1` unit (249), `multitable-record-patch.api`, `multitable-sheet-realtime.api` — green after updating the mock-string SELECT matchers for the added columns. `tsc --noEmit` (backend) + `vue-tsc -b` (web) clean. OpenAPI `MultitableRecord` schema gains `locked`/`lockedBy`/`lockedAt`/`canUnlock`; `verify:multitable-openapi:parity` green.

## CI wiring
Added `tests/integration/multitable-record-lock.test.ts` to the **enumerated DB-gated list** in `.github/workflows/plugin-tests.yml` (not a glob), so the locking canaries run on the real-DB lane.

## Residual for the reviewer
- `multitable-record-form.api.test.ts` has **8 pre-existing failures on the base commit** (`pg_advisory_xact_lock` / `SELECT id, sheet_id FROM meta_records` mock gaps from prior PRs) — unrelated to this change. My delta touched only the 2 `/view` tests in that file (mock SELECT + the `toEqual` rows expectation now include the top-level lock metadata); verified the failing-test set is **identical to base** (no new regressions).
- `MultitableWorkbench.vue` has a **pre-existing** ESLint error (`'apiFetch' is defined but never used`) on base — left untouched to avoid scope creep.
- The new `POST /records/:recordId/lock` route is not yet in the OpenAPI path list (the response *schema* is covered via `MultitableRecord`); flag if you want the path documented too.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -190,6 +190,7 @@ jobs:
             tests/integration/multitable-dashboard-preview-data.test.ts \
             tests/integration/multitable-write-echo-field-mask.test.ts \
             tests/integration/multitable-create-echo-field-mask.test.ts \
+            tests/integration/multitable-record-lock.test.ts \
             tests/integration/multitable-cross-sheet-related-echo-mask.test.ts \
             tests/integration/multitable-lookup-foreign-field-mask-view.test.ts \
             tests/integration/multitable-lookup-foreign-field-mask-export.test.ts \
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -191,6 +191,7 @@ jobs:
             tests/integration/multitable-write-echo-field-mask.test.ts \
             tests/integration/multitable-create-echo-field-mask.test.ts \
             tests/integration/multitable-record-lock.test.ts \
+            tests/integration/multitable-record-lock-bypass.test.ts \
             tests/integration/multitable-cross-sheet-related-echo-mask.test.ts \
             tests/integration/multitable-lookup-foreign-field-mask-view.test.ts \
             tests/integration/multitable-lookup-foreign-field-mask-export.test.ts \
--- a/apps/web/src/multitable/api/client.ts
+++ b/apps/web/src/multitable/api/client.ts
@@ -1321,6 +1321,20 @@ export class MultitableApiClient {
     return this.parseJson(res)
   }
 
+  // Record locking (design #2278 follow-up): { locked: true } locks, { locked: false } unlocks.
+  async setRecordLock(
+    recordId: string,
+    locked: boolean,
+    params?: { sheetId?: string; viewId?: string },
+  ): Promise<{ recordId: string; locked: boolean; lockedBy: string | null; lockedAt: string | null }> {
+    const res = await this.fetch(`/api/multitable/records/${recordId}/lock`, {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify({ locked, ...params }),
+    })
+    return this.parseJson(res)
+  }
+
   async patchRecords(input: PatchRecordsInput): Promise<PatchResult> {
     const res = await this.fetch('/api/multitable/patch', {
       method: 'POST',
--- a/apps/web/src/multitable/components/MetaAutomationRuleEditor.vue
+++ b/apps/web/src/multitable/components/MetaAutomationRuleEditor.vue
@@ -1264,6 +1264,10 @@ const SUPPORTED_SELECTABLE_ACTION_TYPES: AutomationActionType[] = [
   'wait_for_callback',
   'condition_branch',
   'parallel_branch',
+  // lock_record is now a complete contract (record-locking storage + write-path enforcement, rank 8),
+  // so it is a first-class selectable action — both for new rules and to keep existing lock_record rules
+  // editable (it must appear in SUPPORTED so an existing rule's option isn't dropped to empty).
+  'lock_record',
 ]
 
 // A6-3-2a: BRANCH_AUTHORABLE_ACTION_TYPES (the v1 in-branch action subset) is imported from
--- a/apps/web/src/multitable/components/MetaAutomationRuleEditor.vue
+++ b/apps/web/src/multitable/components/MetaAutomationRuleEditor.vue
@@ -1269,7 +1269,9 @@ const SUPPORTED_SELECTABLE_ACTION_TYPES: AutomationActionType[] = [
 // A6-3-2a: BRANCH_AUTHORABLE_ACTION_TYPES (the v1 in-branch action subset) is imported from
 // ../utils/conditionBranchAuthoring — the single source of truth shared with the seam + tests.
 
-const UNSUPPORTED_SELECTABLE_ACTION_TYPES: AutomationActionType[] = ['lock_record']
+// lock_record is now a complete contract (record-locking storage + write-path enforcement), so it is
+// re-exposed in the rule-editor dropdown (reversing the #2278 stop-gap that hid the then-broken action).
+const UNSUPPORTED_SELECTABLE_ACTION_TYPES: AutomationActionType[] = []
 
 function isUnsupportedSelectableActionType(type: AutomationActionType): boolean {
   return UNSUPPORTED_SELECTABLE_ACTION_TYPES.includes(type)
--- a/apps/web/src/multitable/components/MetaGridTable.vue
+++ b/apps/web/src/multitable/components/MetaGridTable.vue
@@ -157,6 +157,22 @@
               <td class="meta-grid__row-num" :style="{ left: `${rowNumLeft}px` }">
                 <button class="meta-grid__expand-btn" :class="{ 'meta-grid__expand-btn--open': expandedRowIds.has(row.id) }" :aria-label="expandedRowIds.has(row.id) ? l('grid.collapseRow') : l('grid.expandRow')" @click.stop="toggleRowExpand(row.id)">&#x25B6;</button>
                 <span>{{ startIndex + ri + 1 }}</span>
+                <span
+                  v-if="isRowLocked(row.id)"
+                  class="meta-grid__lock-indicator"
+                  :title="l('grid.lockedIndicator')"
+                  :aria-label="l('grid.lockedIndicator')"
+                  data-test="row-lock-indicator"
+                >&#x1F512;</span>
+                <button
+                  v-if="canShowLockAction(row.id)"
+                  type="button"
+                  class="meta-grid__lock-action"
+                  :aria-label="isRowLocked(row.id) ? l('grid.unlockRow') : l('grid.lockRow')"
+                  :title="isRowLocked(row.id) ? l('grid.unlockRow') : l('grid.lockRow')"
+                  data-test="row-lock-action"
+                  @click.stop="emit('toggle-lock', { recordId: row.id, locked: !isRowLocked(row.id) })"
+                >{{ isRowLocked(row.id) ? '🔓' : '🔒' }}</button>
                 <button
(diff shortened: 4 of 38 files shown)


zensgit:
## Rank-8 review fixes — record lock now holds across EVERY mutation path (centralized + durable guard)

Applied the §2a.3 lesson: rather than patch sinks one at a time, centralized the lock rule into **ONE helper** and added a **durable structural guard** so no future mutation path can bypass silently.

### Centralized rule
`ensureRecordNotLocked(actorId, row, makeError)` in `record-lock.ts` — throws when `row.locked && !canEditWhileLocked(actorId, …)`. Each surface injects its own error factory so it keeps its native error type / HTTP shape. The **named** single-PATCH, bulk-PATCH and single-DELETE guards were refactored onto it too — there is now exactly one rule.

### Blockers fixed
- **B1 — automation `update_record`** (`automation-executor.ts`): reads `locked/locked_by/created_by`, routes through the helper with `context.actorId`; a non-locker/owner automation is blocked and the step returns `failed: Record is locked` (surfaces in the execution log). Decision f: the rule must `lock_record{locked:false}` first to write through a lock.
- **B2 — form-submit EDIT** (`POST /views/:viewId/submit`): added `locked, locked_by` to the SELECT and the guard before the UPDATE → `403`.
- **B3 — `DELETE /attachments/:attachmentId`** on a record: loads the record's lock columns and guards before the data-mutating UPDATE → `403`; the attachment row is not soft-deleted either.
- **M1 — plugin-SDK `patchRecord`/`deleteRecord`** (`records.ts`): **GUARDED**. The plugin path is **actor-less** (no per-record actor in the input), so `ensureRecordNotLocked(null, …)` makes a locked record **hard read-only** to plugins (`canEditWhileLocked(null,…)` is always false) — the lock can only be lifted via the explicit unlock action.

### Complete mutation-path enumeration (every `UPDATE`/`DELETE meta_records` in audited runtime src)
| # | site | path | disposition |
|---|---|---|---|
| 1 | `record-service.ts` (single PATCH) | user edit | **GUARDED** |
| 2 | `record-service.ts` (single DELETE) | user delete | **GUARDED** |
| 3 | `record-write-service.ts` (bulk/multi-record PATCH) | user edit | **GUARDED** |
| 4 | `automation-executor.ts` `executeUpdateRecord` | **B1** | **GUARDED** |
| 5 | `automation-executor.ts` `executeLockRecord` LOCK | lock action | **LOCK-MGMT** |
| 6 | `automation-executor.ts` `executeLockRecord` UNLOCK | unlock action | **LOCK-MGMT** |
| 7 | `routes/univer-meta.ts` form-submit EDIT | **B2** | **GUARDED** |
| 8 | `routes/univer-meta.ts` attachment-delete record edit | **B3** | **GUARDED** |
| 9 | `routes/univer-meta.ts` lock action | lock action (own canEditRecord authority) | **LOCK-MGMT** |
| 10 | `routes/univer-meta.ts` unlock action | unlock action (own canUnlock authority) | **LOCK-MGMT** |
| 11 | `routes/univer-meta.ts` people-directory sync | internal system sheet | **SYSTEM-EXEMPT** |
| 12 | `routes/univer-meta.ts` field-delete (`data - $1`) | schema op, sheet-wide key drop | **SYSTEM-EXEMPT** |
| 13 | `records.ts` plugin patchRecord | **M1** | **GUARDED** |
| 14 | `records.ts` plugin deleteRecord | **M1** | **GUARDED** |
| 15 | `formula-engine.ts` formula materialize | derived value, no user actor | **SYSTEM-EXEMPT** |
| 16 | `auto-number-service.ts` backfill | system-derived value | **SYSTEM-EXEMPT** |

Sweep also checked: import-xlsx (INSERT-only, no record UPDATE), link/relation writes (live in `meta_links`, not a record-data edit), bulk/cascade delete (no bulk-delete route exists). **No 5th bypass found** — the four named are the complete set.

### Durable guard — `tests/unit/multitable-record-lock-guard.guard.test.ts`
Enumerates every `UPDATE/DELETE meta_records` statement across the 7 audited runtime files and **fails** unless each carries a disposition marker on the line above the SQL (`// lock-guarded:` / `// lock-exempt:` / `// lock-mgmt:`). A new or moved mutation site with no marker trips RED by construction, forcing the author to route through `ensureRecordNotLocked` or document the exemption. Cross-checked: every GUARDED-labelled file really calls `ensureRecordNotLocked(`; the helper is defined once. **Verified the guard goes RED when a marker is removed**, then restored.

### RED → GREEN evidence (real DB `metasheet_test`)
New canaries `tests/integration/multitable-record-lock-bypass.test.ts` (seed record locked by OWNER → mutate as a plain `multitable:write` STRANGER → assert reject + row unchanged):

| canary | HEAD (RED) | after fix (GREEN) |
|---|---|---|
| LRX-B1 automation update_record | step `success`, value overwritten | step `failed`/`Record is locked`, unchanged |
| LRX-B2 form-submit EDIT | `200`, overwritten | `403`, unchanged |
| LRX-B3 attachment-delete | `200`, attachment removed | `403`, attachment + row intact |
| LRX-M1a plugin patchRecord | resolved, overwritten | throws `Record is locked`, unchanged |
| LRX-M1b plugin deleteRecord | resolved, deleted | throws `Record is locked`, still present |

All 5 RED on HEAD, all 5 GREEN after. Canary file wired into `plugin-tests.yml` DB-gated list.

### Suites green
- LR-T1..T10 (real DB): **8/8** ✓
- §2a.3 mask suites: **124 / 0** ✓ (unchanged)
- record-service + record-write-service + automation-v1 + taint-chokepoint guard + new lock guard units: **257/257** ✓
- frontend `meta-grid-table-record-lock.spec` **6/6** ✓
- `tsc --noEmit` ✓ · `vue-tsc -b` ✓ · `verify:multitable-openapi:parity` ✓ · zero new ESLint errors/warnings on changed files

### Residual (pre-existing, NOT introduced here)
`multitable-records.test.ts` / `multitable-query-service.test.ts` have **8 pre-existing failures** (the base PR's `patchRecord`/`getRecord` now return `locked/lockedBy/lockedAt` but those unit *expectations* weren't updated) — reproduced identically with this commit stashed. Same fixture-drift class as the `multitable-record-form.api` 8 the review documented. Left for a separate fixture-cleanup PR to avoid scope creep; provably not worsened (sheet-realtime + record-form form-submit SELECT mocks were updated to the new column list to keep them honest). One design sentence ratifying system-recompute as lock-exempt is now encoded structurally in the durable guard's SYSTEM-EXEMPT markers.

PR #2549, opened 2026-06-12
feat(multitable): mask foreign-field-level over-read in lookup/rollup + formula export (§2a.3)

zensgit:
## Security fix — §2a.3 "lookup foreign-field-level over-read"

A lookup/rollup column leaked a foreign field the actor is **explicitly denied at field level**. `applyLookupRollup` gated only whether the foreign **sheet** was readable (`resolveReadableSheetIds`), then read `data[targetFieldId]` raw — no foreign-**field**-level check. Two sinks.

### Sink A — read/JSON (primary)
`applyLookupRollup` now, for each distinct foreign sheet, resolves the actor's readable-foreign-field-id set (one scope-map load per foreign sheet — batched, never per record) + a cross-base flag (source `base_id` vs foreign `base_id`). The decision is made **once per `(foreignSheetId, targetFieldId)`**. In `resolveLookupValues` (shared by lookup AND rollup), if the foreign target field is masked, the value never reaches `data[fieldId]` (returns `[]`) so it's absent from every read/JSON consumer (`/view` grid, single-record GET, dashboards, related-record echo, form-submit recompute).

### Sink B — export-xlsx (secondary, fail-closed)
Formula values **are** materialized into `meta_records.data`, and export reads them raw (no `applyLookupRollup`). After the exportable `fieldIds` set is built, each exportable **formula** field is taint-checked against `formula_dependencies`: if it (transitively, via same-sheet formula→formula edges) depends on a lookup/rollup whose foreign target field is masked for the exporting actor, the **whole formula column** is dropped — consistent with the existing field-level export masking.

## Ratified semantics (design §2a.3)
- **cross-base** (foreign `base_id` ≠ source `base_id`): MASK **unconditionally** (opt-out ignored).
- **same-base** (equal, incl. both `null`): MASK **by default**, UNLESS the lookup/rollup config carries `skipForeignFieldMasking: true` (deliberate projection).
- actor CAN read the foreign field: value flows normally (no change).

### Cross-base detection with null base_ids
`crossBase = foreignBaseId !== sourceBaseId` after normalizing both to `string | null`. `null !== 'x'` ⇒ cross-base (mask). `null === null` and `'a' === 'a'` ⇒ same-base. So a sheet with a real base and a legacy null-base sheet are correctly treated as cross-base.

## Fail-first RED → GREEN evidence
Both real-DB (`describeIfDatabase`).

**GA-T2a** `multitable-lookup-foreign-field-mask-view.test.ts` (read/JSON via `/view`):
[...]

PR #2555, opened 2026-06-12
fix(multitable): harden rollup masking for §2a.3

zensgit:
Follow-up to #2549 review. This branch contains the rollup-count hardening commit on top of the current #2549 branch; if GitHub still treats #2549 as the PR for this head, close this attempt and keep #2549 as canonical.


zensgit:
Closing this accidental replacement PR. GitHub exposed that #2549's branch/PR ref was out of sync, and the branch carried the larger #2549 stack. I split the actual review follow-up into a clean branch based on current main so it contains only the rollup-mask hardening.

PR #2556, opened 2026-06-12
fix(multitable): mask denied rollup counts as null

zensgit:
## Summary
- distinguish permission-masked lookup/rollup reads from genuine empty values inside applyLookupRollup
- keep lookup masking as an empty array, but return null for rollups whose foreign sheet/field is masked so count rollups do not emit a misleading concrete 0
- add a real-DB /view regression covering denied count rollup -> null and readable count rollup -> 1

## Verification
- git diff --check
- pnpm --filter @metasheet/core-backend exec vitest run tests/integration/multitable-lookup-foreign-field-mask-view.test.ts --watch=false (skipped locally because DATABASE_URL is not set; CI should exercise the real DB path)

Follow-up to #2549 review P2; #2549 current head already covered the formula taint/read-sink P1 on current main.

### DIFF
--- a/packages/core-backend/src/routes/univer-meta.ts
+++ b/packages/core-backend/src/routes/univer-meta.ts
@@ -2281,6 +2281,8 @@ async function applyLookupRollup(
   ): { values: unknown[]; masked: boolean } => {
     const foreignSheetId = cfg.foreignSheetId ?? linkConfigById.get(cfg.linkFieldId)?.foreignSheetId
     if (!foreignSheetId) return { values: [], masked: false }
+    const linkIds = getLinkIds(record, cfg.linkFieldId)
+    if (linkIds.length === 0) return { values: [], masked: false }
     if (!readableForeignSheetIds.has(foreignSheetId)) return { values: [], masked: true }
     // §2a.3: fail-closed foreign-FIELD mask — if the actor can't read the foreign target field,
     // mask it (cross-base unconditional / same-base default unless opt-out). Same gate for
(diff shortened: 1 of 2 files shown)


zensgit:
Review patch pushed after James caught a real P1 in the empty-link case.\n\nIssue: the new masked-vs-empty distinction still checked readableForeignSheetIds before checking the current row's link ids. If a response/page/single-record read contained only rows with no links, the foreign sheet never entered foreignIdsBySheet, so a readable count rollup could still become null instead of 0. The existing test missed it because the empty row was read in the same /view response as a linked row, which primed readability.\n\nFix: resolveLookupValues now returns {values: [], masked: false} immediately when the current row has no link ids. The empty-count regression now uses GET /records/:id for REC_M_EMPTY so the batch contains only the empty-link row.\n\nLocal verification after rebase onto current origin/main:\n- pnpm --filter @metasheet/core-backend exec vitest --config vitest.integration.config.ts run tests/integration/multitable-lookup-foreign-field-mask-view.test.ts --watch=false --reporter=dot (loads; skipped locally because no DATABASE_URL)\n- pnpm --filter @metasheet/core-backend build\n- git diff --check origin/main...HEAD\n\nFresh CI is running now.

zensgit:
Rebased again after #2571 landed on main; new head is aebf4ed9074d63a304061078c7c7ea086c7efb93.\n\nLocal verification after rebase:\n- pnpm --filter @metasheet/core-backend exec vitest --config vitest.integration.config.ts run tests/integration/multitable-lookup-foreign-field-mask-view.test.ts --watch=false --reporter=dot -> spec loads; skipped locally because no DATABASE_URL\n- pnpm --filter @metasheet/core-backend build -> pass\n- git diff --check origin/main...HEAD -> pass\n\nFresh CI restarted.

zensgit:
Rebased onto current `origin/main` and force-pushed `71336c816`.

Fresh CI is settled green: 15/15 pass, 0 fail; merge state is `CLEAN`. Local checks before push:

- `git diff --check origin/main...HEAD`
- `pnpm --filter @metasheet/core-backend exec vitest --config vitest.integration.config.ts run tests/integration/multitable-lookup-foreign-field-mask-view.test.ts --reporter=dot` (skipped locally because `DATABASE_URL` is not set; CI provides the real signal)
- An accidental non-`run` Vitest invocation entered watch mode and completed the broader backend suite green before being cleaned up.

Parked for review/merge approval; not auto-merging.

PR #2569, opened 2026-06-13
feat(attendance): finalize schedule dispatch approvals

zensgit:
## Summary
- Implements D3 final approval writer for schedule_dispatch: final approve now materializes one published attendance_shift_assignments row with schedule_dispatch provenance and an optional schedule-group membership window.
- Revalidates dispatch scope from the reloaded active target schedule group, including latest department_ref, before writing schedule facts.
- Reuses assignment conflict detection, shiftEditPolicy, and shiftCompliance guards; guard failures rollback before status flip and leave the request pending.
- Backfills the dispatch design-lock and DingTalk benchmark tracker as D3 code-green, with D4 UI and D5 staging still pending before dispatch can flip ✅.

## Verification
- node --check plugins/plugin-attendance/index.cjs
- pnpm --filter @metasheet/core-backend build
- fresh DB: DATABASE_URL=postgresql://metasheet:metasheet@localhost:5435/metasheet_dispatch_d3_test pnpm --filter @metasheet/core-backend migrate
- fresh DB: ATTENDANCE_TEST_DATABASE_URL=postgresql://metasheet:metasheet@localhost:5435/metasheet_dispatch_d3_test pnpm --filter @metasheet/core-backend exec vitest --config vitest.integration.config.ts run tests/integration/attendance-schedule-dispatch.test.ts --reporter=dot (11/11)
- fresh DB: ATTENDANCE_TEST_DATABASE_URL=postgresql://metasheet:metasheet@localhost:5435/metasheet_dispatch_d3_test pnpm --filter @metasheet/core-backend run test:integration:attendance (8 files / 158 tests)

## Notes
- One non-fresh local run reproduced the known shiftCompliance settings-pollution hazard; rerunning on a fresh migrated DB passed the full attendance gate.
- This does not include D4 UI or D5 staging smoke.


zensgit:
Patch incoming from review: Harvey caught a real P1 in the D3 writer. Finalization used the generic conflict helper only, so with multi-shift enabled a schedule-dispatch approval could coexist with protected generated rows in another non-conflicting slot/window, despite the design lock forbidding dispatch over temporary / shift_swap / auto_shift_match rows.\n\nI patched the branch to add a dedicated finalization guard before materializing the assignment: any active published overlapping row with assignment_kind='temporary' or producer_type in {shift_swap, auto_shift_match} now fails closed with 409 SCHEDULE_DISPATCH_PROTECTED_ASSIGNMENT_CONFLICT, leaving the request pending and writing no assignment/membership. I also added the real-DB regression for the exact missed shape: multiShift enabled, protected auto_shift_match row in slot 1 with non-overlapping time, dispatch request in slot 0.\n\nI folded in one D5-smoke hardening while here: if the target schedule membership already exists, finalization now returns and records that existing membership id instead of leaving membership_id null. Test covers reuse: no new schedule_dispatch membership row, detail/metadata point at the reused membership.\n\nLocal verification in the temporary worktree: core-backend build PASS; dedicated attendance-dispatch spec loads under vitest.integration.config.ts and skips locally because this machine has no DATABASE_URL; git diff --check PASS. Fresh CI will provide the real-DB proof.

zensgit:
Follow-up note: #2569 merged before this review patch could be pushed. I moved the P1 hardening into a dedicated post-merge PR: #2570.\n\n#2570 carries the protected-generated-row finalization guard plus the membership-reuse auditability fix described above. It is intentionally narrow D3 hardening only; no D4/D5 scope.

PR #2576, opened 2026-06-13
feat(multitable): block base-delete orphaning cross-base links + sheet-create TOCTOU guard + dangling-link sweep (②a §2a.4)

zensgit:
Completes the ②a cross-base governance wall — §2a.4 (owner-ratified 2026-06-13 \"按推荐\", decision O3). Off origin/main 78f7d2acd (the merged ②a wall: §2a.2 `validateLinkFieldConfig` + §2a.1 `resolveBaseReadable`).

## Recon findings

**Is base-delete API-reachable? NO.**
- There is **no `DELETE /bases` (or PATCH/PUT) route** anywhere — `univer-meta.ts` exposes only `GET /bases` and `POST /bases`. No `deleteFrom('meta_bases')` / `DELETE FROM meta_bases` in any code path.
- `meta_bases.deleted_at` (soft-delete column) has **no writer** — soft-delete was designed but never built.
- FK introspection (test DB): the **only** FK referencing `meta_bases` is `meta_sheets.base_id … ON DELETE SET NULL`; `meta_bases` references **no parent table** (its `workspace_id` is a plain `text` column, no FK) — so there is **no workspace (or any) cascade that reaches `meta_bases`**.
- → The `meta_sheets.base_id` SET NULL cascade fires only on a hard `DELETE FROM meta_bases`, which no application path performs. **Neither hard nor soft base-delete is API-reachable, and no cascade reaches bases.**

**SET NULL mechanism:** migration `zzzz20260318110000` — `meta_sheets_base_id_fkey FOREIGN KEY (base_id) REFERENCES meta_bases(id) ON DELETE SET NULL`. `meta_links.foreign_record_id` is `text` with **no FK** (can dangle).

**Sheet-create caller-chosen-id path:** `POST /sheets` (`univer-meta.ts` ~6737) accepts a caller-chosen `id` AND `baseId` (both zod-optional), inserting `meta_sheets(id, base_id, …)` with `ON CONFLICT (id) DO NOTHING`. This is the second leg of the TOCTOU vector the Lane-1 review flagged.

## As-built

**2a.4-a — base-delete guard: MOOT (not shipped).** Base-delete is not API-reachable (above), so there is nothing to block at a route and no force/unlink path to build. Per the task's fallback, the cross-base **detection** is instead folded into the sweep (real caller) + pinned by a pure unit test — not shipped as a dead route-layer helper. **No force param / no unlink path.**

**2a.4-c — sheet-create TOCTOU close (the real shipped guard).** The §2a.2 wall (`validateLinkFieldConfig`) only compares bases when both sheets exist; a link created against a not-yet-existent foreign sheet id no-ops the wall. Closed from the other side: `POST /sheets` now rejects (**400 VALIDATION_ERROR**) when creating this sheet (with its resolved baseId) would retroactively make an **existing** link field cross-base — some live `type='link'` field's foreignSheetId equals the new sheet's id while that field's source sheet is in a different base. Symmetric to the wall:
- **alias-aware** — matches `foreignDatasheetId → foreignSheetId → datasheetId` in the same precedence as `parseLinkFieldConfig` (a foreignSheetId-only scan is bypassable, per the wall's own alias test).
- **null-aware compare** — reuses the new shared `baseIdsAreCrossBase(a,b) = a !== b` (also now used by the wall, so they can't drift; the sweep replicates it in SQL via `IS DISTINCT FROM`).
- **type coverage** — `WHERE mf.type = 'link'` matches the wall: `person` fields normalize to stored `type='link'` and are same-base by construction (`ensurePeopleSheetPreset(baseId)`); no path persists `type='person'` into `meta_fields`.
- distinct `CrossBaseLinkError` → maps to 400 (the route's `ValidationError` maps to 403; using it would have produced the wrong code). Runs inside the existing transaction, after baseId resolution (incl. `ensureLegacyBase`), before the INSERT.

**2a.4-b — dangling / cross-base link ops sweep (always).** `scripts/ops/multitable-dangling-link-sweep.mjs` (read-only psql, mirrors `check-yjs-retention-health.mjs`). Enumerates (i) cross-base link fields (`src.base_id IS DISTINCT FROM fgn.base_id`, alias precedence) and (ii) dangling `meta_links` (`LEFT JOIN meta_records … WHERE r.id IS NULL`). **VALUES-FREE** (field/sheet/base ids + counts only, never cell data), **read-only** (no writes). npm script `ops:multitable-dangling-link-sweep`; exits 0 clean / 2 hazards found / 1 usage. **Why it still matters with base-delete moot:** `DELETE /sheets/:id` IS reachable and `meta_records.sheet_id` is `ON DELETE CASCADE`, so sheet-delete → records vanish → inbound `meta_links` dangle → caught by category (ii). That is the real reachable orphaning vector.

## O3 decision (BLOCK + force)
O3 default = BLOCK (vs silent SET NULL) + read-only ops sweep. As-built: the BLOCK lands at the **reachable** chokepoint (sheet-create TOCTOU = 400) since base-delete itself is unreachable; the sweep is the ops half. No `?force=true` / unlink was built because there is no reachable base-delete to force past.

## Fail-first RED → GREEN (real DB where stateful)
- **unit** `multitable-cross-base-compare.test.ts`: RED 4 failed (`baseIdsAreCrossBase is not a function`) → GREEN 4/4.
- **2a.4-c** `multitable-cross-base-sheet-create-toctou.test.ts`: RED 3 failed (cross-base / legacy-base / alias cases returned 200 — the live hole) | 2 passed (sentinel + same-base compat) → GREEN 5/5.
- **2a.4-b** `multitable-dangling-link-sweep.test.ts`: RED 2 failed (script MODULE_NOT_FOUND) | 1 passed → GREEN 3/3 (enumerates the 2 seeded cross-base fields incl. alias + 1 dangling link, excludes same-base/non-link/good-link).

## Neighbors green
§2a.2 wall 9/9 · §2a.1 resolver 6/6 · §2a.3 mask suite 9 files / 49 · egress-coverage GOLDEN (13, unchanged — no egress helper touched) · cross-base-link-wall-guard + taint/lock/projection guards 18/18 · `tsc --noEmit` clean · OpenAPI parity (no spec change — `POST /sheets` already declares `400 ValidationError`, reused; no force param added).

## No migration
Route-level guard + ops script only. New integration tests wired into `.github/workflows/plugin-tests.yml` DB-gated list (after the §2a.2 wall entry). The DB-gated job uses `ankane/setup-postgres` (psql on PATH — the sweep test spawns it).

## Residuals (named, not fixed)
1. **No sheet base-reparent path exists** — there is no PATCH/PUT `/sheets` and no runtime `UPDATE meta_sheets` (the only one is the one-time migration backfill). So the symmetric reparent bypass (same-base create → re-parent to base B) is not reachable today; if such a route is ever added it must re-fire the wall.
2. **Provisioning sheet-create (`multitable/provisioning.ts` `createSheet`/`ensureSheet`) is internal-only** — the route layer imports only `ensureLegacyBase` from provisioning, not the sheet-create fns; it is the PLM/integration programmatic trust domain, classified OUT-OF-SCOPE here exactly as the existing wall-guard classifies provisioning field-writes. Gated separately if it ever accepts cross-base targets.
3. **Link fields targeting a deleted/missing foreign sheet** are caught by neither sweep category (cross-base requires the foreign sheet to exist for a base compare; dangling is about records) — a separate residual, noted in the script header.


zensgit (commented):
Findings

[P1] The sheet-create guard still races with concurrent link-field creation
`packages/core-backend/src/routes/univer-meta.ts:1068`, `packages/core-backend/src/routes/univer-meta.ts:1097`, `packages/core-backend/src/routes/univer-meta.ts:5554`, `packages/core-backend/src/routes/univer-meta.ts:6799`

This closes the sequential deferred-binding case, but not the actual concurrent TOCTOU window. `POST /sheets` scans existing link fields before inserting the new sheet, while `POST /fields` validates a link by loading the foreign sheet and explicitly returns null when that sheet does not exist yet. There is no shared lock or serialization between those two transactions.

A bypass remains under the default read-committed shape:

1. Tx A starts `POST /sheets` for caller-chosen id `future` in base B and runs `validateSheetCreateNoRetroactiveCrossBaseLink`; no existing link field is visible yet.
2. Tx B starts `POST /fields` on a sheet in base A with `foreignSheetId: future`; `validateLinkFieldConfig` loads no foreign sheet and returns null.
3. Tx B inserts/commits the link field.
4. Tx A inserts/commits the sheet.

The final state is still a retroactive cross-base link. The new integration test seeds the deferred link before sheet creation, so it proves the sequential guard, but it does not exercise the concurrent race that the TOCTOU wording promises to close.

Suggested fix: serialize “materialize sheet id” and “create/update link to sheet id” on the same transaction-scoped advisory lock keyed by the target sheet id. For example, acquire a lock like `pg_advisory_xact_lock(hashtext('multitable-link-target:' || targetSheetId))` before link field create/update validation when a link foreign key is in play, and acquire the same lock before the sheet-create retroactive scan/insert. Then either transaction ordering becomes safe: if the sheet commits first, the link validator sees it and blocks cross-base; if the link commits first, the sheet-create scan sees it and blocks cross-base.

The rest of the slice is directionally good, especially the alias-aware scan and read-only sweep, but I would not merge this as a TOCTOU close until the concurrent path is locked.

zensgit:
## BLOCKER fixed — centralized sheet-create TOCTOU guard (`createSeededSheet`)

Addresses the review BLOCKER: `GET /view?seed=true` was a second, **unguarded** `createSeededSheet` caller reproducing the exact silently-cross-base TOCTOU §2a.4-c closes.

### The second sink
`createSeededSheet` had two callers, only one guarded:
- `POST /sheets` (`univer-meta.ts:6852`) — route-guarded by `validateSheetCreateNoRetroactiveCrossBaseLink` (`:6799`).
- `GET /view?seed=true` (`:7397`) — NO guard. `resolveMetaSheetId` returns the caller-chosen `sheetId` verbatim (even for a not-yet-existent sheet) → `createSeededSheet` inserts it at `ensureLegacyBase()` = `base_legacy`. A plain `multitable:write` member whose source sheet sits in a non-legacy base + a link field with `foreignSheetId = FUTURE_ID` could materialize a cross-base link from the other side.

### The fix — CENTRALIZE (not per-caller patch)
Moved the guard **into the shared chokepoint `createSeededSheet` itself**, keyed to a GENUINELY-NEW insert and validated against the base the sheet is ACTUALLY created at:
- Inside `createSeededSheet.run`: `SELECT id FROM meta_sheets WHERE id = $1` → if the sheet doesn't exist yet (`isGenuinelyNew`), run `validateSheetCreateNoRetroactiveCrossBaseLink(query, sheetId, base_legacy)` before the INSERT; on hit throw `CrossBaseLinkError`.
- `GET /view` catch now maps `CrossBaseLinkError → 400 VALIDATION_ERROR` (mirrors POST /sheets `:6838`).

**No-false-positive (the reviewer's trap):** `POST /sheets` (seed or not) inserts the sheet at its caller-chosen base *before* reaching `createSeededSheet`, so the existence check sees the row → `isGenuinelyNew=false` → chokepoint no-ops → no re-reject against the wrong (legacy) base. The route-level guard at `:6799` is **kept** — it is NOT subsumed: non-seed POST /sheets never calls `createSeededSheet`, and it validates the *caller-chosen* base, which the chokepoint cannot (the chokepoint only ever inserts at legacy).

### RED → GREEN (real wire, `DATABASE_URL=…/metasheet_test`)
New fail-first canary `multitable-cross-base-seed-view-toctou.test.ts` (auth = plain NON-admin member, `multitable:read/write`, mirrors review PROBE2):
- **RED (PR HEAD):** exploit case `GET /view?seed=true&sheetId=FUTURE_CROSS` → **200**, `FUTURE_CROSS` created at `base_legacy` ≠ source base (`expected 200 to be 400`).
- **GREEN (after fix):** exploit → **400 VALIDATION_ERROR**, `FUTURE_CROSS` NOT created. No-false-positive case (clean future id, no inbound link) still → **200** + created at `base_legacy`.

### `meta_sheets` INSERT enumeration (independently re-grepped over `packages/core-backend/src`) — `createSeededSheet` is NOT the sole sink; full disposition:
| Site | Path | Disposition |
|---|---|---|
| `univer-meta.ts createSeededSheet` (~`:3901`) | seed sink for `POST /sheets` + `GET /view?seed=true` | **GUARDED** — centralized chokepoint (this fix), fires on genuinely-new insert |
| `univer-meta.ts POST /sheets` (~`:6829`) | caller-chosen id + caller-chosen base | **GUARDED** — route-level `validateSheetCreateNoRetroactiveCrossBaseLink` (`:6799`), validates the caller-chosen base; distinct INSERT, not via `createSeededSheet`, kept |
| `univer-meta.ts ensurePeopleSheetPreset` (~`:3350`) | system people sheet | **EXEMPT** — id is server-generated (`buildId('sheet')`), inserted into the SAME `baseId` it was passed → not caller-influenced, same-base by construction |
| `provisioning.ts createSheet` (`:269`) / `ensureSheet` (`:293`) | PLM/object provisioning | **EXEMPT (not route-reachable)** — `univer-meta` imports ONLY `ensureLegacyBase` from `provisioning` (`:70`), never the sheet-create fns; not reachable with a caller baseId |

No kysely `insertInto('meta_sheets')` exists. `createSeededSheet` has exactly the two callers above (re-grepped).

### Neighbors (all GREEN)
§2a.2 wall **9/9** · POST /sheets TOCTOU **5/5** · new seed-view canary **3/3** · dangling-link sweep **3/3** · §2a.1 base-readable resolver **6/6** · §2a.3 lookup-foreign-field-mask **37/37** (no mask/egress code touched) · GOLDEN guards (cross-base-compare 4 + wall-guard 3 + egress-coverage + taint-chokepoint 3 + record-lock 4) **15/15** · `tsc --noEmit` clean · OpenAPI parity GREEN (reuses existing `400 VALIDATION_ERROR`, no spec drift). Added the canary to the `plugin-tests.yml` CI allow-list.

### Residual (pre-existing, NOT introduced here, NOT CI-reachable)
`multitable-context.api.test.ts` (a hand-rolled **mock-SQL** harness, excluded from `vitest.config.ts` AND absent from the `plugin-tests.yml` allow-list) throws `Unhandled SQL` on the guard's `SELECT`. Bisected: pre-PR baseline = 2 failing (unrelated DB-state `sheet_ops`); **this PR's own POST /sheets guard** added 3; **this fix adds 0**. Out of scope for the BLOCKER; flagging so the mock allow-list can be extended separately if that file is ever re-included.

zensgit:
### Follow-up (`fa517ffd5`) — no-false-positive now backed by a wire test + GOLDEN count clarified

**1. POST /sheets + seed=true SAME-base no-false-positive — now a real wire test (was prose only).**
Added to `multitable-cross-base-sheet-create-toctou.test.ts`: source sheet + a same-base link in `BASE_A` pointing at `FUTURE_SEED_SAME`, then `POST /api/multitable/sheets { id: FUTURE_SEED_SAME, baseId: BASE_A, seed: true }` → asserts **200**, `seeded:true`, and the sheet created at **BASE_A** (not rejected, not relocated to legacy). This exercises the `isGenuinelyNew=false` skip branch (route inserts at the caller base before `createSeededSheet` runs in the same txn → existence check sees the row → guard skipped). File is now **6/6**.

**Teeth verified:** mutating the gate to always-run-against-`base_legacy` turns **only this new case RED** (`expected 400 to be 200` — the exact false-positive the review warned about); the other 5 stay green. Restored byte-identical; `tsc --noEmit` clean.

**2. "GOLDEN" count clarification.** The review named a `raw-projection` guard I hadn't located in my first pass — it is `tests/unit/multitable-raw-record-data-projection.guard.test.ts` (**3/3 GREEN**). With it, the GOLDEN structural/egress guard set = egress-coverage (1) + raw-record-data-projection (3) + stored-data-taint-chokepoint (3) + record-lock-guard (4) + cross-base-link-wall-guard (3) = **14/14**; plus the cross-base-compare unit (4/4). No egress/projection code is touched by this fix, so regression risk there remains nil.

Full §2a.4 cluster (wall 9/9 · POST /sheets TOCTOU **6/6** · seed-view canary 3/3 · dangling-link sweep 3/3) all GREEN. Ready for re-review; not merging.

zensgit:
Follow-up note: this PR merged before I could push the remaining materialization-lock P1 fix. I did not mutate the merged branch; I opened the fix as #2578 instead.

#2578 serializes link-field writes and target-sheet creation on the same `multitable:link-target:<sheetId>` advisory lock, with real route tests that hold the session lock and assert the request waits while locked.

PR #2578, opened 2026-06-13
fix(multitable): serialize link target materialization

zensgit:
## Summary

Follow-up to #2576. Closes the remaining P1 materialization race by making link-field writes and target-sheet creation contend on the same transaction-scoped advisory lock:

- `POST /fields` link create/update locks `multitable:link-target:<foreignSheetId>` before validation/insert/update.
- `POST /sheets` and the seeded-sheet centralized path lock `multitable:link-target:<sheetId>` before retroactive cross-base validation/materialization.
- Adds real HTTP route tests that hold the session-level advisory lock and assert the request waits while locked, proving the wire path uses the lock rather than a fixture-only helper.

## Verification

- `pnpm --filter @metasheet/core-backend run build`
- `pnpm --filter @metasheet/core-backend exec vitest run tests/integration/multitable-cross-base-sheet-create-toctou.test.ts --watch=false` *(local machine has no DATABASE_URL, so the DB integration file is collected and skipped; CI should execute it with the DB matrix)*

## Boundaries

- Multitable-only.
- No permission/RBAC/auth changes.
- No frontend changes.
- No broad schema or sweep changes beyond serializing the already-locked #2576 chokepoints.

### DIFF
--- a/packages/core-backend/src/routes/univer-meta.ts
+++ b/packages/core-backend/src/routes/univer-meta.ts
@@ -1030,6 +1030,19 @@ export function baseIdsAreCrossBase(a: string | null, b: string | null): boolean
   return a !== b
 }
 
+function linkTargetMaterializationLockKey(sheetId: string): string {
+  return `multitable:link-target:${sheetId}`
+}
+
+async function acquireLinkTargetMaterializationLock(query: QueryFn, sheetId: string): Promise<void> {
+  await query('SELECT pg_advisory_xact_lock(hashtext($1))', [linkTargetMaterializationLockKey(sheetId)])
+}
+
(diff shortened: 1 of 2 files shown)


zensgit:
CI settled green on the follow-up P1 fix: all checks pass with Strict E2E skipped as expected.

What is locked here:
- link field create/update and target sheet create contend on the same `multitable:link-target:<sheetId>` advisory lock;
- both `POST /sheets` and seeded-sheet creation lock before retroactive cross-base validation/materialization;
- real route tests hold the session-level advisory lock and assert the request waits while locked, so this is a wire lock rather than a helper-only fixture.

Local verification before push:
- `pnpm --filter @metasheet/core-backend run build` ✅
- targeted DB integration file collected locally but skipped because this machine has no `DATABASE_URL`; CI executed the matrix and passed. ✅

PR #2595, opened 2026-06-15
feat(multitable): bidirectional/mirror links MVP (derived reverse)

zensgit:
## Bidirectional / mirror links MVP — derived reverse (single edge)

Implements the merged design-lock `docs/development/multitable-bidirectional-links-design-20260614.md`.

### The spine: one edge, two projections (no materialized mirror row, no migration)
A `meta_links` row `(field_id=F_A, record_id=A, foreign_record_id=B)` is the **single canonical edge**. Both fields are projections of that one row:

| Field | Query against the existing edge |
|---|---|
| Forward — A's `F_A` | `WHERE field_id=F_A AND record_id=A` → B (today's read) |
| **Mirror — B's `F_B`** | `WHERE field_id=F_A AND foreign_record_id=B` → A (new read, served by the existing `idx_meta_links_foreign`) |

There is **no write-back sync** (the forward write path is unchanged) and **no second stored row**, so loops are structurally impossible and the existing delete cascade (`record-service.ts` `... WHERE record_id=$1 OR foreign_record_id=$1`) already cleans the reverse.

### What's in the MVP (5 commits)
1. **Config** — `twoWay` / `mirrorFieldId` / `mirrorOf` added to the `link` field type, promoted explicitly through **both** codecs (`univer-meta.ts` `sanitizeFieldProperty` + `multitable/field-codecs.ts`) and OpenAPI (`base.yml` + regenerated dist). `mirrorOf` (the derived-side marker) additionally promotes `readOnly:true`. Wire-vs-fixture round-trip asserted (C4).
2. **Derived reverse resolution** — centralized in `loadLinkValuesByRecord` (`univer-meta.ts:~1841`): link fields are partitioned by side; a derived field (`cfg.mirrorOf` set) is resolved by the reverse query and its source `record_id`s are surfaced under the mirror field id, then flow through the **existing `buildLinkSummaries`** with source/foreign roles swapped (so masking is inherited, not re-implemented). One change covers `/view`, single-record read, and write-echo.
3. **Forward-write invalidation fan-out** — in `record-write-service.ts` the link-write loop collects the affected `foreign_record_id`s (`toInsert ∪ toDelete`) for `twoWay` fields, keyed by the mirror-side sheet + mirror field id, and folds them into the established **FOL-1 `affectedRelatedBySheet`** map after commit → the mirror records ride the existing post-commit Yjs invalidation + realtime broadcast. Read-invalidation only (no values, no `recordPatches`, no forward write back).
4. **Spine guards** — `isFieldAlwaysReadOnly` returns true when `property.mirrorOf` is set, so **both** write services reject a PATCH on the mirror regardless of property-load path (keeps the single edge from gaining a second row). `validateLinkFieldConfig` rejects a `twoWay` link whose two sides are cross-base (400), even with a `foreignBaseId` opt-in.
5. **Real-DB test matrix** (`multitable-bidirectional-mirror-links.test.ts`, `describeIfDatabase`, wired into `plugin-tests.yml`).

### Fail-first (RED→GREEN, real DB)
Design §8 matrix — all 10 GREEN; with the reverse-read / read-only guard / codec injection disabled, **6 canaries fail** (incl. C1 reverse + C9 read-only spine), restored = 10/10:
- C1 reverse resolves · C2 add-then-remove (no stale) · C3 forward-write fans invalidation to the mirror sheet (no write back) · C4 wire round-trip · C5 cross-base pairing rejected (400) · C6 delete clears the edge · C7 masked mirror (denied actor sees no leak) · C8 non-twoWay → no mirror (regression) · C9 mirror read-only.

### Neighbors green
FOL-1 invalidation, cross-base wall + opt-in, link-summary display mask, cross-sheet related echo mask, dangling-link sweep (41), plus the §2a.3 lookup-foreign-field mask family + read/write/create echo mask (34), plus record-write-service unit (55). `tsc --noEmit` clean; OpenAPI parity test passes.

### Deferred (each its own named opt-in, per design §7/§9)
- Editing from the mirror side (MVP mirror is read-only)
- Auto-create-on-toggle UX for the mirror field
- Cross-base bidirectional (this MVP is same-base only)
- Full symmetric pairing-integrity validation + `F_A`-delete → `F_B.mirrorFieldId` unset (no canary depends on the back-reference: reverse read keys on `mirrorOf`; cross-base reverse summaries are masked by the swapped Sink B-1 gate)
- Rollup / lookup over a mirror field

> Note: branch is off `741810a15`; `main` has advanced since — the PR diff is computed against the merge-base.


zensgit:
## Closed the B3 review HEADLINE nit (fail-closed): mask raw `data[mirrorField]` for foreign-denied actors

The review's headline nit was that the mirror's raw `record.data[mirrorField]` array (a SECOND projection of the reverse edge, separate from `linkSummaries`) leaked the inbound source-record ids + count to an actor who cannot read the mirror's foreign sheet — e.g. `[A1,A2]` returned to a sheet-A-denied actor reading a B record. `buildLinkSummaries` gated the summary; the raw array did not.

### Fix
New `maskDerivedMirrorFieldIds` (`univer-meta.ts` ~:1935) reuses the SAME `resolveReadableSheetIds` gate keyed on `cfg.foreignSheetId` that `buildLinkSummaries` uses, and blanks the raw value to `[]` when the actor can't read that sheet. Wired on **every read path the raw value reaches** (right after the populate loop, before `filterRecordDataByFieldIds`):

| Path | Site |
|---|---|
| GET `/view` | ~:7936 |
| Write-echo (public form) | ~:8624 |
| Write-echo (patch) | ~:8844 |
| Single-record GET `/records/:id` | ~:9117 |

**Scoped to mirror (`cfg.mirrorOf`) fields only.** Forward link fields are deliberately left byte-identical — their raw-id posture is pre-existing and shared repo-wide (an outbound edge the record authored); this fix only addresses the *inbound* edges the mirror newly makes reachable. Same-base by construction (cross-base twoWay/mirror is rejected at field-create per C5), so sheet-level read is the exact gate. No new schema/field → OpenAPI parity unaffected.

### Fail-first canaries (real DB)
`C7` only inspected `linkSummaries`; added **C7-RAW-DATA** asserting the raw `data[mirrorField]` — split per endpoint so each wire witnesses the leak independently:
- foreign-DENIED actor (`USER_NOA`) → `data[FLD_B_MIRROR]` is `[]` on `/view` **and** single-record GET
- PERMITTED actor (`USER_FULL`) → `[A1,A2]` on both (no over-masking)

Both proven **RED on the prior HEAD** (each leaked `['rec_a1','rec_a2']`) → **GREEN** after the mask.

### Regression (real DB, `DATABASE_URL`)
- Bidirectional/mirror matrix: **12/12** (C1–C9 + 2× C7-RAW-DATA + sentinel)
- Neighbors: FOL-1 invalidation, cross-base wall + opt-in, link-summary display mask S1–S4, cross-sheet related echo, dangling-link sweep, records-read mask, write-echo (R1–R5), create-echo (R1–R3): **58/58**
- §2a.3 lookup-foreign-field-mask family (view/create/write/export/sibling/aggregate/convergence): **40/40**
- `tsc --noEmit` (core-backend): **clean (exit 0)**

Commit `5227f8f48`. Does not merge.

PR #2598, opened 2026-06-15
feat(multitable): rich-text longText backend — write-path sanitizer (XSS-safe)

zensgit:
Backend half of rich-text longText (per design #2594). The authoritative XSS defense: a server-side sanitizer chokepoint inside `validateLongTextValue` (covers all record write sites — no route/API/automation bypass), `sanitize-html` with a strict allow-list, a `rich` flag on `longText` (default plain; toggle rejected on populated fields), and a unified plain-text projection for xlsx export + search. FE render/editor is a follow-on lane.


zensgit:
## REQUEST-CHANGES addressed — plugin-SDK XSS blocker + secondary toggle bypass closed (`0ce00698c`)

Both findings from the adversarial review are fixed; the sanitizer config itself held the 61-payload bypass battery unchanged.

### BLOCKER — plugin-SDK record write bypassed the sanitizer ✅
`records.ts` `normalizeFieldValue` had no `longText` case, so a `{rich:true}` value written via the plugin-SDK `records.createRecord`/`patchRecord` (wired to every plugin via `plugin-scope.ts`) was stored RAW.
- **Fix:** `records.ts:~233` — `case 'longText': return validateLongTextValue(value, field.id, field.property)` (+ import). Routes the plugin-SDK write through the SAME chokepoint as the HTTP / form-submit / automation paths.
- **Canary (richtext-longtext-sanitizer.test.ts PART 8):** drives the REAL `createRecord`/`patchRecord` and captures the persisted `data`. RED on HEAD (`<script>steal()</script>…` stored verbatim), GREEN after (`<a rel="noopener noreferrer" target="_blank">x</a>…<b>ok</b>` — script/onerror/javascript: all gone). Plain (non-rich) write asserted verbatim (regression).

### SECONDARY — plugin could flip `rich` on a populated field ✅
`provisioning.patchObjectFieldProperty` wrote `meta_fields.property` with no sanitization and no populated-field gate → a plugin could toggle `rich` ON an already-populated longText field, bypassing the HTTP toggle gate.
- **Fix:** `provisioning.ts:~369` — route the property write through `sanitizeFieldProperty` + the SAME backward-compat gate. The gate logic is factored into a shared `field-codecs` helper (`isRichLongTextTurningOn` + `assertRichLongTextToggleAllowed`) that **both** the HTTP field-PATCH route and provisioning now call.
- **Canary (PART 9):** flipping `rich` on a POPULATED field via the real `patchObjectFieldProperty` → REJECTED (matches the HTTP gate); EMPTY field → allowed; junk `rich:'true'` → dropped (strict `=== true`). RED→GREEN.

### Nit ✅
The `as any` at `univer-meta.ts:6042` is removed — subsumed by the shared-helper refactor (typed query result, `(rows?.length ?? 0) > 0`).

### Structural guard (prevents the 3rd multi-sink miss)
New `multitable-richtext-longtext-write-sink.guard.test.ts` enumerates **every** `meta_records.data` writer file (whole-`src`) and requires each to either reference the rich-longText sanitizer (CHOKEPOINT) or be documented SAFE (auto-number / formula-result / schema). Verified it goes RED when the `records.ts` fix is removed (the file stops referencing the sanitizer). Mirrors the §2a.3 chokepoint-guard precedent.

### Verification
- Canaries proven **RED on the unfixed paths → GREEN after** (4 security canaries fail with the fixes stashed).
- Neighbors green: richtext-longtext-sanitizer (32), write-sink guard (3), field-types-batch1, record-service, record-write-service, provisioning, plugin-scope, §2a.3 taint-chokepoint guard — **201 unit tests pass**; after-sales-provisioning + write-echo-field-mask integration green (real DB).
- `tsc --noEmit` clean.

Files: `records.ts`, `provisioning.ts`, `field-codecs.ts`, `routes/univer-meta.ts`, `richtext-longtext-sanitizer.test.ts`, `multitable-richtext-longtext-write-sink.guard.test.ts`. Not merging — re-review.

PR #2623, opened 2026-06-15
fix(multitable-ai): estimate-aware quota admission (close single-request overshoot)

zensgit:
## The gap (from a code-review finding, verified)

`checkAiUsageQuota` decided purely on `prior window SUM >= cap`, so a request at **990/1000 was admitted** (`990 >= 1000` is false), called the provider, and pushed the window **over** the cap — only the *next* request was blocked. This contradicted the documented **no-overshoot** guarantee (`ai-usage-ledger.ts` header :27-28 / reserveAiUsage docstring :263 / `multitable-ai.ts:298`). The in_flight RESERVE-THEN-SETTLE mechanism prevented **parallel** compounding overshoot, but not a **single** request crossing the cap from under it (admission counted everyone's in-flight estimate *except this request's own*).

## Fix

`checkAiUsageQuota` now takes this request's own conservative `estimate` and rejects on **`current + estimate > cap`**. `reserveAiUsage` passes the same conservative value it inserts into the in_flight row. Properties:
- An **exactly-fitting** request (`sum + estimate == cap`) is still allowed.
- A real request (estimate ≥ 1) still blocks at/over a full window → the "already full" behaviour is unchanged (no regression).
- In production a real daily cap ≫ one ~1024-token estimate, so strict admission only rejects a max-size request in the last sliver of budget (correct — it genuinely can't fit). It only over-blocks in a test where `cap < one estimate`, which is why the reserve happy-path test's cap was raised (1000 → 5000, above the 25+1024 estimate) with the exhausted sibling set **at** the cap.

## Tests
- Unit no-overshoot keystones: token **under→over** (990 + 20 > 1000 → blocked; **RED on pre-fix**), exactly-fits allowed, cost-cap under→over.
- Route-level **M4-T5b**: window 999/1000 → **429 `quota_exhausted` + ZERO provider fetch** (the end-to-end no-overshoot the finding asked for).
- 33/33 AI suites green; tsc clean. Docs updated to state the dual-axis guarantee truthfully.


zensgit:
Reviewed + fixed the CI-red (`test (20.x)`). Root cause was an **incomplete** change, not the production logic (which is sound): the estimate-aware admission (`sum + own conservative estimate vs cap`) broke the two real-DB integration suites whose cap fixtures encoded the prior SUM-only model. Unit + route-mock tests were updated in the original commit; `multitable-ai-shortcut-run.test.ts` (6) and `multitable-ai-suggest-formula.test.ts` (1) were missed — the repo's known unit-green / integration-red blind spot (these only run in the 20.x lane).

Follow-up commit `80fbc97fa`:
- Size token caps off the conservative estimate (`prompt.length + maxOutputTokens ≈ 1.1k`), not the actual: caps `1000 → 1500` so one estimate admits (window 0) but the second doesn't (window settles to the ~1000 actual, + estimate > 1500). Concurrency probe cap fits exactly one in_flight estimate.
- **Leak-proof `afterEach`**: the daily-cap test threw at its first assertion *before* its trailing inline `delete …CAP`, leaking a 1000 cap into every later test (one break → six red). `afterEach` now resets only the per-test cap keys; suite-level enabling vars stay intact.
- INSTANCE-USD + version-conflict needed no cap change (they only failed via the leak).
- Refreshed stale comments (estimate is `prompt.length`, not `ceil(chars/4)`).

Verified against a real Postgres (CI's exact command): AI integration **14/14**, PR-touched unit **28/28**, full CI integration list **56 files / 438 tests** green.

**Design note (confirmed acceptable, no action):** counting the conservative estimate (completion bound = `maxOutputTokens` 1024) makes token-cap admission ~2.5× pessimistic vs typical actual completions — a user hits the cap after ~`cap/1024` runs rather than `cap/actual`. At the default 100k/day cap this is a non-issue; it only bites under a very tight token cap near `maxOutputTokens`. For the USD cap (real money) the conservatism is intended. Keeping as-is per the no-overshoot goal.

PR #2627, opened 2026-06-15
feat(attendance): annual-leave L0 — latent config + deductLeaveBalance extraction

zensgit:
## What (L0 of the annual/statutory leave-balance engine — design-lock #2622, tracker §0.4)

Two **no-behavior-change** changes. Implementation gated; nothing accrues/deducts for annual until L2/L3.

**(a) Latent `annualLeavePolicy` org-settings config (default OFF)** — `enabled` / `tenureMode` / `standardDayMinutes` / statutory `tiers` preset (累计工作时间 1–10y=5d / 10–20y=10d / ≥20y=15d) / `carryover.enabled` / `timezone`, wired through `DEFAULT_SETTINGS` + normalize + `mergeSettings` + the PUT/GET zod schema. **Nothing reads it yet** (L2 accrual consumes it). Defensive `normalizeAnnualLeaveTiers` falls a malformed ladder back to the preset (never silently persists a broken ladder).

**(b) Extracted `deductCompTimeBalance` → generalized `deductLeaveBalance`** (`leaveTypeCode, amountMinutes, sourceType, insufficientCode, insufficientLabel, sourceId`); `deductCompTimeBalance` is now a thin comp_time wrapper → **call site unchanged, comp_time byte-identical**.

## Validation (local, real DB)

- `node --check` ✓
- **④ C3 comp_time deduct test passes unchanged** through the generalized helper — the design-lock §8 L0 byte-identical lock. ✓
- New `annualLeavePolicy` round-trip assertion (custom tiers ladder + timezone) survives PUT→persist→GET. ✓
- Full `attendance-plugin.test.ts`: my two target tests pass; the only failures are **pre-existing auto-shift local-pollution flakes** (proven identical on clean origin/main via stash — `deductLeaveBalance` count=0). CI's fresh DB won't hit them.

## One judgment call for your review

The design-lock §5 listed a `deductionBasis` param on the engine. **L0 omits it**: the *caller* owns amount semantics (comp_time passes actual minutes; 年假 L3 will pass `requestedDays × standardDayMinutes`), so including it in L0 with no annual caller would be a dead param (PR-hardening: no dead params). The "annual must not pull actual shift minutes" guardrail moves to **L3**, where the annual approval computes the standard-day amount. Flagging in case you'd rather pin it in the signature now.

Per connected-run A: opening for your merge; on merge I immediately start **L1** (expiry `source_type` generalization, comp_time no-regress).

### DIFF
--- a/packages/core-backend/tests/integration/attendance-plugin.test.ts
+++ b/packages/core-backend/tests/integration/attendance-plugin.test.ts
@@ -7455,6 +7455,65 @@ attendanceIntegrationDescribe(
     expect(emptyTemplateBody.error?.message).toContain('required')
   })
 
+  it('④/年假 L0 — config integrity: enabling requires a timezone (422), and a broken tiers ladder falls back to the preset', async () => {
+    if (!baseUrl) return
+    const dbUrl = process.env.ATTENDANCE_TEST_DATABASE_URL || process.env.DATABASE_URL
+    if (!dbUrl) return
+    const userId = `attendance-annual-l0-guard-${Date.now().toString(36)}`
+    const tokenRes = await requestJson(
+      `${baseUrl}/api/auth/dev-token?userId=${encodeURIComponent(userId)}&roles=admin&perms=attendance:read,attendance:write,attendance:admin`
+    )
+    const token = (tokenRes.body as { token?: string } | undefined)?.token
+    expect(token).toBeTruthy()
+    if (!token) return
+    const headers = { Authorization: `Bearer ${token}`, 'Content-Type': 'application/json' }
+    const readAnnual = async () => {
+      const r = await requestJson(`${baseUrl}/api/attendance/settings`, { headers: { Authorization: `Bearer ${token}` } })
+      return (r.body as { data?: { annualLeavePolicy?: { timezone?: string | null; tiers?: Array<{ minYears: number; maxYears: number | null; days: number }> } } } | undefined)?.data?.annualLeavePolicy
+    }
+    const putAnnual = (annualLeavePolicy: Record<string, unknown>) =>
+      requestJson(`${baseUrl}/api/attendance/settings`, { method: 'PUT', headers, body: JSON.stringify({ annualLeavePolicy }) })
+    const statutoryPreset = [
+      { minYears: 1, maxYears: 10, days: 5 },
+      { minYears: 10, maxYears: 20, days: 10 },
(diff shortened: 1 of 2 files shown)


zensgit:
## Review fixes — both P2s addressed (commit `3bce90773`)

**[P2-1] enabled=true without timezone** → fixed. Post-normalize guard in the settings PUT (`mergeSettings` → guard → `saveSettings`): `annualLeavePolicy.enabled===true` with no `timezone` returns **422 `ANNUAL_LEAVE_TIMEZONE_REQUIRED`** — no UTC fallback, honoring the #2622 lock. L0 requires an **explicit** timezone; inherited-tz resolution (attendance default rule / org) is a documented **L2/L4** extension (the guard is the place it'll live). Disabled policy persists freely.

**[P2-2] tiers only per-entry validated** → fixed. `normalizeAnnualLeaveTiers` now validates the **whole ladder**: contiguous, ascending, non-overlapping, ≤1 trailing open-ended band. Overlaps / gaps / out-of-order / multiple-open-ended fall the whole list back to the statutory preset (a broken ladder is never persisted). **Gaps are treated as malformed** (statutory bands are contiguous) — documented in the code comment + the new test.

**New real-DB reverse test** (`④/年假 L0 — config integrity …`): enable-without-tz → 422; enable+tz+overlapping ladder (1–10 & 5–20) → 200 with GET returning the preset; a contiguous custom ladder is preserved (fallback is targeted, not blanket).

Non-blocking items: `deductionBasis`-to-L3 unchanged (per your + Avicenna's concurrence). Local: `node --check` + the guard test + ④ C3 byte-identical + round-trip all green.

zensgit:
## Re-review fix — closed-tail ladder (new P2) addressed (commit `ec159a732`)

`normalizeAnnualLeaveTiers` now requires a non-empty ladder to **end in exactly one open-ended (`maxYears=null`) band**, else the whole list falls back to the statutory preset. So `[{minYears:1, maxYears:10, days:5}]` (closed tail → senior staff have no matching tier) no longer persists. Gaps and a closed tail are both malformed (statutory bands are contiguous and cover all eligible tenure); an empty list stays allowed (explicit no-bands).

New reverse-test case (5): closed-tail ladder → 200 with GET returning the statutory preset. Local: node --check + guard test (incl. closed-tail) + ④ C3 byte-identical + round-trip green.

PR #2644, opened 2026-06-15
docs(multitable): button/action field scope gate (B1)

zensgit:
**Docs-only scope gate. Runtime / migration / frontend NOT started.** Grounded on `origin/main@92fc9726d`.

## What this is
A contract/boundary doc for **B1**: a new `'button'` multitable field type. The field stores **config only** (a label + ONE bound action from the existing action set) and **no per-record data value**. Its cell renders a clickable control; on click it runs that one action against the button's own row record by **reusing the existing automation single-action execution engine** — without authoring, persisting, or triggering a full automation rule.

## Key design decision
Run via the existing **`AutomationService.executeRule`** (`automation-service.ts:1024`) with a **synthetic, non-persisted, trigger-less, condition-less, single-action `workflow_job_v1` envelope** — the same shape as the existing `testRun` (`automation-service.ts:1446`), but with the **real clicked record** and the **live clicker's** id. This reuses the C1 job/log plane (one job row + one execution-log row) AND every lock / cross-base / permission / redaction gate baked into the write-action executors — for free.

Rejected: the private `executeSingleAction` (`automation-executor.ts:1338`) — it returns a bare result with no execution/job rows and would force re-implementing the C1 wiring + a context shim.

**Prereq verified — no backbone gap (write AND read):**
- `multitable_automation_jobs.rule_id` / `multitable_automation_executions.rule_id` are `TEXT NOT NULL` with **no FK** → a synthetic `btn_`-prefixed `rule_id` inserts cleanly.
- The C1 read path is sheet-scoped (`listExecutions({sheetId})`, no `automation_rules` join) + execution-id-keyed detail → button runs **appear in the existing runs list and hydrate detail** without a parent rule row.
- Only edge accommodations (not engine changes): the run endpoint performs the record-write capability check explicitly (no `requireRecordWritable` middleware exists; `requireRecordReadable` does); `canEditRecord` gated uniformly at the route (exact for writes, deliberate strict over-approximation for `send_*`); runs view labels `btn_` runs.

## Lock-safe posture (out of scope)
NO new action types · NO multi-action / conditional / triggered / scheduled runs · NO suspending/branching bound actions (`wait_for_callback`/`start_approval`/`condition_branch`/`parallel_branch`) in v1 · NO new RBAC tier · NO bypass of locks / permission / redaction / governance · NO cross-base beyond existing governance · NO public/unauthenticated endpoint · NO stored per-record value · NO persisted synthetic rule. Each expansion = its own gated opt-in.

Runtime is split into **three separate opt-in slices** (B1-a field-type+codec → B1-b run endpoint → B1-c FE cell); landing this gate authorizes **none** to start.

Doc: `docs/development/multitable-button-action-field-scope-gate-20260615.md` (+472, single file).

### DIFF
--- a/docs/development/multitable-button-action-field-scope-gate-20260615.md
+++ b/docs/development/multitable-button-action-field-scope-gate-20260615.md
@@ -1,0 +1,490 @@
+# Multitable Button / Action Field Scope Gate (B1) - 2026-06-15
+
+Status: docs-only scope gate; runtime, migration, and frontend not started.
+
+Scope: add a new `'button'` multitable field type whose cell renders a clickable
+control that, on click, runs **exactly one** pre-configured action on the
+button's own row record by **reusing the existing automation single-action
+execution engine** — without authoring, persisting, or triggering a full
+automation rule.
+
+Grounded on: `origin/main@92fc9726d`.
+
+Companions:
+
+- `multitable-automation-a6-execution-plan-20260601.md`
+- `multitable-automation-run-governance-todo-20260527.md`
+- `multitable-automation-a6-3-3-branch-local-wait-scope-gate-20260615.md`
+
+## 0. Verdict
+
+### What B1 IS
+
+B1 is a new **value-less field type** (`'button'`). Its field definition stores
+*configuration only* — a display label plus **one** bound action chosen from the
+already-shipped action set. No per-record data value is ever stored, read, or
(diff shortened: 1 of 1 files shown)


zensgit:
Superseded by merged design-lock #2645, which deliberately chose a different (flat-shape, inert-record_click) design. Closing.

PR #2663, opened 2026-06-15
feat(multitable): color-scale + icon-set conditional formatting backend (A5-2/A5-3)

zensgit:
## Summary

Adds two range-based conditional-formatting **scale** kinds to the multitable backend, both extending the A5-1 data-bar scale machinery (the shared `sanitizeConditionalFormattingScaleRule` / `buildFieldScaleMap` pipeline that derives per-record presentation off an auto/fixed numeric range):

- **A5-2 color scale** — 2-stop (min/max) or 3-stop (min/mid/max) hex gradients. `buildFieldScaleMap` normalizes each value to `t ∈ [0,1]` over the field's range and emits an interpolated `scaleColor`. 3-stop is piecewise with `mid` anchored at `t=0.5` (`t≤0.5` lerps min↔mid over `t/0.5`, else mid↔max over `(t-0.5)/0.5`).
- **A5-3 icon set** — `arrows3` / `traffic3` / `signs3` sets banded by two absolute ascending thresholds (`v<t0` low / `t0≤v<t1` mid / `v≥t1` high), emitting a per-record `iconKey`.

Both reuse the same sanitizer (enum-strict kind dispatch, hex/anchor/threshold validation, rule-cap + order-sort) and the same `buildFieldScaleMap` field map, with `colorScale` populating `scaleColor` and `iconSet` populating `iconKey` (data-bar fields stay absent on each, and vice-versa). The FE mirror spec (`apps/web/tests/multitable-cf-scale.spec.ts`) locks the front-end's mirror of this contract byte-for-byte.

## Why A5-2 and A5-3 are combined in one PR

They edit the **identical** surface: the same `sanitizeConditionalFormattingScaleRule`, the same `buildFieldScaleMap`, and the same FE-mirror spec. Splitting them across two branches would put both branches in conflict on those exact files and stall on merge cadence (each rebase re-touching the shared sanitizer/build-map). Keeping them in one PR avoids the conflict while preserving review clarity — each kind has its **own dedicated test block** (separate `describe` for A5-2 color scale and A5-3 icon set) so the two contracts are validated independently.

## Hardening in this PR

The 3-stop color-scale build tests previously exercised `t ∈ {0, 0.25, 0.5, 1.0}`, so the upper segment (`mid↔max`, the `t>0.5` branch) only ran at its `t=1` endpoint — where the lerp returns `stops[2]` regardless of the stop indices/denominator, masking a plausible bug (wrong stop index or wrong denominator) in that branch. Added an interior record at value 75 (`t=0.75`) asserting `scaleColor` `#80ff00` (lerp `#ffff00`→`#00ff00` at 0.5) so the upper-segment interior is now covered. Mirrored byte-identically into the FE mirror spec to keep the mirror-drift guard intact. (Verified non-vacuous: injecting `stops[0]` for `stops[1]` in the upper branch yields `#808000` and fails the new assertion, while the old `t=1`-only test stayed green.)

## Scope note

This is backend/contract + unit coverage only. The grid render path (**MetaGridTable**) and the authoring dialog (**ConditionalFormattingDialog**) remain **browser-gated** and are not exercised here.

## Tests

- `packages/core-backend/tests/unit/multitable-conditional-formatting.test.ts` — 70 passed (A5-1/A5-2/A5-3 sanitizer + build-map)
- `apps/web/tests/multitable-cf-scale.spec.ts` — 24 passed (FE mirror)
- `tsc --noEmit -p tsconfig.json` on `@metasheet/core-backend` — 0 errors

### DIFF
--- a/apps/web/src/multitable/types.ts
+++ b/apps/web/src/multitable/types.ts
@@ -132,6 +132,19 @@ export interface ConditionalFormattingDataBarConfig {
   negativeColor?: string
   showValue?: boolean
 }
+/** A5-2 color scale: 2 stops (min/max) or 3 stops (min/mid/max), each a hex color. */
+export interface ConditionalFormattingColorScaleConfig {
+  stops: Array<{ at: 'min' | 'mid' | 'max'; color: string }>
+}
+/**
+ * A5-3 icon set: a 3-icon set keyed by two ascending thresholds (`[t0, t1]`).
+ * Thresholds are ABSOLUTE values (band by `v < t0` / `t0 <= v < t1` / `v >= t1`),
+ * not percentiles — a percentile mode over the field's min/max is a follow-up.
+ */
+export interface ConditionalFormattingIconSetConfig {
+  set: 'arrows3' | 'traffic3' | 'signs3'
+  thresholds: [number, number]
+}
 export interface ConditionalFormattingScaleRule {
   id: string
   order: number
@@ -140,6 +153,8 @@ export interface ConditionalFormattingScaleRule {
   enabled: boolean
   range: ConditionalFormattingScaleRange
   dataBar?: ConditionalFormattingDataBarConfig
+  colorScale?: ConditionalFormattingColorScaleConfig
+  iconSet?: ConditionalFormattingIconSetConfig
 }
 export const CONDITIONAL_FORMATTING_SCALE_RULE_LIMIT = 20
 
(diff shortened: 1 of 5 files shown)


zensgit:
Superseded by merged #2664, which landed the same A5-2/A5-3 backend AND the kind-aware grid render (the very render blocker this PR flagged). Closing as redundant — our two independent builds converged on the same sanitizer/buildFieldScaleMap design (confirmatory). Branch retained.

PR #2676, opened 2026-06-16
fix(multitable): close per-field-restore change-timeline oracle for hidden columns (#2672 follow-up)

zensgit:
## The oracle (security bug on merged #2672 code)

The per-field record-restore route (`POST /sheets/:sheetId/records/:recordId/restore` with `fieldIds`) is a **change-timeline oracle** for `visible=false` / read-only columns, defeating the #2144 record-history redaction mask.

The forbidden-field gate ran over `selectedDiff` = (caller `fieldIds` ∩ fields-that-**actually changed** between target version and current). So selecting a forbidden field `F`:
- returns **403** when `F` **changed** at the target version, vs
- returns **200 no-op** when `F` was **unchanged**.

That is an exact per-field oracle for *"did this confidential column change at version N"* — sweep `targetVersion` → recover the full modification timeline. Values stay hidden; the leaked bit is the change-timeline. #2144's `redactRecordRevisionEntry` masks exactly that history for a `visible=false` field, so this route silently un-masks it.

## The fix (`packages/core-backend/src/routes/univer-meta.ts`)

In **per-field** mode (`fieldIds` present), run the forbidden check over the caller-**selected** `fieldIds`, independent of whether they changed, using the **same** permission predicate the route already uses (static guard `!hidden && !readOnly` ∧ layer-3 `visible !== false && readOnly !== true`). Any selected forbidden id → **403 `RESTORE_FORBIDDEN` regardless of diff** → constant response → **oracle closed**, and a selected field you cannot write can never silently fall through to a no-op bypass (**M2**).

Constraints honored:
1. **FULL-restore** mode (no `fieldIds`) keeps its existing atomic-reject behavior verbatim — the gate still runs over the changed diff (`gatedFieldIds = fieldIds ?? selectedDiff.map(...)`).
2. No forbidden id is echoed into `skippedFieldIds` (success path only reaches when every selected field passed, so `skippedFieldIds` stays `[]`).
3. The 403 message stays generic (no field id). A forbidden id is now indistinguishable from a non-field id (both fail the predicate) — no metadata re-leak.

This **restores the #2144 mask property**: a `visible=false` column's change-timeline can no longer be recovered through the restore response.

## Tests (real-DB integration, `tests/integration/multitable-record-restore.test.ts`)

- **T18 (oracle keystone)** — a `visible=false` field `FLD_HIDDEN` (the column #2144 redacts from the read-only actor's history) selected via `fieldIds` returns an **IDENTICAL** response (same status **and** body) whether or not it changed at vN. Both are a constant generic 403 with no field id in the body and the id not in `skippedFieldIds`; live data unchanged (atomic) in both. Chosen as `visible=false` (not the read_only-but-visible `FLD_SECRET`, which is already unmasked in the history endpoint) so the test exercises the exact mask-defeating case.
- **T19 (M2)** — selecting a write-denied field as the read-only user → 403 `RESTORE_FORBIDDEN`, body excludes the field id, live data unchanged.
- **T20 (regression)** — a writable per-field restore still applies; full-restore atomic-reject (changed-forbidden → 403, unchanged-forbidden → only the writable field in the diff) unchanged.

**Non-vacuity proof**: with the pre-fix `selectedDiff` gate temporarily reverted, **T18 fails** with `expected 403 to be 200` — the changed case is 403 while the unchanged case is a 200 no-op (the exact oracle signature). It passes only with the gate-over-selected-fields fix. (T19/T20 hold on both old and new code by design — they assert the M2-differing and regression cases, which were already correct; T18 is the single bug-discriminating keystone.)

## Verification
- `pnpm --filter @metasheet/core-backend exec tsc --noEmit -p tsconfig.json` → 0 errors.
- New + existing record-restore suite: **29/29 green** (26 prior + T18/T19/T20).

### DIFF
--- a/packages/core-backend/src/routes/univer-meta.ts
+++ b/packages/core-backend/src/routes/univer-meta.ts
@@ -5794,24 +5794,38 @@ export function univerMetaRouter(): Router {
         }
       }
 
-      // Per-field (column-level) restore: when the caller passes `fieldIds`, restrict the diff to that
-      // selection. The atomic gate below then runs over only the selected subset — so a user can restore
-      // the writable fields they picked even if another changed field in the revision is forbidden. A
-      // requested field that is unchanged / non-restorable / unknown simply isn't in the diff (no error).
+      // Per-field (column-level) restore: when the caller passes `fieldIds`, restrict the applied diff to
+      // that selection. A selected field that is unchanged / non-restorable simply isn't in the diff and is
+      // a clean no-op for that field — it is NOT silently treated as writable below (see the forbidden gate).
       const selectedDiff = fieldIds ? diff.filter((ch) => fieldIds.includes(ch.fieldId)) : diff
 
-      // Lock B: gate every (selected) diff field on BOTH the static guard and restore's own layer-3 pre-check.
-      const hasForbidden = selectedDiff.some((ch) => {
-        const guard = fieldById.get(ch.fieldId)
-        const perm = fieldPermissions[ch.fieldId]
+      // Lock B (the forbidden gate). The permission predicate (static guard ∧ restore's layer-3 pre-check)
+      // is the SAME in both modes; only the SET it runs over differs:
+      //
+      //  • FULL restore (no `fieldIds`): gate the changed diff — an atomic reject when any CHANGED field is
+      //    forbidden. Unchanged forbidden fields are not in the diff and don't block (existing behavior,
+      //    constraint (1) — kept verbatim). This carries no per-field bit (the caller named no field).
+      //
+      //  • PER-FIELD (`fieldIds` present): gate the caller-SELECTED ids INDEPENDENT of whether they changed.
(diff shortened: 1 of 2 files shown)


zensgit:
Superseded by #2677 as the canonical per-field hidden-field strategy. Both close the change-timeline oracle, but #2677 is the better mainline: it drops a hidden/visible=false selected field to a 200 no-op (indistinguishable from unknown/deleted), so a legitimate partial restore proceeds while silently skipping fields the actor can't see — whereas this PR's uniform 403 fails the whole request if it merely names a hidden field. #2677 also covers schema-drift scope + the link-reorder no-op. Closing to keep one canonical strategy; the parallel session's #2677 wins.

Available to port if useful: this PR's T18 oracle keystone asserts the changed-vs-unchanged hidden-field responses are byte-identical (the direct anti-oracle proof), adversarially verified PASS. Happy to contribute it adapted to #2677's 200-noop semantics if #2677 lacks the explicit changed-vs-unchanged equality assertion.

PR #2814, opened 2026-06-17
feat(multitable): non-string-field realtime sync — Yjs bridge scalar core (inert until FE binding)

zensgit:
Realtime sync for SCALAR fields (number/currency/percent/rating/duration/select/multiSelect/date/dateTime/boolean) — the BE core. Today only string fields (Y.Text) sync; scalars don't. Correct semantic for atomic values = **LWW via the record `fields` Y.Map** (Yjs Y.Map is a CRDT, LWW per key).

`yjs-record-bridge` observer + `PendingWrite.fields` + `scheduleFlush` widened `string`→`unknown`; the observer now also collects plain Y.Map values (skipping Yjs shared types — Y.Text handled separately, Y.Map/Y.Array out of scope — and `undefined` deletes). Scalars flush through the **same validated `patchRecords` path** as Y.Text (no validation bypass), field-type-agnostic.

**Safe + inert:** nothing writes scalar values into the Y.Map until the FE scalar binding lands (forking now), so this changes no behavior. New unit test `yjs-bridge-scalar-fields.test.ts` 2/2 (scalar collection w/ native types; Y.Text regression; nested-shared-type exclusion); tsc clean; full suite 4432/0.

**Follow-up (forking now):** the FE scalar cell binding (read/write the Y.Map key as a typed plain value) + grid wiring — the user-facing half + a two-client e2e test.

### DIFF
--- a/packages/core-backend/src/collab/yjs-record-bridge.ts
+++ b/packages/core-backend/src/collab/yjs-record-bridge.ts
@@ -25,7 +25,7 @@ export interface YjsRecordBridgeConfig {
 export type ActorResolver = (socketId: string) => string | undefined
 
 interface PendingWrite {
-  fields: Record<string, string>
+  fields: Record<string, unknown>
   actorIds: Set<string>
   timer: NodeJS.Timeout
   firstSeen: number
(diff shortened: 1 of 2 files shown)


zensgit:
Superseded by the complete non-string-CRDT PR (BE bridge core + seed + FE primitive + grid wiring) — this branch's BE-core commit is included there.

PR #2815, opened 2026-06-17
feat(multitable): useYjsScalarCell — FE scalar CRDT binding primitive (inert until grid wiring)

zensgit:
The reusable FE primitive for scalar realtime sync (pairs with the BE bridge core #2814). `useYjsScalarCell` — type-agnostic (caller passes the typed value; never coerces), mirrors useYjsCellBinding's flag-gate / synced-and-key-exists activation / REST-fallback / cleanup; reads/writes the record `fields` Y.Map key.

**Inert** until the grid calls it + the BE seeds scalars (deferred — see below). Tested 6/6 (seeded read native types; setValue writes typed number/boolean; two-client Y.Map LWW sync; unseeded key inactive + no-op; remote refresh; flag-off inert). vue-tsc clean.

**Deferred — the corruption-risky coupling (needs a dedicated pass + the field-type registry):** the BE seed of non-string scalars into the Y.Map (yjs-sync-service only seeds strings today), the per-field-type representation decision (Y.Text vs plain-value — esp. select/date, which are strings today), and the grid wiring. Guessing the representation wrong = persisted corruption (no flag-off safety here), so it's explicitly NOT rushed.

### DIFF
--- a/.github/workflows/multitable-web-guard.yml
+++ b/.github/workflows/multitable-web-guard.yml
@@ -34,6 +34,8 @@ on:
       - 'apps/web/src/multitable/components/MetaGridTable.vue'
       - 'apps/web/src/multitable/views/MultitableWorkbench.vue'
       - 'apps/web/tests/multitable-sheet-cursor-state.spec.ts'
+      - 'apps/web/src/multitable/composables/useYjsScalarCell.ts'
+      - 'apps/web/tests/multitable-yjs-scalar-cell.spec.ts'
       - '.github/workflows/multitable-web-guard.yml'
   push:
     branches: [main]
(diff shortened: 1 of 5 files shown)


zensgit:
Closing — the rebase dropped the useYjsScalarCell commit on a web-guard.yml conflict, so this PR ended up a duplicate of the BE core (#2814). The FE scalar binding primitive (useYjsScalarCell, commit 938fd0129) is preserved and will land with the deferred non-string-CRDT functional pass (seed + per-field-type Y.Map representation decision + grid wiring), where the corruption-risk per-type representation is resolved together rather than guessed.

PR #2825, opened 2026-06-18
docs(multitable): close last CRDT comment tails (bridge class JSDoc + dateTime example)

zensgit:
Follow-up to #2823 (reviewer P3). Comment-only: yjs-record-bridge class JSDoc reworded from 'When a Y.Text field changes' → 'a field inside that map changes — Y.Text edit OR plain scalar value' (it collects both); yjs-sync-service head + body examples gain `dateTime` to fully match the stated boundary. Diff is comments only; tsc clean. 🤖 Generated with Claude Code


zensgit:
Superseded by #2859, the single P3 ledger/comment reconciliation pass (reconciles plan/TODO + completion doc + the CRDT comments + field-codecs to the true post-#2849/#2854 state). Closing to avoid re-landing stale doc state.

PR #2851, opened 2026-06-18
docs(multitable): close 2a in gated-remainder plan/TODO (dateTime #2849 shipped)

zensgit:
Refreshes the gated-remainder plan + TODO now that **2a is complete** — select/date (#2832), duration (#2838), dateTime (#2849, 2a-DT-S2). Moves 2a to the §0 closed list in both docs, adds CLOSED banners to §1 (kept as history), and narrows the active remainder to **2b** (conditional permission rules) + **2c** (person org-directory). Addresses the #2840 review's non-blocking follow-up (don't let the ledger overstate 2a). Docs-only.


zensgit:
Hold — do not merge as final. Per review: this only updates plan/todo, but the runtime contract hardening (#16 write-path route parity + admin-bypass at form-context/trash + /patch person guard) is not landed yet. After those land, the plan/todo + completion doc + code comments (esp field-codecs.ts:293 'NOT YET ENFORCED') will be updated together in one accurate pass. Keeping this open/blocked until then.

PR #2914, opened 2026-06-19
fix(multitable): enforce per-subject field read-only on the main bulk PATCH (P0 audit defect)

zensgit:
## What

Closes **P0 audit defect #3** (verified): per-subject `field_permissions.read_only` was enforced only on the **restore + copy** paths, **not** on the everyday `POST /patch` grid edit. The write spine guards only PROPERTY-level field flags, so a field set read-only for a subject via `field_permissions` was still writable through the grid — a real permission hole.

## Fix

`/patch` now uses the `fieldPermissions` that `buildRecordPatchContext` **already derives** (over the full visible-property-field set) and rejects (403, fail-closed) any submitted change whose field is layer-3 read-only / not-visible for the actor — parity with the restore and copy gates.

## Why it's safe (no hot-path regression)
- The spine already **throws** on property-readonly/hidden/computed fields (`record-write-service.ts:449-465`); the only NEW rejection is the intended per-subject case. Normal editable fields are unaffected.
- **No block-all trap:** `fieldPermissions` is derived over `visiblePropertyFields`, so every visible field has an entry — a field with no explicit per-subject rule stays writable (proven by the regression-guard test).
- **Atomic** (restore-style): a mixed request with any forbidden field is rejected whole; nothing is written.
- `/patch` ids are caller-submitted → naming rejected ids leaks nothing (nonexistent and read-masked fields fail identically; no existence oracle).
- Per-subject read-only applies regardless of role (no admin special-case) — same as restore/copy; admins manage by editing the rule, not bypassing it at write time.

## Test (real-DB golden, in `plugin-tests.yml`)
per-subject read-only field rejected (403, nothing written) · **normal field still writes** (regression guard vs block-all) · mixed request rejected atomically (normal field not written). `tsc --noEmit` clean.

P0 batch: this is 1 of the 3 audit defects. #1 (dashboard provider wiring) is next; #2 (open-API token auth) I'm surfacing as a design decision rather than auto-building.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -193,6 +193,7 @@ jobs:
             tests/integration/multitable-rowlevel-readdeny-enforce.test.ts \
             tests/integration/multitable-conditional-rule-enforce-realdb.test.ts \
             tests/integration/multitable-conditional-rule-trash-realdb.test.ts \
+            tests/integration/multitable-fieldperm-write-gate-patch-realdb.test.ts \
             tests/integration/multitable-conditional-rules-api.test.ts \
             tests/integration/multitable-rowlevel-readdeny-xrec.test.ts \
             tests/integration/multitable-rowlevel-readdeny-flag-endpoint.test.ts \
--- a/packages/core-backend/src/routes/univer-meta.ts
+++ b/packages/core-backend/src/routes/univer-meta.ts
@@ -11619,7 +11619,25 @@ export function univerMetaRouter(): Router {
       if (!patchContext) {
         return res.status(404).json({ ok: false, error: { code: 'NOT_FOUND', message: `Sheet not found: ${sheetId}` } })
       }
-      const { fields, readableEchoFields, readableEchoFieldIds, attachmentFields, fieldById } = patchContext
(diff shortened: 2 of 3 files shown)


zensgit:
Reviewed the fix — **APPROVE** (security-correct, no changes requested):

- **Fail-closed** is right: `!perm || perm.visible === false || perm.readOnly === true` → forbidden. Since `fieldPermissions` is derived over the full visible-property-field set, a missing entry means not-visible/unknown ⇒ rejected — never a false block of a normal field.
- **Atomic / restore-style**: the whole request is rejected before any write (no silent-drop). The mixed-request test proves the normal field is NOT persisted when a locked field is present.
- **Parity + no admin bypass**: matches the restore/copy layer-3 gates; per-subject `read_only` applies regardless of role, and the test note correctly documents that admin authority = edit/remove the rule, not write-time bypass.
- **No info leak**: `/patch` fieldIds are caller-submitted; nonexistent vs read-masked fail identically (no existence oracle).
- Real-DB golden covers enforce(403, nothing written) + regression-guard(normal field writes) + atomicity; registered in `plugin-tests.yml`.

Closes a genuine write-bypass hole (per-subject field read-only was enforced on restore/copy but not the everyday grid PATCH). Good to land on green.

PR #2945, opened 2026-06-19
feat(multitable): 2a nested AND/OR filter groups — backend (tree-aware eval + anti-oracle) [SECURITY REVIEW]

zensgit:
2a view-filter operator depth — **nested AND/OR condition groups** (backend evaluator + permission guards). **Opened for your security review — NOT auto-merged.**

## Correction up front
I earlier called this "regression-risk only, not security-sensitive." That was **wrong**, and the max-effort investigation proved it: nested groups touch the **anti-oracle guards**. A denied-field condition nested in a group must be pruned-before-eval, redacted-on-read, and never-erased-on-resave — exactly as for the flat top level. A miss **leaks permission data**. So this routes through your review.

## Design — additive, backward-compatible by construction
`MetaFilterInfo` gains an **optional** `group?: MetaFilterGroup`. Flat filters carry no `group` → every existing path is unchanged. `parseMetaFilterInfo` parses the tree, enforces **caps (depth ≤ 5, ≤ 50 leaves)**, and flattens leaves into `conditions` as a mirror (so field-reference consumers like computed-field detection work unchanged).

## Tree-aware at all 8 consumer sites (the security surface)
| site | handling |
|---|---|
| eval ×3 (view-filter flows A/B/C) | shared `matchesFilterInfo` (group when present, else flat) |
| **anti-oracle prune ×2** | `pruneFilterInfoByField` drops denied leaves at **every depth** — a flat-only prune would leave a denied leaf live inside a group → surviving-row-set oracle |
| **redaction** | `redactViewConfigFilterLiterals` recurses the raw tree; no longer early-returns on a group-only filterInfo |
| **re-save merge guard** | **fail-closed** — rejects a field-denied user's resave of a group filter (tree-position restoration intentionally unsupported; never silently erase) |
| computed-field detect | rides the flattened `conditions` mirror (unchanged) |
| rollup field's own config filter | different system, intentionally out of scope (flat by design) |

**Adversarial self-review:** enumerated every `parseMetaFilterInfo` call-site (exactly the 3 flows) and every `evaluateMetaFilterCondition` site; **no leak found** across eval/prune/redact/merge.

## D-B4 decision surfaced (per #2938)
The re-save guard **rejects** a denied user's group-filter resave rather than attempt fragile tree-position literal preservation — the provably-safe v1 choice. Please confirm or redirect.

## Verification
Pure-helper goldens with the security properties as the centerpiece (`multitable-view-filter-groups.test.ts`): denied leaf pruned at depth (no-leak), redaction recurses incl. group-only, merge fail-closed, caps reject depth/count, flat path byte-identical. CI runs tsc + the suite.

**Scope:** backend only; FE group builder is a follow-up to coordinate with the active 2a FE session (#2935/#2941).

### DIFF
--- a/packages/core-backend/src/routes/univer-meta.ts
+++ b/packages/core-backend/src/routes/univer-meta.ts
@@ -1986,10 +1986,23 @@ function isNullishSortValue(value: unknown): boolean {
 
 type MetaFilterConjunction = 'and' | 'or'
 type MetaFilterCondition = { fieldId: string; operator: string; value?: unknown }
+// 2a nested groups: a node is either a leaf condition or a sub-group. Backward-compatible — `group` is
+// OPTIONAL and only the new nested form populates it; flat filters are untouched.
+type MetaFilterNode = MetaFilterCondition | MetaFilterGroup
+type MetaFilterGroup = { conjunction: MetaFilterConjunction; children: MetaFilterNode[] }
 
 type MetaFilterInfo = {
   conjunction: MetaFilterConjunction
-  conditions: MetaFilterCondition[]
+  conditions: MetaFilterCondition[] // flattened leaves; ALWAYS the full leaf set (mirror of `group` when present)
+  group?: MetaFilterGroup // when present, the source of truth for eval/prune/redact; `conditions` mirrors its leaves
+}
+
+// DoS guards for the nested form (the flat form is unchanged / uncapped to preserve existing behavior).
+const MAX_FILTER_GROUP_DEPTH = 5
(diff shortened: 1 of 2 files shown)


zensgit:
Superseded by #2947 (parallel session) — same feature (2a nested AND/OR filter groups backend), merged first with the full recursive parse/eval/prune + redaction + depth guard. Closing mine to avoid a divergent second implementation of the anti-oracle recursion. I'm auditing #2947's tree-prune coverage at every eval site to confirm no leak-miss (the security property my PR's adversarial tests targeted); will flag/fix if any site evaluates the unpruned tree.

PR #2978, opened 2026-06-21
feat(multitable): 1b Slice A — relation-scoped criteria aggregation (RELSUMIF) + foreign-write fan-out (A.1+A.2)

zensgit:
First implementation slice of the ratified **1b record-set/range** design-lock: a relation-scoped criteria aggregation formula function, built narrow and with the safety rails locked — **not** a casual SUMIF bolt-on.

## Naming (re-confirmed per the design-lock's binding gate)

The function is **`RELSUMIF`** — a deliberately **relation-scoped** name. It sums a foreign target over the records **linked to the current row**, keeping only those matching one criteria. It is **NOT** classic whole-sheet `SUMIF(range, criteria, sum_range)` and makes no such promise; the whole-sheet record-set scan remains a **separate gated extension** (§3 of the lock). The relation-scoped name is chosen precisely so a user can't mistake it for whole-sheet parity.

## Architecture (Block-6: wrapper materializes, pure engine aggregates)

`resolveRelationAggregation` is the wrapper — it materializes the **permission-filtered** linked-record set for one call (reusing the rollup core: `resolveReadableSheetIds` / `resolveForeignFieldReadability` / `shouldMaskForeignField` / `loadDeniedRecordIds`), then the **pure `aggregateRollup`** reduces it. The pure engine never does IO. The criteria *value* may be a current-record `{fld}` — the genuine capability rollup-with-filter lacks (its filters are static literals). The regex-substitute-into-the-expression approach was rejected; `RELSUMIF` is sole-call and parsed with a quote-aware splitter that refuses nested parens.

## Rails locked in this slice

- **`#PERM!`** — unreadable foreign target/criteria, or a cross-base foreign sheet (out of scope at lock time).
- **`#LIMIT!`** — `MAX_RELATION_SCAN_RECORDS`, checked count-first (never an unbounded scan).
- **Read-path taint leak edge** — `resolveTaintedFormulaFieldIds` now parses relation-agg formulas and **drops the field on read** for any reader who can't see its foreign target/criteria. This is the §2a.3 leak surface; the **fail-first golden** proves it (a denied reader gets the field *absent*, never the writer's materialized aggregate).
- **Composition cliff** — `RELSUMIF(...)+1` is rejected fail-loud (validator at save + `#ERROR!` at recompute). Composition is deferred (A.3).

## Permission model — Block-2 / Block-7 reconciliation (an amendment)

The two blocks appeared to conflict (per-actor `#PERM!` vs materialize-once). They don't, once paired: the value is **materialized at write** (writer's req; the existing tainted-skip leaves an unreadable writer's prior value untouched) **and masked per-reader at read**. **Observable (decided):** a denied reader sees the field **absent** (the shipped FOL model), not a `#PERM!` cell — this amends Block 2's "denied → `#PERM!`, never silent null" *for relation-aggregation formulas* to the FOL field-drop. Documented in the design-lock.

## Scope — what ships vs what's deferred (your call on fan-out)

**Ships (A.1):** sole-call `RELSUMIF`, **same-record** recompute (link/criteria-`{fld}` change → recompute), leak-free, real-DB goldens.

**Deferred, disclosed:**
- **A.2 — foreign-write fan-out.** You asked to lock fan-out in this slice. I built the cross-sheet **dependency + taint** so the read side is safe, but the foreign-write → recompute trigger for a *direct-foreign* formula needs a careful extension to `computeDependentLookupRollupRecords` (the existing chain only fans out through lookup/rollup intermediaries). Since the leak is fully closed by A.1's read mask, the only gap without fan-out is **staleness** (a foreign target/criteria edit isn't reflected until the next same-record write) — **not a leak**. This mirrors FOL's own **A-min → A-full** phasing. I held it out rather than rush a correctness-sensitive extension into a 9k-line function. **If you want it in this PR, say so and I'll add it next; otherwise it's the immediate follow-up.**
- **A.3 — `RELCOUNTIF`/`RELAVGIF` + composition.**

## Tests

- 11 pure unit tests (grammar + composition-cliff detector).
- Real-DB golden (registered in `plugin-tests.yml`): relation-scoped sum (30); the **fail-first read-leak gate** (denied reader → field absent; allowed → 300); composition cliff → `#ERROR!`.
- Full core-backend unit suite green locally (no guard regression); tsc clean.


zensgit:
**Updated scope — Slice A core complete (A.1 + A.2).** This PR now contains both:
- **A.1** — `RELSUMIF` relation-scoped criteria aggregation (materialize-at-write + per-reader read-mask leak edge, `#PERM!`/`#LIMIT!`/caps, current-record `{fld}` criteria).
- **A.2** — foreign-write **fan-out** via the FOL reverse-edge (a foreign target/criteria edit recomputes the source RELSUMIF), with the write-side taint-skip.

Plus the **development & verification MD** (`docs/development/multitable-1b-slicea-relation-aggregation-dev-verification-20260620.md`): permission model + Block-2/7 reconciliation + the materialize-model known property + the verification matrix + the research-complete designs for the next sub-slices (A.3 `RELCOUNTIF`/`RELAVGIF` + composition, Slice B lookup, Slice C array).

Real-DB goldens (CI test 20.x): relation sum=30 · **fail-first read-leak gate** (denied reader → field absent) · composition cliff → `#ERROR!` · **A.2 fan-out** (foreign edit → source recompute 30→1020). Auto-merge stays disarmed pending your review.

PR #3003, opened 2026-06-21
fix(multitable): export "all rows" via mask-preserving full-sheet route (+csv), not the 50-row client page

zensgit:
## The bug (verified, re-confirmed)

The multitable export picker (`MetaExportDialog.vue`) is shipped and wired, but the handler serialized the **client-loaded grid page** instead of the full sheet:

- `MultitableWorkbench.vue` → `onExportDialogConfirm` exported over `grid.rows` (client-side `doExportCsv` / `doExportXlsx`).
- The grid is page-replaced at `DEFAULT_PAGE_SIZE = 50` (`useMultitableGrid.ts`).
- So choosing **"all rows"** exported at most 50 rows (the loaded page) — a data-loss correctness bug (Feishu exports the whole sheet).
- Meanwhile a mask-preserving backend route existed but was unused by the FE: `GET /sheets/:sheetId/export-xlsx` (`univer-meta.ts`, PR #2591) — it applies `field_permissions` + view-hidden + §2a.3 formula-taint masking **after** the `fieldIds` selection, cursor-paginates the full sheet up to `XLSX_MAX_ROWS`, and 400s on an empty selection.

## The fix (rewiring)

**"All rows" → the mask-preserving backend route.** A new `MultitableApiClient.exportSheet({ sheetId, viewId, fieldIds, format })` hits `GET /sheets/:sheetId/export-xlsx` (auth attached via `apiFetch`, so the request carries the bearer token — a bare `<a href>` download could not), returns the binary blob + the server's `Content-Disposition` filename, and throws a `MultitableApiError` on non-2xx so the caller toasts the server message instead of downloading an error body. The picker's chosen visible columns are sent as `fieldIds` (selection may only **narrow** within the allowed set), plus the active `viewId` (drives the view hidden-field mask + scope validation).

**CSV: `?format=csv` on the SAME route** (chosen over a sibling route or FE-side streaming). The field/view/§2a.3-taint mask runs **before** the format branch — the `headers`/`rows` pipeline is format-agnostic — so CSV inherits the **identical** enforcement with no separate masking path (the cleanest option: it reuses the same masked cell projection, incl. the rich-longText→plain-text step). The route emits CSV with a UTF-8 BOM (Excel-friendly for CJK) + CRLF. An unsupported `format` value is a **400** (no silent fallback to xlsx — the repo's enum-strictness convention). Both formats emit `X-MetaSheet-XLSX-Truncated`.

**"Selected rows" stays client-side** over `grid.rows` (documented in code). Those rows come from `GET /view`, which masks via `filterRecordDataByFieldIds(row.data, allowedFieldIds)` where `allowedFieldIds` is the **taint-masked** set (`maskStoredRecordFieldIds`) — i.e. `grid.rows` are already field-permission AND §2a.3-taint masked, so serializing the chosen subset adds no egress surface. The selected ids are necessarily within the loaded page, so completeness is not at stake. This keeps the change minimal (no new `recordIds` route param).

## Security property (preserved)

A field-permission-hidden or formula-tainted column does **not** appear in the export even if its id is passed in `fieldIds`. The mask is computed **before** the format branch, so it holds for **both** xlsx and csv on the full-sheet path. Selection narrows, never widens — a denied/tainted id requested in `fieldIds` simply falls out of the intersection. Verified by tests (b) and (d-2). The three structural guards (raw-record-data-projection / egress, §2a.3 taint-chokepoint, record-lock) stay green — no new `filterRecordDataByFieldIds` / raw-projection / taint-resolver call site was added.

## Scope notes (named for review)

- **"All rows" = the full sheet, NOT the view-filtered subset.** The route applies the view's hidden-fields but not its row filter/sort (retrofitting in-memory filter/eval would abandon cursor streaming, add a row cap, and a 422-on-computed-filter failure mode — out of scope for this surgical fix). View-filtered export is a follow-up.
- **Column order**: the route emits sheet field order; the old client path used on-screen column order. If a user reordered view columns, the all-rows export column order now follows the sheet, not the view. Cosmetic.

## Tests

- **Real-DB integration** (`multitable-export-allrows-maskroute-realdb.test.ts`, registered in `plugin-tests.yml` inside the `DATABASE_URL`-gated real-DB step): seeds **120 rows** (> the 50-row page) and asserts **ALL** are exported (xlsx + csv); a `field_permissions`-denied column is **excluded even when its id is in `fieldIds`** (xlsx + csv); an allowed subset **narrows**; all-foreign selection and unsupported format both **400**. 8/8 pass via the exact CI command (`vitest --config vitest.integration.config.ts run`).
- **FE unit** (`export-sheet-client.test.ts`): `exportSheet` requests the route with `viewId` + comma-joined `fieldIds` + `format`, omits empty `fieldIds`, returns blob + `Content-Disposition` filename, and throws the server error (message + code) on non-2xx. 4/4 pass.

## Validation

- `pnpm install --frozen-lockfile` — clean (no lockfile drift).
- `cd packages/core-backend && npx tsc --noEmit -p tsconfig.json` — **0 errors**.
- `pnpm --filter @metasheet/web exec vue-tsc -b` — **0 errors**.
- `CI=true pnpm --filter @metasheet/core-backend test` — **357 files / 4676 tests passed, 0 failed** (103 files / 815 real-DB tests skipped without `DATABASE_URL`, as expected).
- Structural guards (egress / taint-chokepoint / record-lock) — green.
- Web suite: the 17 failing files / 111 failing tests are **pre-existing on clean origin/main** (verified by stash → identical baseline); this PR adds **+1 passing file / +4 passing tests** and **zero** new failures.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -235,6 +235,7 @@ jobs:
             tests/integration/multitable-lookup-foreign-field-mask-view.test.ts \
             tests/integration/multitable-lookup-foreign-field-mask-export.test.ts \
             tests/integration/multitable-export-permission-canary.test.ts \
+            tests/integration/multitable-export-allrows-maskroute-realdb.test.ts \
             tests/integration/multitable-person-directory-route.test.ts \
             tests/integration/multitable-lookup-foreign-field-mask-write.test.ts \
             tests/integration/multitable-lookup-foreign-field-mask-create.test.ts \
--- a/apps/web/src/multitable/api/client.ts
+++ b/apps/web/src/multitable/api/client.ts
@@ -322,6 +322,20 @@ function firstFieldError(fieldErrors?: Record<string, string>): string | null {
   return first ?? null
 }
 
+// Extract the download filename from a Content-Disposition header (handles both `filename="x.csv"` and
+// RFC-5987 `filename*=UTF-8''x.csv`). Falls back to a generic name so the download always has one.
+function parseContentDispositionFilename(header: string | null): string {
+  if (header) {
+    const star = header.match(/filename\*=(?:UTF-8'')?([^;]+)/i)
+    if (star?.[1]) {
+      try { return decodeURIComponent(star[1].trim().replace(/^["']|["']$/g, '')) } catch { /* fall through */ }
+    }
+    const plain = header.match(/filename="?([^";]+)"?/i)
+    if (plain?.[1]) return plain[1].trim()
+  }
+  return 'export'
+}
+
 type RawComment = Partial<MultitableComment> & {
   spreadsheetId?: string
   rowId?: string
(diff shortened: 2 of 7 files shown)


zensgit:
**Review (holding before merge) — strong fix, one product-semantic decision needed.**

✅ What's right: fixes the real data-loss bug (export "all rows" was capped at the 50-row loaded page); routes "all rows" through the mask-preserving full-sheet route; **the field-permission + §2a.3 taint mask is enforced for BOTH xlsx and csv** (tests b/d-2 prove a denied column is excluded even when its id is in `fieldIds`); 8/8 real-DB + 4/4 FE; full backend suite + structural guards green.

⚠️ Merge-blocking semantic: **"export all rows" now exports the ENTIRE sheet, not the view-filtered subset** (the route applies view hidden-fields but not `filter_info` row filter/sort). Before this PR, "all rows" exported the loaded page *of the filtered view* (filter-respected, just capped at 50). After, a user who filters to N rows and clicks "export all" gets the whole sheet — a filter-semantic regression and not Feishu-parity behavior (Feishu exports the filtered view).

**Options:**
- **(A) Make the route respect the view filter** (apply `parseMetaFilterInfo`/the same filter eval per page in the cursor loop) so "export all" = all rows of the *current filtered view*. Correct + parity-accurate; the agent flagged a cursor-streaming tension (computed-field filters need the in-memory slow path) — acceptable for filtered exports.
- **(B) Ship as "Export entire sheet"** — relabel the scope so it's honest (not "all rows"), and add "Export current view (filtered)" as a follow-up. Smaller, no surprise.
- **(C) Merge as-is** (accept full-sheet semantics for "all rows").

Recommendation: **(A)** for true parity, or **(B)** as the pragmatic interim. Holding for the owner's call — this is a product/UX decision, not a code defect.

zensgit:
Superseded by #3010, which contains this commit unchanged and additionally resolves the held open semantic (export now respects the view's row filter + sort — full filtered set, masked, csv+xlsx). Suggest closing this PR in favor of #3010.

zensgit:
Superseded by #3010, which contains this commit + makes export respect the view's row filter + sort (the held semantic, resolved as option A). Closing in favor of #3010.

PR #3043, opened 2026-06-22
feat(attendance): #7 leave cancellation + balance reversal (销假)

zensgit:
Builds **#7 leave cancellation + balance reversal (销假)** per design-lock #3034, with your recommended §3. **PR-not-merge.**

## What it does
Cancelling an **approved** leave now reverses the balance its approval deducted — instead of leaving the employee silently short (the reserved re-entry point the annual-leave engine deferred).

- **Migration** `zzzz20260622150000` — extends the `attendance_leave_balance_events` CHECKs to admit `event_type='reverse'` (δ>0): **both** `chk_event_type` and `chk_delta_sign` (drop + re-add; additive — the four existing types unchanged). *(Per the advisor: both constraints, or a reverse INSERT passes one and violates the other.)*
- **`reverseLeaveBalanceDeduction`** — keys on the request's **own per-lot `deduct` events** (`source_id=requestId`), so it reverses **whatever** was deducted: annual_leave **and** comp_time_leave both deduct with `source_id=requestId`, mirroring each event's `source_type` onto its reverse. *(The advisor's top correctness fix — a hardcoded `annual_leave` would silently restore nothing on a comp_time cancel.)*
  - **§3b idempotent** — no-op if a `reverse` for the requestId already exists.
  - **§3a expired lot is NOT resurrected** — skipped; the unrecoverable amount is surfaced in the result and stays queryable as `Σdeduct − Σreverse` (durable, beats a log line).
  - Restores `remaining_minutes` capped at the lot's `amount_minutes`, `exhausted→active`, `FOR UPDATE OF` the lots.
- **`cancelRequest`** — loosened to allow `approved→cancelled` **only for `request_type='leave'`** (every other type stays pending-only, unchanged); calls the reverse in the same txn; §3c authority reuses the existing own-or-`canAccessOtherUsers` gate.

## Tests
- **7 reverse-logic unit tests** (mock `trx`): restore + positive reverse-event, **source_type mirroring** (comp_time), §3b idempotency, §3a expired-skip, mixed multi-lot, headroom cap, no-op-when-nothing-deducted.
- **3 migration source-guards** (DB-free, C1 pattern).
- **91 existing index.cjs unit tests stay green**; `node -c` + `tsc` clean.

## Coordination + remaining verification
- **Touches the #1 family's ledger** (`attendance_leave_balance_events` + lot rows) — additive only (no change to grant/accrual/expiry); built worktree-isolated. Flagged in design-lock §4.
- The **real-DB route-integration matrix (§5)** — create→approve→deduct→cancel→reverse through the live route — is the **staging-smoke step** (§6 governance), not in this PR (it needs a booted-server harness I can't validate locally); the reverse *logic* + the migration are unit/guard-verified here.

### DIFF
--- a/docs/development/multitable-button-side-effect-1-send-notification-designlock-20260616.md
+++ b/docs/development/multitable-button-side-effect-1-send-notification-designlock-20260616.md
@@ -1,0 +1,84 @@
+# 多维表 Button 首个有副作用动作 (B1-S1 = send_notification) — 设计锁 — 2026-06-16
+
+> Status: **DESIGN-LOCK(待评审,不实现)**。owner opt-in 2026-06-16:先锁定第一个有副作用 button action 的合同,审过再实现。
+>
+> 命名:称 **B1-S1**(side-effecting action #1)。**不复用 "B1-d"**(B1-d 已是 §8 验证 closeout 行)。本动作是 B1-S0 §3.3 升级序里 `record_click → **send_notification** → update_record/send_webhook` 的第二步。
+>
+> 价值:把按钮从"可点"变成"有用",并把两条**刻意 deferred 的安全承诺**一起落地——confirm 从 parsed→**强制**;durable audit 从"等首个有副作用动作"→**强制记录**。`record_click` 继续保持 inert/logger-only,不混淆。send_notification 成为后续动作(update_record/send_webhook)扩展的**可复用合同样板**。
+>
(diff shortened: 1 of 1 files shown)


zensgit:
Created from the wrong branch (canonical checkout, not the #7 feature branch) — superseded by the correct PR off feat/attendance-leave-cancellation-reversal-20260622.

PR #3046, opened 2026-06-22
feat(integration): FOS-2b — field-option-sync sync-mode merge runtime (no-delete / no-silent-write / replace zero-drift)

zensgit:
## What (FOS-2b, owner-gated)

Makes the FOS-1 enums real runtime (on FOS-2b-pre's read-only `getObjectField`). **Owner locks, all tested: no-delete, no-silent-write, replace zero-drift.**

- Kernel `syncFieldOptions` + `mergeFieldOptions`: `append` (union, nothing removed), `disable_missing` (source + current-not-in-source **disabled, never deleted**), conflictPolicy `keep_existing` (preserve current human fields) / `update_from_source` (take source) / `manual_confirm` (**patches nothing**, returns values-free `held` counts). **replace + update_from_source = FOS-2 fast path (no read/merge) → stock-prep byte-identical.** Non-default mode without `readCurrentOptions` → fail-closed.
- Generic route: passes preset syncMode/conflictPolicy + a `getObjectField`-backed reader; returns held evidence for manual_confirm.
- Stock-prep wrapper: unchanged (defaults → fast path).

## Verification (the locks)
LOCK1 no-delete, LOCK2 no-silent-write (0 patch calls + values-free held), LOCK3 replace zero-drift (default does NOT read; raw set patched byte-identical) + keep_existing/update_from_source + disable_missing integration + fail-closed guard. **Zero-drift proven non-vacuous** (break kernel patch → stock-prep test fails). All affected plugin suites green.

## Boundary
Metadata-only; getObjectField read-only; no business-row/external/K3/production write; no migration; admin-gated; values-free. New modes become end-to-end reachable when a non-default preset is added (**FOS-4**).

### DIFF
--- a/plugins/plugin-integration-core/__tests__/field-option-sync-runtime.test.cjs
+++ b/plugins/plugin-integration-core/__tests__/field-option-sync-runtime.test.cjs
@@ -7,7 +7,7 @@ const assert = require('node:assert/strict')
 const fs = require('node:fs')
 const path = require('node:path')
 
-const { syncFieldOptions } = require(path.join(__dirname, '..', 'lib', 'field-option-sync-runtime.cjs'))
+const { syncFieldOptions, mergeFieldOptions } = require(path.join(__dirname, '..', 'lib', 'field-option-sync-runtime.cjs'))
 
 // ZERO-DRIFT STRUCTURAL LOCK: the stock-prep option-sync module MUST route through this kernel and
 // MUST NOT carry its own patch loop / patch call (else it would be a parallel copy that could drift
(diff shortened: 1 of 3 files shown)


zensgit:
Review note (P2, coverage — deferred to FOS-4, no lock impact): the generic route's non-default merge/manual_confirm path (the `readCurrentOptions` closure that reads `getObjectField().property.options`) is currently **HTTP-dormant** — the only catalog preset is default-mode (replace/update_from_source → fast path), and any non-default preset throws `FIELD_OPTION_SYNC_PRESET_NO_READINESS` until FOS-4 wires its readiness binding. So the real `getObjectField`-backed closure is kernel-tested-with-fixtures but not yet wire-tested (the documented wire-vs-fixture blind spot). **FOS-4 action:** when introducing the first non-default preset, add an http-routes test that drives it through the real `getObjectField` closure and asserts the merged patch + values-free `held` round-trip the wire. Dormant = strictly safe today; all 3 owner locks (no-delete / no-silent-write / replace zero-drift) hold.

PR #3048, opened 2026-06-22
docs(attendance): plan remaining benchmark development

zensgit:
## Summary

- Add a docs-only execution plan for the remaining attendance benchmark development.
- Ground it on current `origin/main` and live PR checks as of 2026-06-22.
- Separate completed work (#1 annual leave, #3 email notification channel, #7 design-lock) from the actual remaining build waves.

## Plan locked in this doc

- Wave 0: finish #3044 leave cancellation / balance reversal merge window.
- Wave 1: build #4 structured makeup-punch and #2 unified employee self-service in parallel.
- Wave 2: build #5 report-tiering, then #6 pending-leave/team-availability.
- Wave 3: build #8 night-shift / cross-midnight deepening as the final hot-core arc.

## Notes

- This does not replace `attendance-dingtalk-benchmark-target-and-tracker-20260601.md`; the tracker remains the status source of truth.
- It does not flip any tracker rows or claim #3044 merged.
- It explicitly marks #3 as complete via #3018 + #3033, so we do not reopen notification-channel work by accident.

## Verification

- `git diff --check`
- live PR state checked with `gh pr view` for #3014, #3018, #3033, #3034, #3044

PR #3055, opened 2026-06-22
docs(attendance): #5 report-tiering — severe/absence-late first-class rules (design-lock RT-0, PROPOSED)

zensgit:
**PROPOSED** design-lock for **#5 report-tiering** (Wave 2 / RT-0 from the remainder plan #3048). **Design-lock only — no code; your 拍板 on §3 before build.**

## Why I produced this now
#4/#2 are your Codex session's lane (Wave 1), so **my lane is the serial #5→#6→#8** — and #5 is unambiguously non-Codex. A design-lock is collision-safe to produce ahead of the serial Wave-2 build, and it's the RT-0 slice the plan calls for. (If you'd rather I wait or take a different arc, redirect — it's cheap.)

## What #5 does
`lateMinutes` is already computed truthfully from the rule's `lateGraceMinutes`. Severe/absence-late are just **tiers of the same lateness**, but today they're **meta-passthrough** (`index.cjs ~4574` reads `summary.severeLateCount` from record meta). v1 promotes them to **first-class self-calc**:
- add `severeLateThresholdMinutes` / `absenceLateThresholdMinutes` to the **per-group rule** (beside `lateGraceMinutes`),
- **tier `lateMinutes`** at compute (≥ severe → severe; ≥ absence → absence-late), like `late_count`,
- **persist** the tier counts onto the summary so exports round-trip; the existing meta read becomes a **legacy fallback** (no history restatement).

## §3 — your 拍板
- **3a defaults** — severe **30 min** / absence-late **60 min** over grace (per-group configurable).
- **3b persist vs read** — recommend **persist** (the v3 "self-calc + 落库 + round-trip" bar).
- **3c legacy** — **meta fallback**, don't recompute closed periods.
- **3d scope** — defer the v3's 出勤口径 (free-hours/unscheduled-not-counted) + 公式函数 (TEXT/FILTER/REGEX) to a follow-up; v1 = late tiers only.

## Guardrails
The **wire-vs-fixture round-trip landmine** (assert the tier survives the real export projection, not a hand-built record) + enum-strict normalizer (#1776) + **serial with #6/#8** (shared record-compute; never land concurrently). RT-1..RT-4 slices mirror the plan. MetaSheet 口径; no competitor names.


zensgit:
**Strictness reconciled (addressing the recurring §2/§6 review finding).** Pushed as `9d9a3a294` on this branch — note any review snapshot under `/tmp/metasheet2-review-3055/` predates it, so re-pull to see it.

§2 (line 20) and §6 now state: a **supplied** non-integer / negative / out-of-order threshold → explicit **400 reject** (the #1776 rule); an **omitted** field **resolves to existing/default** (whole-rule replace, no UI yet); the `absence ≥ severe ≥ grace` ordering is enforced on the **resolved** rule. This matches the shipped RT-1a impl (optional+default, reject only supplied-invalid) per your review — rather than forcing the fields required on write.

Merge order unchanged and respected: **#3055 (your §3 拍板) → #3069 → rebase #3070 onto main → full gate → merge.** Nothing is merged (all four PRs OPEN).

PR #3071, opened 2026-06-23
docs(attendance): design-lock for cross-midnight overtime windows (#8 NS-0)

zensgit:
## #8 NS-0 — design-lock: cross-midnight overtime windows

Docs-only design-lock for the last attendance arc on my lane (plan #3048, Wave 3, heaviest, serial-last). PROPOSED — owner 拍板s §3 before any build.

**The grounding sharpened the scope.** A read of the code (overnight handling, the reject, the time math, the tests) shows overnight **shifts already work** — `computeMetrics` moves `shiftEndAt` to the next day and an integration test covers 22:00→06:00. The *only* cross-midnight hard-reject is `validateOvertimeSegmentationWindow`'s `OVERTIME_CROSS_MIDNIGHT_UNSUPPORTED` (~10154, test-locked) for overtime **windows** that span a day boundary. So #8 is narrowly: make cross-midnight overtime windows computable, not "add overnight support."

**The proposal:** boundary-aware segmentation — split a window at local midnight, bucket each sub-span by its own date — with a one-midnight cap, and the plan's NS-2 discipline of **lifting the reject only after** a reverse/adversarial matrix proves the split correct + idempotent (don't unguard on a half-proven split). §3 leaves the owner the calls on the midnight cap, boundary attribution, and lift-timing.

Serial with #5/#6. MetaSheet 口径; no competitor names.


zensgit:
**§P2 addressed** in `03ee66440`: §2 + §5 (NS-1) now pin the reject-stays mechanism — the one-midnight split lands BEHIND the existing reject (a separate `splitOvertimeSegmentationWindowAtMidnight` helper not on the default path, OR an `allowCrossMidnight` option defaulting false), since `maybeBuildOvertimeSegmentationSnapshot` calls the validator directly. The route keeps returning `OVERTIME_CROSS_MIDNIGHT_UNSUPPORTED` until NS-3, proven by a route-level test (NS-1/NS-2 assert one-midnight STILL rejects). Consistent with §1.2/§3c/§4/NS-3/§7.

PR #3085, opened 2026-06-23
feat(multitable): BS-3 scoped restore batch-execute — the multi-record write (PARTIAL) [HELD for review]

zensgit:
> **HELD for your review — do not auto-merge.** The gated write slice (BS-3), built to exactly the ratified boundary. Merge gate retained per "merge 前保留审阅闸口".

The multi-record fan-out **write**. Consumes the BS-1 scoped identity (#3073) + BS-2 preview (#3078).

## Two-layer rejection model (advisor)
- **DIFF-LEVEL** — a record missing / no-revision / delete / schema-drifted / *restorable diff changed* / empty since preview → excluded from `contributing` → the recomputed `scopeHash` diverges → **WHOLE-BATCH 409**, re-preview. No per-record noop-skip (that would execute a smaller set than the token authorized).
- **WRITE-LEVEL** — row-deny appeared / version conflict / field-forbidden during the per-record fan-out → **PARTIAL skip + report** (machine-distinguishable `denied`/`conflict`/`forbidden`). The split's proof case: data **edited-then-reverted** → diff identical (scopeHash passes) but version moved → the in-transaction CAS skips it.

## Boundaries (exactly as ratified)
- **verify-before-write** — `scopeHash` verified before ANY 2xx, **including an all-noop set** (a forged token over an already-restored set 4xxes, never a 200 noop short-circuit).
- **write-set ≡ hashed-set** — one canonical sorted/deduped `recordIds` source; each diff computed **once**, fed to both the hash and `patchRecords`; `expectedVersion` never enters the hash.
- **per-record fresh gates** — row-deny re-check + **authoritative in-transaction `expectedVersion` CAS under `SELECT FOR UPDATE`** + field-write gate. The concurrency guard the advisor flagged as the blocker — confirmed authoritative (record-write-service.ts L711), not the TOCTOU outside pre-check.
- **PARTIAL default** — per-record `patchRecords` (its own txn) so one conflict/forbidden/denied skips only that record.
- **forward-only** (`source='restore'`, no destructive delete = T8); **bounded ≤100** fail-closed. all-or-nothing deferred to BS-3.1 (a named-need follow-up; `patchRecords`' single-call path is already atomic).

## Verification
**11 real-DB goldens** (run locally on `metasheet_test`): real preview→execute happy path (forward revisions, all rows written) · **[P2]** forged token over all-noop → 4xx · **diff-level** restorable-diff-changed → whole-batch 409 nothing written · **write-level conflict** (edited-then-reverted) → PARTIAL · **write-level denied** (1 of 3) → PARTIAL · **BS-7 keystone** (execute a subset → 409) · **D6** (single-record token → 409) · **actor binding** → 409 · bounded 400 · expectedVersions fail-closed 400.
**Mutation-proven** (3 new surfaces, Python-revert, residue 0): noop-short-circuit-before-verify → [P2] fails; strip the per-record `expectedVersion` → conflict fails; disable the fresh row-deny gate → denied fails. Full restore surface **56/56** across 6 files (no regression); `tsc` clean; control bytes 0; added to the plugin-tests allowlist.

## How I'd like the gate handled
Review at your pace — I can post a self-review summary, run `/ultrareview`, or you take it. I won't merge until you say.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -214,6 +214,7 @@ jobs:
             tests/integration/multitable-restore-execute-realdb.test.ts \
             tests/integration/multitable-restore-per-field-realdb.test.ts \
             tests/integration/multitable-restore-batch-preview-realdb.test.ts \
+            tests/integration/multitable-restore-batch-execute-realdb.test.ts \
             tests/integration/yjs-scalar-seed-realdb.test.ts \
             tests/integration/yjs-scalar-flush-realdb.test.ts \
             tests/integration/yjs-scalar-strdate-migration-realdb.test.ts \
--- a/packages/core-backend/src/multitable/restore-preview-identity.ts
+++ b/packages/core-backend/src/multitable/restore-preview-identity.ts
@@ -114,15 +114,20 @@ export interface ScopedRestorePreviewIdentityClaims {
 }
 
 /**
- * Canonical, ORDER-INVARIANT scope hash: binds the EXACT record set AND each record's per-record changesHash.
- * Sorted by recordId, each entry serialized as `[recordId, changesHash]` (D4). A record added/removed changes the
- * set → a different hash; a changed per-record diff → a different hash. This is what makes BS-7 hold — a scoped
- * identity for {A,B,C} cannot execute {A,B} or {A,B,C,D}, because the execute re-hashes the ACTUAL set and diverges.
+ * Canonical, ORDER-INVARIANT scope hash: binds the EXACT record set, each record's per-record changesHash, AND
+ * each record's per-record expected `version`. Sorted by recordId, each entry serialized as
+ * `[recordId, changesHash, version]` (D4). A record added/removed changes the set → a different hash; a changed
+ * per-record diff → a different hash; a changed per-record version → a different hash. This is what makes BS-7
+ * hold (a scoped identity for {A,B,C} cannot execute {A,B} or {A,B,C,D}) AND what binds the per-record
+ * optimistic-concurrency anchor: the version is FOLDED IN here at mint (BS-2) and re-folded at verify (BS-3) from
+ * the CLIENT-SUBMITTED expectedVersion — never trusted as free side-input — so a client that submits the current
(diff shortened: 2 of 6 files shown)


zensgit:
## Both [P1] findings fixed (force-pushed `bdaee91c6`, still held)

Verified both against the code (not reflex-agreed) — both real, both fixed:

**[P1 #1] layer-3 per-subject write gate** — `allowed` is a *read* mask (visible ∧ field_permissions-visible); patchRecords enforces only *field-definition* readonly, not per-subject `field_permissions.read_only`. A visible-but-readOnly field was hashed + written. **Fix:** gate each record's diff at the write fan-out with the *exact* legacy `/restore` derive (`fieldPermissions` from `buildRecordPatchContext`) → `skip:forbidden` (PARTIAL, whole-record). Hash + BS-2 unchanged (scopeHash stays read-masked, preview↔execute consistent).

**[P1 #2] unbound `expectedVersions`** — the identity didn't bind per-record version, so submitting the *current* version (after edit-then-revert) slipped the CAS. **Fix:** fold version into `hashScope` → `[recordId, changesHash, version]`; BS-2 binds the preview-time version + returns `previewVersion`; BS-3 folds the **client-submitted** value (not a fresh read). Honest client → hash matches → CAS fires; tamper → hash diverges → 409.

**Verification:** 60/60 restore-surface goldens. 2 new real-DB goldens (visible+readOnly → forbidden-skip PARTIAL; version-tamper `{B:4}` → 409, no write) + a BS-1 unit version-bind golden — both **mutation-proven** (disable the gate → forbidden golden fails; fold a constant version → tamper golden fails). The "authoritative CAS" claim is now true (was client-bypassable before #2).

## Separate [P1] — flagging, not fixing here
The **shipped** single-record `restore-execute` (T6-2, ~L7886) has the **same** layer-3 write gap — it also masks with `allowed` (read) + relies on patchRecords. I did **not** fold a fix for shipped code into this held PR (scope creep). Your call: separate follow-up PR, or fold in. (The single *preview* offering readonly fields is a read-only misleading-preview nit, not a write vuln.)

Still held for your merge decision.

PR #3056, opened 2026-06-22
docs(attendance): #6 pending-leave overlay + team availability (design-lock TA-0, PROPOSED)

zensgit:
**PROPOSED** design-lock for **#6 pending-leave overlay + team availability** (Wave 2 / TA-0 from plan #3048, **serial after #5**). **Design-lock only — no code; your 拍板 on §3 before build.**

## What it does
Today only **approved** leave shows in the calendar (`resolveEffectiveCalendar`, `index.cjs ~11365 status='approved'`); pending is invisible and there's no team-availability surface. v1:
- an **opt-in `includePending`** overlay adds `kind:'pending_leave'` items to `resolveEffectiveCalendar` — **default false**, so existing callers are byte-identical;
- **display-only**: the write/records path ignores `pending_leave` (mirrors the taint-skip discipline — a provisional item must never reach `attendance_records`);
- a **scope-gated team-availability** read endpoint (group/dept/date → each member's state).

**The hard line:** visualization + provisional overlay only — never alters a final record, deduction (#1/#7), or report number.

## §3 — your 拍板
- **3a capacity** — rec **show-only/tentative** (don't subtract unconfirmed leave from headcount).
- **3b visibility** — rec **owner/sub-owner/admin** (reuse the group-scope gate; not peers).
- **3c marking** — distinct provisional style + 口径 tooltip (the "precise definitions" humanization lesson).
- **3d granularity** — faithful to whatever the leave request already models.

## Guardrails
`includePending` default-false (zero perturbation); **serial with #5/#8** (shared resolver — never land concurrently; runs after #5); a **real-wire guard** that pending items never reach `attendance_records`; reject→disappears / approve→formal, records unchanged. TA-1..TA-4 mirror the plan. MetaSheet 口径; no competitor names.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3087, opened 2026-06-23
feat(attendance): team-availability read endpoint (#6 TA-2)

zensgit:
## #6 TA-2 — team-availability read endpoint

Per design-lock #3056 §3 (owner 拍板, recommended). `GET /api/attendance/team-availability?groupId=&from=&to=` aggregates each group member's per-date state from `resolveEffectiveCalendar(includePending:true)` (TA-1, on main).

### Honours §3 exactly (product semantics)
- **§3e group-only** — `groupId` required (`400` otherwise); range ≤ 366 days; `members × days ≤ 6000` guard.
- **§3b scope-gated** — `admin OR userManagesAttendanceGroup` (owner/sub-owner) → else `403`. Reuses the shipped group-scope helpers; `RBAC_BYPASS` short-circuits for tests.
- **§3a pending ≠ subtraction** — per-date summary keeps pending in a **separate tentative bucket**, and `availableFormal = scheduled + pendingLeaveTentative`. A pending-leave member is **still counted in the formal available headcount** (their leave isn't approved); only approved leave subtracts.
- per-member `state` precedence: `approved_leave` (personal_leave overlay) > `pending_leave` overlay > `unscheduled` (`isUserScheduledForDate=false`) > `scheduled` (`isWorkingDay`) > `rest`. `pending_leave` is a **distinct** state — never folded into approved.
- **§3c** provisional UI marking ("待审批，未生效") is **TA-3**; TA-2 just returns `pending_leave` distinctly so the UI can mark it.

### Test (real-DB)
A group with a pending-leave member + a scheduled member → `m1.state=pending_leave`, `summary.pendingLeaveTentative=1`, `approvedLeave=0`, `availableFormal = scheduled + 1` (§3a); missing `groupId` → `400` (§3e); an outsider (non-admin/non-manager) → `403` (§3b).

### Scope
Read-only aggregation; no change to records/approved-leave. v1 group-only; dept deferred (§3e). TA-3 (UI) next. Serial vs #8. (Advisor was overloaded at review time; relying on the real-DB gate + self-review.)


zensgit:
**§P2 addressed** in `7e75e9e6a` (this branch — the `/tmp/metasheet2-review-3087` snapshot predates it; re-pull to see it).

In the team-availability handler, **after** the scope gate (so an unauthorized caller still gets 403 first — no existence leak), I now verify the group:
```
const group = await loadAttendanceGroupByIdOrCode(db, orgId, groupId)
if (!group) { res.status(404)... 'Attendance group not found'; return }
```
So a valid-UUID but non-existent/deleted group → **404**, not the misleading 200 with members:0. Regression test added: admin + a random valid-UUID groupId → **404**. CI is re-running the full gate with it; PR is BEHIND.

Per your guidance I'll then update/rebase → fresh-green → merge via the stable path.

PR #3097, opened 2026-06-23
feat(multitable): reconcile orphaned AI bulk-fill jobs after hard restart (B-4 BJ-5 follow-up)

zensgit:
## What

Narrow follow-up to **#3075** closing the one non-blocking gap from its review: **stale `queued`/`running` job reconciliation after a hard restart**.

`runJob` is the in-process worker. `markErroredIfRunning` only covers the in-process exception path — its own comment notes a **hard process restart is a B-4 follow-up**. On restart the queue + in-process plan registry are gone, so a job left `queued`/`running` has no live worker and can never progress, yet it stays "active" (the BJ-7 partial unique index counts `queued/running/suspended`) and **blocks a fresh start for that `(actor, sheet, field)` until `expires_at` GC**.

## Change (3 files, +96)

- **`reconcileOrphanedBulkJobs(query, { staleAfterMs })`** — flips stale `queued`/`running` headers → `errored` (header-only, mirroring `markErroredIfRunning`; already-`generated` rows stay committable). **Excludes `suspended`** (intentionally paused for review). An **age guard** avoids racing a just-claimed job in a multi-instance deploy; the `UPDATE` is idempotent + race-safe, so **no leader election** is needed.
- **Startup wiring** (`index.ts`) — best-effort, mirroring the ledger-retention scheduler. Degrades gracefully on a DB error (verified: `server-lifecycle.test` exercises it, logs degraded-mode, and the server continues).
- **Hard-restart golden** (real-DB integration) — a 5-min-stale `running` **and** a stale `queued` job → `errored`; a just-updated `running` job and a `suspended` job are **untouched**.

## Validation
- `tsc --noEmit` passes; integration file parses + collects (the golden runs under `DATABASE_URL` in CI).
- Full `core-backend` **unit suite: 298 files / 3896 pass**, including `server-lifecycle` which runs the new startup path and degrades cleanly.

Does **not** touch any #3075 code — purely additive.

### DIFF
--- a/packages/core-backend/src/index.ts
+++ b/packages/core-backend/src/index.ts
@@ -32,6 +32,8 @@ import type {
 } from './types/plugin'
 import type { User } from './auth/AuthService'
 import { poolManager } from './integration/db/connection-pool'
+import { reconcileOrphanedBulkJobs } from './services/ai-bulk-job-service'
+import type { AiUsageQueryFn } from './services/ai-usage-ledger'
 import { eventBus } from './integration/events/event-bus'
 import { initializeEventBusService } from './integration/events/event-bus-service'
 import { messageBus } from './integration/messaging/message-bus'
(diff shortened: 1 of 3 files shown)


zensgit:
Addressed in `ed64d986` — floored `reconcileOrphanedBulkJobs`'s stale window at 30s (`Math.max(30, …)`) as suggested. Purely defensive (prod default is 10min and provider timeout maxes at 60s); the floor just stops the exported helper from being misused into reconciling a just-claimed, still-live job. No other code changes — keeping this PR narrow.

PR #3110, opened 2026-06-24
feat(attendance): NS-2 adversarial matrix pinning NS-1 split (#8)

zensgit:
## #8 NS-2 — adversarial matrix pinning NS-1's split (reject UNCHANGED)

Per the owner: **NS-2 pins down NS-1's split logic**; **NS-3** (not this) lifts `OVERTIME_CROSS_MIDNIGHT_UNSUPPORTED`; **TA-4** stays a separate staging gate. Test-heavy + one pure primitive — **no route change**.

### Added
- **`bucketOvertimeSegmentationWindowAtMidnight({startAt, endAt, timeZone, classifyDate})`** — classifies each split sub-span by **its own** local date (workday/restday/holiday) and sums per bucket, so a restday portion after midnight counts as **restday** and never bleeds into the prior workday (§1.1/§3d). Pure, still behind the guard.

### Adversarial matrix (8)
same-day unchanged · 23:00→01:00 → 1h/1h by date · **restday-after-midnight buckets correctly** · holiday override · **§3b Shanghai Z-string buckets per LOCAL date** · reversed + multi-midnight keep the stable reject · **idempotent** (byte-identical re-run, total == window minutes → no boundary double-count) · **guard test**: `maybeBuildOvertimeSegmentationSnapshot` still throws `422` for one-midnight (NS-2 did not open the route).

### Notes
Local 21/21, tsc 0. Reuses NS-1's ICU-robust split (no new exact-midnight `getZonedParts`). **NS-3** lifts the reject + wires the bucketing into the snapshot over real DB; **TA-4** is the staging smoke. Neither is in this PR.

---
**Design-lock reconciled (same PR):** moved `real-DB integration over the actual segmentation snapshot` from NS-2 → **NS-3**, since the snapshot rejects cross-midnight until NS-3 — so NS-2 proves split+bucketing at the pure-primitive level and the only real-DB-provable thing now is the preserved guard. Two NS-3 watch-items recorded in the doc: replay-through-persistence idempotency, and aligning the bucketing `classifyDate` contract with `resolveOvertimeDayTypeFromEffectiveCalendarItem` (one day-type model).


zensgit:
**§P2/§P3 addressed** in this branch (the `/tmp/ms2-ns2` review snapshot predates it):
- §P2: flipped bucketing precedence to **workday-wins-over-holiday**, matching `resolveOvertimeDayTypeFromEffectiveCalendarItem` (调休 makeup-workday = workday). Holiday test now uses a true holiday; added a makeup-workday precedence test.
- §P3: retitled §6 → **Verification plan — NS-3/NS-4** with a note that the real-DB snapshot round-trip is NS-3.
Local 22/22, tsc 0, guard unchanged. Updating to main → fresh-green → auto-merge per your note.

PR #3116, opened 2026-06-24
feat(attendance): NS-3 lift cross-midnight reject + wire bucketing into the snapshot (#8)

zensgit:
## #8 NS-3 — lift the cross-midnight reject + wire bucketing into the snapshot

Per the owner's go (3 items): the route now **accepts** a one-local-midnight overtime window; reversed/multi-midnight keep the stable reject; NS-2 bucketing is wired into the **actual snapshot**; real-DB + replay matrix added; the old test-locks flip reject→accept. **NS-4/TA-4 staging are not in this PR.**

### Build (advisor recipe)
- **`validateOvertimeSegmentationWindow`** — the tz-aware split is now the **single source** of cross-midnight detection (the old UTC/literal compare is gone, so a Z-string sharing a UTC date but crossing local midnight is no longer masked). One local midnight → `ok` + per-date spans; reversed/multi-midnight → stable reject. Removed the orphaned `extractOvertimeSegmentationDateOnly`.
- **`maybeBuildOvertimeSegmentationSnapshot`** — loads `workDate` + the next local date (and the work timezone) once; **same-local-date → the UNCHANGED `buildOvertimeSegmentationSnapshot` (byte-identical)**; cross-midnight → the new builder.
- **`buildCrossMidnightOvertimeSegmentationSnapshot`** — rule-normalize the total **once** (identical to same-day), then **partition** across per-date window weights — integer, **last date gets the remainder** so `Σ buckets === total` (conservation + replay-determinism). Each date bucketed by its **own** canonical day-type (workday wins over a holiday layer — 调休). **Never** re-applies the rule per sub-span. **Fails closed** (`OVERTIME_SEGMENTATION_UNRESOLVED`) if any spanned date's calendar item is missing.

### Tests
- **Unit 25/25** — partition/rule-once/last-remainder/conservation, holiday-vs-makeup-workday precedence, fail-closed, and the flipped accept locks (`:144` + the NS-1 default).
- **Real-DB + replay matrix** — a one-midnight overtime request is accepted, bucketed per own date (after-midnight holiday portion = 60), conserved; corrupt-then-approve recomputes byte-identical; multi-midnight → 422. Learns the work tz so the window crosses **local** midnight regardless of the org default.

### Watch-items (NS-3 doc reconciliation, recorded for follow-up)
The `classifyDate` contract is converged onto `resolveOvertimeDayTypeFromEffectiveCalendarItem` (one day-type model). Same-day byte-identical is guaranteed by routing same-day to the untouched builder + the existing O2/O3/O4 real-DB tests.

### DIFF
--- a/packages/core-backend/tests/integration/attendance-plugin.test.ts
+++ b/packages/core-backend/tests/integration/attendance-plugin.test.ts
@@ -5879,6 +5879,96 @@ attendanceIntegrationDescribe(
     }
   })
 
+  it('加班三段 NS-3 — a one-midnight overtime window is ACCEPTED, bucketed per OWN date, conserved, replay-idempotent; multi-midnight rejects (real DB)', async () => {
+    if (!baseUrl) return
+    const dbUrl = process.env.ATTENDANCE_TEST_DATABASE_URL || process.env.DATABASE_URL
+    if (!dbUrl) return
+    const runSuffix = Date.now().toString(36)
+    const userId = `attendance-ns3-${runSuffix}`
+    const previousRbacBypass = process.env.RBAC_BYPASS
+    const pool = new Pool({ connectionString: dbUrl })
+    const createdRequestIds: string[] = []
+    const workDate = '2037-06-12' // June → no DST transition in any tz, so the local-midnight offset is stable
+    const nextDate = '2037-06-13'
+    const farDate = '2037-06-14'
+    const holidayId = randomUUID()
+    let overtimeRuleId: string | undefined
+    // UTC ISO for a local wall-clock in tz (DST-safe here because the dates avoid transitions).
+    const zonedUtcIso = (dateStr: string, timeStr: string, tz: string): string => {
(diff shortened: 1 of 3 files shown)


zensgit:
**§P1 + §P2 addressed** (this branch; the `/tmp/ms2-ns3` snapshot predates it):
- **§P1 workDate anchoring** — after the split, `validateOvertimeSegmentationWindow` now requires `spans[0].date === workDate`. A one-midnight window naturally spans D→D+1; a window whose first local date isn't workDate (e.g. both ends on D+1) keeps the stable `OVERTIME_CROSS_MIDNIGHT_UNSUPPORTED` (restoring the pre-NS-3 anchoring). Added a unit regression (workDate=D, times on D+1 → reject).
- **§P2 test hygiene** — the NS-3 real-DB test now snapshots+restores settings in `finally` (global toggle) and deletes `attendance_records` for the synthetic user/dates.
Unit 25/25, tsc 0. Core NS-3 behavior unchanged.

PR #3121, opened 2026-06-24
feat(approval): edit approval nodes in complex graphs

zensgit:
## Summary

- add the G-5 complex-graph approval-node editor for preserved approval graphs
- edit approval-node assignee source, approval mode, empty-assignee policy, and merge-with-requester without changing graph topology
- fail closed on complex approval-node configs the structured editor cannot represent, instead of flattening them
- update the amount-tier and complex-graph design locks so Gate A is marked satisfied while amount-tier presets remain unbuilt

## Scope

- UI/docs/guard only: no backend runtime, migrations, or approval-result write-back
- implements the approval-node editor prerequisite for amount-tier presets; does not add the amount-tier preset payloads
- preserves all node keys, all edges, and untouched condition/parallel/cc/approval nodes byte-identical

## Verification

- pnpm --filter @metasheet/web exec vitest run approval-common-template-presets approvalTemplateAuthoring approval-detail-field approval-template-authoring-detail approval-template-authoring-graph-preserve approval-template-authoring-condition-edit approval-template-authoring-parallel-edit approval-template-authoring-cc-edit approval-template-authoring-approval-node-edit --reporter=dot (166/166)
- pnpm --filter @metasheet/web exec vue-tsc -b
- pnpm --filter @metasheet/web exec eslint src/approvals/approvalNodeEdit.ts src/approvals/templateAuthoring.ts src/views/approval/TemplateAuthoringView.vue tests/approvalTemplateAuthoring.spec.ts tests/approval-template-authoring-approval-node-edit.test.ts
- node -e "const fs=require('fs'); const yaml=require('js-yaml'); yaml.load(fs.readFileSync('.github/workflows/approval-web-guard.yml','utf8')); console.log('yaml ok')"
- git diff --check

### DIFF
--- a/.github/workflows/approval-web-guard.yml
+++ b/.github/workflows/approval-web-guard.yml
@@ -34,6 +34,7 @@ on:
       - 'apps/web/tests/approval-template-authoring-condition-edit.test.ts'
       - 'apps/web/tests/approval-template-authoring-parallel-edit.test.ts'
       - 'apps/web/tests/approval-template-authoring-cc-edit.test.ts'
+      - 'apps/web/tests/approval-template-authoring-approval-node-edit.test.ts'
       - '.github/workflows/approval-web-guard.yml'
   push:
     branches: [main]
@@ -55,6 +56,7 @@ on:
       - 'apps/web/tests/approval-template-authoring-condition-edit.test.ts'
       - 'apps/web/tests/approval-template-authoring-parallel-edit.test.ts'
       - 'apps/web/tests/approval-template-authoring-cc-edit.test.ts'
+      - 'apps/web/tests/approval-template-authoring-approval-node-edit.test.ts'
       - '.github/workflows/approval-web-guard.yml'
 
 jobs:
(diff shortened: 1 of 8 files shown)


zensgit:
Closing as superseded — G-5 (complex-graph approval-node source editing) shipped via **#3124** (squash `597de701c`, on `main`).

This draft and #3124 are two codelines for the same capability; keeping both open risks divergence. #3124 is the landed path: approver-source editing as a fourth disjoint edit pass (topology + within-node config byte-identical), with a fail-closed shape-level allowlist (top-level + nested `assigneeSources`/`autoApprovalPolicy`/`fieldPermissions` against the backend's exact per-kind/per-field shapes) so unknown keys can't silently flatten on save, plus a mounted SFC wiring test. approval-web-guard 178/178.

If anything in this draft's approach is worth preserving (e.g. the rich directory picker for approval-node IDs, which #3124 deliberately scoped to a compact ID-typed control as a follow-up), let's fold it into a follow-up PR rather than reopen this line. Reopen only if #3124 needs to be reverted.

PR #3124, opened 2026-06-24
feat(approval): G-5 — complex-graph approval-node source editing (topology-preserving)

zensgit:
Makes an `approval` node inside a **preserved complex graph** editable for its approver **source** (`assigneeSources`) — the **#3114 Decision-4 prerequisite** the amount-tier preset runtime depends on ("admins must tune which role/person approves at a threshold").

## Changes
- **`approvalNodeEdit.ts`** (pure, no `.vue`): `applyApprovalNodeEditsToGraph` is the **fourth disjoint edit pass** after condition/parallel/cc. The `assigneeSources`-present guard + spread-original-first means an untouched node is **byte-identical by construction**, a **legacy** node (`assigneeType`/`assigneeIds`, no `assigneeSources`) is never seeded → cloned verbatim + read-only, and the edited node's own `approvalMode`/`emptyAssigneePolicy`/`autoApprovalPolicy` **survive**. `validateApprovalNodeEdits` mirrors the backend (`ApprovalProductService.ts:457-480`): a `form_field_user` source must reference a **top-level user field**.
- **View**: a compact per-kind source control on complex-graph approval nodes; condition/parallel/cc keep their editors; topology read-only.

## Scope (deliberately honest)
- **Approver-source editing only.** `approvalMode` / `emptyAssigneePolicy` are **preserved, not yet editable** (a later slice).
- View is a **compact ID-typed control**, not the rich directory picker the linear editor uses.
- Multi-source nodes: edit the **primary** source, preserve extras.

So **G-5 ≠ "approval-node editing done"** — it's the source-editing unlock #3114 needs, with the rest scoped to follow-ups.

## Tests
13 tests — topology-preservation, **within-node** config survival on the edited node, **legacy-node** round-trip byte-identical, **four-phase compose** (condition+parallel+cc+approval all land), `form_field_user`→top-level-user validation. Wired into `approval-web-guard` (**168/168**). vue-tsc + lint clean.

UI-only — no runtime/validator/type changes.

### DIFF
--- a/.github/workflows/approval-web-guard.yml
+++ b/.github/workflows/approval-web-guard.yml
@@ -20,6 +20,7 @@ on:
       - 'apps/web/src/approvals/conditionEdit.ts'
       - 'apps/web/src/approvals/parallelEdit.ts'
       - 'apps/web/src/approvals/ccEdit.ts'
+      - 'apps/web/src/approvals/approvalNodeEdit.ts'
       - 'apps/web/src/approvals/detailField.ts'
       - 'apps/web/src/approvals/commonTemplatePresets.ts'
       - 'apps/web/src/approvals/templateAuthoring.ts'
@@ -34,6 +35,7 @@ on:
       - 'apps/web/tests/approval-template-authoring-condition-edit.test.ts'
       - 'apps/web/tests/approval-template-authoring-parallel-edit.test.ts'
       - 'apps/web/tests/approval-template-authoring-cc-edit.test.ts'
+      - 'apps/web/tests/approval-template-authoring-approval-node-edit.test.ts'
       - '.github/workflows/approval-web-guard.yml'
   push:
     branches: [main]
@@ -41,6 +43,7 @@ on:
       - 'apps/web/src/approvals/conditionEdit.ts'
       - 'apps/web/src/approvals/parallelEdit.ts'
       - 'apps/web/src/approvals/ccEdit.ts'
+      - 'apps/web/src/approvals/approvalNodeEdit.ts'
       - 'apps/web/src/approvals/detailField.ts'
       - 'apps/web/src/approvals/commonTemplatePresets.ts'
       - 'apps/web/src/approvals/templateAuthoring.ts'
@@ -55,6 +58,7 @@ on:
       - 'apps/web/tests/approval-template-authoring-condition-edit.test.ts'
       - 'apps/web/tests/approval-template-authoring-parallel-edit.test.ts'
       - 'apps/web/tests/approval-template-authoring-cc-edit.test.ts'
+      - 'apps/web/tests/approval-template-authoring-approval-node-edit.test.ts'
       - '.github/workflows/approval-web-guard.yml'
 
 jobs:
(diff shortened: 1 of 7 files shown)


zensgit:
Addressed in `46cbada95` — both P1s.

**[P1] complex approval-node config fail-closed.** `unsupportedTemplateAuthoringReason` no longer short-circuits `return null` for complex graphs without checking approval-node config keys. A complex approval node carrying a key the backend (`normalizeApprovalGraph` :899-911) won't preserve → unsupported (`graphReadOnly=false` ⇒ editor not rendered, save disabled), never silently flattened. The complex allowlist is **separate** from the linear one (the linear path rebuilds via `buildStepConfig`, which can't preserve `fieldPermissions` — sharing would flatten it there). Verified: the existing linear "fails closed on fieldPermissions" test still passes, and a new test proves the **complex** path *allows* `fieldPermissions`. Note: this also closes a latent G-1 gap, not just G-5.

**[P1] mounted SFC wiring test.** Added to `approvalTemplateAuthoring.spec.ts`: load a preserved complex graph, change `approval_1`'s source via the real `approval-node-source-kind` control, save, assert the `updateTemplate` payload's `assigneeSources` changed while its own `approvalMode`/`emptyAssigneePolicy`/`autoApprovalPolicy` + the cc node + all edges stay byte-identical. Plus a legacy-node-shows-no-editor assertion. (It also surfaced that an empty `static_role` correctly **blocks** save via the validation preview — validation is wired end-to-end.)

**Non-blocking:** `validateApprovalNodeEdits` now requires integer levels; the "mirrors backend" claim is narrowed (the manager-chain level cap is server-side). On #3121 — agreed it's the overlapping G-5 draft; will close it once #3124 merges so the two codelines don't diverge.

+5 tests (3 allowlist + 2 mounted wiring); approval-web-guard **173/173**, vue-tsc + lint clean.

zensgit:
Addressed in `fd7adb497` — round-2 P1 (nested unknown keys).

**Shape-level fail-closed.** The top-level allowlist let nested unknown keys through; the backend rebuilds the nested objects from fixed fields too and silently drops the rest. `complexApprovalConfigHasBackendDrop` now checks the full shape against the exact backend allowlists:
- `assigneeSources[]` per kind (`ApprovalProductService.ts:408-453`) — `{kind}` + kind-specific only (and an unknown KIND is flagged)
- `autoApprovalPolicy` (`:371-376`) — the 4 fields
- `fieldPermissions[]` (`:786-799`) — `{fieldId, access}`

All three nested shapes bottom out in primitives / string-arrays — **there is no fourth object level** inside approval-node config — so the 2-level check is complete for this surface. +5 tests (3 nested-drop cases + unknown-kind + a fully-known-nested *allowed* case proving I didn't over-strict). approval-web-guard **178/178**, vue-tsc + lint clean.

**Named boundary (NOT fixed here — pre-existing, outside G-5 scope).** The identical unknown-key drop exists for **cc / condition / parallel** nodes: the backend rebuilds their configs from fixed shapes too (e.g. cc → `{targetType, targetIds}`, `:920-923`) and drops anything else, while the FE spread-preserves it. This predates G-5 (G-2/3/4) and their nested shapes are deeper (`branches[].rules[]`, `branches[]`), so it's sized as a sibling shape-check pass — tracked in the TODO out-of-scope as a declared boundary rather than silently left.

(Pre-empting one likely question: the backend also *transforms values* — trims `userIds`/`assigneeIds`, collapses `fieldPermissions: []` → undefined. Those aren't round-trip breaks: a loaded template was already backend-normalized on its prior save, so trimming is idempotent and empty-arrays were already collapsed. The real threat is unknown-keys from non-normalized data — future schema / manual seed / older backend — which is exactly what this allowlist defends.)

PR #3130, opened 2026-06-24
feat(multitable): record T9-R2 config revisions

zensgit:
## Summary
- extend the T9 config-revision recorder beyond fields with generic diff helpers
- record permission, view, sheet-config, form-share, default-view, and field-delete cascade changes in the same mutation transaction
- add real-DB goldens for permission grants, view CRUD, sheet config, form-share, and cascade batch grouping

## Scope
- recording only: no read API, no restore, no FE surface
- diff-first: no-op updates do not create history rows
- field-delete view cleanup uses the field-delete batch id

## Verification
- pnpm --filter @metasheet/core-backend test:unit
- pnpm --filter @metasheet/core-backend exec vitest run tests/integration/multitable-config-revisions-realdb.test.ts --reporter=dot (local DATABASE_URL absent, 16 skipped; CI test 20.x is the real-DB gate)


zensgit:
## 🚫 T9-R2 review gate — BLOCKING: not mergeable until BAR-1 rollback tests + the [P3] land (CI-green is NOT sufficient)

Strong PR overall — but it does **not** yet hold the review bar set for T9-R2, and CI-green does not change that. (A parallel review approved on "CI green / no blocking findings"; that misses the gap below — please don't merge on that basis.)

### BAR 1 — transaction-consistency is **not proven** (the blocker)
The per-entity goldens prove the **commit direction** (mutation succeeds → revision exists), but a recording golden **cannot distinguish** "config write + history write in one transaction" from "config write succeeds, then a separate path appends history." Only the **rollback direction** proves the recording is bound to the transaction — and there are **0** such tests. The boundary *is* structurally correct (the recorder uses the mutation's own txn `query`), so this is a **test gap, not a correctness bug** — but proving it is the entire point of this slice (the one most prone to transaction-ization hidden bugs).

**Required in this PR (minimum, not open-ended):**
- **field-permission** — fault-inject the `meta_config_revisions` INSERT to fail; assert the permission DELETE+INSERT **also rolls back**.
- **view** — ≥1 view update/delete txn: history INSERT fails → view config **unchanged**.
- **sheet-config** — history INSERT fails → sheet config **unchanged**.
- **cascade** — add the field-delete→view-cascade rollback if cheap; otherwise **document** in the PR that the cascade shares the field-delete txn and is covered by the view + field-delete boundary tests above.

### [P3]
`multitable-view-config.api.test.ts` inlines the broad `sql.includes('INSERT INTO meta_config_revisions')` — import the shared `configRevisionNoop` (the narrowed `/^\s*INSERT\s+INTO…/` regex agreed in R1-round-3) instead.

### What passes (genuinely good — keep it)
- **BAR 2** transaction-ized from the real direct-`pool.query` fact (field-perm DELETE+INSERT in a txn; views wrapped incl. view-delete's both-in-one-txn, no half-wrap; sheet-config wrapped).
- **BAR 3** field-delete→view cascade records under the **field-delete `batchId`**, in R2, only when view config changed.
- **BAR 4** records only (0 read-API/restore/FE).
- **BAR 5** all 10 mock tests swept, 0 unswept breakages; allowlist correctly unchanged.
- Entity-agnostic recorder; per-entity recording goldens (16/16); permission diffs scoped to the changed subject (no whole-list leakage); behavior preserved (view-config 7/7).

**Re-review after the rollback goldens + [P3].** Full review: `/tmp/pr3130-review-claude-20260624-gate.md`.

zensgit:
## ✅ BAR-1 blocker + [P3] resolved — added in `6cba56e9` (clean fast-forward, no work lost)

Per the owner's "继续补," I added the required transaction-consistency tests directly:

**BAR-1 rollback goldens** (in-DB fault-injection — a `BEFORE INSERT` trigger on `meta_config_revisions` that raises, so a failing history insert proves the config write rolls back *with* it):
- `BAR-1 field-permission` — failed config-rev insert → the permission DELETE+INSERT rolls back (grant unchanged)
- `BAR-1 view` — view update rolls back (config unchanged)
- `BAR-1 sheet-config` — row-deny toggle rolls back (unchanged)
- `BAR-1 cascade` — field-delete + its view-config cleanup **both** roll back (field not deleted, view cleanup undone)

**20/20** config-revisions goldens (16 recording + 4 rollback). **Teeth-verified**: no-op'ing the injected trigger (so the mutation succeeds) makes the rollback golden fail — its unchanged-assertion genuinely catches a changed config.

**[P3]** — `multitable-view-config.api.test.ts` now imports the shared `configRevisionNoop` (narrowed `/^INSERT INTO…/`) instead of the broad `sql.includes(...)`.

This closes the only blocker; BARs 2/3/4/5 already passed. **Mergeable once CI re-greens** (the rollback goldens run via the already-allowlisted `multitable-config-revisions-realdb.test.ts`). Final merge call is the owner's.

PR #3132, opened 2026-06-24
feat(approval): amount-tier preset runtime (reimbursement + purchase) — design-lock #3114

zensgit:
Builds the two amount-tier presets from design-lock **#3114**, now that G-5 (#3124, approval-node editing) and #3129 (fail-closed complex config) are on main.

## Presets
- **reimbursement_amount_tier** — direct manager → condition gate (top-level `amount` ≥ 5000) → high path adds a dept-head approval. Reuses the basic reimbursement schema.
- **purchase_amount_tier** — budget owner → condition gate (`amount` ≥ 20000) → high path forks a **parallel 会签** (joinMode 'all') of two distinct dynamic approvers that **join at the terminal `end` node**. Reuses the basic purchase schema.

## De-risked against the *real backend* first (the part the FE can't see)
Per the design's build gates, I built the graphs + ran them against real Postgres **before** the FE scaffolding:
- Both presets **create** → 201 + draft (the presets smoke `it.each` covers any list entry).
- The purchase parallel **`joinNodeKey:'end'` round-trips through normalize** (create→fetch) — catches a future join-target rewrite.
- **Terminal-join runtime:** a static-assignee parallel joining at `end` forks → both approve → join-all → instance **completes** (`approved`). This proves the design's "two approvers then done" intent *executes* (`resolveFromNode(joinNodeKey)` handles a terminal join, `ApprovalGraphExecutor.ts ~676/761`) — not just create-valid. New regression guard, and the parallel-gateway suite is **now wired into the real-DB CI job** (it was untested in CI — the smoke trap).
- Distinct **dynamic** parallel sources avoid the create-time duplicate-assignee reject; the role node is a labelled starter to retune before publishing.

## FE
The presets are complex graphs → load-preserve (`graphReadOnlyReason != null`, `unsupportedReason == null`) and round-trip **byte-identical** (G-1 anti-flatten floor), asserted per preset. The library cards fall out of the wired list; clicking creates a complex draft and drops into the G-1..G-5 editor.

Backend **12/12** (local real-DB) + FE **73/73**; vue-tsc + lint clean.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -306,6 +306,7 @@ jobs:
             tests/integration/approval-delegation-api.db.test.ts \
             tests/integration/approval-detail-subform.db.test.ts \
             tests/integration/approval-common-template-presets.api.test.ts \
+            tests/integration/approval-wp1-parallel-gateway.api.test.ts \
             --reporter=dot
 
       - name: Run comment-reaction keystone (real wire, real PG)
--- a/apps/web/src/approvals/commonTemplatePresets.ts
+++ b/apps/web/src/approvals/commonTemplatePresets.ts
@@ -1,3 +1,4 @@
+import { APPROVAL_ROLE_CONFIGURE_SENTINEL } from '../types/approval'
 import type {
   ApprovalAssigneeSource,
   ApprovalGraph,
@@ -231,9 +232,10 @@ function reimbursementAmountTierGraph(): ApprovalGraph {
 // :4152/4181), so a user-typed and a role-typed branch can NEVER collide — regardless of org data.
 // (Two USER-resolving dynamic sources — e.g. an earlier dept_head here alongside manager_at_level —
 // CAN resolve to the SAME person at runtime → APPROVAL_ASSIGNEE_PARALLEL_DYNAMIC_CONFLICT 409; that
-// was the bug.) The role is a STARTER: configure it before publishing — an unconfigured role resolves
-// empty → emptyAssigneePolicy 'error' fails the high path CLOSED (a high-value approver is never
-// silently skipped). The role node name signals "发布前配置".
+// was the bug.) The role is a STARTER using the APPROVAL_ROLE_CONFIGURE_SENTINEL placeholder: the
+// backend FAIL-FASTS at publish (assertNoUnconfiguredPlaceholderRoles → 400 APPROVAL_ROLE_PLACEHOLDER_
+// NOT_CONFIGURED) until the admin replaces it with a real role — so an UNTOUCHED preset cannot be
+// published at all (a verifiable state, not merely a runtime stuck-flow). The node name signals 发布前配置.
 function purchaseAmountTierGraph(): ApprovalGraph {
   return {
     nodes: [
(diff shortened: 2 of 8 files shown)


zensgit:
Two boundaries named proactively so they read as chosen, not missed:

**1. Newly-wired parallel-gateway tests are first-run in CI.** This PR pulls the 4 pre-existing `approval-wp1-parallel-gateway` tests into the real-DB job alongside the new terminal-join test (the suite was untested in CI before — the smoke trap). They're green locally (4/4 + the new one), but green-local ≠ green-CI (DB state / ordering / timing). If a *pre-existing* one flakes in CI, it'll surface on this PR; I'll watch that job and, if so, show it's pre-existing rather than introduced here.

**2. The runtime test isolates terminal-join, it doesn't traverse the full preset high-path.** The join=end runtime proof uses a minimal static-assignee fork→join-at-end graph, so it proves the *novel* part (terminal-join completion) cleanly without the directory. The purchase preset's full `condition → parallel fork → join-at-end` composition is **create-validated** (reachability + duplicate-assignee) and its condition-routing is G-2-covered, but not runtime-traversed end-to-end in one test. Deliberate scoping — the novel behavior is runtime-proven, the rest is covered by existing layers — flagged so it's visible.

Everything else: both presets create 201+draft against real PG, joinNodeKey='end' round-trips through normalize, FE anti-flatten floor holds per preset. Backend 12/12 local + FE 73/73; vue-tsc + lint clean.

zensgit:
Addressed — you're right, and I owned the gap in the commit.

**The miss was mine:** I reasoned "both dynamic → no *static* duplicate at create" and stopped, but the conflict is enforced at **runtime resolution** (`assertNoActiveAssignmentConflicts`, gated on `metadata.resolvedFrom`), and my terminal-join test used static assignees so it never exercised the dynamic path. `manager_at_level(2)` and `dept_head` can resolve to the same person → `APPROVAL_ASSIGNEE_PARALLEL_DYNAMIC_CONFLICT` (409).

**Fix:** the second parallel branch is now a `static_role` — which is what #3114 actually specified ("Configured Role / Person"); `dept_head` was my deviation. The conflict key is `${assignmentType}:${assigneeId}`, so a user-typed and a role-typed branch can **never** collide regardless of org data.

**Locked at the enforcement layer** (not just a comment — the draft is publishable, as you noted): two real-DB cases in the now-CI-wired parallel-gateway suite, using `form_field_user` (dynamic, resolves from form data — no directory): two user-resolving branches → same user → **409**; user + role → **clean fork**. Plus an FE shape lock so the preset can't regress to two user-resolving branches. (Note on mechanism: the conflict only fires for *dynamically-resolved* assignments — static assignees can't reproduce it, so I used `form_field_user` rather than static.)

**Product decision, named:** the role is a starter with `emptyAssigneePolicy: 'error'`, so publishing **without configuring it** hard-fails every high-amount purchase (empty role → fail-closed). Deliberate — a high-value approver is never silently skipped — and the node name signals 发布前配置. Flagging it so it's a chosen boundary, not a surprise.

zensgit:
Addressed — you're right that a publishable placeholder still jams, so I made it a **verifiable state with a publish fail-fast**, not a comment.

- **Sentinel, not a normal role id:** `APPROVAL_ROLE_CONFIGURE_SENTINEL` ('__APPROVAL_ROLE_PLACEHOLDER__'), defined in the backend and mirrored in the FE types.
- **Publish fail-fast:** `assertNoUnconfiguredPlaceholderRoles` runs in `publishTemplate` (after the existing assignee/field validations, before `buildRuntimeGraph`) → any static_role still carrying the sentinel → **400 `APPROVAL_ROLE_PLACEHOLDER_NOT_CONFIGURED`**. An untouched preset **cannot be published**.
- **Start is protected by construction:** the guard runs before the `published_definition` is written, so no published graph can ever contain the sentinel — start can't encounter it. (So I didn't add a redundant start guard; flagging in case you expected one — happy to add belt-and-suspenders if you'd prefer.)
- **Removed the wrong claim:** the "`emptyAssigneePolicy` catches it" comment is gone — it doesn't catch it, it jams.

**Lock (real-DB, wired presets smoke):** create the *real* `purchase_amount_tier` (draft 201) → publish → **400** + the code. Plus an explicit `FE_SENTINEL === BACKEND_SENTINEL` assertion so the mirror can't silently drift (a drift would make the guard miss the preset). A real (non-sentinel) static_role still publishes 200 (the parallel-gateway no-conflict case), so the guard doesn't over-block.

Backend 15/15 local + FE 65; vue-tsc + backend tsc + lint clean.

On your option B (require a real role at save/publish, with a G-5 UI guard too): the backend publish guard is the hard gate; a G-5 editor hint that surfaces the sentinel before save would be a nice UX follow-up, but I kept this PR to the enforcement layer. Say the word if you want the UI hint in-scope here.

PR #3154, opened 2026-06-25
feat(multitable): T9-R3 config-history read API (gated per-entity-type + view-literal redaction)

zensgit:
## T9-R3 — config-history READ API

Implements the read API over `meta_config_revisions` (recorded by R1 #3126 / R2 #3130) per the design-lock **#3139**. Backend only; unified timeline + FE deferred.

### Gate口径 (symmetric with the write routes; fail-closed)
One predicate (`config-history-read.ts`): field→canManageFields, view→canManageViews, sheet_config→canManageSheetAccess, permission/{sheet→canManageSheetAccess, view→canManageViews, field→canManageFields}; unknown type / unparseable permission scope → **DENY**. Never the record-history mask.

### Six read endpoints (each a single structural cap gate)
`GET /sheets/:sheetId/config-history/{fields,views,sheet-config}` + `…/permissions/{sheet,view,field}` — mirroring the record-history endpoint's shape (pagination `created_at DESC,id DESC`, actor enrichment). Belt-and-suspenders: each row re-checked against the predicate.

### View payload redaction — proven END-TO-END (data-deep, not fixture-deep)
The `/views` endpoint redacts `before/after.filterInfo` literals via the requester's `loadAllowedFieldIds` + `redactViewConfigFilterLiterals` (#2052/R9 — `canManageViews` ≠ field-read). The golden creates+patches a view filtered on a denied field **through the real routes** (so R2 records it), asserts the raw recorded row holds the literal in `filterInfo.conditions[].value`, then asserts a field-denied view-manager does **not** see it through the API (and a non-filter key survives).

### Verification (all green)
- Unit (9): every entity_type, all permission subtypes, fail-closed.
- Real-DB goldens (8): per-surface scoping; cross-cap 403 (write-only → 403 on sheet-config + sheet-perm; read-only → 403 on all); view redaction (fixture **and** end-to-end real-recorder); fail-closed; deterministic pagination.
- `tsc` 0; R1/R2 recording goldens unaffected (20/20); real-DB test registered in `plugin-tests.yml`.

Design + verification doc: `docs/development/multitable-t9-r3-config-history-read-verification-20260624.md`.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -216,6 +216,7 @@ jobs:
             tests/integration/multitable-restore-batch-preview-realdb.test.ts \
             tests/integration/multitable-restore-batch-execute-realdb.test.ts \
             tests/integration/multitable-config-revisions-realdb.test.ts \
+            tests/integration/multitable-config-history-read-realdb.test.ts \
             tests/integration/yjs-scalar-seed-realdb.test.ts \
             tests/integration/yjs-scalar-flush-realdb.test.ts \
             tests/integration/yjs-scalar-strdate-migration-realdb.test.ts \
--- a/docs/development/multitable-t9-r3-config-history-read-verification-20260624.md
+++ b/docs/development/multitable-t9-r3-config-history-read-verification-20260624.md
@@ -55,7 +55,8 @@ fully-allowed reader does; a non-filter key (the view name) survives.
 ## Verification (all green)
 
 - **Unit** — `tests/unit/config-history-read.spec.ts` (9): every entity_type, all 3 permission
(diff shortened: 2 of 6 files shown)


zensgit:
Superseded by the T9-R3.1 hotfix (claude/multitable-t9-r3.1-view-literal-hotfix-20260625). Main shipped R3 as a single /config-history?entityType= endpoint (#3153) + R4 FE on it (#3155); the six-endpoint shape here diverges and wouldn't fix the shipped path. The view payload redaction + the real-DB golden are carried into the hotfix against the single endpoint; the colonless-predicate fix isn't needed (main gates via SQL LIKE 'field:%').

PR #3161, opened 2026-06-25
docs(approval): server-side amount total-check design-lock

zensgit:
## What

Docs-only design/verification record for the server-side amount total-check line, reconciled to the current main state.

The runtime capability has already landed in #3176. This PR records the as-built design and the remaining follow-up gaps rather than claiming runtime is still unbuilt.

## Current status

- Implemented by #3176: backend `formSchema.amountConsistencyCheck` normalization + `createApproval` submit-time check + unit/real-DB coverage.
- Persistence home: versioned `formSchema.amountConsistencyCheck`, not a separate parent-template column. The earlier dedicated-column idea is withdrawn because `form_schema` / `approval_graph` live on `approval_template_versions`.
- Remaining gaps tracked in the doc: **§1 FE authoring silent-drop — CLOSED by #3197** (the authoring save now carries `amountConsistencyCheck` through verbatim — round-trip unit + active-exposure mounted guard); **§2 amount-tier preset mappings — DONE by #3183** (`4f6cd83b2`); §3 the visibilityRule save-vs-submit policy — **ratified as-built (doc-only, see #3207)**: structural-at-save + semantic-at-submit; a save-time semantic reject stays a deferred P2 (not rushed).

## Key decisions

- Explicit declared mapping, never inferred.
- Server-side, fail-closed, at submit.
- Exact money-safe comparison using declared precision / exact decimal handling.
- One total field to one detail amount column in v1.
- No auto-fill / auto-sum in this lock.

## Validation

- Docs-only diff.
- CI green on #3161.
- #3176 is the runtime implementation PR; follow-up code PRs should be split from this doc record.

### DIFF
--- a/docs/design/approval-amount-server-side-total-check-design-lock-20260624.md
+++ b/docs/design/approval-amount-server-side-total-check-design-lock-20260624.md
@@ -56,9 +56,13 @@ rows.
 3. **Exact comparison on a money-safe representation.**
    Compare `total` against `sum(rows[*][amountColumnId])` in integer minor units
    (or a fixed-scale decimal), NEVER raw IEEE float — `0.1 + 0.2` drift must not
-   create a phantom mismatch. The implementation scales each value to a fixed number
-   of decimal places (default 2), sums, and compares exactly. No tolerance band in
-   v1 (a money control is exact); a configurable epsilon is out of scope.
+   create a phantom mismatch. The scale is derived from the mapped number fields'
+   declared precision (approval `FormField.props`, aligned with the multitable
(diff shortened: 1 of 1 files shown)


zensgit:
Superseded: the total-check design + verification is on origin/main via the dev-verification MD (approval-amount-total-check-dev-verification-20260624.md, #3180/#3187), and the runtime shipped (#3176). This upfront design-lock draft was never merged; closing to avoid landing a stale draft over the documented shipped state. Recorded in the amount/formula line roadmap (PR above).

PR #3158, opened 2026-06-25
feat(attendance): v1-1b OT-bank source-tagged lots (migration + gated per-source grant)

zensgit:
Second slice of the ratified 加班银行 design-lock. **Money path — held for your review, not auto-merged** (per the per-slice review rigor; a CI-green-but-wrong grant = wrong pay).

- **Migration**: `attendance_leave_balances.overtime_source` TEXT NULLABLE → existing + dormant rows stay NULL (byte-identical).
- **`partitionOvertimeBankGrantLots`** (pure, 7 unit tests): DORMANT → single NULL-source lot of the whole total; ENABLED → partition the **rule-normalized total** by per-source weights (rule once; **last non-zero source** absorbs remainder — a 0-weight source can't be mis-attributed, caught in test); only `pooledSources` become comp_time lots (keyed `…:${source}` → per-source replay idempotency); non-pooled = must-pay (账4). `holiday → statutory_holiday` **conservatively** (never poolable §6; finer holiday sub-types + special_hours deferred until the calendar exposes them).
- **Grant**: DORMANT branch = the existing single-lot INSERT **literally untouched**; ENABLED = per-source loop. Existing C2 (sections 1–3) proves dormant byte-identical; new C2 **section (4)** proves ENABLED per-source lots + **conservation** + **replay no double-credit** on real DB.

Unit 13/13, tsc 0. NS-3 partition discipline reused. Migration `zzzz20260624160000`.

PR #3162, opened 2026-06-25
feat(attendance): v1-2a LeaveOffsetPolicy latent config + normalizer (加班银行 §4 账2)

zensgit:
Third slice — 账2 foundation. Split v1-2 into **v1-2a** (this: dormant config + normalizer + wire, no behavior change, v1-1a pattern) + **v1-2b** (the deduction-order wiring on the existing `deductLeaveBalance`, gated). Off main (independent of v1-1b). **Held for review.**

- `leaveBalanceDeductionPolicy` LATENT { enabled:false, rules:[] } → existing per-type deduction unchanged.
- enum-strict normalizer (deductFrom ∈ {comp_time, annual, unpaid}, insufficient ∈ {block, partial_unpaid_absence}; invalid rules dropped). v1 locks single-pool in v1-2b; cross-pool = v2.
- Wired DEFAULT_SETTINGS + normalizeSettings + **mergeSettings** + zod. Unit 4/4 + **real-DB wire regression** (full shape · partial-preserve · bad-pool→400).

### DIFF
--- a/packages/core-backend/tests/integration/attendance-plugin.test.ts
+++ b/packages/core-backend/tests/integration/attendance-plugin.test.ts
@@ -8665,6 +8665,34 @@ attendanceIntegrationDescribe(
       return ((res.body as { data?: Record<string, unknown> } | undefined)?.data ?? {}) as Record<string, unknown>
     }
 
+    it('round-trips leaveBalanceDeductionPolicy (wire lock): PUT→GET full shape, partial PUT preserves rules, bad pool → 400', async () => {
+      if (!baseUrl) return
+      const runSuffix = Date.now().toString(36)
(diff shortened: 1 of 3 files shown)


zensgit:
**§P2 single-pool addressed** — v1 now locks deductFrom to one pool:
- normalizer **truncates** deductFrom to the first valid pool (array shape kept forward-compatible for v2);
- PUT zod `deductFrom .min(1).max(1)` → a multi-pool PUT is **rejected (400)**;
- tests: unit now expects `personal_leave → ['comp_time']` (truncated) + a single-pool-lock test; wire test (4) asserts multi → 400.
Cross-pool order stays v2 (explicit opt-in). node -c + unit + tsc clean. Still **held** — merge after #3158 lands, per your sequence.

PR #3163, opened 2026-06-25
docs(multitable): T9-W config restore/rollback design-lock (PROPOSED)

zensgit:
## T9-W — config restore / rollback (write side) — design-lock, for review

The T9 read side is complete (R1–R4 record + the gated, redacted `/config-history` read, with the R3.1 view-literal hotfix in #3159). T9-W is the natural next main line: reverting a config entity to a recorded past state. It's security-sensitive (a restore can re-grant permissions, resurrect a deleted field, revert row-deny), so this locks scope + the security model before any code.

### Scope — v1 = single-revision revert
`meta_config_revisions` is diff-first, so PIT reconstruct = replaying every diff (deferred). **v1 reverts ONE entity from ONE revision** (apply its `before`): undo a single change. Batch / PIT / FE are deferred, each its own opt-in.

### Locks
- **W-L1** append-only / forward-only — a restore is a new write recording a new revision; never edits history.
- **W-L2** write-symmetric per-entity/subtype gate (field→canManageFields, view→canManageViews, sheet_config + sheet-perm→canManageSheetAccess, view-perm→canManageViews, field-perm→canManageFields), reusing the read-side fail-closed derivation.
- **W-L3** dry-run preview→execute, **bound** by a hashed identity token (mirrors record-restore); a stale preview fails closed.
- **W-L4** same-transaction config-write + history-row (the BAR-1 rollback model).
- **W-L5** security-relevant changes (permission re-grant, field resurrect, row-deny flip) surfaced in preview + separately gated — no coarse "restore anything" cap.
- **W-L6** idempotent / no-op-safe; single-use/TTL token.
- **W-L7** view-restore **preview** reuses `redactViewConfigFilterLiterals`; **execute** applies the real stored target server-side (a field-denied restorer never sees the denied literal).

### Open questions (decide before impl)
1. v1 single-revision vs a bounded batch day one? 2. Permission restore in v1 or held (highest blast radius)? 3. The restore-provenance marker shape.

Reuses the record-restore preview/execute + `restore-preview-identity` machinery. Full doc: `docs/development/multitable-t9-w-config-restore-design-lock-20260625.md`.

### DIFF
--- a/docs/development/multitable-t9-w-config-restore-design-lock-20260625.md
+++ b/docs/development/multitable-t9-w-config-restore-design-lock-20260625.md
@@ -1,0 +1,76 @@
+# T9-W — config restore / rollback (write side) — design-lock (PROPOSED) — 2026-06-25
+
+> **Design-lock (PROPOSED).** The T9 read side is complete (R1–R4 record + the gated, redacted
+> `/config-history` read). T9-W adds the WRITE side: reverting a config entity to a recorded past
+> state. This is design-heavy and **security-sensitive** — a config restore can re-grant
+> permissions, resurrect a deleted field, or revert a row-deny — so it gets its own lock before any
+> code. No implementation yet. Review this first.
+
+## Scope — v1 is SINGLE-REVISION REVERT (not PIT reconstruct)
+
+`meta_config_revisions` is **diff-first** (each row is `before`/`after` for one change, not a full
+snapshot). Reconstructing a whole sheet's config at an arbitrary time T = replaying every diff —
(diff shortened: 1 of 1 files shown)


zensgit:
Superseded by #3164, which shipped its own T9-W design-lock + a narrower safe-subset (field name/order + view config-update revert; permission/sheet_config/create/delete gated). This PROPOSED six-entity design-lock no longer matches the landed implementation. The preview-redaction follow-up #3164 needed is folded into the T9-R3.1 safety PR (#3159). A post-#3164 *delta* design-lock for the still-gated unsafe-restore slices can be opened when those are taken up.

PR #3169, opened 2026-06-25
feat(multitable): T9-W-3 config-restore FE panel (faithful client)

zensgit:
The FE panel for T9-W config-restore (pairs with backend #3164). Each *update* row in the config-history view gets a **Revert** action → `getConfigRestorePreview` → a confirm panel rendering `target←current` per changed key. The **server's** decision is rendered as-is: `opKind==='gated'` shows the reason and offers no confirm; `driftConflict` shows a warning and disables confirm; on confirm `executeConfigRestore(id, baselineHash)` → reload. Faithful client — no client cull.

## Verification
**11 jsdom specs** (5 existing R4 + 4 revert: hidden-when-read-only / safe preview→confirm→execute / gated-disabled-with-reason / drift-disabled-with-warning + a client wire-lock asserting `getConfigRestorePreview` unwraps the real `{ok,data:{preview}}` envelope). `vue-tsc -b` 0. In the web-guard.

### DIFF
--- a/apps/web/src/multitable/api/client.ts
+++ b/apps/web/src/multitable/api/client.ts
@@ -1023,6 +1023,20 @@ export interface MetaConfigRevision {
   createdAt: string
 }
 
+// T9-W: the would-be revert of a recorded config change (server-computed; the FE renders it as-is).
+export interface ConfigRestorePreview {
+  revisionId: string
+  entityType: string
+  entityId: string
+  changedKeys: string[]
+  current: Record<string, unknown>
+  target: Record<string, unknown>
+  driftConflict: boolean
+  opKind: 'safe' | 'gated'
+  gatedReason?: string
+  baselineHash: string
+}
+
 export interface AiShortcutPreviewData {
   status: 'succeeded'
   action: 'preview'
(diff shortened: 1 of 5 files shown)


zensgit:
**REQUEST CHANGES (blocking) — post-restore UI doesn't refresh the actual config.**

`MultitableWorkbench.onConfigReverted()` only calls `loadConfigHistory(...)`. A revert changes field `name`/`order` or view `filter`/`config`, but the field/view *save* paths reload via `loadSheetMeta(...)` + `grid.loadViewData(grid.page.value.offset)` — the revert handler does neither. So after a successful revert the user sees "撤销成功" while the field name / view filter / grid stay stale until a manual refresh (same class as the B-4 grid-refresh bug #3125).

Fix:
```
async function onConfigReverted() {
  await loadSheetMeta(activeSheetId)                 // reverted field name/order + view config
  await grid.loadViewData(grid.page.value.offset)    // re-render grid against the reverted view
  void loadConfigHistory(configHistory.value.entityType)
}
```
Then re-run `multitable-web-guard`. (Filed as REQUEST CHANGES rather than a direct push — this branch is actively held by another worktree with an unpushed commit; not grabbing it to avoid a #3154/#3163-style collision.)

PR #3193, opened 2026-06-25
feat(attendance): v1-3b 满勤 full-attendance compute on the live summary (加班银行 §3 账3)

zensgit:
The 账3 满勤 flag on the landed AttendanceBonusPolicy config. Money-path-adjacent — **held for review**. Per advisor's 3 guards + an Explore map of the formula-report system.

- **`resolveFullAttendanceEligible(summary, policy)`** → bool|null. **§4 KEY**: breaks on ANY leave **EVEN IF fully offset by the comp-time pool** — reads RAW `leave_minutes` (Σ approved leave, pre-offset), not net. **Deduction-INDEPENDENT** (real absence = 账4/v1-4). null when policy off.
- **DORMANT-CLEAN**: the cached/fingerprinted report path is NOT dormant-safe for new fields (a new field changes the SHA1 source fingerprint → re-sync for every org; static provisioning — Explore-confirmed). So 满勤 is computed on-the-fly on the **LIVE GET /summary** read, field **OMITTED when policy off** → byte-identical; cached report untouched.
- **Tests**: unit 7/7 (the §4 discriminator: fully-offset leave → still false; toggle relaxation; policy-off→null) + real-DB (policy OFF → field absent; ON + clean → present & true).
- v1: any late/early day breaks 满勤 (toggle); finer #5 RT severe-threshold = follow-up. real-absence stays in v1-4.

node -c OK, unit 7/7, tsc 0.


zensgit:
**§P1 fixed** — 满勤 compute now counts `late_early_days`:
- `hasLateOrEarly` now also checks `late_early_days` (the same-day both-late-AND-early status that's aggregated separately at index.cjs:11643/11698) → a both-late+early person is correctly **not 满勤** when lateBeyondThresholdBreaksFullAttendance is on.
- unit reverse-test added (`{late_days:0,early_leave_days:0,late_early_days:1}` → false); **real-DB false-path** added (approved leave in period → 满勤 false), since the prior route test only had OFF-absent + clean-true.
node -c + unit + tsc clean. Re-running CI; held for your re-review.

PR #3196, opened 2026-06-25
docs(multitable): add Global History acceptance runbook

zensgit:
## Summary
- add a non-destructive Global History / point-in-time restore acceptance runbook
- map product entry points, trial data, permission/security checks, expected statuses, and release criteria
- keep T8-2 Reset-to-T, T9-W data-loss ops, and record-history keyset hasMore explicitly gated

## Verification
- git diff --check
- rg -n "Time Machine|timemachine" docs/development/multitable-global-history-nondestructive-acceptance-runbook-20260625.md (no matches)

### DIFF
--- a/docs/development/multitable-global-history-nondestructive-acceptance-runbook-20260625.md
+++ b/docs/development/multitable-global-history-nondestructive-acceptance-runbook-20260625.md
@@ -1,0 +1,190 @@
+# Multitable Global History non-destructive acceptance runbook
+
+Grounding: `origin/main` at `2b56e749b` on 2026-06-25. This runbook is for the
+first product acceptance pass after the non-destructive Global History /
+point-in-time restore line closed on main.
+
+Recent closeout anchors:
+
+- #3169 (`0d5dfe511`) shipped the T9-W FE token plumbing;
+- #3177 (`ab92ae3c`) shipped R4 diff polish and the full mount-to-fetch wire test;
+- #3184 (`6e666871`) reconciled the canonical verification docs;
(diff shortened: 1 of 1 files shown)


zensgit:
**Review: APPROVE with one Medium finding** (docs-only, CI green). Thorough non-destructive acceptance runbook; correctly holds T8-2 destructive Reset out as a remaining gate (§8). It even captures this session's fixes (revert→meta/grid refresh; previewToken-not-baseline-hash; no-existence-oracle; 413 fail-closed; all-or-nothing zero-writes; page-reload step).

**Medium — missing the view filter-literal redaction criterion.** §3.D verifies the config-history ACCESS gate ("read gate matches mutation gate") but not the view-filter-literal PAYLOAD redaction — the exact leak fixed this session in #3159 (the `/config-history` read) **and** the `config-restore-preview` path. `canManageViews` does not imply field-read, so a view's `filterInfo.conditions[].value` literals are redacted per-requester (#2052/R9), independent of the entity-type gate. With no acceptance criterion, a regression re-exposing denied filter literals would pass acceptance. Suggest adding to §3.D (+ a §5 step):
- [ ] A `canManageViews`-but-field-denied actor reads config-history for a view filtered on a denied field and does NOT see the denied field's filter literal (redacted; fieldId/operator kept).
- [ ] The same actor's view restore-preview redacts `current`/`target.filterInfo` literals; execute still applies the real server-side target.

Mergeable as-is; this criterion should land before the non-destructive pilot sign-off so the just-fixed leak is regression-gated.

PR #3201, opened 2026-06-25
feat(attendance): v1-4 账4 settlement export — per-source remaining + must-pay marker (加班银行 §4)

zensgit:
The 账4 export contract on the landed v1-3b summary surface. **Money-path — held for review.** Per advisor + §4 literal contract (各来源剩余 + 可折算/必须直付标记, **no amounts**).

- **`buildOvertimeSettlementExport`** → null (dormant) | { `convertibleBySource` {workday,restday} = comp_time lot **remaining** by `overtime_source`; `mustPayBySource` {statutory_holiday} = holiday OT total (never banked → must-pay, §6) }. **must-pay is an attendance fact, not a payroll inference** — computed from facts we already have. No amounts (倍率/金额 = payroll).
- **DORMANT-CLEAN**: on the LIVE GET /summary read, gated on overtimeBankPolicy; field **omitted + lots query skipped when off** → byte-identical; cached report untouched (same constraint as v1-3b).
- Tests: unit **4/4** + real-DB (OFF → absent; ON → present, banked restday lot → convertibleBySource.restday=90, must-pay has statutory_holiday).

real-absence stays out (账3 result). node -c, tsc 0.

### DIFF
--- a/packages/core-backend/tests/integration/attendance-plugin.test.ts
+++ b/packages/core-backend/tests/integration/attendance-plugin.test.ts
@@ -6067,6 +6067,53 @@ attendanceIntegrationDescribe(
     }
   })
 
+  it('④ 加班银行 v1-4 — overtimeSettlement on GET /summary when bank ON (convertible by source + must-pay); ABSENT when OFF', async () => {
+    if (!baseUrl) return
+    const dbUrl = process.env.ATTENDANCE_TEST_DATABASE_URL || process.env.DATABASE_URL
+    if (!dbUrl) return
+    const runSuffix = Date.now().toString(36)
+    const userId = `attendance-v14-${runSuffix}`
+    const previousRbacBypass = process.env.RBAC_BYPASS
+    const pool = new Pool({ connectionString: dbUrl })
+    let token: string | undefined
+    let originalSettings: Record<string, unknown> = {}
+    try {
+      process.env.RBAC_BYPASS = 'true'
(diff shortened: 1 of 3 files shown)


zensgit:
**Closing — deferring the §4 settlement export to v1-5 (snapshot), per your [P1].**

Your [P1] is correct and load-bearing: `overtimeSettlement` conflates **period facts** (must-pay from the from/to summary) with the **current cumulative comp_time balance** (`loadCompTimeRemainingBySource` has no period/asOf/cycle boundary). A September query returns September must-pay but the *current* convertible balance — and a from/to consumer reads it all as period-bound → mis-pay/漏发 risk. The [P2] (the route test never locks the must-pay minutes / doesn't prove must-pay isn't balance-derived) is real too.

Both of your stabler fixes — convertible by 结算时点/周期, or defer to the v1-5 snapshot — require the **point-in-time/snapshot machinery that v1-5 provides**. A §4 settlement export (周期末各来源剩余 + 可折算/直付) is inherently a period-end/snapshot concept; it can't be made temporally coherent on a live mixed-temporal read. This is exactly the prematurity I flagged before building it — I should have held it. So:

- **Closing this PR** rather than shipping a mis-pay-risky export or a relabeled current-balance stub.
- The §4 settlement export **folds into v1-5**: snapshot the active PayrollSettlementPolicy + reconstruct each source's banked-remaining **as-of the settlement point** (so convertible + must-pay share one temporal basis), then emit the export with must-pay proven from period OT facts (not balance lots) — and the v1-5 real-DB matrix locks the must-pay **minutes** (your [P2]) + the poison-lot anti-derivation case.

Reusable groundwork from here: `loadCompTimeRemainingBySource` (the by-overtime_source aggregation) carries into v1-5's as-of reconstruction. No code from this branch lands as-is.

PR #3206, opened 2026-06-25
docs(attendance): v1-5 settlement-snapshot design-lock (snapshot-at-close, ratified)

zensgit:
Converges your v1-5 ratification into an implementable design-lock — **design-lock-first, before any v1-5 runtime**. Absorbs the §4 export deferred from v1-4 (#3201).

- **snapshot-at-close** (reject reconstruct-as-of first); **new typed table** `attendance_payroll_cycle_settlements` + UNIQUE(org,cycle,user,source); cycle.metadata holds only id/hash/status; **no amounts**.
- **trigger** locked at cycle→closed (update→closed AND create-as-closed share one txn-local path).
- **snapshot scope**: the 3 policies + PolicyAssignment/effectiveFrom/version/hash; money = payroll.
- **must-pay from the cycle period OT facts**; **convertible = balance-at-close** (cumulative, labeled) — a deliberate two-temporal pairing (your refinement-accepted [P1] fix).
- **immutable + idempotent replay-close** (ON CONFLICT DO NOTHING) + explicit recompute (out).
- **§7 invariant**: must-pay never balance-derived + **poison-lot real-DB test** + lock the minutes (your [P2]).
- **§8 build slices v1-5a..d**, each held for review.

Docs-only; v1-5 build opens after this lands.


zensgit:
**CHANGES REQUIRED → all 3 addressed (docs-only).** Sharp catches — each would have been a v1-5 runtime money-path bug; the design-lock-first gate earned its keep again.

- **[P1] closed-cycle period not frozen.** Did **both** you offered: (a) settlement table gains `period_start_date`/`period_end_date`/`closed_at` so the snapshot carries its own period; (b) §3 forbids changing `start_date`/`end_date`/`template_id` once `status=closed` (only archive / explicit reopen-recompute). §8d read-out keys off the table's frozen period, not the cycle's current dates → a later cycle PUT can't hang an old snapshot under a new period.
- **[P2] legacy/NULL `overtime_source`.** `source NOT NULL`; NULL `overtime_source` COALESCEs to a `legacy_unsourced` archive source at settlement (so Postgres NULL≠NULL can't bypass the UNIQUE key or drop dormant balance from the group). `legacy_unsourced` = **convertible** (banked comp_time), not must-pay; settlement enumerates lots COALESCE-then-group.
- **[P2] settlement target population.** §8a pins it = (OT/leave-offset/attendance facts in the cycle period) ∪ (active comp_time balance at close), org-bounded; `active user_orgs` is only a permission reference, never the sole source — the annual-leave 'only-scan-active drops offboarded-with-balance' trap. v1-5b's real-DB matrix gains that offboarded-person case.

Header now reflects 'CHANGES REQUIRED → 3 closed'. Re-review when you can; build opens on your re-ratify.

PR #3211, opened 2026-06-25
feat(attendance): v1-5a settlement snapshot table migration (dormant) — 加班银行 §2

zensgit:
First v1-5 build slice on the ratified design-lock (#3206). **DORMANT** — creates the table only; nothing writes it until v1-5b. Held for review (it's the settlement home).

- **migration** `attendance_payroll_cycle_settlements` with the **3 #3206 constraints baked in**: [P1] `period_start_date`/`period_end_date`/`closed_at` (frozen period — a later cycle PUT can't hang an old snapshot under a new month); [P2] `source NOT NULL` (NULL→`legacy_unsourced` at settlement so the UNIQUE key isn't NULL-bypassed); `UNIQUE(org_id,cycle_id,user_id,source)` = replay-close idempotency key. **No amount columns**.
- **real-DB schema test**: cols present, source NOT NULL, UNIQUE rejects duplicate source, distinct source (incl. legacy_unsourced) allowed, NULL source rejected.

tsc 0. Lands after #3206. v1-5b (the snapshot-at-close compute + poison-lot matrix) is next, held.


zensgit:
**[P2] fixed — FK is now ON DELETE RESTRICT.** Sharp catch: ON DELETE CASCADE would have let a cycle delete cascade-wipe the immutable closed settlement snapshots (#3206), bypassing archive/reopen-recompute — and once v1-5b writes, that's silent loss of accounting facts.
- migration: `cycle_id ... ON DELETE RESTRICT` (DB refuses to delete a cycle while settlement rows exist — the DB-level backstop; v1-5b adds the route-layer closed-cycle delete/update guard).
- real-DB test added: with a settlement row present, `DELETE FROM attendance_payroll_cycles` throws.
Rest unchanged per your review (period_start/end/closed_at · source NOT NULL · UNIQUE · legacy_unsourced). tsc 0; re-running CI. Held for re-review.

PR #3228, opened 2026-06-26
feat(attendance): v1-5b-i settlement compute (pure, dormant) — 加班银行 §5 correctness core

zensgit:
The **decisive correctness gate** of v1-5b, built pure + **locally proven**. DORMANT (nothing calls it yet → byte-identical runtime). Held for review.

- **`buildCycleSettlementRows`** (pure §5): convertible = comp_time remaining by source; must-pay = cycle-period statutory holiday OT **from period facts**. **UNCONDITIONAL** (advisor #1/§6) — not bank-gated, so a bank-off org with holiday OT still gets a must-pay row.
- **`loadCompTimeRemainingBySource`**: COALESCE NULL→`legacy_unsourced` (§2 [P2]).
- **unit 5/5 LOCAL**: **poison-lot** (bogus statutory balance lot can't move must_pay) · **unconditional must-pay** · convertible-by-source · legacy_unsourced · clamp. tsc 0.

**Split**: the settlement-WRITE wiring (`snapshotCycleSettlementOnClose` + the two cycle→closed hooks + §8a population + real-DB matrix) is **v1-5b-ii** — per advisor, the local real-DB matrix is the gate, and this env has no local DB, so I'm not pushing can't-locally-verify money-path settlement writes. v1-5b-ii wants a pass with a DB.

### DIFF
--- a/packages/core-backend/tests/unit/attendance-settlement-compute.test.ts
+++ b/packages/core-backend/tests/unit/attendance-settlement-compute.test.ts
@@ -1,46 +1,60 @@
 import { describe, expect, it } from 'vitest'
 
 const attendancePlugin = require('../../../../plugins/plugin-attendance/index.cjs')
+type Row = { source: string; convertibleMinutes: number; mustPayMinutes: number }
 const helpers = attendancePlugin.__attendanceOvertimeBankForTests as {
   buildCycleSettlementRows: (input?: {
-    holidayOvertimeMinutes?: number
+    periodOtBySource?: Record<string, number>
     compTimeRemainingBySource?: Record<string, number>
-  }) => Array<{ source: string; convertibleMinutes: number; mustPayMinutes: number }>
+    pooledSources?: string[]
+  }) => Row[]
 }
 
 describe('#加班银行 v1-5b — buildCycleSettlementRows (settlement compute §5)', () => {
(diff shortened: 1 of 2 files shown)


zensgit:
**[P1] fixed — must-pay now includes un-pooled sources, not just statutory.** Real under-pay catch: with pooledSources=['restday'], a period's workday OT was un-pooled (never granted as comp_time) but my helper dropped it → payroll short by those minutes.

- `buildCycleSettlementRows` now takes **periodOtBySource + the effective pooledSources** (+ compTimeRemainingBySource). must-pay = statutory_holiday ALWAYS (§6, unconditional) **+ workday/restday period OT when that source is NOT in pooledSources** (un-pooled → must-pay; pooled → 0, it became convertible). bank-off/pooledSources=[] → all non-statutory must-pay. convertible + must-pay coexist on one row.
- Poison-lot unchanged: must-pay reads only period facts.
- **unit 7/7 LOCAL**: your exact scenario (pooledSources=['restday'] → workday 120 must-pay) + all-pooled→0 + bank-off-all-must-pay + statutory-always + poison-lot + coexist. tsc 0.

Still dormant; v1-5b-ii wires + runs the real-DB matrix (which will re-prove this per-source split end-to-end). Held for re-review.

PR #3233, opened 2026-06-26
feat(attendance): v1-5b-ii settlement snapshot-at-close wiring (money-path, held) — 加班银行 §3/§5/§8a

zensgit:
The settlement **WRITE** — `snapshotCycleSettlementOnClose` hooked into **all three cycle→closed paths** (PUT update→closed, POST create-as-closed, generate-as-closed), each in the **same txn** as the status write, **transition-only** (re-close doesn't re-snapshot). **Held for review.**

- §8a population = facts-in-period ∪ active comp_time balance (offboarded-with-balance still settles). Per user: period OT by source + close-time comp_time remaining + snapshotted pooledSources → `buildCycleSettlementRows` → INSERT with frozen period + closed_at + §4 policy snapshot, **ON CONFLICT DO NOTHING** (idempotent). **Unconditional** (advisor #1).
- real-DB matrix: convertible · **POISON statutory balance lot → no row** · frozen period · **idempotent re-close** · balance-only population.

**Honest limits (please scrutinize):**
1. Built **without a local DB** — the real-DB matrix runs only on CI; CI + your review are the gate (I couldn't run the matrix locally, contrary to the ideal).
2. **must-pay-from-period-OT end-to-end** (the [P1] un-pooled case) is **unit-proven** but not yet wired-e2e (needs the OT-segmentation harness) — flagged as v1-5b-iii.
3. route-layer closed-cycle delete/update guard is still a follow-up (FK RESTRICT is the DB backstop).

node -c + tsc 0; unit compute green. This is the money-path settlement write — your review is the gate.

### DIFF
--- a/packages/core-backend/tests/integration/attendance-plugin.test.ts
+++ b/packages/core-backend/tests/integration/attendance-plugin.test.ts
@@ -6125,7 +6125,7 @@ attendanceIntegrationDescribe(
       expect([201, 409]).toContain(tmplRes.status)
       const tmplId = (tmplRes.body as { data?: { id?: string } } | undefined)?.data?.id as string
       const genRes = await requestJson(`${baseUrl}/api/attendance/payroll-cycles/generate`, { method: 'POST', headers, body: JSON.stringify({ templateId: tmplId, anchorDate: '2026-12-15', count: 1, status: 'closed', namePrefix: `Gen ${runSuffix}` }) })
-      expect(genRes.status).toBe(201)
+      expect(genRes.status).toBe(200) // generate responds 200 with { created, skipped } (not 201)
       const generated = (genRes.body as { data?: { created?: { id?: string }[] } } | undefined)?.data?.created ?? []
       expect(generated.length).toBeGreaterThanOrEqual(1)
       const genId = generated[0]?.id as string
--- a/packages/core-backend/tests/integration/attendance-plugin.test.ts
+++ b/packages/core-backend/tests/integration/attendance-plugin.test.ts
@@ -6017,6 +6017,69 @@ attendanceIntegrationDescribe(
     }
   })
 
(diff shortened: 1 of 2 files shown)


zensgit:
**[P1] + [P2] both fixed.** You were right that I shouldn't have deferred the guard the design-lock mandated.

- **[P1] closed-cycle freeze (now in this PR):** PUT rejects changing start_date/end_date/template_id once `status='closed'` (409 `CYCLE_CLOSED_PERIOD_FROZEN`; status-only closed→archived still allowed); DELETE of a closed cycle → 409 `CYCLE_CLOSED_NOT_DELETABLE` (FK RESTRICT is the DB backstop). So a closed cycle can't be re-hung to a new period while its snapshot stays on the old one.
- **[P2] both close entries now in the matrix:** added two real-DB reverses — POST create `status:'closed'` writes settlement, POST generate `status:'closed'` writes settlement — proving neither bypasses the §3 close-snapshot path; plus the [P1] guard assertions (PUT-change→409, DELETE-closed→409, status-only→200).

node -c + tsc 0; re-running CI. The route guard is no longer a follow-up. Held for re-review.

zensgit:
**[P2] archived-bypass fixed.** Real hole — you're right that 'archived' is once-closed, so checking only `status === 'closed'` let close → archive → change re-open the period (and close → archive → delete fall through).

Both guards (PUT period-freeze + DELETE) now freeze if **status === 'closed' OR 'archived'**, OR — durable backstop — the cycle **already has settlement rows** (covers a reopened-with-settlements edge + the zero-row closed/archived case you flagged). The freeze survives archive; status-only transitions + name/metadata stay allowed; open-with-no-settlements stays mutable.

**Regression added:** close → archive → period change → **409**; close → archive → delete → **409**.

node -c + tsc 0; re-running CI. Held for re-review.

PR #3252, opened 2026-06-26
docs(approval): narrow residual total-check wording missed by #3243

zensgit:
## What

Tiny follow-up: two threat-model-wording residues that **#3243 was meant to clear but didn't reach main** — its merge head (`52f98dbd`) had dropped the "verified value" fix in a concurrent rebase. Both are still on main.

- presets design-lock D1: `the amount tier routes on a verified value` → `… a consistency-checked total`
- amount-formula-line roadmap (the shipped money-chain paragraph): `formula conditions route on those backend-verified values` → `… those consistency-checked totals`

Same honest boundary as #3243: the total-check verifies **top-level total == detail-row sum** (consistency), not the business truthfulness of the amounts. The roadmap's *line-math* `tamper-proof` use stays untouched — it's a correct, distinct claim.

Docs-only, 2 files.

### DIFF
--- a/docs/design/approval-amount-tier-template-presets-design-lock-20260624.md
+++ b/docs/design/approval-amount-tier-template-presets-design-lock-20260624.md
@@ -29,7 +29,7 @@ approval-result write-back.
    The branch condition reads the top-level total field (`amount`). AS BUILT (updated 2026-06-26,
    post-closeout): the detail total is now read-only **auto-filled** from the detail rows (#3198) and a
    server-side **total-check** (#3176) rejects any total ≠ the detail sum — so the backend total-check is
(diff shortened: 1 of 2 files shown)


zensgit:
Superseded: `main` already carries the honest threat-model wording for both spots — presets D1 reads "consistency-checked total — one verified only against the submitted detail-row sum, not for truthfulness" (my fix + an explicit honesty clause), and the roadmap reads "consistency-checked values". Another session landed these, so this PR's two fixes are redundant and the branch now conflicts with main. Closing.

Note: the roadmap uses `consistency-checked values` rather than the `consistency-checked totals` suggested in review — both are honest (neither claims truthfulness); flag if the exact `totals` wording is wanted and it's a trivial separate change.

PR #3264, opened 2026-06-26
feat(multitable): T9-W Tier 1 — sheet_config revert (flag-gated)

zensgit:
## T9-W Tier 1 — sheet_config revert (owner-greenlit; behind `MULTITABLE_ENABLE_SHEET_CONFIG_REVERT`)

First unsafe config-restore slice. `classifyRevert` stays **pure** (sheet_config = intrinsically gated); the **route** supports it when the per-tier flag is on. Builds on U-1a (#3258, the redactor).

### Constraints baked in (per review)
- **opKind override** — `computeRevertPreview` returns `opKind:'gated'` (from the pure `classifyRevert`), which the FE uses to hide confirm. For flag-on sheet_config the route overrides the **preview response** `opKind→'safe'` so the confirm path shows. Execute gates on the flag+guard, not opKind.
- **Fail-closed `entity_id===sheet_id` guard** — a sheet_config revision whose `entity_id` ≠ its `sheet_id` → `400 INVALID_REVISION`, so a malformed/forged revision can't restore config into another sheet.
- **U-L6 redaction** — preview `current`/`target` `conditionalReadRules` literals redacted per requester (reuses U-1a); execute applies the raw server-side target.
- **U-L1 per-tier flag** — default off; preview AND execute → `403 SHEET_CONFIG_REVERT_DISABLED`.

### Runtime
`SHEET_CONFIG_COLUMN` + `applyConfigRevert`/`loadSheetConfigSnapshot`/`loadEntityConfigSnapshot` for sheet_config. `loadSheetConfigSnapshot` returns the **raw jsonb** (not re-parsed) so it matches the recorder's stored shape — `stable()` is plain `JSON.stringify`, so a re-parse's JS key order would trip a false drift.

### Goldens (7, real-DB)
flag-off→403 (preview+execute) · gate (write-not-share)→403 · `entity_id` mismatch→400 · **flag-on happy: preview confirmable (opKind safe, no drift) + execute restores rules + forward `source=restore` revision** · drift→409 · **U-L6 field-denied reader can't see the secret rule literal** (opKind still safe). Regression green (config-restore + both redaction goldens); tsc 0; registered in `plugin-tests.yml`.

Default-off, so inert until explicitly enabled. Tier 2 (lossy retype) is next, after the U-L8 re-confirm on #3254.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -219,6 +219,7 @@ jobs:
             tests/integration/multitable-restore-batch-allornothing-realdb.test.ts \
             tests/integration/multitable-config-revisions-realdb.test.ts \
             tests/integration/multitable-config-restore-realdb.test.ts \
+            tests/integration/multitable-sheet-config-revert-realdb.test.ts \
             tests/integration/multitable-revert-pit-realdb.test.ts \
             tests/integration/multitable-reset-pit-realdb.test.ts \
             tests/integration/multitable-config-revision-retention-realdb.test.ts \
--- a/packages/core-backend/src/multitable/config-restore.ts
+++ b/packages/core-backend/src/multitable/config-restore.ts
@@ -111,9 +111,24 @@ export async function loadViewConfigSnapshot(query: QueryFn, viewId: string): Pr
   }
 }
 
+export async function loadSheetConfigSnapshot(query: QueryFn, sheetId: string): Promise<Record<string, unknown> | null> {
+  const res = await query('SELECT conditional_read_rules, row_level_read_permissions_enabled FROM meta_sheets WHERE id = $1', [sheetId])
+  const row = res.rows[0] as Record<string, unknown> | undefined
+  if (!row) return null
+  // Return the RAW jsonb value (NOT re-parsed): the recorder stored before/after as jsonb too, so both go through the
+  // same Postgres key-ordering. stable() is a plain JSON.stringify (no key-sort), so re-parsing here (which rebuilds
+  // objects in JS key order) would mismatch the recorded `after` and trip a FALSE drift. `?? []` matches the recorder's
+  // empty shape.
+  return {
+    conditionalReadRules: row.conditional_read_rules ?? [],
+    rowLevelReadPermissionsEnabled: row.row_level_read_permissions_enabled === true,
+  }
+}
+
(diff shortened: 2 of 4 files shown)


zensgit:
**Post-merge review note — [P1] flag-on `sheet_config` bypass is too broad (needs a follow-up narrowing PR)**

This landed (`f69fa535`) with an over-broad flag-on bypass. The flag (`MULTITABLE_ENABLE_SHEET_CONFIG_REVERT`) is default-off and the live recorder only emits `action:'update'` (univer-meta.ts:6639/6724), so this is **not an active breach** — but it should be narrowed before the flag is ever enabled, and it is forgeable (the PR already adopts the forged-revision threat model via the `entity_id===sheet_id` guard).

Both flag-on hooks key off `rev.entity_type === 'sheet_config'` **alone**, bypassing `classifyRevert`'s action/key gating:
- Preview (univer-meta.ts:7746): `preview.opKind = 'safe'; delete preview.gatedReason` — unconditional.
- Execute (univer-meta.ts:7797): `if (classify.kind === 'gated' && rev.entity_type !== 'sheet_config') return 422` — the `!== 'sheet_config'` escape neutralizes the `action !== 'update'` gate (config-restore.ts:30).

**Impact:** a forged sheet_config revision with `action='create'|'delete'` (or unknown `changed_keys`) but mapped keys → confirmable preview + an executed `UPDATE meta_sheets` + an appended restore revision = violates #3254's hold on Tier 3/4.

**Follow-up fix** — narrow predicate in `config-restore.ts` next to `classifyRevert` (shared + unit-testable):
```
isSupportedSheetConfigRevert(rev) =
  rev.action === 'update' &&
  rev.changed_keys.length > 0 &&
  rev.changed_keys.every(k => k === 'conditionalReadRules' || k === 'rowLevelReadPermissionsEnabled')
```
- Preview: override to `safe` **only if** the predicate passes (else leave `gated`).
- Execute: skip the 422 **only if** the predicate passes → unsupported sheet_config returns a clean **422 `RESTORE_NOT_SUPPORTED`** (not a 500 via `applyConfigRevert`'s throw).

**Required golden** (real DB, flag-on, `canManageSheetAccess`): sheet_config `action='create'`, `action='delete'`, and unknown `changed_keys` each → gated preview (not `safe`) AND execute 422, with **zero writes** and **no `source='restore'` revision**.

Rest of U-1b Tier 1 verified good (U-1a redaction reuse, `entity_id===sheet_id` guard, flag-off/gate/drift/redaction/happy-path goldens).

PR #3274, opened 2026-06-26
docs(attendance): propose anomaly result edit guardrails

zensgit:
## What changed

Adds a PROPOSED design-lock for attendance anomaly result edits: a dedicated admin action for correcting abnormal attendance results with reason/evidence, bounded edit window, immutable audit, and affected-employee-only notification.

This is docs-only. It does not unlock runtime code; AE-1..AE-4 remain separate owner-reviewed slices.

## Why

Current `/api/attendance/anomalies` is read-only and the UI only guides admins into creating correction requests. The missing contract is a first-class, audited admin result edit for confirmed anomalies, without silently overwriting records or notifying unrelated managers.

## Grounding

- `GET /api/attendance/anomalies` is read-only and returns `suggestedRequestType`.
- `AttendanceView.vue` anomaly action currently calls `prefillRequestFromAnomaly(item)`.
- `upsertAttendanceRecord` / `computeAttendanceRecordUpsertValues` are the consistency path for metrics and report-tier meta.
- C5 delivery tables can back affected-employee-only notification.

## Validation

- `git diff --cached --check`
- Brand/competitor term grep: zero hits in the new design doc

## Owner decisions before runtime

The doc leaves five explicit ratification questions: edit-window default/range, allowed target statuses, evidence shape, closed-cycle behavior, and whether notification can be disabled while preserving audit.

### DIFF
--- a/docs/development/attendance-anomaly-result-edit-guard-design-lock-20260626.md
+++ b/docs/development/attendance-anomaly-result-edit-guard-design-lock-20260626.md
@@ -56,7 +56,7 @@ POST /api/attendance/anomaly-result-edits
   "evidence": [
     { "type": "url", "label": "现场签到照片", "url": "https://..." }
   ],
-  "idempotencyKey": "optional-client-key"
+  "idempotencyKey": "client-generated-key"
 }
 ```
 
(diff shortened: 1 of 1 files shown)


zensgit:
Addressed Lorentz review findings in c9cb0884e.

- [P1] Idempotency is now request-required: `idempotencyKey` must be supplied by the client (frontend may generate it when opening the modal), missing key => 400, route never generates a random key. Audit table keeps `idempotency_key text not null`, and AE-1 tests now include missing-key no-write.
- [P1] Closed-cycle guard is now exact: any same-org `attendance_payroll_cycles` row with `status IN (closed, archived)` and `work_date BETWEEN start_date AND end_date` rejects with `ATTENDANCE_RESULT_EDIT_CYCLE_CLOSED`; no settlement row required; closed/archived wins over overlapping open cycles; missing cycle table fails closed/DB_NOT_READY.
- Wording tightened from global admin to attendance admin.
- Evidence URL support no longer claims an existing attendance URL allowlist; AE-1 must either accept attachmentId/text-only or add a tested HTTPS-only validator.
- UI scope now names both anomaly surfaces (`AttendanceView.vue` and `AttendanceRequestCenterSection.vue`).

Local checks: `git diff --check`; brand/competitor grep clean.

PR #3275, opened 2026-06-26
docs(attendance): propose report digest subscriptions

zensgit:
## What changed

Adds a PROPOSED design-lock for attendance report digest subscriptions: org-level daily/weekly/monthly cadence config, bounded channel/recipient config, a scheduler producer that writes C5 outbox rows, small explainable digest payloads, and admin UI / staging slice boundaries.

Docs-only; no runtime is unlocked. RD-1..RD-5 remain separate owner-reviewed slices.

## Why

Current main has the building blocks: attendance summary/request report read surfaces, C5 notification outbox + delivery worker, and AttendanceScheduler secondary jobs. It does not yet have a subscription policy or producer that sends regular attendance summaries.

## Grounding

- `AttendanceScheduler` supports secondary jobs.
- `AttendanceNotificationDeliveryWorker` routes by row.channel and already has a default work-notification channel plus `email_smtp`.
- `GET /api/attendance/notification-deliveries` is read-only admin observability.
- `GET /api/attendance/summary` and `/reports/requests` provide digest facts.
- `AttendanceView.vue` has delivery status UI but no digest subscription card.

## Validation

- `git diff --cached --check`
- Brand/competitor term grep: zero hits in the new design doc

## Owner decisions before runtime

The doc leaves five explicit ratification questions: complete-period vs same-day daily cadence, default cadence states, default recipients, channel scope, and UI placement.

### DIFF
--- a/docs/development/attendance-report-digest-subscription-design-lock-20260626.md
+++ b/docs/development/attendance-report-digest-subscription-design-lock-20260626.md
@@ -1,0 +1,251 @@
+# 考勤统计通知订阅 — 设计锁（PROPOSED）
+
+> **Status**: 🟡 **PROPOSED（待 owner 拍板）**。本设计锁把“按日/周/月主动发送考勤统计摘要”收敛成 MetaSheet 自己的契约：订阅配置、周期 producer、C5 outbox、可解释 payload、以及 delivery observability。本文只锁设计，不写运行时代码；实现仍按 RD-1 → RD-5 一刀一 PR。
+
+---
+
+## 1. 背景与现状
+
+当前 main 上已有三块可复用基座：
+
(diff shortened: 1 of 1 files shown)


zensgit:
Planck review fixes pushed in b94ab1872.

Closed the two contract drifts:
- C5 delivery worker does not resolve owner/sub_owner. The doc now pins RD-3 as the place that resolves active attendance group membership + managers and materializes concrete recipient_user_id rows; self maps to existing recipient_role=subject. Zero-group and multi-group behavior are now explicit.
- Env gates are now layered: ATTENDANCE_SCHEDULER_ENABLED for the shared scheduler, ATTENDANCE_REPORT_DIGEST_ENABLED for this producer, and ATTENDANCE_NOTIFICATION_DELIVERY_WORKER_ENABLED + channel envs for real delivery. RD-5 must prove producer row creation and worker-visible send/fail separately.

Also addressed the non-blocking ambiguities: source_key now includes the normalized delivery channel, matching the current per-channel C5 producer pattern, and the verification matrix/table were updated so they do not drift from the body.

PR #3278, opened 2026-06-26
feat(attendance): add self-service rules read endpoint

zensgit:
## Summary
- Add `GET /api/attendance/rules/me` for employee self-service rule visibility.
- Pin subject/org to authenticated context, reject `userId`/`orgId`/`groupId` query/body/header overrides before DB reads.
- Return a whitelist-only explainable summary: attendance groups, schedule groups, runtime rule, optional configured group rule marker, punch-policy subset, and warnings.
- Add an integration test covering token-subject behavior, override rejection, multi schedule warnings, configured-rule divergence, and no leakage of `approvalFlowId` / rule-set config / `geoFence`.

## Verification
- `node -c plugins/plugin-attendance/index.cjs`
- `pnpm --filter @metasheet/core-backend exec vitest --config vitest.integration.config.ts run tests/integration/attendance-plugin.test.ts --reporter=dot -t "attendance rules /me"` (collects; local DB env absent, existing harness skips integration tests locally)
- `git diff --check`

## Notes
- This intentionally does not expose raw attendance settings, geofence config, approval flow ids, or rule-set config to the employee endpoint.
- The route is stricter than the older annual leave `/me` read: spoofing fields are rejected instead of ignored.

### DIFF
--- a/packages/core-backend/tests/integration/attendance-plugin.test.ts
+++ b/packages/core-backend/tests/integration/attendance-plugin.test.ts
@@ -5482,6 +5482,168 @@ attendanceIntegrationDescribe(
     }
   })
 
+  it('attendance rules /me — returns a token-subject rule summary, rejects overrides, and does not leak admin settings', async () => {
+    if (!baseUrl) return
+    const dbUrl = process.env.ATTENDANCE_TEST_DATABASE_URL || process.env.DATABASE_URL
+    if (!dbUrl) return
+    const pool = new Pool({ connectionString: dbUrl })
+    const runSuffix = Date.now().toString(36)
+    const meId = `rules-me-${runSuffix}`
+    const otherId = `rules-me-other-${runSuffix}`
+    const adminId = `rules-me-admin-${runSuffix}`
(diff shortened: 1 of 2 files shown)


zensgit:
Review follow-up fixed in head f55618332:\n\n- Closed the tenant-header spoof path: /api/attendance/rules/me now rejects x-tenant-id and x-workspace-id alongside x-org-id/x-user-id, and treats tenantId/tenant_id/workspaceId/workspace_id query/body keys as subject/org override attempts.\n- Added the x-tenant-id reverse test to the real DB integration case.\n\nVerification:\n- node -c plugins/plugin-attendance/index.cjs\n- pnpm --filter @metasheet/core-backend type-check\n- pnpm --filter @metasheet/core-backend exec vitest --config vitest.integration.config.ts run tests/integration/attendance-plugin.test.ts -t "attendance rules /me" --reporter=dot (local collect; DB unavailable so skipped)\n- CI: test (18.x), test (20.x), coverage all green.\n\nSubagent review residual risk is closed; PR is back to CLEAN.

PR #3280, opened 2026-06-26
feat(attendance): add makeup punch policy settings

zensgit:
## Summary\n- add dormant makeupPunchPolicy to attendance settings with bounded defaults and enum-strict normalization\n- wire the shape through normalizeSettings, mergeSettings, and PUT/GET schema validation\n- add real settings wire regression coverage for full round-trip, nested partial merge, array replacement, invalid timezone, and invalid supplied values\n\n## Scope / sequencing\nThis is MP-1 only: config/default/normalize/schema/merge. No request create/update/runtime path reads makeupPunchPolicy yet.\n\nDepends on owner review/merge of the MP-0 design-lock (#3276) before this should merge.\n\n## Verification\n- node -c plugins/plugin-attendance/index.cjs\n- pnpm --filter @metasheet/core-backend type-check\n- pnpm --filter @metasheet/core-backend exec vitest --config vitest.integration.config.ts run tests/integration/attendance-plugin.test.ts -t "makeupPunchPolicy" --reporter=dot (collects locally; no local DB, so CI is the real DB gate)\n- git diff --check

### DIFF
--- a/packages/core-backend/tests/integration/attendance-plugin.test.ts
+++ b/packages/core-backend/tests/integration/attendance-plugin.test.ts
@@ -9300,6 +9300,65 @@ attendanceIntegrationDescribe(
       }
     })
 
+    it('round-trips makeupPunchPolicy (wire lock): full shape, nested merge, array replacement, invalid supplied values → 400', async () => {
+      if (!baseUrl) return
+      const runSuffix = Date.now().toString(36)
+      const adminToken = await getAdminToken(`attendance-makeup-policy-wire-${runSuffix}`)
+      expect(adminToken).toBeTruthy()
+      if (!adminToken) return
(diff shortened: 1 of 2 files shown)


zensgit:
CI is green on da52d9579.\n\nVerified locally before push:\n- node -c plugins/plugin-attendance/index.cjs\n- pnpm --filter @metasheet/core-backend type-check\n- targeted attendance integration spec collected clean locally (no local DB, so CI is the real DB gate)\n- git diff --check\n\nGitHub checks now green: test (18.x), test (20.x), migration-replay, contracts, coverage, pr-validate, and neighbors; Strict E2E is the expected skip.\n\nScope remains MP-1 only: dormant settings/default/normalize/schema/merge + wire test. No request runtime reads makeupPunchPolicy. This should still wait on #3276 design-lock review/merge before landing.

PR #3281, opened 2026-06-27
feat(attendance): add report digest policy settings

zensgit:
## Summary
- add dormant `attendanceReportDigestPolicy` settings with default-off report digest subscription config
- wire the shape through `DEFAULT_SETTINGS`, `normalizeSettings`, `mergeSettings`, and the settings PUT schema
- add a real PUT/GET wire-lock integration test for full shape, top-level and nested partial merges, and supplied-invalid rejects

## Verification
- `node -c plugins/plugin-attendance/index.cjs`
- `pnpm --filter @metasheet/core-backend type-check`
- `pnpm --filter @metasheet/core-backend exec vitest --config vitest.integration.config.ts run tests/integration/attendance-plugin.test.ts -t "attendanceReportDigestPolicy" --reporter=dot` (local no-DB skip; CI is the real DB gate)
- `git diff --check`

## Notes
- Stacked on #3275 (RD-0 design-lock).
- RD-1 is latent only: no scheduler registration, no outbox writes, and no delivery worker path reads this setting yet.

PR #3282, opened 2026-06-27
feat(attendance): add report digest period builder

zensgit:
## Summary
- add pure RD-2 report-digest helpers for previous-complete daily/weekly/monthly periods
- add monthly day-of-month clamp helper for future scheduler wiring
- add compact digest payload builder that emits summary/request totals only, not raw records

## Scope
- stacked on #3281 / RD-1
- pure helpers + tests only
- no scheduler, worker, outbox, or runtime producer reads this policy yet

## Verification
- `node -c plugins/plugin-attendance/index.cjs`
- `pnpm --filter @metasheet/core-backend exec vitest run tests/unit/attendance-report-digest-policy.test.ts --watch=false`
- `git diff --check`

### DIFF
--- a/packages/core-backend/tests/unit/attendance-report-digest-policy.test.ts
+++ b/packages/core-backend/tests/unit/attendance-report-digest-policy.test.ts
@@ -1,0 +1,148 @@
+import { describe, expect, it } from 'vitest'
+
+const attendancePlugin = require('../../../../plugins/plugin-attendance/index.cjs')
+const helpers = attendancePlugin.__attendanceReportDigestForTests as {
+  buildAttendanceReportDigestPayload: (input: {
+    cadence?: string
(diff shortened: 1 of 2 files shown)


zensgit:
RD-2 verification note:

- Pure period/payload helper only; no scheduler, worker, outbox, or runtime producer reads this policy yet.
- Local verification:
  - `node -c plugins/plugin-attendance/index.cjs`
  - `pnpm --filter @metasheet/core-backend exec vitest run tests/unit/attendance-report-digest-policy.test.ts --watch=false` → 7/7
  - `git diff --check`
- CI: `pr-validate` green.
- Self-review tightening added before final push: invalid timezone now fails closed instead of relying on Intl fallback / UTC behavior.

PR #3295, opened 2026-06-27
fix(multitable): narrow T9-W Tier 1 sheet_config revert to update+mapped-keys (#3264 follow-up)

zensgit:
## Summary

Follow-up to **#3264** (merged with a [P1]): the T9-W Tier 1 flag-on path keyed off `rev.entity_type === 'sheet_config'` **alone**, bypassing `classifyRevert`'s action/key gating. A forged sheet_config revision with `action='create'|'delete'` (Tier 3 un-create, **held** by #3254) or unknown `changed_keys` was surfaced as confirmable (preview `opKind='safe'`) and skipped the execute 422 — executing an update-shaped revert + appending a restore revision, **violating #3254's hold on Tier 3/4**.

Latent today (the live recorder emits only `action='update'`) but **forgeable** — and the slice already adopts the forged-revision threat model via the `entity_id===sheet_id` guard, so fail-open on a forged `action`/`changed_keys` was inconsistent. The flag `MULTITABLE_ENABLE_SHEET_CONFIG_REVERT` is default-off, so this was not an active breach, but it must be narrowed before the flag is ever enabled.

## Change

- New shared predicate `isSupportedSheetConfigRevert(rev)` in `config-restore.ts` (next to `classifyRevert`): `entity_type === 'sheet_config' && action === 'update' && non-empty changed_keys ⊆ mapped read-rule columns`. **Set membership** (not `key in obj`) so a forged `'toString'`/`'constructor'` key can't match via the prototype chain; `Array.isArray` guards a malformed forged row.
- Gate **both** the preview `opKind='safe'` override **and** the execute 422-skip on the predicate — they can never diverge. Unsupported sheet_config now returns a clean **422 `RESTORE_NOT_SUPPORTED`** (was: 500-via-throw, or an update-shaped write).
- Preview redaction is unchanged — it still runs for **every** sheet_config preview, so a gated preview can't leak rule literals.

`classifyRevert` stays pure. No behavior change for the supported Tier 1 happy path, view/field reverts, or the flag-off/gate/entity_id/drift paths.

## Verification

- **DB-free unit lock** (runs in the default `test` job): `tests/multitable-config-restore-narrowing.test.ts` — supported shape, create/delete, unknown/mixed/empty keys, proto-chain footgun, wrong entity_type, malformed non-array `changed_keys`. 6/6 green locally.
- **Real-DB golden (h)** in `multitable-sheet-config-revert-realdb.test.ts`: flag-on, `canManageSheetAccess`, `entity_id===sheet_id` — forged `create`/`delete`/unknown-key each → preview not-`safe` + execute 422, **zero writes + no `source='restore'` revision**.
- `tsc --noEmit`: 0 errors.

## Scope

Default-off flag unchanged; enabling it stays a separate runbook step (now safe-gated by this narrowing). Tier 2 (lossy field retype) is a separate slice.

### DIFF
--- a/packages/core-backend/src/multitable/config-restore.ts
+++ b/packages/core-backend/src/multitable/config-restore.ts
@@ -147,9 +147,13 @@ const SHEET_CONFIG_COLUMN: Record<string, ColumnMap> = {
   conditionalReadRules: { col: 'conditional_read_rules', jsonb: true },
   rowLevelReadPermissionsEnabled: { col: 'row_level_read_permissions_enabled' },
 }
-// The ONLY sheet_config revert shape this slice supports. A Set (not `key in obj`) so a forged `changed_keys`
-// entry like 'toString'/'constructor' can't match via the prototype chain.
-const SHEET_CONFIG_REVERT_KEYS = new Set(Object.keys(SHEET_CONFIG_COLUMN))
+// The Tier-1 revert surface — the ONLY sheet_config keys this slice will revert. Deliberately an EXPLICIT literal,
+// NOT derived from Object.keys(SHEET_CONFIG_COLUMN): growing that column map (for any reason) must NOT silently widen
+// what's revertible. Adding a key here is a Tier-1 SCOPE change — update the T9-W design-lock + add goldens FIRST;
+// the unit test pins this exact set as a tripwire. A Set (not `key in obj`) so a forged 'toString'/'constructor'
+// changed_key can't match via the prototype chain. Every key here must also exist in SHEET_CONFIG_COLUMN (else
+// applyConfigRevert throws) — the happy-path real-DB golden exercises that direction.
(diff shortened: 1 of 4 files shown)


zensgit:
Closing as **superseded by #3291**, which already landed the identical runtime narrowing on main: `isSupportedSheetConfigRevert` gating both the preview `opKind='safe'` override and the execute 422-skip, the over-broad `entity_type === 'sheet_config'` path removed, `SUPPORTED_SHEET_CONFIG_REVERT_KEYS` as an explicit literal, and the real-DB create/delete/unknown-key golden (gated preview, execute 422, no write, no restore revision).

Merging this would only add duplicate-implementation + constant/comment-naming churn. The two test locks that #3291 didn't include — a DB-free predicate unit test (incl. the prototype-chain footgun) and an exact-set tripwire — are salvaged tests-only in #3297.

PR #3325, opened 2026-06-28
docs(multitable): T9-W Tier 3 (U-3) un-create design-lock (PROPOSED)

zensgit:
**Docs-only design-lock (PROPOSED).** Unblocks the umbrella lock's `🔒 U-3 (HOLD) Tier 3 un-create — until a separate destructive-delete sign-off`. **No runtime here.**

### What Tier 3 is
Config-restore is **revision-scoped**; `classifyRevert` routes every `create`-action revision to `gated` → execute 422 today. Tier 3 turns "revert a `create` revision" into a real **un-create** (drop the entity that revision created), behind a new default-off flag. **Per-revision, not PIT-bulk** — no config-at-T reconstruction (that's the deferred config-reset). Inverse of Tier 4 (undelete), so this sets up Tier 4.

**Premise verified:** forward creates record `action:'create'` rows (field `9672`/`fieldCreateDiff`, view `10445`/`11170`/`configCreateDiff`) with `before=null`, `after={full config}` — so there's a real row to revert.

### v1 surface — LOCKED NARROW
IN: a single `create` revision for `field` or `view` only. OUT (each separate): `delete`→undelete (Tier 4), permission create/delete (permission-revert, held), `sheet_config` create, bulk/PIT config-reset, cross-base, FE.

### The five merge-bar findings → locks
- **[P1] absorb #3323 + rebase** → runtime gate ②; inherits the schema-only/lossless Tier-2 wording (no "Tier 2 lossy" revived).
- **[P1] narrow v1** → v1 surface + U3-L1 (field:create + view:create only).
- **[P1] real field delete sink** → U3-L2 (reuse `DELETE /fields/:fieldId`'s full cascade — meta_fields + auto-number seq + meta_links + strip column from `meta_records.data` + reorder trailing fields + clean referencing view configs, one `batchId`; extract a txn-parameterized `dropFieldCascade`/`dropViewCascade` shared with the forward route, behavior-preserving).
- **[P1] data-loss + no-oracle** → U3-L5 (**undisclosed, no counts** — a per-cell/row count leaks row existence under read-deny; first load-bearing U-L8 use in T9-W).
- **[P2] preview identity binds the deletion plan** → U3-L4 (the existing `driftConflict` false-409s on a benign post-create rename and `baselineHash` can't see new view refs / new column data → purpose-built **plan digest**: entity-alive + cascade set + column-data marker; execute re-enumerates, 409 on drift).

### Awaiting ratification
Three owner OKs gate runtime: the undisclosed no-oracle choice (U3-L5), the hard-drop/no-undo recoverability statement (U3-L7), and the shared-delete-helper refactor that touches the forward route (U3-L2). Plus #3323 merged/rebased, and no live parallel T9-W runtime session.


zensgit:
**Amended per your REQUEST CHANGES — U3-L4 oracle leak fixed.**

The blocking finding is resolved. U3-L4 no longer binds any raw plan field into the client-decodable JWT or the preview response:
- The drift control is now a **single opaque `planHash`**. The raw cascade set / order-shift set / column-data-presence marker are **server-side-only inputs** to the digest — never claims, never response fields.
- **HMAC-keyed, not plain sha256**: I locked HMAC-with-server-secret specifically because the presence marker is ~1 bit, so a plain `sha256` (the existing `baselineHash` style) in a decodable token would let a client hash both possibilities and confirm whether the hidden column has data. HMAC closes that.
- Execute recomputes server-side and returns **one generic `PLAN_DRIFT`** — no sub-reason differentiating new-view-ref / new-column-data / cascade-changed / entity-gone (a sub-reason is itself an oracle).
- `planHash` **excludes cosmetic config** (e.g. `name`), so a benign post-create rename does not false-409; the un-create path does not apply the existing `baselineHash`/`driftConflict` 409.

Test matrix updated: plan-drift cases all assert the same `PLAN_DRIFT`; added a **token/response-opacity** golden (decode the JWT + inspect the response → assert only the opaque `planHash`, no raw plan fields/counts).

**Sequencing accepted**: merge order locked as #3323 → rebase #3325 → merge #3325 (runtime gate ②). The other three owner decisions (undisclosed/no-counts, hard-drop/no-undo, sanctioned shared-helper extract with forward-delete behavior pinned) are recorded as accepted.

PR #3335, opened 2026-06-28
feat(multitable): T9-W Tier 3 (U-3) un-create runtime (default-off)

zensgit:
Implements the ratified Tier 3 design-lock (#3325) — reverting a field/view config `create` revision **un-creates** (drops) the entity it created. **Default-off** (`MULTITABLE_ENABLE_CONFIG_UNCREATE`); inert until a runbook enables it. Per-revision, not PIT-bulk.

### Locks → code (U3-L1..L8)
- **isSupportedUncreate** (pure; field/view `create` only) — `classifyRevert` stays pure; sheet_config/permission creates still 422.
- **dropFieldCascade / dropViewCascade** — the forward `DELETE /fields|/views` sinks extracted txn-parameterized and **shared**. Forward routes call them with `source='mutation'` (behavior-preserving — response/error handling unchanged); un-create calls them with `source='restore'` + `restored_from_id`. Field cascade = delete field + auto-number seq + `meta_links` + strip column from every `record.data` + order-shift + per-view config cleanup, one `batchId`.
- **Opaque HMAC `planHash`** identity (disjoint `config-uncreate-preview`) — binds entity-alive + cascade view-id set + order-shift set + column-data-presence **server-side only** (never claims/response). HMAC-keyed (the ~1-bit presence input would be brute-forceable under a plain hash in the decodable JWT). Excludes cosmetic config → no benign-rename false-409. Execute recomputes under `FOR UPDATE` → one **generic `PLAN_DRIFT`**.
- **Undisclosed no-oracle preview** (no counts, no raw plan); flag + entity-cap (`canManageFields`/`canManageViews`) + typed `confirm:'uncreate'` + single-txn atomicity; realtime parity with forward delete.

### Goldens (12, real-DB, registered in `plugin-tests.yml`)
(a) flag-off 403 · (b) field perm via view-only actor 403 · (c) happy field un-create (full cascade + one `batchId` + `source='restore'` delete-rev) · (d) happy view un-create · (e) **forward-delete parity** (shared cascade, `source='mutation'`) · (f) **plan-drift → one generic `PLAN_DRIFT`** (new view-ref / new column data / new trailing field) · (g) **no benign-rename false-409** · (h) **token/response opacity** · (i) **no-oracle no-count** · (j) typed-confirm · (k) **single-txn atomicity rollback** · (l) audit delete-rev.

### Gate before enabling the flag
Watch goldens (c)/(k) (destructive write + atomicity) in CI; flag-on live smoke before any rollout. Tier 4 undelete / permission-revert remain out (each its own design-lock).


zensgit:
**Both Mediums + the precision note are resolved as of the current head `aceb878b5`** (the review above is pinned to `823789ea3`, the commit *before* the fix — a refresh should pick it up). The only commit since `823789ea3` is the fix; the un-create real-DB suite is **green in CI** (`✓ multitable-uncreate-config-realdb.test.ts (15 tests)`).

- **Medium 1 (sheet-consistency guard)** → golden **(m)** `sheet-consistency guard: a revision whose entity lives on ANOTHER sheet → 400 INVALID_REVISION, zero drops` (test file line ~394). Seeds a malformed `create` revision recorded under one sheet but pointing at a field on another → execute `400 INVALID_REVISION`, field + column data untouched. Guard runs before the planHash verify, so the token is moot.
- **Medium 2 (atomicity overclaim)** → golden **(n)** `late-cascade atomicity: a failure at the column-strip (AFTER field+links delete) rolls the WHOLE cascade back` (line ~410) forces the failure at `UPDATE meta_records` (after `DELETE meta_fields` + `DELETE meta_links`) and asserts `meta_links` is restored. **(k)** renamed to `early-drop atomicity` so neither test overclaims.
- **Precision note** → dev-verification MD now states column-data drift fires only when `columnDataPresent` flips **false→true** (the coarse no-oracle bit), not on every value change. (f)(ii) already tests exactly that flip.

Default-off flag unchanged. Happy to walk through any of the three on `aceb878b5`.

PR #896, opened 2026-04-17
feat(dingtalk): close self-bind, stable user deep-link, and mobile CAS loop

zensgit:
## What changed

This PR closes three reviewed DingTalk gaps:

- completes self-service DingTalk bind for already-authenticated users
- stabilizes `/admin/users?userId=...` deep-links when the target row is outside the first paginated page
- closes the mobile profile CAS loop so concurrent admin edits no longer overwrite silently

## User-facing behavior

- Starting DingTalk bind from settings now completes a true bind flow for the current logged-in account.
- DingTalk callback always posts back to the backend when `code` is present, and the frontend now branches on backend `data.mode`:
  - `bind`: keep current session/token, redirect back to settings
  - `login`: preserve existing login behavior
- Directory-managed users are explicitly blocked from self-bind with `409`.
- User management deep-links no longer depend on the target user being present in page 1.
- Admin mobile edits now send an `expectedMobile` CAS witness.
- On `PROFILE_MOBILE_CONFLICT`, the UI refreshes access and shows the true latest value if refresh succeeds; if refresh fails, it falls back to a neutral retry message instead of showing stale data.

## Backend changes

- DingTalk OAuth state now carries bind intent and `bindUserId`
- `GET /api/auth/dingtalk/launch` supports authenticated bind mode
- `POST /api/auth/dingtalk/callback` now supports explicit `bind` vs `login` branching
- self-bind is blocked when the current user is directory-managed
- `GET /api/admin/users` supports deep-link pinning via `userId`
- `PATCH /api/admin/users/:userId/profile` supports `expectedMobile` CAS
- mobile CAS SQL guard now normalizes both sides to avoid false conflicts on legacy whitespace-formatted phone data

## Frontend changes

- `DingTalkAuthCallbackView` no longer relies on a frontend-local bind hint; backend `data.mode` is now the source of truth
- `SessionCenterView` launches bind without writing local intent state
- `UserManagementView`
  - supports stable deep-link focus by `userId`
  - sends `expectedMobile` on profile save
  - handles `PROFILE_MOBILE_CONFLICT` with refresh + correct messaging

## Files included

- `apps/web/src/views/DingTalkAuthCallbackView.vue`
- `apps/web/src/views/SessionCenterView.vue`
- `apps/web/src/views/UserManagementView.vue`
- `apps/web/src/utils/browserLocation.ts`
- `apps/web/tests/dingtalk-auth-callback.spec.ts`
- `apps/web/tests/sessionCenterView.spec.ts`
[...]

PR #3340, opened 2026-06-28
fix(plm): remove aspirational BOM write-back pact until provider slice lands

zensgit:
## Why

#3332 (方案1) froze a `PATCH /api/v1/bom/multitable/{partId}/lines/{bomLineId}` write-back interaction in the **consumer pact** and added `PLMAdapter.updateBomMultitableLine`. But **no production path calls that method** and **the provider endpoint does not exist** — so the Yuantus pact-broker `can-i-deploy --pacticipant Metasheet2 --main-branch` gate verifies the interaction and gets a **404**, reddening every Yuantus PR (e.g. provider-side #896).

The consumer pact's own contract is to freeze **only what `PLMAdapter` actually calls — never aspiration**. This interaction violated that, so it's removed until the real call exists.

## What (surgical — not a full #3332 revert, per owner decision)

- Remove the write-back interaction from the **frozen pact JSON** (34 → 33 interactions).
- Remove the PATCH path from `PLM_ADAPTER_PACT_PATHS` + its pact test.
- Remove `PLMAdapter.updateBomMultitableLine` + `BomMultitableLinePatch` / `BomMultitableLineUpdateResult` + the unit-test block — so consumer `main` stops modelling a call to an unbuilt provider.
- **Keep** the 方案1 design-lock doc, annotated **SUPERSEDED by #896 Slice 0**.

## Verification

- `tests/contract` green (19/19); `tests/unit/plm-adapter-bom-multitable` green (10/10).
- `tsc --noEmit` clean (0 errors) — no dangling imports of the removed interfaces.
- Pact JSON valid; 0 write-back interactions remain; only these 5 files changed.

## Sequencing (how #896 goes green)

1. Merge this → 2. metasheet2 CI republishes the cleaned `main` pact to the broker (publish on `push:main`) → 3. re-run #896's contracts (won't auto-rerun). #896 is green only after all three.

## ⚠️ Cross-session alignment required

This corrects a **merged, design-locked parallel line**. The 方案1 session must align to the **provider-first** path: the write-back returns as a **real** pact in the consumer-relay slice, *after* the provider write slice ships as one complete cut (endpoint + replay guard + domain AuditLog/idempotency + ECO create-intent semantics). If the PATCH interaction is re-added to `main` before the provider exists, the Yuantus broker reddens again.

**Owner-gated: do not auto-merge.** Merge after CI green + explicit owner go.


zensgit:
Closing as a **superseded duplicate**. #3337 already depublished the aspirational BOM write-back interaction from the main pact (broker unblocked; #896 restored to MERGEABLE/CLEAN, contracts pass), which is the substance of this PR.

Note for the record: #3337 kept `PLMAdapter.updateBomMultitableLine` as a dead client method (no caller; provider endpoint unbuilt) plus two 'PLMAdapter.ts keeps the client method' comments and the `endpointsToFind` assertion. The rebased version of this branch additionally removed that dead method/types/unit-tests/comments. That residual cleanup is **harmless to defer** (the method is not in the pact and has no caller); if you want it removed from main later, it's a ~4-file no-logic deletion I can reopen as a tiny follow-up. Not merging this to avoid the confusion you flagged.

PR #3347, opened 2026-06-28
feat(multitable): date-reminder timeOfDay-aligned scheduling + scanIntervalMs decouple (follow-up to #3329)

zensgit:
## What & why
Follow-up to #3329 (`schedule.date_field`). Closes the owner-found **P2**: `timeOfDay` only gated the due-predicate — it did **not** drive when the scan ran. The boot-anchored `setInterval` (default 24h) fired a 09:00 rule whenever the daily tick landed relative to server boot (boot 03:00 ⇒ ~18h late), the cadence was untested, and `scanIntervalMs` was load-bearing twice (cadence **and** `scanWindow = 2×cadence`).

Built to the owner-ratified design-lock (`docs/design/multitable-date-reminder-timeofday-alignment-design-lock-20260628.md`): option 1, grace = 48h, decouple `scanIntervalMs`.

## Changes (DR-A…E)
- **DR-A** — `schedule.date_field` now arms a `setTimeout` to the next UTC `timeOfDay` boundary, re-arming daily (re-arm guarded by `this.timers.get(id) === timer` so unregister/replace stops it). No boot anchor. `setInterval` retained for cron/interval (callback extracted to `runScheduledCallback`).
- **DR-B** — bounded catch-up: if today's `timeOfDay` already passed at register/restart, run **one** immediate scan — still gated by `occ ≥ ruleCreatedAt` + the recent window, so never a backfill-blast.
- **DR-C** — claim/fire ledger unchanged: PK `(rule_id, record_id, occurrence_ts)` + `ON CONFLICT DO NOTHING RETURNING`, at-most-once.
- **DR-D** — window is a fixed `DATE_REMINDER_GRACE_WINDOW_MS = 48h` constant; `dateReminderScanWindowMs`/`dateReminderScanIntervalMs` removed; `scanIntervalMs` retained-but-`@deprecated`/ignored so it can no longer distort the window or masquerade as the trigger time.
- **DR-E** — UI hint + accurate label/comments (the `setInterval` fix makes "触发时间(UTC)" true).

## Verification
- **Fake-timer golden 5/5** (NEW — closes the untested-cadence gap): boot-03:00 no-fire / fires-at-09:00 / restart-10:00 catch-up-once / re-arm-next-day no-same-day-double / unregister-cancels (+ non-leader no-arm). G1 fails on the old `setInterval`.
- pure-core **17/17**, scheduler suites **32/32** (no cron/interval regression), real-DB **11/11** (DR-1/2/3/6 + VAL — proves the 48h-constant window = old 2×24h didn't move scan/claim/fire), editor spec **97/97** (coexists with #3336 form.submitted), backend `tsc` + frontend `vue-tsc` clean.

## Boundary (named)
DR-B runs one scan per `date_field` rule on register, so a leader boot/failover after `timeOfDay` with N rules kicks ~N near-simultaneous catch-up scans (each bounded + idempotent; a future slice can stagger/coalesce). Unchanged deferred: tz-aware bucketing, sub-day offsets, ledger retention.


zensgit:
### CI status — the red is a pre-existing attendance flake, not this diff

The failing `test (18.x / 20.x)` jobs fail in the **attendance integration suite**, a sibling step in the same job — **not** in this PR's date-reminder lane (the `multitable-date-reminder-trigger.test.ts` plugin-tests step passes).

**Proven unrelated (A/B on the exact CI attendance suite, 10 files / one process):**
- this branch's source → 1–2 failures (`shift-compliance cap`, varying set)
- `origin/main`'s source with the date-reminder work fully reverted → **13** failures (a different, larger set)

A failing set that changes every run and gets *worse* when this diff is removed is shared-state / ordering nondeterminism in the attendance suite itself — independent of date-reminder. Re-running has not cleared it (3/3 red).

**Plan (agreed with maintainer):** a separate narrow CI-health PR fixes the attendance suite's test-isolation; once it merges, this PR rebases onto it and goes green through the front door. Holding — not admin-merging.

The P1 (`effectiveAt` activation floor) is closed as specified: server-stamp on create + conversion, preserve on stay/edit, never client-trusted, fallback to `createdAt` for pre-fix rules; covered by real-DB DR-7 + 4 floor units.

PR #3362, opened 2026-06-29
feat(automation): add date reminder ledger retention

zensgit:
## Summary

Adds bounded retention for the `schedule.date_field` idempotency ledger:

- fixes the v1 retention choice at 365 days;
- ages `meta_automation_date_reminder_fires` by `fired_at` rather than `occurrence_ts`;
- exposes `AutomationService.sweepDateReminderLedger()` as an ops/test seam;
- runs a best-effort once-per-day opportunistic sweep from active date-reminder scans, without adding a new scheduler;
- updates the original date-reminder docs and migration comment so they no longer claim ledger retention is unbuilt.

At-most-once delivery and the `(rule_id, record_id, occurrence_ts)` dedupe key are unchanged.

## Verification

Local:

- `pnpm -C packages/core-backend exec vitest run src/multitable/automation-date-reminder.test.ts --reporter=dot` → 23/23
- `pnpm --filter @metasheet/core-backend type-check` → clean
- `git diff --check HEAD~1..HEAD` → clean

Real-DB:

- Added `DR-RET` to `tests/integration/multitable-date-reminder-trigger.test.ts`: inserts old and recent ledger rows, runs the sweep, and proves only the old `fired_at` row is removed.
- Local dev Postgres could not run the suite because this checkout's local 5435 DB is not migrated to `meta_bases` (the migration chain stops on an older approval migration). CI's DB lane is the required proof for this case.

### DIFF
--- a/docs/design/multitable-date-reminder-ledger-retention-design-lock-20260629.md
+++ b/docs/design/multitable-date-reminder-ledger-retention-design-lock-20260629.md
@@ -1,0 +1,35 @@
+# Multitable date-reminder ledger retention — design lock (2026-06-29)
+
+> Status: RATIFIED + BUILT in this slice. Owner chose the 365-day retention default for the
+> `schedule.date_field` idempotency ledger.
+
+## 1. Problem
+
+`meta_automation_date_reminder_fires` is the claim ledger that gives date reminders their at-most-once
+property. Each fired occurrence writes `(rule_id, record_id, occurrence_ts)` before executing the automation,
+and later scans skip duplicates by the primary key. Rule deletion cascades ledger rows, but a long-lived active
+rule can accumulate rows forever.
+
+## 2. Decision
+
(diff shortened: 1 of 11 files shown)


zensgit:
Addressed the review P2 in `f0df0d609`: added a new migration `zzzz20260628120200_add_date_reminder_fires_fired_at_index` creating `idx_date_reminder_fires_fired_at`, and updated the design/dev docs to state the retention sweep is backed by that index. This is intentionally a new migration, not an edit to the shipped table migration DDL.\n\nLeft the P3 batching note as a follow-up boundary: the daily cadence keeps backlog risk low, and batching can be added separately if retention sweeps ever need chunking under large accumulated ledgers.\n\nRe-verified after rebase: pure date-reminder unit 23/23, backend type-check clean, diff whitespace clean.

PR #901, opened 2026-04-18
feat(dingtalk): link projected member groups to default governance

zensgit:
## What changed

This stacked PR extends DingTalk directory department projection so projected `platform_member_groups` can also apply a safe default governance baseline to already linked local users.

Added integration config:

- `memberGroupDefaultRoleIds`
- `memberGroupDefaultNamespaces`

Behavior:

- selected DingTalk departments still project to `platform_member_groups`
- linked local users inside those projected groups can now receive additive default governance during sync
- only missing grants are filled
- no automatic revocation is introduced in this round

## Safety constraints

Backend validation blocks unsafe config:

- configured roles must exist
- `admin` is rejected
- delegated-admin style roles are rejected
- namespaces must be admission-controlled resources

## User-facing changes

Directory integration settings now include:

- `成员组默认业务角色`
- `成员组默认插件开通`

Sync success feedback now reports:

- governed user count
- added role assignment count
- added namespace admission count

## Files

- `packages/core-backend/src/directory/directory-sync.ts`
- `packages/core-backend/tests/unit/directory-sync-member-group-projection.test.ts`
- `packages/core-backend/tests/unit/admin-directory-routes.test.ts`
- `apps/web/src/views/DirectoryManagementView.vue`
- `apps/web/tests/directoryManagementView.spec.ts`
- `docs/development/dingtalk-directory-member-group-governance-linkage-development-20260418.md`
- `docs/development/dingtalk-directory-member-group-governance-linkage-verification-20260418.md`
- updated design docs for directory sync governance

## Verification

```bash
pnpm install --frozen-lockfile
pnpm --filter @metasheet/core-backend exec vitest run tests/unit/directory-sync-member-group-projection.test.ts tests/unit/admin-directory-routes.test.ts --watch=false
pnpm --filter @metasheet/web exec vitest run tests/directoryManagementView.spec.ts --watch=false
pnpm --filter @metasheet/core-backend build
pnpm --filter @metasheet/web build
```

Results:

- backend tests: `23 passed`
- frontend tests: `32 passed`
- backend build: passed
- web build: passed

## Notes

This PR is intentionally additive and stacked on top of #900.
It does not yet implement automatic governance revocation or role-scope templates.

PR #3367, opened 2026-06-29
feat(plm): re-add Phase-7 BOM write-back pact interaction + Idempotency-Key (#3332 follow-up)

zensgit:
**#3332 follow-up.** The Yuantus provider write-back endpoint **landed** (#905, ratified design #901), so this re-publishes the governed PATCH interaction that #3337 temporarily depublished while the provider was unbuilt.

## What
- **pact**: re-add `PATCH /api/v1/bom/multitable/{W1}/lines/{W3}` → `200 {ok, bom_line_id}`, providerState naming the **distinct write SKU** `plm.bom_multitable_writeback`, with an `Idempotency-Key` header. The interaction's request/response/providerStates **byte-match** the provider's checked-in (#905) interaction, so the broker provider-verify passes. `PACT_PATHS` + the documenting `it()` restored.
- **`PLMAdapter.updateBomMultitableLine`** now **sends a per-edit `Idempotency-Key`** (`crypto.randomUUID`) — the provider 400s without it (#901 §2); merged on top of the connection auth/tenant headers by `HTTPAdapter.select`.

## Effect / safety
On `push:main`, the Yuantus Pact Consumer republishes the main pact **with** this interaction; the Yuantus provider (already on main) then verifies it green — closing the #3332 → #3337 → #905 loop. The interaction is aligned exactly to the provider's seeded W1/W3 fixture, so it will **not** re-red the blocking broker Phase-B gate.

Verify: PACT_PATHS↔JSON exact ordered match (34); write-back request/response/providerStates equal the provider's verified shape.

### DIFF
--- a/packages/core-backend/src/data-adapters/PLMAdapter.ts
+++ b/packages/core-backend/src/data-adapters/PLMAdapter.ts
@@ -1,3 +1,4 @@
+import { randomUUID } from 'node:crypto'
 import { Optional } from '@wendellhu/redi'
 import { IConfigService, ILogger } from '../di/identifiers'
 import { HTTPAdapter } from './HTTPAdapter'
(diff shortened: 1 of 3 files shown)


zensgit:
Redundant: the Phase-7 write-back pact interaction was re-added to metasheet2 main independently (3e349cf2, republished to the broker at 07:54). The consumer loop is already closed; this PR would duplicate/conflict with the landed interaction. Closing. (Yuantus Phase-B re-verify against the new 34-interaction pact is being driven separately.)

PR #3400, opened 2026-06-30
fix(multitable): row-lock the permission-revert live-grant load (never-escalate-under-concurrency)

zensgit:
## What

Closes the documented concurrency gap in the **flag-gated** multitable permission-revert execute path (`MULTITABLE_ENABLE_PERMISSION_REVERT`, **default off**). The execute path re-checks the de-escalation direction against the **LIVE** grant inside its txn, but the live-grant load was a plain `SELECT`. A concurrent forward grant/revoke write that **lowered** the grant between the revert's direction check and its apply could turn a checked de-escalation into a **net escalation** (e.g. forward revokes `admin→none`, then the revert "de-escalates" `admin→read` = `none→read`, raising the subject).

The original code even documented this: the `meta_sheets FOR UPDATE` carried a comment that "the forward grant/revoke routes must take the same lock for full coverage — tracked in the dev-verification honest gaps."

## Fix (confined to the gated path, no forward-route change)

The execute-path live-grant load now takes **`FOR UPDATE`** (`loadLivePermissionGrant` gains a `forUpdate` flag, set `true` **only** on the execute path; preview stays unlocked). A de-escalation always reads a **non-null** live grant (no-row ⇒ `null` ⇒ rank 0, and de-escalation requires live rank ≥ 1), so the `FOR UPDATE` always locks a **real row**; a concurrent forward write on the same subject contends on that row, so the drift re-check → direction re-check → apply are **atomic** against it. The existing `meta_sheets FOR UPDATE` is **kept** (coarse per-sheet mutex; also serializes the sheet-delete FK cascade — see below); lock order is always `meta_sheets → permission-row`, so concurrent reverts can't deadlock. Under **READ COMMITTED** (verified the pool sets no isolation override), a blocked `FOR UPDATE` re-reads the committed-forward value → drift → **409**.

## Why row-lock over CAS

`FOR UPDATE` reuses the existing `GRANT_DRIFT` check verbatim (no new `rowCount==0` branch), avoids per-scope NULL-normalization/delete-then-insert CAS fiddliness, and needs zero forward-route change. The "field rank-2 = no-row" subtlety that would have forced CAS gymnastics **doesn't arise**: the loader returns `null` for no-row, and de-escalation never reads `null`.

## Verification

- **15/15** `multitable-permission-revert-realdb` goldens pass against a real Postgres, including **new (m)/(n)**: a forward **REVOKE** (sheet) / **HIDE** (field) committed **WHILE** the revert executes — staged deterministically via a held row lock — makes the revert **re-read the lowered grant and 409 `GRANT_DRIFT`** instead of escalating. (The serial drift case (h) does *not* exercise the lock; (m)/(n) overlap two txns.)
- File is already in the `plugin-tests.yml` real-DB allowlist (line ~232) → (m)/(n) ride along.
- `tsc --noEmit` clean.
- **Forward-path serialization audit** (independent): every forward path that *lowers/removes* a field/view/sheet grant does so via UPDATE/DELETE on the **same lock key** → all contend. Sheet-delete FK cascade is serialized by the kept `meta_sheets FOR UPDATE`. **No** `permission-row → meta_sheets` lock order anywhere → deadlock-free.

## Scope / non-goals

- **Does NOT enable the flag.** This is the engineering prerequisite that makes a future *owner-gated* enablement safe.
- The legacy `/permissions/grant` INSERT-without-DELETE path (sheet scope) is non-contending but **benign** for never-escalate: the **revert's own writes never raise the subject above the live it direction-checked**. The final *derived* level isn't guaranteed to equal the target — a concurrent grant of a higher `perm_code` that commits after the revert's apply-`DELETE` survives — but that higher value is an **independent authorized grant** (forward-vs-forward last-write-wins), not a revert-blessed escalation. Field/view are single-PK (fully serialized; no new-row escape). Optional future owner-gated hardening if grant/revert mutual exclusion is ever wanted.

### DIFF
--- a/docs/development/multitable-permission-revert-concurrency-rowlock-dev-verification-20260629.md
+++ b/docs/development/multitable-permission-revert-concurrency-rowlock-dev-verification-20260629.md
@@ -1,0 +1,109 @@
+# Permission-revert concurrency hardening — row-lock the live-grant load (dev + verification)
+
+**Date:** 2026-06-29  **Grounding:** `origin/main @ 57fd06962`  **PR:** #3400
+**Branch:** `claude/permrevert-concurrency-rowlock-20260629`
+**Status of the flag:** `MULTITABLE_ENABLE_PERMISSION_REVERT` stays **default-off**. This change enables nothing; it is the engineering prerequisite that makes a future, separately-signed-off enablement safe.
+
+> Positioning: this is a focused engineering hardening of one already-shipped, flag-gated path. It does **not** advance the gated Global History remainder (PIT/T9-W flag enablement, Reset-UI product entry, value-level undelete) — those remain owner-gated per-item decisions, re-listed in §7 for sign-off.
+
+---
+
+## 1. Context — what permission-revert is, and the gap
+
+The Global History line includes a **config-history → config-restore** capability. One restorable kind is a `permission` revision: reverting it re-applies the revision's `before` grant, but **only when restoring `before` reduces** the subject's access on the entity's single total-order access rank (field `hidden < read-only < read-write`; view `none < read < write < admin`; sheet `none < read < write-own < write < admin`). Escalation / no-op is refused fail-closed (422). The path sits behind the default-off flag `MULTITABLE_ENABLE_PERMISSION_REVERT`, on top of the `canManageSheetAccess` capability floor and a typed `revert-permission` confirm.
+
+The load-bearing safety property is **never-escalate**: the execute path re-checks `permissionRevertDirection(before, LIVE)` against the **current live grant** (not the recorded `after`), inside its transaction. A server-minted preview token binds the live grant via an HMAC `currentGrantHash`, so a grant changed between preview and execute → 409 `GRANT_DRIFT`.
+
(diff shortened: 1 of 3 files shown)


zensgit:
Closing as **superseded by #3402**.

#3402 (merged to main, commit 3dbc4684) adopts **option (a)** for this concurrency gap: all three forward grant/revoke routes now take the same \`meta_sheets … FOR UPDATE\` the permission-revert execute path holds → two-sided serialization. It also adds its own \`(m)\` double-writer golden to \`multitable-permission-revert-realdb.test.ts\`. The engineering gate this PR targeted is therefore **already closed on main**.

This PR took **option (c)** — row-lock the revert's execute-path live-grant load (\`forUpdate\`). With #3402 on main it is redundant runtime, its \`(m)/(n)\` goldens collide with #3402's \`(m)\`, and its dev-verification doc now reads contrary to main ("no forward-route change needed" / "option (a) rejected" — option (a) is exactly what shipped). Rather than rework this branch and risk dragging the superseded reasoning back in, closing it clean.

Any further hardening goes in a **fresh** PR — e.g. #3402's explicitly-deferred legacy \`spreadsheet-permissions.ts\` grant/revoke route (the remaining un-serialized writer), locked #3402-style — not a rework of this one.

PR #3410, opened 2026-06-30
fix(multitable): lock legacy spreadsheet-permissions writer — close permission-revert writer-set

zensgit:
## What & why

Completes the writer-set guarantee that **#3402** started. #3402 serialized the univer-meta forward grant/revoke routes against `permission-revert`, but left **one residual writer**: the legacy `/api/spreadsheets/:id/permissions` grant|revoke route (`routes/spreadsheet-permissions.ts`) wrote `spreadsheet_permissions` via **bare autocommit `pool.query`, no lock** — so a grant/revoke through it could still commit between a revert's live-grant re-check and its apply, turning a re-checked de-escalation into a **net escalation**.

## The fix (`routes/spreadsheet-permissions.ts`)

Both DB-branch writes (grant INSERT, revoke DELETE) now run inside `poolManager.get().transaction()` taking `SELECT 1 FROM meta_sheets WHERE id=$1 FOR UPDATE` **first** — the same row lock the revert execute path and the univer-meta grant routes take. Key facts that make this correct:
- `db/pg`'s `pool` **is** `poolManager.get().getInternalPool()` (same pool, same DB) → the `FOR UPDATE` genuinely serializes across surfaces.
- `auditLog` and the response read stay **outside** the txn (parked window scoped to the lock); same lock order (sheet row → grant rows) → **no new deadlock**.
- This **serializes** the legacy route; it is *not* made a first-class peer (it still records no config revision). Retiring it stays a future option only if confirmed externally dead — a breaking change for an OpenAPI-documented endpoint, so out of scope here.

## Completeness — the audit that justifies "closed"

Whole-backend writer audit of the three grant tables: de-escalation apply helper (under the revert lock ✓), the three univer-meta routes (#3402 ✓), and this legacy route (**now locked** ✓). `permissions.ts` (`/api/permissions/*`, the FE-used route) writes `user_permissions` — a **global RBAC table, not** sheet-scoped — and is correctly out of set. So this was the **sole** remaining un-serialized writer; with it locked, **all writers to `spreadsheet_permissions` / `meta_view_permissions` / `field_permissions` serialize on the `meta_sheets` row.**

## Golden (n) — placed where it actually runs

Added to the **already-CI-wired** `multitable-permission-revert-realdb.test.ts` (a brand-new realdb file would silently not run without a `plugin-tests.yml` edit). Mounts the legacy router in a local app with `req.user={roles:['admin']}` (clears `rbacGuard`'s admin short-circuit), then the **same FK-aware discriminator as (m)**:
- hold **`FOR KEY SHARE`** on the sheet; POST the legacy grant of a **non-pre-seeded** perm_code.
- **with the fix**: the txn requests `FOR UPDATE` → conflicts → grant **parks** (not settled, code absent) until release, then applies.
- **without the fix**: the bare INSERT needs only the FK's `FOR KEY SHARE` → compatible → commits during the hold → **RED**.

Non-seeded probe code → the `ON CONFLICT DO NOTHING` is a real observable write (absent→present), not a no-op. Shared pool `max=20` → no connection-starvation false-green.

## Verification

- `tsc --noEmit` (core-backend) — **clean (exit 0)**, run against the canonical checkout's deps.
- Golden (n) runs in the CI-only realdb suite (same job as (a)–(m)); teeth rest on the FK `FOR KEY SHARE`/`FOR UPDATE` lock-compatibility argument.
- Flag stays **default-off**; no FE, no contract change, no behavior change to the legacy API.

## Docs

Flips dev-verification §2 from **◑ PARTIAL → ✅ CLOSED** (writer-set-complete, no longer a HARD enablement gate); adds a dated dev-verification doc.


zensgit:
Superseded by #3414 (merged) — closing to avoid two colliding legacy-lock branches.

#3414 closes the same residual with the **same lock model** (both grant + revoke writes wrapped in a `meta_sheets … FOR UPDATE` txn), but is **better in two ways** after deep review:

- **More accurate framing.** This PR (#3410) treated grant and revoke as equal gaps. #3414 states it precisely: `spreadsheet_permissions.sheet_id REFERENCES meta_sheets(id)`, so a grant **INSERT** already takes the parent's `FOR KEY SHARE` which conflicts with the revert's `FOR UPDATE` — grant was **already serialized**; locking it is uniformity. The **revoke** (child `DELETE`, no parent lock) is the **genuine** un-serialized gap.
- **More complete golden.** #3410 tested only the grant path (golden n); #3414 tests **both** grant (a) **and revoke (b)** — and (b) is the one that actually proves the real-gap closure (held `FOR KEY SHARE`; the row stays present during the hold and is gone after release; a bare DELETE would have removed it → red pre-fix). #3414 also confirmed 2/3 RED with the runtime reverted.

No unique value here that #3414 lacks. Closing.

Issue #3317, opened 2026-06-27
[Attendance] AE-1b/AE-2/AE-2.1/AE-3 shipped · AE-4 staging smoke pending


zensgit:
**Update 2026-07-01 — AE queue current after #3461/#3462.**

AE-1b corrected-fact durability is CLOSED by #3377 (`d9875212e`, merged 2026-06-30T05:50:37Z). The `attendance_records.meta.manual_result_edit` sticky marker (with `correctedAgainst` over workDate / firstInAt / lastOutAt / isWorkday) is honored by `computeAttendanceRecordUpsertValues`: no-`statusOverride` recomputes preserve the correction, and material fact divergence sets `reviewConflict.state='needs_review'`. Design-lock: #3371. F1 writer-enumeration gate was honored at runtime.

This issue remains OPEN as the umbrella for the remaining AE gated follow-ups, not because AE-1b durability is still undone.

### Current AE follow-up queue

- [x] **AE-2 — notify corrected employee**. Design-lock #3407 + runtime #3413 are on `main`. A successful correction enqueues an affected-employee-only attendance notification row.
- [x] **AE-2.1 — honor `notifyAffectedEmployee`**. Runtime #3419 + design-lock #3423 + test-isolation fix #3425 are on `main`. `notifyAffectedEmployee=false` suppresses the outbox row and records `notification_skipped_reason='policy_disabled'`; default true/omitted enqueues and back-links `notification_delivery_id`.
- [x] **AE-3 — admin anomaly-correction modal**. Design-lock #3422 + runtime #3461 (`b4207da73`, merged 2026-07-01T17:13:58Z) are on `main`. The UI action is available through the ratified explicit capability probe, uses a modal-open snapshot for submit, and widens the correction write path to the existing audited backend route.
- [ ] **AE-4 — staging smoke**. Runbook #3462 is on `main` with status `PREPARED`; actual AE-4 closure requires deploy SHA, staging PASS stamp, and residue=0.

---

_Original AE-1 review context retained below._

From the AE-1 (#3316) adversarial review — APPROVE-WITH-NITS, all 9 ratified locks hold, no P1. Owner decisions and planning-matrix test gaps have now been resolved; the only remaining item in this umbrella is AE-4 staging smoke.

### Owner decisions
[...]

PR #3419, opened 2026-06-30
fix(attendance): honor result-edit notification toggle

zensgit:
## Summary
- honor attendanceResultEditPolicy.notifyAffectedEmployee in the AE-2 result-edit notification producer
- when enabled, enqueue the employee-only delivery and back-link notification_delivery_id to the immutable audit row
- when disabled, skip the outbox row and persist notification_skipped_reason='policy_disabled'
- add a real-DB route regression for the disabled-notify path and tighten the default path to assert the audit delivery backlink

## Verification
- node -c plugins/plugin-attendance/index.cjs
- git diff --check origin/main...HEAD
- pnpm --filter @metasheet/core-backend exec vitest --config vitest.integration.config.ts run tests/integration/attendance-result-edit.test.ts --reporter=dot (local no-DB path: 19 skipped; CI runs this against real Postgres)
- subagent review found P2 skipped-reason literal drift; fixed to the design/migration contract value policy_disabled

Refs #3317

### DIFF
--- a/packages/core-backend/tests/integration/attendance-result-edit.test.ts
+++ b/packages/core-backend/tests/integration/attendance-result-edit.test.ts
@@ -123,6 +123,11 @@ describeDb('AE-1 attendance anomaly result edit (real DB, route-level)', () => {
     }
   }
 
+  function resultEditPolicyWithNotifyAffectedEmployee(policy: unknown, notifyAffectedEmployee: boolean) {
+    const base = resultEditPolicyWithEnabled(policy, true)
+    return { ...base, notifyAffectedEmployee }
+  }
+
   async function seedRecord(input: {
     userId: string
     workDate: string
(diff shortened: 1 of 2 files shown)


zensgit:
Independent re-review complete: APPROVE, no P1/P2/P3 findings.

Verified live head bf22b7481d4a74e7c377d4bcc9d7e07e216866d9 / mergeState=CLEAN. The runtime now honours `attendanceResultEditPolicy.notifyAffectedEmployee`: enabled writes one employee-only delivery with `source_id = auditRow.id`, a `:channel:` source_key, redacted payload, and `notification_delivery_id` back-link; disabled writes no outbox row and records `notificationSkippedReason=policy_disabled` in audit/API. Enqueue remains strict in-transaction with no best-effort try/catch.

CI evidence: required test (18.x) and test (20.x) are green; subagent also checked the workflow/log coverage and confirmed `attendance-result-edit.test.ts` executed against real Postgres rather than skip-shaped green. Ready for owner review/merge.

PR #3422, opened 2026-06-30
docs(attendance): ratify AE-3 admin modal design-lock

zensgit:
## Summary
- propose the AE-3 admin anomaly-correction modal design lock
- lock placement as an anomaly-row action in both existing anomaly surfaces, with no new admin-rail section or anchor-nav count change
- require an overview-safe admin capability predicate instead of gating the overview-mounted anomaly table by showAdmin alone
- require modal-open snapshot/idempotency, no metric overrides in v1, explicit backend error mapping, refresh/stale-clearing behavior, and request-center parity
- gate AE-3 runtime on owner ratification and #3419/equivalent notification-toggle honoring

## Verification
- git diff --check
- text grep: no brand terms, Status remains PROPOSED, no runtime/backend/schema/staging authorization
- subagent review found two P2s (overview capability predicate + missing stale-clearing matrix); both fixed and re-reviewed APPROVE

Refs #3317

### DIFF
--- a/docs/development/attendance-anomaly-result-edit-admin-modal-design-lock-20260630.md
+++ b/docs/development/attendance-anomaly-result-edit-admin-modal-design-lock-20260630.md
@@ -1,8 +1,8 @@
-# Attendance AE-3 — admin anomaly-correction modal — design lock (PROPOSED)
+# Attendance AE-3 — admin anomaly-correction modal — design lock (RATIFIED)
 
-> Date: 2026-06-30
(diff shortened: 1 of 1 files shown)


zensgit:
Design-lock handoff update: #3422 is OPEN/CLEAN and all required checks are green, including test (20.x). It remains PROPOSED/docs-only and does not authorize runtime.

Subagent review found two P2s and both are fixed in the PR: (1) the overview anomaly table must not rely on `showAdmin` alone, so AE-3 runtime needs an explicit overview-usable capability predicate derived from the backend route authority; (2) modal/result/error state must clear synchronously on filter/date/org/user changes or load-start to avoid stale range-A state while range-B loads.

Ready for owner ratification. Runtime remains gated on ratification plus #3419 or equivalent notification-toggle semantics landing on main.

PR #3428, opened 2026-07-01
docs(workflow): R1 design-lock — BPMN HTTP-task SSRF egress boundary (no runtime)

zensgit:
## R1 first cut — design-lock only, authorizes no runtime

Per the agreed discipline: this PR is **the design-lock + TODO for the BPMN/SSRF boundary**. It adds **no runtime**, no egress code, no config scaffold, no stub, no "temporary allow," and does **not** modify `executeHttpTask`. Implementation slices are named but not started; they begin only after you ratify the open decisions.

## The hole it locks around (grounded)

`BPMNWorkflowEngine.executeHttpTask` (`packages/core-backend/src/workflow/BPMNWorkflowEngine.ts:539`) today does `fetch(resolveExpression(props.url, variables))` with **no scheme/host/IP validation**, and stores the response into a workflow variable → **full-read SSRF** (attacker reads e.g. the cloud-metadata credential document, not just a blind request). And it's not only `url`: `headers` and `body` are field-level `resolveExpression`-templated inputs; `method` is BPMN-definition-controlled (`props.method || 'GET'`), **not** workflow-variable-templated, and is constrained by L8.

## What's locked vs what's yours to decide

- **Locked invariants (L1–L8):** scheme allowlist; post-resolution IP deny-list (metadata `169.254.169.254` / loopback / RFC-1918 / IPv6 ULA / IPv4-mapped); required positive egress allowlist; redirect re-validation; **rebinding/TOCTOU defeat via IP pinning**; timeout + response-size limits; fail-closed default; header/method hardening.
- **Decisions to ratify (D0–D5):** deploy-time provenance; `http` vs `https`-only; allowlist granularity; **rollout for existing http-tasks** (fail-closed is a behavior change — a governance call, not a silent default); header/method policy shape; in-process IP-pinned dispatcher vs central egress proxy.

## Two things the doc deliberately flags

- The current call site **can't be fixed by a pre-check** — `validate(url); fetch(url)` re-resolves DNS and reopens the rebinding window. The fetch-replacement slice needs an **IP-pinned dispatcher or a vetted egress proxy**, so it's scoped realistically.
- `offsetDays` `RangeError` and the `HTTPAdapter` / connector-action / expression-injection surfaces are **explicitly out of scope**.

Please ratify D0–D5 (or amend the locked set); I'll carve the implementation slices only after that, each with its own opt-in. 🤖 Review PR — metasheet2, not for self-merge.

### DIFF
--- a/docs/development/workflow-bpmn-http-task-ssrf-boundary-design-lock-20260630.md
+++ b/docs/development/workflow-bpmn-http-task-ssrf-boundary-design-lock-20260630.md
@@ -28,7 +28,8 @@ inferred target.
 
 `executeHttpTask` (verbatim behavior today):
 
-- resolves `url`, `method`, `headers`, `body` via `resolveExpression(props.*, instance.variables)`;
+- resolves `url`, `headers`, `body` via `resolveExpression(props.*, instance.variables)`, and reads
+  `method` directly as `props.method || 'GET'` (BPMN-definition-controlled, **not** variable-templated);
 - calls `fetch(url, { method, headers, body })` with **no scheme/host/IP validation**;
 - stores the response JSON into `props.responseVariable` when set.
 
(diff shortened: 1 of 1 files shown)


zensgit:
**[P2] fixed.** You're right — `method` is read directly as `props.method || 'GET'` (`BPMNWorkflowEngine.ts:545`); only `url`/`headers`/`body` go through `resolveExpression`. Corrected §2 and §3 to: *url, headers, body are field-level resolveExpression inputs; `method` is BPMN-definition-controlled, not workflow-variable-templated today.* **L8 method hardening retained** — a definition author can still configure a side-effecting method (just not via variable templating). Also added a matching clarifier on the §3 `method` row.

Also **rebased onto current `main`** (`ce7bbc856`) to clear the BEHIND state you flagged.

Still doc-only, no runtime/config/stub. Held for your ratification of D0–D5 (or amendment of the locked set); I'll carve slice 1 only after that.

zensgit:
PR body synced to match the final doc: *headers/body are field-level `resolveExpression` templated; `method` is BPMN-definition-controlled (`props.method || 'GET'`), not variable-templated, and constrained by L8.* No doc/code change — body text only, so a squash/closeout won't carry the stale claim. Still doc-only; holding merge for your ratification + the 18.x/20.x test gates.

PR #3430, opened 2026-07-01
feat(integration): external-API read self-service S1 — config model + validator (#1709)

zensgit:
## What — 刀2 / S1 of the self-service track (#3424 ladder)

The first **code** slice: a **validation-only** config model + save-time validator (`lib/read-source-config.cjs` + test, wired into the plugin `test` chain). **For your review** — CI-green then your go; **not** auto-merged.

## What it does

- **Config model** for a consultant-authored read source, generalizing the proven preset shape: `systemId` (reference to the registered external system that holds baseUrl + credential), `requiredKind`, `object`, `mode` ∈ the four proven modes, `readPath`, `readMethod`, `operations:['read']`, per-mode container/key fields, `fieldMap`, `version`.
- **`validateReadSourceConfig(config)`** → `{valid, normalized}` (frozen, read-only) or `{valid:false, errors:[{code,field,reason}]}`, fail-closed on every guardrail.

## Guardrails (config-time two-tier safety)

- **SSRF endpoint guard** (crown jewel) — strict safe-relative-path allowlist: rejects scheme, protocol-relative `//`, backslash, encoded slash/backslash (`%2f`/`%5c`), path traversal `..`, control chars, non-path chars; relative-only (host-allowlist deferred). **Adversarially test-locked** against the full nasty-set.
- **Read-only** — any `savePath`/`submitPath`/`auditPath` or non-`['read']` operations → `WRITE_CONFIG_REJECTED`.
- **Backend-reference-only credentials** — config references the system via `systemId`; inline credential keys **and** secret-shaped values are rejected (no raw token can be stored).
- Mode/method/encoding enum allowlists; container-path case-aware allowlist; per-mode required fields (hardcoded for the four — not a generic schema engine); strict top-level key allowlist; positive-int `version`.
- **Values-free errors** — `{code, field-name, coarse-reason}` only; never the offending URL or secret value (leak-bait test-locked).

## Scope / non-goals

**Validation ONLY.** No persistence/config-store, no audit-log, no route, no runtime executor, no network, no UI, no write. **Marker-gating enforcement is DEFERRED to S3-runtime** (design-lock requires new presets to be marker-gated — flagged, not silently dropped). Advisor-reviewed (endpoint guard, values-free errors, per-mode-not-generic-schema). Test wired into the CI `test` chain.


zensgit:
Addressed the P1 blocker + P2 in delta commit `94723e6ca` (rebased onto current main):

**P1 (SSRF, blocking):** `isSafeRelativeReadPath` now rejects **all** percent-encoding, not just `%2f`/`%5c`. The gap was `%2e` — Node's URL layer normalizes `/%2e%2e/admin` → `/admin` *after* the guard runs. Demonstrated + closed:
```
/%2e%2e/admin      -> URL normalizes to /admin  | guard: REJECT
/a/%2e%2e/b        -> URL normalizes to /b      | guard: REJECT
/%2E%2E/%2E%2E/etc -> URL normalizes to /etc    | guard: REJECT
/%2e/admin         -> URL normalizes to /admin  | guard: REJECT
```
Rationale: a config endpoint never needs percent-encoding, so one blanket reject kills the whole encoded-char bypass class (`%2e`/`%2f`/`%5c`/`%00`). Clean paths (`/K3API/BOM/GetDetail`) still accepted.

**P2 (fieldMap):** entries validated as config metadata — `source` → field/container path (`isValidContainerPath`), `target` → bounded id (`isBoundedIdentifier`). Value-shaped source (`MAT-001`), traversal (`../../x`), and whitespace targets now reject; a well-shaped map still passes; errors stay values-free.

Tests added: the four `%2e` cases + `%20`/`%00`, and valid + malformed/leak-bait fieldMap cases. Scope unchanged (validation-only). Not merging pending your re-review.

PR #3434, opened 2026-07-01
feat(workflow): egress-guard module for BPMN SSRF boundary (R1 slice 1, not wired)

zensgit:
## R1 slice 1 — the egress-guard module (per your `拍 D0–D5` + slice-1 go)

Implements the first slice of the design-locked BPMN HTTP-task SSRF boundary (**#3428**): a **standalone** validator + reusable IP classifier + fail-closed policy config. It **resolves no DNS, opens no sockets, and is NOT wired to `executeHttpTask`** — exactly the scope you specified.

### Exports
- **`isBlockedIp(ip, nat64Prefixes?)`** — the L2 core (reused by slice 2 on resolved IPs). Blocks loopback / link-local (incl. cloud metadata `169.254.169.254`) / private / CGNAT / IPv6 ULA / multicast / reserved / unspecified, and **decomposes embedded-IPv4 tunnel forms** (IPv4-mapped, 6to4, NAT64) so a v6 wrapper can't smuggle a blocked v4. Unparseable → fail-closed.
- **`validateEgressUrl(url, policy)`** — synchronous L1 (scheme + creds) + L3 (allowlist) + L2 for IP-literal hosts. Canonicalizes via WHATWG `URL` so decimal/octal/hex/short IPv4 forms normalize before classification; **L2 overrides L3** (an allowlisted internal IP is still blocked).
- **`defaultEgressPolicy()`** — `https`-only + empty allowlist ⇒ **deny all** (fail-closed).

### New runtime dependency (security path)
Adds **`ipaddr.js@^2.4.0`** for vetted IP/CIDR parsing — chosen over hand-rolling IPv6/mapped/CIDR classification (`package.json` + lockfile; lockfile delta is only this package).

### Adversarial verification (and a real fix)
Two independent adversarial probes wrote + ran scratch tests against the actual functions:
- URL-parsing probe → **no bypass** (the guard reads the same WHATWG `url.hostname` a downstream fetch would).
- IP-representation probe → surfaced **one real gap**: NAT64 on a **Network-Specific Prefix** (only the well-known `64:ff9b::/96` was decoded). **Fixed** by making `EgressPolicy.nat64Prefixes` configurable (well-known always decoded; operators running NAT64 declare their prefix), with the residual documented (unconfigured NSP / non-`/96` lengths; L3 allowlist is the backstop). New tests pin it.

### D0–D5 ratification (companion doc)
D1/D2 encoded as `EgressPolicy` config; **D3 (rollout)/D4 (header-method)/D5 (dispatcher) explicitly bind at their wiring slices and remain owner-amendable** — slice 1 does not settle them.

### Verification
`vitest` — **59 pass**; `tsc --noEmit` — **clean**.

Slices 2–4 (IP-pinned dispatcher, `executeHttpTask` wiring + L8, rollout) are **not started** and each takes its own opt-in. 🤖 Review PR — metasheet2, not for self-merge.

### DIFF
--- a/docs/development/workflow-bpmn-http-task-ssrf-D-ratification-and-slice1-20260630.md
+++ b/docs/development/workflow-bpmn-http-task-ssrf-D-ratification-and-slice1-20260630.md
@@ -1,0 +1,74 @@
+# R1 SSRF — D0–D5 ratification + slice 1 (egress-guard module), 2026-06-30
+
+Companion to the design-lock
+`docs/development/workflow-bpmn-http-task-ssrf-boundary-design-lock-20260630.md` (`#3428`).
+This record (1) ratifies the D0–D5 decisions and (2) records slice 1 — the standalone
+egress-guard module. **No runtime is wired to `executeHttpTask`;** the module resolves no DNS and
+opens no sockets.
+
+## 1. D0–D5 ratification
+
+Ratified per the design-lock recommendations. The two decisions that shape slice 1 (D1, D2) are
+encoded as **`EgressPolicy` config**, not hard-coded; the rest bind at their *wiring* slices and
+remain owner-amendable — so slice 1 does **not** silently settle the genuinely owner-scoped calls
+(D3 rollout especially).
+
+| # | Decision | Ratified | Binds at |
+|---|---|---|---|
+| D0 | Deploy provenance / who sets http-task fields | Workflow authors are untrusted for network egress; the boundary holds regardless of deploy authz. | (assumption; no code input) |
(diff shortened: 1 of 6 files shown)


zensgit:
Superseded by the parallel R1-A egress guard landed in `3915c1ca4` (same design-lock, same file paths — a genuine duplicate built concurrently). The landed version is the better-structured incumbent (internal-DNS-name blocking, zone-id rejection, typed reason codes). Closing this duplicate rather than clobbering it.

The one gain this branch carried that the incumbent lacks — the **configurable `nat64Prefixes` fix** my adversarial probe surfaced (the incumbent decodes only the well-known `64:ff9b::/96`, so a NAT64 Network-Specific-Prefix can still smuggle an internal v4) — will land as a small follow-up ported onto the incumbent's structure.

PR #3437, opened 2026-07-01
fix(workflow): block embedded IPv4 egress literals

zensgit:
## Summary

Follow-up to #3432 / R1-A review P3 before R1-B consumes `isBlockedEgressIp` for resolved DNS answers.

This PR:
- decodes deprecated IPv4-compatible IPv6 (`::a.b.c.d` / `::7f00:1`) through the existing IPv4 deny classifier
- decodes ISATAP-style embedded IPv4 (`...:0000:5efe:a.b.c.d` and `...:0200:5efe:a.b.c.d`) through the same IPv4 deny classifier
- adds URL-level tests for both forms, not only helper-level tests
- preserves the mainline configurable NAT64 `/96` prefix handling while routing it through the same embedded-IPv4 classifier
- records the R1-B caller contract: fetch/use `decision.normalizedUrl`, not the raw input string

No BPMN engine wiring, no dispatcher, no fetch behavior change.

## Verification

- `pnpm --filter @metasheet/core-backend exec vitest run src/guards/__tests__/egress-guard.test.ts --reporter=dot` — 49/49
- `pnpm --filter @metasheet/core-backend exec vitest run src/guards --reporter=dot` — 65/65
- `pnpm --filter @metasheet/core-backend run build` — pass
- `pnpm --filter @metasheet/core-backend exec eslint src/guards/egress-guard.ts src/guards/__tests__/egress-guard.test.ts` — 0 errors; test file ignored by existing ESLint ignore pattern

### DIFF
--- a/docs/development/workflow-bpmn-egress-guard-slice1-dev-verification-20260701.md
+++ b/docs/development/workflow-bpmn-egress-guard-slice1-dev-verification-20260701.md
@@ -22,6 +22,9 @@ workflow behavior.
 - The existing `send_webhook` SSRF guard was reviewed and left untouched. It is a product-specific
   resolve-and-pin path; this slice adds the BPMN policy primitive under `src/guards` and keeps
   dispatcher reuse/wiring for later slices.
+- Callers must use the returned `decision.normalizedUrl` when a URL is allowed. It is the WHATWG
+  canonicalized URL that this guard evaluated; a later caller that fetches the raw input string could
+  reintroduce parser-split ambiguity outside this module.
 
 ## Explicitly not shipped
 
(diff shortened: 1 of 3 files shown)


zensgit:
**Adversarially verified the ISATAP / IPv4-compatible extraction** (reused the SSRF probe harness against this PR's `isBlockedEgressIp`, 13 cases, all pass):

- **ISATAP behind different outer prefixes** with an internal inner v4 — doc (`2001:db8::5efe:…`), **global** (`2606:4700:4700::5efe:10.0.0.1`), and link-local (`fe80::5efe:127.0.0.1`), with inner metadata `169.254.169.254` / RFC-1918 / CGNAT `100.64` / loopback, u/g bit both 0 and 1 → **all blocked**. `isIsatapAddress` checks only the `00:00:5e:fe` marker at bytes 8–11, so it's **prefix-agnostic** — it does **not** have the NSP-NAT64-class gap (a global-prefix ISATAP wrapping an internal v4 is caught).
- **IPv4-compatible `::a.b.c.d`** with internal inner → all blocked.
- **Public inner** (`::8.8.8.8`, `2001:db8::5efe:8.8.8.8`) → correctly **not** blocked (allowlist gates it downstream).

No leak found. Minor benign note: `::/96` also matches `::1` / `::` → treated as IPv4-compatible `0.0.0.x` and still blocked (unspecified range), so safe. LGTM on the extraction — the DIRTY status just needs a rebase onto post-`#3438` main.

PR #3442, opened 2026-07-01
docs(workflow): refresh R1 BPMN SSRF egress closure ledger

zensgit:
## What this is

Docs-only R1 BPMN HTTP-task SSRF egress line ledger, refreshed after the line moved through A2.

Grounded against `origin/main` at `5193432a8` (`#3451`). This replaces the stale point-in-time survey that still said `#3437` was open and slices 2/3 were unbuilt.

## Current facts recorded

- `#3428` design-lock is merged and ratified.
- `#3432` R1-A guard is merged.
- `#3438` NAT64 NSP hardening is merged.
- `#3437` IPv4-compatible / ISATAP hardening is merged.
- `#3443` helper API hardening is merged.
- `#3447` A1 IP-pinned dispatcher is merged.
- `#3451` A2 engine wiring is merged: `executeHttpTask` no longer uses raw `fetch`; it routes through the pinned dispatcher + HTTPS JSON transport and is default fail-closed.

## Boundary

This PR opens no runtime and changes no behavior. It records that the original live full-read SSRF path is closed by default, while keeping A3 rollout / configured policy enablement as a separate owner/governance gate.

## Verification

- `git diff --check`
- Source facts checked against merged PR metadata and `origin/main`.


zensgit:
Synced the PR body to the final doc口径: #3437's ISATAP/IPv4-compatible extraction is **verified** (13-case adversarial probe, no leak — review comment on #3437), and the only remaining step for it is **rebase + land**. Removed the stale "not verified here" line and the "adversarial review as next step" (it's done), so a squash/PR page won't carry the old claim.

PR #3445, opened 2026-07-01
[codex] feat(integration): add read-source probe contract

zensgit:
## What changed

Adds the S2-a pure-function contract for external-API read self-service:

- `read-source-probe-contract.cjs` normalizes a future probe plan from an S1-normalized read-source config only.
- Adds values-free probe evidence sanitization using the existing read-smoke container evidence sanitizer.
- Exports `readSmokeResponseShapeContainerEvidence` so S2-a reuses the existing evidence spine instead of inventing a parallel shape.
- Wires `read-source-probe-contract.test.cjs` into the plugin CJS test chain.

## Boundary

This is S2-a only: pure functions, no route, no outbound call, no external-system lookup, no credential resolution, no persistence, no audit write, no UI, no generic runtime, no resolver composition, and no write/delete path.

The probe contract rejects raw top-level path/method/body/credential fields, requires structurally normalized S1 config, uses platform-fixed timeout/cap constants, and only emits alias-scoped `{type,arrayLength}`/counts/booleans/coarse error evidence.

## Verification

- `node __tests__/read-source-probe-contract.test.cjs`
- `node __tests__/read-source-config.test.cjs && node __tests__/read-smoke.test.cjs`
- `git diff --check`
- `pnpm -C plugins/plugin-integration-core test`

Note: the first full plugin test attempt failed because the fresh worktree had no `node_modules`; after `pnpm install` with an unchanged lockfile, the full plugin CJS suite passed.


zensgit:
**Independent review — strong approve on scope & design, one P2 to resolve.** Verified against S2 design-lock #3439 (S2-a = probe contract + evidence schema, pure functions, no outbound/persistence/write/route).

**What it gets right:** pure functions (only S1 validator + read-smoke, no http/db/credential/adapter, no runtime consumer); consumes only S1 normalized config *and enforces it* (`config_not_normalized` round-trip rejects a valid-but-raw config); `operations:['read']` frozen; platform-fixed timeout/cap with no input channel; evidence is an **allowlist-pull** (never spreads the result), reusing `readSmokeResponseShapeContainerEvidence` (the mandated reuse, not a parallel path); plan/evidence separation correct (`readPath`/`systemId` never in evidence); adversarial values-free leak-baits; test wired into the CI chain; green locally.

**P2 (recommend fixing in this PR) — `errorCode`/`errorType` is shape-validated, not values-free by construction.** It's the sole evidence channel that passes a producer string through instead of extracting/gating. Empirically, value-shaped identifiers ride through:
```
errorCode "MAT-001" -> "MAT-001"            errorType "evil.example.com" -> "evil.example.com"
errorCode "PO-2024-000123" -> unchanged     errorType "Data.Page2.FItemId" -> unchanged
```
Realistic, not contrived (an adapter error embedding the failing doc number, or `errorCode: \`READ_FAILED_${matNo}\``). Fix splits: **errorType** — drop `.`/`:` from the charset (error-class names don't need them → kills host/field-path shapes cheaply); **errorCode** — `MAT-001` vs `READ_FAILED` are shape-identical, so either validate against a registered coarse-code set (the contract-like fix, on-scope for a contract slice) or a **tested** best-effort acceptance. **Not** "document that S2-b must be careful" — that betrays the module's own by-construction principle (containers/counts don't trust the producer). Pin with a red test (`errorCode:'MAT-001'`) either way.

Honest framing: **latent, no live producer today** — S2-a generates none of these values, it sanitizes a future S2-b result. So not unsafe to land; but the PR's deliverable *is* the values-free schema, so a values-free hole in it is worth closing here. Merge-timing is the owner's call.

P3: document the `config_not_normalized` caller obligation (callers pass `validateReadSourceConfig(x).normalized`, not raw); errorType is somewhat redundant with errorCode.

Full review: /tmp/pr3445-review-claude-20260701.md

zensgit:
**Re-review @ 0497787cc — P2 + P3 resolved. Approve.**

**P2 fixed in the robust form.** `safeErrorCode`/`safeErrorType` now validate against **exact registered Sets** (`Set.has`, 11 coarse codes / 8 class names) — not a shape regex, and not "document that S2-b must be careful." The sets are exported so S2-b consumes the canonical taxonomy. Empirically verified against the real module:
```
READ_SOURCE_PROBE_TIMEOUT / K3WiseWebApiAdapterError -> pass (registered)
READ_SOURCE_PROBE_MAT-001                            -> READ_SOURCE_PROBE_FAILED   ← prefix-attempt DROPPED (exact-set, not startsWith)
MAT-001 / PO-2024-000123                             -> READ_SOURCE_PROBE_FAILED
evil.example.com / Data.Page2.FItemId               -> Error
```
The prefix-attempt dropping is the decisive proof the earlier passthrough class is fully closed — errorCode/errorType are now values-free **by construction** like the container/count/boolean channels.

**Tests strengthened:** valid case uses a registered code (`READ_SOURCE_PROBE_REJECTED`); unsafe case is a 4-way loop (`MAT-001`, `PO-2024-000123`, `READ_FAILED_MAT_001`, spaces) asserting downgrade **and** no-leak. Green locally.

**P3 resolved:** `assertS1NormalizedConfig` now documents the caller obligation (pass `validateReadSourceConfig(raw).normalized`, not raw).

Clear to merge once CI is green on this head. Optional nit (non-blocking): add one explicit `READ_SOURCE_PROBE_MAT-001` prefix-attempt case to the test loop to lock exact-set-vs-prefix in the suite — the impl is provably exact-set via `Set.has`, so it's belt-and-suspenders.

PR #3446, opened 2026-07-01
fix(approval): T2-4 threshold re-entry quorum-bypass — scope tally to current node-entry round

zensgit:
Closes the **T2-4 threshold re-entry quorum-bypass** (confirmed real bug; Lane B / approval engine, independent of the in-flight R1 lane).

## Bug
The threshold quorum tally counts `COUNT(DISTINCT actor_id) … action='approve' AND metadata->>'nodeKey'=$2` with **no round scoping**. The 退回/return path admits any previously-visited approval node — including an already-resolved `threshold` node — and `approval_records` are append-only. So on re-entry the prior-round approve records still count toward N: **a 2-of-3 that A+B already approved re-resolves on a single fresh vote.** Self-verified reachable.

## Fix (schema-free — uses existing data)
The current round begins at the `to_version` of the most recent return/jump that re-entered the node; approve records carry `to_version` (bumped), so scope the tally to **`to_version >= cutoff`**. `>=` (not `>`) is deliberate: a return/jump can fire an auto-approval cascade AT the re-entered node in the **same transaction** (`insertAutoApprovalEvents` writes those approve records at `to_version = re-entry version = cutoff`) — legitimate current-round votes that must count; prior-round approves are strictly `< cutoff`, so `>=` admits the cascade without re-admitting stale votes. **No re-entry → cutoff null → falls back to the exact old whole-node query (zero regression for the normal path).** The graph is validated **acyclic**, so a threshold node is only re-enterable via return/jump (marker complete — no forward-loop bypass). Only the `threshold` branch changes.

## Verification
tsc 0 · real-DB `approval-nofm-threshold` **4/4** incl. a **new fail-first re-entry test** (2-of-3 → A+B approve → advance → RETURN → one fresh vote must NOT re-resolve; a 2nd fresh vote still resolves). **RED-before confirmed.** N>M fail-closed + single-reject paths unchanged.

Open follow-up (flagged): a dedicated regression for the same-version-cascade `>=` boundary — the `>=` correctness is reasoned + the cascade write-site confirmed, but not yet locked by its own test.

### DIFF
--- a/packages/core-backend/src/services/ApprovalProductService.ts
+++ b/packages/core-backend/src/services/ApprovalProductService.ts
@@ -4175,14 +4175,52 @@ export class ApprovalProductService {
         // across role rows / re-dispatch, and auto-approvals are included because they too
         // write an action='approve' record carrying metadata.nodeKey (Q5).
         const threshold = executor.getApprovalThreshold(currentNodeKey)
-        const priorApproversResult = await client.query<{ count: string }>(
-          `SELECT COUNT(DISTINCT actor_id) AS count
+        // T2-4 RE-ENTRY quorum fix: scope the DISTINCT-approver tally to the CURRENT node-entry round. A
+        // `threshold` node can be RETURNED/JUMPED back to after it already resolved (the 退回/return path admits
+        // any previously-visited approval node), and `approval_records` are append-only — so without scoping,
+        // approve records from a PRIOR entry still count toward N and re-satisfy the node on stale votes (a
+        // quorum bypass: a 2-of-3 that A+B already approved would resolve on a single fresh vote). The current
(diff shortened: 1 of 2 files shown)


zensgit:
Follow-up from adversarial review (self-verified): the cutoff uses `to_version >= cutoff` (not `>`) because a return/jump can fire an auto-approval cascade AT the re-entered node in the same transaction — `insertAutoApprovalEvents` writes those approve records at `to_version = re-entry version = cutoff`, so they're current-round votes that must count; prior-round approves are strictly `< cutoff`. Also confirmed the graph is validated **acyclic**, so a threshold node is only re-enterable via return/jump (the marker is complete — no forward-loop bypass). **Open follow-up:** a dedicated test for that same-version-cascade boundary is reasoned + the write-site confirmed but not yet test-locked.

PR #3450, opened 2026-07-01
feat(automation): add T2-6 event dedup ledger

zensgit:
## Summary

Implements T2-6 on the ratified #3385 register defaults: event-driven automation triggers now carry a transport `_eventId`, and each matching rule claims `(rule_id, dedup_key)` before executing so redelivered events are at-most-once while legitimate later events still fire.

## Scope

- Adds `meta_automation_event_fires` with `PRIMARY KEY (rule_id, dedup_key)` and a `fired_at` retention index.
- Adds `automation-event-dedup.ts` helpers:
  - `_eventId` stamping for supported emit sites.
  - `dedupKey = ${eventType}:${_eventId}`.
  - absent/invalid `_eventId` fail-open during rollout.
  - 7-day retention cutoff.
- Wires `_eventId` onto `multitable.record.created`, `multitable.record.updated`, `multitable.record.deleted`, and `multitable.form.submitted` emit sites, including chained automation writes and W7 backwrite fan-out.
- `field.value_changed` rides `multitable.record.updated`, so it dedups on the underlying event id.
- Leaves `webhook.received` and `approval.*` out of scope, per the register.
- Wires the real-DB test file into the multitable integration lane and excludes it from the no-DB default run so it cannot skip-green.

## Verification

- `pnpm --filter @metasheet/core-backend exec vitest --config vitest.config.ts run tests/unit/automation-event-dedup.test.ts tests/unit/multitable-automation-service.test.ts tests/unit/record-service.test.ts tests/unit/record-write-service.test.ts --reporter=dot` — 121/121
- `DATABASE_URL=postgresql://postgres@localhost:5432/metasheet_test pnpm --filter @metasheet/core-backend exec vitest --config vitest.integration.config.ts run tests/integration/multitable-event-dedup-trigger.test.ts tests/integration/multitable-form-submit-trigger.test.ts tests/integration/multitable-date-reminder-trigger.test.ts --reporter=dot` — 29/29
- `pnpm --filter @metasheet/core-backend exec eslint src/multitable/automation-event-dedup.ts --max-warnings=0`
- `git diff --check HEAD~1..HEAD`

## Notes

A full-file targeted ESLint run over the long touched legacy files still reports pre-existing lint debt unrelated to this PR (`record-write-service.ts`, `univer-meta.ts`). The new helper itself is lint-clean, and the behavior is covered by unit + real-DB integration tests.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -297,6 +297,7 @@ jobs:
             tests/integration/multitable-automation-start-approval.test.ts \
             tests/integration/multitable-automation-start-approval-http.test.ts \
             tests/integration/multitable-form-submit-trigger.test.ts \
+            tests/integration/multitable-event-dedup-trigger.test.ts \
             tests/integration/multitable-date-reminder-trigger.test.ts \
             tests/integration/multitable-webhook-event-bridge.test.ts \
             tests/integration/multitable-webhook-retry-tick.test.ts \
--- a/packages/core-backend/src/db/migrations/zzzz20260701120000_create_automation_event_fires.ts
+++ b/packages/core-backend/src/db/migrations/zzzz20260701120000_create_automation_event_fires.ts
@@ -1,0 +1,29 @@
+/**
+ * Migration: `meta_automation_event_fires` — event-driven automation idempotency ledger.
+ *
+ * Event-driven triggers (record.created/updated/deleted, form.submitted, and field.value_changed via
+ * record.updated) receive a transport `_eventId` at every emit site. Each matching rule claims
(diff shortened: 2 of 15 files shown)


zensgit:
Addressed the two review notes in 070ebdb37:\n\n- buildAutomationEventDedupKey now accepts unknown payloads and fails open for null/undefined/array/non-stamped payloads instead of throwing.\n- The opportunistic event-dedup ledger retention sweep is kicked off best-effort in the background, so stamped event handling is not blocked by the daily DELETE. Added a unit test with a never-resolving sweep to lock that hot-path behavior.\n\nRe-verified locally after rebase onto current main: backend build, 122/122 focused unit specs, 29/29 real-DB specs, and diff check.

PR #3455, opened 2026-07-01
feat(workflow): add R1 A3 egress policy normalizer

zensgit:
## What

Implements R1 A3-a from #3452: a pure BPMN HTTP-task egress policy config normalizer.

- Adds `normalizeBpmnHttpTaskEgressPolicyConfig(rawConfig)` for server-owned policy input.
- Returns either a frozen normalized `EgressPolicy` plus values-free metadata, or a fail-closed disabled policy plus coarse error code/field.
- Locks exact ASCII DNS host normalization, explicit punycode acceptance, Unicode IDN rejection, public IP literal rejection, wildcard/suffix/path/scheme/port/userinfo/CIDR rejection, duplicate host rejection, and `/96` IPv6 NAT64 prefix validation.
- Requires NAT64 policy prefixes to be true network roots (host bits zero), so semantically-duplicate `/96` prefixes cannot drift counts/fingerprints.
- Keeps evidence values-free: counts + fingerprint only, no raw host/config/token echo.

## Boundary

A3-a only. No route, engine, transport, DNS, database, server config loader, workflow deployment path, or outbound call is touched. No destination is enabled; runtime remains fail-closed by default.

## Verification

- `pnpm install --frozen-lockfile`
- `pnpm --filter @metasheet/core-backend exec vitest run src/guards/__tests__/egress-policy-normalizer.test.ts src/guards/__tests__/egress-guard.test.ts src/guards/__tests__/egress-dispatcher.test.ts --reporter=dot` — 100/100
- `git diff --check`

### DIFF
--- a/packages/core-backend/src/guards/__tests__/egress-policy-normalizer.test.ts
+++ b/packages/core-backend/src/guards/__tests__/egress-policy-normalizer.test.ts
@@ -113,6 +113,8 @@ describe('R1-A3-a BPMN HTTP-task egress policy normalizer', () => {
 
   test.each([
     '2a00:1098:2c::/64',
+    '2a00:1098:2c::1/96',
+    '64:ff9b::c000:0201/96',
     '8.8.8.0/24',
     'not-a-prefix',
     '2a00:1098:2c::',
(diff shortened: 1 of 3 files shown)


zensgit:
Follow-up fix pushed at a283067d2: reject an explicit 64:ff9b::/96 entry in nat64Prefixes as a duplicate of the guard's built-in well-known NAT64 prefix. This keeps nat64PrefixCount/policyFingerprint scoped to extra deployment-specific prefixes only. Local verification: focused egress-policy-normalizer vitest 36/36 + core-backend tsc --noEmit.

PR #3492, opened 2026-07-02
fix(multitable): tolerate missing member group table for record permissions

zensgit:
## Summary

- add the same optional `platform_member_groups` fallback to the record-permissions list endpoint that view/field permission lists already have
- keep member-group labels best-effort: if the optional directory table is absent, the endpoint still returns permission rows with null group label/subtitle instead of 500
- add a focused route test for the missing-table fallback path

## Why

#3408/#3426 continue to report permission-list 500s on the entity environment. Current main already has this fallback for view and field permission lists, but record permissions still joined `platform_member_groups` without the same retry path. This closes that asymmetric backend gap.

This PR does not claim to fix the formula echo/materialization failures in #3408/#3426; current main still needs separate deployment/evidence alignment for that path.

## Validation

- `pnpm install --frozen-lockfile`
- `pnpm --filter @metasheet/core-backend exec vitest --config vitest.integration.config.ts run tests/integration/multitable-sheet-permissions.api.test.ts -t "lists record permissions when the optional member-group directory table is absent" --reporter=dot`
- `pnpm --filter @metasheet/core-backend exec tsc --noEmit --pretty false`
- `git diff --check`

Note: the full `multitable-sheet-permissions.api.test.ts` file currently has unrelated mock-harness failures on existing tests after recent SQL additions; the new focused fallback case passes.

### DIFF
--- a/packages/core-backend/src/routes/univer-meta.ts
+++ b/packages/core-backend/src/routes/univer-meta.ts
@@ -7550,37 +7550,70 @@ export function univerMetaRouter(): Router {
         return res.status(404).json({ ok: false, error: { code: 'NOT_FOUND', message: `Record not found: ${recordId}` } })
       }
 
-      const result = await pool.query(
-        `SELECT
-            rp.id,
-            rp.sheet_id,
-            rp.record_id,
-            rp.subject_type,
-            rp.subject_id,
-            rp.access_level,
-            rp.created_at,
-            rp.created_by,
-            u.name AS user_name,
-            u.email AS user_email,
-            u.is_active AS user_is_active,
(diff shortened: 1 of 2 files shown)


zensgit:
Thanks. I’m keeping this PR in the existing route pattern deliberately: `view` and `field` permission lists already use the same primary query + missing-`platform_member_groups` fallback query shape, and this change makes `record` permissions symmetric with them.

A shared query builder would be reasonable as a separate cleanup across all three endpoints, but doing it inside this hot fix would broaden the diff and risk changing two already-covered routes. This PR stays scoped to the missing record-permissions fallback plus its regression test.

PR #3489, opened 2026-07-02
feat(automation): add signed inbound webhook trigger

zensgit:
## Summary
- add signed `POST /api/multitable/automation/webhooks/:ruleId` delivery for `webhook.received` rules
- require non-empty per-rule secrets at create/update, reject legacy secret-less rules on edit/ingest, and redact secrets on rule reads
- verify HMAC over `timestamp.rawBody`, enforce 5-minute replay window, 1MB JSON limit, 60/min/rule rate limit, uniform 401 rejects, and rejection metrics
- wire the real-DB integration file into CI and document the T1-2 dev/verification record

## Security boundary
- caller-supplied `recordId` / `sheetId` / actor-shaped fields remain body data only; the execution context is synthetic record-less with `actorId=null`
- accepted webhooks execute under the stored automation rule author, same as scheduled/event triggers
- malformed / stale / unsigned / wrong-trigger / disabled / missing-secret paths all fail closed

## Not included
- editor UI exposure for configuring webhook secrets
- secret rotation history / dual-secret windows
- source IP allowlists

## Verification
- `pnpm --filter @metasheet/core-backend exec tsc --noEmit`
- `pnpm --filter @metasheet/core-backend exec vitest --config vitest.config.ts run tests/unit/automation-inbound-webhook.test.ts tests/unit/automation-routes-wiring.test.ts --reporter=dot`
- `pnpm --filter @metasheet/core-backend exec vitest --config vitest.config.ts run tests/unit/automation-v1.test.ts tests/unit/multitable-automation-service.test.ts tests/unit/automation-inbound-webhook.test.ts tests/unit/automation-routes-wiring.test.ts --reporter=dot`
- `DATABASE_URL=${DATABASE_URL:-postgresql://postgres@localhost:5432/metasheet_test} pnpm --filter @metasheet/core-backend exec vitest --config vitest.integration.config.ts run tests/integration/multitable-inbound-webhook-trigger.test.ts tests/integration/multitable-event-dedup-trigger.test.ts tests/integration/multitable-form-submit-trigger.test.ts tests/integration/automation-approval-completed-trigger.test.ts --reporter=dot`

PR #3467, opened 2026-07-02
feat(automation): approval.completed trigger — template-routed, record-less v1 (T1-3)

zensgit:
**First-batch ballot GO (T1-3 all-✅).** Approval completions become a first-class automation trigger: approve/reject/revoke/cancel → matching rules fire (notification/webhook/email/DingTalk) — the approval → automation product loop (拒绝通知、审批后 webhook、审批后消息 all land here).

## Ballot decisions → code
- **Q1 routing:** `trigger_config.templateId` REQUIRED + must reference an existing template; cross-sheet dispatch via `loadEnabledApprovalCompletedRules` (JSONB filter, no index v1). **templateId-null completions = v1 out-of-contract** (never match; debug-log + return).
- **Q2 authz:** creator must hold `approvals:read` — at SAVE **and re-checked at fire** (deny+skip, fail-closed on lookup errors). Principal = route-set server-side `createdBy`.
- **Q3 record-less v1:** payload `recordId:''` with the approval event riding along; actions save-restricted (top-level AND nested) to `send_notification / send_webhook / send_email / send_dingtalk_*`; record-targeting + `start_approval` rejected → the approval→approval loop is structurally broken (Q4a).
- **Q4 depth:** bridge-originated approvals continue the parent chain at depth+1 (read-only bridge lookup); human-started = depth 0; MAX_AUTOMATION_DEPTH retained.
- **Q5 idempotency:** reuses the **T2-6 ledger** with `dedup_key = approval.completed:{event.eventId}` — no new ledger, redelivery = no-op.
- **Q6:** single `approval.completed` type + optional `outcomes` filter (default approved-only; invalid persisted values degrade fail-closed).
- **Q7:** fresh rules fire for EVERY completion, independent of the W6/W7 bridge resume.
- **Q8:** non-empty conditions rejected at save.
- **DB:** `chk_automation_trigger_type` CHECK widened by migration (date_field-widening pattern). `updateRule` validates the full RESULTING shape (no smuggling via action-only edits).

## Verification
[...]

PR #3495, opened 2026-07-02
feat(automation): expose approval.completed + webhook.received triggers in the rule editor (UI slice)

zensgit:
Owner-ordered UI-exposure cut: **the fastest way to make the shipped capabilities visible.** T1-3 `approval.completed` (#3467) and T1-2 signed inbound webhook (#3489) were API-only; the editor's webhook comment still claimed runtime-inert (stale since #3489).

## What ships
- **Editor options** for both triggers + FE type union + zh/en labels + Manager card summaries; stale comment replaced with the current contract.
- **approval.completed config UI**: template select (reuses the existing `approvalTemplates` loader, text fallback) + outcomes checkboxes (empty → backend default approved-only) + **FE mirror of the backend save gate** with a visible block reason (templateId required / no conditions / notification-family actions only).
- **webhook.received config UI**: write-only secret input + endpoint/signature hint (`POST /api/multitable/automation/webhooks/:ruleId`, timestamp-bound HMAC, ±300s). **Keep-or-rotate semantics**: the read API redacts the secret as `<redacted>` — the editor strips the placeholder at hydrate and, when an existing rule's secret field is left blank, **omits `triggerConfig` from the payload entirely** so the backend preserves the real secret; typing a value rotates it.
- **Backend defense-in-depth (P1-grade find while building this)**: the save gate accepted the literal `<redacted>` as a valid secret — any client naively round-tripping a fetched rule would silently corrupt the stored secret and break all future signatures. `validateInboundWebhookTriggerAtSave` now rejects the placeholder with a keep-or-rotate message.

## Verification
`vue-tsc -b` 0 · backend tsc 0 · **rule-editor spec 111/111** (4 new cases: options+labels · approval.completed gate-mirror + outcomes serialize · new-rule secret required+trimmed · existing-rule keep-stored-secret omits triggerConfig, then rotate) — the old "webhook NOT selectable / exactly 8 options" contract test flipped to the new 10-option contract · labels 12/12 · Manager 79/79 · backend inbound-webhook unit 5/5 (new sentinel-reject) · real-DB inbound-webhook trigger 5/5.

## Not in this slice
Approval-field condition vocabulary (named follow-up from #3467) · approval-center batch/urge UX (next slice per owner order).

### DIFF
--- a/apps/web/src/multitable/components/MetaAutomationManager.vue
+++ b/apps/web/src/multitable/components/MetaAutomationManager.vue
@@ -1840,6 +1840,10 @@ function describeTrigger(rule: AutomationRule): string {
       const fid = rule.triggerConfig?.fieldId as string | undefined
       return automationCardTriggerSummary(rule.triggerType, fid ? fieldNameById(fid) : '', isZh.value)
     }
+    // T1-2/T1-3: editor-exposed triggers get their localized type label in the card summary.
+    case 'webhook.received':
+    case 'approval.completed':
+      return automationTriggerTypeLabel(rule.triggerType, isZh.value)
     default:
       return String(rule.triggerType)
   }
--- a/apps/web/src/multitable/components/MetaAutomationRuleEditor.vue
+++ b/apps/web/src/multitable/components/MetaAutomationRuleEditor.vue
@@ -39,12 +39,14 @@
             <option value="schedule.interval">{{ automationTriggerTypeLabel('schedule.interval', isZh) }}</option>
             <option value="schedule.date_field">{{ automationTriggerTypeLabel('schedule.date_field', isZh) }}</option>
             <!--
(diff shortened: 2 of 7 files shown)


zensgit:
Fixed the Gemini TDZ finding in this branch.

Change:
- Moved `INBOUND_WEBHOOK_REDACTED_SECRET` above `validateInboundWebhookTriggerAtSave()` so the save-time validator cannot reference the const before initialization.

Validation run on the rebased branch:
- `pnpm --filter @metasheet/core-backend exec vitest run tests/unit/automation-inbound-webhook.test.ts --reporter=dot` — 5/5 pass
- `pnpm --filter @metasheet/core-backend exec tsc --noEmit --pretty false` — pass
- `git diff --check` — pass

No runtime scope change beyond declaration order.

PR #3501, opened 2026-07-02
docs(approval): nodeEntryEpoch design-lock — durable T2-4 threshold round-scoping

zensgit:
The structural close you called for: after the **3rd** patch to the same threshold round-scoping heuristic (#3446 name-X · #3453 same-version cascade · #3499 through-X) + an open condition/cc 4th-vector verify-item, this promotes the register's **Option-B `nodeEntryEpoch`** to a design-lock.

## Why
Reconstructing "which round am I in" by pattern-matching `nextNodeKey`/`targetNodeKey`/`toNodeKey` across append-only records is inherently whack-a-mole. Record the round at the source instead.

## Mechanism
Mint a per-node-entry **epoch** at each activation (`approval_instances.node_activation_seq`, stamped on the activation's assignment rows as `entry_epoch`, copied onto approve records as `metadata.nodeEntryEpoch`); tally only approves whose epoch equals the node's **current** epoch (read off its active assignments). The epoch keys off **activation, not transition metadata**, so it's complete for every entry vector — direct return/jump, forward re-entry, timeout-jump, **and condition/cc-in-the-middle** — **by construction**.

## Migration (the hard part) — dual-read fallback, zero backfill
Additive columns; a threshold instance pending mid-round at deploy (legacy NULL-epoch rows) keeps the current cutoff behavior until its node next re-activates, then transparently switches to the epoch path. No record mis-counted, no round reconstruction.

## In the doc
Stamp sites · exact tally query · full edge matrix (activation/add-sign/reduce-sign/cascade/return/jump/timeout-jump/condition-cc → same/new epoch) · open owner decisions · fail-first verification plan (adds the condition/cc vector + a mid-round-deploy migration test).

**PROPOSED — no runtime until ratified.** The cutoff fixes (#3446/#3453/#3499) remain the in-place guard meanwhile. On ratification → implement in Lane B (`ApprovalProductService.ts` + a migration), fail-first + real-DB.

### DIFF
--- a/docs/development/node-entry-epoch-threshold-round-scoping-design-lock-20260703.md
+++ b/docs/development/node-entry-epoch-threshold-round-scoping-design-lock-20260703.md
@@ -1,0 +1,151 @@
+# nodeEntryEpoch — durable T2-4 threshold round-scoping · DESIGN-LOCK (PROPOSED) · 2026-07-03
+
+> **Status: PROPOSED design-lock — awaiting owner ratification.** Owner-directed 2026-07-03: after the 3rd
+> cutoff-heuristic patch (#3446 name-X · #3453 same-version cascade · #3499 through-X) plus an open condition/cc
+> 4th-vector verify-item, promote the register's **Option B (`nodeEntryEpoch`)** to the next design-lock — the
+> structural fix that retires this bug family. **No runtime is written until this is voted GO.**
+
+## 1. Problem — the cutoff heuristic is inherently incomplete
+
+The threshold (门槛会签) quorum tally counts DISTINCT approver ids among append-only approve records at node X.
+Because X can be re-entered (return/jump to X, or return to an ancestor then forward into X), prior-round
+approves get re-counted → a single fresh vote re-satisfies an N-of-M. Every fix so far reconstructs "which
+round am I in" by pattern-matching `nextNodeKey`/`targetNodeKey`/`toNodeKey` across records (the `cutoff`):
+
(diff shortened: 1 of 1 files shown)


zensgit:
Rev 2 pushed — both REQUEST-CHANGES points addressed:

**P1 (insertAssignments shared / activation-vs-mutation):** §4 now splits the two semantics explicitly. **(A) Activation insert** (initial / forward advance / return / jump / timeout-jump) mints a new epoch (bump `node_activation_seq`). **(B) Same-round mutation** — add-sign, reduce-sign, **manual transfer, timeout transfer, admin reassign/handover** — preserves the current node epoch and **never bumps**, with an explicit *caller-passes-epoch* implementation contract (so a shared "insert → bump" rule can't split a round or leave a transferred-in approve at NULL epoch). §7 edge matrix + §9 add the three mid-round-mutation cases (new assignee's approval counts under the same `nodeEntryEpoch`; `node_activation_seq` unchanged).

**P2 (active + fail-closed on mixed epoch):** §5 resolves the current epoch from `is_active = TRUE` assignments via `SELECT DISTINCT entry_epoch`; all-NULL → cutoff fallback; exactly-one-non-NULL → tally; **mixed or multiple distinct epochs → fail closed + structural-error log**, not `MAX()`-collapse. §9 adds the mixed-epoch fail-closed test.

Core mechanism / dual-read migration / condition-cc-by-construction unchanged. Ready for re-review.

PR #3502, opened 2026-07-03
docs(attendance): MP-4 makeup-punch admin config-card design-lock (RATIFIED)

zensgit:
## Why

MP-2/MP-3 makeup-punch enforcement is live on main, but there is no admin surface to configure `makeupPunchPolicy` — the only "runtime without a config face" gap in the attendance line. This locks the MP-4 admin config card so the implementation is a mechanical, separately-opt-in slice. **Design-lock-first: this PR is docs-only; the Vue card + behavioral test does not begin until this lock is ratified.**

## What (1 file, +117)

`docs/development/attendance-makeup-punch-admin-config-card-design-lock-20260702.md`, Status **PROPOSED**:

- **§1 Scope** — MP-4 = the config card only. Frontend-only: `makeupPunchPolicy` already exists end-to-end on main (`DEFAULT_SETTINGS`/`mergeSettings`/normalizer/zod/MP-2-3 enforcement), so no backend/migration change. Explicitly OUT: MP-5 (request UX), MP-6 (staging), and the AE staging-window closure acceptance (feeds none of the AE-4/RD-4/5/v1-8 PASS stamps).
- **§2 Policy contract** — a field table with one row per top-level `makeupPunchPolicy` field (9 fields; `cycle`/`quota`/`submitWindow` expanded to sub-rows), each with its default, its enum-strict zod bound, and a verified "runtime reads it?" column. Coverage is defined by the table (no bare prose count). Honors the ratified §9 numbers.
- **§3 Card UI** — mirrors the RD-4 report-digest card precedent (`#3283`/`97b363977`) exactly: same 4-file set, `data-makeup-punch=*` selectors, control types per field. States the **honest anchor-nav consequence**: adding a rail section bumps `attendance-admin-anchor-nav.spec.ts` 32 → 33 at both assertions (lines 127 and 835) + adds the label — required, not a no-op.
- **§4 Load/save** — `applyMakeupPunchPolicyToForm` into `loadSettings`; `saveMakeupPunchPolicy` PUTs only the `makeupPunchPolicy` key with full sub-objects (deep-merge discipline); `showAdmin`+`shouldShowAdminSection`+403→`adminForbidden` gate (not `provisionUserIsAdmin`).
- **§5 Enablement preconditions** — the RBAC single-tenant posture, elevated to a first-class section: `attendance:admin` is a global permission and `org_id` is a partition key, not an auth boundary, so this card must be enabled only under a single-tenant posture — and that line goes into the **release/operator note** at customer enablement, not an issue comment.
- **§6 Test plan** — put the behavioral test inside the already-guarded `attendance-admin-regressions.spec.ts` so no `attendance-web-guard` workflow edit is needed (avoids skip-shaped-green); update the anchor-nav counts.
- **§7 Deferrals** — runtime is a separate opt-in; MP-5/MP-6 stay gated.

## Verification

Built via a grounded design workflow (parallel readers over the runtime contract, the RD-4 UI precedent, and scope/RBAC) then self-adversarially reviewed: every §2 key/default/zod bound checked against `index.cjs:284`/`:21407`; the "runtime reads it?" column against the MP-2/3 enforcement path `:26735-26781`; the anchor-nav counts against the live spec (`toHaveLength(32)`@127, `.toBe(32)`@835); the §5 route/permission citations re-verified and corrected against origin/main (GET/PUT settings `:41083`/`:41408`, `getOrgId:5823`, `withPermission:19911`). No bare field count, no external brand names, exactly one new file (no runtime code).

## After ratify

The runtime slice is a single frontend-only PR mirroring `97b363977`'s 4-file set (card + rail id/nav + the two test edits). MP-5/MP-6 remain separate owner-gated slices.

### DIFF
--- a/docs/development/attendance-makeup-punch-admin-config-card-design-lock-20260702.md
+++ b/docs/development/attendance-makeup-punch-admin-config-card-design-lock-20260702.md
@@ -1,8 +1,8 @@
-# Attendance MP-4 — makeup-punch admin config card design-lock
+# Attendance MP-4 — makeup-punch admin config card design-lock (RATIFIED)
 
-**Status:** PROPOSED (awaiting owner ratify) · **Date:** 2026-07-02 · **Slice:** MP-4 (admin config UI for `makeupPunchPolicy`)
+**Status:** RATIFIED (owner review, 2026-07-02) · **Date:** 2026-07-02 · **Slice:** MP-4 (admin config UI for `makeupPunchPolicy`)
 
-This design-lock writes **no runtime code**. It locks the shape of the makeup-punch admin config card so the follow-up implementation (Vue card + behavioral test) is a mechanical, separately-opt-in slice. The runtime slice does not begin until this lock is ratified.
+This design-lock writes **no runtime code**. It locks the shape of the makeup-punch admin config card so the follow-up implementation (Vue card + behavioral test) is a mechanical, separately-opt-in slice. The runtime slice remains separate and starts only after an explicit owner opt-in.
 
 ---
 
--- a/docs/development/attendance-makeup-punch-admin-config-card-design-lock-20260702.md
+++ b/docs/development/attendance-makeup-punch-admin-config-card-design-lock-20260702.md
@@ -1,0 +1,118 @@
+# Attendance MP-4 — makeup-punch admin config card design-lock
+
+**Status:** PROPOSED (awaiting owner ratify) · **Date:** 2026-07-02 · **Slice:** MP-4 (admin config UI for `makeupPunchPolicy`)
+
+This design-lock writes **no runtime code**. It locks the shape of the makeup-punch admin config card so the follow-up implementation (Vue card + behavioral test) is a mechanical, separately-opt-in slice. The runtime slice does not begin until this lock is ratified.
+
+---
+
+## §1 Scope
+
+**In scope (MP-4):** an admin settings card that reads and writes the already-live `makeupPunchPolicy` org setting through `GET`/`PUT /api/attendance/settings`, plus its admin-rail registration and the two test edits that registration forces. Nothing else.
+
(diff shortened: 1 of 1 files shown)


zensgit:
Revised after review (force-pushed `257083c63`) — two findings fixed, still PROPOSED / not ratified:

**[P1] `allowedRequestTypes` was mis-marked as runtime-consumed.** Verified against `origin/main`: the request-type gate uses the frozen `MAKEUP_PUNCH_ALLOWED_REQUEST_TYPES` constant (`index.cjs` ~L12650, referenced ~L12882/26737/27940/27979/28562); `policy.allowedRequestTypes` is normalized + zod-validated but **never read** by `enforceMakeupPunchPolicy`/`buildMakeupPunchPolicySnapshot`. Fix: the §2 "runtime reads it?" column is now three-state (**Enforced** / **Snapshot-only** / **Not consumed**); `allowedRequestTypes` = Not consumed and the card renders it **read-only/informational** (not an editable checkbox group that would save-but-do-nothing). Chose the owner's option 1 (keep MP-4 frontend-only, read-only field); option 2 (wire the field into enforcement) is written up as a separate owner-gated **backend** slice in §7, since it changes accept/reject behavior. The same audit reclassified `cycle.type`/`quota.principal`/`submitWindow.unit` as Snapshot-only (single-valued, no behavioral branch) — also read-only in the card.

**[P2] line-ref drift.** `normalizeMakeupPunchPolicySetting` is defined ~L12671 (L12165 is only its call site); settings routes re-verified at ~L41083 (GET) / ~L41408 (PUT). All citations switched to **symbol + approximate `~L`** anchors so the design-lock doesn't hard-lock a drifting line.

No bare field count, no brand names, one file, no runtime code. Ready for a fresh ratify decision.

PR #3504, opened 2026-07-03
docs(approval): T3-6 approvals-as-multitable-records design-lock (proposed)

zensgit:
Makes the highest-differentiation remaining item (T3-6 — approval data as first-class multitable records) build-ready. Turns the third-batch ballot (Q1–Q10 + build contract) into an implementable architecture:

- **One-way read-model projection** (engine authoritative; editing the row never mutates the approval).
- **Triggers:** an explicit **create hook** (completion events only fire terminal, so pending-create needs it) + the existing `ApprovalCompletionEventV1` bus for terminal; auto-approve-at-create → the SAME row.
- **Columns:** allowlisted system fields only (requestNo/template/status/requester/approver/timestamps/outcome); no `form_snapshot` (PII gated behind Q7).
- **Link + idempotency:** side mapping table `instance_id PK → record_id` (not a `meta_records` column); idempotent via the T2-6 ledger pattern / `projected_version`.
- **Sheet:** lazily-provisioned, admin/owner-scoped, system-managed, per template family.
- Best-effort + observable (never fails the approval flow); fail-first verification plan (create/terminal/auto-approve/redelivery/no-write-back/visibility/best-effort).

**PROPOSED — no runtime until the T3-6 votes.** The payoff: approval analytics becomes "reuse multitable" (飞书-grade reporting + re-automation for free) instead of a bespoke reporting module.

### DIFF
--- a/docs/development/t3-6-approvals-as-multitable-records-design-lock-20260703.md
+++ b/docs/development/t3-6-approvals-as-multitable-records-design-lock-20260703.md
@@ -3,8 +3,13 @@
 > **Status: PROPOSED design-lock — awaiting the T3-6 votes** in
 > `approval-automation-third-batch-ballot-20260702.md` (Q1–Q10). This turns the ballot decisions into an
 > implementable architecture. **No runtime until voted GO.** T3-6 is the differentiation move: once approval
-> data is a multitable read-model, 飞书-grade reporting / dashboards / re-automation come for free on the
-> existing multitable analytics — instead of a bespoke approval-reporting module.
+> data is a multitable read-model, **飞书-grade reporting / dashboards / views / formulas come for free** on the
+> existing multitable analytics — instead of a bespoke approval-reporting module. **Re-automation ON the
+> projected rows is deliberately v1-OUT** and specified as a gated follow-up (§6a) — see the P1 decision.
+>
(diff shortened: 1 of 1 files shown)


zensgit:
Rev 2 pushed — both load-bearing decisions filled:

**P1 (§6a) — projection event semantics:** decided **event-SILENT in v1**. The projection write does NOT emit `record.created`/`record.updated` (direct system-path materialization). This scopes 'for free' precisely to **reporting/dashboards/views/formulas** — NOT re-automation — and eliminates the projection→automation loop + permission surface. Re-automation on projected rows is a named **v2** requiring source-marker + loop/depth guard + `approval.projection:{instanceId}:{version}` dedup + a permission boundary. (Fixes the earlier over-broad 're-automation for free'.)

**P2 (§6b) — reconcile/backfill is v1, not 'noted':** `reconcileApprovalProjection(instanceId)` replays the authoritative state from `approval_instances` + `approval_records` → idempotent upsert into the mapping row **guarded by `projected_version`** (never clobbers a newer live projection); a leader-gated **sweep** finds rows behind their source (or missing) + an **on-demand admin endpoint**; observability via `projected_version`/`last_projected_at`/`last_error` + an `approval_projection_stale_total` drift metric. §7 adds the event-silence + reconcile fail-first tests.

Ready for re-review.

PR #3505, opened 2026-07-03
feat(approval): T1-4 node field-permissions authoring (linear editor, hidden+readonly)

zensgit:
Implements T1-4 per its build-spec + your hidden/readonly steer. **Scope correction (honest):** the runtime `hidden` echo-redaction was **already shipped in #2799** (`redactHiddenFormFields` wired in `ApprovalBridgeService.toUnifiedDTO`) — so this PR builds only the genuinely-missing **authoring UI** + the **§4 contract tests**, not a duplicate.

## What ships
- **Authoring UI** (linear step editor, `TemplateAuthoringView.vue`): per-step, per-form-field access selector (可编辑/只读/隐藏) → persists `node.config.fieldPermissions`. `editable` is the absent default (untouched linear templates stay byte-stable). `readonly` carries a non-blocking "enforced in a later slice (T1-4b)" hint.
- **Owner reconcile settled ✏️:** hidden AND readonly are authorable; **only `hidden` is runtime-enforced** this rung (via the pre-shipped redaction); `readonly` persists + round-trips but is **runtime-inert**.
- **Never-flatten / fail-closed:** `fieldPermissions` added to the linear `allowedConfigKeys`; `buildStepConfig` prunes deleted fields; a malformed entry fails **closed** to read-only (never silently re-saved); complex graphs preserve `fieldPermissions` verbatim (deep-clone path). A routing-driver-hidden authoring hint (form_field_user).

## Verification (independently re-run)
- **FE `vue-tsc -b` 0** · authoring tests **77/77** (field-permissions **11 new** — hydrate/emit/allowlist/never-flatten; `approvalTemplateAuthoring` 46 incl. 2 new mounted write-wire + routing-hint; graph-preserve 20). **RED-before (genuine, FE):** pre-change round-trip drops `fieldPermissions` / locks read-only / routing-driver template errors → GREEN after.
- **Backend `tsc` 0** · `approval-p1c-field-permissions` **7/7** real-DB (4 existing + **3 new §4 locks**): hidden field redacted from the node echo while assignee resolution (form_field_user) + condition routing still read the stored snapshot (echo-only proof); readonly round-trips runtime-inert.

## Caveats (honest)
- The 3 backend contract tests are **regression locks** (no RED-before — they lock behavior #2799 already ships correctly).
- 17 pre-existing FE suite failures (featureFlags/multitable/attendance/etc., `auth.getToken` fixtures) — proven pre-existing by stash-and-rerun; **not caused by this change**; touched areas fully green.

### DIFF
--- a/apps/web/src/approvals/templateAuthoring.ts
+++ b/apps/web/src/approvals/templateAuthoring.ts
@@ -14,6 +14,8 @@ import type {
   FormFieldVisibilityRule,
   FormOption,
   FormSchema,
+  NodeFieldAccess,
+  NodeFieldPermission,
   CreateApprovalTemplateRequest,
   UpdateApprovalTemplateRequest,
 } from '../types/approval'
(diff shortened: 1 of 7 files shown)


zensgit:
P2 fixed — malformed `fieldPermissions` now fails closed, no silent flatten. Extracted one full-validation predicate `isBackendDroppedFieldPermission` (non-object OR extra key OR **non-string/empty fieldId** OR **out-of-enum access**) used by the linear guard, `complexApprovalConfigHasBackendDrop`, AND the hydrate filter (single source of truth). Added 2 fail-first tests (out-of-enum access / non-string fieldId → unsupported/read-only); **RED-before confirmed** (both fail with the shape-only guard). vue-tsc 0 · authoring 79/79.

PR #3506, opened 2026-07-03
feat(approval): T3-5 W7 cross-base resultWriteback (implements design-lock)

zensgit:
Implements T3-5 per its ratified design-lock — extends W7 `resultWriteback` to a **literal cross-base target**, reusing the executor's ratified cross-base write gate.

## What ships (design-lock §2)
- **Config (Q2):** `resultWriteback` gains optional literal `targetBaseId`+`targetSheetId`+`targetRecordId` (no templating). Save gate: all-three-or-none required when any is set (create + update).
- **Effective actor (Q1):** the **trigger actor** (`readTriggerActorId(bridge.triggerEvent)` → `triggerEvent.actorId`), NOT the resume actor. Null/system approvals **fail closed** at the gate (no owner/requester fallback). Same actor governs the target-record lock. VALUES still come from the event (approver = `event.actor?.id`).
- **Gate reuse (§3.4):** extracted the executor gate to a public context-agnostic `evaluateCrossBaseWriteGate(queryFn, actorId, triggerSheetId, targetSheetId, declaredTargetBaseId)`; the 4 record-action call sites keep the old adapter signature; the service calls it on the **same executor instance** → **shares the singleton per-target-base quota** with update/create/delete/lock (Q5).
- **Audit (Q3):** no new table; step output carries the target triple; actor omitted from the target-base realtime fan-out.

## Verification (independently re-run)
tsc 0 · **start-approval 22/22** (16 existing W7 + 6 new T3-5) · cross-base gate suites **30/30** incl. a new **Q-6 quota-sharing** test. **RED-before confirmed** — incl. the sharpest **anti-misroute** (a source-resident field id + cross-base triple mutated the SOURCE record pre-change; untouched post-change). Proven: authorized target write · unauthorized fail-closed · null-actor fail-closed (no requester fallback) · **source untouched** · **non-merge into resume tail** (tail webhook lacks the cross-base patch) · step-output triple · target-sheet fan-out. Self-reviewed (2 finder agents: correctness clean; 4 cleanup items applied) + advisor.

## Named follow-ups (honest, disclosed)
- **`onNonApproved` + cross-base** is reachable (a non-approved terminal with `onNonApproved:true` + a cross-base triple routes through the same cross-base branch) but **not pinned by a test** — by-construction covered by T3-4's `onNonApproved` gating × T3-5's cross-base branch (both independently tested); the reject-flow test needs custom plumbing. Worth a follow-up test.
- Realtime **actor-omission** is implemented but asserted only via the eventBus emit, not a publish spy.
- **Resume-retry quota double-spend** documented as a v1 limitation in-code.

### DIFF
--- a/packages/core-backend/src/multitable/automation-executor.ts
+++ b/packages/core-backend/src/multitable/automation-executor.ts
@@ -1793,8 +1793,11 @@ export class AutomationExecutor {
    * record-create guard at record-service.ts). The state is unreachable today (sheets are hard-deleted)
    * but this hardens against a future sheet-soft-delete feature.
    */
-  private async resolveSheetBaseId(sheetId: string): Promise<string | null | undefined> {
-    const res = await this.deps.queryFn('SELECT base_id FROM meta_sheets WHERE id = $1 AND deleted_at IS NULL', [sheetId])
+  private async resolveSheetBaseId(
+    sheetId: string,
+    queryFn: AutomationDeps['queryFn'],
+  ): Promise<string | null | undefined> {
+    const res = await queryFn('SELECT base_id FROM meta_sheets WHERE id = $1 AND deleted_at IS NULL', [sheetId])
     const row = (res.rows as Array<{ base_id?: unknown }>)[0]
     if (!row) return undefined
     return typeof row.base_id === 'string' ? row.base_id : null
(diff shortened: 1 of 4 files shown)


zensgit:
Fixed the CI blocker + added the suggested test.

**Lock-guard marker (CI red):** the `multitable-record-lock-guard.guard.test.ts` structural audit requires a disposition marker on the line above every `UPDATE meta_records`. The cross-base UPDATE already calls `ensureRecordNotLocked` immediately above (with the trigger actor), so this was a missing audit marker, not a missing check — added `// lock-guarded: …` (backed by the real call in the same file, as the guard requires). **Guard test now 4/4.**

**Locked-target test (your suggestion):** new test — the cross-base target record is locked by ANOTHER actor while the trigger actor IS authorized to write the target base → **backwriteSkipped, target + source unchanged**. Pins that `ensureRecordNotLocked` is evaluated with the TRIGGER actor for the cross-base target. start-approval now 23/23, tsc 0.

zensgit:
Correction to the previous comment: my shell interpreted the inline backticks while posting. The actual P2 follow-up is:

- Runtime: after the shared cross-base gate resolves the target, `gate.crossBase === false` now throws `cross-base resultWriteback target must resolve to a different base`, surfaced as `backwriteSkipped`; no source write and no same-base target write.
- Test: added a real-DB resume-path regression where `targetBaseId = source base`, `targetSheetId = source sheet`, and `targetRecordId` points at a different same-base record. It asserts source and same-base target stay unchanged and no `crossBaseBackwrite` output is recorded.

Local verification:
- `pnpm --filter @metasheet/core-backend build` ✅
- `git diff --check` ✅
- The integration suite requires a migrated `DATABASE_URL`; sourcing the local `.env` hit an unmigrated DB (`automation_rules` missing), so the new real-DB test is left to the PR CI lane.

PR #3511, opened 2026-07-03
docs(attendance): ratify MP-5 makeup-punch request UX lock

zensgit:
## Summary
- Add the MP-5 makeup-punch request-side UX design-lock as PROPOSED.
- Scope the runtime to Request Center UX for `MAKEUP_PUNCH_*` rejections, makeup attachment URL exposure, shared anomaly/reminder prefill path, and guard wiring.
- Keep runtime, MP-6 staging, org-configurable `allowedRequestTypes`, and backend enforcement changes as separate opt-ins.

## Review notes
- Subagent review initially caught two blockers: `attendance-selfservice-dashboard.spec.ts` is not currently in `attendance-web-guard`, and `MAKEUP_PUNCH_ATTACHMENT_REQUIRED` would point users at a non-rendered field.
- The lock now requires the runtime PR to wire the chosen spec into `attendance-web-guard`, and it includes the makeup `attachmentUrl` field + retry-body proof in MP-5 scope.

## Verification
- `git diff --check`
- brand / bare closing-keyword grep clean
- subagent re-review: no remaining P1/P2/P3

### DIFF
--- a/docs/development/attendance-makeup-punch-request-ux-design-lock-20260703.md
+++ b/docs/development/attendance-makeup-punch-request-ux-design-lock-20260703.md
@@ -1,0 +1,147 @@
+# Attendance MP-5 — makeup-punch request-side UX design-lock (PROPOSED)
+
+**Status:** PROPOSED · **Date:** 2026-07-03 · **Slice:** MP-5 (Request Center UX for `makeupPunchPolicy` rejections)
+
+This document proposes the request-side UX slice that follows the ratified makeup-punch policy runtime and the MP-4 admin config card. It writes **no runtime code**. Runtime starts only after owner ratification and remains a separate PR.
+
+---
+
+## §1 Scope
+
+**In scope (MP-5):** make the employee Request Center explain `MAKEUP_PUNCH_*` policy rejections in a human-actionable way, expose the already-supported `attachmentUrl` field for the three makeup request types, and prove anomaly / missed-punch-reminder prefilled drafts still submit through the same `POST /api/attendance/requests` path.
+
+**Explicitly OUT:**
+
+- No backend enforcement change. `enforceMakeupPunchPolicy` remains the server authority for quota/window/type/reason/attachment.
+- No admin policy-card changes. MP-4 already shipped the config surface; this slice reads no settings directly.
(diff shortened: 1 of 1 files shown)


zensgit:
Independent adversarial review (Claude, second reviewer) — **APPROVE, ratify-ready, no P1/P2.** Re-derived every load-bearing claim against `origin/main` rather than trusting the prior pass:

- **All 7 `MAKEUP_PUNCH_*` codes are real + server-emitted** (`plugins/plugin-attendance/index.cjs`, 1 hit each) — no invented code.
- **`MAKEUP_PUNCH_CROSS_USER_FORBIDDEN` is PUT/update-path only** (`index.cjs:27950`, cross-user edit guard) — §3/§6 correctly forbid faking a POST path to cover it.
- **Attachment field is leave-only today** (`AttendanceView.vue:895`, `v-if="isLeaveRequest"`) and **the submit payload already sends `attachmentUrl` unconditionally** (`:19742`, POST at `:19745`) — so MP-5 exposes (not invents) the field for makeup types. Accurate.
- **`attendance-selfservice-dashboard.spec.ts` exists but is NOT in `attendance-web-guard`** (run-list line 77 has no selfservice entry) — §6 gate 7 / §7's "add it to the workflow or it's skip-shaped-green" is correct. Mirrors the AE-3 F1 pattern.
- All named FE symbols exist (`submitRequest`, `createApiError`, `prefillRequestFromAnomaly`, `openMissingPunchQuickAction`, `classifyStatusError`, `setStatusFromErrorWithContext`). Citations are by symbol name (drift-proof). No brand names, docs-only, no runtime.

The two earlier blockers (web-guard inclusion, attachment field) are correctly fixed.

**One optional P3 for the runtime PR (not a lock defect):** when exposing the attachment field for makeup types, broaden the existing `v-if="isLeaveRequest"` on `#attendance-request-attachment` (e.g. `|| isMakeupRequest`) rather than add a second input with the same `id` — a duplicate DOM `id` would break the selector and the §6 gate-6 retry assertion.

Full review: `/tmp/pr3511-review-claude-20260703.md`. Ratify/merge is the owner's call.

PR #3512, opened 2026-07-03
feat(approval): T3-3 node signaturePolicy (declared-inert slice 1)

zensgit:
Third-batch **T3-3 first rung** on the ballot default (Q2 **declared-inert**): a node carries an optional `signaturePolicy { required, kind?, appliesTo? }` that is **persisted + round-tripped** through the approval graph but **NOT enforced at runtime** — approve/reject never blocks on a signature until enforcement is a separately-ratified rung (mirrors the `fieldPermissions` readonly / `auto_*` declared-but-do-not-wire precedent).

## What ships
- `SignaturePolicy` type + `signaturePolicy?` on `ApprovalNodeConfig`. `kind` is **contract-open** (Q1: v1 authoring will offer typed/click; image capture is a later slice); `appliesTo` defaults **approve-only** (Q7).
- Shape-only normalizer `normalizeNodeSignaturePolicy` (required boolean · kind non-empty string · appliesTo ∈ {approve, approve_reject}) — **fails closed** on a malformed value, never a coerced default / silent flatten. Wired into the node-config whitelist next to `fieldPermissions`/`timeout`; **default-absent nodes stay byte-identical** (key omitted).
- No SQL migration (lives in node-config JSON). Q8 signature-ref-in-record + runtime enforcement + authoring UI are **later rungs** (this is the inert floor the ballot's build contract asks to prove first).

## Verification
tsc 0 · new real-DB **`approval-node-signature-policy` 5/5**: create→publish→reload **round-trip byte-identical** · **default-absent byte-identical** · **DECLARED-INERT** (a `required:true` policy does NOT block approve with no signature) · malformed `required`/`appliesTo` **rejected at save**. **RED-before confirmed** (round-trip fails when the whitelist emit is removed → policy dropped). Regression: p1c field-permissions 7/7 · nofm-threshold 6/6.

Per the T3-3 build contract: proves `signaturePolicy` survives normalize→publish→reload→re-normalize, default-absent stays byte-identical, and the floor cannot imply runtime enforcement.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -347,6 +347,7 @@ jobs:
             tests/integration/approval-node-timeout-effects.test.ts \
             tests/integration/approval-nofm-threshold.test.ts \
             tests/integration/approval-bulk-reassign.api.test.ts \
+            tests/integration/approval-node-signature-policy.api.test.ts \
             --reporter=dot
 
       - name: Run comment-reaction keystone (real wire, real PG)
--- a/packages/core-backend/src/services/ApprovalProductService.ts
+++ b/packages/core-backend/src/services/ApprovalProductService.ts
@@ -988,6 +988,12 @@ function normalizeNodeSignaturePolicy(
   if (!isRecord(value)) {
     failValidation(context, `${path} must be an object`)
   }
+  const allowedKeys = new Set(['required', 'kind', 'appliesTo'])
+  for (const key of Object.keys(value)) {
+    if (!allowedKeys.has(key)) {
+      failValidation(context, `${path}.${key} is not supported`)
+    }
+  }
   if (typeof value.required !== 'boolean') {
     failValidation(context, `${path}.required must be a boolean`)
   }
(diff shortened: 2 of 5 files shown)


zensgit:
Review fixes pushed in `2407ff854`:\n\n- Closed the signaturePolicy unknown-key drop: `normalizeNodeSignaturePolicy` now rejects any key outside `required` / `kind` / `appliesTo`, so a future/unsupported nested key cannot be silently flattened by the backend allowlist rebuild.\n- Added fail-first coverage for that path, plus an explicit `required:false` preservation test so false is not confused with default-absent.\n- Wired `approval-node-signature-policy.api.test.ts` into the approval real-DB CI lane and excluded it from the default no-DB config, so it cannot skip-green.\n\nLocal verification: `git diff --check`, `pnpm --filter @metasheet/core-backend build` pass. Target integration runs 5/6 locally; the remaining failure is the local DB missing the already-main `node_activation_seq` migration column, so CI's migrated real-DB lane is the authoritative run for that file.

PR #3526, opened 2026-07-03
feat(integration): resolver_lookup R2 — S3-1 runtime executor integration, standalone-only (#1709)

zensgit:
## What

**R2** of resolver_lookup — the S3-1 runtime executor integration (owner-authorized 2026-07-03 with hard boundaries). Wires the merged R1 pure evaluator into `executeConfiguredRead` so an approved resolver_lookup config actually executes at runtime. **Backend only — ZERO frontend changes** (the resolver config-authoring form is R3, explicitly out of R2).

## How

- `read-source-read-runtime.cjs`: `resolver_lookup` added to `CONFIGURED_READ_SUPPORTED_MODES`; the S1-normalized `config` (byte-equal to the validator's normalized output per assertS1NormalizedConfig) threaded through `prepareConfiguredRead`; in `executeConfiguredRead`, after the outbound keyed read yields `raw`, resolver_lookup **returns `evaluateResolver(config, raw, {rowCap})` verbatim**, BYPASSING the generic field-map data plane (which stays for the other three modes).
- **Outbound read reuses the existing builder** — `buildReadSourceProbeRequest` already routes resolver_lookup down the single_record keyed branch (filter by keyField=inputs.key); the candidate ARRAY container is what the evaluator selects from. No new outbound path.

## Hard boundaries (owner-listed, enforced + seam-commented + tested)

- **No composition** — the resolver resolves one key to one value and returns it; its output is NEVER fed into a second read / BOM lookup / chained call. Test asserts **exactly ONE outbound read** (`readArgs.length === 1`).
- **No recursive BOM, no material→FBillNo chaining.**
- **No write** — read-only; test asserts `writeCalls.length === 0`.
- **No UI expansion** — zero frontend changes (config form is R3).
- **approved-only** (getForRuntime; draft/retired → 409), **key-only** ({inputs} strict allowlist), **values-free** (evidence leak-scanned: resolved value, field names, candidate values, key, credential, path all absent).

## Verification

Full plugin `pnpm test` exit 0 (all read-source / resolver / http-routes suites); type-check clean. Route test covers: exactly_one resolves → data:{resolver:{target,value}}; >1 → AMBIGUOUS + data:null (still 200 — a resolved outcome); container-missing coarse code; single outbound read (no chaining); zero write; values-free leak scan; approved-only 409. Reviewed the no-composition seam (immediate return, no downstream), no-write reachability, and config-threading safety directly.

Completes the owner-authorized resolver R0+R1+R2 (standalone). **R3 (route/config UI), composition, recursive BOM, and write remain separately gated.**

### DIFF
--- a/plugins/plugin-integration-core/__tests__/http-routes.test.cjs
+++ b/plugins/plugin-integration-core/__tests__/http-routes.test.cjs
@@ -6428,6 +6428,109 @@ async function testReadSourceConfiguredReadRoute() {
   console.log('  testReadSourceConfiguredReadRoute OK')
 }
 
+// R2 (#1709): resolver_lookup end-to-end through the S3-2 route — approved-only, key-only, read-tier,
+// standalone (data carries ONLY the resolver target+value), no write, values-free.
+async function testReadSourceResolverReadRoute() {
+  const readArgs = []
+  const writeCalls = []
+  const resolverConfig = {
+    version: 1,
+    systemId: 'sys_1',
+    requiredKind: 'erp:k3-wise-webapi',
+    object: 'material',
+    mode: 'resolver_lookup',
+    readPath: '/K3API/Material/GetList',
+    readMethod: 'POST',
+    operations: ['read'],
+    keyField: 'FNumber',
+    containerPaths: ['Data.Rows'],
+    resolverRule: 'exactly_one',
+    fieldMap: [{ source: 'FItemID', target: 'item_id' }],
+  }
+  const services = createMockServices({
+    externalSystemRegistry: {
+      async getExternalSystemForAdapter(input) {
+        return { id: input.id, kind: 'erp:k3-wise-webapi', credentials: { bearerToken: 'secret-token' }, config: { objects: {} } }
+      },
+    },
+    readSourceConfigStore: {
+      async saveVersion() { return {} },
+      async list() { return [] },
+      async get() { return {} },
+      async approve() { return {} },
+      async retire() { return {} },
+      async listAudit() { return [] },
+      async getForRuntime(input) {
+        if (input.id === 'rsc_draft') {
+          const error = new Error('not approved')
+          error.name = 'ReadSourceConfigNotApprovedError'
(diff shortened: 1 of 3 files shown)


zensgit:
Review result from this pass: PASS / no blocking findings.

I re-read the live head `b57a9afe6ed267b65dd83d70ba2a52523d9187ca` rather than relying on the existing approval summary. The R2 boundary holds: `resolver_lookup` is only admitted through the approved-config runtime path, the request body remains strict `{inputs}`, outbound is the existing keyed read builder, `executeConfiguredRead` returns the R1 evaluator outcome directly with no downstream read/write composition, and resolver evidence is still routed through the values-free `readSourceProbeEvidence` sanitizer.

Local verification run from an isolated worktree:
- `node plugins/plugin-integration-core/__tests__/read-source-read-runtime.test.cjs`
- `node plugins/plugin-integration-core/__tests__/read-source-resolver-evaluator.test.cjs`
- `node plugins/plugin-integration-core/__tests__/http-routes.test.cjs`

CI note: `test (20.x)` is green; `coverage` was still pending at my last check. Merge mechanics still need the branch to catch up with main.

Low / non-blocking nit: `plugins/plugin-integration-core/__tests__/read-source-read-runtime.test.cjs:55-56` still says S3-1 fail-closes resolver_lookup until R2 wires the evaluator. That is stale on this PR; it should say the fixture is the R2-supported resolver config shape. This does not affect runtime or test coverage, but it is worth cleaning to avoid the stale-audit trap we keep tripping over.

Note: GitHub will not let this account submit an APPROVE review on a PR authored by the same account, so this is recorded as a review comment rather than a formal approval state.

zensgit:
Low nit fixed on the PR branch.

Changed only the stale resolver fixture comment in `plugins/plugin-integration-core/__tests__/read-source-read-runtime.test.cjs` so it now reflects the R2-supported runtime path instead of the old pre-R2 fail-closed state.

Local verification: `node plugins/plugin-integration-core/__tests__/read-source-read-runtime.test.cjs` → PASS.

Commit: `33cecfd18`.

PR #3535, opened 2026-07-03
feat(approval): surface requester/department analytics in the SLA dashboard (arc B)

zensgit:
Arc **B** — with an honest reorientation surfaced by grounding: the approval **SLA dashboard already ships** (`ApprovalMetricsView.vue` at `/approvals/metrics`, `requiresAdmin`, nav-linked "审批 SLA") and the metrics endpoints (`/summary /report /people /teams /breaches /instances`) already exist and are permission-tested. The one **computed-but-unsurfaced** metric was `getMetricsByDimension` (per-requester / per-department). So this slice **completes the existing dashboard**, it does not build one.

## What ships (FE-only; backend + routes + nav untouched, already existed)
- Two read-only sections added to the existing view: **按部门汇总** (`/teams`, `approvals:admin`) and **按发起人汇总** (`/people`, `approvals:analytics`) — 总量/通过/驳回/撤回/平均耗时 + SLA 超时率 with inline breach-rate bars, wired into the shared date-range `loadAll`.
- `api.ts`: `fetchApprovalMetricsTeams` / `fetchApprovalMetricsPeople` (GET-only, since/until/limit).

## Verification
`vue-tsc -b` 0 · new `approvalMetricsView.spec.ts` **5/5**, **RED-first verified** (4/4 failing before the view change): sections render from mocked fetchers with exact args; requester section present+fetched only with `approvals:analytics`, hidden+not-fetched otherwise; read-only; shared date-range propagates since/until. Backend `tsc` 0; `approval-metrics-router` 9/9; `approval-metrics-people-teams` 3/3 real-DB; existing approval FE specs 61/61.

## Honest scope notes
- **The deeper T3-6 differentiation payoff** — building multitable-native views/charts *on the projection base* ("approval analytics = reuse multitable") — is **NOT this slice**; it's a separate, larger slice gated behind arc A's visibility. B completes the *aggregate* SLA dashboard.
- FE `approvals:analytics` gate is defense-in-depth (the page is `requiresAdmin` so it short-circuits true for admins) — the real gate is the backend `rbacGuard('approvals:analytics')`; the test exercises both branches.
- No chart added (endpoints are aggregates, not time-series → inline CSS bars, no new dep); aggregate node-breakdown out of scope (no backend aggregate — would invent metrics math); i18n consistent with the module (no i18n infra).

### DIFF
--- a/apps/web/src/approvals/api.ts
+++ b/apps/web/src/approvals/api.ts
@@ -594,6 +594,22 @@ export interface ApprovalMetricsReport {
   breachedTemplates: ApprovalMetricsTopTemplateRow[]
 }
 
+// Mirrors backend `MetricsDimensionRow` — one aggregation bucket keyed by the
+// requester (person) or the requester's frozen department (team). `key` is the
+// requester id / department string (`null` = the unattributed bucket); `name`
+// is a representative display name. Note: the backend surfaces `slaBreachRate`
+// (a ratio) here, NOT the raw breach/candidate counts.
+export interface ApprovalMetricsDimensionRow {
+  key: string | null
+  name: string | null
+  total: number
+  approved: number
+  rejected: number
+  revoked: number
+  avgDurationSeconds: number | null
+  slaBreachRate: number
+}
+
 export interface ApprovalMetricsNodeBreakdownEntry {
   nodeKey: string
   activatedAt: string | null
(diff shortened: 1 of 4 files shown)


zensgit:
Reviewed current head `543a3b7a`.

Verdict: APPROVE on content. This is correctly scoped as a FE-only completion of the existing Approval Metrics dashboard, not a new analytics subsystem:

- `fetchApprovalMetricsTeams` / `fetchApprovalMetricsPeople` are GET-only wrappers over the already-existing `/teams` and `/people` endpoints, preserving `since` / `until` / `limit` query shape.
- `ApprovalMetricsView.vue` wires both sections into the existing shared date-range `loadAll`; department aggregation is always shown on the admin-gated page, while requester/person aggregation is hidden and not fetched unless `hasPermission('approvals:analytics')` is true.
- Backend hard gates are already in place and tested: `/people` is `rbacGuard('approvals:analytics')`, `/teams` is `rbacGuard('approvals:admin')`, with router tests covering admin-only vs analytics-only 403/200 behavior.
- Service grounding is real: `getMetricsByDimension` groups via `approval_metrics` joined to `approval_instances.requester_snapshot`, and the existing real-DB test proves both requester and `directoryDepartment` department grouping.
- The new FE spec covers the load-bearing UI behavior: section rendering, no-fetch when requester analytics is not permitted, shared date-range propagation, and read-only refresh.

Notes, non-blocking:
- The page route itself remains `requiresAdmin`, so an analytics-only user can hit the backend `/people` endpoint but cannot navigate to this admin dashboard. The PR calls this out honestly; changing that access model would be a separate product decision.
- Local focused Vitest did not run in my temp worktree because it has no installed `node_modules`/Vitest binary. `git diff --check origin/main...HEAD` is clean, and GitHub test (18.x/20.x) plus the other checks are green.

Merge hygiene: PR is BEHIND. Rebase/update to current main and require fresh green before merging.

PR #3536, opened 2026-07-03
feat(approval): delegation self-service + audit visibility (arc C · UX/governance layer)

zensgit:
Arc **C** — the UX/governance layer on the **already-shipped** delegation (委托) runtime. Grounded the shipped surface first (`ApprovalDelegations` resolve-time frozen map · `ApprovalDelegationConfig` CRUD · admin-only `/api/approval-delegations` · `DelegationSettingsView` · `delegatedFrom` metadata trail · self/overlap/chain guards) and **did not rebuild it** — only filled the two genuine gaps.

## Gaps filled
- **Self-service (was 100% admin-only):** participant-gated (`approvals:read`) `/api/approval-delegations/mine` GET/POST/DELETE + `MyDelegationView.vue` at `/my-delegation` (`requiresAuth`, no manage perm). **Security:** create **FORCES `delegator = actor`** (`body.delegatorUserId` is ignored — structural, not validated-away); disable is **403 (not 404)** on another's row via a SELECT-then-check discriminator (`disableOwnDelegation`).
- **Audit visibility:** `listDelegations` gained `includeInactive` (history); the admin view gets a "显示历史" toggle surfacing active+historical + a per-delegator `routedApprovalCount` (`COUNT(DISTINCT instance_id)` from the real `delegatedFrom` trail). Efficiency-review fix: the routed-count aggregate runs **only in audit mode**, so the default operational view keeps its single small query.
- **Boundaries:** self-delegation / range-overlap / delegatee-chains are **already guarded** (service check + table CHECK + one-active unique index + single-hop resolver) — documented, not rebuilt.

## Verification (real-DB `metasheet_cdeleg_test`)
`tsc` 0 · `vue-tsc` 0 · **RED-before proven** (self-service test hit `Cannot POST /api/approval-delegations/mine` 404 before impl) · self-service **5/5** (incl. forced-delegator + 403-not-404) · api **5/5** (history + real-resolver routed count `toBe(1)`) · unit 22 · regression 118 · FE 20. Reviewed by two agents (correctness: zero bugs) + advisor.

## For your confirmation (flagged)
- `/mine` gate = `approvals:read` (participant floor, matches `/pending`); the **self-ownership constraint is the real authorization**. Defensible vs `approvals:act` — your call.
- Routed attribution is **per-delegator**, not per-window (metadata records the delegator id, not which delegation window) — stated in the doc + FE tooltip.

### DIFF
--- a/apps/web/src/approvals/delegations.ts
+++ b/apps/web/src/approvals/delegations.ts
@@ -14,6 +14,9 @@ export interface DelegationRecord {
   startAt: string
   endAt: string
   active: boolean
+  // Audit-only: approvals routed via this delegator (per-delegator, derived server-side).
+  // Present on the admin audit list + the self-service /mine list; absent otherwise.
+  routedApprovalCount?: number
 }
 
 export interface CreateDelegationPayload {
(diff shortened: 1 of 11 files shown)


zensgit:
Reviewed current head `61a76604`.

Verdict: APPROVE on content. The self-service delegation boundaries are implemented in the right place:

- `/api/approval-delegations/mine` is registered before the admin `/:id` routes, so `/mine` cannot be shadowed.
- create-own forces `delegatorUserId` to the authenticated actor; the body value is ignored.
- list-own is scoped to the actor and includes history for the self-service audit view.
- disable-own does a SELECT first, so another user's row is a real 403 rather than a masking 404, and the target row remains active.
- the admin audit list computes `routedApprovalCount` only for audit/history mode, keeping the default active-list path small.

I checked the Gemini notes against the actual schema/code:

- The invalid-UUID concern is refuted: `approval_delegations.id` is `TEXT PRIMARY KEY`, not UUID, so a non-UUID path segment does not trigger Postgres `22P02`; it just behaves like an unknown text id.
- The count query is a real aggregate over `approval_assignments.metadata->>'delegatedFrom'`, but it is limited to audit/self-history surfaces rather than the default operational list. Worth watching, not a merge blocker for this slice.
- Frontend date validation is permissive for malformed non-empty strings: `new Date(...).getTime()` can be `NaN`, and the comparison then does not reject. Backend validation still rejects invalid dates, and the date picker normally emits controlled values, so I do not consider this blocking. A small follow-up test/helper that explicitly rejects `Number.isNaN(start/end)` would make the client match the backend.
- UI disable is one-click. Because it is a soft disable and can be recreated, this is UX hardening rather than a correctness blocker; an `ElMessageBox.confirm` would be a good follow-up.

Verification note: PR checks are green, but the branch is currently BEHIND. My local `git diff --check origin/main...HEAD` was clean. Focused local Vitest did not run in the detached temp worktree because that worktree has no `node_modules`/Vitest binary; I am relying on the GitHub test (18.x/20.x) checks for execution proof.

Merge hygiene: rebase/update to main and require fresh green before merging.

PR #3537, opened 2026-07-03
feat(approval): T3-6 projection base admin-only read guard (arc A)

zensgit:
Arc **A** — closes the leak flagged in the T3-6 closeout: the projection base `base_apr_projection` (materialized approval outcomes) has no per-sheet share, so it fell to the standard capability gate where `deriveCapabilities` grants `canRead` to **any** global `multitable:read` holder. **Owner-default: admin-only** (per-row `visibility_scope` inheritance = a separate slice).

## Guard (every content-read path covered)
Three shared read chokes, each gated (admins bypass, so the lookup never touches the admin hot path):
- **`resolveSheetCapabilities`** — single sheet + record read (record read funnels here via `resolveSheetReadableCapabilities`).
- **`filterReadableSheetRowsForAccess`** — the shared sheet-listing gate (base list + 3 sheet lists); filters projection sheets → **also hides the base itself** (a base surfaces only with ≥1 readable sheet).
- **`resolveReadableSheetIds`** — a second listing path.
- `resolveSheetCapabilitiesForUser` has **no read-path callers** (verified) → no gate.

## Side-effect-free constant
New `approval-projection-constants.ts` holds `APPROVAL_PROJECTION_BASE_ID` + `restrictApprovalProjectionCapabilities`; the projection service **re-exports** the id from here so the permission hot path references it **without** importing the service's `eventBus`/scheduler subscription surface. `loadApprovalProjectionSheetIds` = one lightweight `SELECT … base_id = $2`, non-admin only.

## Verification (fail-first, real-DB)
`approval-projection-visibility.db.test.ts` **6/6**: non-admin (`multitable:read`) **denied** single sheet/record read while reading an ordinary sheet fine; **admin allowed**; listing filters the projection sheet for non-admin (base absent from base list), keeps it for admin; `resolveReadableSheetIds` excludes it. **RED-before confirmed** — neutralizing the lookup → 4 fail (leak returns across all chokes). `tsc` 0 · **T3-6 write/reconcile 14/14** (constant re-export inert to writes) · **full unit 4169/4169** (fixed 3 button-route mock-pool builders to answer the new projection lookup — test-drift, same class as caught earlier this arc).

## Out of scope (explicit)
Per-row `visibility_scope` inheritance (a later slice); the projection write path (unchanged).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3538, opened 2026-07-03
docs(approval): Tier-1 arc A/B/C dev & verification MD (line completion)

zensgit:
The `/goal` deliverable — design + verification for the Tier-1 line-completion arc: **A** projection visibility hardening (#3537), **B** analytics dashboard completion (#3535), **C** delegation self-service + audit (#3536). Per-slice as-built, independent verification, the honest reorientations (B/C found much already shipped → scoped down), and the line-complete statement (remaining = Tier-2/3, each its own future ballot).

### DIFF
--- a/docs/development/approval-line-completion-arc-abc-dev-verification-20260703.md
+++ b/docs/development/approval-line-completion-arc-abc-dev-verification-20260703.md
@@ -8,7 +8,7 @@
 
 | # | Slice | PR | Honest as-built |
 |---|---|---|---|
-| **A** | T3-6 projection base visibility hardening (security prereq) | **#3537** | Full runtime — admin-only read guard at all 3 read chokes, per-path RED-before. |
+| **A** | T3-6 projection base visibility hardening (security prereq) | **#3537** | Full runtime — admin-only read guard at all **4** read chokes (incl. the collab/Yjs/api-token resolver, review P1), per-path RED-before. |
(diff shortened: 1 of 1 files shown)


zensgit:
Reviewed current head `83b915fd`.

Verdict: HOLD — this close-out doc is ahead of current reality.

Two blocking accuracy issues before merge:

1. It states the Tier-1 A/B/C arc is line-complete, but the referenced implementation PRs are not all merged or merge-ready. #3535 and #3536 are still open/BEHIND; #3537 is still open/BEHIND and currently has a P1 review blocker. This document should not land on main as an authoritative completion record until the implementation PRs have actually landed (or the wording is changed to explicitly say "pending PRs / not yet complete").

2. The A section claims #3537 now has the review-P1 collab/Yjs/api-token resolver choke covered and describes the projection read-model as scoped. That is not true for the current #3537 head I reviewed: `restrictApprovalProjectionCapabilities` still does not clear the real write/manage capability fields (`canCreateRecord`, `canEditRecord`, `canDeleteRecord`, `canManageFields`, `canManageViews`, `canManageAutomation`, `canSendNotification`, `canComment`, etc.). So a non-admin with global `multitable:write` can still keep write/manage capabilities while `canRead=false`. Until #3537 fixes that and re-greens, this MD over-claims the security state.

The document shape is useful as a final close-out artifact, and the B/C summaries align with my reviews. But completion docs need to be more conservative than implementation PRs: they should lag verified main, not predict it. Suggested order: fix + merge #3537, merge #3535/#3536 after fresh green, then rebase this doc and update the SHAs / statuses to the actual merged state.

PR #3541, opened 2026-07-03
fix(multitable): schedule config-revision retention sweep (T9 D4 one-knob-ages-both)

zensgit:
## The gap
`startMetaRevisionRetention` wired only `sweepMetaRevisionRetention` (record revisions). `sweepConfigRevisionRetention` existed and was real-DB tested, but the scheduler **never called it** — so `meta_config_revisions` grew unbounded even with retention enabled, contradicting #3168's design intent ("T9 D4 config-revision retention sweep **same policy as records**" / "one knob ages both").

Found during a primary-source re-assessment of the Global History / version-restore line (the line is otherwise essentially complete; this was the one genuine runtime gap — everything else remaining is owner-gated or a doc-staleness fix).

## The fix
`runSweep` now runs **both** sweeps each tick, isolated (one failing never blocks the other), under the same flag/policy/interval, each keeping its own never-delete-latest floor. Inert until `MULTITABLE_META_REVISION_RETENTION_ENABLED=1` (unchanged default-off gate) — so this changes nothing in prod today.

## Verify
Unit test asserts one tick sweeps BOTH `meta_record_revisions` and `meta_config_revisions` (would have failed before the wiring). `8/8` pass in `tests/unit/meta-revision-retention.test.ts`. The sweep functions themselves are unchanged (their real-DB tests still hold).

### DIFF
--- a/packages/core-backend/src/multitable/meta-revision-retention.ts
+++ b/packages/core-backend/src/multitable/meta-revision-retention.ts
@@ -198,7 +198,8 @@ export interface MetaRevisionRetentionSchedulerOptions {
 }
 
 /**
- * Runtime entry: start the periodic retention sweep. **No-op (returns immediately) when retention is
+ * Runtime entry: start the periodic retention sweep of BOTH `meta_record_revisions` and
+ * `meta_config_revisions` (T9 D4 — one knob ages both). **No-op (returns immediately) when retention is
  * disabled** — which is the default — so wiring this into bootstrap changes nothing until the owner
  * sets `MULTITABLE_META_REVISION_RETENTION_ENABLED=1`. Mirrors `startMultitableAttachmentCleanup`
  * (setInterval + returned stop fn; errors are logged, never crash the process). Returns a stop fn.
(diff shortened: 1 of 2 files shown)


zensgit:
Reviewed current head `402699cd0`.

Verdict: APPROVE on content. The scheduler gap is real and the fix is correctly scoped: `startMetaRevisionRetention` now schedules both `sweepMetaRevisionRetention` and `sweepConfigRevisionRetention` under the same flag/policy/interval, while preserving default-off behavior and separate error handling. The added unit test proves one tick reaches both tables, and the existing real-DB config-retention suite already covers the config sweep semantics (`keep-last-n`, `keep-days`, batch bound, never-delete-latest).

Local verification:
- `pnpm --filter @metasheet/core-backend exec vitest run tests/unit/meta-revision-retention.test.ts` → 8/8 pass
- `git diff --check origin/main...HEAD` → clean

Non-blocking P2: the scheduler remains fire-and-forget, so a long sweep can overlap the next interval tick. That risk already existed for record revisions before this PR; this PR extends the same scheduler shape to config revisions, so I don't consider it a merge blocker. A follow-up `isRunning`/`finally` guard plus a long-sweep skip test would make it match the stronger `LedgerRetentionScheduler` pattern.

Merge hygiene: branch is currently BEHIND, so rebase/update onto main and require fresh green before merging.

PR #3307, opened 2026-06-27
docs(multitable): T8-1 undelete-execute (PIT Revert undelete) design-lock (PROPOSED)

zensgit:
## Summary

Design-lock (PROPOSED, docs-only) for **T8-1 undelete-execute** — the next Time Machine write-slice per the agreed order, and the foundation for T9-W Tier 4 (field undelete). Locks scope + per-decision bars **before any runtime code**.

The gap: the PIT/version-restore path refuses to resurrect a deleted record (`univer-meta.ts:8000/8102/8238`). T8-1 wires a *gated resurrect*, reusing the existing outbound-link rebuild + bound-token preview→confirm discipline.

## Locked scope (narrow v1)
- **PIT Revert undelete only** — single-record restore / batch restore / PIT **Reset** undelete each stay a separate explicit decision.
- Resurrect the **original record id**; **id-collision → reject** (never overwrite a live record).
- Preview-identity **binds the resurrect-set + target snapshot/version/hash**; execute re-enumerates → **409** on drift (not the revert hash).
- **Outbound** links rebuilt from the resurrected snapshot.
- **Inbound** links = the core decision (delete drops both directions; existing restore rebuilds outbound only) — **(A) lazy/repair-on-read (recommended) vs (B) eager re-materialize**, **gated on #3306** (cross-base dangling-link repair, OPEN) which moves the dangling boundary.
- Permission + default-off flag + typed confirm + single-tx atomicity + audit/revision/realtime → test matrix.

## Runtime gate (no code until all hold)
1. This lock ratified — especially the **L4 inbound** decision.
2. **#3306 re-checked / merged** (inbound/dangling boundary).
3. No live parallel undelete session.

## Open questions for the owner
1. L4 inbound — (A) lazy/repair-on-read vs (B) eager re-materialize?
2. L5 — exact cap (`canDeleteRecord` vs `canEditRecord`) + flag name.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3542, opened 2026-07-03
docs(multitable): global-history verified state map + owner-decision menu (2026-07-03)

zensgit:
Answers **"进度如何 / 同设计MD规划还差哪些"** with a consolidated, primary-source re-assessment of the Global History / version-restore line against `origin/main` (5-cluster parallel audit).

## Headline: 基本已完成
Every substantive capability is **built** — record-history + Layer-1 restore (7 error codes, per-field restore, FE), PIT view/revert/reset/undelete (destructive paths behind default-off flags), config-history + all 5 T9-W tiers, T5 preview / T6 scoped / diff-unify / batch-scope identity, the Global History Center UI, and flag-enablement readiness (staging smoke + operator checklist + harnesses).

## What this doc adds
1. **Verified capability map** per cluster (built / gated-off / missing, with evidence).
2. **Reconciles stale doc statuses** — several per-slice headers still say "docs-only / NOT built" though the runtime shipped (the exact stale-draft trap); corrected here + inline in `remaining-dev-plan-20260625` §4a (undelete-execute is BUILT #3307).
3. **The one runtime gap** found + fixed: config-revision retention wiring → **PR #3541** (companion).
4. **Owner-decision menu** for the remainder: per-flag enablements, product entries (history-anchored T-source, FE for destructive tiers), irreversible-semantics sign-offs (lossy retype, tombstone-capture, record undelete Slice 2b), and one **honestly-impossible** item (value-level undelete of already-deleted field columns).

Two assessor claims were **overturned by direct verification** (legacy permission route already locked #3414; config-revision pruning confirmed in-scope #3168) — the map is grounded, not drafted. Docs-only.

### DIFF
--- a/docs/development/multitable-global-history-remaining-dev-plan-20260625.md
+++ b/docs/development/multitable-global-history-remaining-dev-plan-20260625.md
@@ -1,5 +1,13 @@
 # Global History / point-in-time-restore line — remaining-dev plan, design & verification
 
+> **STATUS UPDATE (2026-07-03) — §4a partly OVERTAKEN BY EVENTS; see the current state map.**
+> Since this 2026-06-25 plan, **PIT undelete-execute (§4a) was BUILT** (#3307, behind default-off
+> `MULTITABLE_ENABLE_PIT_UNDELETE`) — only its *enablement* is still gated, not the code. §4b (T9-W lossy
+> value-transform / field-undelete data recovery) remains genuinely un-built + gated (and value-level field
+> undelete is *impossible* under the current hard-delete cascade — recorded as such). The authoritative
+> current-reality view is **`multitable-global-history-verified-state-map-and-decision-menu-20260703.md`**;
+> read that for "还差哪些". This plan is preserved as the historical sign-off vehicle.
+
 Grounding: `origin/main` 2026-06-25, after the non-destructive line + the debrand/runbook closeout. This doc answers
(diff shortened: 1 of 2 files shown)


zensgit:
Reviewed current head `be28e33f`.

Verdict: HOLD / needs doc fixes before merge. The direction is useful: a current Global History state map is exactly the right artifact, and most of the built-vs-gated reconciliation matches the code and runbooks. Two accuracy issues need fixing because this doc presents itself as the authoritative current-reality view:

1. #3541 is still OPEN/BEHIND, but §3 and the headline describe the config-revision retention scheduler gap as already "fixed". Either merge #3541 first and then rebase this PR, or change the wording to "tracked by PR #3541 / pending merge". Otherwise main would contain a status map claiming a runtime fix that is not actually on main.

2. §4a gives `PIT_RESET` a stop condition of "keep `META_REVISION_RETENTION` off (or set trash retention >= approval window)". The established runbook says the opposite nuance: `meta_records_trash` has no purge and is not the stop-ship; the real condition is `MULTITABLE_META_REVISION_RETENTION_ENABLED` off while `PIT_RESET` is on, unless the owner explicitly accepts that revert-undo becomes time-windowed. Trash retention is not a valid substitute for revision-retention safety.

Checks are green but stale; PR is also BEHIND. After those wording fixes, rebase to main and run fresh CI.

PR #3562, opened 2026-07-04
fix(integration): stock-prep S1 wizard residual hardening (#3561 follow-up)

zensgit:
Completes the non-blocking residual items from the #3561 independent review (both review channels; deliverable at `/tmp/pr3561-review-claude-20260704.md`). #3561 merged before the residuals landed, so they ship as this follow-up. Frontend-only, zero backend change, no wire change.

## What
1. **Readiness scope type narrowed** (`workbench.ts`): `getIntegrationStockPreparationTargetReadiness` now takes plain `IntegrationScope` — the signature **structurally forbids** `baseId` on readiness (server handler only consumes `baseId` on `ensure`; readiness inspects the already-bound canonical target). No wire change: the caller already sent `tenantId`/`workspaceId` only, and the existing test lock (readiness URL never carries `baseId`) stays.
2. **In-flight input lock** (`IntegrationWorkbenchView.vue`): the baseId input is `:disabled` while a request runs — closes the operator-confusion window where editing baseId mid-flight silently dropped a completed `ensure` receipt (server had created/bound; FE showed nothing). The staleness signature keeps covering `baseId` as the **belt** behind the disabled input; a mid-flight `disabled === true` assertion + the existing programmatic-edit drop test lock both layers.
3. **Leak lock widened** (spec): private-binding non-render assertions now cover the **whole** `s1-result` panel textContent, not just the evidence `<pre>` (the mock's `targetBinding` carries realistic private `sheetId`/`fieldIdMap` values; none may render anywhere in the panel).
4. **Deny-path test** (spec): an `integration:write` (non-admin) user sees no S1 panel and no `/stock-preparation/target/` wire.

## Recorded-as-is (deliberate, no code)
- Signature conservatism on readiness (covers baseId though the request doesn't) — intentional belt.
- `IntegrationStockPreparationTargetScope.projectId` stays for `ensure` forward-compat (server consumes it).

## Verification
- `vitest run tests/IntegrationWorkbenchView.spec.ts` → **48/48** (47 existing + 1 new deny-path)
- `vue-tsc -b` → clean

### DIFF
--- a/apps/web/src/services/integration/workbench.ts
+++ b/apps/web/src/services/integration/workbench.ts
@@ -1352,14 +1352,15 @@ export async function syncIntegrationStockPreparationOptions(
   return parseIntegrationResponse<IntegrationStockPreparationOptionSyncResult>(response)
 }
 
+// Readiness inspects the already-bound canonical target; the server only consumes baseId on
+// ensure (create/bind). The plain IntegrationScope parameter structurally forbids callers from
+// passing a baseId here, so the request always mirrors what the server evaluates.
 export async function getIntegrationStockPreparationTargetReadiness(
-  scope: IntegrationStockPreparationTargetScope = {},
+  scope: IntegrationScope = {},
 ): Promise<IntegrationStockPreparationTargetReadinessResult> {
   const query = buildQueryString({
     tenantId: scope.tenantId,
     workspaceId: scope.workspaceId,
-    projectId: scope.projectId,
-    baseId: scope.baseId,
   })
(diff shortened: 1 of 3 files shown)


zensgit:
Review verdict: PASS — no blocking findings.\n\nReviewed live head 2f8a11a3f against the S1 residual hardening goals: readiness no longer serializes baseId/projectId, baseId is disabled during in-flight target ops with the signature belt still catching programmatic changes, leak assertions now cover the whole result panel, and the non-admin deny path has a sibling mount anchor plus no stock-prep target wire. CI is green.\n\nNote: GitHub would not let me submit a formal APPROVE review on this PR because it is treated as my own PR; this comment records the review verdict instead.

PR #3553, opened 2026-07-04
feat(integration): read-source composition C-R1 config model (#1709)

zensgit:
## What

Implements the authorized C-R1 slice for #1709 read-source resolver composition: chain config model + validator + content-keyed versioned store.

Scope is intentionally narrow:

- two-hop ordered approved-read composition config (`material -> resolver output -> second resolver key` shape);
- save-time validator for read-only operations, bounded two-step chain, previous-step handoff, approved `resolver_lookup` read-source refs, and single resolver output target;
- content-keyed versioned store + draft/approved/retired lifecycle + values-free audit;
- migration `063_create_integration_read_source_composition_configs.sql` for composition config/audit tables;
- plugin CI wiring for the new tests.

## Merge order

This PR owns migration `063_*` and should merge before #3548, which intentionally uses migration `064_*` for W1 write-target configs.

Do not merge #3548 ahead of this PR unless #3548 is first renumbered.

## Boundary

No runtime execution, no route, no chaining executor, no adapter/network call, no resolver composition runtime, no recursive BOM expansion, no Save/Submit/Audit, no external/production write.

Referenced read configs are checked through `readSourceConfigStore.getForRuntime()` at save time, so C-R1 only accepts approved read-source configs. Runtime re-check remains a later C-R3 responsibility.

## Verification

- `node --check` on new C-R1 modules/tests
- `pnpm --filter plugin-integration-core run test:read-source-composition-config`
- `pnpm --filter plugin-integration-core run test:read-source-composition-config-store`
- `pnpm --filter plugin-integration-core run test:read-source-composition-config-migration`
- `pnpm --filter plugin-integration-core test`
- `git diff --check`
- Fresh GitHub CI green, including test (18.x) and test (20.x)

## Notes

This is Draft for owner review. It does not authorize or implement C-R2/C-R3.

PR #3563, opened 2026-07-04
feat(integration): read-source composition C-R2 pure chain planner/evaluator

zensgit:
## Summary

C-R2 rung of the composition ladder (#1709, `docs/development/integration-read-source-resolver-composition-design-lock-20260703.md`): a **PURE chain planner/evaluator**. Builds directly on merged C-R1 (#3553 config model) and the R0–R2 standalone resolver.

- `planReadSourceComposition` — reuses the **C-R1 save-time validator verbatim** (approved-only, resolver_lookup-only, write-shaped-step rejected) and shapes an identifier-only ordered plan (never endpoints/filters/credentials/values).
- `deriveCompositionStepInput` — the lock-2 **typed single-scalar handoff**: prior hop's resolver output target must match the config-declared wiring; the value is validated by the **same `normalizeReadSourceProbeInputs` predicate the runtime enforces** (imported, not re-stated — no mirror drift). Failure results carry only `{ ok, errorCode }`.
- `evaluateCompositionOutcome` — lock-5 **values-free stitched evidence**: `{step, ok, rule?, errorCode?}` vector only; per-hop + chain-level fail-closed (lock 4), downstream hops marked `STEP_NOT_RUN`; chain data **re-projects only the last hop's single resolver output** (the intermediate FItemID never exposed).

## Boundary (scope fence)

```text
routeAdded=false
outboundCallExecuted=false
adapterCreated=false
persistenceTouched=false
writePathTouched=false
recursionSupported=false
clientMirrorTouched=false   (no UI/route surface — client vocab mirror obligation starts at C-R3)
```

C-R3 (chain runtime executor) and recursion remain separate gated opt-ins.

## Verification

- `node __tests__/read-source-composition-planner.test.cjs` — OK (13 test groups: purity require-allowlist tripwire, plan approved-only/write-rejection, 10 derive fail-closed cases incl. runtime key bounds 128/control-char/non-scalar, hop-failure abort, incomplete-chain fail-closed, malformed-success-data fail-closed, foreign-token sanitization, step-vector key discipline, 7-code vocabulary)
- Full related suite locally green: read-source-{config, probe-contract, probe-runtime, read-runtime, resolver-evaluator, composition-config} + planner
- Wired into the plugin `test` chain in package.json


zensgit:
**Adversarial-verify round (ultracode, 3 lenses × refute-verification, 9 agents) — 5 confirmed findings, all fixed in `bdc73e9ab`:**

| # | Sev | Finding | Fix |
|---|---|---|---|
| 1 | P2 | Accessor-bearing hop data could **throw** out of `evaluateCompositionOutcome`; TOCTOU getter could pass validation with a clean scalar and smuggle an unvalidated object into chain data via a second read | `snapshotResolverData` — every hop field read **once**; validation + projection share the read; all exports get a fail-closed catch (exception messages never escape) |
| 2 | P2 | No red test locked L5 chain-data stripping — a `{...spread}` regression shipped green | deepEqual shape pin + sentinel scans (host / candidateValues / rawRows) |
| 3 | P2 | `isResolverData` narrower than R1's real emission set (R1 can resolve boolean/object standalone) → self-contradictory stitched evidence | chain-only scalar narrowing made **explicit**: dedicated `STEP_OUTPUT_NOT_SCALAR` code (vocab 7→8), header note; R1/R2 standalone untouched |
| 4 | NIT | Non-scalar prior value mapped to `VALUE_MISSING` (the 'never resolved' code); tests hedged | vocabulary pinned: MISSING=never resolved / MISMATCH=wrong target / INVALID=resolved-not-key-usable |
| 5 | NIT | A success outcome supplied after a failed ordinal stitched as `ok:true` — evidence could certify an L4 violation | post-failure entries clamped `STEP_NOT_RUN` (void); vector can no longer represent 'ran after abort' |

Plus the earlier review round: P1 zero-step crash (bounds + handoff arity in `isPlanShaped`) and P2 registered-set errorCode guard — both in `d1b05b479` with red tests.

PR #3565, opened 2026-07-04
feat(integration): read-source composition C-R3 chain runtime executor

zensgit:
## Summary

C-R3 rung of the composition ladder (#1709, `docs/development/integration-read-source-resolver-composition-design-lock-20260703.md`): the **chain runtime executor**. Builds on merged C-R1 (#3553 config model) and C-R2 (#3563 planner). Orchestrates an approved two-hop composition end-to-end, **read-only**, reusing the shipped pieces verbatim — no forked read/resolver path.

- **C-R2 planner** for ordering / typed single-scalar handoff / values-free chain evidence + last-hop-only projection
- **S3-1 `executeConfiguredRead`** for each hop's bounded/capped/SSRF-reguarded outbound read
- **R1 resolver** (inside `executeConfiguredRead`) for per-hop multiplicity selection

## Locks

- **L1 approved-only** — builds the planner `readConfigsById` from the resolved per-step bundle so the planner re-validates `status==='approved'` / resolver_lookup / read-only / handoff-target as **defense-in-depth over the route load**; a draft/missing/write-shaped/mis-wired step aborts at plan time (values-free `PLAN_INVALID` + the C-R1 validator's values-free triples), **no hop runs**.
- **L2 typed single-scalar handoff** — `deriveCompositionStepInput` is the single gate between hops; the intermediate key is **derived by the platform**, never runtime-supplied.
- **L3 key-only** — strict `{ inputs: { key } }` allowlist at the chain boundary (config override / per-hop key / extra field **rejected**, not dropped); the key value goes through the same bounded-key predicate every read uses.
- **L4 chain abort** — a failing derive aborts; the missing tail is stitched `STEP_NOT_RUN`; a failed/ambiguous/non-scalar hop never runs downstream.
- **L5 values-free** — intermediate resolved value never exposed; chain data = **only** the last hop's single resolver output.
- **L6 read-only** — every hop is `executeConfiguredRead`; adapter write surface never touched (test asserts `writes===0`).

## Boundary (scope fence — same as read-source-read-runtime)

```text
routeAdded=false
persistenceLookupInside=false   (route passes loaded composition config + resolved per-step bundle in)
writePathTouched=false
recursionSupported=false
clientMirrorTouched=false       (route + config-store load + client vocab mirror + UI = a LATER gated rung)
newCredentialPath=false
```

The route wiring (+ config-store approved-only load, client vocab mirror, UI) and recursion remain separate gated opt-ins.

## Verification

- `node __tests__/read-source-composition-runtime.test.cjs` — OK (10 groups): happy-path two-hop (derived intermediate key on hop-1 filter, last-hop-only data, values-free evidence, `writes===0`); key-only request rejection (6 bad shapes + long-key downstream fail-closed); plan gate (draft/missing/malformed → PLAN_INVALID, **no hop runs**); hop-0 ambiguate/not-found abort (hop 1 never reads); hop-1 failure after hop-0 resolve; adapter-throw fail-closed (error message never leaks); kind-mismatch fail-closed; createAdapter required; proto-pollution-safe bundle lookup
- Full related suite locally green (config, probe-contract, probe-runtime, read-runtime, resolver-evaluator, composition-config, composition-planner, composition-runtime)
- Wired into the plugin `test` chain in package.json


zensgit:
**Adversarial-verify pass (ultracode, 3 lenses × refute-verification) + 1 hardening in `71d89ea75` (rebased to `3727bde21`):**

The workflow's own verifiers refuted all 3 surfaced items as *caller-contract* rather than live defects. I independently reproduced the L1 one and hardened anyway — it makes the module's **own stated defense-in-depth invariant** actually hold, matches the hardening class accepted in the C-R2 round, and gives the not-yet-built C-R4 route a **fail-closed** (not fail-open) contract.

| # | Item | Disposition |
|---|---|---|
| L1 | A per-step bundle **lacking a `status` property** slipped through as approved and ran the full chain (reproduced: `ok=true, reads=2`) — the C-R1 validator's status check is optional at save-time, so `undefined` was skipped | **Fixed** (1 ternary): runtime coerces any non-`'approved'` status → explicit `'not_approved'`; statusless bundle now fails closed (PLAN_INVALID / STEP_NOT_APPROVED, 0 reads). Safe direction — if C-R4 forgets `status`, the chain fails closed loudly, never silently runs an unapproved step |
| P2 | No end-to-end coverage of the chain-only scalar-narrowing abort | **Red test added**: hop 0 resolves a K3 boolean → `STEP_OUTPUT_NOT_SCALAR`, hop 1 never reads, `writes===0`, values-free |
| NIT | L3 named request-shape vectors untested | **Red tests added**: JSON `__proto__` own-key rejected at body+inputs, getter-on-key read exactly once (TOCTOU-safe), object/array key fails closed at hop 0 with no outbound read |

All test-only except the one-ternary status coercion. Local suite green; rebased onto latest main.

PR #3568, opened 2026-07-04
feat(integration): fail-close the C-R2 planner seam (approval-context + evidence-agreement)

zensgit:
## What

Two defense-in-depth tightenings on the merged C-R2 chain planner (`read-source-composition-planner.cjs`), as a follow-up to #3563, so the C-R3 chain runtime executor (#3565) builds on a seam that fails closed **by construction**.

Both close "untested guard" items surfaced by an independent review of #3563. That review ran against an early PR head that predated #3563's own pre-merge hardening — the largest item (F1, last-hop re-projection leak) was already closed on merge by `testChainDataStripsExtraFields`, so only these two P3 items remain. Both are **unreachable via the real R1 producer** (which nulls `data` on failure) and **already satisfied by the #3565 runtime**, but the executor seam should not depend on a correct caller.

### 1. Approval is a plan-time gate — fail closed without an approval map
The C-R1 validator only vets `approved / resolver_lookup / read-only` **when `options.readConfigsById` is supplied** and returns early otherwise, so `planReadSourceComposition` could emit an approval-**unverified** `ok:true` plan when a caller omits the map. The planner is the C-R3 executor seam, so it now returns `PLAN_INVALID` when the approved-config map is absent (`undefined`/`null`). Omission is a contract violation, not a preview.

### 2. A hop is a success only when evidence AND data agree
`hopOk` keyed on the data snapshot alone, ignoring `evidence.ok`. A contradictory hop (`evidence.ok === false` but resolver-shaped scalar data) was certified as chain success — a **fail-OPEN** that would expose a stale/aborted value. Now a hop succeeds only when it **both** claims success (`evidence.ok === true`) **and** carries a chain-usable scalar; a hop that claimed failure keeps its own coarse code. The mirror direction (ok evidence + non-resolver data = failed) was already enforced; this closes the other direction.

## Compatibility with #3565 (C-R3 runtime, open)
`read-source-composition-runtime.cjs` always builds `readConfigsById = {}` and passes `{ readConfigsById }` (its own comment asks the planner to "re-validate approved-only / resolver-only / read-only ... as defense-in-depth"), and its success hops carry `evidence.ok:true`. So **#3565 satisfies both guards** and only needs a trivial rebase; re-verify it green after rebasing. If any #3565 fixture omits `evidence.ok` on a success hop, guard 2 will (correctly) surface that drift.

## Verification
- Both guards **mutation-verified**: neutering each reds its new test (`testPlanRequiresApprovalContext`, `testEvaluateContradictoryHopFailsClosed`).
- `node __tests__/read-source-composition-planner.test.cjs` OK + neighbors (config / resolver-evaluator / probe-runtime) green.
- The planner has **no other consumer on main** (only its own test) — the change is isolated; no `package.json` test-chain edit (tests added to an already-chained file), so no merge-order hazard.

Scope fence unchanged: pure lib, no route/adapter/persistence/write. Independent review MD at `/tmp/pr3563-review-claude-20260705.md`.


zensgit:
**Cross-verification vs the now-merged C-R3 runtime (#3565, `e68fe39ce`):**

I rebased this branch onto `origin/main` (which now contains C-R3) in a throwaway worktree and ran the full composition suite together — **both guards are sound and compatible with C-R3**:

- `read-source-composition-planner.test.cjs` — OK
- `read-source-composition-runtime.test.cjs` — **OK** (the C-R3 executor)
- neighbors (config / resolver-evaluator / read-runtime / probe-runtime) — OK

Specifics that matter for the interaction:
- **Guard 2 (evidence.ok agreement)** — C-R3's success hops come from the real `executeConfiguredRead → evaluateResolver`, which sets `evidence.ok:true` on success, so they satisfy `evidence.ok===true && scalar`. My `testNonScalarHandoffAborts` (hop 0 resolves a K3 boolean) still yields `STEP_OUTPUT_NOT_SCALAR` under the new branch, because R1 emits `evidence.ok:true` for a resolved-but-non-scalar value → guard 2's error-code branch (`evidence.ok===true && non-scalar`) fires exactly as before.
- **Guard 1 (approval-context required)** — C-R3 always builds `readConfigsById` and passes `{ readConfigsById }` (never undefined/null), so it satisfies the new mandatory-map gate; and it *populates* the map from the resolved per-step bundle, so a real chain still validates approved-only per step.

Net: this hardening does not regress C-R3; it's clean to land after the rebase (0 conflicts — C-R3 touched only `read-source-composition-runtime.cjs`, not the planner). Not touching this branch since it's under an active worktree — flagging the green cross-check for whoever drives the merge.

PR #3573, opened 2026-07-04
feat(integration): read-source composition C-R4-1 HTTP surface (authoring + approved-only key-only run route)

zensgit:
## Summary

**C-R4-1** (opt-in authorized 2026-07-05) of the composition line (#1709): the **HTTP surface** over the merged C-R1 config store (#3553) + C-R3 runtime executor (#3565). **Server-only** — no client code, no UI. C-R4-2 (client vocab mirror + tripwire) and C-R4-3 (UI) are staged after this.

- **Store registration** (index.cjs): registers the composition config store as a plugin service. It validates each step's read config is approved **at save time** (via `readSourceConfigStore.getForRuntime`) and the run route re-loads them at runtime.
- **Authoring routes** (mirror the read-source-config routes): `save` (write), `list`/`get`/`audit` (read), `approve`/`retire` (write) + a values-free composition-config error mapper.
- **Run route** `POST /api/integration/read-source-compositions/:id/run` — the runtime tier:
  - **read-tier** + strict `{ inputs }` body allowlist (config override / extra field rejected **before any store access**; per-hop key rejected by the executor's own `{ inputs: { key } }` normalizer);
  - **approved-only DOUBLE gate** — the composition config **and** each referenced step read config are loaded via their stores' `getForRuntime` (throws `NOT_APPROVED` for a non-approved version), and the C-R2 planner re-validates the assembled bundle inside the executor as defense-in-depth;
  - intermediate keys **derived by the platform** (never runtime-supplied);
  - **values-free** evidence, chain data = last hop's single output only, **read-only** (no adapter write surface).

## Boundary

```text
clientCodeTouched=false     (C-R4-2 = vocab mirror + tripwire; C-R4-3 = UI)
uiTouched=false
writePathTouched=false
recursionSupported=false
newRuntimeSuppliedInput=false   (only the first business key; intermediate keys derived)
approvedOnlyDoubleGate=true
```

## Verification

- `node __tests__/http-routes.test.cjs` — `testReadSourceCompositionRoutes` **OK** + full suite green: happy-path two-hop (derived intermediate key on hop-1 filter, last-hop-only data, values-free evidence, no write), approved-only gate #1 (composition draft → 409, **no hop runs**), approved-only gate #2 (step config draft → 409, **no hop runs**), key-only allowlist (config-override / extra field → 400 before store access; per-hop key → 400), hop-0 ambiguate abort (hop 1 never reads, 200 values-free), authoring tier enforcement (run=read, save=write), authoring lifecycle.
- `node --check` clean on index.cjs + http-routes.cjs; composition modules + http-routes load OK.
- Neighbors green (composition-config-store / composition-runtime / composition-planner / read-source-config-store).

> The composition vocab **client/server mirror + CI tripwire is C-R4-2** (landing before the C-R4-3 UI consumes these codes); no client code is touched here, so no existing mirror is affected.

### DIFF
--- a/plugins/plugin-integration-core/__tests__/http-routes.test.cjs
+++ b/plugins/plugin-integration-core/__tests__/http-routes.test.cjs
@@ -6586,7 +6586,7 @@ async function testReadSourceCompositionRoutes() {
     ],
   }
 
-  function compositionServices({ itemRows, bomRows, compositionStatus = 'approved', bomConfigStatus = 'approved' } = {}) {
+  function compositionServices({ itemRows, bomRows, compositionStatus = 'approved', bomConfigStatus = 'approved', bomConfigMode = 'resolver_lookup' } = {}) {
     const reads = []
     const writes = []
     const services = createMockServices({
@@ -6618,7 +6618,9 @@ async function testReadSourceCompositionRoutes() {
             error.details = { id: input.id, status: bomConfigStatus }
             throw error
           }
-          const config = input.id === 'cfg-bom' ? BOM_CONFIG : ITEM_CONFIG
+          // bomConfigMode lets a test return an APPROVED (getForRuntime passes) but WRONG-SHAPE step
+          // config, so the C-R2 planner's defense-in-depth re-validation is exercised at the route level.
+          const config = input.id === 'cfg-bom' ? { ...BOM_CONFIG, mode: bomConfigMode } : ITEM_CONFIG
           return { id: input.id, systemId: 'sys_1', status: 'approved', config }
         },
       },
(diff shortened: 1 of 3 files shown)


zensgit:
**Adversarial-verify pass (ultracode, 3 lenses × refute-verification) — `confirmed: []`, no security defects.** The 3 lenses (approved-only + gate order / key-only + injection / values-free + S3-2 parity) found no reproducible bypass, leak, or fail-open.

One **NIT coverage gap** was surfaced and closed in `46e899c0d`: the run-route tests exercised gate #2 with a *non-approved* step config (getForRuntime throws before the planner runs), but not the design-lock's explicit *'planner re-validates the bundle'* layer. Added **gate #2b** — an *approved-but-wrong-shape* step config (mode≠resolver_lookup, so getForRuntime passes but the C-R2 planner rejects it) → values-free PLAN_INVALID, no hop runs. Locks the defense-in-depth that would otherwise regress silently.

Note: the earlier `test (20.x)` red was an unrelated attendance-integration flake (`attendance-plugin.test.ts` idempotent-retry, expected 2 got 0) — zero overlap with this diff (composition routes only), and `test (18.x)` passed the same suite; re-run triggered.

PR #3574, opened 2026-07-04
test(integration): #3365 OAPI allowlist⟺guard dual-reachability tripwire

zensgit:
Tier-B gap #5 from `docs/research/multitable-feishu-refresh-audit-20260629.md` line 58 (permission golden matrix): *"pin that the OAPI allowlist set equals the set of `apiTokenAuth`-guarded routes (currently lockstep-by-hand)."* **Test-only, zero `src` changes.**

## What it does

Two independently-edited call sites currently have to move together by hand: the allowlist regex/route arrays in `oapi-read-allowlist.ts`, and the `router.<verb>(...)` registrations in `univer-meta.ts` / `comments.ts` (the only two files that ever mount `apiTokenAuth` — confirmed by whole-`src` grep). Nothing forces them to stay in sync, so they can drift in either direction:

- a route gains `apiTokenAuth` + `requireScope` but the allowlist entry is never added → dead guard (harmless, but silently breaks whatever was meant to be exposed)
- an allowlist entry is widened/added without a route mounting both guards → **silent no-auth bypass** (`requireScope` fail-opens when no token scopes are attached)

This test source-scans every route registration in both files, splits them into GUARDED vs UNGUARDED by whether `apiTokenAuth` + `requireScope(` appear on the same registration line, and asserts both directions against the real exported `isOapiAllowlistRequest` — no hand-duplicated list of its own. Style follows the `multitable-stored-data-taint-chokepoint.guard.test.ts` precedent (whole-file structural scan, not a few hand-picked cases).

## Proof of non-vacuousness (deliberate mutate → RED → revert → GREEN)

- Stripped the guard from `GET /fields` → **test 3 went RED**, naming the exact leaked route (`routes/univer-meta.ts: GET /api/multitable/fields`) — proves the bypass direction is actually caught.
- Added a new guarded-but-unallowlisted probe route → **test 2 went RED**, naming the exact dead-guard route — proves the dead-guard direction is actually caught.
- Reverted both mutations; `diff` confirmed byte-identical to `origin/main` after each.

## Verification

- New test: **3/3 pass** today (the invariant currently holds — PR #3365 wired it correctly).
- Adjacent suite unaffected: existing hand-written `multitable-oapi-read-allowlist.test.ts` (14/14) + the `taint-chokepoint` precedent guard (3/3) still green alongside it — **20/20 total**.
- `git status`/`git diff --stat` after commit: only the one new test file, zero `src` touched.

Awaiting review — not merging.

### DIFF
--- a/packages/core-backend/tests/unit/multitable-oapi-allowlist-guard-tripwire.test.ts
+++ b/packages/core-backend/tests/unit/multitable-oapi-allowlist-guard-tripwire.test.ts
@@ -21,22 +21,24 @@
  * are the same set, following the same whole-file structural-scan discipline as
  * `multitable-stored-data-taint-chokepoint.guard.test.ts`:
  *
- *   1. Scan every `router.<verb>('<path>', ...)` registration in `routes/univer-meta.ts` and
- *      `routes/comments.ts` — confirmed (via `rg -ln apiTokenAuth packages/core-backend/src`) to be the
- *      ONLY two files where `apiTokenAuth` is ever mounted on a route — and split them into GUARDED
+ *   1. First scan every TypeScript file under `src/routes` to prove the explicit `ROUTE_FILES` scope
+ *      covers every route file that references `apiTokenAuth`.
+ *   2. Scan every `router.<verb>('<path>', ...)` registration in `routes/univer-meta.ts` and
+ *      `routes/comments.ts` — currently the ONLY two files where `apiTokenAuth` is mounted — and split
+ *      them into GUARDED
  *      (the registration's own argument list contains both `apiTokenAuth` and `requireScope(`) and
  *      UNGUARDED (every other registered route in the same two files).
- *   2. For each GUARDED route, assert the exported `isOapiAllowlistRequest` (the real gate function,
+ *   3. For each GUARDED route, assert the exported `isOapiAllowlistRequest` (the real gate function,
  *      not a re-typed copy) admits it.
- *   3. For each UNGUARDED route, assert `isOapiAllowlistRequest` REJECTS it — this is what actually
+ *   4. For each UNGUARDED route, assert `isOapiAllowlistRequest` REJECTS it — this is what actually
  *      proves the allowlist isn't over-broad: it's checked against every OTHER real route Express would
  *      dispatch on these two routers, not just a few hand-picked "adjacent path" guesses.
  *
  * If this test fails, you added/changed a route's guard-mounting or the allowlist's shape without
  * updating the other — see the two failure messages below for which direction drifted and what to fix.
  */
-import { readFileSync } from 'node:fs'
-import { join } from 'node:path'
+import { readdirSync, readFileSync } from 'node:fs'
+import { join, relative, sep } from 'node:path'
 
 import { describe, expect, test } from 'vitest'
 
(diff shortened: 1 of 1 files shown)


zensgit:
# PR #3574 review — OAPI allowlist/guard dual-reachability tripwire — 2026-07-05

Verdict: one P2 should be fixed before merge.

What looks good:
- Scope is test-only: one new unit test file, no runtime/src changes.
- The core invariant is the right one: every route with `apiTokenAuth + requireScope` must be admitted by `isOapiAllowlistRequest`, and every other scanned route must be rejected.
- The test uses the real exported `isOapiAllowlistRequest`, not a retyped allowlist.
- It checks both drift directions and the PR body records fail-first mutations for dead-guard and overmatch/bypass cases.
- Current CI is green (Strict E2E skip aside), and the adjacent OAPI route state on `origin/main` matches the test's current expected population.

Finding:

P2 — The `ROUTE_FILES` scope is a load-bearing premise but is not itself test-locked.

The new test scans only:

```ts
const ROUTE_FILES = ['routes/univer-meta.ts', 'routes/comments.ts'] as const
```

The file comments and PR body say those are the only route files that mount `apiTokenAuth`, and that is true on current `origin/main` by grep. But the test does not enforce it. If a future route file adds `router.get(..., apiTokenAuth, requireScope(...))`, this tripwire will not include it in either `GUARDED` or `UNGUARDED`, so the dead-guard direction for that new route is invisible. If a future allowlist pattern is widened to reach a route outside these two scanned files, that route is likewise invisible to the "unguarded route must reject" half.

Because the whole purpose of this PR is to convert "kept in lockstep by hand" into a tripwire, the "only these two files" assumption should be locked in the test, not left as a one-time grep fact.

Suggested minimal fix:

- Scan `packages/core-backend/src/routes/**/*.ts` (or the route files under `SRC/routes`) for actual route registrations whose same registration line contains `apiTokenAuth`, e.g. `router.<verb>(..., apiTokenAuth`.
- Assert every file containing such a route registration is in `ROUTE_FILES`.
- Keep the existing `ROUTE_FILES` scanner for the richer guarded/unguarded parity check.

That makes a future third route file fail loudly with an instruction to either add it to the tripwire scope or avoid mounting token auth there.

No other blockers found.

zensgit:
P2 fix pushed in `600e47e6d` (`test(multitable): lock OAPI token route file scope`).

What changed:
- Added a source-scope tripwire that scans every TypeScript file under `packages/core-backend/src/routes`.
- Any route source file referencing `apiTokenAuth` must now be listed in `ROUTE_FILES`.
- Any stale `ROUTE_FILES` entry that no longer references `apiTokenAuth` also fails.

That closes the gap from my review: a future third route file cannot mount `apiTokenAuth` and remain invisible to both the guarded and unguarded allowlist parity checks.

Local verification:
- `pnpm --filter @metasheet/core-backend exec vitest run tests/unit/multitable-oapi-allowlist-guard-tripwire.test.ts` → 4/4 pass
- `pnpm --filter @metasheet/core-backend type-check` → pass
- `git diff --check` → pass

CI should rerun on the updated head.

zensgit:
Codex review verdict: PASS. The allowlist<=>guard tripwire closes the #3365 drift risk in the right direction: it derives guarded and unguarded route populations from source, verifies every guarded route is allowlisted, and verifies adjacent unguarded routes are rejected. Checks are green; merge mechanics only.

Note: GitHub would not allow this account to submit a formal APPROVE because it is the PR author, so recording the verdict as a PR comment.

PR #3584, opened 2026-07-04
feat(multitable): AI-fields S1 runtime — write provenance + commit-action batch grouping

zensgit:
Implements the ratified, merged design-lock: `docs/development/multitable-ai-write-provenance-batch-grouping-s1-designlock-20260705.md` (design-lock PR #3569, squash `6b0e27bf2`). **Awaiting human review — not merging.**

## LOCK-A — source attribution
- `'ai-shortcut'` added to `RecordWriteSource` (`post-commit-hooks.ts`). `RecordRevisionSource` needed no change — already an open `| string` union — verified the value flows end-to-end through the sole `recordRecordRevision` call site inside `patchRecords` (an update-only function; one call site).
- All three AI write call sites (inline shortcut-run + the shared `commitOneRecord` helper used by both bulk-commit and job-commit) tag `source: 'ai-shortcut'`.
- `HistoryCenterModal`'s inline `sourceLabel` map gets an `'ai-shortcut'` entry — NOT extracted to a shared module, per the lock's explicit A5 instruction.

## LOCK-B — commit-action batch grouping
- Optional `RecordPatchInput.batchId?: string` seam on `patchRecords`; default behavior is **byte-identical** for every existing caller (falls back to the same `randomUUID()` as before this seam existed).
- bulk-commit and job-commit **ROUTES** (not `commitOneRecord` itself) mint ONE `batchId` per commit REQUEST, threaded through every per-row `commitOneRecord` call — including across BJ-10's internal chunk boundaries, so a chunked job-commit still yields one batch.
- `batchId` is server-minted only; a client-supplied value in a request body has no effect (the request schemas never read it — G7 proves this).

## Golden matrix — G1-G8, all passing
- **Backend real-DB** (`packages/core-backend/tests/integration/multitable-ai-write-provenance-batch-grouping-realdb.test.ts`): G1, G2, G3, G4a, G4b, G5, G5b, G6, G7 — **10/10 pass** (incl. sentinel).
- **Frontend render test** (`apps/web/tests/multitable-history-center-ai-shortcut-label.spec.ts`): G8 — **2/2 pass**.
- **Non-vacuousness proof**: temporarily reverted the `batchId ?? randomUUID()` line back to unconditional `randomUUID()` and reran — exactly the 4 batch-**sharing**-dependent goldens (G2/G3/G4b/G5) went RED, everything else (G1/G4a/G5b/G6/G7) stayed GREEN as expected; reverted back to 10/10 GREEN.

## Judgment call (recorded in the goldens file header too)
The design-lock's G3 outcome vocabulary (`not_in_cache / skipped_no_perm / stale-drop / committed`) is bulk-commit's `RecordCommitOutcome`, **not** job-commit's `JobCommitOutcome` (which has no `not_in_cache` — a job row is either generated-and-confirmed or `pending_not_generated`). G3 here drives bulk-commit accordingly; **G5 covers the job-commit-specific claim instead** (BJ-10 chunking spanning multiple `patchRecords` calls still shares one batch), which is where a job-commit-specific proof adds unique value.

G4 ("two commit requests → two distinct batches") is proven via **two independent commit actions** — two bulk-commit calls (G4a), AND two separate jobs each committed once (G4b) — rather than two commits on ONE job. A job's commit is architecturally single-shot: every commit call (whether confirming all or a subset of generated rows) unconditionally resolves the job via `setHeaderAggregate(..., 'resolved')`, and `setHeaderRunning`'s committable set is `{suspended, errored, rejected}` — never `resolved`. G4b's test proves this live: a second commit attempt on an already-resolved job 409s `BULK_JOB_NOT_COMMITTABLE`.

## Verification
- Backend `tsc --noEmit`: clean. `apps/web` `vue-tsc -b`: clean.
- Full backend unit suite: **4186/4186 pass**, unaffected.
- Targeted real-DB regression sweep (AI + restore + history suites, the ones touching the modified write spine): green, except 2 flakes that reproduce **identically on a clean, unmodified `origin/main` baseline** when the same files run together (cross-file env-var/rate-limiter state bleed across concurrently-scheduled real-DB files — a pre-existing suite-concurrency artifact, not caused by this change). Each affected file passes 100% in isolation, on both `origin/main` and this branch.

### DIFF
--- a/apps/web/src/multitable/api/client.ts
+++ b/apps/web/src/multitable/api/client.ts
@@ -1184,6 +1184,8 @@ export interface AiBulkCommitRowOutcome {
 export interface AiBulkCommitData {
   outcomes: AiBulkCommitRowOutcome[]
   counts: Record<AiBulkCommitOutcome, number>
+  /** S1 LOCK-B6: server-minted history batch id for this commit action; null when nothing was written. */
+  batchId: string | null
 }
 
 // AI BULK fill — ASYNC JOB (B-4). When a bulk-preview request exceeds the inline
@@ -1293,6 +1295,8 @@ export interface AiBulkJobCommitData {
   state: AiBulkJobStatus
   counts: Record<AiBulkJobCommitOutcome, number>
   attempted: number
+  /** S1 LOCK-B6: server-minted history batch id for this commit request; null when nothing committed. */
+  batchId: string | null
 }
 
 /** Cancel response (multitable-ai.ts BJ-4 ~1316). `state` is `rejected` on success. */
--- a/apps/web/src/multitable/components/HistoryCenterModal.vue
+++ b/apps/web/src/multitable/components/HistoryCenterModal.vue
@@ -134,7 +134,7 @@ function sourceLabel(source: string): string {
   const map: Record<string, [string, string]> = {
     rest: ['手动', 'Manual'], multitable: ['手动', 'Manual'], button: ['按钮', 'Button'],
     automation: ['自动化', 'Automation'], import: ['导入', 'Import'], restore: ['恢复', 'Restore'],
-    admin: ['系统', 'System'], 'global-rbac': ['系统', 'System'],
+    admin: ['系统', 'System'], 'global-rbac': ['系统', 'System'], 'ai-shortcut': ['AI 填充', 'AI fill'],
   }
   const pair = map[source]
   return pair ? t(pair[0], pair[1]) : source
--- a/apps/web/tests/multitable-ai-bulk-fill-composable.spec.ts
+++ b/apps/web/tests/multitable-ai-bulk-fill-composable.spec.ts
@@ -43,6 +43,7 @@ const COMMIT_WIRE: AiBulkCommitData = {
     { recordId: 'rec_gone', outcome: 'not_in_cache' },
   ],
   counts: { written: 1, stale_reprev: 1, write_conflict: 0, not_in_cache: 1, skipped_no_perm: 1 },
+  batchId: 'batch_1',
 }
 
 function jsonResponse(body: unknown, init?: { status?: number; headers?: Record<string, string> }): Response {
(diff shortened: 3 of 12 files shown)


zensgit:
**Review fixes pushed** (`a01687f80`):

- **Medium (LOCK-B6 response contract)**: both `bulk-commit` and `job-commit` responses now return `batchId` (`commitBatchId` when written/committed > 0, else `null`). Updated FE types + the 4 existing test fixtures typed against them — including fixing a pre-existing exact-shape lock (`Object.keys(data).sort()`) that needed `batchId` added rather than left stale. Added real-DB response-batchId assertions to G2/G3/G4a/G4b/G5/G7 (all the bulk/job-commit goldens, not just G2) — G7 additionally confirms the response carries the server-minted id, never the client-supplied one.
- **Low (G3 no-perm coverage)**: added a `skipped_no_perm` row to G3's mixed batch (owned by `OTHER`, re-gated at commit — same recipe as the existing `multitable-ai-bulk-commit.test.ts` own-write-scope test), asserting it produces no revision and never joins the batch.

Re-verified independently on a fresh DB: real-DB goldens 10/10, 5 FE spec files 55/55, `vue-tsc -b` clean, backend `tsc` clean, full backend unit suite 4186/4186 unaffected.

zensgit:
**Second fix pushed** (`b44a375e5`): CI (test 20.x) caught a real gap in my previous fix — `tests/integration/multitable-ai-bulk-commit.test.ts` (the original, pre-S1 B-2 test file) has its own pinned-shape test (`Object.keys(res.body).sort()`) asserting the bulk-commit response's exact top-level keys. This is a sibling to the FE fixture lock I already updated, but on the backend integration-test side — a file I hadn't touched or run, since it lives in `tests/integration/` (needs `DATABASE_URL`) and wasn't part of my `tests/unit/` regression sweep. Added `batchId` to the pinned set there too.

Re-verified on a fresh DB + fresh migrations, single clean run: all 7 `multitable-ai-*.test.ts` integration files green (85/85). Also grepped the whole repo again for any other exact-shape lock on this response family, and for any closed-set enumeration of `RecordWriteSource` — nothing else found. Backend `tsc` clean, full unit suite 4186/4186 unaffected.

PR #3582, opened 2026-07-04
docs(multitable): grid perf measure-first baseline (Tier-B gap #4) — flat path already solved, grouped path is the real gap

zensgit:
Measurement-only pass on Tier-B gap #4 from `docs/research/multitable-feishu-refresh-audit-20260629.md`: *"Grid performance/virtualization — 100k+ rows. (Scale, not feature; measure-first.)"* No implementation change — per the audit's own discipline, this establishes a baseline before anyone attempts a rewrite.

## Headline: the gap as originally framed is largely stale — measuring surfaced a different, more urgent one

- **Flat (ungrouped) path**: already windowed — A1 row-virtualization shipped in #3008 (`085ce92be`, merged **2026-06-22, 7 days before** the 06-29 audit that lists this gap). Real measured numbers (Vue+jsdom mount, `apps/web/tests/multitable-grid-perf-baseline.spec.ts`): mount time and DOM row count are **flat across 1k → 100,000 rows** (15–36ms, ~33 DOM rows, regardless of scale). The audit's "100k+ rows" framing doesn't hold for this path — it's solved.
- **Grouped path**: genuinely **NOT windowed at all** (`flatWindowEnabled` requires `!groupedRows.value` unconditionally) — confirmed empirically, not just by reading the code comment. **3,000 rows / 30 groups took ~11.6s to mount** vs 15–36ms for the flat path at **100,000** rows. This is a narrower, more urgent problem than "100k+ rows generically" — grouped views (an ordinary, common configuration) degrade at a far lower threshold.

## Method + honesty caveats (see doc §2)
Real Vue+jsdom render (this repo's own `createApp`/`mount` test convention, e.g. `multitable-grid-databar.spec.ts`) — real signal for DOM-node-count-vs-scale and relative mount-cost, but **not** a real-browser paint/scroll-jank measurement (jsdom has no layout/paint pipeline; the grouped-path absolute timing should be read as directional severity, not a literal production millisecond figure). A Playwright-driven, live-dev-server measurement at production scale is named as explicit follow-up in the doc, not silently substituted.

## What's in this PR
- `apps/web/tests/multitable-grid-perf-baseline.spec.ts` — 5 tests, all passing, producing the numbers above (run `pnpm exec vitest run tests/multitable-grid-perf-baseline.spec.ts` in `apps/web`).
[...]

PR #3591, opened 2026-07-05
docs(multitable): DESIGN-LOCK (PROPOSED) — grouped-path grid row windowing (GW)

zensgit:
The design-lock that PR #3582's measure-first baseline called for — windowing the multitable grid's **grouped** rendering path, the real remaining half of audit Tier-B gap #4 (the flat path shipped in #3008 / `085ce92be`). **Docs-only; PROPOSED — awaiting ratification; no runtime code here.**

## Grounded, with one honest reframing of the baseline
- The mechanism gap is real: `flatWindowEnabled` hard-requires `!groupedRows` (`MetaGridTable.vue:622-627`); the grouped `<tbody>` renders every `groupRenderItems` entry — baseline measured 3,000 mounted rows / ~9.7-11.6s jsdom mount at just 3,000 rows/30 groups.
- **But grouped mode is pager-bounded in production today** (`infiniteScrollEnabled` also excludes grouped `:677`; classic footer pager stays, workbench `limit: 50`). So this lock is the **prerequisite for scaling grouped views** (+ a guard for any large-`rows` caller), not a live production fire — recorded explicitly in §1/§7 per code-truth-over-framing discipline.
- Two verified simplifiers: **subtotals are server-computed by composite path** (windowing cannot corrupt values — `:474-476`), and **grouped mode has no expand-rows** (`:181`/`:271` are flat-branch only).

## The lock
- **LOCK-GW-A**: window the EXISTING flattened `groupRenderItems` stream (`:838-861` — header|data|subtotal typed items) via a **per-kind-height prefix-sum offset table** + the same two spacer `<tr>`s the flat path uses; absolute `navIndex`/`path` identity is the keystone (same #3008 lesson). Rejected alternatives recorded: per-group windowing (DOM not viewport-bounded), uniform-height approximation (the exact reason the A1 comment says uniform math doesn't apply), full dynamic measurement (overkill for 3 near-constant kinds).
- **LOCK-GW-B**: `groupedWindowEnabled` predicate (reuses `VIRTUALIZE_MIN_ROWS=60` over ITEMS); printing/small sets force full render; **flat path and the grouped pager untouched** — grouped infinite-scroll accumulation is a named follow-up this lock *unblocks*, explicitly out of scope.
- **LOCK-GW-C**: 6 invariants — spacer-sum geometry, keyboard-nav parity through the offset table (`focusRow * rowHeightPx` stays flat-only), selection integrity, collapse-toggle clamp, subtotal render parity, group semantics untouched.
- **GW1–GW10 fail-first golden matrix**, including flipping the baseline spec's grouped test from "documents the gap" (asserts 3,000 mounted) to "locks the fix" (asserts bounded).
- **Binding real-browser acceptance**: the runtime PR must ship Playwright before/after (mount + scroll-frame) at ≥5k rows/≥50 groups — the baseline doc's own jsdom caveat, made a closing condition.

## Relationship to in-flight PRs
References #3582 (baseline, still open — this lock cites its branch head `f709acb9b`; no file overlap). No overlap with the AI-fields arc (#3584) or #3574.


zensgit:
# PR #3591 review — grouped-path grid row windowing design-lock — 2026-07-05

Verdict: APPROVE on design direction, with one P2 merge-order condition and one P3 doc-accuracy fix before merge.

What I verified against current `origin/main`:
- The grouped path is genuinely unwindowed today: `MetaGridTable.vue` renders `groupRenderItems` directly in the grouped `<tbody>`, while `flatWindowEnabled` requires `!groupedRows.value`.
- `groupRenderItems` is already the right abstraction for the proposed design: a flat typed stream of `header | data | subtotal`, with absolute `navIndex` preserved for data rows.
- The flat path and grouped path are already separate: `flatWindowEnabled` is flat-only, and `MultitableWorkbench.vue` only enables load-more on `gridIsFlatPath`, so the design's "grouped keeps classic pager today" framing is correct.
- The model is sound: per-kind height constants + prefix-sum offsets + binary-search windowing is the right conservative extension. It avoids the uniform-height trap, keeps subtotal values server-owned, keeps absolute nav/selection identity, and explicitly leaves grouped infinite-scroll/cross-page grouping out of scope.
- The Playwright/live-browser acceptance requirement is the right closure bar; jsdom row-count/timing alone should not close a rendering-performance lock.

Finding:

[P2] Merge-order dependency on #3582 needs to be enforced or made non-dangling.
This new design-lock references `docs/development/multitable-grid-perf-baseline-20260705.md` from #3582 as a file-level provenance source, but #3582 is still open and the file is not on `origin/main` yet. If #3591 lands first, main will contain a design-lock pointing at a non-existent evidence document. Simplest fix: merge #3582 first, then rebase #3591 and merge it. Alternative: adjust the provenance wording so it references PR #3582 without claiming a main-resident file until that PR lands. I prefer the first option because #3582 is already content-approved and it is the actual baseline this lock depends on.

[P3] The workbench `limit: 50` line reference points at the wrong current code.
The doc says the workbench pages at `limit: 50` around `MultitableWorkbench.vue:1682`, but current main at that location is a link-option/import lookup (`workbench.client.listLinkOptions`), not the main grid row loader. The underlying claim is still true: `useMultitableGrid.ts` has `DEFAULT_PAGE_SIZE = 50`, uses that `pageSize` for `loadView`, and `MultitableWorkbench.vue` gates load-more to `gridIsFlatPath`. Please retarget that citation to the composable plus the `gridCanLoadMore`/`gridIsFlatPath` lines, instead of the stale workbench line number.

No runtime blockers found. This remains docs-only and does not authorize grouped-windowing runtime.

zensgit:
P3 citation fix pushed in `ddb9945d3` (`docs(multitable): retarget grouped windowing pager evidence`).

What changed:
- Removed the stale `MultitableWorkbench.vue:1682` / `limit: 50` citation, which pointed at link-option lookup rather than the main grid row loader.
- Replaced it with the current evidence chain:
  - `useMultitableGrid` default page size is 50 and is passed as `loadView({ limit: pageSize, ... })`.
  - `MultitableWorkbench.vue` only feeds load-more when `gridIsFlatPath` / `gridCanLoadMore` are true.

That closes my P3 doc-accuracy finding.

Remaining condition from my review:
- P2 merge-order still stands: #3582 should land first, or #3591 should avoid a main-file reference to the #3582 baseline doc until that file exists on `main`.

zensgit:
Codex review verdict: PASS / RATIFY. The grouped-windowing design matches the current MetaGridTable architecture and keeps backend/data semantics out of scope. Owner should consciously accept the activation threshold as written: groupRenderItems.length > 60 counts headers/subtotals as viewport items, not only data rows. With that threshold call accepted, content is mergeable after #3582.

Note: GitHub would not allow this account to submit a formal APPROVE because it is the PR author, so recording the verdict as a PR comment.

zensgit:
## 独立审阅(Claude 值守会话)— APPROVE-for-ratification

行号引用抽查全部命中(:622/:601/:604/:677/:838/:864/:794)。**§1.2 的诚实重定性是本锁最有价值的部分**——核实 `infiniteScrollEnabled`(:677)同样排除 grouped,生产分组视图今天 pager-bounded ~50 行,故「扩容前置而非 hotfix」的自我定位准确,并正确把 grouped 无限滚动/跨页分组数据模型划出界(§6)。

模型判断:对既有 `groupRenderItems` 扁平流做 per-kind 常数高度 offset table + 二分 + 双 spacer,是 #3008 平铺架构的最小正确延伸;三个 rejected alternatives 的否决理由都对;C1–C6 覆盖真实回归面;GW1 显式翻转基线分组测试(与 #3582 耦合闭环);Playwright 真浏览器前后数字设为硬性条款是对 jsdom 告警的正确回应。

非阻断 [NIT]:§2 A1 的 H_header/H_subtotal 首测存在标准 remeasure 跳变风险(首次真实测量 ≠ CSS 默认时 offset table 重建使 scrollTop 对应内容位移)——建议 runtime 在高度常数变更时按锚定 item 校正 scrollTop,或 GW3 加「首测后视口无跳变」断言。不改变模型,ratify 后处理即可。

合并顺序维持 #3582 → #3591。ratify/合并动作留给 owner 与所属会话。全文:/tmp/pr3582-pr3591-review-claude-20260705.md

zensgit:
Review verdict: APPROVE after W0 prerequisite #3582 landed. I checked the lock against the current MetaGridTable implementation and the merged baseline:
- flatWindowEnabled excludes groupedRows, so flat-path windowing stays disjoint;
- grouped rendering consumes the flattened groupRenderItems stream without window slicing today;
- the design correctly narrows the problem to grouped-path scaling, not a current production fire for pager-bounded default grouped views;
- the runtime acceptance matrix is appropriately strict: offset-table geometry, nav/selection/subtotal parity, print/full-render fallback, and a real-browser Playwright before/after requirement.

I pushed a one-line provenance fix so the doc now points at #3582's merged squash b891780bd instead of the old open-branch head. Docs-only; authorizes no runtime.

PR #3609, opened 2026-07-05
feat(multitable): config-tier acceptance harness (uncreate/config-undelete/permission-revert) + runbook

zensgit:
Extends the established acceptance-harness rollout pattern (reset-acceptance.mjs / pit-undelete-acceptance.mjs) to the three built-but-unsmoked config tiers from the verified-state-map §4a (#3542), so each tier's staging acceptance can be operator-run ahead of its (separate, owner-gated) enablement decision.

## Harness — `scripts/config-tier-acceptance.mjs`
One binary, three independently-skippable tier sections (config-uncreate / config-undelete / permission-revert). Per tier:
- **flag-off contract**: preview + execute both → 403 `<TIER>_DISABLED` (flag state detected from the server's own first response, per tier — works in both staging postures)
- **flag-on walk**: provision precondition history on a THROWAWAY sheet → preview (masked shape + previewToken asserted) → typed-confirm execute → effect verified by follow-up read
- **key negatives**: no/wrong confirm → 400 `CONFIRM_REQUIRED`; stale tokens → 409 `PLAN_DRIFT` / `ID_COLLISION` / `GRANT_DRIFT`; escalation → 422 `RESTORE_NOT_SUPPORTED` (permission tier; automates the 5 smoke points of the 20260630 runbook)

Safety: writes ONLY to the base+sheet it creates (deleted at the end); `EDITOR_TOKEN` used solely as the de-escalation subject on the throwaway sheet and the whole tier SKIPs cleanly without it; tokens never printed; exit 1 on any FAIL.

Every asserted status/code is source-traced to `univer-meta.ts` / `config-restore.ts` and cross-checked against the three real-DB golden suites (mapping table in the harness header comments + runbook). Notably `GRANT_DRIFT` (409) is asserted BEFORE the escalation 422 in the drift fixture because the execute path checks the grant hash first (~8379 < ~8385).

## Runbook — `docs/development/multitable-config-tier-acceptance-runbook-20260705.md`
Env contract, per-tier scenario tables, expected codes, stop-conditions, evidence capture. States explicitly: **production flags are NOT changed by this work**; each tier's staging flip is a separate operator action gated on owner sign-off; **the harness has not yet been live-run — the first live run is the staging acceptance itself** (local run attempt hit two pre-existing, unrelated dev-DB migration-chain bugs, documented in the PR discussion below).

## Verify
`node --check` clean; codes source-traced (none invented); style mirrors the two precedent harnesses.

### DIFF
--- a/docs/development/multitable-config-tier-acceptance-runbook-20260705.md
+++ b/docs/development/multitable-config-tier-acceptance-runbook-20260705.md
@@ -1,0 +1,162 @@
+# Config-restore tier acceptance — runbook (staging, pre-enablement)
+
+Three default-off config-restore tiers ship built and real-DB-golden-tested but have **no staging acceptance
+harness yet**:
+
+| Tier | Flag | What it does |
+|---|---|---|
+| config-uncreate | `MULTITABLE_ENABLE_CONFIG_UNCREATE` | Reverting a field/view **create** = drop the entity (typed confirm, permanent, column data lost) |
+| config-undelete | `MULTITABLE_ENABLE_CONFIG_UNDELETE` | Reverting a field/view **delete** = recreate the **definition only** (typed confirm; values/links/auto-number are NOT restored) |
+| permission-revert | `MULTITABLE_ENABLE_PERMISSION_REVERT` | Reverting a `permission` change = re-apply its earlier grant, but **only when that reduces access** (de-escalation-only; escalation/no-op is refused) |
+
+All three are driven through the same two routes (`POST /sheets/:id/config-restore-preview` and
+`…/config-restore-execute`); the route branches on the shape of the recorded `meta_config_revisions` row the caller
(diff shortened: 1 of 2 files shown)


zensgit:
**Pre-existing dev-DB migration-chain debt found during the (attempted) local live-run** — unrelated to this PR, recorded for the owner:

1. Migrating a fresh/stale local DB fails at `tables_owner_id_fkey` (type mismatch `integer` vs `text`) on one path.
2. On the isolated docker dev Postgres, `zzzz20260411120100_approval_templates_and_instance_extensions.ts` has a swallowed error inside a `.catch()`-guarded `ALTER TABLE ADD CONSTRAINT` that poisons the migration transaction for the next statement.

Both reproduce with zero changes from this branch; both rolled back cleanly (no residual state). They only affect bootstrapping a local dev DB from scratch — CI's migration-replay gate stays green because it replays against its own baseline. Worth a separate hygiene fix if local-from-scratch bootstrap matters.

PR #3612, opened 2026-07-05
feat(attendance): override-import guard — forced-preview confirm + backup export (G1-G3)

zensgit:
## Summary

Implements the override-import guardrail per the RATIFIED design-lock (posture A) at
`docs/development/attendance-override-import-guard-design-lock-20260705.md` on PR #3597
(`claude/attendance-override-import-guard-lock-20260705`, head `6d48d9626`). Frontend-only,
zero backend changes, default import mode unchanged.

- **G1 — override submit gate (posture A, forced Preview-first).** `importMode==='override'`:
  the Import button is disabled until a fresh Preview completes (mirrors the accrual
  dry-run-required idiom — `:disabled="!annualAccrualDry"` + param-change clears the dry-run);
  any edit to file/mode/rule-set/profile/user-map/group-sync/timezone/payload invalidates
  that freshness live. Clicking Import then opens an in-DOM two-step confirm (same shape as
  `annualOpsConfirm`, kept as its own state since it also carries the backup action):
  restates mode + the Preview's `rowCount` + date range + distinct-user count, shows a
  **generic** "cannot be restored via rollback" warning (never a precise overwrite count —
  hard boundary, see below), requires a mandatory checkbox, and offers one-click "export
  backup first" against the existing `GET /api/attendance/export` route. Only after
  confirming does the original `runImport()` fire — the gate sits before the call, so it
  covers both the sync commit and the `/import/commit-async` lane identically.
- **G2 — merge stays one-click.** Non-destructive union never gates; the mode-selector's
  hint text now states the tradeoff explicitly (override = overwrite + non-recoverable;
  merge = non-destructive union).
- **G3 — manual-adjustment ledger note.** A negative (deduction) manual balance adjustment's
  existing `annualOpsConfirm` lines get one extra line: the deduction is written to the
  ledger and needs a reverse adjustment to restore (pure copy, no behavior change).

New pure module `apps/web/src/views/attendance/importOverrideGuard.ts` (a `<script setup>`
block can't export functions for direct unit testing) holds the confirm-gate's logic: line
building, warning/checkbox copy, preview-range summarizing, backup-target resolution, and
the export query shape. `AttendanceView.vue` wires it into the existing import state/handlers
(`requestRunImport()` replaces the old direct `@click="runImport"`).

### Precise-N hard boundary (design-lock §4)

The confirm modal never shows a precise "N existing rows will be overwritten" count. `GET
/api/attendance/export` (re-verified at `plugins/plugin-attendance/index.cjs` ~L41529) is
strictly single-`userId`-scoped (`targetUserId = query.userId ?? requesterId`) with no
"all users in org" mode — computing an exact overwrite count would need either a backend
change or one read call per distinct user, both out of scope here (design-lock §6 Deferred).
The modal only restates the Preview's own already-resolved scope (parsed row count + best-
effort date range/user set over whatever preview rows are currently loaded).

### Backup-export tradeoff (documented per design-lock §4 guidance)

The same single-`userId` scoping means a multi-user override import (CSV + user-map, no
top-level "User ID" set) has no honest single-call backup target. Rather than silently
export the *acting admin's own* range-scoped records (misleading — it would look like a real
backup but cover the wrong person), `resolveImportBackupTargetUserId()` only resolves a
target when the form's "User ID" field is set **or** the loaded Preview resolves to exactly
one distinct user; otherwise the "Export backup first" button is disabled with an inline
explanation. This is the documented tradeoff: date-range-level, single-user backup via the
existing route, not an all-employees-in-one-click backup (which the export route cannot do
without a backend change).

## Test plan

- [x] New `apps/web/tests/attendance-import-override-guard.spec.ts` (18 tests): 8 pure-module
      unit tests (line/warning/checkbox copy, range summarizing, backup-target/query
      resolution — including an explicit "never contains a digit" assertion on the warning
      text) + 10 mounted-component integration tests (override disabled-without-preview →
      enabled-after-preview, confirm modal opens with exact 4-line contents and no 5th
      overwrite-count line, confirm-submit gated on the checkbox and proceeds to the real
      `/import/commit` call, param-edit invalidates a fresh preview, merge one-click with no
      modal, exact `GET /api/attendance/export` wire-shape assertion (org/user/date-range
      query, parsed via `URLSearchParams`) + backup-disabled-when-ambiguous case, G2 hint
      copy, G3 ledger-note present-for-negative/absent-for-positive).
- [x] Wired into `attendance-web-guard` (both `pull_request`/`push` path-filter lists + the
      `vitest run` filter argument).
- [x] Local real run: `pnpm install --prefer-offline` → new spec alone (18/18) → full guard
      set `attendance-admin-regressions attendance-admin-anchor-nav
      AttendanceRequestCenterSection.result-edit attendance-selfservice-dashboard
      attendance-caliber-transparency attendance-punch-outcome attendance-import-override-guard`
      (220/220) → adjacent import specs (`attendance-import-preview-regression`,
      `attendance-import-ops-summary`, `AttendanceImportBatchesSection`,
      `AttendanceImportWorkflowSection`, `useAttendanceAdminImportBatches`,
      `useAttendanceAdminImportWorkflow`, 39/39) → `vue-tsc -b` (exit 0).
      (`attendance-import-batch-timezone-status.spec.ts` fails both before and after this
      branch — confirmed pre-existing on `origin/main`, unrelated to this change, and not
      part of the attendance-web-guard targeted set.)
- [x] Mutation self-check: committed, then cut the G1 gate (`requestRunImport()` forced to
      call `runImport()` unconditionally) — 4 of the 18 tests (the ones asserting the confirm
      modal actually opens/gates/backs-up) went red as expected; reverted via
      `git checkout --`, full guard set re-verified green (220/220), `vue-tsc -b` re-verified
      (exit 0).

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/attendance-web-guard.yml
+++ b/.github/workflows/attendance-web-guard.yml
@@ -12,9 +12,14 @@
 # for the shared AE-3 result-edit callback), attendance-selfservice-dashboard.spec.ts,
 # attendance-caliber-transparency.spec.ts (display-only 口径透明 slice — #3575 design-lock:
 # G1 records-header title, G2 team-availability legend equation, G3 caliber-guide card),
-# and attendance-punch-outcome.spec.ts (punch outcome clarity: 202 pendingApproval /
+# attendance-punch-outcome.spec.ts (punch outcome clarity: 202 pendingApproval /
 # OUTDOOR_NOTE_REQUIRED / LOCATION_RESTRICTED — see
-# docs/development/attendance-punch-outcome-clarity-design-lock-20260705.md). The full
+# docs/development/attendance-punch-outcome-clarity-design-lock-20260705.md), and
+# attendance-import-override-guard.spec.ts (override-import confirm gate: forced-preview
+# submit-disable, in-DOM confirm with rowCount/date-range/user-count lines + generic
+# irreversible warning (never a precise overwrite count), export-backup wire shape, merge
+# one-click, G3 manual-adjustment ledger note — see
+# docs/development/attendance-override-import-guard-design-lock-20260705.md, PR #3597). The full
 # `@metasheet/web` vitest suite still has historical/flaky failures, so wiring the whole
 # suite as a required gate now would be red on arrival. Keep expanding the `vitest run`
 # filter (or move to the full web suite) as more web specs are cleaned up.
@@ -31,6 +36,7 @@ on:
       - 'apps/web/tests/attendance-selfservice-dashboard.spec.ts'
       - 'apps/web/tests/attendance-caliber-transparency.spec.ts'
       - 'apps/web/tests/attendance-punch-outcome.spec.ts'
+      - 'apps/web/tests/attendance-import-override-guard.spec.ts'
       - '.github/workflows/attendance-web-guard.yml'
   push:
     branches: [main]
@@ -43,6 +49,7 @@ on:
       - 'apps/web/tests/attendance-selfservice-dashboard.spec.ts'
       - 'apps/web/tests/attendance-caliber-transparency.spec.ts'
       - 'apps/web/tests/attendance-punch-outcome.spec.ts'
+      - 'apps/web/tests/attendance-import-override-guard.spec.ts'
       - '.github/workflows/attendance-web-guard.yml'
 
 jobs:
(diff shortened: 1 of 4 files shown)


zensgit:
## 对抗审阅 + 修复回合（Claude, opus 审阅 → 主循环修复）

审阅判定 = 忠实实现 RATIFIED 姿势-A 锁，但发现一处真实确认门旁路（P2）+ 两个 polish（P3）。已全部在分支修复并 mutation-锁死：

- **P2-1（真旁路，已修）**：`retry-run-import` 状态动作原先直接 `runImport()`，绕过 override 确认/预演门——例如「merge 导入失败→下拉切 override→点重试」会提交一次从未确认的 override。现改为经 `requestRunImport()` 同门（merge 保持一键）。两个回归测试锁死不变量（无静默 re-commit；重试必重现 modal-or-preview-required 门）。**mutation 验证**：还原该修复 → 两个新测试翻红。
- **P3-1（已修）**：截断样本（shown<total）不再自动解析单用户备份目标（样本里恰一个用户不代表全量），且模态新增「样本已截断 N/total」披露行。
- **P3-2（已修）**：死 export `importModeRequiresConfirm` 现为 `requestRunImport()`/retry 的真实谓词（原为内联 `!== 'override'`）。

验证：guard 全集 **223/223**、新 spec **21/21**、`vue-tsc -b` clean。审阅 MD：`/tmp/pr3612-review-claude-20260705.md`。

**判定：修复后 APPROVE**——0 P1 / 0 P2 残留。这是**行为变更**切片（确认门）+ RATIFIED 锁背书，按约不自动合并——**就绪待 owner 点头合并**。

PR #3617, opened 2026-07-05
feat(integration): composition authoring-tier client service calls

zensgit:
## Summary

**N1** of the authoring-surface round (#1709): the config-time client service calls the composition **authoring panel** (N2, next slice) needs — closing the runbook-flagged gap where composition authoring required raw curl. Client-only; the C-R4-1 server routes are untouched.

- `buildReadSourceCompositionPayload(draft)` — pure builder; `version:1` / `operations:['read']` / step ids / `fromStep` / `toInput:'key'` all **pinned constants** (only the 4 named draft fields are ever read — a smuggled extra draft field can't reach the payload; test-locked).
- `validateReadSourceCompositionDraft` — coarse bounded-identifier pre-checks; field-keyed messages that **never echo raw values** (test: secret-shaped overlong values not present in any message).
- `saveReadSourceCompositionVersion` (POSTs exactly `{config}`; 201→`reused:false`, 200→`reused:true`) + `approveReadSourceComposition` / `retireReadSourceComposition`.
- `listReadSourceCompositionAudit` — values-free rows: action closed-enum (unknown actions dropped whole), `detail` clamped to only `{from,to}` status enums (a `version` or smuggled field is dropped).
- `ReadSourceCompositionApiError.fieldErrors` — save-time CONFIG_INVALID triples clamped with the SAME three patterns as #3588's planErrors (entry failing any clamp dropped whole; sentinel-scanned).

## Verification

- 33/33 (service 24 + mirror 3 + run panel 6) — re-verified independently in a clean worktree; `type-check` clean.
- Only 2 files touched (service + its spec); run panel / mirror spec / server untouched.

## Boundary

Client config-time surface over existing routes; no server change, no write path, no recursion.

Drafted by a Sonnet-5 agent against a discipline-pinned spec; reviewed line-by-line (clamps, pins, audit normalizer, no-echo) + locally re-verified before this PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3619, opened 2026-07-05
feat(integration): composition authoring panel

zensgit:
## Summary
- add an independent read-source composition authoring panel for consultant/admin use
- select two approved resolver_lookup read-source configs, save a pinned read-only composition version, and approve/retire/audit saved compositions
- mount the authoring panel separately from the existing key-only composition run panel

## Boundaries
- no server/runtime changes; consumes the #3617 authoring-tier client service only
- no chain execution from the authoring panel, no runtime key submission, no recursion, no write/delete path
- error and audit rendering stays values-free via the existing client clamps

## Verification
- pnpm install --frozen-lockfile --prefer-offline
- pnpm --filter @metasheet/web exec vitest run tests/IntegrationReadSourceCompositionAuthoringPanel.spec.ts --watch=false
- pnpm --filter @metasheet/web exec vitest run tests/IntegrationReadSourceCompositionPanel.spec.ts tests/readSourceCompositions.service.spec.ts --watch=false
- git diff --check

Related: #1709

### DIFF
--- a/apps/web/src/components/integration/IntegrationReadSourceCompositionAuthoringPanel.vue
+++ b/apps/web/src/components/integration/IntegrationReadSourceCompositionAuthoringPanel.vue
@@ -1,0 +1,386 @@
+<template>
+  <div class="integration-read-source-composition-authoring" data-testid="rscomp-author-panel">
+    <p class="integration-read-source-composition-authoring__hint">
+      顾问/管理员在这里把两个已审批 resolver_lookup 读取源编排成只读组合链。面板只保存组合配置版本,
+      不执行组合、不提交逐跳 key、不含写入能力。
+    </p>
+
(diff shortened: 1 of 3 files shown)


zensgit:
Self-review delta pushed in 4b7d34c3d: the authoring panel now renders raw Error.message only for the two clamped service error classes (ReadSourceApiError / ReadSourceCompositionApiError); unexpected thrown errors fall back to a fixed coarse message. Added a red test with `transport failed near MAT-001 SECRET` to prove the raw message/key cannot reach the panel. Local checks: authoring panel 5/5, existing composition panel+service 30/30, web type-check, git diff --check.

zensgit:
Closing as a duplicate earlier N2 implementation. The canonical authoring panel landed via #3621, and the one extra defensive clamp from this draft has been isolated into follow-up #3625.

PR #3621, opened 2026-07-05
feat(integration): composition authoring panel — consultant config-time UI

zensgit:
## Summary

**N2** of the authoring-surface round (#1709): the consultant/config-time **authoring panel** — closes the runbook-flagged gap (composition authoring previously required raw curl). Consumes the N1 service layer (#3617) exclusively.

- Dedicated component `IntegrationReadSourceCompositionAuthoringPanel.vue` (NOT inlined into the 22k workbench SFC): two step pickers (approved `resolver_lookup` configs only, distinct-selection enforced), name + `sourceTarget` inputs, client pre-check list, save → saved row (id/version/status/reused) with **clamped fieldErrors** on 400, approve/retire with disable-on-inflight + **requestId+signature stale guard**, values-free audit list, boundary hint.
- **Payload discipline**: the component never assembles the wire body — the draft goes to `saveReadSourceCompositionVersion` (pinned builder inside the service), so a tampered draft field can't reach the wire (test-locked).
- **Permission idiom**: `integration:write` actions-denied (disabled + handler guard), deliberately NOT panel-hide — matches the server's exact route split (save/approve/retire = write-tier; list/get/audit/run = read-tier), so pickers/audit stay readable while mutations are gated.
- Mount: 11-line diff in the workbench view, between the read-source config panel and the run panel.

## Verification

- **87/87** re-verified independently in a clean worktree: authoring 6 + run panel 6 + service 24 + vocab mirror 3 + full `IntegrationWorkbenchView.spec.ts` 48; `vue-tsc -b` clean; `git diff --check` clean.
- Sentinel scans in tests (7 assertions): business-value triples dropped from rendered HTML; 409 coarse render with no value echo.

## Boundary

Client config-time surface over existing routes; no server change, no write path, no recursion. Entity-runbook Phase 1 can now be panel-driven end-to-end.

Drafted by a Sonnet-5 agent against a discipline-pinned spec; reviewed (payload-discipline, permission-tier decision, stale guard) + independently re-verified before this PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3624, opened 2026-07-05
docs(integration): composition authoring-surface round — design & verification MD

zensgit:
Round ledger: **N1** authoring service #3617 (`4c7665752`) + **N2** authoring panel #3621 (`9e77045a5`) — closes the runbook-flagged raw-curl gap (entity acceptance Phase 1 now panel-driven). Design points (pinned builder / fieldErrors clamp / values-free audit / write-tier actions-denied aligned to the server route split), verification matrix (33/33 + 87/87 independently re-verified), model-assignment record (third validated round of Sonnet-drafts-behind-Fable-gate). Docs only; authorizes nothing.

### DIFF
--- a/docs/development/integration-composition-authoring-surface-dev-verification-20260705.md
+++ b/docs/development/integration-composition-authoring-surface-dev-verification-20260705.md
@@ -1,0 +1,76 @@
+# 组合 authoring 面(config-time UI)一轮 — 设计与验证 — 2026-07-05
+
+## 0. 本轮定位
+
+实体机验收 runbook(#3598)诚实标注了组合线最后一块产品缺口:**组合 authoring 无 UI,Phase 1 只能裸
+curl**。本轮固定节奏关掉该缺口——纯 client、config-time、走既有 C-R4-1 路由,零 server 改动:
+
+```text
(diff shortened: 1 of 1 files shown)


zensgit:
P2 ledger sync before merge, because #3621 merged just before the generic-error clamp review note landed.

#3625 is now the follow-up that closes that last N2 client render edge: unknown/generic `Error.message` in the authoring panel is collapsed to a fixed fallback, while `ReadSourceCompositionApiError` / `ReadSourceApiError` remain the only trusted clamped message sources.

Once #3625 lands, please update this MD before merging (or I can do a tiny follow-up doc PR):

- line 10: N2 should be `#3621 9e77045a5 + #3625 <sha> hardening`, not only #3621
- §1.2 add one bullet for generic-error values-free render: only known clamped service error classes render message; generic/transport Error gets fixed fallback
- §2 N2 verification should become authoring `7` tests, total `88/88` if the rest is unchanged
- §5 boundary/discipline can mention #3625 as same-surface hardening, no server/write/recursive change

This is not a design issue with #3624; just making the final ledger match the actual main sequence after the late P2 clamp.

PR #3630, opened 2026-07-05
feat(attendance): report-sync A2 scheduled trigger runtime

zensgit:
## Summary

- Implements the A2 report-sync scheduled trigger runtime: the only missing piece of the already-shipped report-records -> multitable sync machinery (the `mode:'enqueue'` job path had no worker).
- S1: latent settings `reportSync.scheduledTrigger.{enabled,cadence,maxOrgsPerRun,maxUsersPerRun}` (default fully off, enum-strict zod) + a dormant `AttendanceScheduler` job `attendance-report-sync-scheduled` (env flag AND settings both required — double-gate).
- S2: once both gates are open, the runner claims/resumes a `plugin_attendance_report_sync_jobs` `mode:'enqueue'` row per org (bounded by `maxOrgsPerRun`/`maxUsersPerRun`) and delegates every write to the **existing** `syncAttendanceReportRecordsForUsers` writer via the existing `runAttendanceReportSyncJobNextPage` runner — zero write-semantics change. Repeat ticks are no-ops via the job table's `(org_id, idempotency_key)` partial unique index plus the writer's own dual source-fingerprint compare.
- Settings are wired into both the normalizer and the PUT `settingsSchema` zod (the #1829 lesson: a key added to only one silently strips on round-trip).
- Registered as an independent `AttendanceScheduler` job so one failure never skips expiry/reminder/delivery/auto-write/digest (composite-registry discipline).
- `maxUsersPerRun`'s valid range is tightened to 1..100 (not 2000) to match `ATTENDANCE_REPORT_RECORDS_SYNC_MAX_PAGE_SIZE`, the real per-page ceiling the runner's `pageSize` is silently clamped to downstream — keeps the validated range honest so a configured value never does less than it claims.
- No admin UI (backend-only v1, stays off the FE serial lane, per design-lock §3).

Spec source: `docs/development/attendance-report-sync-a2-scheduled-trigger-design-lock-20260705.md`, authored on the design-lock branch tracked by PR #3623 (RATIFIED, owner-delegated). **Note:** at push time #3623 is still open (not yet on `main`) — its content was read directly from that branch since this runtime task was dispatched assuming it had already merged. Recommend landing #3623 so `main` matches the spec this PR implements; no action taken here beyond reading its content.

## Recovery / continuation note

This resumes a session-limit-interrupted WIP (`ecf0381af`) that had already written the full runtime + most of the test matrix. Assessment: the runtime and almost all required tests were already complete and correct. Rebasing onto current `main` plus a full local real-DB test pass surfaced and fixed three pre-existing WIP bugs (all in test code / validation-range consistency, not the core claim/page/delegate logic):

1. `maxUsersPerRun`'s zod/normalizer max (2000) didn't match the real per-page ceiling the writer silently clamps to (100) — tightened the validated range instead of leaving a "configured value does less than advertised" gap.
2. The integration test's fake multitable fixture hardcoded the row-key field id as `fld_rowKey` instead of the real `fld_row_key` (matching the sibling reference test file's own convention) — this silently broke `queryRecords` matching once any record existed, causing later users in the same tick to patch the first record's data instead of creating their own.
3. The "PUT round-trip" test's cleanup did a whole-settings-blob restore (the convention used elsewhere in this file) which is fragile against an unrelated pre-existing schema asymmetry (`holidaySync.lastRun.error` round-trips as `null` from GET but the zod schema demands a `string` on PUT) — rescoped this test's restore to just the `reportSync` key it actually touches, sidestepping that unrelated issue entirely (not fixed here — out of scope for A2).

Also added one test the WIP was missing: the original double-gate test only ever exercised the *settings* half of the gate (env was toggled while settings stayed off in both cases, so it could never have caught a deleted env check). Added a second scenario — settings enabled, env unset — to independently prove the env half of the double-gate, confirmed via mutation below.

## Test plan

- `node --check plugins/plugin-attendance/index.cjs` — syntax OK.
- Unit (no DB): `tests/unit/attendance-report-sync-scheduled-trigger.test.ts` (16 tests, new) + `tests/unit/attendance-scheduler.test.ts` (22 tests, +1 new isolation case) — 38/38 pass.
- Integration (real DB, ephemeral Postgres 15 matching CI's `ankane/setup-postgres` + migration set): ran the new `attendance report sync scheduled trigger (A2, ...)` describe block in `tests/integration/attendance-plugin.test.ts` (5 tests: both double-gate halves independently / gate-open write-delegation + same-period no-op repeat / `maxUsersPerRun` page-by-page resumption across ticks / PUT settings round-trip incl. sibling-preserving partial update + invalid-enum 400) — 5/5 pass.
- Ran the **full** `attendance-plugin.test.ts` file once against a completely fresh migrated DB (matching CI exactly): 152/154 pass. The 2 failures (`auto shift matching preview` date off-by-one) are pre-existing and unrelated — reproduced identically with this branch's changes fully `git stash`-ed, and traced to my local sandbox's `Asia/Taipei` system timezone vs. CI's UTC runner; not touched by this PR.
- This file (`tests/integration/attendance-plugin.test.ts`) is already wired into the existing attendance real-DB CI gate (`.github/workflows/plugin-tests.yml`, "Run attendance integration tests" step) — no workflow changes needed.
- Mutation self-check (on the committed clean state, each mutation reverted immediately after observing red):
  - Removed the settings half of the double-gate check → the original double-gate test's settings-off scenario went red (env-half test unaffected).
  - Removed the env half instead → the newly-added env-isolation test went red (the original test unaffected) — confirms the two double-gate tests together, not either alone, cover both halves.
  - Removed `AttendanceScheduler.runCycle()`'s per-job try/catch → both the pre-existing generic isolation test and this PR's new `attendance-report-sync-scheduled`-specific isolation test went red.
  - All three mutations reverted; working tree confirmed clean before push.

## Risk notes

- `docs/development/attendance-report-sync-a2-scheduled-trigger-design-lock-20260705.md` is not yet on `main` (PR #3623 open) — flagging per above so it lands alongside or before this runtime.
- S3 (staging smoke, residue=0) from the design-lock's completion criteria is not run here — no staging access from this sandbox; left for the same follow-up step used for the prior A2 auto-write feature.
- `maxOrgsPerRun` has no equivalent secondary clamp (unlike `maxUsersPerRun`/`ATTENDANCE_REPORT_RECORDS_SYNC_MAX_PAGE_SIZE`) — left its existing 1..50 range as-is; no hidden ceiling found there.


zensgit:
## opus 对抗审阅 + P2-1 修复回合

审阅判定 = **APPROVE，0 P1 / 0 P2-blocking**——三条最重红线（零写入语义变更 / 单 job 失败隔离 / env+settings 双门）经代码通读 + mutation + 真 DB 三重验证全部干净；fld_row_key 确认为测试专属修复非生产 bug；holidaySync 规避为真·pre-existing 无关 bug。

审阅提的 **P2-1 latent（多组织 fan-out 无序 + 静默饥饿）已修**——虽非 v1 阻断（off-by-default、小 N 首用、锁定义 cap 为限流非公平），但"静默 + 不确定"是危险的一半，用最小改动堵死、不引入 lock 明确 deferred 的完整轮转复杂度：

- **(1) `ORDER BY org_id`** → 被覆盖的 org 确定（消除不确定性）；
- **(2) 可选 logger warn** → org 数超 cap 时输出精确覆盖数（totalOrgs/cap/uncoveredOrgs），短缺**可观测非静默**；
- 真正的跨-tick 公平轮转仍为文档化 follow-up（锁 throttle-not-fairness 框架）。

+4 单测（ORDER BY 确定性 / 超 cap warn 精确数 / cap 内不 warn），**两 mutation 锁死**（砍 ORDER BY→确定性测试红；砍 warn→可观测测试红），type 注解同步（tsc clean），resolver+scheduler 单测 41/41。off-by-default v1 零回归。

**判定：修复后 0 P1/P2 残留，就绪待落。** backend 切片 → 待 owner 点头 + 落地次序 **#3623 锁先 → #3630 runtime**（rebase 待 CI 绿即合）。审阅 MD：`/tmp/pr3630-review-claude-20260705.md`。

PR #3646, opened 2026-07-05
docs(multitable): W1-1 formula freshness — audit + DESIGN-LOCK (PROPOSED)

zensgit:
Converts the owner-passed W1-1 pre-drafts now that the W0 review batch cleared (per the goal-pool plan's owner-approved sequence: W1-1 lock first, W1-2 B1 side-door goldens follow as a cross-referenced separate PR).

**Audit** (owner-passed; anchors refreshed): Q1 **GAP** — Yjs bridge flush stubs `recalculateFormulaFields` ('REST PATCH path for now'), collab edits leave formula values stale until next REST write. Q2-c — expression change triggers **no bulk recompute** (no entry point exists). Q3 (1-hop fan-out) / Q4 (on-save only) self-consistent → non-goals.

**Design-lock** (PROPOSED — awaiting ratification): bridge recompute via optional `actorId` helper seam, real impl exported from `univer-meta.ts` (its two `req` consumers are chokepoint-guarded there); taint discipline byte-for-byte, **no system/full-access bypass** ('system context' rejected as a mask-laundering channel); lookup/rollup stubs OUT (computed-on-read; named residual + pinned golden). Expression-change bulk recompute: synchronous, pre-commit COUNT gate, env cap default 1000, over-cap **422 + config NOT saved** (fail-closed); writer context = config editor's own req; derived raw UPDATE, zero revisions (fan-out precedent); idempotent. GF1–GF11 fail-first matrix.

Sibling: W1-2 **B1 side-door goldens** (permission angle of the same Yjs-bridge side-door) — separate PR, mutual cross-reference.

Docs-only. Not merging — awaiting ratification.


zensgit:
Codex review verdict: REQUEST CHANGES.

Medium — §3 has an internal trigger/idempotency contradiction. B1 locks the bulk recompute trigger to a formula PATCH where the normalized property.expression actually changed, but B6/GF8 require a mid-chunk failure to be recoverable by re-saving the same expression, with the second save converging to full freshness. Those cannot both be true: after the first request commits the expression, a same-expression retry is no longer an actual expression change, so the B1 trigger would not fire and GF8's convergence path would be impossible.

Recommended fix: choose and write one explicit contract. My preference is to broaden B1 to include an explicit same-expression retry when the request carries property.expression for a formula field, with no config revision on the no-op retry but recompute still allowed; then keep B6/GF8. If you want to keep B1 strictly actual-change-only, then B6/GF8 must drop the re-save convergence promise and name a separate retry/sweep follow-up. Either way, the design-lock needs one coherent trigger model before ratification.

zensgit:
**B1/B6/GF8 fix pushed** (`2c38068b8`) — took your preferred direction (widen B1): trigger is now 'request explicitly carries `property.expression`' (fires on both actual-change AND same-expression re-save = the recovery lever), not 'value differs'. Recompute-trigger ↔ config-audit decoupled: same-expression re-save is a config no-op so writes no `meta_config_revisions` row (verified: recorder comment 'empty diff → null'), but recompute still fires — retrying to freshness never inflates the audit trail. B6 reworded, GF8 spells out the re-save, new GF12 pins the decoupling. B6/GF8 convergence promise intact. Awaiting re-ratify.

zensgit:
Codex re-review verdict: PASS / RATIFY on content. The REQUEST CHANGES item is resolved on head 2c38068b8: B1 now triggers on an explicit property.expression request rather than actual-change-only, B6 uses same-expression re-save as the recovery lever, GF8 names that retry path, and GF12 pins the decoupling (recompute fires, zero new meta_config_revisions rows). This is now internally coherent and matches the intended idempotent recovery model.

Merge note: at re-review time test (18.x) and test (20.x) were still pending and the branch was BEHIND, so land after CI is green/current. No remaining content findings.

PR #3649, opened 2026-07-05
test(multitable): W1-2 permission-matrix B1 — Yjs bridge / OAPI-token / base-write side-door goldens

zensgit:
## Summary

Lands the W1-2 permission-matrix spec (`docs/development/multitable-w1-2-permission-matrix-spec-20260705.md`, landed verbatim) and its **B1 batch only** (highest-risk, first — B2/B3/B4 are explicitly out of scope for this PR). Test-only; **zero runtime code changes**.

18 new real-DB goldens across 3 gaps:

- **G-1 (11 goldens)** — `multitable-permmatrix-b1-yjs-bridge-flush-realdb.test.ts`: the Yjs collab bridge's scalar flush now resolves capabilities via the REAL `resolveSheetCapabilitiesForUser` (DB-backed RBAC + sheet scope + admin role) instead of the hardcoded grant the pre-existing `yjs-scalar-*-realdb` suites use (those isolate value fidelity, not permissions — see their own file comments). Covers actor tiers A3 (global read-only → denied) / A4 (sheet-scoped full write, no global write → **allowed**, proving the bridge honors a real per-sheet grant) / A5 (write-own: own allowed, not-own denied) / A6 (no permission → denied), plus M3 record-lock (a write-capable actor is still denied on a locked record — a true independent gate) and M2/M4 target goldens (see finding below).
- **G-2 (4 goldens)** — `multitable-permmatrix-b1-oapi-token-write-realdb.test.ts`: an `mst_` token PATCH is the *identical* handler the interactive session hits (`apiTokenAuth` only substitutes `req.user`) — pins that parity explicitly. M3 lock → 403 zero side effects; M2/M4 targets → 200 (parity, see finding below).
- **G-3 (3 goldens)** — `multitable-permmatrix-b1-basewrite-no-sheet-write-realdb.test.ts`: an actor holding *only* `multitable:base:write`, or the base's bare owner (`meta_bases.owner_id`) with zero permission codes, gets 403 on an interactive record PATCH — `resolveBaseWritable` is automation-executor-only and never feeds `deriveCapabilities`, so base-level write authority structurally cannot leak into sheet/record write (the "M6 crux" the spec names but had never golden-locked).

### Finding recalibrated during implementation (documented, not a defect)

[...]

PR #3651, opened 2026-07-05
test(multitable): W1-2 B1 side-door permission goldens (G-1/G-2/G-3)

zensgit:
## What this is

**W1-2 B1 batch** — real-DB integration test PR (test-only, no runtime changes) for the three highest-risk side-door / write-modifier clusters named in the owner-passed spec `docs/development/multitable-w1-2-permission-matrix-spec-20260705.md` §3 (gaps G-1/G-2/G-3, batch B1 in §5).

**Sibling cross-ref**: #3646 (W1-1 formula-freshness design-lock, MERGED) locks the *freshness* angle of the Yjs-bridge side-door (bridge writes previously skipped formula recompute). This PR locks the *permission* angle of the same side-door, plus two adjacent write-modifier clusters (OAPI token write, base-write-vs-sheet-write). Deliberately a separate PR, per #3646's own §7 arc placement.

## Framing (important, owner-corrected)

**G-1 is a test blind-spot, NOT a runtime hole.** The Yjs bridge already resolves real capabilities via `resolveSheetCapabilitiesForUser` ("Resolve real user capabilities — same path as REST", `index.ts:2471,2473`) and writes through the same `RecordWriteService.patchRecords` spine as REST (`yjs-record-bridge.ts:226`). Authorization machinery is already in place on this path — existing `yjs-scalar-{flush,seed}-realdb.test.ts` suites just isolate *value fidelity* and grant `multitable:write` unconditionally by design ("since this test isolates value fidelity, not permissions" — their own comments), so nobody had pinned that a non-writer's flush is actually rejected via *real* DB-backed capability resolution. **These goldens lock that authorization stays enforced on the bridge path going forward — they do not claim to close a bypass.**

## What each cluster locks

**G-1** — `packages/core-backend/tests/integration/multitable-permmatrix-b1-yjs-bridge-flush-realdb.test.ts` (12 tests)
- Actor tiers A3 (global read-only) / A4 (sheet-scoped full write, no global write) / A5 (sheet-scoped write-own, own vs. not-own record) / A6 (zero permissions) — non-writers denied with zero side effects (no data change, no version bump, no revision row); A4/A5-own prove the bridge honors a *real* per-sheet grant, not just global permissions.
- M3 (record lock): a globally write-capable actor is still denied on a record locked by someone else — the lock gate applies on the bridge path too.
- M2 (row-level read-deny) / M4 (conditional rule-deny): both are **read-only constructs** — verified by exhaustively grepping every write-path file in the repo (`RecordWriteService`, `RecordService`, and the `RuleEffect` type itself, which is literally typed as `'deny_read'` only, no write variant exists to express). Each gets a non-writer-denied golden (defense-in-depth) *and* a writer-succeeds golden (documented non-gate, parity with interactive REST — same convention as `docs/development/permission-matrix-golden-20260525.md` §2).
- **New beyond the spec's minimum**: a property-level field-mask ("masked-target") assertion — a writer's flush touching a `formula`-type field alongside a normal field is rejected **atomically** (`RecordWriteService.validateChanges` throws `RecordFieldForbiddenError` for the whole flush, before any transaction opens), matching what both REST routes (bulk `/patch` and single-record `PATCH /records/:recordId`) do for the identical field classes. This directly satisfies the task's "a writer flush must still respect field-mask... targets the same way REST does" ask with a mechanism that's genuinely, uniformly enforced everywhere (not just capability-gated).

**G-2** — `packages/core-backend/tests/integration/multitable-permmatrix-b1-oapi-token-write-realdb.test.ts` (4 tests)
- An `mst_` `records:write` token PATCH re-gates the same three modifiers as interactive: **locked record → 403, zero side effects, denied audit row** (M3, true gate); **conditional-rule-denied write → still succeeds** (M4); **row-denied write → still succeeds** (M2). The token route (`univer-meta.ts:13851`) is not token-exclusive — `apiTokenAuth` falls through to session `req.user` when no `mst_` Bearer token is present — so this is the general single-record PATCH handler under both auth modes, with no second authorization implementation to drift.
- **Re-grounding correction**: the spec/task text describes all three modifiers as "blocked." Re-anchoring against the write spine shows only M3 is a true gate; M2/M4 are non-gates everywhere (confirmed above). The goldens assert the *actually enforced* parity (token channel matches interactive channel exactly, whichever way that cuts) rather than a stronger property that isn't true anywhere in the codebase.

**G-3** — `packages/core-backend/tests/integration/multitable-permmatrix-b1-basewrite-no-sheet-write-realdb.test.ts` (3 tests)
- Pins the C2 crux fact (documented in the spec, never golden-locked): `resolveBaseWritable`'s base-level write codes (`multitable:base:write`/`multitable:base:admin`, or `meta_bases.owner_id` ownership) are a separate, kernel-internal primitive for the cross-base automation path — `deriveCapabilities` (which gates interactive record PATCH) has zero knowledge of them. An actor holding *only* base-write authority, or the base owner with zero permission codes, gets 403 on an interactive record PATCH, zero side effects. A regression here would be a silent privilege escalation with no code path announcing it.

## Adjacent finding (out of scope, not golden-locked, flagged per the "discover a runtime defect → stop, report separately" convention)

While grounding G-1's masked-target case, direct code reading turned up an asymmetry: the **per-subject** `field_permissions.read_only` write gate (distinct from the property-level guard G-1 locks) is enforced at only **3 route-level call sites** — the bulk `/patch` route's own pre-check (whose comment literally says *"RecordWriteService.patchRecords enforces only PROPERTY-level guards ... it does NOT enforce per-subject field_permissions.read_only"*), record-copy, and revision-restore. Neither `RecordWriteService.patchRecords` (bridge, and bulk `/patch`'s underlying spine) nor `RecordService.patchRecord` (the single-record `PATCH /records/:recordId` handler used by G-2/G-3 and by ordinary interactive single-record edits) consults `field_permissions` for a write decision at all — verified by reading both functions in full. Unlike M2/M4 (confirmed-intentional non-gates by design), this looks like an incomplete rollout of a feature that 3-of-5ish write paths already enforce, so it is **not** asserted as a "documented non-gate" golden here — it's reported as a candidate follow-up for the owner to triage, not fixed or test-locked in this PR.

## Verification (real numbers)

- Fresh dedicated DB (`metasheet_w12b1_test`, single clean run, migrated via `pnpm run db:migrate` to the current tip — `zzzz20260705150000_create_meta_view_personal_configs`).
- All 3 new files, single run: **3 test files passed, 19 tests passed** (12 + 4 + 3; includes 3 DATABASE_URL sentinels).
- **Non-vacuousness proof (G-1)**: temporarily replaced the bridge test harness's real `resolveSheetCapabilitiesForUser` call with a hardcoded permissive grant (the exact pattern the *other*, unrelated `yjs-scalar-flush-realdb.test.ts` uses for its own value-fidelity isolation) — **7 of 12 assertions flipped from pass to fail** (A3, A5-not-own, A6, both M2 sub-tests, both M4 sub-tests), while 5 correctly stayed green (the sentinel, A4/A5-own — already-legitimate sheet-scope grants, M3 — the independent lock gate, and the new masked-target test — the independent field-guard gate). Reverted before commit; re-ran clean on a rebuilt fresh DB to rule out state bleed: 19/19 green again.
- `tsc --noEmit`: clean, exit 0.
- No cluster was dropped as redundant — the spec's own inventory (§1) confirms none of G-1/G-2/G-3 were previously golden-locked.

## Not in this PR

No runtime code changes. No fix for the adjacent per-subject field-permission finding above. Do not merge — opening for review per the requested workflow.

### DIFF
--- a/packages/core-backend/tests/integration/multitable-permmatrix-b1-basewrite-no-sheet-write-realdb.test.ts
+++ b/packages/core-backend/tests/integration/multitable-permmatrix-b1-basewrite-no-sheet-write-realdb.test.ts
@@ -1,0 +1,118 @@
+/**
+ * B1-G3 — permission golden matrix: base-level write codes do NOT intersect the sheet/record
+ * write axis (real DB). Spec: W1-2 permission-matrix (docs/development/multitable-w1-2-permission-
+ * matrix-spec-20260705.md) §3 gap G-3 / §2 axis M6. Sibling: #3646 (W1-1 formula-freshness
+ * design-lock) locks the FRESHNESS angle of the Yjs-bridge side-door; this B1 batch (G-1/G-2/G-3)
+ * locks the PERMISSION angle across the side-door / write-modifier surface more broadly — this
+ * cluster in particular pins the C2 crux fact rather than a bridge/token entry point.
+ *
+ * `resolveBaseWritable` (permission-service.ts:1599-1645) is a SEPARATE, kernel-internal primitive
+ * introduced for the (b) cross-base automation slice: it grants base-level write authority via
+ * `BASE_WRITE_PERMISSION_CODES` (`multitable:base:write` / `multitable:base:admin` / `multitable:admin`,
+ * permission-service.ts:145-148) OR via `meta_bases.owner_id` ownership. It is consulted ONLY by the
+ * automation executor's base-scoped write path — it is NEVER called by the interactive record PATCH
+ * route. `deriveCapabilities` (sheet-capabilities.ts:75-100 + the sheet-scope composition at :122-158),
+ * which DOES gate interactive PATCH via `capabilities.canEditRecord`, has no knowledge of
+ * `multitable:base:*` codes or of `meta_bases.owner_id` at all — verified directly: `hasPermission`
+ * (sheet-capabilities.ts:69-73) does an exact/wildcard string match against the QUERIED code
+ * (`'multitable:write'`), and the literal string `'multitable:base:write'` satisfies neither that nor
+ * any `SHEET_*_PERMISSION_CODES` set membership.
+ *
+ * This is the "M6 crux" documented in the spec but never golden-locked: an actor who holds ONLY a
+ * base-level write grant (§1) or IS the base's owner (§2) — with NO sheet-scope grant and NO global
+ * `multitable:write` — must get 403 on an interactive record PATCH. A regression here would be a
+ * SILENT privilege escalation (a base-write/base-owner actor suddenly able to edit every record in
+ * every sheet of that base) with no code path announcing the change.
+ *
+ * Runs only with DATABASE_URL (sentinel fails-not-skips in CI per the suite convention).
+ */
+import express, { type Express } from 'express'
+import request from 'supertest'
+import { afterAll, beforeAll, describe, expect, test } from 'vitest'
+
+import { poolManager } from '../../src/integration/db/connection-pool'
+import { univerMetaRouter } from '../../src/routes/univer-meta'
+
+const describeIfDatabase = process.env.DATABASE_URL ? describe : describe.skip
+
+const TS = Date.now()
+const BASE_ID = `base_b1g3_${TS}`
+const SHEET_ID = `sheet_b1g3_${TS}`
+const FLD_NAME = `fld_b1g3_name_${TS}`
+const REC_ID = `rec_b1g3_${TS}`
+
+// §1 — holds ONLY the base-level write code (no sheet grant, no global multitable:write).
+const BASE_WRITE_CODE_ACTOR = `u_b1g3_basewritecode_${TS}`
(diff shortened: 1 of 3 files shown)


zensgit:
Closing as a **duplicate** of #3649 (a second parallel build session opened this for the same W1-2 B1 slice). #3649 is the survivor: it has the CI wiring (plugin-tests.yml), the spec doc, and the ledger update that this PR lacks. **Salvaged from here into #3649** (`7e9fda556`): this PR's superset test files, including the net-new G-1 masked-target field-guard golden. Also carried forward: the adjacent runtime finding this session surfaced (per-subject `field_permissions.read_only` write-gating enforced only at 3 route call sites, not in `patchRecords`/`patchRecord`) — I'm reporting that to the owner as a separate candidate-defect item, not folding it into the test PR. Nothing lost.

PR #3653, opened 2026-07-05
docs(multitable): W1-3 per-subject field-write gate — DESIGN-LOCK (PROPOSED)

zensgit:
**PROPOSED — awaiting your ratify OR close. Docs-only; no runtime ships here.**

Converts the W1-3 pre-draft into a reviewable design-lock — the concrete form of the opt-in decision I offered. **Ratify** = greenlight the runtime fix (a separate future PR); **close** = archive the finding. Zero-cost either way; nothing runtime changed by this doc.

## The gap (verified at line level, independently re-confirmed)
The write spines (`patchRecords` / `patchRecord`) are property-level only **by design**; each write ROUTE applies the layer-3 per-subject `field_permissions.read_only` gate itself. **Six** route sites do (grid /patch + restore-family + revert-execute). **Two paths skip it**:
1. **Yjs collaborative-edit bridge** (`index.ts` write-input builder → `yjs-record-bridge.ts:226`) — no fieldPermissions, no layer-3 gate.
2. **Single-record OAPI `PATCH /records/:recordId`** (`univer-meta.ts:13673`) — coarse `canEditRecord` then `recordService.patchRecord`, no layer-3 gate. (The #3649 build flagged this 'to be checked'; this lock confirmed it's a genuine 2nd uncovered path.)

So a field set read-only for a specific actor via `field_permissions` is blocked on the everyday grid PATCH but bypassable by that actor through the **collab channel** OR the **token write channel**. Structural read-only (formula/lookup/rollup/mirror) stays blocked (property-level); only the per-subject layer leaks.

## What it locks
Gate added to the two paths, **never the spine** (F1); predicate-only shared-helper extraction, not the route-shaped flow which differs deliberately to avoid existence oracles (F2); one indexed `field_permissions` read per debounced flush build, acceptable (F3); fail-closed + atomic — whole write rejected, zero side effects (F4); GW1–GW7 fail-first real-DB matrix incl. **GW7** a durable structural guard enumerating every write path against a frozen has-a-layer-3-gate allowlist (same discipline as the taint-chokepoint guard). **Fail-closed by construction**: can only ADD denials for the exact per-subject-restricted case the six routes already deny — cannot newly block a normal field.

## Non-goals
No `field_permissions` model change; no spine change (property-level posture is the intentional contract); no behavior change to the six working gates; not Yjs GA; not W1-1 recompute / W1-2 capability (siblings).

Sibling arc: W1-1 freshness (merged `26af7a560`) · W1-2 B1 capability (#3649).


zensgit:
Codex review verdict: REQUEST CHANGES.

Medium — the bridge denial outcome in F4 does not match the current Yjs bridge metrics contract. The lock says the bridge builder should log coarsely and return null, and that the deny is observable via log + bridge metric. But today `executePatch()` treats null input as `return false`; the caller increments `flushSuccessCount` only when `applied` is true and increments `flushFailureCount` only in `.catch(...)`. So a null-return deny would be neither success nor failure; no bridge metric records it. Please either (a) make the lock say the builder throws a coarse permission error for this deny path so `.catch` increments `flushFailureCount`, or (b) explicitly drop the metric claim and make the golden assert no success/no mutation rather than a failure count. My preference is (a), because permission-denied flushes should be visible as failed flushes, while null should stay reserved for unavailable/deleted context.

Low precision fix to fold in at the same time: `PATCH /records/:recordId` is not OAPI-only; `apiTokenAuth` is token-specific, but the route is also the session single-record PATCH path when no `mst_` token is present. Please call it `single-record PATCH route (including OAPI token and session callers)` rather than `single-record OAPI PATCH`, and keep the runtime fix at the handler so both callers are covered.

zensgit:
**Both fixes pushed** (`1464c5d60`) — both verified at line level first:
- **F4 metric contradiction**: confirmed `executePatch` maps a null input → `false` into the `.then(applied=>…)` branch (`yjs-record-bridge.ts:211-218`), which touches NEITHER counter, so a null-return deny is metric-invisible. Fixed to your preferred direction: a per-subject deny **throws a coarse permission error** → routes to `.catch` → `flushFailureCount++` + coarse log. F1-bridge and F4-bridge both updated with the why-throw-not-null rationale.
- **single-record PATCH not OAPI-only**: confirmed `apiTokenAuth` only intercepts `mst_` tokens and `next()`s everything else (`api-token-auth.ts:47-56`), so `PATCH /records/:recordId` serves both token AND session/JWT callers. Retitled 'single-record OAPI PATCH' → 'single-record PATCH route'; §1.2 now states the leak is reachable by any caller (token or session); GW4 asserts BOTH an `mst_` and a session caller.

Awaiting re-ratify.

zensgit:
Codex re-review verdict: PASS / RATIFY on content.

Both requested fixes are resolved on head 1464c5d60: F4 now throws a coarse permission error for bridge per-subject denial so the existing `.catch` path increments `flushFailureCount` instead of silently returning null, and the single-record route is correctly described as `PATCH /records/:recordId` for both `mst_` token and session/JWT callers. GW4 also now asserts both caller shapes. The design-lock is coherent and ready to ratify.

Low tidy before merge: §6 still says `W1-2 B1 side-door capability goldens — PR #3649 (open)`. #3649 has since merged, so update that ledger row to merged/on-main with the squash SHA. Optional same sweep: refresh the stale single-record line anchor in F1 if you want the line-level provenance to stay crisp. No content blocker.

PR #3673, opened 2026-07-06
docs(multitable): S4 AI cost visibility — estimate + per-field dimension (DESIGN-LOCK, PROPOSED)

zensgit:
**PROPOSED — awaiting owner ratification. Docs-only; no runtime ships here.**

Drafts the S4 slice from the arc ledger ("cost visibility polish — estimate UI + per-field/per-run usage dimensions"). Grounded: the AI cost machinery is already built — S4 is a **read/surface slice**, not new accounting.

## Grounding (the crux: what does the ledger already record?)
- `multitable_ai_usage_ledger` rows carry `field_id` (**populated** per shortcut/bulk write, `ai-bulk-shared.ts:105`) + `estimated_cost_usd` → **per-field cost is queryable today by GROUP BY field_id, no schema.**
- The pre-run estimator (`n()`/caps) already runs server-side for the quota pre-check but is **never surfaced** to the user before they spend.
- `GET /ai/usage-summary` (`multitable-ai.ts:237`, `requireAdminRole`) returns aggregate windows only; the FE `AiUsageSummary` type mirrors those four fields; **zero FE consumers** — the cost surface is genuinely unbuilt.
- **The ledger has NO run key** (no `run_id`/`jobId` column) → "cost of THIS run" is NOT answerable without a migration.

## What it locks
- **C1** pre-run estimate `{estTokens, estCostUsd, rowCount}` in the bulk-preview response (same math the quota check uses; estimate only, never billed from).
- **C2** `usage-summary` gains a `byField` aggregate (GROUP BY field_id over existing columns).
- **C3** values-free (aggregate numbers + fieldId only; no prompt/field/record data).
- **§3** keep the existing admin gate for `byField` (don't fragment onto canManageFields — it's a finer cut of the same admin-scoped, values-free aggregate).
- **GS1–GS6** golden matrix (estimate shows pre-run; per-field GROUP BY truth; NULL-field-id excluded; admin gate holds; values-free; scope guard).

## Explicit split (honesty over forcing scope)
**Per-RUN cost is NOT this slice** — the runless ledger means it needs a `run_id` column + backfill (a migration), so it's split to a gated **S4b** with its own design-lock. GS6 pins that S4 surfaces only ledger-recorded dimensions and invents no per-run split.

Sibling arc: S1/S2 merged; S3/S5/S1b gated.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3681, opened 2026-07-06
docs(multitable): S3 AI-shortcut staleness lineage — DESIGN-LOCK (PROPOSED)

zensgit:
**PROPOSED — awaiting ratify. Docs-only design-lock; no runtime here.** Converts the S3 pre-draft to a reviewable artifact (grounding independently spot-checked).

## Concept: flag-don't-auto-spend
An `aiShortcut` output is derived from its `sourceFieldIds`. When a source value changes the output is STALE — but re-running an AI shortcut **costs tokens / hits a provider**. So S3 does the **opposite of W1-1 formula freshness** (which auto-recomputes because formulas are cheap/deterministic): it **flags stale, never auto-recomputes**. Auto-recompute here would be a token-budget footgun — explicitly forbidden (LOCK-C).

## Key decisions
- **LOCK-A** — staleness = server-side `sha256(source values)` fingerprint stored at generation, **derived on read** (stale ⟺ stored hash ≠ current). Self-healing (re-run rewrites hash → clears; source reverted → not stale); no invalidation-cache drift.
- **LOCK-B (the one gated sub-decision)** — ONE narrow table `multitable_ai_shortcut_generation`, **no existing table touched**. Ratify = greenlight this single migration. Verified no zero-storage path exists: `previewVersion` is ephemeral (30-min TTL cache), `formula_dependencies` is formula-only (neither anchors aiShortcut source provenance). If you'd rather not add storage, S3 can't ship as a visibility slice — stated plainly, not smuggled.
- **LOCK-C (anti-footgun, the whole point)** — a source edit **NEVER** calls the provider / writes the usage ledger / spends a token. **GS1 asserts the ledger is byte-untouched** after a source edit = tested invariant. Re-run is explicit + priced (reuses S4 #3673 estimate) + behind `MULTITABLE_AI_CONFIRM_LIVE_REQUESTS`.
- **LOCK-D** — **ZERO write-path fan-out** (staleness read-time-derived; source edits do zero extra write work — low hot-path risk); transitivity free via hash-over-values.
- **LOCK-E** — values-free (`source_hash` server-only; client sees a boolean).
- **GS1–GS7** fail-first real-DB matrix (anti-footgun, values-free, transitivity, regression).

Sibling arc: S1 ✅ · S2 ✅ · S4 #3673 (PROPOSED) · S4b 🔒 (per-run, split honestly) · W1-1 runtime ✅ (#3679) · W1-3 runtime 🔒 (#3676, pending your A/B).

Docs-only. Awaiting ratify (and the §3-B3 storage opt-in) — not merging.


zensgit:
**⚠ Independent adversarial gate: CHANGES_REQUESTED (light) — ratify after LOCK-A pins the hash value-set.** (Gate ≠ ratify.) Architecturally sound: flag-dont-auto-spend is the right inversion of W1-1; it does **NOT** silently re-open the RFC-deferred C1 (LOCK-B is a per-CELL value fingerprint `sha256(source values)`, not the per-FIELD edge graph C1 = — governance-clean, and §1.3/§3-B2 explicitly name why `formula_dependencies` isnt reused). **No missed-staleness (false-negative) path** — every error direction is false-POSITIVE (safe), and LOCK-C forbids silent spend, so no user can unknowingly trust an out-of-date AI value. All 5 load-bearing claims HELD.

**P2 (must pin before ratify) — LOCK-A hash value-set is internally contradictory + reader-dependent.** §2 A1 says the fingerprint is `sha256(current values of the FULL declared sourceFieldIds)` AND, same breath, "the SAME values `assembleMaskedPrompt` assembled = its **readable set**." Those are different bases. `assembleMaskedPrompt` (`multitable-ai.ts:1719-1733`) intersects `sourceFieldIds` with the actor’s `readableEchoFields`, dropping unreadable sources → "readable set" is **per-actor**. Concrete failure: generator R1 reads {A,B,C}, hash stored over {A,B,C}; reader R2 is denied C → if read-time uses R2’s readable set {A,B}, `sha256({A,B}) ≠ sha256({A,B,C})` **forever** → R2 sees a permanent false-positive `stale` badge on an unchanged output (tempting a wasted priced re-run), and it contradicts E1’s "server-side, reader-independent" framing. **Fix:** pin the fingerprint to the FULL declared `sourceFieldIds` with **server-authoritative values**, computed identically at generation AND read (reader-independent, per-cell-stable); drop the "readable set" gloss in A1+A2.

**P3 (fold into lock text):** (a) full-set hashing gives a reader denied a source a 1-bit change-oracle ("that source changed") when the badge flips — acknowledge in E2’s leak envelope or scope the badge to the reader’s readable subset; (b) the hash covers source VALUES only — editing the shortcut’s own params/kind leaves cells stale-vs-new-config unflagged (adjacent to S2); state S3 covers source-value staleness only; (c) require the hash upsert to be atomic with the value `patchRecords` write (a crash between them → permanent not-stale via A3 fail-open). NIT: "lineage" in the title is a mild misnomer (it’s a fingerprint, not stored edges — which is exactly why it’s not C1). Hard rules PASS. Report `/tmp/ai-s3-staleness-3681-gate-20260708.md`. Still PROPOSED — do not merge as ratification.

zensgit:
**Gate remediation pushed (`d8a9bb37c`) — remediation only, still PROPOSED, awaiting re-gate + owner ratify. Nothing merged/ratified here.** Verdict-item → fix mapping:

- **P2 (LOCK-A hash value-set contradictory/reader-dependent) → PINNED as an explicit enumeration.** New §2 A1a–A1e: basis = the **FULL declared `aiShortcut.sourceFieldIds`** in persisted declared order (validator dedupes order-preserving, `ai-shortcut-config.ts:74-79`), values = **server-authoritative** `meta_records.data[sourceFieldId]` via `normalizeJson` (`multitable-ai.ts:1708`), canonical form pinned (per-field `fieldId + '=' + canonicalJson(value)`, keys sorted recursively, absent→`null`, NUL-joined, lowercase-hex SHA-256), ONE shared helper at both write sites AND the read derivation. The `assembleMaskedPrompt` "readable set" gloss is **dropped** from A1/A2 — A1a states explicitly why hashing that per-actor subset (`multitable-ai.ts:1719-1733`) is forbidden (your R1/R2 permanent-false-stale scenario, quoted as the counterexample). A2 now says reader-independent outright. A1e records the honest residual: full-set basis vs a masked run's subset prompt (`masked = readableSourceCount < configSourceCount`, `:848-849`, `:937-938`) can over-flag — false-POSITIVE direction only, accepted.
- **P3a (1-bit change-oracle) → folded into §6 E2's leak envelope**: a reader denied a declared source can see the badge flip = 1 bit "some declared source changed", never which field nor any value; per-reader badge scoping explicitly REJECTED (it would recreate the per-actor basis P2 removed).
- **P3b (config-staleness unflagged) → scope pinned** in the What-S3-is-NOT list: fingerprint covers source VALUES only; shortcut `kind`/`params` edits don't flag stale-vs-new-config (S2 prompt-config-history territory); S3 = source-value staleness only.
- **P3c (crash window → permanent not-stale via A3) → new §2 A4**: `sourceHash` upsert MUST commit in the SAME transaction as the `patchRecords` value write; post-commit best-effort upsert forbidden, hash-before-value equally forbidden (409'd value write would strand a hash for a generation that never landed).
- **NIT ("lineage" misnomer) →** clarified in the NOT-list: lineage here = per-cell generated-against value fingerprint, no dependency-edge graph introduced. (Filename/title kept to preserve the PR artifact path.)
- **Extra pin**: GS8 golden added — two readers (full-read vs denied-one-source) derive the SAME `stale` boolean for the same cell.

All new `file:line` cites verified against `origin/main` @ `e837c508f`. Status header untouched (PROPOSED). Docs-only diff: 1 file, +12/−5.

zensgit:
## 独立重闸裁决(修复后)— MERGE_CLEAN ✅

对 d8a9bb37c 的 refute-first 重审(报告 `/tmp/pr3681-regate-20260709.md`;全部引用对 origin/main @ 4820c3c3e 逐条 EXACT):

| 原缺陷 | 状态 | 关键证据 |
|---|---|---|
| P2 hash 基集矛盾/reader 依赖 | **RESOLVED** | A1a–A1e 钉死「全声明 sourceFieldIds × server-authoritative 值 × 单一 helper(写=读逐字节同源)」;决定性 grounding:写点候选数据来自**未掩码**的 readRecordOnce(:1695)→ 全集哈希在写点可实现且 reader-independent;构造 R1/R2 denied-source 输入两侧同 `stale:false`,被拒读者不再永久 false-stale;false-negative hunt 通过(任何入 prompt 字段皆声明源→值变必翻哈希) |
| P3a 1-bit oracle | RESOLVED | E2 承认信号存在;拒 per-reader scoping 论证成立(会重造被 P2 移除的 per-actor 基集)——在原裁决明列选项内 |
| P3b config 范围 | RESOLVED | 指纹=source VALUES only;kind/params=S2 领地,边界干净 |
| P3c 原子性 | RESOLVED | A4 双向禁(post-commit→经 A3 永久 not-stale;hash-before-value→expectedVersion stale-drop 悬挂哈希,:525/:1180+:1339/:1600 有码可依) |
| NIT lineage 误称 | RESOLVED | lineage=值指纹,无边图,强化 C1 不重开 |

新增 GS8 golden(全读 vs 单源被拒读者同 stale 布尔)使 reader-independence 可测。1 个非阻塞 NIT:§1 既有 formula_dependencies 行号随 main 漂移(修复未触碰区,实质架构论断仍真,可后补刷新)。

**结论:P1/P2/P3 清零,本锁进入 owner ratify 队列就绪状态。仍为 PROPOSED——merge 即 ratify,由 owner 决定。**

PR #3686, opened 2026-07-06
test(multitable): W1-2 permission-matrix B4 — export⊆read differential + comments×visibility (G-7, G-8)

zensgit:
## Summary

Final batch (B4) of the W1-2 permission golden matrix — the two **differential properties** left in
`docs/development/multitable-w1-2-permission-matrix-spec-20260705.md` §3: **G-7** (export ⊆ interactive
read, same actor) and **G-8** (comments × sheet-visibility). Continues B1 (#3649, merged) + B2 (#3682,
merged). **Note:** B3 (#3685) is still **open** (not merged) as of this PR — B4's scope (G-7/G-8) doesn't
depend on B3's content (G-6, management-surface 403 matrix), so it was built independently off
`origin/main` per instructions; flagging so "B1–B4 complete" isn't assumed until B3 also lands.

TEST-ONLY. No `src/` changes. Do not merge — requesting review/merge by the repo owner.

## G-7 — export ⊆ interactive-read (same actor) — GREEN, wired into CI

`packages/core-backend/tests/integration/multitable-permmatrix-b4-g7-export-subset-read-realdb.test.ts`
**6/6 tests pass.**

Property: for the same actor, `export-xlsx` must never reveal more than that actor's `GET /records`
output — every field masked/absent on read stays masked/absent on export; every row-denied record absent
from read stays absent from export; a tainted formula column (M7, formula-over-masked-cross-base-lookup)
masked on read stays masked on export. `GET /records` and export's no-view branch share the identical
chokepoints (`deriveFieldPermissions` → `maskStoredRecordFieldIds` for M1+M7; `loadRowLevelReadDenyEnabled`
+ `loadDeniedRecordIds` for M2), which is why a differential between them is meaningful rather than two
independently-written masks that happen to agree.

Tests: sentinel · non-vacuous ALLOW-actor control (proves the fixture is genuinely readable/exportable,
not just two empty sets) · **G-7a** M1 field-mask differential · **G-7b** M2 row-deny differential ·
**G-7c** (headline) M1+M2+M7 combined in one fixture, generic subset assertion over dynamic
field-id/row-key sets (not hardcoded arrays) · admin-bypass control (M2's explicit `isAdminRole` guard).

**Non-vacuousness verified**: temporarily injected a phantom leaked field-id and row-key into the
export-side projection in G-7c and confirmed `assertSubset` flips RED (`expected [ 'MUTATION_PHANTOM...' ]
to deeply equal []`), then reverted. Diffed the reverted file back to the pre-mutation copy — byte-identical.

**Redundant-drops** (not re-asserted — already locked elsewhere):
- Export masks M1 field_permissions in isolation → `multitable-permission-golden-d3d1.test.ts`,
  `multitable-export-allrows-maskroute-realdb.test.ts`.
- Export masks M7 taint (whole formula column dropped) → `multitable-lookup-foreign-field-mask-export.test.ts`.
- Export excludes an M2 row-denied record (zero presence) → `multitable-permmatrix-b2-g5-collection-noleak-realdb.test.ts`.

None of those assert the **cross-surface** relation (export's masked output ⊆ that same actor's
interactive-read masked output) — that's the actual gap this file closes.

## G-8 — comments × sheet-visibility — CONFIRMED REAL LEAK, NOT wired into CI

`packages/core-backend/tests/integration/multitable-permmatrix-b4-g8-comments-visibility-realdb.test.ts`
**6 tests: 2 pass (sentinel + control), 4 RED against unmodified `main`.**

This is a genuine, reproducible finding, not a test bug — per the task contract ("if a differential
golden goes RED against current main... STOP and report as a candidate runtime finding, do NOT weaken the
golden, do NOT fix runtime"), I stopped there: the assertions encode the correct/intended contract
verbatim, `src/` is untouched, and **this file is deliberately NOT added to
`.github/workflows/plugin-tests.yml`'s real-DB allowlist** — wiring a confirmed-red suite into that shared
step would turn CI permanently red for every unrelated future PR. The file's header carries the same
notice in detail.

### The finding

`routes/comments.ts` gates every route through `rbacGuard('comments', 'read'|'write')` — a **global** RBAC
permission-code check (`userHasPermission(userId, 'comments:read')`) that takes no `sheetId`/
`spreadsheetId` parameter and structurally cannot be sheet-scoped. `CommentService.getComments` /
`getCommentPresenceSummary` (`services/CommentService.ts`) then query `meta_comments` purely by
`spreadsheet_id` (+ optional rowId/fieldId/resolved) — **there is no call anywhere in the comments code
path to `resolveSheetReadableCapabilities` or any other multitable per-sheet capability/ACL check.**

Contrast with `multitable/sheet-capabilities.ts`'s `deriveCapabilities`, where `canRead` (multitable) and
`canComment` (comments) are computed as **independent** booleans from independent permission codes. Even
more tellingly, `permission-service.ts`'s `applySheetPermissionScope` (~line 1249) already encodes the
*intended* per-sheet contract — `canComment: capabilities.canComment && scope.canRead` for sheets with
explicit ACL assignments — but that composed value is only a UI-capability hint surfaced in `GET /view`'s
response payload; `routes/comments.ts` never consults it. **The result: holding `comments:read` /
`comments:write` acts as a system-wide bypass of multitable sheet visibility.**

**Not theoretical** — this exact actor shape is a shipped product role:
- `auth/access-presets.ts`'s `plm-collaborator` preset ("PLM 协作成员" / PLM Collaboration Member —
  "适用于查看 PLM 工作台、审批与评论协作") grants exactly
  `['spreadsheets:read', 'workflow:read', 'approvals:read', 'comments:read']` — comments:read with **no**
  `multitable:*` permission at all.
- Migration `zzzz20260320163000_add_comment_permissions.ts` seeds **both** `comments:read` and
  `comments:write` onto the generic `'user'` role.

### Empirically confirmed (fresh DB, unmodified `main` @ `1db165fe5`)

Actor: `comments:read`/`comments:write` RBAC only, zero `multitable:*`, zero sheet-scoped grant, on a
private sheet with one real comment on it.

| Request (same actor) | Result |
|---|---|
| `GET /api/multitable/records?sheetId=<sheet>` (interactive read, control) | **403** (correctly denied) |
| `GET /api/comments?spreadsheetId=<sheet>` | **200** + full comment content, rowId, fieldId |
| `GET /api/comments/summary?spreadsheetId=<sheet>` | **200** + real counts/rowIds |
| `GET /api/multitable/:id/comments/presence` | **200** + real counts/rowIds |
| `POST /api/comments {spreadsheetId:<sheet>, rowId, fieldId, content}` | **201 Created** (unauthorized write succeeded) |

The 4 corresponding goldens (G-8a/b/c/d) are RED for exactly this reason; the sentinel and the "actor is
genuinely 403'd on interactive read" control both pass, so the RED isn't a fixture/harness bug.

### Suggested next step (not undertaken here — runtime is out of scope for this PR)

Gate `routes/comments.ts`'s read/write routes on `resolveSheetReadableCapabilities` (or an equivalent
per-sheet check) for the resolved `spreadsheetId`, in addition to the existing global `rbacGuard`, before
touching `CommentService`. Once fixed, this file's 4 red tests should pass unmodified — flip the file's
header notice and add it to `plugin-tests.yml`'s allowlist in the same follow-up PR. Alternatively, if the
owner decides `comments:read`/`:write` is *intentionally* a global, sheet-visibility-independent axis, the
CONTRACT (this file **and** the spec's G-8 entry) needs to be revised together, in the open — not silently.

## Verification (real numbers)

- Fresh dedicated DB: `createdb metasheet_w12b4_test`; `pnpm run migrate` (clean, all migrations applied);
  single run per file, rerun together for cross-file collision check.
- G-7: 6/6 green in isolation and when run alongside G-8 + the 9 existing sibling permmatrix/export/oapi
  real-DB files (no interference, no shared-state collisions).
- G-8: 2 pass / 4 fail, identically reproducible across 3 separate runs (isolated, combined-with-G-7,
  combined-with-siblings).
- `tsc --noEmit`: clean. (The repo's `tsconfig.json` structurally excludes `**/*.test.ts`, so I verified
  with an ad hoc project config — original compiler options, `**/*.test.ts`/`__tests__` exclusions intact,
  plus explicit `include` entries for these 2 new files only — 0 errors. Also reran the plain
  `pnpm --filter @metasheet/core-backend type-check` baseline: clean.)
- Zero orphaned rows: post-run id-prefix count query (`base_b4g%`, `sheet_b4g%`, `fld_b4g%`, `rec_b4g%`,
  `cmt_b4g%`, `record_permissions`/`field_permissions`/`meta_links`/`formula_dependencies` by sheet,
  `user_b4g%`) across all 3 test runs — all zero.
- Workflow YAML parses (`yaml.safe_load`); the giant real-DB step's descriptive name updated to mention
  the new G-7 coverage, matching the B1/B2 precedent.

## This completes the W1-2 test-matrix build-out

B1 (#3649, merged) + B2 (#3682, merged) + B4 (this PR) are done; **B3 (#3685) is still open** — once it
merges, all of §3's G-1..G-8 gaps identified in the spec are golden-locked (G-1/G-2/G-3 → B1; G-4/G-5 →
B2; G-6 → B3; G-7/G-8 → B4), except G-8's runtime gap, which is now a tracked, documented finding rather
than a silent hole.


zensgit:
Closing in favor of #3688, a clean G-7-only rebuild off current main.

Reason for the split:
1. This branch was cut before B3 (#3685) landed, so it removed B3's `multitable-permmatrix-b3-mgmt-403-matrix-realdb.test.ts` from the CI allowlist — a naive merge would have regressed B3 out of CI. #3688 rebuilds off current main with B3 intact.
2. The G-8 (comments × sheet-visibility) golden in this branch is a **confirmed real runtime security leak** against main (independently reproduced this session, 4 RED / 2 GREEN on a fresh DB), not a mergeable test. It is surfaced separately as an owner-gated security finding and must not enter the tree/CI as a permanent red until the runtime gap is triaged.

G-7 (the green, correct-behavior differential) lives on in #3688.

PR #3693, opened 2026-07-06
feat(approval): CFG-1 card-config resolvers — env-first, stored-encrypted fallback (one-tap config lock)

zensgit:
Implements **CFG-1** of the card-config settings design-lock `docs/development/approval-card-config-settings-ui-design-lock-20260706.md` (#3690, **PROPOSED**) — merging this PR = ratifying the lock's direction for the resolver layer.

## What
- **New `integrations/dingtalk/approval-card-config.ts`**: `resolveApprovalCardLinkSecret()` + `resolveApprovalCardPublicAppUrl()` — **env FIRST** (operator override unchanged), then stored `directory_integrations.config` fallback (`approvalCardLinkSecret` decrypted via `security/encrypted-secrets` — same path as appSecret/workNotificationAgentId; `approvalCardPublicAppUrl` plaintext), else **fail-closed empty**. Store/decrypt failures resolve empty — never a fallback secret.
- **Two read-point switches** (the only as-built `process.env.APPROVAL_CARD_LINK_SECRET` readers):
  - `automation-executor.ts` `executeSendDingTalkApprovalCard` — sign side
  - `ApprovalCardDeliveryAction.ts` `verifyApprovalCardLinkToken` — verify side, now async with `deps.query` passthrough
  - **Same-source invariant**: both resolve env-first-else-the-same-stored-row → a token signed on send always verifies on decision.
- **Behavior with env set is byte-identical** (env short-circuits before any query) — the ratified #3594 chain is untouched; this only adds the stored fallback CFG-2/CFG-3 will write into.

## Verification
- Unit 9/9 (`tests/unit/dingtalk-approval-card-config.test.ts`): env priority skips store · stored decrypt · missing row/key + query failure + undecryptable value all fail-closed · **empty-key forgery rejected** · sign/verify same-source round-trip · URL env order/trailing-slash/stored fallback.
- Real-DB (fresh `ms2_cfg1_verify` DB, migrations from zero): one-tap set **18/18** — A-2b action 5/5 + A-4 wrapper 8/8 (incl. **new** stored-secret fallback round-trip: env unset → wrapper verifies stored-signed token, rejects env-signed token; fixture removed in finally per shared-DB discipline) + task_created trigger 5/5.
- `tsc --noEmit` clean. No new CI wiring needed: unit file rides the default config; the wrapper file is already in plugin-tests.yml.

## Out of scope (lock checklist)
- CFG-2 generate endpoint (admin-gated, `{configured:true}` never echoes the value)
- CFG-3 DirectoryManagementView UI (generate button + chip + regenerate confirm + URL field)

### DIFF
--- a/packages/core-backend/src/integrations/dingtalk/approval-card-config.ts
+++ b/packages/core-backend/src/integrations/dingtalk/approval-card-config.ts
@@ -1,0 +1,93 @@
+/**
+ * CFG-1 (card-config settings lock, §3.1): runtime resolvers for the one-tap approval-card
+ * configuration — the deep-link HMAC signing secret and the public app URL the links point at.
+ *
+ * Resolution order (mirrors readDingTalkMessageConfigFromRuntime): env FIRST, stored
+ * directory_integrations.config fallback second. The stored secret is written through
+ * security/encrypted-secrets (same path as appSecret / workNotificationAgentId) and is only
+ * ever decrypted here, server-side.
+ *
+ * Invariants:
+ * - SAME-SOURCE: the card sender (sign) and the card-delivery wrapper (verify) must both call
+ *   these resolvers so a token signed on send always verifies on decision — guaranteed by
+ *   "env first, else the one stored dingtalk integration".
+ * - FAIL-CLOSED: any miss (no env, no row, decrypt failure, query failure) resolves to '' —
+ *   the card action refuses to send and the wrapper refuses to verify. Never a fallback secret.
+ */
+import { query } from '../../db/pg'
+import { decryptStoredSecretValue } from '../../security/encrypted-secrets'
+
+const DEFAULT_PROVIDER = 'dingtalk'
+export const APPROVAL_CARD_LINK_SECRET_CONFIG_KEY = 'approvalCardLinkSecret'
+export const APPROVAL_CARD_PUBLIC_APP_URL_CONFIG_KEY = 'approvalCardPublicAppUrl'
+
+type QueryFn = (sql: string, params?: unknown[]) => Promise<{ rows: unknown[] }>
+
+type JsonRecord = Record<string, unknown>
+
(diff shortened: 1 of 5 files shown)


zensgit:
## Adversarial review (Opus, refute-first) — APPROVE-with-hardening

Full review: `/tmp/pr3693-review-claude-20260706.md` (local). **No P1/P2 code defect.** All attacked invariants held:

- **SAME-SOURCE**: sign and verify share one process/pool, identical env-first + identical `ORDER BY` stored pick — no single-instant divergence path found.
- **No auth bypass from sync→async**: `verifyApprovalCardLinkToken` has exactly 2 callers, both awaited; routes only call the two wrappers; nothing else calls it.
- **Fail-closed mutation-proven**: dropping `if (!secret) return false` → empty-key-forgery test RED; making the catch return a constant → fail-closed test RED (both mutations reverted, tree clean).
- **Byte-identical with env set**: env short-circuits before any query; no test anywhere asserts the changed error strings; ratified #3594 chain intact.
- Re-ran tests on real PG: unit 9/9 · wrapper DB 8/8 (incl. stored-secret fallback round-trip) · sign-side action 5/5.

**Findings & disposition:**
- **P2 (governance)**: the governing design-lock (#3690) is PROPOSED/unmerged — **merge of this PR is gated on owner ratify** (this PR's body already states merging = ratifying the resolver direction). No code change.
- **P3 → CFG-2**: enforce `normalizeStoredSecretValue` at the generate-endpoint write (plaintext passthrough parity noted); add http(s) scheme allowlist for `approvalCardPublicAppUrl` + re-verify admin gate.
- **NIT → fixed in `032a77b2a`**: stored-secret test fixture now uses `now()` instead of a far-future timestamp, so a crash-skipped finally cannot permanently shadow real integrations.

Awaiting owner ratify of the #3690 lock before merge.

PR #3698, opened 2026-07-06
feat(approval): CFG-2 card-config self-service — server-side secret generate + URL save (admin-gated, no echo)

zensgit:
Implements **CFG-2** of the card-config settings design-lock (#3690, ratified; CFG-1 = #3693 merged).

## What
- **`generateApprovalCardLinkSecret`**: `crypto.randomBytes(32)` **server-side**, stored via `normalizeStoredSecretValue` (encrypted-at-rest, same path as appSecret — review P3 enforced at the write). The plaintext never leaves the function; every response carries presence booleans only (`valuePrinted: false`).
- **`saveApprovalCardPublicAppUrl`**: absolute **http(s)-only** scheme allowlist at the write (review P3 — the stored URL steers where approval deep links point); empty clears the key.
- **`getApprovalCardConfigStatus`**: presence + source (`env|stored|missing`) + `envOverrideActive` flags so the UI can warn that env wins at runtime.
- **Routes** (`admin-directory`, all `ensurePlatformAdmin`-gated with redacted audit entries): `GET/PUT /integrations/:id/approval-card-config`, `POST .../secret/generate`.
- **WIPE-TRAP FIX**: `updateDirectoryIntegration` rebuilds the config JSONB from a whitelist — every generic integration-form save would silently drop the stored secret and kill all in-flight card links. Added carry-through (encrypted value carried as-is, never decrypted). **Mutation-proven**: removing the carry-through turns the new real-DB regression RED.
- Integration summaries now expose `approvalCardLinkSecretConfigured` + `approvalCardPublicAppUrl` (presence/plain, mirrors `appSecretConfigured`) for the CFG-3 chip.

## Verification
- Unit **17/17** (encrypted write, no 64-hex leak in any response, scheme allowlist incl. `javascript:` rejection before any DB access, env-override reporting)
- Route gates **31/31** (403 non-admin on all three routes, 404 unknown, redacted audit, 400 invalid URL)
- Real-DB **9/9** on fresh DB incl. **CFG-2 closed loop** (generate → resolver → sign → wrapper verify) + wipe regression; A-2b sign-side 5/5 unchanged
- `tsc --noEmit` 0 errors

## Out of scope
- CFG-3 settings UI (generate button + chip + regenerate confirm + URL field) — next slice, Sonnet agent per model split.

### DIFF
--- a/packages/core-backend/src/directory/directory-sync.ts
+++ b/packages/core-backend/src/directory/directory-sync.ts
@@ -277,6 +277,8 @@ export type DirectoryIntegrationSummary = {
     appKey: string
     appSecretConfigured: boolean
     workNotificationAgentIdConfigured: boolean
+    approvalCardLinkSecretConfigured: boolean
+    approvalCardPublicAppUrl: string | null
     rootDepartmentId: string
     baseUrl: string | null
     pageSize: number
@@ -976,6 +978,8 @@ function parseIntegrationConfig(row: Pick<DirectoryIntegrationRow, 'config'>): D
 
 function summarizeIntegration(row: DirectoryIntegrationRow): DirectoryIntegrationSummary {
   const config = parseIntegrationConfig(row)
+  // Approval-card keys are presence-only in summaries (secret never decrypted here).
+  const rawConfig = parseJsonRecord(row.config)
   return {
     id: row.id,
     orgId: row.org_id,
@@ -995,6 +999,8 @@ function summarizeIntegration(row: DirectoryIntegrationRow): DirectoryIntegratio
       appKey: config.appKey,
       appSecretConfigured: Boolean(config.appSecret),
       workNotificationAgentIdConfigured: Boolean(config.workNotificationAgentId),
+      approvalCardLinkSecretConfigured: Boolean(normalizeText(rawConfig.approvalCardLinkSecret)),
+      approvalCardPublicAppUrl: normalizeText(rawConfig.approvalCardPublicAppUrl) || null,
       rootDepartmentId: config.rootDepartmentId,
       baseUrl: config.baseUrl ?? null,
       pageSize: config.pageSize,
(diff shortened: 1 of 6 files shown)


zensgit:
## Adversarial review (Opus, refute-first) — APPROVE

Full review: `/tmp/pr3698-review-claude-20260706.md` (local). **No P1/P2/P3.** All attacked surfaces held:

- **Secret leakage**: none — `buildStatus` returns booleans + plain URL only; encrypted blob never decrypted in status/summary/audit; error messages static; GET-list summary emits presence booleans matching the `appSecretConfigured` pattern.
- **Authz**: all three routes gate on `ensurePlatformAdmin` before any service call; the three service functions have exactly one caller each.
- **Carry-through**: cleared URL stays absent everywhere (no resurrection); encrypted secret carried as-is; create path correctly needs nothing.
- **URL allowlist**: `javascript:`/`ftp:`/relative/protocol-relative all rejected; SSRF genuinely out of scope (deep-link base, not fetched server-side, admin-gated).
- **jsonb_set**: paths from module constants, everything else parameterized.
- **Rotation**: no memoization anywhere — sign/verify both resolve fresh; rotation atomic, in-flight links fail-closed (intended semantics, UI confirms before regenerate).
- **Mutation 3/3 RED**: scheme allowlist · admin early-return · carry-through key — each removal turns a test red; reverted, tree clean.
- Tests re-run locally: 17+31(now 32) unit · 9 real-DB (closed loop + wipe regression) · 5 neighbors — green.

**NIT fixed in `e835d398b`**: PUT missing `publicAppUrl` now 400s instead of silently clearing (explicit `""` required to clear). Remaining NITs (400-vs-500 on infra error, raw-vs-canonical stored URL, illustrative route-path difference vs lock) documented as non-blocking.

Queued behind #3696/#3697 in the serial lander.

PR #3732, opened 2026-07-06
fix(multitable): W1-2 B4 G-8 — enforce per-sheet read on comments surface (close cross-sheet visibility leak, decision A)

zensgit:
## What & why
Closes the confirmed **G-8 cross-sheet comment-visibility leak** (owner **decision A**: enforce sheet-read on the comments surface). Previously the comments routes were guarded only by the coarse global `rbacGuard('comments', read|write)` (no sheetId) + `CommentService` queried purely by `spreadsheet_id`, so any holder of `comments:read`/`comments:write` — a shipped role shape (`plm-collaborator` preset; generic `user` via the comment-permissions migration) — could read private comment content and write comments on a sheet the interactive multitable read path 403s them from.

## Fix
`routes/comments.ts` gains `ensureSheetReadable(req, res, spreadsheetId)` — reuses the SAME `resolveSheetReadableCapabilities` chokepoint the record read path uses; the global `comments:*` code supplies the `canComment` half (rbacGuard), this adds the missing per-sheet `canRead` half (the codebase's own intended contract, `permission-service.ts` `applySheetPermissionScope`: `canComment && canRead`). Gates the **9 spreadsheetId-taking routes** (list / summary / presence / mention-candidates / mention-summary read + create / mark-read / mark-all-read write). Denied → 403, same shape as `GET /records` on a no-read sheet (no new existence oracle).

## Golden
`multitable-permmatrix-b4-g8-comments-visibility-realdb.test.ts` — the differential authored as the original RED finding, now GREEN against the gate (6/6): the no-read `comments:*` actor is 403'd across read/summary/presence + create write, response carries no content/rowId/fieldId; control (interactive `GET /records` → 403) proves a genuine no-read sheet. Wired into the real-DB allowlist so it guards against regression.

## Verify
Backend type-check clean. Golden **6/6 green** on a fresh migrated Postgres (was 4/5 RED against pre-fix main @ 1db165fe5).

## Residual (documented, NOT in this PR — distinct mechanism)
`:commentId`-addressed mutations (need per-comment sheet lookup) + user-scoped `inbox`/`unread-count` cross-sheet aggregates (need readable-sheet filtering). Follow-up.


zensgit:
⚠️ **Converted to DRAFT — needs an owner product/design decision before merge.**

CI `test (20.x)` is RED: this G-8 sheet-read gate (decision A) correctly extends to the **OAPI comment-token path**, but that path is design-locked (`multitable-openapi-token-auth-designlock-20260619.md` §4, `min(token scope, creator RBAC)` — no sheet-read term). The gate adds a sheet-read term, so 4 design-locked goldens now fail:
- `multitable-oapi1-comments-read-realdb.test.ts` ×3 (`/api/comments`, `/summary`, `/presence` → 403, expected 200)
- `multitable-oapi2a-comments-write-realdb.test.ts` ×1 (`POST /api/comments` → 403, expected 2xx)

These assert the OLD contract (a `comments:read` token whose creator has NO sheet-read → 200), i.e. the **token-flavored version of the exact leak G-8 closes**. So decision A arguably extends here too — but that's a **behavior change to a shipped, design-locked feature** and I will **not** rewrite a design-lock's goldens autonomously.

**Decision needed** — (A′, recommended) extend decision A to tokens: update the OAPI-1/2a goldens (grant creator sheet-read → 200; add no-read → 403) + revise §4 to `min(scope, RBAC, per-sheet read)`; existing `comments:*` token integrations without creator sheet-read start getting 403. vs (B′) exempt the token path, leaving that leak open.

The session-path fix (9-route gate) + G-8 golden (6/6 green) are correct and ready; only this collision blocks. Full writeup: `/tmp/finding-g8-breaks-oapi-comment-token-contract-20260706.md`.

PR #3758, opened 2026-07-07
feat(approval): T36-1 — projection per-row participant read (Plan A, ratified)

zensgit:
Implements **T36-1** of the ratified per-row visibility lock (#3687, Plan A — owner 2026-07-07). Ordinary users now see THEIR approval outcomes (我发起的 + 我做终态决策的) as multitable rows; everyone else keeps the #3537 admin-only fence bit-for-bit.

## Mechanism (single-choke, all surfaces inherit)
- **Row narrowing rides the W1-2-locked deny choke**: `loadRowLevelReadDenyEnabled` always-on for projection sheets + `loadDeniedRecordIds`/`loadRecordPermissionScopeMap` union in non-participant rows (DENY-WINS) → the 5 records:read routes, export and history inherit narrowing with zero per-surface edits (their choke coverage is already golden-locked).
- **Capabilities**: participants keep `canRead`/`canExport` ONLY; write/fields/views/automation stay denied for every non-admin at all three chokes (REST, Yjs/OAPI `resolveSheetCapabilitiesForUser`, sheet-listing filter). Strangers: full fence, sheet/base never surfaces.
- **Fail-closed hardening the tests caught**: SQL three-valued logic let corrupt rows (missing participant fields) ESCAPE the denied set — `NOT(NULL OR NULL)` is NULL. Closed with COALESCE; empty actor denies all; lookup errors → non-participant; unexpected denied-query errors THROW (read fails, never serves unfiltered rows).

## Verification (fresh DB, migrations from zero)
- New participant suite **9/9** + **#3537 fence goldens 7/7 in the same run** (combined correctness)
- **RED-before mutation**: neutering the participant predicate → exactly the 4 participant tests red, fence stays green (additive, not replacement); reverted
- Generic deny-choke regressions **24/24** (conditional-rule enforce/trash + W1-2 B4 G-7 export⊆read)
- `tsc` 0 · two-point CI wiring — **including the #3537 fence file itself, which was previously wired into NO workflow (skip-green)**

Adversarial review running before merge (security slice). T36-2 (as-built flip + combined regression doc) follows.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -320,6 +320,8 @@ jobs:
             tests/integration/automation-approval-task-created-trigger.test.ts \
             tests/integration/automation-dingtalk-approval-card-action.test.ts \
             tests/integration/approval-card-delivery-wrapper.db.test.ts \
+            tests/integration/approval-projection-visibility.db.test.ts \
+            tests/integration/approval-projection-participant-read.db.test.ts \
             tests/integration/multitable-date-reminder-trigger.test.ts \
             tests/integration/multitable-webhook-event-bridge.test.ts \
             tests/integration/multitable-webhook-retry-tick.test.ts \
--- a/packages/core-backend/src/index.ts
+++ b/packages/core-backend/src/index.ts
@@ -2355,6 +2355,7 @@ export class MetaSheetServer {
         // Auth gate: uses the same sheet + record-level capability resolution as REST
         const sheetCaps = await import('./multitable/sheet-capabilities')
         const { resolveSheetCapabilitiesForUser, canWriteRecord } = sheetCaps
+        const { isRecordReadDeniedForUser } = await import('./multitable/permission-service')
         yjsWsAdapter.setAuthChecker(async (userId, recordId) => {
           try {
             const pool = poolManager.get()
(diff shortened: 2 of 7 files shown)


zensgit:
## Both review P1s fixed (head updated)

**P1-A (adversarial review — main record-read surface)**: `hasRecordPermissionAssignments` now reports true for projection sheets (single point activates the row-filter block on all 5 surfaces: view / list / single-GET / record-history / AI preview). **Wire-level test added** driving `requireRecordReadable` end-to-end — participant reads own record, 403 on others'/corrupt, stranger 403, admin unaffected. RED-before: neutering the branch turns exactly that test red. This kills the choke-only fake-green both reviews called out.

**P1-B (owner review — Yjs record auth)**: new `isRecordReadDeniedForUser` (gated on `loadRowLevelReadDenyEnabled`, DENY-WINS, fail-closed on missing ids) wired into the index.ts authChecker after sheet-level canRead — denied record → `{canRead:false, canWrite:false}`, no doc seed. Helper matrix test + static wiring tripwire (pins the authChecker calls the helper, so a refactor can't drop the rung silently).

**P3 also closed**: sibling projection sheet without the participant's rows stays hidden (per-sheet participant status pinned).

Verification: projection suites **20/20** (participant 13 + #3537 fence 7 in the same run) · generic deny-choke regressions 24/24 · tsc 0 · both new RED-before mutations verified and reverted.

Review MDs: `/tmp/pr3758-review-claude-20260707.md` (adversarial). Awaiting owner re-check → merge call.

PR #3773, opened 2026-07-07
feat(web): multitable UI P2-1c slice-1 — migrate MetaToolbar action buttons to shared MtButton/MtIconButton

zensgit:
First **P2-1c migration** (behavior-adjacent — NOT self-merged; yours to review). Migrates MetaToolbar's standalone action buttons onto the shared primitives; no behavior change.

## Migrated (7)
undo/redo → MtIconButton (ghost); fit/print/import/export-csv/export-xlsx → MtButton (ghost, icon+label); 新建记录 → MtButton primary. Preserved every `:disabled`/`v-if`/`:title`/`:aria-label`/`@click` emit.

## Left as-is (deliberate)
- Dropdown triggers (字段/排序/筛选/分组) — need MtPopover/MtMenu = later slice.
- **Row density ("行高")** — FLAGGED: structurally a dropdown-trigger (toggles a `.meta-toolbar__panel--density`), so treated like the others and deferred, NOT migrated. (The task named it a candidate, but it opens a panel — I kept the panel-trigger rule consistent rather than guess. Say the word if you want it in this slice.)
- reset-to-shared — untouched (green styling; MtButton has no success variant; `data-testid` intact).

## Invariant proof + tests
- **emit() call-site set byte-identical vs main** (all 24 event types, same names + counts) — no emit drift.
- New `meta-toolbar-action-buttons-migration.spec.ts` (11 tests): each migrated button click → same emit; disabled → native `disabled` + no emit; gated buttons absent when prop false; all migrated controls are `<button>` (keyboard-operable). Real emit assertions, not snapshots.
- Regression: `meta-toolbar-group-picker` + `filter-builder` specs still pass → **41/41** across the 3 MetaToolbar specs. vue-tsc clean; token-only; diff = only MetaToolbar.vue + the new test.

Note: some `MultitableWorkbench`-mounting specs fail in a fresh worktree with a pre-existing `grid.filterGroups`/`auth.getToken` fixture gap — verified identical on unmodified main (unrelated to this change).

### DIFF
--- a/apps/web/src/multitable/components/MetaToolbar.vue
+++ b/apps/web/src/multitable/components/MetaToolbar.vue
@@ -116,10 +116,10 @@
         </div>
       </div>
 
-      <!-- Undo/Redo -->
+      <!-- Undo/Redo (UI-P2-1c: migrated to shared MtIconButton — ghost, icon-only) -->
       <span class="meta-toolbar__divider" aria-hidden="true"></span>
-      <button class="meta-toolbar__btn" :disabled="!canUndo" :title="l('toolbar.undoTitle')" :aria-label="l('toolbar.undo')" @click="emit('undo')"><el-icon class="meta-toolbar__btn-icon"><component :is="ICON.undo" /></el-icon></button>
-      <button class="meta-toolbar__btn" :disabled="!canRedo" :title="l('toolbar.redoTitle')" :aria-label="l('toolbar.redo')" @click="emit('redo')"><el-icon class="meta-toolbar__btn-icon"><component :is="ICON.redo" /></el-icon></button>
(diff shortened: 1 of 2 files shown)


zensgit:
**Medium fixed.** `078f574b4`

`mountToolbar()` no longer shares a single global app/container — it uses local vars and pushes each mount to a `mounts` array; `afterEach` unmounts **all** of them (the double-mount present/absent tests left the first tree behind before) and asserts `document.querySelectorAll('.meta-toolbar').length === 0` (residue guard). 11/11 still pass — the guard is the proof the pollution is gone.

Row-density: agreed, left as a dropdown-trigger for the MtPopover/MtMenu slice. Not re-armed for auto-merge — yours to re-review.

PR #3805, opened 2026-07-07
docs(multitable): completion master plan — remaining-dev sequencing + autonomy guardrails

zensgit:
Execution anchor for the full-auto completion goal. Lanes: **A** UI-P2-1c migration sweep (57 components, self-merge on the proven bar), **B** draft remaining design-locks (S5/UI-P2-2/W3-6/AI-L0.5/S1b → ratify queue), **C** owner-gated (ratifies + AI lighting canary/cap + runtime-behind-lock — not done unattended). Red lines: no self-ratify, no production AI env flip, no permission/security semantics; self-merge only presentation-only + docs; branch protection only via restored strict-off bursts. Full doc in the diff.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3814, opened 2026-07-07
docs(multitable): AI S1b — true batch rollback — DESIGN-LOCK (PROPOSED)

zensgit:
**PROPOSED — awaits ratify** (Lane B of completion plan #3805). Batch-level rollback on top of S1's batch_id foundation: locate each record's pre-batch revision → atomic restore (fail-closed, partial→whole-batch reject) through the existing permission/mask write path (no new bypass) → rollback itself provenance-tracked (source=rollback, re-rollbackable). New-record rollback = existing recycle-bin. History Center gains an explicit-confirm controlled write entry. v1 = whole-batch only, reuses existing restore permission gate. Runtime specifics (pre-batch version location, restore gates) flagged for impl-time verification — NOT assumed (learned from the AI-lighting overclaim). Rungs S1b-1/2/3 🔒.

### DIFF
--- a/docs/development/multitable-ai-s1b-batch-rollback-designlock-20260707.md
+++ b/docs/development/multitable-ai-s1b-batch-rollback-designlock-20260707.md
@@ -1,0 +1,47 @@
+# AI 字段 S1b · 真批次回滚 · 设计锁（PROPOSED）
+
+> 状态：**PROPOSED — 待 owner ratify**。docs-only；不改 runtime。
+> 前置（已落 main）：**S1 = 写入血缘 + 批次分组 FOUNDATION**（lock #3569 + runtime #3584；round-2 评审把"batch rollback"**明确降级为 grouping foundation**、S1b 独立门禁——即 S1 造了"一次批量=一个 batch_id"的地基但**明示不承诺回滚**）· record-level version restore（既有，单记录版本还原）· HistoryCenterModal（**当前只读**）· `meta_record_revisions`(带 batch_id)。
+> ⚠ 本锁的运行时具体点（restore 路径能否定位"批次前一版本"、restore 的权限门）**需在 impl 时对当前 head 逐一核实**，不得照抄本文假设——本文定的是形状与不变式，不是实现细节。
+> 模型分档：设计 = Fable/Opus；runtime = Sonnet；回滚是写操作 → 点亮/权限相关的对抗审阅 = Opus。
+
+## 1. 原则
+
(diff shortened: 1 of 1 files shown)


zensgit:
**⚠ Independent adversarial gate: SUPERSEDED-BY-#3940 — recommend CLOSE (do not ratify both).** (Gate ≠ ratify.) This lock and **#3940** (`AI-fields S1b true history-batch rollback`, gated MERGE_CLEAN this session) lock the **identical slice** — same S1 `batch_id` foundation, same goal, same History-Center-write deliverable. Ratifying both leaves two competing S1b locks for one slice.

**And #3940 is the correct one:** this lock’s core premise ("find the revision before the batch, then group-restore") is exactly the naive **nearest-prior** resolution that **#3940 §2 proves silently clobbers** intervening non-revision writes. Verified against `origin/main`: `automation-executor.ts:2163` / `records.ts:489` bump `version` and emit **no** revision, and `history-projection.ts:546` documents `loadPreviousSnapshots` as "nearest surviving prior, **not `version-1`**" — so the exact `v-1` predecessor a batch member needs can be unrecoverable. #3814 never identifies this hazard; #3940 handles it. #3814 also lacks the mutex/seam declaration vs live restore-surface work (#3922/#3928/#3931) that #3940 supplies.

**Recommendation:** ratify **#3940**, close this as superseded — **with one rider:** closing this silently adopts #3940’s *partial-by-default* rollback semantics. This lock makes rollback *all-or-nothing* a hard invariant — the one genuine behavioral disagreement. If you want atomic all-or-nothing at runtime, record it on #3940 (flip its §9 atomic-mode from deferred to in-scope) before ratify. Report `/tmp/ai-s1b-3814-vs-3940-gate-20260708.md`.

zensgit:
Closing as superseded by #3940, which locks the identical S1b slice (same S1 `batch_id` foundation, same History-Center write deliverable) and carries a MERGE_CLEAN adversarial gate (2026-07-08, all load-bearing claims verified at a pinned baseline). This lock's core mechanism — locate 'the revision before the batch' and group-restore to it — is the nearest-prior predecessor resolution that #3940 §2 proves unsound: several user-data writers bump `version` without emitting a revision (automation-executor.ts:2163, automation-service.ts:2793, records.ts:489, univer-meta.ts public-form/attachment edits), so a batch member's true `v−1` predecessor may have no revision row and a nearest-prior revert silently clobbers those intervening writes (history-projection.ts:546 documents loadPreviousSnapshots as 'nearest surviving prior, not version-1'). #3940 replaces that with an exact-`version−1` fail-closed resolver, field-scoped diffs, a live-data-only base, and the mutex/seam declarations vs #3922/#3928/#3931 that this lock omits. One decision travels with this close: this lock's all-or-nothing invariant is NOT carried over — #3940 is partial-by-default with atomic mode deferred to its §9. If all-or-nothing rollback is wanted at runtime, record it on #3940 (flip §9 atomic-mode to in-scope) before ratifying. Nothing else here needs porting; ratification proceeds on #3940 only.

PR #3839, opened 2026-07-07
feat(web): multitable UI P2-1c — migrate MetaImportModal <buttons> to MtButton

zensgit:
## Summary
- Migrate MetaImportModal's paste-step footer **Cancel** + **Preview** buttons from bespoke `<button>`s to the shared `MtButton` primitive (Preview → `variant="primary"`; Cancel → default ghost).
- Behavior byte-identical: `@click` bindings (`requestClose` / `parseAndPreview`) and `:disabled` expressions (`isImporting` / `!rawText.trim()`) unchanged; no `emit(...)` call-site touched.
- **Divergence from sibling P2-1c migrations (call this out for review):** the `.meta-import__btn` / `.meta-import__btn--primary` bespoke CSS is intentionally **kept**, not removed. Those classes are shared by 6+ sibling bespoke buttons in the preview/importing/result steps (back, import, cancelImport, backToMapping, retry, applyFixes, close, per-failure picker) that are out of scope for this slice — so the rules are not dead. This is the migration standard's "keep the sibling's part" case. The migrated buttons keep their original classes for selector stability; since the bespoke CSS still applies at equal specificity, they retain their existing visual look (accepted trade-off, not a defect).
- Close-× glyph and every other step's buttons (behind the multi-step import pipeline) stay bespoke, untouched.

## Test plan
- [x] Added `apps/web/tests/meta-meta-import-modal-migration.spec.ts`: real mount/Teleport (`document`-scoped) interaction test, 4 cases — native `<button>` rendering, `:disabled` bindings on both controls, Cancel click still emits `close`, Preview click still drives `parseAndPreview()` into the mapping/preview step (would go red if either `@click` were dropped).
- [x] `git diff origin/main -- ...MetaImportModal.vue | grep emit(` → empty (no handler changes).
- [x] `git diff origin/main -- ...MetaImportModal.vue` new-line hex scan → empty (no new hardcoded hex).
- [x] `pnpm --filter @metasheet/web run type-check` → passes.
- [x] `pnpm --filter @metasheet/web exec vitest run tests/meta-meta-import-modal-migration.spec.ts` → 4/4 pass.
- Note: `tests/multitable-workbench-import-flow.spec.ts` has 7 failures reproducible identically on `origin/main` (pre-existing `scaleStatsKey`/`filterGroups.value` bug), unrelated to this change — not a regression.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/apps/web/src/multitable/components/MetaImportModal.vue
+++ b/apps/web/src/multitable/components/MetaImportModal.vue
@@ -30,8 +30,8 @@
           ></textarea>
           <div v-if="parseError" class="meta-import__error">{{ parseError }}</div>
           <div class="meta-import__actions">
-            <button class="meta-import__btn" :disabled="isImporting" @click="requestClose">{{ l('import.cancel') }}</button>
-            <button class="meta-import__btn meta-import__btn--primary" :disabled="!rawText.trim()" @click="parseAndPreview">{{ l('import.preview') }}</button>
+            <MtButton class="meta-import__btn" :disabled="isImporting" @click="requestClose">{{ l('import.cancel') }}</MtButton>
+            <MtButton class="meta-import__btn meta-import__btn--primary" variant="primary" :disabled="!rawText.trim()" @click="parseAndPreview">{{ l('import.preview') }}</MtButton>
           </div>
         </div>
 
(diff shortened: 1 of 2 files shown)


zensgit:
Deferring this slice. Adversarial gate found a **shared-class double-styling** defect (P2): the migrated Cancel/Preview keep `meta-import__btn`, whose hardcoded-hex CSS is still live (shared with 11 non-migrated buttons across the preview/importing/result steps) and falls through onto the MtButton root. The clean fix is to migrate **all 13** `meta-import__btn` buttons at once so the shared CSS can be removed — but 11 sit behind the multi-step import pipeline (not cleanly mount-testable), so this isn't a clean single presentation-only slice. Deferring MetaImportModal to a dedicated full-component slice; closing this partial migration to avoid shipping the double-styling. (Gate report /tmp/pr3839-rereview-20260707.md.)

PR #3884, opened 2026-07-08
fix(attendance): CSV header-detection parity on the streaming import path (DT-HARDEN-10)

zensgit:
The large-file/streaming import path trusted row 0 as header unconditionally while the inline path requires a name+date header row, so DingTalk exports with a leading title row silently emptied workDate above the upload threshold. Streaming path now uses the same name+date detection. Adds a byte-identical inline-vs-streaming parity test (+ mutation self-checked).

Roadmap: DT-HARDEN-10. Verify: new parity test 3/3; full backend unit suite green.

### DIFF
--- a/packages/core-backend/tests/unit/attendance-import-csv-header-stream-parity.test.ts
+++ b/packages/core-backend/tests/unit/attendance-import-csv-header-stream-parity.test.ts
@@ -23,6 +23,10 @@ const DINGTALK_STYLE_CSV = [
   '李四,2026-06-02,1002,09:05,18:02',
 ].join('\n')
 
+// A file with no name+date row anywhere: the readers must still fall back to the
+// first non-empty row so upload validation keeps producing its existing 400.
+const HEADERLESS_CSV = ['随便一行,不是表头', 'a,b'].join('\n')
+
 async function collectInlineRows(csvText: string) {
   const rows: Array<{ workDate: string; fields: Record<string, string>; userId?: string }> = []
   const summary = await helpers.iterateImportRowsFromCsv({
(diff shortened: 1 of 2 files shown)


zensgit:
⛔ Auto-merge disarmed. Adversarial review found a **P1: this PR breaks two of the product's own shipped CSV templates.**

`readImportCsvHeaderFromFile` falls back to row 0 when no name+date row is found, so upload *validation* passes — but `iterateImportRowsFromCsvFileAsync` has **no such fallback**, so every row is skipped, `rowCount` is 0, and the upload 400s with "CSV must include at least 1 non-empty data row". Commit ① copied `detectCsvHeaderIndex`'s scan but dropped its tail (`return found ? detectedIndex : 0`).

Reachable via the ≥5 MiB upload path. Regressing templates (both shipped by `IMPORT_MAPPING_PROFILES`):
- `dingtalk_api_columns` — `workDate,userId,…`
- `manual_rows` — `workDate,userId,firstInAt,…`

Root cause of the class: `isLikelyImportHeaderRow` accepts **one** name token (`姓名`/`name`) while `IMPORT_HEADER_CONTEXT_KEYS` declares 17 (`工号`, `userid`, `empno`, …), and it skips `normalizeImportHeaderLookupKey`, so `Work Date` and `姓 名` validate but fail detection. Profile 1's own `userMapKeyField` is `'工号'` — a 姓名-less header is a *designed* input.

Fixing before merge: two-pass streaming (bounded header-index peek, then stream via the existing `explicitHeaderRowIndex` seam), fold the predicate onto the real key sets and call it from all three sites, table-driven parity test over `IMPORT_MAPPING_PROFILES[*].templateColumns`, and wire the new test into CI.

zensgit:
## Red resolved — `4a240264c`. Reproduced your exact 10-case symptom and fixed it.

You were right that this was the same class as the "400 before the fixed parser" hazard. Two independent gaps, both in `plugins/plugin-attendance/index.cjs`:

1. **`isLikelyImportHeaderRow` only recognised a literal `姓名`/`name` cell**, and never applied `normalizeImportHeaderLookupKey`. Two of the three shipped templates (`dingtalk_api_columns`, `manual_rows`) have **no name column** — they key on `workDate`/`userId`. Folded onto the real `IMPORT_HEADER_DATE_KEYS` / `IMPORT_HEADER_CONTEXT_KEYS` vocabulary with normalisation, so header *detection* can no longer diverge from upload *validation*.
2. **The streaming parser lost the `found ? idx : firstNonEmpty` fallback** its predecessor had. A one-line fallback is impossible there (it discards rows unbuffered), so it now does a bounded two-pass peek (`detectCsvHeaderRowIndexFromFile`, capped at `IMPORT_HEADER_SCAN_MAX_ROWS`) then streams through the existing `explicitHeaderRowIndex` seam.

### Your 10 cases, reproduced and explained

The two guards are **independent rescues** — each alone saves a header-at-row-0 file, which is why a single mutation doesn't reproduce the failure. It takes **both** broken together (the PR-head state):

| state | `attendance-plugin.test.ts` (clean DB) |
|---|---|
| PR head `aa88d6f5f` (both broken) | **10 csvFileId cases fail**, 201→400 — your symptom, exactly |
| this fix `4a240264c` | **154 / 154 pass** |

### Verification

- **Clean DB** (fresh `createdb` + migrations, not my polluted local): full `attendance-plugin.test.ts` = **154/154**. (A dirty local DB shows ~14 unrelated failures — holidaySync error-shape etc. — that are pollution, absent on a clean DB.)
- New **`tests/unit/attendance-import-csv-header-stream-parity.test.ts`** (21 tests): table-driven parity over all three shipped templates proving each passes detection AND upload validation; normalisation (`Work Date` / `姓 名` / `EMP_NO`); headerless fallback. It runs in CI via the default unit lane (`pnpm --filter @metasheet/core-backend test`, plugin-tests.yml:138) — no allowlist wiring needed for a `tests/unit` file.
- **Mutations** (each reddens the parity unit test — the real CI guard):

| mutation | parity unit | CSV integ |
|---|---|---|
| drop the streaming fallback | **RED ×2** | green (vocab still rescues) |
| strip normalisation from detection | **RED ×5** | green (each alone rescues) |
| revert to narrow name+date vocabulary | **RED ×3** | green |
| **both** (= PR head) | **RED ×7** | **RED ×10** (your cases) |

The CSV *integration* cases pass under any single mutation because the two guards back each other up; the **parity unit test is the tight regression net**, and it catches every individual mutation.

Full unit sweep of import/csv/header/attendance: clean (the one `multitable-button-routes` failure is a known flake this diff doesn't touch).

Not re-armed for auto-merge — awaiting the same independent-verdict discipline as the rest of the line.

PR #3896, opened 2026-07-08
fix(dingtalk): stop auto-admission from committing orphan users (DT-HARDEN-02)

zensgit:
Sync auto-admission hardcoded `enableDingTalkGrant: true`. A corp-scoped directory account without an `open_id` cannot hold a DingTalk login grant, so the bind assertion threw — **after** `INSERT INTO users`. The sync loop's catch swallowed it and the transaction committed, leaving an ACTIVE local user with no DingTalk identity, no linked directory account, and no presence in the review queue.

**Fix**
- auto-admission computes the grant from openId presence (`resolveDirectoryAutoAdmissionCanGrantDingTalkLogin`, mirroring the bind assertion) → such accounts are admitted with the grant OFF, directory binding intact;
- the grant assertion now runs **before** the `users` INSERT → user+bind is all-or-nothing for every caller;
- new sync stat `autoAdmittedWithoutGrantCount`;
- `scripts/ops/dingtalk-directory-orphan-inventory.mjs` (read-only) inventories pre-existing orphans — remediation stays an owner decision.

**Verification**: 25 unit tests pass (orphan-guard 3 + auto-admission 9 + bind-account 13); `tsc --noEmit` clean. **Mutation-proven**: removing the assert-before-insert turns the orphan-guard test red with the exact orphan symptom (a `users` row inserted before the throw).

**Coverage boundary (honest)**: `syncDirectoryIntegration` orchestration has no test harness repo-wide, so a regression at the call site would not go red. The safety property (no orphan) is protected independently by the mutation-proven assert-before-insert; a call-site regression would only make such accounts fail admission, not corrupt data.

Roadmap: DT-HARDEN-02.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -368,6 +368,7 @@ jobs:
             tests/integration/approval-postgate-acceptance.api.test.ts \
             tests/integration/dept-head-sync-plumbing.test.ts \
             tests/integration/approval-manager-chain.db.test.ts \
+            tests/integration/directory-sync-admission-orphan-guard.db.test.ts \
             tests/integration/approval-requester-department.db.test.ts \
             tests/integration/approval-requester-title.db.test.ts \
             tests/integration/approval-requester-role.db.test.ts \
--- a/packages/core-backend/src/directory/directory-sync.ts
+++ b/packages/core-backend/src/directory/directory-sync.ts
@@ -997,6 +997,22 @@ export async function upsertDirectoryAccountDepartments(
   return summary
 }
 
+/**
+ * DT-HARDEN-02: whether an auto-admitted account can be granted DingTalk login.
+ * A grant needs a stable identity key: corp-scoped accounts key on corpId+openId,
+ * so a corp account WITHOUT an openId cannot be granted (directory 通讯录's
+ * user/get does not return openId — it only appears after a real DingTalk OAuth
+ * bind); non-corp accounts key on unionId and are grantable. Mirrors
+ * `assertDirectoryAccountCanEnableDingTalkGrant` exactly so auto-admission never
+ * requests a grant the assertion would reject (which previously threw AFTER the
+ * users row was inserted, leaving a committed orphan).
+ */
+export function resolveDirectoryAutoAdmissionCanGrantDingTalkLogin(
+  account: { corp_id: string | null; open_id: string | null },
+): boolean {
+  return !normalizeText(account.corp_id) || Boolean(normalizeText(account.open_id))
+}
+
 function normalizeDirectorySyncAuditUserId(adminUserId: string): string | null {
   const normalized = normalizeText(adminUserId)
   if (!normalized || normalized.startsWith('system:')) return null
@@ -2516,6 +2532,7 @@ export async function syncDirectoryIntegration(
       let autoAdmissionCandidateCount = 0
       let autoAdmittedCount = 0
       let autoAdmittedNoEmailCount = 0
+      let autoAdmittedWithoutGrantCount = 0
       let autoAdmissionSkippedMissingEmailCount = 0
       let autoAdmissionExcludedCount = 0
       let autoAdmissionFailedCount = 0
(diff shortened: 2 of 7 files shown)


zensgit:
## Independent adversarial review — APPROVE-with-hardening (no P1), but hardening REQUIRED to close DT-HARDEN-02

Head `6c30d94657` (matches `gh pr view`). **The suspected sibling-class hole is refuted** — unlike the four P1 PRs, this guard is genuinely falsifiable and tested against the REAL function. The orphan-guard test uses a passthrough fake client (records SQL, returns `{rows:[]}`, does NOT return canned answers ignoring the WHERE) and drives the real `createDirectoryAdmittedUserInTransaction`. Mutation A (remove the assert-before-insert at `:3316`) reddens it with the exact orphan symptom (`length 0 → 1`), confirmed on real Postgres.

### P2 (proven on real PG) — the fix closes only the *grant* throw-point; other post-INSERT bind throws still commit orphans
The assert moved before INSERT (`:3316`) covers only the grant assertion. But `applyDirectoryAccountBindInTransaction` (called at `:3373`, **after** the `users` INSERT at `:3367`) still has three throw points: `:3162` (identity key missing), `:3193`/`:3208` (conflicts). The sync catch at `:2530` swallows them with no SAVEPOINT/ROLLBACK, and the outer tx commits.

**Scenario B, proven on real PG:** a corp account with **no unionId and no openId** (grant computed `false`) still INSERTs an active `users` row, then throws at `:3162` → **orphan committed.** This violates §6.2 ("No local user row is inserted if binding cannot be safely completed") and falsifies the PR body's "all-or-nothing for every caller." P2 not P1 only because DingTalk normally returns a unionId, so the common corp+unionId case is now correct.

**Fix:** a per-account SAVEPOINT around INSERT+bind (covers all three throw points), or validate `buildDingTalkIdentityExternalKey` feasibility before the INSERT (covers the identity-missing one). And soften the PR body's "all-or-nothing" claim.

### P3 (disclosed, confirmed) — `syncDirectoryIntegration` orchestration untested
Mutation C (revert the `enableDingTalkGrant` wiring to `true`) = 60/60 GREEN. Mitigated: the tested assert-before-insert protects the safety property regardless, so a wiring regression degrades to "account fails admission," not "orphan committed."

Verdict: **safe to land — it strictly reduces orphans and the guard is real — but do not mark DT-HARDEN-02 complete until the P2 SAVEPOINT/feasibility hardening lands.** I'll take the P2 fix next. `tsc` 0, CI coverage verified (both tests `tests/unit`, run via plugin-tests.yml:138). Full report: `/tmp/pr3896-review-claude-20260709.md`.

zensgit:
## P2 fixed — INSERT+bind is now all-or-nothing via a SAVEPOINT (`87aa0b64e`)

Your finding was exactly right: the pre-INSERT grant assertion covered only one throw point, and `applyDirectoryAccountBindInTransaction` (after the `users` INSERT) still threw — with the grant withheld — when the account had **no openId and no unionId**, or when the identity was already bound to another local user. The sync loop's swallow then committed an orphan.

Wrapped INSERT+bind in a per-account `SAVEPOINT directory_admit_user`: any bind throw rolls the `users` row back and leaves the transaction usable for the next account, so the swallow is safe. This covers **all three** post-INSERT throw points, not just the identity-key-missing one — which is why I chose the SAVEPOINT over a pre-INSERT feasibility check (that would only cover Scenario B, not the identity-conflict vector).

**Proof — new real-DB test** `tests/integration/directory-sync-admission-orphan-guard.db.test.ts` (the fake-client unit test can't prove a rollback):
- no-openId/no-unionId account → throws, **0 orphan** committed (your Scenario B, reproduced);
- identity already bound to another user → throws, **0 orphan**;
- a valid account still admits **after** a failed one in the same transaction (the run isn't poisoned);
- a well-formed account is admitted and its login identity pre-seeded (control).

Mutation table (real PG):

| mutation | result |
|---|---|
| remove `ROLLBACK TO SAVEPOINT` | **RED ×2** (both orphan-assertion tests) |
| remove the SAVEPOINT entirely (pre-fix) | **RED** (orphan committed) |
| control | 4/4 green |

Wired at both points (`vitest.config.ts` exclude + whole file in the approval real-DB step). Softened the code comment's "all-or-nothing" claim to name the SAVEPOINT as what guarantees it. `tsc` 0; full unit 4366/4367 (the one failure is `correlation.test.ts`, a known flake, passes in isolation, untouched by this diff).

DT-HARDEN-02's orphan invariant now holds for every post-INSERT throw. Not armed — held for your merge decision. Full review: `/tmp/pr3896-review-claude-20260709.md`.

zensgit:
## Rebased onto current main — DIRTY resolved (`dd0e56c00`)

The conflict was in `directory-sync.ts`: main's landed #3904 block (`resolveDirectoryPrimaryDepartmentId` + `upsertDirectoryAccountDepartments`) and this PR's `resolveDirectoryAutoAdmissionCanGrantDingTalkLogin` were inserted at the same spot. Resolved by **keeping both** (independent functions). The P2 SAVEPOINT survived the rebase intact (4 refs), and the plugin-tests.yml / vitest.config.ts allowlist entries merged cleanly.

Re-verified on the rebased head: `tsc` 0; the orphan real-DB test 4/4; the fake-client unit test 3/3; full unit **328 files / 4401 tests**. Now BLOCKED (checks running) rather than DIRTY. Still held for your merge decision.

PR #3897, opened 2026-07-08
fix(dingtalk): delivery hygiene — strict robot response, rule indexes, length guards (DT-HARDEN-08)

zensgit:
Three independent delivery-hygiene fixes.

1. **Robot response validation**: a malformed 200 (proxy-hijacked HTML, non-object body) or a *string* `errcode` was recorded as a **successful send**. Now requires a real object with numeric `errcode === 0`.
2. **Delivery telemetry indexes**: both list services do `WHERE automation_rule_id = $1 ORDER BY created_at DESC` with no covering index → sequential scans as history grows. Adds `(automation_rule_id, created_at DESC)` on `dingtalk_group_deliveries` and `dingtalk_person_deliveries` (additive, `IF NOT EXISTS`, reversible).
3. **Message length guards**: titles/bodies are truncated with an ellipsis. The 快捷入口 link block is budgeted **first** (`composeDingTalkBodyWithLinks`) — truncating the assembled string would have silently dropped the links, i.e. shipped a card nobody can act on.

**Verification**: 248 unit tests pass across automation-v1, breach-channel, robot-response, migration, both delivery services, and the new length-guard suite; `tsc --noEmit` clean.

Retention sweep deliberately out of this slice (needs scheduler wiring).

Roadmap: DT-HARDEN-08.

### DIFF
--- a/packages/core-backend/src/db/migrations/zzzz20260708090000_add_delivery_rule_indexes.ts
+++ b/packages/core-backend/src/db/migrations/zzzz20260708090000_add_delivery_rule_indexes.ts
@@ -1,0 +1,28 @@
+import { sql, type Kysely } from 'kysely'
+
+/**
+ * DT-HARDEN-08 — delivery telemetry indexes.
+ *
+ * Both dingtalk-group-delivery-service.ts and dingtalk-person-delivery-service.ts
+ * list deliveries with `WHERE automation_rule_id = $1 ORDER BY created_at DESC
+ * LIMIT $2`. Neither dingtalk_group_deliveries nor dingtalk_person_deliveries has
+ * an index covering that filter+sort shape (only destination_id/local_user_id and
+ * source_type are indexed), so those list queries fall back to a sequential scan
+ * as delivery history grows. Add the missing composite indexes.
+ */
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await sql`
+    CREATE INDEX IF NOT EXISTS idx_dingtalk_group_deliveries_rule_created_at
+    ON dingtalk_group_deliveries(automation_rule_id, created_at DESC)
+  `.execute(db)
+
+  await sql`
+    CREATE INDEX IF NOT EXISTS idx_dingtalk_person_deliveries_rule_created_at
+    ON dingtalk_person_deliveries(automation_rule_id, created_at DESC)
+  `.execute(db)
+}
+
+export async function down(db: Kysely<unknown>): Promise<void> {
+  await sql`DROP INDEX IF EXISTS idx_dingtalk_person_deliveries_rule_created_at`.execute(db)
+  await sql`DROP INDEX IF EXISTS idx_dingtalk_group_deliveries_rule_created_at`.execute(db)
+}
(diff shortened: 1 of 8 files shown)


zensgit:
## Independent adversarial review — APPROVE (no P1/P2)

Head `af3469ddc6` (matches `gh pr view`). All three DT-HARDEN-08 claims are real, wired into the production send path, and **mutation-proven** — the sibling-PR "fixture stands between the test and the guard" pattern does not apply here (4 of 5 test files redden when the guard is broken).

**Mutation proof:**
- Revert the errcode-strictness in `robot.ts` (a malformed 200 / string errcode counted as success) → **9 tests RED** — 7 unit on the REAL `validateDingTalkRobotResponse` (not a re-implemented copy) + 2 wiring through the real breach channel; `success:true` is gated at `automation-executor.ts:3405-3444` on the real JSON parse.
- Revert link-budgeting to the pre-fix "truncate the assembled string" → **1 RED**. `composeDingTalkBodyWithLinks` budgets the 快捷入口 block first; boundary probes (body == limit, body > limit) both stay ≤ 20000 with links intact.
- Neuter title truncation → **2 wiring tests RED**, asserted on the actual `fetch` payload (non-tautological).

**Real-DB (PG :5435):** ran the exact `up()/down()` DDL in a rolled-back txn — both tables + the `automation_rule_id`/`created_at` columns exist, the composite index was genuinely absent, `up()` is idempotent (IF NOT EXISTS), `down()` reversible + idempotent, additive-only, and the index exactly matches the real list query (`WHERE automation_rule_id = $1 ORDER BY created_at DESC LIMIT $2`). Retention sweep correctly deferred (declared DT-HARDEN-08 tail). CI: all 5 files are `tests/unit/**`, run by the core-backend test job.

**Two NITs (non-blocking, optional defense-in-depth):**
- NIT-1: the migration test is a source-text grep (`fs.readFile` + `toContain`), not real-DB execution — can't catch DDL drift. Mitigated (I DB-verified the DDL independently).
- NIT-2: `composeDingTalkBodyWithLinks` never clamps the link block itself — if `linkSection` alone exceeded 20000 the output would too. Producer-unreachable (`linkLines` ≤ 2 entry links + ≤ 2 short lines from system IDs). A final clamp would be belt-and-suspenders.

Verdict: **APPROVE — safe to land as-is.** Full report: `/tmp/pr3897-review-claude-20260709.md`.

zensgit:
## Rebased onto current main — window-4 conflict resolved (`9e98413bf`)

The predicted #3900↔#3897 mutual-file conflict materialized once #3900 merged: six interleaved conflict triplets in `automation-v1.test.ts`, all one region — main's DT-HARDEN-06 batch-partial-failure test and this PR's person-message truncation test competing for the same insertion point. Resolved by **reconstructing both complete tests from their pristine sources** (not stitching fragments; a first extraction grabbed the group-message twin by loose pattern — caught by duplicate-title scan and replaced with the person-message block).

Per the content-merge discipline, the merged spec was **re-run and re-mutated**:
| mutation | result |
|---|---|
| revert H06 partial-failure (`unsentRecipients = resolvedRecipients`) | **RED ×1** (the H06 test) |
| drop truncation (`return value`) | **RED ×2** (both truncation tests) |
| control | 3/3 green |

`tsc` 0; merged spec + both H08 suites 233/233; full unit 4441/4442 (the one failure `data-source-test-ephemeral` passes 14/14 in isolation, zero data-source files in this diff — ordering flake). Armed (window 4); auto-merge fires at green+current.

PR #3898, opened 2026-07-08
fix(dingtalk): encrypt group-robot credentials at rest (DT-HARDEN-03)

zensgit:
`dingtalk_group_destinations.webhook_url` embeds the robot `access_token` and `.secret` is the HMAC signing key — together, enough to inject messages into a customer's group. Both were stored as **plain text**, while every other DingTalk credential here (appSecret, agentId, approvalCardLinkSecret) already goes through `security/encrypted-secrets`. A DB dump leaked live send capability.

**One boundary** (`dingtalk-group-destinations.ts`): encrypt on write, decrypt on read.
- create/update validate the plaintext, persist `enc:` ciphertext;
- `rowToDestination`, test-send, the masked log line, and **the automation executor's send path** all decrypt before validation / HMAC signing / masking;
- `decryptStoredSecretValue` passes non-`enc:` values through → pre-existing plaintext rows keep working and re-encrypt on next write (compatibility reader, per roadmap §6.3);
- `scripts/encrypt-dingtalk-destination-secrets.ts` backfills untouched legacy rows — idempotent, and refuses to run without an explicit `ENCRYPTION_KEY` (running with the dev default would write values production cannot decrypt).

API responses unchanged: URL masked, secret presence-only.

**Verification**: 245 unit tests pass; `tsc --noEmit` clean. **Mutation-proven twice**: (a) making the executor read the raw column turns the new send-path test red — that is the production-outage failure mode; (b) making `create()` store plaintext turns the service test red — that is the P1 itself.

No schema migration: `text` columns hold the `enc:` form.

Roadmap: DT-HARDEN-03.

### DIFF
--- a/packages/core-backend/scripts/encrypt-dingtalk-destination-secrets.ts
+++ b/packages/core-backend/scripts/encrypt-dingtalk-destination-secrets.ts
@@ -46,6 +46,8 @@ export interface BackfillOptions {
 export interface BackfillResult {
   encrypted: number
   alreadyEncrypted: number
+  /** Rows whose value changed between this script's SELECT and its UPDATE — left untouched. */
+  skippedConcurrent: number
   total: number
 }
 
@@ -106,6 +108,7 @@ export async function runBackfill(pool: QueryablePool, options: BackfillOptions)
 
   let encrypted = 0
   let alreadyEncrypted = 0
+  let skippedConcurrent = 0
 
   for (const row of rows) {
     const webhookNeedsEncrypting = !isEncryptedSecretValue(row.webhook_url)
@@ -118,27 +121,42 @@ export async function runBackfill(pool: QueryablePool, options: BackfillOptions)
       continue
     }
 
-    encrypted += 1
     if (options.dryRun) {
+      encrypted += 1
       console.log(`[dry-run] would encrypt destination ${row.id} (webhook=${webhookNeedsEncrypting}, secret=${secretNeedsEncrypting})`)
       continue
     }
 
-    await pool.query(
+    // Compare-and-swap on the exact values this run read: a credential rotated by an
+    // admin between the SELECT and this UPDATE must NOT be overwritten with the
+    // re-encryption of the stale value we are holding. (Encrypt-on-write means the
+    // concurrent writer already stored ciphertext — the row no longer needs us.)
+    const updated = await pool.query<{ id: string }>(
       `UPDATE dingtalk_group_destinations
           SET webhook_url = $2,
               secret = $3
-        WHERE id = $1`,
+        WHERE id = $1
+          AND webhook_url = $4
+          AND secret IS NOT DISTINCT FROM $5
+        RETURNING id`,
       [
         row.id,
         normalizeStoredSecretValue(row.webhook_url),
         row.secret ? normalizeStoredSecretValue(row.secret) : null,
+        row.webhook_url,
+        row.secret,
       ],
     )
+    if (updated.rows.length === 0) {
+      skippedConcurrent += 1
+      console.log(`skipped destination ${row.id}: concurrently modified since this run read it`)
+      continue
+    }
+    encrypted += 1
     console.log(`encrypted destination ${row.id}`)
   }
 
-  return { encrypted, alreadyEncrypted, total: rows.length }
+  return { encrypted, alreadyEncrypted, skippedConcurrent, total: rows.length }
 }
 
 async function main(): Promise<void> {
@@ -154,7 +172,8 @@ async function main(): Promise<void> {
     const result = await runBackfill(pool, { dryRun, allowFirstRun })
     console.log(
       `${dryRun ? '[dry-run] ' : ''}done: ${result.encrypted} row(s) ${dryRun ? 'would be' : ''} encrypted, ` +
-      `${result.alreadyEncrypted} already encrypted, ${result.total} total`,
+      `${result.alreadyEncrypted} already encrypted, ${result.skippedConcurrent} skipped (concurrently modified — re-run to re-check), ` +
+      `${result.total} total`,
     )
   } finally {
     await pool.end()
--- a/packages/core-backend/scripts/encrypt-dingtalk-destination-secrets.ts
+++ b/packages/core-backend/scripts/encrypt-dingtalk-destination-secrets.ts
@@ -7,75 +7,165 @@
  * rows that are never edited again.
  *
  * MUST run with the same ENCRYPTION_KEY / ENCRYPTION_SALT as the application, or
- * the values it writes cannot be decrypted at runtime. Idempotent: values already
- * carrying the `enc:` prefix are left untouched (normalizeStoredSecretValue is a
- * no-op on them), so a re-run is safe.
+ * the values it writes cannot be decrypted at runtime. The derived key is
+ * `pbkdf2(ENCRYPTION_KEY, ENCRYPTION_SALT)` — presence of ENCRYPTION_KEY alone does
+ * NOT prove the pair matches the application: a correct key with the wrong (or
+ * default) salt derives a *different* key and still "succeeds" at encrypting, just
+ * with ciphertext the application can never decrypt again. Before touching any row
+ * this script therefore runs a pre-flight: it decrypts an already-`enc:`-prefixed
+ * value that some previous run of the application wrote, using the KEY/SALT this
+ * invocation holds. A round-trip failure means the pair is wrong and the script
+ * aborts with zero writes. (A fresh encrypt+decrypt "sentinel" cannot catch this —
+ * it always round-trips against whatever key happens to be derived, right or
+ * wrong; the check has to reuse ciphertext the *application* actually produced.)
+ * If every row is still plaintext (first run ever), there is no such value to
+ * check against — pass --confirm-first-run once you have independently confirmed
+ * ENCRYPTION_KEY/ENCRYPTION_SALT match the application's env.
+ *
+ * Idempotent: values already carrying the `enc:` prefix are left untouched
+ * (normalizeStoredSecretValue is a no-op on them), so a re-run is safe.
  *
  * Usage:
  *   DATABASE_URL=… ENCRYPTION_KEY=… ENCRYPTION_SALT=… pnpm dlx tsx \
- *     packages/core-backend/scripts/encrypt-dingtalk-destination-secrets.ts [--dry-run]
+ *     packages/core-backend/scripts/encrypt-dingtalk-destination-secrets.ts [--dry-run] [--confirm-first-run]
  */
 import pg from 'pg'
-import { isEncryptedSecretValue, normalizeStoredSecretValue } from '../src/security/encrypted-secrets'
+import { decryptStoredSecretValue, isEncryptedSecretValue, normalizeStoredSecretValue } from '../src/security/encrypted-secrets'
 
-const dryRun = process.argv.includes('--dry-run')
+export type DestinationSecretRow = { id: string; webhook_url: string; secret: string | null }
 
-async function main(): Promise<void> {
-  const databaseUrl = process.env.DATABASE_URL
-  if (!databaseUrl) throw new Error('DATABASE_URL is required')
-  if (!process.env.ENCRYPTION_KEY) {
+export interface QueryablePool {
+  query: <T = unknown>(text: string, params?: unknown[]) => Promise<{ rows: T[] }>
+}
+
+export interface BackfillOptions {
+  dryRun: boolean
+  allowFirstRun: boolean
+}
+
+export interface BackfillResult {
+  encrypted: number
+  alreadyEncrypted: number
+  total: number
+}
+
+/** Both halves of the KDF input must be explicit — presence of ENCRYPTION_KEY alone is not the invariant that matters. */
+export function assertEncryptionConfigPresent(env: NodeJS.ProcessEnv = process.env): void {
+  if (!env.ENCRYPTION_KEY) {
     throw new Error('ENCRYPTION_KEY is required — running with the default dev key would write values production cannot decrypt')
   }
+  if (!env.ENCRYPTION_SALT) {
+    throw new Error(
+      'ENCRYPTION_SALT is required — the derived key is pbkdf2(ENCRYPTION_KEY, ENCRYPTION_SALT); running with ' +
+      'the default salt while the application uses a different one would write values production cannot decrypt',
+    )
+  }
+}
+
+/** Finds an already-`enc:`-prefixed value in the current row set to verify the running KEY/SALT against. */
+export function findCanary(rows: DestinationSecretRow[]): string | undefined {
+  for (const row of rows) {
+    if (isEncryptedSecretValue(row.webhook_url)) return row.webhook_url
+    if (isEncryptedSecretValue(row.secret)) return row.secret as string
+  }
+  return undefined
+}
+
+/**
+ * Pre-flight, BEFORE any UPDATE: prove this invocation's ENCRYPTION_KEY/ENCRYPTION_SALT
+ * are the same pair that encrypted rows already at rest, by decrypting one of them.
+ * Throws (writes nothing) on mismatch or on an unacknowledged first run.
+ */
+export function verifyEncryptionConfig(rows: DestinationSecretRow[], allowFirstRun: boolean): void {
+  const canary = findCanary(rows)
+  if (!canary) {
+    if (allowFirstRun) return
+    throw new Error(
+      'No already-encrypted row found to verify ENCRYPTION_KEY/ENCRYPTION_SALT against, so this run cannot prove ' +
+      'they match the application. If every row really is still plaintext (first run ever), re-run with ' +
+      '--confirm-first-run after independently confirming ENCRYPTION_KEY/ENCRYPTION_SALT match the application env.',
+    )
+  }
+  try {
+    decryptStoredSecretValue(canary)
+  } catch (error) {
+    throw new Error(
+      'ENCRYPTION_KEY/ENCRYPTION_SALT do not match the pair that encrypted the existing rows ' +
+      `(decrypting a known-good row failed: ${error instanceof Error ? error.message : String(error)}). ` +
+      'Refusing to write — this would overwrite recoverable ciphertext with ciphertext nobody can decrypt.',
+    )
+  }
+}
+
+export async function runBackfill(pool: QueryablePool, options: BackfillOptions): Promise<BackfillResult> {
+  const { rows } = await pool.query<DestinationSecretRow>(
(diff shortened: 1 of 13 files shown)


zensgit:
## Independent adversarial review — CHANGES REQUESTED

Reviewed at head `c4f649fb1`. **The shipped runtime code is correct** — I could not break it. Read points are genuinely exhausted, no credential escapes to any API response, legacy plaintext and encrypted rows sign identically, crypto errors leak nothing. Both mutation claims in the PR body reproduce.

What blocks is what protects this code from the *next* change.

### P2-2 (blocking) — 3 of the 5 surfaces §6.3 names have zero coverage

The roadmap's acceptance line reads: *"Tests cover create, update, list, test-send, and delivery send."*

| guard | mutation | result |
|---|---|---|
| `createDestination` encrypt | store plaintext | **RED** |
| executor decrypt | read raw column | **RED** |
| **`updateDestination` encrypt** | **store plaintext** | **GREEN** — 4350/4350 + 12/12 |
| `rowToDestination` decrypt (list/get) | return raw | **GREEN** |
| `testSend` decrypt | read raw | **GREEN** |
| masked-log decrypt | read raw | **GREEN** |

The `updateDestination` row is the serious one: neutered so it writes **plaintext credentials — the exact P1 this PR exists to fix** — the entire suite stays green. Root cause is fixture shadowing (`dingtalk-group-destination-service.test.ts:64-81` stores plaintext `webhook_url`/`secret`, so every decrypt is a pass-through no-op), plus the update test asserting only the *returned* object and never the *persisted* shape — unlike the create test at L133-136. Four assertions close it. The route integration test `vi.doMock`s the whole service, so it can never close them.

### P2-1 (must fix before running; does not block merge)

The backfill guard is asymmetric: it enforces `ENCRYPTION_KEY` presence but silently accepts the default `ENCRYPTION_SALT`, though the derived key is `pbkdf2(KEY, SALT)`. Probe-confirmed: correct key + wrong salt → `Unsupported state or unable to authenticate data`. The `UPDATE` overwrites the plaintext, so the drift is irreversible. Deliberately **not** rated P1: the catastrophic case needs prod running a non-default salt, which I could not confirm, and merging doesn't execute the script. What is established regardless of config: *presence is not the invariant that matters.* Fix = symmetric check + key-fingerprint pre-flight, or drop the script — the compatibility reader alone makes the PR correct.

**P2-3** — the script is outside `tsconfig.include` (`--listFiles | grep -c` → 0), so "tsc clean" never covered it. No test, not in `package.json`.

### P3 / NIT
- One undecryptable row makes `GET /api/multitable/dingtalk-groups` **hang with no response** — `rows.map` has no per-row isolation and the express@4 route has no try/catch. Verified with supertest. Reachable via a wrong-key backfill or a key rotation: this PR adds a table to the encrypted-at-rest set, but `ConfigService.rotateEncryptionKey` re-encrypts only `system_configs`.
- Backfill `SELECT`→`UPDATE` without CAS → lost update on concurrent edit.
- `encryptDingTalkDestinationValue('')` returns `''` unencrypted (unreachable today); `maskDingTalkWebhookUrl` returns an `enc:` blob verbatim.

**Refuted, do not relitigate:** missed read point in `dingtalk-group-delivery-service.ts:66-85` (it joins the table but selects only `g.name`); API ciphertext leak (contract unchanged); `enc:`-prefix collision (both validators reject it); secrets in crypto error strings; idempotency; name-only PATCH preserving the secret.

One accuracy note: the full-suite failures (`approvals-bridge-routes`, `plm-workbench-capabilities-routes`, one `ECONNRESET`) are pre-existing flakes — each passes in isolation and the diff touches no approval/PLM file.

Full report: `/tmp/pr3898-review.md`.

zensgit:
## Closure — all 4 review items addressed (rebased onto origin/main, head `7b4f130e2`)

### 1. Update-path ciphertext test (already on branch, `440790d09`) — verified
Ran the described mutation: dropped the `encryptDingTalkDestinationValue()` call on
`updateDestination`'s webhook path. `persists rotated credentials encrypted at rest on
update (DT-HARDEN-03)` went **RED** (`expected false to be true` on
`isEncryptedSecretValue(setArg?.webhook_url)`). Restored, green again. This closes the
P2-2 `updateDestination` row from the review's mutation table.

### 2. Backfill script safety (P2-1 / P2-3)
`scripts/encrypt-dingtalk-destination-secrets.ts` now:
- Requires `ENCRYPTION_SALT` symmetrically with `ENCRYPTION_KEY` (the derived key is
  `pbkdf2(KEY, SALT)`; presence of KEY alone doesn't prove the pair matches production).
- Runs a **pre-flight before any UPDATE**: finds an already-`enc:`-prefixed row and
  decrypts it with the KEY/SALT this invocation holds. A round-trip failure aborts with
  **zero writes**. A fresh encrypt+decrypt "sentinel" cannot catch a wrong-salt run — it
  always round-trips against whatever key happens to be derived, right or wrong — so the
  check reuses ciphertext the *application* actually produced. First-ever runs (no
  encrypted row to check against) require an explicit `--confirm-first-run`.
- Stayed idempotent and count-reporting (both already true on the branch).
- Added to `tsconfig.json`'s `include` (scoped to this one file, not `scripts/**/*` —
  the glob would have pulled unrelated scripts like `encrypt-dingtalk-integration-secrets.ts`
  into typecheck), so `tsc --noEmit` now actually covers it.
- New `tests/unit/encrypt-dingtalk-destination-secrets.test.ts` (14 tests) against a fake
  `pool.query` (real crypto, faked I/O) — includes the driving case: seed a canary
  encrypted under one salt, run the backfill under a different salt, assert it rejects
  and **zero** `UPDATE` queries were issued.

### 3. Fail-safe per-row decrypt (P3)
`rowToDestination` (called from `Array.map` in `listDestinations`/`getDestinationById`,
under routes with no try/catch) now catches decrypt failures per row instead of letting
them throw and take down the whole response. A corrupt row is flagged
`credentialUnreadable: true` with `webhookUrl=''`/`secret=undefined` — **never** the raw
`enc:` value, since `maskDingTalkWebhookUrl` passes non-URL strings through verbatim and
would otherwise leak the ciphertext blob as a "masked" webhook URL. Covered at the
service layer (mixed good/corrupt row set — good row unaffected, corrupt row flagged,
`toDingTalkGroupDestinationResponse` never emits `enc:`) and at the route layer
(`GET /api/multitable/dingtalk-groups` returns 200 with the corrupt row flagged, JSON
shape intact).

### 4. Credential-surface coverage (P2-2) — the 5 surfaces named in roadmap §6.3
`create`, `update`, `list`, `test-send`, `delivery send`. `create` and `delivery send`
already had real-ciphertext coverage in the base commit; `update` was closed by the
commit already on the branch (verified above). `list` and `test-send` were only ever
exercised against the plaintext `destinationRow()` fixture, where
`decryptDingTalkDestination*` is a pass-through no-op — added `encryptedDestinationRow()`
(real `enc:` ciphertext via `encryptStoredSecretValue`) and one test per surface.

### Mutation table (this round)

| guard | mutation | result |
|---|---|---|
| `updateDestination` encrypt (already on branch) | store plaintext | **RED** |
| backfill pre-flight | remove `verifyEncryptionConfig()` call | **RED** (wrong-salt catastrophe test) |
| `rowToDestination` decrypt (list/get) | read raw `enc:` column | **RED** (2 tests: list-surface + P3 mixed-row) |
| `testSend` decrypt | read raw `enc:` column | **RED** |
| `rowToDestination` unreadable-row fallback | fall back to raw ciphertext instead of `''` | **RED** (leak guard) |

All mutations reverted after confirming RED; real implementation restored and reconfirmed
green before each subsequent step.

### Test evidence
- `tsc --noEmit -p tsconfig.json`: clean (script now in `include`).
- `tests/unit/dingtalk-group-destination-service.test.ts`: 24/24 (was 21; +3: list
  surface, test-send surface, P3 mixed-row).
- `tests/unit/encrypt-dingtalk-destination-secrets.test.ts`: 14/14 (new file).
- `tests/unit/dingtalk-destination-secret-encryption.test.ts`: 7/7.
- `tests/unit/automation-v1.test.ts` (delivery-send surface): 55/55.
- `tests/integration/dingtalk-group-destination-routes.api.test.ts`: 13/13 (+1: P3
  route-level "responds 200 with corrupt row flagged").
- Full `tests/unit` run: 4571/4573 passed; the 2 failures
  (`approval-metrics-router.test.ts`, `multitable-hierarchy-parent-link-guard.test.ts` —
  `ECONNRESET`) are pre-existing flakes unrelated to this diff — both pass in isolation,
  same category the prior review round already flagged for other files.

### Honest gaps
- Backfill first-run edge case (no canary at all) is guarded by an explicit
  `--confirm-first-run` flag rather than an automated cross-check — there is genuinely
  nothing in `dingtalk_group_destinations` to verify against on a from-scratch table, so
  this is a human-confirms-env-vars gate, not a technical guarantee.
- The backfill's `SELECT`→`UPDATE` without CAS (lost-update on concurrent edit) noted in
  the original review as a NIT is unchanged — out of scope for this round.
- `ConfigService.rotateEncryptionKey` still only re-encrypts `system_configs`, not this
  table — key rotation for `dingtalk_group_destinations` remains a manual backfill
  re-run, same as before this PR.
- Route-level try/catch was **not** added to the GET handlers — the P3 fix is at the
  source (`rowToDestination` no longer throws for decrypt failures), which covers the
  reported hang. A non-decrypt exception (e.g. a DB error) in those handlers would still
  propagate unguarded; that's a pre-existing, broader gap not scoped to this PR.

zensgit:
# Adversarial review — PR #3898 `fix(dingtalk): encrypt group-robot credentials at rest (DT-HARDEN-03)`

- **PR head reviewed**: `7b4f130e22a25d985e8ff9fd2c7f37f29e38d559` (branch `claude/dt-harden-03`, base `main`, merge-base `a894d8d0a`, 1 commit behind main)
- **Reviewer**: Claude (independent adversarial gate), 2026-07-09
- **Method**: fresh detached worktree at PR head; `pnpm install --frozen-lockfile`; all touched test files + full `tests/unit` suite run by me; 9 mutations re-fired by me (7 author-claimed + 2 of my own); all readers of `dingtalk_group_destinations` audited at PR head; CI-wiring of every touched test verified; author's tsconfig-scoping justification empirically tested.

## Verdict: **APPROVE (with P3 hardening backlog)** — no P1/P2. Every author claim I re-tested held.

---

## 1. What I actually ran (evidence, not reading)

| Run | Result |
|---|---|
| `tsc --noEmit -p tsconfig.json` (package dir, PR head) | clean |
| `dingtalk-group-destination-service.test.ts` + `encrypt-dingtalk-destination-secrets.test.ts` + `dingtalk-destination-secret-encryption.test.ts` | 47/47 (24 + 14 + 9) |
| `automation-v1.test.ts` + `dingtalk-group-destination-routes.api.test.ts` | 232/232 (219 + 13) |
| **Full `tests/unit` suite** (`CI=true npx vitest run tests/unit`) | **4573/4573, 343 files, 0 failures** (author saw 2 flakes; I saw none — flake claim consistent) |
| `include: scripts/**/*` counter-experiment | REDs typecheck: TS2347 ×2 (`encrypt-dingtalk-integration-secrets.ts`), TS2393 ×2 (`kv-inspect.ts`), TS2345 (`phase5-snapshot-migrate-and-restore.ts`) → single-file include is the correct minimal scope, author justification **verified** |

## 2. Mutation matrix — all re-fired by me, all RED, worktree restored to clean after each

| # | Mutation (file) | Expected red test | Result |
|---|---|---|---|
| M1 | `create()` stores plaintext (service.ts insert) | `creates and lists destinations` (isEncryptedSecretValue) | **RED** (1 failed / 23 passed) |
| M2 | `updateDestination` stores plaintext webhook+secret | `persists rotated credentials encrypted at rest on update` | **RED** |
| M3 | `rowToDestination` reads raw columns (no decrypt) | `listDestinations decrypts real ciphertext` + `flags an undecryptable row` | **RED ×2** |
| M4 | `testSend` reads raw columns before signing | `testSend decrypts real ciphertext before signing and sending` | **RED** |
| M5 | catch-block falls back to raw `row.webhook_url` instead of `''` | `flags an undecryptable row…` (`not.toContain('enc:')` leak guard) | **RED** |
| M6 | `runBackfill` drops `verifyEncryptionConfig()` | `CATASTROPHE GUARD: wrong ENCRYPTION_SALT aborts with ZERO writes` | **RED** |
| M7 | executor reads raw columns (no decrypt) | `decrypts an encrypted destination before signing and sending` (automation-v1) | **RED** (1/219) |
| M8 *(mine)* | `assertEncryptionConfigPresent` drops the SALT branch | `requires ENCRYPTION_SALT even when ENCRYPTION_KEY is set` | **RED** |
| M9 *(mine)* | `credentialUnreadable` flag never set (spread removed) | `flags an undecryptable row…` | **RED** |

Every security-shaped guard in this PR is load-bearing under mutation. No fixture shadowing found: the new list/test-send/executor tests use **real `enc:` ciphertext** produced by the real `encryptStoredSecretValue`, so the decrypt path is genuinely exercised (the P2-2 fixture-shadowing fix is real — the legacy `destinationRow()` plaintext fixture made decrypt a pass-through no-op, and the new `encryptedDestinationRow()` closes that).

## 3. Reader audit — no missed decrypt site

Every consumer of `dingtalk_group_destinations` at PR head:
- `dingtalk-group-destination-service.ts` — list/get (`rowToDestination`, decrypts + fail-safe flag), update (validate-plaintext→encrypt), testSend (decrypts **inside** the try that records delivery + last_test_error → corrupt row yields a recorded failure, not a hang) ✅
- `automation-executor.ts:3444-3462` — decrypts inside per-destination try/catch; a decrypt failure isolates to `failedDestinations` with a delivery record; other destinations still send ✅
- `dingtalk-group-delivery-service.ts:82` — joins for `g.name` only, no credential columns ✅
- backfill script — `normalizeStoredSecretValue` idempotent (enc: pass-through) ✅
- No other reader (migrations/types aside). Error messages on the decrypt/validate path (`robot.ts`, node crypto) embed neither ciphertext nor tokens ✅

Leak surfaces re-checked: response mapper strips `secret`, masks URL; the `''` placeholder (never raw `enc:` blob) is what reaches `maskDingTalkWebhookUrl`, which passes non-URLs through verbatim — the fallback-to-raw variant was mutation-proven red (M5). `enc:`-prefix collision with user input is unreachable: webhook must be `https://oapi.dingtalk.com/robot/send…`, secret must start with `SEC`.

## 4. CI wiring (no invisible tests)

- 3 new/extended `tests/unit/*.test.ts` files: auto-collected by `pnpm --filter @metasheet/core-backend test` in the required `test (18.x/20.x)` job ✅ (18.x already green on this PR; my local full run = the same collection, 4573 passed)
- `tests/integration/dingtalk-group-destination-routes.api.test.ts`: **not** in `vitest.config.ts` exclude → runs in the default required test job ✅ (mock-service pattern, correctly disclosed as route-shape proof, not decrypt proof)
- No new `.db.test.ts` → no two-point wiring needed ✅
- Script's `main()` guard (`argv[1].includes(...)`) does not fire under vitest — verified empirically (no DATABASE_URL in my runs; suite green, no DB dial) ✅
- No migration; `text` columns hold `enc:` form ✅. PR body has no bare closing keywords ✅. Scope: all 11 files in-lane ✅.

## 5. Findings

### P3-1 — backfill lost-update race (new code in this PR)
`scripts/encrypt-dingtalk-destination-secrets.ts:127-140` — SELECT-then-UPDATE without CAS. If a user rotates credentials via `updateDestination` between the script's SELECT and its UPDATE of that row, the backfill overwrites the freshly-rotated ciphertext with the re-encrypted **stale** value; sends then fail with the old token until re-rotated, silently. Window is per-row milliseconds and the script is a manual ops-window tool, so not landing-blocking — but this is new code, and a one-line CAS (`WHERE id=$1 AND webhook_url=$2_old AND secret IS NOT DISTINCT FROM $3_old`) or `SELECT … FOR UPDATE` closes it. **Fix in follow-up.**

### P3-2 — GET handlers remain unguarded for non-decrypt exceptions (pre-existing)
`src/routes/api-tokens.ts:404` (`GET /dingtalk-groups`) and `:508` (`GET /:id/deliveries`) have no try/catch; the P3 fix contains decrypt failures at the source (`rowToDestination`), which I verified covers the reported hang, but any other rejection (e.g. DB error) still yields an unhandled rejection → hung request under Express 4. Pre-existing, disclosed, out of lane scope. Backlog: wrap or add an async error middleware.

### P3-3 — sibling script has zero pre-flight (adjacent, out of scope, disclosed)
`scripts/encrypt-dingtalk-integration-secrets.ts` has **no** ENCRYPTION_KEY/SALT presence check and no canary round-trip at all — the exact catastrophe class P2-1 just closed for destinations is fully open for `directory_integrations`/`attendance_integrations` appSecret backfills. Correctly left untouched per lane scope. **Deserves its own backlog entry (DT-HARDEN-03b).**

### NIT-1 — `withEnv` misused with an async callback
`tests/unit/encrypt-dingtalk-destination-secrets.test.ts:188` — `withEnv`'s `finally` restores env as soon as the async fn returns its pending promise, i.e. **before** `runBackfill` executes; the CATASTROPHE test actually runs under the beforeEach `unit-test-key/unit-test-salt` pair, not the declared `wrong-salt`. It still discriminates (canary was encrypted under `prod-key/prod-salt`) and the true wrong-salt-same-key proof lives in the sync test at :105-115, so coverage holds — but make `withEnv` async-aware (`await fn()` in try) so the test proves what its name says.

### NIT-2 — `hasSecret:false` for unreadable rows
An undecryptable row reports `hasSecret:false` although ciphertext exists at rest; `credentialUnreadable:true` disambiguates for the UI. Acceptable contract; worth a comment if a UI ever offers "clear secret".

### NIT-3 — tsconfig per-file include precedent
Single-file include is verified-correct for this lane (see §1 counter-experiment), but it starts a per-file listing pattern; a `tsconfig.scripts.json` or cleaning the 3 broken scripts would be the durable fix. Owner call.

## 6. Honest-gaps audit of the author's report

Every disclosed gap checked and confirmed accurate: first-run `--confirm-first-run` is procedural by necessity (nothing to verify against on a virgin table); no `rotateEncryptionKey` machinery exists anywhere in `src/` for any encrypted-secrets consumer (grep-verified), so this table is no worse than its siblings; the routes-test-mocks-service caveat is correctly self-flagged; the sibling-script gap is real (§P3-3). Report claims were not inflated — the one discrepancy (author: 4571/4573 with 2 flakes; me: 4573/4573) favors the PR.

zensgit:
Window protocol executed for landing: rebased onto post-#3903 main (`685643cfc`, clean, 7 commits survived), affected suites 47/47 + tsc clean, load-bearing mutation replayed on the rebased head (encrypt seam → identity ⇒ ciphertext-at-rest tests RED ×2, restored). Armed alone per single-window discipline.

PR #3548, opened 2026-07-04
feat(integration): write self-service W1 config model

zensgit:
## Summary
- add the W1 write-target config validator for consultant-authored external write configs
- add a versioned, content-keyed write-target config store plus audit persistence
- add migration `064_create_integration_write_target_configs.sql`
- wire the new W1 tests into `plugin-integration-core`'s CI test chain

## Merge order
This PR intentionally uses migration `064_*` because #3553 owns `063_*` for read-source composition C-R1.

Merge order is therefore:
1. merge #3553 first, or
2. if #3548 must land first, renumber this migration before merge.

Do not merge #3548 ahead of #3553 while leaving the migration as `064_*`.

## Scope / Boundary
This is W1 only: config model + validation + persistence primitives.

Explicitly out of scope:
- no route
- no runtime executor
- no dry-run/apply path
- no adapter or network call
- no external write
- no production write
- no UI

## Guardrails
- strict config key allowlist
- safe relative `writePath` validation
- backend-reference-only systems: `systemId` + distinct `sandboxSystemId`
- write-only operation declaration: `operations: ['write']`
- delete/raw/prod-shaped keys fail closed
- inline credential keys and secret-shaped values fail closed
- structural `fieldMap` only; value-shaped tokens are rejected
- content-keyed idempotent save/reuse semantics
- values-free audit details: only version/status transition metadata

## Validation
- `node --check` on the new modules and tests
- focused W1 tests
- `pnpm --filter plugin-integration-core test`
- Fresh GitHub CI green, including test (18.x) and test (20.x)

PR #3899, opened 2026-07-08
feat(integration): W2-a write-target dry-run runtime — in-memory compose, zero outbound write, sandbox-only (design-lock #3878)

zensgit:
## Summary

Implements **W2-a** under the ratified design-lock
`docs/development/integration-write-self-service-w2-dryrun-design-lock-20260708.md` (#3878):
a pure, in-memory dry-run **preview** module for a write-target config (W1, #3548).
Owner authorized development of this rung 2026-07-08.

- New: `plugins/plugin-integration-core/lib/write-target-dry-run-runtime.cjs` — validates an
  **approved** W1 write-target config record (fail-closed on draft/retired/missing/invalid/
  non-normalized), checks the caller supplied the sandbox-matching system (production `systemId`
  is never contacted), composes the would-be write body **in memory only** from `fieldMap` per
  row for keying/shape preview, and returns values-free evidence (counts/booleans/exact-registered
  coarse codes). **Zero network of any kind** — not even the optional read-only lookup the lock
  admits at §1.1 (deferred to a later opt-in). No token issuance, no route, no redeemer.
- New: `plugins/plugin-integration-core/__tests__/write-target-dry-run-runtime.test.cjs` — 12
  hermetic test groups (happy path, zero-write proof via a hostile throwing adapter, sandbox-only,
  fail-closed config states, payload caps, key-missing total/partial, values-free sentinel scan,
  exact-code family + clamp, write-dispatch guard with a derived-not-hardcoded marker, a static
  source-scan backstop, `__internals` sanity, and a static no-token-export assertion).
- Wired into the plugin's single `test` script chain (UNION, nothing dropped) + a
  `test:write-target-dry-run-runtime` convenience script.
- New: `docs/development/integration-w2a-dryrun-runtime-dev-verification-20260708.md` — the
  evidence contract, the zero-outbound-write structural guarantee, a 5-mutation kill log (all
  confirmed red, then reverted), the sandbox-only proof, the values-free sentinel scan, the
  fail-closed matrix, the `canApply`/`lookupExecuted` safety-gating rationale, and an explicit
  statement of what remains a separate later opt-in (W2-b token issuance, the lookup-classified
  preview, the W2 route/UI, W3 sandbox-redeem, W4 production apply).

## Zero-outbound-write guarantee (the whole point of this rung)

Three independent layers:
1. **Absence** — no call site to `fetch`/`http(s).request`/`adapter.write|upsert|save`/
   `insertRows|updateRows` anywhere in the module; `deps.createAdapter` is accepted only for
   DI/interface parity with `read-source-read-runtime.cjs` and is never referenced.
2. **Static source-scan test** — greps the module's own source for forbidden write/network patterns.
3. **Write-dispatch guard** (mutation-provable) — `createWriteDispatchGuard()`'s `blockDispatch`
   increments a real counter and throws `WRITE_TARGET_DRY_RUN_WRITE_BLOCKED` in the same step;
   `dryRunWriteTarget` re-checks the counter first thing and fails the whole run closed if non-zero;
   `evidence.externalWriteAttempted` is independently derived from the same counter (never hardcoded).

5 mutations were applied directly to the committed module and each turned a test red (guard
re-check disabled, marker hardcoded, sandbox check widened to accept production, an actual
`adapter.upsert()` call injected into the happy path, approved-status gate widened to accept
`draft`) — full log in the verification MD, §3.2. All reverted; working tree confirmed clean;
full chain re-verified green after.

## Test plan

- [x] `node __tests__/write-target-dry-run-runtime.test.cjs` — green in isolation
- [x] `pnpm --filter plugin-integration-core test` — full chain green, exit 0 (run in no CI
      workflow — verified locally per repo convention)
- [x] 5 mutation-kill experiments against the hard locks, each confirmed red then reverted
- [x] Static source-scan + values-free sentinel scan + no-token-export static assertion

## Explicitly NOT in this slice (separate, later, owner opt-ins)

W2-b write-intent token issuance · the lookup-classified preview level (bounded read-only
sandbox existence check) · the W2 route/UI · W3 sandbox-redeem · W4 production apply. Production
`systemId` is never resolved/contacted/decrypted anywhere in this change; no production-write
default switch is added, flipped, or widened.

DO NOT MERGE without owner review.


zensgit:
**Honest gap (self-flagged during review, added in f92d4c2b3):** everything in the verification MD is proven against this module in isolation, hermetically. One thing is **not** proven: the real W1 config-store record has never been exercised against this module.

- `write-target-config-store.cjs`'s `rowToPublicWriteTargetConfig` returns ~14 fields (tenantId, workspaceId, sandboxSystemId, contentKey, createdBy/updatedBy, timestamps, ...). This module's strict 4-key allowlist (`id`/`status`/`version`/`config`) would reject a real store row verbatim with `config_record_unexpected_field`. This is by design (a narrow, values-free projection, not trusting an arbitrary caller object) — but the future wiring rung must explicitly reshape the store row into this shape, or the allowlist must be revisited. That decision is not made here.
- The store's `get()` also runs the stored config through `sanitizeIntegrationPayload` + version-injection before returning it. This module's byte-identity check (`config === validateWriteTargetConfig(config).normalized`) has only been tested against synthetic fixtures, never against a config that round-tripped through the real save→approve→get path. If that path perturbs the config's shape in any way the W1 validator doesn't also normalize away, every real dry-run would fail closed with `CONFIG_NOT_NORMALIZED` — fail-closed, not silently wrong, but untested.

Mirrors the read-line's own precedent (its runtime tests are also synthetic-config-only, no live store round-trip), and store-wiring is explicitly out of scope for W2-a per the design-lock's §7 disposition — so not a defect in this slice, but the one clause I can't claim is fully proven end-to-end. Full writeup in `docs/development/integration-w2a-dryrun-runtime-dev-verification-20260708.md` §10. Recommended before/during the wiring rung: a real-store round-trip test (`saveVersion` → `approve` → `get` → feed into `dryRunWriteTarget`, reshaped as needed).

zensgit:
## 质量闸(Opus,独立复验)— PASS

零写结构性保证 5 守卫独立 mutation 全 **KILLED**:sandbox-only(接受 production id → 红)· 注入真 adapter.upsert(静态扫描+spy → 红)· fail-closed status(放行 draft → 红)· 派生 marker(buildSuccessEvidence 硬编 → 红,**本轮新补 pin 测**)。

**闸内一处修正**:你自证的 "mutation 2 硬编 externalWriteAttempted" 其实是**可达状态等价 mutation**(调用点 `guard.writeAttempts>0` 恒 false,因 >0 已在更早 fail-closed),任何测试都无法 kill;真正可 pin 的是 evidence-builder 忠实携带派生值——已补直接单测钉死。全链 EXIT=0(79 suites,含新 common test)。

**接受你披露的 follow-up**(记档,非阻塞):configRecord 4键投影 vs 真 store 14字段 + byte-identity 仅测合成 fixture 未过真 store sanitize/version-injection——两者均 fail-closed 非静默错,归 W2-b/wiring rung 的真机验证,与读线同精度纪律。W2-b token / W3 sandbox-redeem / W4 prod 各独立 opt-in,不在本 slice。

PR #3900, opened 2026-07-08
fix(dingtalk): request timeout + honest partial-failure delivery semantics (DT-HARDEN-06)

zensgit:
Three reliability fixes.

1. **Request timeout**: `requestDingTalkJson` used a naked `fetch` with no timeout — gettoken, directory sync, work notifications, approval cards and container login could hang for undici's ~300s default *inside an inline automation execution*. Bounded by `AbortSignal.timeout` (`DINGTALK_REQUEST_TIMEOUT_MS`, default 10s) and surfaced as a typed `DingTalkTimeoutError` so callers record a failed run/delivery instead of hanging. The spread order is deliberate: an explicit `signal: undefined` in a caller's init must not clobber the timeout.
2. **Partial-failure semantics**: work notifications batch by 100. When a later batch threw, the catch marked **every** resolved recipient failed — including earlier batches already sent *and already holding a success row*. One send attempt produced both a success and a failed row for the same recipient. Only unsent recipients are failed now; partial success is reported as a partial result.
3. **Card ledger visibility**: approval-card mark-send-failed swallowed its own error (`.catch(() => null)`), so a card stuck `pending` was fail-closed but invisible. Now logged and surfaced on the step output.

**Verification**: 235 unit tests pass; `tsc --noEmit` clean. **Mutation-proven twice**: (a) reverting the catch to mark all recipients turns the new 150-recipient/2-batch test red on the no-double-recording invariant; (b) dropping the abort signal turns the timeout test red.

Roadmap: DT-HARDEN-06.

### DIFF
--- a/packages/core-backend/src/integrations/dingtalk/client.ts
+++ b/packages/core-backend/src/integrations/dingtalk/client.ts
@@ -73,6 +73,8 @@ export interface DingTalkWorkNotificationResult {
 
 interface DingTalkRequestOptions {
   fetchFn?: typeof fetch
+  /** Override the default per-request timeout (DT-HARDEN-06). */
+  timeoutMs?: number
 }
 
 export interface DingTalkDepartment {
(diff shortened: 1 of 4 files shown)


zensgit:
## Independent adversarial review — APPROVE-with-hardening (no P1); P3 fixed in `fac60261c`

**Runtime is correct across all three fixes.** The sibling-class hole is genuinely absent: the partial-failure guard uses a dumb `INSERT` with no `WHERE`/`ON CONFLICT`, so observing INSERT params *is* observing the guard — a mutation flips 150→250 rows RED, and real Postgres confirms two same-recipient rows persist (no dedup masking it). Lock gates 6.6.1/6.6.2/6.6.3/6.7.1 PASS.

### P3 (fixed) — the timeout-precedence test was a tautology
`tests/unit/dingtalk-request-timeout.test.ts` re-implemented the `{ ...init, signal: … }` merge in the test body and asserted on that local object; flipping production's spread order left 5/5 green. Rewritten to drive `fetchDingTalkAppAccessToken` with a spy and assert on the init that **actually reaches fetch** — a live, un-aborted `AbortSignal`. Mutation-proven: `signal: undefined` in production reddens it (and the existing behavioral test). The pure spread-order precedence stays structurally correct but untested through the public surface — `requestDingTalkJson` is not exported and no in-repo caller passes `init.signal`, so the clobber path has no real producer.

### P2 (tracked follow-up, not blocking) — approval-card ledger-visibility is untested at the executor level
`automation-executor.ts:2673-2699`: reverting the new `try/catch` to `.catch(() => null)` leaves 214/214 green — `deliveryStuckPending`/`deliveryLedgerError` appear in zero test files. This is **observability, not data correctness** (the send failure is still reported; the DB helper `markDingTalkApprovalCardDeliverySendFailed` is itself integration-tested), which is why the review graded it P2. I did **not** force a test here: asserting `output.deliveryStuckPending` requires reading the execution-record output from the event-driven approval-card path, and the only cheaper route is a brittle SQL-string-matched mock — the exact anti-pattern this line avoids. Recipe for a clean version, if wanted: an executor-level test where the mark-failed query throws, asserting `output.deliveryStuckPending === true`.

Verdict: **safe to land.** Not armed — held for your merge decision like the rest of the line. Full report: `/tmp/pr3900-review-claude-20260709.md`.

PR #3902, opened 2026-07-08
fix(dingtalk): batch bind/unbind keeps the audit trail of committed items (DT-HARDEN-04)

zensgit:
Batch bind/unbind commit **one account per transaction**, but the loop failed fast: the first bad item threw, so the caller never learned which items had already **committed** — and the route, which wrote its audit entries only *after* the whole batch returned, silently dropped the audit trail for every one of them. A compliance gap, not a cosmetic one.

- `batchBind/batchUnbindDirectoryAccounts` isolate each item and return `{ succeeded, failed }` instead of throwing on one item;
- the routes audit **every committed item**, even when a later item fails;
- partial failure is a normal `200` batch result carrying per-item errors (`failedCount`, `failed[]`);
- when **nothing** commits, the historical error mapping (404/409/400) is preserved, so single-item failures behave exactly as before.

**Web compatibility**: `DirectoryManagementView` reads `data.items` (now the succeeded set) and throws on `!response.ok` — unchanged behavior. Surfacing `failedCount`/`failed` in the UI is a follow-up.

**Verification**: 49 unit tests pass (35 route + 14 service); `tsc --noEmit` clean. **Mutation-proven twice**: (a) reverting the service to fail-fast turns the new service isolation test red; (b) making the route throw whenever anything failed turns the committed-item audit test red.

Roadmap: DT-HARDEN-04.

### DIFF
--- a/packages/core-backend/src/directory/directory-sync.ts
+++ b/packages/core-backend/src/directory/directory-sync.ts
@@ -3261,24 +3261,46 @@ export async function getDirectoryReviewItem(
   return summarizeReviewItem(row, recommendationsByAccount.get(row.directory_account_id) ?? null)
 }
 
+/**
+ * DT-HARDEN-04: batch bind/unbind commit one account per transaction. A fail-fast loop
+ * threw on the first bad item, so the caller never learned which items had already
+ * COMMITTED — and the route, which wrote its audit entries only after the whole batch
+ * returned, silently dropped the audit trail for every one of them. Compliance gap.
+ *
+ * Each item is now isolated: a failure is a per-item outcome, never a lost success.
+ * Partial failure is a normal batch result, so callers can audit every success and
+ * report exactly which items failed and why.
+ */
+export type DirectoryAccountBatchOutcome = {
+  succeeded: DirectoryAccountMutationResult[]
+  failed: Array<{ accountId: string; error: string }>
+}
+
 export async function batchUnbindDirectoryAccounts(
   directoryAccountIds: string[],
   input: DirectoryAccountUnbindInput,
-): Promise<DirectoryAccountMutationResult[]> {
+): Promise<DirectoryAccountBatchOutcome> {
   const normalizedIds = Array.from(new Set(directoryAccountIds.map((item) => normalizeText(item)).filter(Boolean)))
   if (normalizedIds.length === 0) throw new Error('accountIds are required')
 
-  const results: DirectoryAccountMutationResult[] = []
+  const outcome: DirectoryAccountBatchOutcome = { succeeded: [], failed: [] }
   for (const directoryAccountId of normalizedIds) {
-    results.push(await unbindDirectoryAccount(directoryAccountId, input))
+    try {
+      outcome.succeeded.push(await unbindDirectoryAccount(directoryAccountId, input))
+    } catch (error) {
+      outcome.failed.push({
+        accountId: directoryAccountId,
+        error: readErrorMessage(error, 'Failed to unbind directory account'),
+      })
+    }
   }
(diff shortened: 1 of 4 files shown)


zensgit:

Head `a681d124e4` (matches `gh pr view`). The server-side deliverable is **correct and proven against real Postgres**: per-item isolation works, a mid-batch failure leaves **no orphan/partial write** (each item is one atomic BEGIN/COMMIT/ROLLBACK calling the atomic bind/unbind — the #3896 orphan class does NOT apply here), audit-write and error-mapping are enforced-and-tested, retry is idempotent.

### P2-1 (must-fix hardening) — `batchBindDirectoryAccounts` isolation guard is untested/un-falsifiable
`directory-sync.ts:3134-3146`. Mutation M1 (make the bind loop fail-fast) leaves the entire 49-test suite **GREEN** — the route test mocks the whole service, and `batchBind` has **no** service-level test. So a future revert to fail-fast (reintroducing the exact bug this PR fixes) ships green. The twin `batchUnbind` guard *is* tested (M2 RED), so this is an asymmetry. Runtime is correct (real-DB test passes on head, FAILS under M1). **Fix (taking it):** a service/real-DB `batchBind` mid-batch-failure isolation test, symmetric to the unbind one.

### P3-2 (security-adjacent) — deprovision UI toast reports partial failure as full success
`DirectoryManagementView.vue:5092-5115`: after this PR a partial unbind returns 200, so the client's `if (!response.ok) throw` no longer fires and the toast says "已完成 {total} 个" — an admin revoking access is told it fully succeeded when some accounts still have access. Downgraded to P3 because the UI is explicitly deferred and the list refreshes to the true state (only the toast is optimistic). **But the PR body's "unchanged behavior" claim is inaccurate for the partial case** — partial now returns 200 where it used to be a 4xx. Recommend correcting that claim; the toast fix is a UI-remainder alongside the deferred panel.

P3-1: audit is batch-level in the route, not per-item per §6.4's ideal Target (a crash mid-batch loses audit for committed items) — acceptance bullet met, systemic/pre-existing. Plus a NIT (`auditLog` rejection aborting a committed batch).

Mutations: M1 (bind try/catch) GREEN=untested; M2 (unbind try/catch) RED; M3 (bind route audit) RED; M4 (error mapping) RED. Verdict: **APPROVE-with-hardening.** Full report: `/tmp/pr3902-review-claude-20260709.md`.

zensgit:
## P2-1 closed — `batchBind` isolation now tested (`12c51822d`)

Added the missing symmetric test for `batchBindDirectoryAccounts`, mirroring the existing `batchUnbindDirectoryAccounts` isolation test in the same file: a batch of two accounts where one fails (not-found → the bind throws before opening a transaction) and one succeeds, asserting `succeeded` holds exactly the good item, `failed` holds `{accountId, error: /not found/}`, and the transaction ran exactly once (the good item genuinely committed).

**Mutation M1 (the review's finding):** strip the per-item `try/catch` so the bind loop is fail-fast → **exactly the new test goes RED** (the not-found error propagates out instead of landing in `outcome.failed`); the other 14 stay green. Before, M1 left all 49 green. Restored, 15/15.

Test-only change (verified: no `src/` touched), so no CI wiring needed — it extends an already-run unit file. `tsc` 0; the DT-HARDEN-04 neighbourhood 75/75.

This closes the bind/unbind asymmetry, so a future revert to fail-fast can no longer ship green. The P3-2 toast (partial-unbind reported as success) and the PR-body "unchanged behavior" wording remain as noted — the toast fix belongs with the deferred UI panel; recommend correcting the PR body's partial-case claim. Not armed — held for your merge decision.

PR #3904, opened 2026-07-08
fix(dingtalk): make the primary department explicit and deterministic (DT-HARDEN-07)

zensgit:
`directory_account_departments.is_primary` decides which management chain approvals route up — `ApprovalDirectoryOrg` anchors `direct_manager` and the whole `continuous_managers` chain on it. It was written as `departmentIds[0]`: wherever DingTalk happened to place the department in `dept_id_list` (and, for a user first seen through another department's listing, wherever the merge put it). **A multi-department employee could route approvals up the wrong chain.**

## Why this ships default-off

DingTalk's `topapi/v2/user/get` has **no unambiguous "main department" field**. It returns `dept_order_list: [{dept_id, order}]`, whose `order` is the employee's sort position *within* each department — conventionally lowest for the primary one, but that is not contractually documented. Silently switching the signal would change **live approval routing** on an unverified assumption. So the order-based resolver ships behind `DIRECTORY_PRIMARY_DEPT_FROM_ORDER` (default off, current behavior byte-for-byte). Roadmap §6.8 already gates enabling it on *"confirming the field shape in staging"* — that check remains the owner's call and cannot be done from here.

What ships unconditionally: the choice is now an **explicit, deterministic, tested function** instead of an accident of array order, with a documented tie-break stable across syncs.

- `scripts/backfill-directory-primary-department.ts` recomputes existing rows from stored data (no DingTalk calls). A `--dry-run` before/after flipping the flag shows exactly what enabling it would change.
- **Multi-department approval-routing golden**: the same requester resolves a *different* manager and dept head depending solely on which department carries `is_primary`.

**Verification**: 25 unit tests pass; `tsc --noEmit` clean. **Mutation-proven**: reverting the resolver to `departmentIds[0]` turns the lowest-order test red.

Roadmap: DT-HARDEN-07.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -384,6 +384,8 @@ jobs:
             tests/integration/approval-node-signature-policy.api.test.ts \
             tests/integration/dingtalk-approval-card-deliveries.db.test.ts \
             tests/integration/dingtalk-container-login.db.test.ts \
+            tests/integration/directory-primary-department-write.db.test.ts \
+            tests/integration/directory-primary-department-backfill.db.test.ts \
             tests/integration/attendance-csv-export-bom.test.ts \
             --reporter=dot
 
--- a/packages/core-backend/scripts/backfill-directory-primary-department.ts
+++ b/packages/core-backend/scripts/backfill-directory-primary-department.ts
@@ -1,28 +1,37 @@
 /**
  * DT-HARDEN-07 — recompute `directory_account_departments.is_primary` from stored data.
  *
- * The flag decides which management chain approvals route up (ApprovalDirectoryOrg
- * anchors on it). It was historically written as "whichever department happened to come
- * first in dept_id_list", so a multi-department employee may carry the wrong primary.
- * This recomputes it with the same resolver the sync now uses, from
- * `directory_accounts.raw` (which holds the DingTalk user/get payload).
+ * The flag decides which management chain approvals route up (ApprovalDirectoryOrg anchors
+ * on it). It was historically written as "whichever department happened to come first in
+ * dept_id_list", so a multi-department employee may carry the wrong primary. This recomputes
+ * it with the same resolver the sync now uses, reading `directory_accounts.raw` — the stored
+ * DingTalk `user/get` payload. No DingTalk API calls.
  *
- * Idempotent: it only writes rows whose flag actually changes, and it re-derives from
- * data already in the database — no DingTalk API calls. Run it after enabling
- * DIRECTORY_PRIMARY_DEPT_FROM_ORDER (that flag is read here too, so a dry run before and
- * after tells you exactly what the switch would change).
+ * The department ORDER comes from `raw.dept_id_list`, never from row order. There is no
+ * ordering column on `directory_account_departments`, so a `SELECT` returns heap order and
+ * every `UPDATE` moves the tuple to the heap tail. Deriving `departmentIds` from row order
+ * made this script non-idempotent: it walked the primary department one step further on each
+ * run, WITH THE FLAG OFF. It also made `--dry-run` — which the design-lock names as the
+ * operator's go/no-go for flipping a live approval-routing flag — report changes the sync
+ * would never make.
+ *
+ * With `DIRECTORY_PRIMARY_DEPT_FROM_ORDER` unset this script is now a provable no-op against
+ * a synced database: it derives exactly what the sync writes. Turn the flag on and re-run
+ * `--dry-run` to see precisely which people the switch would re-route.
+ *
+ * An account whose `raw` carries no usable `dept_id_list` is SKIPPED and reported, never
+ * guessed at: guessing would silently re-route a live approval chain.
  *
  * Usage:
  *   DATABASE_URL=… [DIRECTORY_PRIMARY_DEPT_FROM_ORDER=true] pnpm dlx tsx \
- *     packages/core-backend/scripts/backfill-directory-primary-department.ts [--dry-run]
+ *     packages/core-backend/scripts/backfill-directory-primary-department.ts \
+ *     [--dry-run] [--integration=<uuid>]
  */
 import pg from 'pg'
 import {
   isDirectoryPrimaryDepartmentFromOrderEnabled,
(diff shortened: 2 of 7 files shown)


zensgit:
⛔ Auto-merge disarmed. Adversarial review returned **CHANGES-REQUESTED** with a P1. The sync-side change is sound (flag-off equivalence was attacked with `['','D1']`, `['  ','D1']`, `[' D1 ','D2']` and proved unreachable — `client.ts:671-672`/`:621-622` already trim+filter, `mergeDepartmentIds:1994` filters again). What blocks it:

**P1 — the backfill script is non-idempotent and `--dry-run` is a lying instrument.**
`scripts/backfill-directory-primary-department.ts:52,66-68`. Its load-bearing comment ("the sync writes rows in dept_id_list order, so the query's ORDER BY keeps the same fallback semantics") is **false**: there is no ordering column on `directory_account_departments` (`created_at` is a single transaction timestamp; the PK's department id is a random uuid). The query returns heap order, and each UPDATE moves the tuple to the heap tail — so the "primary" walks the department list on every run, **with the flag off**, which by definition must be a no-op:

```
state@start : 100=true  200=false 300=false 400=false
run#1       : 100=false 200=true  300=false 400=false
run#2       : 100=false 200=false 300=true  400=false
run#3       : 100=false 200=false 300=false 400=true
```

Reproduced on real Postgres 16 against a correctly-synced DB. The routing damage is mitigated (the next sync rewrites `is_primary`), but two things are not: an ops script advertised as *idempotent* repeatedly rewrites the approval anchor, and **the design-lock (§3.6:114, already on main) names `--dry-run` as the owner's go/no-go for flipping a live approval-routing flag** — while it reports phantom changes with the flag off.

**P2-1** `parseDirectoryDepartmentOrderList` — `Number(record.order)` folds `null` / `''` / `[]` / `false` to `0`, which then always wins lowest-order. The `Number.isFinite` guard catches `undefined` (NaN) and nothing else.

**P2-2** — the one line this PR actually changes behaviour on (`directory-sync.ts:2416`) has **no test**. Reverting it to `departmentIds[0]` leaves 324 files / 4353 tests green. The mutation evidence only covers the resolver's internals.

**P2-3** the routing golden's mock ignores `params`, so it cannot catch a wrong `$2`. **P2-4** the backfill has no per-account transaction (transient, and on crash permanent, double-primary; no unique constraint exists). **P3-1** `tsconfig.include` excludes `scripts/`, so this PR's "tsc clean" says nothing about the new script.

Fixing all of P1, P2-1..P2-4 before re-arming. Full report: `/tmp/pr3904-review.md`.

zensgit:
## P1 + P2-1..P2-4 fixed — `78842f5e1`

**P1 — the backfill walked the primary department.** Its load-bearing comment ("the sync writes rows in dept_id_list order, so the query's `ORDER BY` keeps the same fallback semantics") was false: there is no ordering column on `directory_account_departments`, a `SELECT` returns heap order, and every `UPDATE` moves the tuple to the heap tail.

Order now comes from `raw.dept_id_list` — the stored `user/get` payload, which is exactly what the sync reads (`directory_accounts.raw` ← `user.source`, verified at `directory-sync.ts:2346`). With the flag off the script is a provable no-op, so `--dry-run` means what the design-lock says it means. An account whose `raw` has no usable list is **skipped and reported, never guessed at** — guessing re-routes a live approval chain. Each account's flag move is one transaction (P2-4). Added `--integration=<uuid>` scoping, which the script needed anyway.

**P2-2 — the write had no test, and reverting it left 4353 green.** That was the real defect. `is_primary` was computed inline in `syncDirectoryIntegration`, which needs a live DingTalk and so has no test at all.

Extracted `upsertDirectoryAccountDepartments`. This is a seam, not decoration: a real-DB golden now drives the actual write, and it means **PERF-01 (#3911) can no longer silently revert this fix** — its bulk `unnest` rewrite has to keep the golden passing rather than quietly restoring `departmentIds[0]`.

**P2-3 — the routing golden was a tautology.** `buildQuery('20')` returned "20 is primary" regardless of `params`, so it proved only that routing follows whatever the database says — green no matter what `is_primary` was written as. Deleted, with a comment explaining why, and replaced by a real-DB test where the writer and the router *meet*: a requester in two departments with two different leaders, the real write, then the real `ApprovalDirectoryOrg` SQL naming a manager.

**P2-1 — `Number(record.order)`** folded `null`, `''`, `[]`, and `false` to `0`, and `0` is the winning order. Now typed: a number, or a numeric string; anything else drops the entry.

### Mutation record (2 new real-DB files, 20 goldens)

| # | mutation | real-DB | unit |
|---|---|---|---|
| A | **revert the write to `departmentIds[0]`** | **RED ×4** | green |
| B | drop the belongs-to-department filter in the resolver | green | **RED ×1** |
| C | restore `Number()` coercion on `order` | **RED ×1** | **RED ×1** |
| D | backfill re-derives order from rows (the original P1) | **RED ×5** | green |
| E | backfill guesses instead of skipping a missing `dept_id_list` | **RED ×1** | green |
| F | backfill `--dry-run` writes | **RED ×2** | green |
| — | control | 20/20 green | 11/11 |

Mutation A is the one that matters: it is exactly the revert that used to leave 4353 tests green. Mutation B is caught only at the unit level — that guard is a property of the pure resolver, which is the right place for it, but recording the asymmetry rather than claiming both lanes cover it.

The backfill fixture deliberately makes `dept_id_list` order **disagree** with alphabetical order (`bfz` declared first, `bfa` sorts first), so a regression that re-derives order from rows — heap order *or* my own `ORDER BY external_department_id` — picks the wrong department and goes red. Mutation D confirms it does.

Both files wired at **both** points: excluded from the no-DB job so they cannot skip-green, and added as whole files to the `Run approval real-DB integration` step in `plugin-tests.yml`.

`tsc --noEmit` exit 0. Full unit: 4352/4353 — the one failure is `multitable-button-routes.test.ts`, which passes in isolation and which this diff does not touch (a known flake, also seen by the #3898 reviewer).

`DIRECTORY_PRIMARY_DEPT_FROM_ORDER` stays default-off; §6.8 still gates enabling it on confirming `dept_order_list` against a real tenant in staging. Not re-armed for auto-merge.

PR #3905, opened 2026-07-08
feat(dingtalk): offboarding policy executor, default-off with a preview (DT-OPS-01)

zensgit:
`default_deprovision_policy` and `deprovision_policy_override` have existed since the schema was created and were **never enforced**. Removing a member from DingTalk marked the shadow account inactive and dropped it into the review queue, but their **local account stayed active** — password login kept working, and `unbind` only cleared the identity. Offboarding never closed the access loop.

## Why this ships default-off

Two hazards shape the design:
1. the column's **DB default is already `mark_inactive`**, so naively honouring the stored value would silently start deactivating users on the very next sync of *every existing integration*;
2. deactivating the wrong person locks out a real employee.

So the executor is env-gated: `DIRECTORY_DEPROVISION_ENABLED`, **default off**. With it off — the shipped default — nothing is written and the run reports exactly what it *would* have done (`deprovisionCandidateCount` / `GrantsDisabled` / `UsersDeactivated`). That is the *"show policy effect before enabling it"* preview the roadmap asks for.

## Policies

| policy | effect |
|---|---|
| `manual_review` (and anything unrecognised) | review-only, exactly as today |
| `disable_grant_only` | revoke the DingTalk login grant, keep the local user active |
| `mark_inactive` | revoke the grant **and** deactivate the local user |

An unknown stored value degrades to review-only — **never** to a destructive action.

Applied effects are audited post-commit and invalidate the user's cached permissions. `countInactiveLinkedDirectoryAccounts()` exposes the "offboarded but unprocessed for N days" metric.

**Verification**: 50 unit tests pass (12 new + 38 neighbours); `tsc --noEmit` clean. **Mutation-proven**: bypassing the gate turns the "DEFAULT-OFF writes nothing" test red — the exact regression that would deactivate users on deploy.

Roadmap: DT-OPS-01 (P1-severity, Phase 2 by size).

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -354,7 +354,7 @@ jobs:
             tests/integration/multitable-tombstone-retention-realdb.test.ts \
             --reporter=dot
 
-      - name: Run approval real-DB integration (directory endpoints + P1-C field redaction + P1-B add_sign/reduce_sign + direct_manager create/start + A/E/B/D/G cross-lane acceptance + dept_head sync-plumbing carry-forward + continuous_managers chain walk + common template presets)
+      - name: Run approval real-DB integration (directory endpoints + P1-C field redaction + P1-B add_sign/reduce_sign + direct_manager create/start + A/E/B/D/G cross-lane acceptance + dept_head sync-plumbing carry-forward + continuous_managers chain walk + common template presets + DT-OPS-01 deprovision selection goldens)
         # Lane A (directory endpoints) + Lane B (P1-C hidden field redaction) + Lane D (add_sign/reduce_sign). Wired as WHOLE FILE runs so the
         # gate is robust (a name filter that matches zero tests would exit green and hide
         # regressions). The file is describeIfDatabase-guarded + excluded from the no-DB
@@ -396,6 +396,7 @@ jobs:
             tests/integration/directory-primary-department-backfill.db.test.ts \
             tests/integration/directory-sync-preview-apply-parity.db.test.ts \
             tests/integration/directory-sync-alert-coverage.db.test.ts \
+            tests/integration/directory-deprovision-selection.db.test.ts \
             tests/integration/attendance-csv-export-bom.test.ts \
             --reporter=dot
 
--- a/packages/core-backend/src/directory/directory-sync.ts
+++ b/packages/core-backend/src/directory/directory-sync.ts
@@ -1,5 +1,6 @@
 import * as bcrypt from 'bcryptjs'
 import * as crypto from 'crypto'
+import { auditLog } from '../audit/audit'
 import { buildOnboardingPacket } from '../auth/access-presets'
 import { recordInvite } from '../auth/invite-ledger'
 import { issueInviteToken } from '../auth/invite-tokens'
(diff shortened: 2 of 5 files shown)


zensgit:
⛔ Auto-merge disarmed. Adversarial review returned **CHANGES-REQUESTED with two P1s**. The merged artifact is provably inert (default-off is mutation-verified), but both P1s sit in the *enable* path, the preview meant to guard the flip is generated by the same defective query, and **there is no reactivation path** — `users.is_active=FALSE` here is the backend's only writer of that value, and the only `true` writer is invite-accept. A mistake is not undoable from any admin surface.

**P1-1 — the wrong person gets deactivated.** Candidates are scoped by `integration_id` (`directory-sync.ts:920-931`), but the write is global: `UPDATE users SET is_active=FALSE WHERE id=$1` (`:966`). `directory_account_links` is UNIQUE on `directory_account_id` only; `local_user_id` is a plain index — **N accounts → 1 user**. Rehire, single integration, no exotic state: the removed account stays `linked` (the link loop skips already-linked rows, `:2549`), the new account links via the stable `unionId`, and the next sync deactivates an actively-employed person. Two integrations: removal from B deactivates a user active in A. Needs a `NOT EXISTS` active-sibling guard, deliberately **not** scoped by `integration_id`.

**P1-2 — "absent from the fetch" ≠ "departed", and there is no circuit breaker.** Four non-throwing ways to vanish: a blank `name` is silently filtered (`client.ts:642`), a missing `list` yields zero users with no error (`:616`), `contain_access_limit:false` (`:607`) — *this repo already knows that happens; `sampleRootDepartmentDiagnostics` exists to detect it* — and a narrowed `rootDepartmentId`. The executor inherits the precision of `directory_accounts.is_active`, a flag that was never load-bearing before. Same bug class as #3883, two commits earlier. No max-batch abort, no refuse-on-empty-fetch.

**The money shot (P2-3):** inverting the selection SQL so it deprovisions *every active employee* leaves **4354/4354 tests green**. The `fakeClient` regex-matches SQL text and returns canned rows regardless of the `WHERE`. Every predicate deciding who gets deactivated has zero coverage. Meanwhile mutation 1 (bypassing the gate) is correctly red — the gate is tested, the *selection* is not.

Also: **P2-1** the `:918` comment says "vanished in THIS run" but the predicate matches the lifetime backlog; **P2-2** `countInactiveLinkedDirectoryAccounts` is dead (zero callers) *and* structurally always 0 (the blanket `updated_at=NOW()` re-stamp at `:2452` lacks an `is_active=true` guard); **P2-4** the `try/catch` around `auditLog` is dead code — `audit/audit.ts:22-40` already catches internally and never rejects, so the loud error is unreachable; **P2-5** `outcome.affected` carries identities and is discarded, so the operator sees a number and never a name.

Fixing P1-1, P1-2, P2-1..P2-5 before re-arming. `DIRECTORY_DEPROVISION_ENABLED` must not be set anywhere — staging included — until a real-DB test covers the selection predicate. Full report: `/tmp/pr3905-review.md`.

zensgit:
## P1-1 / P1-2 fixed — `75fed923a`

Both P1s from the adversarial review are addressed, and the reason neither was caught is addressed too.

**P1-1 — rehire deactivates an employed person.** Verified against the live schema before touching code:

```
idx_directory_account_links_account  UNIQUE, btree (directory_account_id)
idx_directory_account_links_user             btree (local_user_id)     <-- not unique
```

N accounts → 1 user, while the write is global (`UPDATE users SET is_active = FALSE WHERE id = $1`). Fixed with a sibling guard: a person is a candidate only if they hold no other linked, **active** directory account anywhere. Deliberately **not** scoped by `integration_id` — scoping it is the plausible wrong fix, and there is now a golden that fails if someone adds that scope back.

**P1-2 — "absent from the fetch" ≠ "departed".** Added a circuit breaker: zero accounts fetched, or a batch above `DIRECTORY_DEPROVISION_MAX_BATCH` (default 25), aborts the pass and deprovisions nobody. `deprovisionAbortedReason` lands in the run stats.

Also: P2-1 candidates are now this run's *transitions* (`AND is_active = true … RETURNING id`), not the lifetime backlog; P2-2 `countInactiveLinkedDirectoryAccounts` deleted (zero callers, structurally always zero); P2-4 the dead `try/catch` around `auditLog` removed; P2-5 affected identities now surface in stats (capped at 100) — an operator deciding whether to flip the flag needs to see *who* would lose access.

### The test shape, which was the actual defect

The old unit test's `fakeClient` regex-matched the SQL and returned canned rows whatever the `WHERE` said. Your P2-3 showed the consequence: invert the selection to target every active employee, 4354/4354 green.

New: `tests/integration/directory-deprovision-selection.db.test.ts` runs the real SQL against the **real migrated tables** — not a hand-rolled replica, which would just re-introduce the drift. The unit file keeps only what a stub can honestly prove (policy dispatch, breaker, env gate) and now carries a scope note plus a throw-on-unknown-SQL stub.

Wired at **both** points, or it would run nowhere: excluded from the no-DB job (`vitest.config.ts`) so it cannot skip-green, and added as a whole file to the `Run approval real-DB integration` step in `plugin-tests.yml`.

### Mutation record (real Postgres, 10 goldens)

| # | mutation | result |
|---|---|---|
| A | delete the sibling guard | **RED** ×2 — rehire + cross-integration |
| B | scope the sibling guard by `integration_id` (valid SQL) | **RED** ×1 — cross-integration, exactly |
| C | revert to lifetime-backlog selection (valid SQL) | **RED** ×1 — backlog, exactly |
| D | neuter the circuit breaker (`return null`) | **RED** ×2 — both breaker arms |
| E | invert selection to target **active** employees | **RED** ×6 (was all-green before) |
| F | let the harsher policy win per person | **RED** ×1 — least-destructive golden |
| — | control | 10/10 green |

B and C were first written as invalid SQL and killed all 10 tests on a bind-param crash; that proves nothing, so both were rewritten as *valid* mutations. The kills above are semantic.

Full suite: 324 files / 4360 tests, `tsc --noEmit` exit 0.

**Still not re-armed for auto-merge, and `DIRECTORY_DEPROVISION_ENABLED` must not be set anywhere — staging included — until an operator has read a preview run's `deprovisionAffected`.**

zensgit:
## Rebased onto current main — DIRTY resolved (`56e5e9d06`)

The conflict was the recurring insertion-point pattern in `directory-sync.ts`: main's landed #3904 primary-department block and this PR's deprovision-executor block both insert after `computeAutoAdmissionScope`. Resolved by **keeping both** (independent sections); the second commit's conflicts were purely additive allowlist unions (`plugin-tests.yml` + `vitest.config.ts`). All three commits survived the rebase (executor → lazy audit import → P1-1/P1-2 fix).

Re-verified on the rebased head: `tsc` 0; the selection goldens **10/10** against real Postgres; **mutation E (invert the selection to target ACTIVE employees) still reddens 6 goldens** — the exact mutation that left 4354 tests green before this PR's fix remains load-bearing after the rebase; control 10/10. Full unit **329 files / 4419 tests green**.

Now current-and-CI-running rather than DIRTY. **Not armed** — one merge window at a time per the train protocol (#3896 holds the window). `DIRECTORY_DEPROVISION_ENABLED` remains default-off and must stay unset everywhere until an operator reads a preview run's `deprovisionAffected`.

zensgit:
## Window-7 rebase (`09035c7b1`) — third insertion-point resolution, all guards re-proven

Post-#3915/#3972/#3914 main required a manual rebase: the recurring ~1000 insertion point (now holding #3896's helper + #3915's identity-match block) vs this PR's deprovision executor — resolved keep-both; plus the three-way allowlist union in `plugin-tests.yml` (parity + coverage + this PR's selection goldens). All three commits survived.

Re-verified on the rebased head: `tsc` 0; **three real-DB suites 18/18** (deprovision selection + preview/apply parity + orphan guard — the two sibling suites prove this rebase didn't disturb the landed cars); **mutation E (invert selection → deprovision active employees) → RED ×6**; full unit **336 files / 4491 tests, zero flakes**. Armed; auto-merge fires at green+current. `DIRECTORY_DEPROVISION_ENABLED` stays unset everywhere until an operator reads a preview run's `deprovisionAffected`.

PR #3910, opened 2026-07-08
fix(dingtalk): resolve work-notification credentials from the recipient's integration (DT-OPS-04)

zensgit:
A DingTalk userid only means anything **inside its own corp**, but the person-notification path called `readDingTalkMessageConfigFromRuntime()` with no integration — so the stored-config resolver fell back to *"whichever integration sorts first as latest active"* (`ORDER BY status='active', updated_at DESC LIMIT 1`). Under two integrations, a user bound to corp A could be notified with **corp B's credentials**.

- the recipient query now carries the integration each recipient is actually bound under, and that scopes the credential lookup;
- recipients spanning **more than one** integration are refused outright rather than silently notified with the wrong corp's credentials — the send never reaches the DingTalk API;
- env-first resolution inside `readDingTalkMessageConfigFromRuntime` is untouched, so env-configured (and single-integration) deployments keep their bootstrap behavior exactly.

**Verification**: 215 unit tests pass; `tsc --noEmit` clean. **Mutation-proven**: dropping the cross-integration guard turns the "refuses to notify recipients bound under different DingTalk integrations" test red.

**Scoped out**: approval-card delivery rows carrying their own `integration_id` (needs an additive migration) — tracked as remaining DT-OPS-04 work.

Roadmap: DT-OPS-04.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -326,6 +326,7 @@ jobs:
             tests/integration/automation-approval-completed-trigger.test.ts \
             tests/integration/automation-approval-task-created-trigger.test.ts \
             tests/integration/automation-dingtalk-approval-card-action.test.ts \
+            tests/integration/dingtalk-person-message-integration-scoping.db.test.ts \
             tests/integration/approval-card-delivery-wrapper.db.test.ts \
             tests/integration/approval-projection-visibility.db.test.ts \
             tests/integration/approval-projection-participant-read.db.test.ts \
--- a/packages/core-backend/src/multitable/automation-executor.ts
+++ b/packages/core-backend/src/multitable/automation-executor.ts
@@ -1773,11 +1773,14 @@ export class AutomationExecutor {
     }
 
     const recipientResult = await this.deps.queryFn(
+      // DT-OPS-04: the rule creator's userid is only meaningful inside their own corp, so
+      // carry the integration they are bound under and notify them with its credentials.
       `SELECT u.id AS local_user_id,
-              linked.external_user_id AS dingtalk_user_id
+              linked.external_user_id AS dingtalk_user_id,
+              linked.integration_id AS integration_id
          FROM users u
          LEFT JOIN LATERAL (
-           SELECT a.external_user_id
+           SELECT a.external_user_id, a.integration_id::text AS integration_id
              FROM directory_account_links l
              JOIN directory_accounts a ON a.id = l.directory_account_id
             WHERE l.local_user_id = u.id
@@ -1794,6 +1797,7 @@ export class AutomationExecutor {
     )
     const row = (recipientResult.rows[0] ?? null) as Record<string, unknown> | null
     const dingtalkUserId = typeof row?.dingtalk_user_id === 'string' ? row.dingtalk_user_id.trim() : ''
+    const ruleCreatorIntegrationId = typeof row?.integration_id === 'string' ? row.integration_id.trim() : ''
 
     if (!row || !dingtalkUserId) {
       await recordDingTalkPersonDeliverySafely(this.deps.queryFn, {
@@ -1812,7 +1816,7 @@ export class AutomationExecutor {
     }
 
     try {
-      const messageConfig = await readDingTalkMessageConfigFromRuntime()
+      const messageConfig = await readDingTalkMessageConfigFromRuntime(ruleCreatorIntegrationId || undefined)
       const accessToken = await fetchDingTalkAppAccessToken(messageConfig, { fetchFn: this.deps.fetchFn })
       const result = await sendDingTalkWorkNotification(
         accessToken,
@@ -2641,10 +2645,11 @@ export class AutomationExecutor {
     const recipientResult = await this.deps.queryFn(
       `SELECT u.id AS local_user_id,
               u.is_active AS local_user_active,
-              linked.external_user_id AS dingtalk_user_id
+              linked.external_user_id AS dingtalk_user_id,
+              linked.integration_id AS integration_id
          FROM users u
          LEFT JOIN LATERAL (
-           SELECT a.external_user_id
+           SELECT a.external_user_id, a.integration_id::text AS integration_id
              FROM directory_account_links l
              JOIN directory_accounts a ON a.id = l.directory_account_id
             WHERE l.local_user_id = u.id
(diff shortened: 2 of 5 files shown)


zensgit:
⛔ Auto-merge disarmed. Adversarial review: **CHANGES-REQUESTED**. The security direction is right — on the stored-config path this genuinely stops corp A's user being notified with corp B's credentials. Two things block it.

**P1 — the refuse aborts the whole rule, and fires where the bug is impossible.** Measured against `origin/main` and PR head with identical inputs (env fully configured, rule = `[send_dingtalk_person_message, update_record]`, recipients spanning two integrations):

| source | execution | steps |
|---|---|---|
| `origin/main` | `success` | `[send: success, update_record: success]` |
| PR head | `failed` | `[send: failed, **update_record: skipped**]` |

The refuse hits `executeActions`' fail-stop and skips **unrelated downstream actions**. And `readDingTalkMessageConfigFromRuntime` returns env config *before* consulting stored config — so in a fully env-configured deployment there is exactly one credential and the wrong-corp bug **cannot occur**, yet the refuse sits upstream and aborts anyway. Zero security benefit, hard failure. The PR's own test sets all three `DINGTALK_*` vars and still asserts `failed`, which refutes this PR body's claim that env-configured deployments are unaffected. Roadmap §7.5 never asks for a refuse — per-integration batching satisfies both its acceptance bullets with no failure, and `AttendanceNotificationDeliveryWorker` already resolves config per recipient. (Also: `directory_integrations` has no uniqueness on `corp_id`, so two rows can share one corp — where the pre-PR send was correct and now hard-fails.)

**P2 — the scoping fix has zero coverage, mutation-proven twice.** Deleting the scoping (`…FromRuntime()`) → **4344/4344 green**. Keeping the variable but corrupting it to `'dir-WRONG-CORP'` → **215/215 green**. Structural cause: the test named *"resolves credentials from the recipient's own integration"* sets the env vars, so the resolver short-circuits and no test that sets them can ever observe scoping. Its only scoping assertion is a SQL substring check.

**P2** — `notifyRuleCreatorOfDingTalkGroupFailure` (`:1786`) uses the same LATERAL and is still unscoped; undisclosed in the PR body. **P2** — intra-user ambiguity: the automation path silently `LIMIT 1`-picks while its sibling worker refuses on `dingtalk_recipient_ambiguous`; a sync bumping `updated_at` can flip a dual-bound user's corp.

On my pointed question about NULL `integration_id` masking: **the mechanism is real** (mixed-NULL → `filter(Boolean)` masks it → one batch on the wrong corp's credentials; all-NULL → latest-active silently restored) but **unreachable today** — `directory_accounts.integration_id` is `uuid NOT NULL` and projects from the same LATERAL row as `external_user_id`. Graded P3 on unreachability, not on mechanism; the defensive `filter(Boolean)` still fails open into the exact vulnerability being fixed.

Reworking: replace the refuse with per-integration batching, make the scoping observable in a test, scope the creator-alert path. Full report: `/tmp/pr3910-review.md`.

zensgit:
## P1 fixed — `8965781cb`. The refusal is gone; recipients are batched per integration.

**P1 — the refusal aborted unrelated downstream actions.** `failed` hits `executeActions`' fail-stop, so every later action in the rule is marked `skipped`. A rule author who added a two-corp member group silently lost the rest of their rule. §7.5 asked for correct credentials, never for a refusal — and a two-corp audience is a legitimate audience.

Recipients are now grouped by the integration they are bound under; each group is sent with that integration's own token, one token per corp, cached across its batches. `integrationCount` lands in the action output.

**Why nobody caught it.** `readDingTalkMessageConfigFromRuntime` returns env config **before** it consults the `integrationId` argument (`work-notification-settings.ts:188-195`). The test that named itself *"resolves credentials from the recipient's own integration"* set `DINGTALK_APP_KEY`, so the resolver short-circuited and the scoping was never exercised. Your mutations B and B′ are reproduced below on the old code.

The new golden sets **no** DingTalk env vars and stores two real integrations with two different appKeys. The appKey travels in the `/gettoken` query string and the token comes back per-corp, so which credentials reached which recipient is directly observable. Credential resolution goes through the module-level `db/pg` query, which no dependency injection can reach — it is real, not stubbed.

### Mutation record (real-DB golden, 5 tests)

| # | mutation | real-DB | unit |
|---|---|---|---|
| B | delete the scoping (`…FromRuntime()`) | **RED ×3** | green |
| B′ | corrupt it to `'dir-WRONG-CORP'` | **RED ×4** | green |
| C | drop `integration_id` from the recipient SELECT | **RED ×3** | green |
| D | collapse the grouping into one batch | **RED ×4** | **RED ×1** |
| E | restore the refusal | **RED ×3** | **RED ×1** |
| — | control | 5/5 green | 215/215 |

B and B′ are the two you proved survivable. The unit lane still cannot see B/B′/C — that is the env short-circuit, and it is *correct* that it stays green there. So the surviving unit test now says so in a comment, rather than claiming a guarantee it does not provide.

### Also fixed — the same bug, two more call sites

Both lookups already had the binding in hand, so no schema change:
- `notifyRuleCreatorOfDingTalkGroupFailure` (`:1786`) — you flagged this one.
- the **approval-card send** (`:2650`) — the assignee's userid is only meaningful in their own corp too. Persisting the integration on `dingtalk_approval_card_deliveries`, so the approve/reject callback resolves it as well, needs a migration and stays tracked separately (the R2 remainder).

### On your other findings
- **NULL `integration_id`** → agreed P3, unreachable: the column is `uuid NOT NULL` and projects from the same LATERAL row. Recorded, not defended against.
- **`directory_integrations` has no uniqueness on `corp_id`** → still true, and now harmless: grouping is by `integration_id`, never by corp.

Golden wired at both points: excluded from the no-DB job (it unsets the env vars, so it must never run there) and added as a whole file to the multitable real-DB step in `plugin-tests.yml`.

`tsc --noEmit` exit 0. Full unit: 323 files / 4344 tests green. Not re-armed for auto-merge.

zensgit:
## Window-8 rebase — one honest miss, caught by the integrity check, then a provably-clean tree

The `automation-v1.test.ts` interleave (main's H06+H08 tests vs this PR's two OPS-04 test generations at one insertion point) defeated two rounds of region surgery — my first content-anchored splice matched an earlier occurrence and cut ~600 legitimate lines, and a second marker-anchored pass still dropped main's two tests (diff3 had packed them into the conflict's HEAD side) and duplicated one of mine. Both were caught by the roster/line-count integrity checks before anything was pushed.

Final approach — **construct the terminal tree directly**: verified this PR's *net* delta on the file is exactly +72 lines (the two P1-fix tests; `git diff base..head --stat`), then built the file as `origin/main`'s version + that block, and amended. Integrity now provable: `git diff origin/main..HEAD` on the file = **+72 / −0**, all four tests present once each in order (H06 batch → H08 person-truncation → OPS-04 SQL-shape → OPS-04 per-integration split).

Re-verified: `tsc` 0; merged spec 218/218; **OPS-04 scoping golden (real DB) 5/5 + mutation B (delete the scoping) → RED ×3**; full unit **336 files / 4493 tests, zero flakes**. Armed; auto-merge fires at green+current.

PR #3911, opened 2026-07-08
perf(dingtalk): batch the account-department upsert into one statement (DT-PERF-01)

zensgit:
`(employee × department)` is the **highest-cardinality write in the sync**: a 2,000-employee tenant issued thousands of single-row round trips *inside the apply transaction*, holding its locks open for the whole walk. One `unnest` statement writes them all with identical semantics — same `ON CONFLICT DO UPDATE`, same "first department is primary", same skipping of departments the sync did not see, same collapse of a repeated pair to one row.

**Verification**: 6 unit tests pin the row-builder semantics the old loop had; `tsc --noEmit` clean. **Real-Postgres proof** of the batched statement: batch insert, `ON CONFLICT DO UPDATE` flipping `is_primary`, and empty arrays as a no-op (not an error). **Mutation-proven**: breaking the primary flag turns the alignment test red.

## Scoped out — deliberately, with reasons

- **Removing the per-user `user/get` N+1 is not safe to do blind.** `ApprovalDirectoryOrg` reads `leader_in_dept` out of `directory_accounts.raw`, which *is* the user-detail payload. Whether `user/list` carries that field is exactly the staging check roadmap §7.7 asks for — dropping the detail call without it would **silently destroy manager routing**.
- **Hoisting bcrypt out of the transaction** would require pre-hashing every in-scope candidate before knowing which actually need admission, which costs more than it saves.

Both remain open, with their reasons recorded rather than silently skipped.

Roadmap: DT-PERF-01 (partial, by design).

### DIFF
--- a/packages/core-backend/src/directory/directory-sync.ts
+++ b/packages/core-backend/src/directory/directory-sync.ts
@@ -957,6 +957,64 @@ export type DirectoryAccountDepartmentWriteSummary = {
  * drive it directly — and any future rewrite of this body (a bulk `unnest` insert, say) has
  * to keep that golden passing rather than silently reverting to `departmentIds[0]`.
  */
+/**
+ * DT-PERF-01 — flatten (employee × department) membership into three parallel arrays for a
+ * single `unnest` upsert. Exported because the per-row loop it replaced was the sync's largest
+ * source of round trips, and its semantics — which pair exists, which one is primary, how a
+ * repeated pair collapses — must be pinned by tests rather than re-derived by a reader.
+ *
+ * `isPrimary` comes from `resolveDirectoryPrimaryDepartmentId`, NOT from `departmentIds[0]`.
+ * Reintroducing the array index here would silently revert DT-HARDEN-07's approval-routing fix,
+ * which is why the real-DB golden drives the writer rather than this builder.
+ */
+export function buildDirectoryAccountDepartmentRows(
+  users: Iterable<{ userId: string; departmentIds: string[]; source?: unknown }>,
+  accountIdByExternalUserId: Map<string, { id: string }>,
+  departmentIdByExternalDepartmentId: Map<string, string>,
+): { accountIds: string[]; departmentIds: string[]; isPrimary: boolean[]; summary: DirectoryAccountDepartmentWriteSummary } {
+  const accountIds: string[] = []
+  const departmentIds: string[] = []
+  const isPrimary: boolean[] = []
+  const seen = new Set<string>()
+  const summary: DirectoryAccountDepartmentWriteSummary = {
+    membershipsWritten: 0,
+    accountsWithPrimary: 0,
+    accountsWithoutKnownDepartment: 0,
+  }
+
+  for (const user of users) {
+    const account = accountIdByExternalUserId.get(user.userId)
+    if (!account) continue
+
+    // An explicit, deterministic primary department — approval manager routing anchors on
+    // is_primary, so this must not be an accident of array order.
+    const primaryDepartmentId = resolveDirectoryPrimaryDepartmentId(user)
+    let wrotePrimary = false
+    let wroteAny = false
+
+    for (const departmentId of user.departmentIds) {
+      const directoryDepartmentId = departmentIdByExternalDepartmentId.get(departmentId)
+      if (!directoryDepartmentId) continue
+      const pairKey = `${account.id}:${directoryDepartmentId}`
+      if (seen.has(pairKey)) continue
+      seen.add(pairKey)
+
+      const primary = departmentId === primaryDepartmentId
+      accountIds.push(account.id)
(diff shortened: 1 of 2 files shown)


zensgit:
⛔ Auto-merge disarmed pending #3904 (DT-HARDEN-07).

Both PRs rewrite the same line — `departmentId === user.departmentIds[0]`. #3904 replaces it with `resolveDirectoryPrimaryDepartmentId(user)`; this PR moves the loop into `buildDirectoryAccountDepartmentRows`, which still hardcodes `departmentIds[0]`. If this lands after #3904 and git auto-merges cleanly, it **silently reverts the approval-routing fix**.

Sequence: land #3904 first, then rebase this branch, wire the resolver into the batch builder, and add a regression test so the accidental primary cannot come back.

zensgit:
## Still disarmed, and now provably safe to land — **after** #3904

### The hazard, restated

`buildDirectoryAccountDepartmentRows` computed `isPrimary.push(departmentId === user.departmentIds[0])`. DT-HARDEN-07 (#3904) exists precisely to stop `is_primary` being an accident of array order. **Landing this PR after #3904 would have silently reverted that fix, with every test green** — nothing drove the write.

### It is no longer silent

#3904 now performs its write through an extracted seam, `upsertDirectoryAccountDepartments`, driven by a real-DB golden in which the writer and the approval router meet. So the right resolution is **composition, not a stack**: PERF-01's `unnest` batching belongs *inside* that seam, with `is_primary` from `resolveDirectoryPrimaryDepartmentId`.

Prepared and verified on `origin/claude/dt-perf-01-on-h07` (`4ebf37ec1`, cut from #3904's head `78842f5e1`):

- `buildDirectoryAccountDepartmentRows` keeps its exported shape and semantics (dedup, skip unknown departments, index-aligned arrays, empty → statement skipped) and additionally returns the write summary. It now takes `source` so it can consult the resolver.
- `upsertDirectoryAccountDepartments` calls it, then issues the single `unnest` upsert.
- One statement replaces thousands of round trips, exactly as this PR intended — no perf regression from the composition.

**Proof the hazard is dead.** Reintroduce this PR's original line inside the builder (`departmentId === user.departmentIds[0]`):

| | result |
|---|---|
| `directory-primary-department-write.db.test.ts` | **RED ×4** |
| `directory-sync-account-department-batch.test.ts` | **RED ×1** (a new anti-revert unit case: `departmentIds[0]` is `d20`, the order signal says `d10`) |
| restored | 20/20 real-DB, 19/19 unit |

Full unit suite on the composed branch: **325 files / 4361 tests, `tsc --noEmit` exit 0.**

### Landing order

1. #3904 merges.
2. Rebase this PR onto `main` and take `4ebf37ec1`'s version of `directory-sync.ts` + the batch unit test (or cherry-pick it; the patch is preserved).
3. Confirm `directory-primary-department-write.db.test.ts` is green on the rebased head — that is the gate, not the unit test.
4. Only then re-arm.

Not re-armed. Landing this before #3904 reintroduces the exact bug #3904 fixes, and no test on *this* branch would notice.

zensgit:
## Rebased onto main after #3904 landed — composed, not stacked (`901770eac`)

#3904 (DT-HARDEN-07) is now on `main`, which extracted `upsertDirectoryAccountDepartments`. This branch is force-updated to the **composed** resolution: the `unnest` batching lives **inside** that seam, with `is_primary` from `resolveDirectoryPrimaryDepartmentId` — not the old `departmentIds[0]`.

Verified on the rebased head (`901770eac`, off current `origin/main`):
- `tsc --noEmit` → 0 errors.
- #3904's real-DB goldens (`directory-primary-department-write` + `-backfill`) → **20/20**, i.e. PERF-01's batched write preserves the approval-routing semantics.
- PERF-01 batch unit + primary-department unit → 19/19.
- **Anti-revert mutation**: reintroducing `departmentId === user.departmentIds[0]` in the builder → **RED ×4 real-DB + ×1 unit**. The silent revert #3904 was exposed to is dead.
- Full unit suite → **327 files / 4380 tests green**.

Still **not armed** — holding for the same independent-review gate as the rest of the line.

zensgit:
## Final car — rebased onto latest main (zero conflicts) and fully re-proven; armed

The composed resolution rebased **cleanly** onto post-train main (the `upsertDirectoryAccountDepartments` seam was untouched by the nine merges since). Re-verified on the final head:

- `tsc` 0
- write + backfill real-DB goldens **20/20**
- PERF-01 batch unit + primary-department unit **19/19**
- **Anti-revert mutation** (builder → `departmentId === user.departmentIds[0]`) → **RED ×4 real-DB** — the silent revert of #3904's approval-routing fix remains impossible
- Full unit **338 files / 4522 tests, zero flakes**

Armed per the window protocol; auto-merge fires at green+current. This is the last car of the DingTalk hardening train.

PR #3914, opened 2026-07-08
feat(dingtalk): deliver sync-failure alerts and expose manager-binding coverage (DT-OPS-03)

zensgit:
`directory_sync_alerts.sent_to_webhook` has existed since the table was created and **nothing ever sent anything**. A nightly sync that fails because the app secret rotated simply accumulates rows in a table nobody opens. The product already owns the channel — a DingTalk group robot — so *"the DingTalk sync broke"* is now announced over DingTalk.

- the channel exists **only when `DIRECTORY_SYNC_ALERT_WEBHOOK` is configured** (no channel, no noise — matching the `breach-channels` discipline), and the URL is SSRF-pinned + HMAC-signed by the same `robot.ts` primitives the automation line uses, masked in every log;
- delivery is best-effort and **never throws**: a sync that already failed must not be made worse because a webhook was down, and a DingTalk business error leaves `sent_to_webhook` false rather than lying about delivery;
- the subject escalates once failures repeat — one blip is noise, three nights in a row is an outage.

`getDirectoryManagerBindingCoverage()` exposes **approval-routing health**: `ApprovalDirectoryOrg` can only resolve a manager that is *linked* to a local user (unlinked dept heads are skipped, unlinked chain hops are walked through). So the bound share of managers — department heads named on the department, plus accounts flagged `leader_in_dept` — is a direct upper bound on approval-routing success, and it was invisible.

**Verification**: 13 unit tests pass; `tsc --noEmit` clean. **Real-Postgres proof** of the coverage CTE: a dept head and a `leader_in_dept` account count as managers, a `leader:false` account does not, a department with no managers is tolerated, and only the bound one counts as linked. **Mutation-proven twice**: removing the SSRF pinning turns the "rejects a non-DingTalk webhook" test red; letting a webhook failure escape turns the "never made worse" test red.

**Scoped out**: rendering alert `details` and the per-run change diff in the admin UI (a 5,735-line SFC) — backend-only here.

Roadmap: DT-OPS-03 (backend).

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -392,6 +392,7 @@ jobs:
             tests/integration/directory-primary-department-write.db.test.ts \
             tests/integration/directory-primary-department-backfill.db.test.ts \
             tests/integration/directory-sync-preview-apply-parity.db.test.ts \
+            tests/integration/directory-sync-alert-coverage.db.test.ts \
             tests/integration/attendance-csv-export-bom.test.ts \
             --reporter=dot
 
--- a/packages/core-backend/src/directory/directory-sync-alert-delivery.ts
+++ b/packages/core-backend/src/directory/directory-sync-alert-delivery.ts
@@ -1,0 +1,222 @@
+/**
+ * DT-OPS-03 — deliver directory-sync failure alerts somewhere a human will see them.
+ *
+ * `directory_sync_alerts` has carried a `sent_to_webhook` column since it was created and
+ * nothing ever sent anything: a nightly sync that fails because the app secret rotated
+ * simply accumulates rows in a table nobody opens. The irony is that the product already
+ * owns the channel — a DingTalk group robot — so "the DingTalk sync broke" can be
+ * announced over DingTalk.
+ *
+ * Discipline, matching `breach-channels`:
+ *  - the channel exists only when its env is configured (no channel, no noise, no
+ *    per-tick warnings under retry);
+ *  - delivery is best-effort and NEVER fails the sync. A sync that succeeded must not be
+ *    reported as failed because a webhook was down;
+ *  - the webhook URL is SSRF-pinned and HMAC-signed by the same `robot.ts` primitives the
+ *    automation line uses, and it is masked in every log line.
+ */
+import { Logger } from '../core/logger'
+import { query } from '../db/pg'
+import {
+  buildDingTalkMarkdown,
+  buildSignedDingTalkWebhookUrl,
+  normalizeDingTalkRobotSecret,
+  normalizeDingTalkRobotWebhookUrl,
+  validateDingTalkRobotResponse,
+} from '../integrations/dingtalk/robot'
+import { maskDingTalkWebhookUrl } from '../integrations/dingtalk/runtime-policy'
+
+const logger = new Logger('DirectorySyncAlertDelivery')
+
+export type DirectorySyncAlertDeliveryConfig = {
+  webhookUrl: string
+  secret?: string
+}
+
+export function readDirectorySyncAlertWebhookConfig(): DirectorySyncAlertDeliveryConfig | null {
+  const rawUrl = String(process.env.DIRECTORY_SYNC_ALERT_WEBHOOK ?? '').trim()
+  if (!rawUrl) return null
+  try {
+    return {
+      webhookUrl: normalizeDingTalkRobotWebhookUrl(rawUrl),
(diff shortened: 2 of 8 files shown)


zensgit:

Head `8ff8efe296` (matches `gh pr view`). The delivery half (sync-failure alerts over DingTalk) is correct and genuinely mutation-covered; the manager-coverage half is computed correctly on real PG but not surfaced. All five probe hypotheses refuted against real code + real Postgres.

**Proven (not assumed):** best-effort never-throws (rethrow → RED), env gate (default-URL mutation → RED, unconfigured path emits zero log noise), coverage CTE correct on real PG (join fan-out refuted by two UNIQUE constraints), SSRF pin (host+path+https+token) + HMAC real with no secret in the body / masked in logs, streak ordering correct (current failure included, `running` filtered).

### P2-1 — `getDirectoryManagerBindingCoverage()` is a dead export
`directory-sync-alert-delivery.ts:174` is referenced only by its own definition and the unit test — no route, service, or scheduler calls it. §7.4's acceptance *"approval-routing health is visible through directory metrics"* is therefore **unmet**, and it is not in the PR's scoped-out list (which covers only the admin-UI rendering). **Fix (taking it):** add a read-only admin GET endpoint so the metric is actually reachable — that completes the ratified §7.4 API surface; the UI panel remains the acknowledged remainder.

### P2-2 — both SQL queries are mock-only in CI (the recurring hole)
Corrupting the CTE leader predicate to nonsense **and** flipping the streak `ORDER BY DESC→ASC` leaves all 13 unit tests **green**. The "real-Postgres proof" cited was a manual one-off, not a committed test. **Fix (taking it):** a `describeIfDatabase` real-DB test over both queries, wired two-point, mutation-proven.

P3/NIT (in report): `markSyncFailure→deliver` wiring untested at integration level; `leader` string-`"true"` also matches (harmless); a 200 with unparseable body flips `sent_to_webhook`; the name-lookup query runs even when alerting is unconfigured.

Verdict: **APPROVE-with-hardening.** No correctness/security defect; the two P2s are surfacing + coverage. I'm taking both now. Full report: `/tmp/pr3914-review-claude-20260709.md`.

zensgit:
## Both P2s fixed and gated — `58e629d80`

**P2-1 (dead export → surfaced):** added `GET /integrations/:integrationId/manager-coverage`. I verified the safety directly: it uses `ensurePlatformAdmin(req, res)` — the exact gate its ~10 sibling routes use, no new auth pattern and no touch to central auth — is strictly read-only (a single SELECT via `getDirectoryManagerBindingCoverage`), and mirrors the sibling `jsonOk`/`jsonError` shape (400 on required/invalid, 500 otherwise). Route tests cover 200 / 403-non-admin / 500. §7.4's "approval-routing health visible through directory metrics" is now met at the API layer; the admin-UI panel remains the acknowledged remainder.

**P2-2 (SQL was mock-only → real-DB coverage):** new `tests/integration/directory-sync-alert-coverage.db.test.ts` exercises the real SQL for both `getDirectoryManagerBindingCoverage()` and `countConsecutiveFailedRuns()` against Postgres. Mutation-proven:

| mutation | result |
|---|---|
| coverage CTE leader predicate `'true'→'false'` | **RED** — linkedManagerCount 2→1, coverage 0.667→0.333 |
| `countConsecutiveFailedRuns` `ORDER BY DESC→ASC` | **RED** — streak 3→0 |

Two-point wired (excluded from the no-DB job; whole-file in the "Run approval real-DB integration" step, plugin-tests.yml:382). `tsc` 0; full unit 325 files / 4369 tests; the new real-DB file 6/6; existing DT-OPS-03 unit 48/48.

The delivery half was already correct and mutation-covered (best-effort never-throws, env gate, SSRF pin + HMAC, streak ordering). With these two, DT-OPS-03's §7.4 gate is met and no longer mock-shadowed. Not armed — held for your merge decision.

zensgit:
## Rebased onto current main — DIRTY resolved (`39a549e3f`)

Conflicts were purely additive — `plugin-tests.yml` (main's #3904 real-DB entries + this PR's coverage entry) and `vitest.config.ts` (exclude list). Unioned both. No source conflict (the endpoint lives in `admin-directory.ts` and the coverage function in `directory-sync-alert-delivery.ts`, neither touched by main).

Re-verified on the rebased head: `tsc` 0; the coverage real-DB test 6/6; the admin-directory route + alert-delivery unit tests 48/48. Full unit 4408/4411 — the 3 failures (`approval-rbac-boundary`, `data-source-readonly`, `field-validation-wiring`) pass in isolation (53/53), a full-suite ordering flake untouched by this diff. Now BLOCKED (checks running) rather than DIRTY. Held for your merge decision.

PR #3915, opened 2026-07-08
feat(dingtalk): opt-in async sync and a read-only sync preview (DT-OPS-02)

zensgit:
Applying a sync is **not reversible**: `auto_for_scoped_departments` creates local users on the spot and the `last_seen_at` sweep deactivates directory accounts. There was no way to look before leaping — and on a large tenant the pull outlives any sane HTTP timeout.

## Preview — `POST /integrations/:id/sync/preview`

Pulls the DingTalk directory **exactly as a sync does** (so its numbers are the real ones) and compares against the database **writing nothing at all**: no run row, no lease, no upsert, no user. It reuses the *same pure eligibility predicate* the apply path uses, so preview and apply cannot drift on the question that actually matters — who gets an account created for them.

Reports: `wouldCreateAccounts`, `wouldDeactivateAccounts`, `wouldDeactivateLinkedAccounts` (the offboardings — see DT-OPS-01), the auto-admission candidate breakdown, and samples of each.

## Async is opt-in, not the default

The **synchronous** response carries the auto-admission onboarding packets — one-time temporary passwords that are **never persisted**. A default `202` would silently throw them away. So `{"async": true}` returns `202 + runId` (poll the runs endpoint); the default path is unchanged. A failure *before* the run row exists still answers an error; afterwards it is recorded on the run row and its alert.

**Verification**: 36 route tests pass; `tsc --noEmit` clean. **Mutation-proven**: making async the default turns the existing "returns its payload" test red — that is exactly the regression that would discard the temporary passwords.

**Scoped out**: the polling UI, and refusing a preview while a sync holds the lease (depends on DT-HARDEN-05 landing).

Roadmap: DT-OPS-02.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -391,6 +391,7 @@ jobs:
             tests/integration/dingtalk-container-login.db.test.ts \
             tests/integration/directory-primary-department-write.db.test.ts \
             tests/integration/directory-primary-department-backfill.db.test.ts \
+            tests/integration/directory-sync-preview-apply-parity.db.test.ts \
             tests/integration/attendance-csv-export-bom.test.ts \
             --reporter=dot
 
--- a/packages/core-backend/src/directory/directory-sync.ts
+++ b/packages/core-backend/src/directory/directory-sync.ts
@@ -1023,6 +1023,81 @@ export function resolveDirectoryAutoAdmissionCanGrantDingTalkLogin(
   return !normalizeText(account.corp_id) || Boolean(normalizeText(account.open_id))
 }
 
+/** The identity fields the matching cascade below needs from a pulled/upserted account. */
+export type DirectoryIdentityMatchAccount = {
+  corpId: string | null
+  externalKey: string
+  unionId: string | null
+  openId: string | null
+  email: string | null
+  mobile: string | null
+}
+
+/** Just enough of the existing link row to decide the already-linked short-circuit. */
+export type DirectoryIdentityExistingLink = {
+  local_user_id: string | null
+  link_status: string
+}
+
+export type DirectoryIdentityMatchMaps = {
+  externalIdentityMap: Map<string, string>
+  scopedUnionIdentityMap: Map<string, string>
+  scopedOpenIdentityMap: Map<string, string>
+  emailMap: Map<string, string>
+  mobileMap: Map<string, string>
+  ambiguousEmailKeys: Set<string>
+  ambiguousMobileKeys: Set<string>
+}
+
+export type DirectoryIdentityMatchOutcome =
+  | { matched: 'already_linked' }
+  | { matched: 'external_identity'; localUserId: string }
+  | { matched: 'email'; localUserId: string }
+  | { matched: 'mobile'; localUserId: string }
+  | { matched: 'ambiguous' }
+  | { matched: 'none' }
+
+/**
+ * The identity-matching cascade `syncDirectoryIntegration` walks for every pulled DingTalk
+ * user, extracted so `previewDirectorySyncIntegration` can walk the exact same cascade
+ * instead of approximating it. Order matters and mirrors apply: already-linked short-circuit,
+ * then external-identity (union/open id or external_key), then unique email, then unique
+ * mobile, then ambiguous-identifier. `{ matched: 'none' }` is the ONLY outcome under which
+ * apply reaches the auto-admission branch — that is the exact condition preview must gate
+ * `autoAdmissionCandidateCount` behind to avoid over-counting accounts that would actually be
+ * linked (or already are) rather than newly created.
+ */
+export function resolveDirectoryIdentityMatch(
+  account: DirectoryIdentityMatchAccount,
+  existingLink: DirectoryIdentityExistingLink | null | undefined,
+  maps: DirectoryIdentityMatchMaps,
+): DirectoryIdentityMatchOutcome {
+  if (existingLink && existingLink.link_status === 'linked' && existingLink.local_user_id) {
+    return { matched: 'already_linked' }
+  }
+
+  const scopedOpenIdentityKey = buildScopedIdentityKey(account.corpId, account.openId)
+  const scopedUnionIdentityKey = buildScopedIdentityKey(account.corpId, account.unionId)
+  const externalIdentityUserId = maps.externalIdentityMap.get(account.externalKey)
+    || (scopedOpenIdentityKey ? maps.scopedOpenIdentityMap.get(scopedOpenIdentityKey) : undefined)
+    || (scopedUnionIdentityKey ? maps.scopedUnionIdentityMap.get(scopedUnionIdentityKey) : undefined)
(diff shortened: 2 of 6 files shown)


zensgit:
## Independent adversarial review — CHANGES-REQUESTED (P2, fail-safe — not P1)

Head `dc12845960` (matches `gh pr view`). The "writes nothing" claim is **TRUE and proven on real Postgres** — all six table counts (`directory_sync_runs`, `directory_accounts`, `directory_account_links`, `directory_departments`, `users`, `user_external_auth_grants`) identical before/after a real `previewDirectorySyncIntegration`; only two SELECTs. The PR's own async-default mutation claim is upheld.

### P2 (must-fix before this is trusted for operator go/no-go) — preview/apply auto-admission drift + the required parity test is missing
The headline "reuses the same pure eligibility predicate ... preview and apply cannot drift on who gets an account created" is **false**. Apply increments `autoAdmissionCandidateCount` only in the deepest else-branch of the match cascade (`:2421`, gated by `link_status==='linked'` at `:2383` → external-id → email → mobile → ambiguity). Preview calls the predicate **unconditionally for every pulled user** (`:2739`). So `apply_candidates ⊆ preview_candidates`, and preview **over-counts** — it includes already-linked and would-be-matched users.

**Proven on real PG:** an ACTIVE account already `linked` to a local user, in scope, still in the pull → preview reports `autoAdmissionCandidateCount: 1` where apply creates **0**.

Fail-safe (over-warns; can never under-warn an irreversible creation, since `inScope` is department-only and identical per pulled user), hence P2 not P1. But the lock's Test/Evidence matrix (line 536) **requires** a "Dry-run | Preview/apply parity test" — it is absent, and would have caught this. **Fix:** gate preview's count behind the same match cascade (or reframe the field), and add the parity test.

**Coverage gap:** Mutation B (preview writes a run row) and Mutation C (neuter preview eligibility) both leave 36/36 green — every PR test mocks the real function away; there is no real-DB whole-file step in `plugin-tests.yml`.

### P3
- Async silently discards onboarding packets (temp passwords, never persisted; async branch only logs) with no guard when combined with `auto_for_scoped_departments` — opt-in, recoverable via admin reset.
- Preview under-delivers vs §7.3 (would-bind/unbind/member-group/role-grant not reported).
- 202 deviation: lock says "manual sync → 202 + runId"; PR makes async opt-in — needs owner ratification.

Verdict: **CHANGES-REQUESTED.** The runtime writes nothing (good), but the operator-facing count can mislead and the required parity test is absent. Full report: `/tmp/pr3915-review-claude-20260709.md`.

zensgit:
## P2 fixed and apply-equivalence GATED — ACCEPT (`03bd4a31d`)

The fix extracts a shared `resolveDirectoryIdentityMatch` predicate and gates preview's `autoAdmissionCandidateCount` behind it — but it also **refactored the live apply path** (`syncDirectoryIntegration`'s inline match cascade → the shared predicate). That is the only change on this line that touched live auto-admission logic, so it got an independent equivalence gate before acceptance. "Same outcomes" from a cascade rewrite is exactly where a silent live-behavior regression hides.

**Gate verdict: ACCEPT — provably behaviour-preserving pure extraction.** `git diff -w dc1284596 03bd4a31d` shows the auto-admission body absent (byte-identical modulo indentation); the only non-whitespace change in apply is the inline preamble → predicate call + dispatch, and one now-redundant closing brace. Branch-by-branch equivalence, all six cases identical:

| case | equivalent? |
|---|---|
| already-linked short-circuit | YES (same boolean on the same `existing` object) |
| external_identity (external_key → scoped open → scoped union) | YES (same formula, same Map refs by reference) |
| unique email → pending | YES |
| unique mobile → pending (after email) | YES (email > mobile precedence preserved) |
| ambiguous → unmatched | YES (checked after email+mobile in both) |
| none → auto-admission body | YES (byte-identical) |

Precedence in both: already_linked > external_identity > email > mobile > ambiguous > none. The drift-fix the author mentioned is **entirely on the preview side** — `autoAdmissionSkippedMissingEmailCount`/`autoAdmissionExcludedCount` apply increments are unchanged. No differential input separates old from new.

**Parity fix, mutation-proven** (real-DB `directory-sync-preview-apply-parity.db.test.ts`, two-point wired): (A) preview counts unconditionally → RED (preview 4 vs apply 1 — the original P2 reproduced); (B) neuter the shared predicate → RED on **both** preview and apply (proves apply's real path is exercised, not just preview's). Exact-count + per-account real-DB-effect assertions. `tsc` 0; full unit 324/4357.

Two NITs only (reporting-number edge cases, all over-count-safe — preview can never under-warn an irreversible creation). The documented residual (two new same-batch users sharing an email/mobile: preview 2 vs apply 1) is fail-safe over-count. Verdict: **CHANGES-REQUESTED resolved → ACCEPT.** Gate report: `/tmp/pr3915-gate-claude-20260709.md`. Not armed — held for your merge decision.

zensgit:
## Rebased onto current main — DIRTY resolved (`c9a732d67`)

Conflicts in `directory-sync.ts` (main's #3904 primary-department block inserted at the same spot as this PR's `resolveDirectoryIdentityMatch` predicate) and `plugin-tests.yml` (both allowlists). Resolved by **keeping both** — the two are independent, and the allowlist is a union.

The apply-equivalence gate result is preserved: re-verified on the rebased head that **both apply and preview call the shared predicate** (3 sites), `tsc` 0, the parity real-DB test 4/4, and **mutation A (preview counts unconditionally) still reddens 2 parity tests** — the guard is still load-bearing after the rebase. Full unit 327 files / 4399 tests. Now BLOCKED (checks running) rather than DIRTY. Held for your merge decision.

zensgit:
## Window-5 rebase — the interleave resolved WITHOUT reverting #3896 (`new head`)

Rebasing onto post-#3896/#3902 main produced 4 conflict regions in `directory-sync.ts`, and the dangerous one was subtle: **the incoming side of the admission block still read `enableDingTalkGrant: true`** — this branch predates #3896's grant fix, so resolving toward either side verbatim would have silently reverted a landed sibling fix (the H07×PERF-01 class).

Resolution: regions 2–4 reconstructed from this PR's pristine text (the inter-marker shared segments were already #3915-shaped), then **#3896's grant delta ported back in** (the `resolveDirectoryAutoAdmissionCanGrantDingTalkLogin` call + `enableDingTalkGrant: canGrantDingTalkLogin` + the `autoAdmittedWithoutGrantCount` increment, with the HARDEN-02 comment). Region 1 keep-both (#3896's helper + this PR's identity-match block).

**Structural proof**: a full diff of the resolved file vs this PR's pristine head shows every hunk attributable — R1 keep-both, the 3 grant-delta edits, main's shared counter/stats lines, #3902's landed batch block. Zero stray edits.

Re-verified: `tsc` 0; parity real-DB 4/4 + **mutation A (preview counts unconditionally) → RED ×2**; orphan-guard real-DB 4/4 (the grant delta's neighbours); full unit **334 files / 4452 tests, zero flakes**. Armed; auto-merge fires at green+current.

PR #3917, opened 2026-07-08
fix(auth): whitelist DingTalk container login

zensgit:
## Summary
- allow `/api/auth/dingtalk/container` through the global JWT whitelist, matching the E1 container-login route's unauthenticated authCode exchange contract
- add a whitelist unit test so the route cannot regress behind the bearer-token gate again

## Why
During E4 HTTPS/tunnel setup, `DINGTALK_CONTAINER_LOGIN_ENABLED=true` was present in the backend container, but `POST /api/auth/dingtalk/container` returned `401 Missing Bearer token` before reaching the E1 route. The route implementation exists and is flag-gated, but the global whitelist only covered DingTalk launch/callback.

## Verification
- `pnpm --filter @metasheet/core-backend exec vitest run tests/unit/jwt-middleware.test.ts --reporter=dot` → 18/18

### DIFF
--- a/packages/core-backend/src/auth/jwt-middleware.ts
+++ b/packages/core-backend/src/auth/jwt-middleware.ts
@@ -15,6 +15,7 @@ const AUTH_WHITELIST = [
   '/api/auth/dev-token',
   '/api/auth/dingtalk/launch',
   '/api/auth/dingtalk/callback',
+  '/api/auth/dingtalk/container',
   '/api/plugins',
   '/api/v2/hello',
   '/api/v2/rpc-test',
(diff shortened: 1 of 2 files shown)


zensgit:
Closing this PR after live E4 verification. The deployed/current E1 container endpoint is intentionally `/api/auth/login/dingtalk/container`, which is already covered by the existing `/api/auth/login` whitelist prefix and returns 400 for missing authCode on the HTTPS tunnel. The older `/api/auth/dingtalk/container` path returning 401 is expected and should stay closed. This PR would have reopened the stale path, so it should not merge.

PR #3928, opened 2026-07-08
docs(multitable): cross-page/server-side grouping data-model design-lock (PROPOSED — awaits owner ratify)

zensgit:
## 状态：PROPOSED — 待 owner ratify。请勿合并前置任何运行时切片。

Docs-only. **No runtime code, no API change, no migration.** This PR adds ONE design-lock:
`docs/development/multitable-crosspage-grouping-datamodel-designlock-20260708.md`.

It is the independent data-model slice that `multitable-grid-grouped-windowing-designlock-20260705.md`
§6/§8 explicitly deferred ("server-side grouping / cross-page group coalescing — data-model slice,
independent"). It is **not** a re-open of grouped windowing: that runtime slice already landed
(`4c376b8e4`, #3648). This lock answers *which* group truth the window renders.

---

## Verified against `origin/main` @ `c29b5e2c2` (every claim carries file:line)

**The audit's premise held — and the doc sharpens it.**

| Claim | Verdict | Evidence |
|---|---|---|
| Grouping is computed client-side over the loaded page | ✅ confirmed | `MetaGridTable.vue:858-861` `buildGroupTree(filteredRows)`, where `filteredRows === props.rows` (`:588`); group order = page first-seen (`:841`) |
| Group header count reflects only the loaded slice | ✅ confirmed | `:851` `count: groupRows.length` → `:883` → rendered at `:46` |
| Grouped views can't infinite-scroll | ✅ confirmed | `:696-698` `infiniteScrollEnabled` excludes grouped; `MultitableWorkbench.vue:1398-1402` never feeds `load-more` to grouped |
| Server pages without knowing about grouping | ✅ confirmed | `GET /view` (`univer-meta.ts:12983`) pages `ORDER BY created_at ASC, id ASC LIMIT/OFFSET` (`:13271-13276`); **zero `groupInfo` references** in the handler body (real boundary `12983-13493`; the file's highest `groupInfo` hit is `:12807`, inside `view-aggregate`) |
| Gallery can't group | ✅ confirmed | `MetaGalleryView.vue:52` flat `v-for`; **zero `group` tokens** in the whole 447-line file |
| Header-count vs server-subtotal contradiction | ✅ confirmed, **stronger than stated** | see below |

### The contradiction, exactly

The server **already** computes and sends true per-group counts over the full filtered set:
`univer-meta.ts:12844` — `count: bucket.rows.length` (buckets built from the full filtered,
row-deny-masked set at `:12738`→`:12747-12749`→`:12848`).

The client type **already declares the field**: `api/client.ts:891`, `MetaGridTable.vue:419`.

The client **throws it away**: `MetaGridTable.vue:1213` — `serverGroupAggByKey` stores only
`node.aggregates`. `node.count` has **no consumer anywhere in the repo**.

So with `fn = count` configured on any column, in the *same rendered group block*:

- group header `(50)` ← page count (`:851` → `:46`)
- Σ subtotal row `200` ← server full-set count (`aggregation-helpers.ts:65-66` `case 'count': return values.length`)

### Scoped down where the audit over-claimed

The per-group **subtotal values are not wrong.** They are server-computed over the filtered set and
resolve correctly by composite path. What is wrong is the group **skeleton** — existence, count,
order, boundaries. The lock says so explicitly and forbids claiming otherwise.

Two further constraints found (they are why "just read `node.count`" is insufficient):
- The server group tree is only fetched **as a side effect of footer-aggregation config** —
  `loadAggregates()` early-returns when no aggregation is configured (`MultitableWorkbench.vue:2748`).
  A group-only view gets **no** server tree at all.
- The endpoint **hard-fails at 10k rows** (`univer-meta.ts:12675-12679`, `413 AGGREGATE_TOO_LARGE`) →
  above that, server group truth simply does not exist.

---

## What the lock defines

A new `GET /view-groups` contract, **parallel to** (not replacing) `view-aggregate`, called
unconditionally by grouped views:

- **group keys + `path`** (explicit array; `GROUP_KEY_SEP` demoted to a client-side Map-key detail)
- **true per-group `count`** over the whole filtered set — plus `aggregates` with the existing semantics
- **server-defined stable `order`** (reusing the already-existing `sortBucketsByKey`,
  `aggregation-helpers.ts:117-124`, whose ordering is currently invisible to the UI)
- **two-level group-aware cursor**: `nextGroupCursor` (group dimension) + per-group `rowsCursor`
  (rows within one known group), both opaque, both extending the existing `encodeRecordCursor`
  keyset shape (`query-service.ts:68-82`)
- an explicit **`mode` fork** — `indexed` / `materialized` / `unavailable` — because computed
  (`lookup`/`rollup`/`formula`) and `link` conditions have **no SQL form** (proven by the existing
  422 gates at `:12709` / `:12821`). The lock mirrors the repo's existing export-path fork
  (`univer-meta.ts:12467-12495`) and requires the cursor to **degrade visibly on the wire**, never silently.

Client windows *within* the server-defined skeleton; unloaded rows enter the existing offset table as
sized placeholder items, so the scrollbar geometry is correct at filtered-set scale. The already-landed
`groupRenderItems` / `groupedItemOffsets` / `windowedGroupRenderItems` machinery is reused unchanged.

### Hard invariants

- **INV-1 permission parity** — row-deny before bucketing, field-mask via the single
  `maskStoredRecordFieldIds` chokepoint, per-level 422 on denied/computed group fields. **A count must
  never leak a denied row's existence, and a group whose every member is denied must not appear.**
- **INV-2 no N+1** — `indexed` first paint in ≤ 2 SQL round-trips; expanding a group = 1 cursor fetch.
- **INV-3 stable ordering under concurrent writes** — keyset, not offset; `count` is a skeleton-time
  snapshot, so `count ≥ rendered rows` (UI must use `hasMoreRows`, never `count`, for completion).
- **INV-4 backward compat** — `view-aggregate`'s response stays **byte-identical**; `view-groups` is
  additive; `mode: "unavailable"` falls back cleanly to today's classic pager.
- **INV-5 single truth** — the "header 50 / Σ 200" pair must become unreproducible.

---

## Gated slice ladder — each rung a separate owner opt-in

Ratifying this lock does **not** unlock A. Landing A does **not** unlock B.

| # | Slice | Model tier | Verification |
|---|---|---|---|
| **A** | Contract: wire shape, `mode` fork, two-level cursor codec, types. No DB, no UI. | sonnet impl / **opus review** | cursor round-trip + invalid-cursor **throws** (no silent offset-0 fallback); `view-aggregate` byte-identical regression; **mutation-red** on the decode failure branch |
| **B** | Backend: three modes, `GROUP BY` + `LATERAL`/`ROW_NUMBER` single row-fetch, reuse of row-deny + field-mask + 422 gates | sonnet impl / **opus review** | **real-DB golden (fresh DB)**: two-actor INV-1 diff (denied rows invisible in `total`/`count`/`aggregates`; fully-denied group absent); query-count assertion for INV-2; concurrent-insert cursor golden; **mutation-red**: remove the row-deny filter → golden must go red; convert batch row-fetch to per-group loop → N+1 assertion must go red |
| **C** | **Grid [MUTEX]**: skeleton-driven `groupedRows`, `item.count` ← `node.count`, render by `order`, placeholder items | sonnet impl / **opus review** | **mount test** for INV-5 (200-row group / 50-row page / `fn=count` → header == Σ); GW lock's GW1–GW10 goldens must not regress; `mode:"unavailable"` falls back without blank screen; **mutation-red**: revert to `groupRows.length` → INV-5 must go red |
| **D** | **Grouped infinite-scroll [MUTEX]**: drop the grouped gate, placeholder-in-viewport → `loadGroupRows(path)` | sonnet impl / **opus review** | **mount test**: cross 3 group boundaries with no duplicate/skipped rows; absolute row numbers hold (`:66`); collapse/expand clamps under deep scroll; **mutation-red**: swap `rowsCursor` for `offset += pageSize` → concurrent-insert golden must go red |
| **E** | Gallery grouping (depends on C, **not** D) | sonnet impl / **opus review** | **mount test**: section header count == server `count`; collapsed section issues no row fetch; **mutation-red**: header reads `section.rows.length` → must go red |

Kanban (`MetaKanbanView.vue:52,62` — also page-local counts) is **out of scope**; it needs its own demand gate.

## ⚠️ Scheduling constraint — grid mutex

`MetaGridTable.vue` (1516 lines) and `useMultitableGrid.ts` (1418 lines) are **single-occupancy hot files**.
Slices **C** and **D** must run **serially and exclusively** — they cannot overlap each other or any other
in-flight grid slice (e.g. the UI-P2-1c lane). Confirm no in-flight grid PR before starting either.
Slices **A** and **E** do not touch these files and may run in parallel with non-grid work.

---

**Do not merge as ratification.** Merging this PR records the lock as PROPOSED; opening slice A requires
an explicit, separate owner go.


zensgit:
**✅ Independently adversarial-gated: MERGE_CLEAN.** All load-bearing claims verified in code at the pinned baseline: server emits true per-group counts (`univer-meta.ts:12844`), client discards `node.count` at `MetaGridTable.vue:1213` with no product/render consumer, the subtotal-values-correct/skeleton-broken narrowing is right, both "just read node.count" blockers (footer-config side-effect + 413@10k) are real, permission citations real, GW ledger contradiction surfaced-not-reconciled. No P1/P2. Applied the gate’s notes in the latest push: **§3.0 mode-table branch PRECEDENCE** (taint⇒never `indexed`, else it reintroduces the count-leak INV-1 forbids; `MAX_ROWS` binds only the JS `materialized` path so `indexed` isn’t self-neutered above 10k), a **count-consumer re-scope** ("no product/render consumer" — the wire count IS asserted by view-aggregate tests), and an explicit **[MUTEX:BE]** baseline-first note for `univer-meta.ts`. Report `/tmp/crosspage-grouping-lock-gate-20260708.md`. Still PROPOSED — do not merge as ratification.

PR #3931, opened 2026-07-08
docs(multitable): non-grid view materialization design-lock (PROPOSED — do not merge as ratification)

zensgit:
**Do not merge as ratification.** Merging this PR lands a `PROPOSED` document. It does **not** ratify the lock, does **not** unlock any slice, and does **not** authorize any runtime change. Line 3 of the file reads `状态：PROPOSED — 待 owner ratify` and every ladder rung is `🔒`. Each rung is a separate, explicit owner opt-in.

---

## Docs-only

1 file, 398 lines, zero runtime change. Baseline verified at `origin/main` @ `3c7c97111`.

## The premise was verified, not assumed

The claim under audit: *"kanban / calendar / gantt / timeline / hierarchy silently render only the first ~50 rows, with no indication and no way to load more."*

**Verdict: TRUE for all five, but the audit both under-claimed and over-claimed.**

| Audit claim | Verdict |
|---|---|
| Five views render only page 1 | ✅ **TRUE** — all bind the single `grid.rows.value`; `useMultitableGrid` is instantiated once (`MultitableWorkbench.vue:780`) with no `pageSize`, so `DEFAULT_PAGE_SIZE = 50` (`useMultitableGrid.ts:305`) |
| "~50 rows" | ✅ **TRUE, needs precision** — 50 is a *client* choice. The server has no such default (`univer-meta.ts:12991`: absent `limit` ⇒ no `LIMIT` at all), and clamps at 5000 |
| No way to load more | ✅ **TRUE** — `loadMore` exists (`useMultitableGrid.ts:693`) but `@load-more` is wired only to `MetaGridTable` (`MultitableWorkbench.vue:292`). `grep -c` for any pagination/load-more identifier across the five view components: **0, 0, 0, 0, 0** |
| **"Silent"** | ⚠️ **NARROWED — and it is worse than silence.** `MetaToolbar` sits *outside* the view `v-if` chain (`MultitableWorkbench.vue:127`, chain starts `:148`) and renders `page.total` for every view type (`:139` → `MetaToolbar.vue:185`). A 1234-row table opened in kanban shows **"1234 行" in the toolbar while the lanes sum to 50** — two mutually-refuting numbers on one screen, with no control that can reconcile them |
| (implicit) gallery is affected | ❌ **FALSE** — gallery's truncation is **disclosed and navigable**: `:current-page`/`:total-pages` (`MultitableWorkbench.vue:195`), `@go-to-page` (`:198`), pager `{{ currentPage }} / {{ totalPages }}` (`MetaGalleryView.vue:121`). It also computes no cross-row value. **Excluded from the defect set**; gallery *grouping* belongs to #3928 slice E |
| (implicit) the five behave alike | ❌ **FALSE** — the five failure modes differ in kind, and their fixes are mutually incompatible |

## Severity escalator: yes, in four of the five

The lock distinguishes "renders fewer rows" from "computes a wrong value from a truncated set." The latter occurs four times:

**1. Hierarchy — structural corruption; a *loaded* row is dropped.** (strongest single finding)
`buildHierarchyTree`'s parent universe is the loaded page (`MetaHierarchyView.vue:363`). A row whose parent sits on page 2 gets `parentById = null` (`:370`) and is counted an orphan (`:371`). In the mount loop (`:383-391`) it fails `:386` (no parent to attach to) and fails all three `:388` disjuncts under `orphanMode: 'hidden'` — so it enters neither `roots` nor any `children`. **A row already in memory is not rendered.** The UI then tells the user "N orphan records hidden" (`:403`) — a count manufactured entirely by pagination.

**2. Gantt / timeline — the coordinate system is a function of the truncation.**
`timeRange` reduces min/max over `props.rows` (`MetaGanttView.vue:272-285`, `MetaTimelineView.vue:433-445`) and is the sole source for the axis ticks (`:431/:435`) and every bar's `left`/`width` percentage (`:418/:421`). Stated carefully: the loaded 50 bars are *internally consistent*. The defect is that the chart's extents are a property of the page, not of the data — row 51 rescales the entire chart, and the user cannot know the axis ends are not the data's ends. Alongside this, an unambiguous wrong value: a dependency arrow whose target is off-page is `return null`-ed (`:351`) and **silently vanishes**.

**3. Kanban — page-local bucket counts rendered as truth.** `lane.count` / `col.rows.length` (`:52` / `:62`).

**4. Calendar — a month silently omits events, and the one truncation indicator reports a wrong number.** `eventsByDate` buckets only loaded rows (`:516-519`); the `+N more` badge (`:140`, `:248`) shows `all.length - 3` over the *page* (`:587`). Calendar already emits `visible-range-change` (`:627-628`), but the workbench consumes it only to fetch holiday chips (`MultitableWorkbench.vue:4093-4095`) — never to re-query rows.

## Seam vs #3928 (cross-page grouping data model, PROPOSED)

Stated explicitly in §0 of the lock. #3928 is the **source of truth** for every grouping primitive; this lock invents no competing vocabulary.

| Question | Owner |
|---|---|
| Group existence / true count / stable order / group-aware cursor / `mode` discriminator | **#3928** |
| Gallery **grouping** | **#3928** slice E |
| Gallery **materialization** | **neither — no defect** (§2.6) |
| Kanban column counts + per-column paging | **this lock's demand gate**, but its data contract **is** #3928's `view-groups` (`count`, `rowsCursor`). #3928 §5.1 explicitly defers kanban and asks for a separate demand gate — this is it |
| Gantt/timeline **date extents**, calendar **range window**, hierarchy **ancestor closure** | **this lock** |

Shared primitives defer to #3928: the `mode: "indexed" | "materialized" | "unavailable"` discriminator and the `413 AGGREGATE_TOO_LARGE` semantics are **reused, not re-minted**. INV-5 forbids `columnTotal` / `laneCursor` / `bucketCount` from appearing anywhere in this line's code, enforced by a repo-wide grep in slice E's acceptance.

## The fork the audit's framing hides

"Just add a load-more button to all five" is wrong. The four data shapes are not unifiable:

- **kanban** → per-group cursor (`view-groups.rowsCursor`)
- **gantt / timeline** → a full-filtered-set `MIN`/`MAX` reduction — a paging cursor can *never* produce it
- **calendar** → a range-window predicate, orthogonal to `created_at` offset paging
- **hierarchy** → an ancestor closure; a child's correct position is determined by an ancestor that may sort arbitrarily late

The lock names this fork and hands each resolution to the slice behind its own gate, rather than papering it over — mirroring how #3928 handled `mode: indexed | materialized | unavailable`.

## Mutex declaration — three classes

**1. Grid mutex** (`MetaGridTable.vue` 1516 L, `useMultitableGrid.ts` 1418 L) — owned by #3928 slices C/D. **This lock's slices never touch either file.** New fetches land in a new `useNonGridMaterialization.ts`; continuation goes only through `useMultitableGrid`'s public `loadMore()` export (`:693`). Slice C's acceptance includes a `git diff --stat` assertion of zero change to both. Deliberate seam choice so the two ladders can run in parallel.

**2. Workbench mutex** — `[MUTEX:WB]` — `MultitableWorkbench.vue` (4419 L, single-occupancy shared assembly). Slices **B / C / E / F / G** are tagged and **cannot run concurrently with #3928 slices C / D / E**, nor with each other.

**3. Backend mutex** — `[MUTEX:BE]` — *this class was previously unnamed by either lock; §5.0-C adds it.*

| File | #3928 | this lock |
|---|---|---|
| `packages/core-backend/src/routes/univer-meta.ts` (hottest route file; the `/view` body alone spans `:12983-13492`) | slice **A** (`view-groups` route skeleton), **B** (three-branch impl) | slice **D** (`view-extents` route), **G** (calendar range predicate) |
| `apps/web/src/multitable/api/client.ts` | slice **A** (wire types) | slice **D** (wire types) |
| `packages/core-backend/src/multitable/aggregation-helpers.ts` | slice **B** (bucketing) | slice **D** (date `min`/`max`; must widen `fnApplies` `:53` / `isNumericFieldType` `:17`) |

These are append-style edits (new route / new exported type / new branch), so **no exclusive gate** — but **baseline-first is mandatory**: no two in-flight route PRs on `univer-meta.ts`; whichever lands second rebases onto the first and **re-derives every `:12xxx`/`:13xxx` citation** (§6 rule 1 — line drift in that file is guaranteed). And because `univer-meta.ts:12785`'s leak gate *depends on* `isNumericFieldType`'s current semantics, slice D may widen it for dates only — never to "any comparable type."

Slices **A** (hierarchy stop-the-bleed, zero backend, doesn't touch the workbench) and **D** (backend only) carry no `[MUTEX:WB]` and may run in parallel with the frontend lanes.

## Permission parity — reuse by citation, invent nothing

Central `rbac`/`auth` is locked (K3). The lock cites the existing chokepoints and requires each new read path to inherit them layer by layer:

- Field mask: `maskStoredRecordFieldIds` (`univer-meta.ts:13044`) → `filterRecordDataByFieldIds` (`:13345`)
- Row-read-deny: `:13354-13373` (`hasRecordPermissionAssignments` `:13355` → `loadRowLevelReadDenyEnabled` `:13366` → `rows.filter(deriveRecordPermissions(...).canRead)` `:13367-13370`)

Three egress surfaces are named explicitly, each with a required golden (§4 INV-2):
1. **Calendar's range predicate** is a new server-side predicate — it must be evaluated *after* the field mask, or a denied/tainted date column becomes bit-by-bit probeable by varying the range and observing hit counts.
2. **Gantt/timeline extents are a column's distribution endpoints.** They must replicate the existing numeric-only leak gate at `univer-meta.ts:12785` (`if (!fieldType || !isNumericFieldType(fieldType)) continue`). Note that `isNumericFieldType` (`aggregation-helpers.ts:17`) and `fnApplies` (`:53`) restrict `min`/`max` to numeric types today, so **date-column extents do not exist in the repo** — the extent slice is a new capability and must be computed *after* row-deny.
3. **Hierarchy ancestor closure** turns "an ancestor I cannot read exists" into an observable.

**INV-3 records an inherited constraint and forbids fixing it here.** `/view`'s `page.total` is computed *before* row-deny (`COUNT(*) OVER()` at `:13272`, or `sorted.length` at `:13265`; the deny filter runs at `:13354-13373`), and is already displayed by `MetaToolbar.vue:185`. (Row-level read-deny is per-sheet opt-in, default-off, so the two agree on sheets that never enabled it.) That is central-permission-surface territory. The lock therefore (a) forbids any slice from "fixing it along the way," and (b) forbids any slice from rendering an `N of M` where `M` is not a post-deny count — which is why the disclosure slice (§3.2) reduces to a *boolean* "am I truncated" plus an action, and renders no `M` at all. A grep assertion in slice B's acceptance enforces this.

## Gated ladder

Seven rungs, all `🔒`. Each carries `unblockedBy`, a model tier (**sonnet impl / opus review** throughout), a named verification requirement, and **the exact mutation that must go RED**. Representative examples:

- **A** (hierarchy stop-the-bleed, zero backend, doesn't touch the workbench) — mutation: restore `:370`'s unconditional `null` + `:371`'s count ⇒ both the "row still rendered under `orphanMode:'hidden'`" and "no fabricated `orphanCount`" assertions must go red **together**.
- **B** (truncation disclosure) — mutation: render `${rows.length} / ${total}` ⇒ the INV-3 grep assertion ("no `1234` in the rendered text") must go red.
- **D** (backend extent contract, real-DB golden, new DB single-run) — mutation: comment out the leak gate at `:12785` ⇒ "denied date column yields no key" goes red; move the extent computation *before* row-deny ⇒ the two-user golden (A denied the row holding `max` must see the second-largest) goes red.
- **F** (gantt/timeline extents) — mutation: revert `timeRange` to the local `props.rows` min/max ⇒ the INV-6 assertion (axis ticks and bar geometry **byte-identical** across `pageSize` = 10 / 50 / 200) must go red.

A guard whose mutation cannot be named is not a guard; §6 rule 2 makes this a pass/fail criterion, not a bonus.

## Self-verification performed before commit

- **102 `file:line` citations checked mechanically** against the worktree immediately before committing, and re-swept after the final amend (102/102 OK). **1 self-correction:** the `/view` handler body ends at `:13492` (`:13493` is blank, next `router.*` at `:13494`), not `:13493` as first written.
- The "same-screen contradiction" claim was checked at the render level, not just by string match: `<MetaToolbar>` is a direct child of the root `div.mt-workbench` (`:2`), a sibling of the `div.mt-workbench__content` (`:146`) that holds the view `v-if` chain, with no conditional wrapper; `MetaToolbar.vue`'s root (`:2`) has no `v-if`; and the component references no view type (`grep -c "viewType|activeViewType"` → 0).
- **Zero external-product references.** A case-insensitive grep for the 18-name competitor/vendor blocklist (the standard list used for these locks) over the document → **0 matches**. The document states MetaSheet's own principles only, and cites no outside product.
- Line 3 byte-compared against `状态：PROPOSED — 待 owner ratify`.
- All five mount specs named in the ladder's acceptance criteria (`apps/web/tests/multitable-{kanban,calendar,gantt,timeline,hierarchy}-view.spec.ts`) confirmed to exist.

---

**Do not merge as ratification.** This is a `PROPOSED` design-lock awaiting owner review. Ratifying the document does not unlock slice A; landing slice A does not unlock slice B. Every rung is an independent owner opt-in.


zensgit:
**✅ Independently adversarial-gated: MERGE_CLEAN.** All decisive claims verified in code: 5 views bind the same 50-row page with no pagination, toolbar shows the pre-deny total for every view, **hierarchy drops an in-memory row** under `orphanMode:hidden` (re-derived end-to-end), gantt/timeline coordinate system is a function of the truncation, gallery correctly excluded (real navigable pager), INV-3 pre-deny ordering correct and the lock touches no permission code. No P1/P2. Applied the gate’s P3 in the latest push: the **deny-ON residual FOURTH state** — on a row-read-deny sheet `page.total` is pre-deny (INV-3, untouchable), so `rows.length<total` stays true forever after loading all readable rows while `canLoadMore=false && capped=false`; the 3-state disclosure model didn’t enumerate it and would show a "narrow your filter" CTA that can’t help. Added the INV-4 carve-out (neutral terminal "all visible records loaded", no M) + wired it into §5.1-B’s criteria + mount assertion. Report `/tmp/nongrid-materialization-lock-gate-20260708.md`. Still PROPOSED — do not merge as ratification.

PR #3853, opened 2026-07-08
ci(web-guard): gate the UI-P2-1c migration proof-specs (task #30 systemic CI-coverage)

zensgit:
**Owner review requested — CI-config, not self-merged** (outside the presentation-only self-merge bar).

Closes the systemic gap the #3852 adversarial gate escalated to P2: the 18 `*-migration.spec.ts` UI-P2-1c proof-specs (12 MtButton + 6 MtPopover/MetaToolbar) run in **no** CI workflow — the web-guard's curated `vitest run` list only covered `meta-toolbar-filter-builder`, and there's no full-suite web job. A refactor breaking the migrated `@click`/emit wiring would ship green (the #3769 failure mode).

**Fix (additive test-coverage only, zero product/runtime change):**
- Run-list: one positional token `migration` → runs ALL `*-migration.spec.ts`. Verified it matches **only** the 18 UI migration specs (no slow/DB test); **24 files / 108 tests green in ~2s** locally with the exact web-guard command. Future migration specs auto-covered.
- Paths: `apps/web/tests/*-migration.spec.ts` added to both `pull_request` + `push` triggers.

Retroactively gates all 12 landed MtButton migrations (#3823/26/28/32/33/34/35/36/42/44/50/52) + the 6 toolbar slices.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
**2026-07-09 gate-P2 fix**: `migration` token now ALSO added to `apps/web/scripts/run-required-web-tests.sh` (required `web-tests` gate) per the two-point discipline installed by #3967 — the advisory guard alone left 23/24 migration specs outside the merge-blocking gate. Spec-count figures refreshed: the token currently matches 24 files / 108 tests (auto-covers new *-migration.spec.ts by design).

PR #3934, opened 2026-07-08
feat(web): multitable formula catalog parity — surface 12 engine-supported functions, replace an inert parity fixture with a real drift guard

zensgit:
**Do not merge as a presentation-only self-merge.** This is a **product-surface** change (users see 12 new functions in the formula editor's catalog/autocomplete), and it now also touches CI config and `packages/core-backend`. Open for owner review.

## What this does

Surfaces 12 functions the engine has **always executed** but the editor never advertised: `DAYS, EDATE, IFERROR, IFS, ISBLANK, ISERROR, ISNUMBER, ROUNDDOWN, ROUNDUP, SECOND, WEEKDAY, XOR`.

Not just discoverability — **a real defect fix**: `formula-docs.ts` warns `formulaUndocumentedFunction(fn)` for any called function absent from the catalog. Before this PR, typing `=XOR(a,b)` — which the engine happily evaluates — produced a spurious *"XOR is not yet documented in this editor"* warning. Gone for all 12.

All 12 verified against the live registry (`new FormulaEngine().getRegisteredFunctionNames()`), not against the audit that proposed them. `COUNTIF` is **deliberately excluded**: it is engine-registered but takes a *range*, which the `{fld}`-reference model cannot express — advertising it would be a false-capability bug. The parity assertion is therefore a **subset**, never an equality, so it can never demand COUNTIF.

## The interesting part: the first version of this PR shipped a guard that guarded nothing

An adversarial gate (report: `/tmp/formula-catalog-parity-gate-20260708.md`) proved the original parity spec was an **inert fixture**, two ways:

- **Structural** — the spec had *no import edge to the engine at all*. It compared the catalog against a hand-copied `ENGINE_REGISTRY` snapshot the same PR wrote. No change to `engine.ts` could ever turn it red. (This is provable by reading the import list; you don't have to run it.)
- **Mutation A/B** — deleting `this.functions.set('XOR', …)` from `engine.ts:243` left the guard **GREEN (3 passed)** while users typing `=XOR(a,b)` would get `#NAME?`.

The mirror was byte-exact the day it was written (78 = 78). That precision is exactly what made it dangerous: correct now, silent when it stops being.

The stated blocker for a real guard — *"`apps/web` can't import `@metasheet/core-backend`; the engine pulls in a live Kysely/pg pool at import time"* — was true-but-harmless: `new Pool()` constructs an object, `pg` opens no socket until first query. Ten backend test files already import `FormulaEngine`. And the cited precedent argued the **opposite** way: it asserts `expected ⊆ getRegisteredFunctionNames()` (live registry, catches removal), while the new spec hardcoded the registry itself. Structurally inverted.

**Fix:** the guard moved to the backend, where the engine *can* be imported live.

- `packages/core-backend/tests/unit/formula-catalog-web-parity.test.ts` — live `getRegisteredFunctionNames()` + `fs`-read of the web catalog; asserts `catalog ⊆ (live ∪ REL* ∪ operators)`.
- The web spec's `ENGINE_REGISTRY` mirror is **deleted**; it keeps only intent assertions (the 12 are present, COUNTIF is absent). Its header comment — which stated the refuted justification — is rewritten.

### Anti-fixture guard on the guard

A guard that `fs`-reads and regex-scrapes has the *same* failure mode: if the path breaks or the formatting changes, the regex matches 0 names, the "uncallable" set is `[]`, and it goes **silently green**. So the new test asserts its own harvest is plausible — non-empty source, `names.length >= 80`, known anchors (`SUM`, `IFERROR`) present — and that is mutation-proven below (M-D).

## Mutation record (every guard is load-bearing; all reverted, tree clean after each)

| Mutation | Target | Result |
|---|---|---|
| **M-A** delete `functions.set('XOR')` from `engine.ts:243` | backend guard | **RED** — `catalog advertises uncallable: XOR` |
| **M-B** rename catalog `'XOR'` → `'XORX'` | backend guard | **RED** (web spec independently red too) |
| **M-C** inject a `COUNTIF` catalog entry | web spec | **RED** — absence assertion fires |
| **M-D** break the scrape (wrong path / broken regex) | anti-fixture sanity asserts | **RED both ways** — never silently green |
| **M-E** remove the `DAYS` arity entry | arity test | **RED** |

## Also fixed: the 12 were exempt from arg-count validation

`formula-docs.ts` does `const arity = FORMULA_FUNCTION_ARITY[call.name]; if (!arity) continue` — a **silent skip**. That table (46 keys) contained none of the 12, so `=IFERROR({fld_a})` produced no diagnostic and a silently wrong value. All 12 added, each checked against the engine's actual argument handling. Note this is a **third** name-keyed map, distinct from `FORMULA_FUNCTION_DOCS` and `FUNCTION_DESCRIPTIONS`; adding a catalog function requires updating all three.

Text corrections while here: `ROUNDUP`/`ROUNDDOWN` bracket their optional `[digits]`; `IFS` documents its `#N/A` fall-through; `WEEKDAY` states its default numbering (type 1 → Sun=1 … Sat=7). en + zh both.

## CI wiring (this is why the PR is not self-mergeable)

- **The backend guard runs in a required check.** `.github/workflows/plugin-tests.yml` → `test` job → *"Run core-backend tests"*. Verified fail-closed: no `if:`, no `continue-on-error`, no `|| true`, no path filter. `test (20.x)` is one of the 5 required contexts.
- **`multitable-web-guard.yml`** — the web specs ran in **no** workflow at all. Added the two source files + 4 spec paths to both `pull_request.paths` and `push.paths`, and appended 4 positional tokens to the **single** `vitest run` line (verified by YAML parse: exactly 1 such step; a second `run:` line would silently drop specs).

⚠️ **Land-order dependency with #3853.** That PR edits this same file's `paths:` + run-list to add a `migration` token/glob. The edits are additive at different anchors, but whoever lands second must **union the run-list into one line by hand**.

## Honest disclosure — one guard this PR *newly runs in CI* overclaims

`apps/web/tests/multitable-formula-editor.spec.ts` contains a test named *"documents every backend registered formula function"* which hardcodes **its own** `backendFunctions` list — the same structural pattern as the mirror deleted above, in the opposite direction (`backendFunctions ⊆ catalog`). Because this PR wires that spec into CI, it now *executes*, but its name promises coverage it cannot deliver: if the engine **adds** a function, the hardcoded list doesn't know, the spec stays green, and the catalog silently fails to document it. Left untouched (out of this PR's scope) and flagged for a follow-up. The new backend guard covers the other, more dangerous direction (catalog → uncallable).

Also found: `meta-formula-labels.ts`'s `FUNCTION_DESCRIPTIONS` overrides `doc.description` for **both** locales whenever a matching entry exists — so editing a catalog description in `formula-docs.ts` alone is a silent no-op. Both files updated together here; future catalog-text edits should watch for it.

## Verification

- `type-check` clean on both `@metasheet/web` and `@metasheet/core-backend`.
- Target web specs 22/22; full guard run-list 53 files / 537 tests green; new backend guard + sibling formula backend tests 20/20.
- **Pre-existing-failure baseline, measured on BOTH trees:** 20 failed files / 111 failed tests, failing file sets **identical by name** → zero new failures from this PR. (Dominant signatures `SQLSERVER_EXECUTOR_MISSING` and `connect failed:` — unrelated.)

Zero engine change. Zero runtime change.


zensgit:
⚠️ **Same-line collision with #3853** (both edit the single `vitest run` line in `multitable-web-guard.yml`; #3853 adds a `migration` token, this PR adds 4 formula tokens). **Whoever lands SECOND must hand-union** so the final one line carries BOTH `migration` AND `multitable-formula-catalog-parity multitable-formula-catalog-1a meta-formula-labels multitable-formula-editor`. #3853 has been rebased onto current `main`; this PR was rebased onto `096490866`. Stacking a second `run:` line silently drops specs — keep it one line.

zensgit:
## 欠账重闸裁决 — MERGE_CLEAN ✅(head 7f6a8a9e3)

初闸抓出惰性 fixture 后的返工 + post-#3959 union 一并独立重审(报告 `/tmp/pr3934-regate-20260710.md`,4 刀 mutation 全真跑全还原):

| 面 | 结论 |
|---|---|
| 漂移守卫真咬 | 真 import 边到活注册表(engine.ts:488);Knife-A 删引擎 XOR→RED;Knife-B 注入假目录项→RED;M-D 破正则→反 fixture 双 RED;守卫跑在 **required** test(20.x) fail-closed |
| COUNTIF 分层 | M-C 注入 COUNTIF→web absence 断言 RED 而 backend 守卫仍 GREEN(可调用)——职责分层正确 |
| union 严格 additive | DOCS 91→103 / ARITY 59→71(+12/0 丢/0 重),labels en+zh 12×2 全齐,4 项签名语义抽检 doc↔引擎一致 |
| yml 严格 UNION | run 行 43→47 token 零丢失,恰 1 个 vitest step,12 path 条目,YAML 有效 |
| F0(历史已修) | union 曾丢 CHOOSE 闭合 `},`(CI 三红抓出),7f6a8a9e3 修复,当前 head 解析 OK |
| 本地+CI | web 22/22、backend 守卫 3/3、tsc 干净;head 全 required 绿 |

残余:1×P3(editor spec 反向惰性,PR body 已如实披露为 follow-up,仅影响可发现性)+ 1 NOTE(「12 present/COUNTIF absent」intent 断言在 advisory 门,required 面覆盖 catalog⊆callable——本次已本地实证 22/22)。

**P1/P2 清零,落地。**

PR #3941, opened 2026-07-08
docs(research): provider-agnostic org-transfer model + DingTalk corp-switch assessment (Rev 4 — as-built)

zensgit:
**Rev 4 (2026-07-13) — as-built refresh.** File content is Rev 4 (head `1fd3e4278`); this metadata now matches so review targets the right version.

Rev 4 corrections vs the pinned baseline (~268 commits behind current main):
- **§3** corp_id-editable kept TRUE for current main + forward-note that PR #4181 (`claude/dingtalk-org-transfer-phase1-guardrails`, gate APPROVE 0 P1/P2, CI-green, unmerged) makes corp_id **unconditionally immutable once set**; correction path = delete+recreate before first sync; no production escape hatch.
- **§10 DT-OPS-01** stale "not yet merged" **corrected** → merged via #3905 (`8bc8d5aad`), on main, env-gated default-off; clarified it guards ONLY the downstream local-user deprovision, never the `directory_accounts`/`directory_departments` is_active sweep.
- **§4** unguarded-sweep claim kept + sharpened (do not conflate with DT-OPS-01).
- **§12/§13** org_id-anchor pointer (true cross-corp/cross-provider anchor; #3944 Rev 3 formalizes it).
- Anchor-drift warning added (every file:line must be re-verified; concrete `:1596→:2173` example).

§5/§6 and §9.3 single-transaction-rebind untouched. Docs-only. refs #3944 (Rev 3 companion).


zensgit:
## Rev 3 — migration composition corrected (your finding)

You're right: §5/§6/§13 still prescribed a naive **unbind-A-then-bind-B** using the public `unbindDirectoryAccount()` + `bindDirectoryAccount()`, which contradicts the P1 we fixed. The public `unbindDirectoryAccount` **deletes** the identity (`:4056`), so composing it with a later bind opens a no-identity window, and bind-then-unbind throws on the still-linked guard.

Corrected §5, §6 (step 3), and §13 to the **dedicated single-transaction rebind** — clear the source link *without deleting the identity*, upsert the one `user_external_identities` row in place source→target, enable the grant, then link the target — and marked **`provider-org-transfer-development-plan-20260709.md` §9.3 (#3944 Rev 2) as the authoritative spec that supersedes this assessment's migration prescription**. The description of what the primitives *themselves* do is unchanged (that part was correct); only the "how to compose them" prescription was wrong. Rebased onto current main.

PR #3945, opened 2026-07-08
docs(multitable): ordering-substrate design-lock (manual row order) — PROPOSED, do not merge as ratification

zensgit:
Do not merge as ratification. PROPOSED design-lock, docs-only (1 file, zero runtime), for cluster B of the benchmark goal pool: manual/explicit ordering.

**⚠ NOT independently adversarially-gated.** The drafting subagent hit a *weekly* API limit (resets Jul 12); the adversarial-gate pipeline is paused, so this lock was written verify-first in the main loop and self-reviewed only. Scrutinize its load-bearing claims (§1 current-state, §4 primitive fork) harder than the gate-passed #3928/#3931/#3940.

## Verify-first narrowed the premise (two corrections vs the brief)
- **Manual row order = the real gap.** `meta_records` reads `ORDER BY created_at ASC, id ASC` at multiple sites; no `order_index`/`position` column exists. Users cannot drag a row into a persistent position.
- **Kanban = narrower than briefed.** Cross-column drag ALREADY works — `onDrop` (`MetaKanbanView.vue:407`) emits `patch-cell` on the group field. Only intra-column *order* is missing → a gated rider (R5), rendering deferred to #3931.
- **Dashboard = excluded.** Panels already carry a linear integer `order` (`MetaDashboardView.vue:746`); only 2D free-layout would be additive and it is not shown benchmark-necessary.

## The load-bearing fork (§4), not papered over
Existing field-reorder uses integer positions with **shift-on-insert** (T9-R1) — O(n) rewrites per move, fine for <100 fields, **wrong for thousand-row collaborative tables** (its O(n) writes collide with #3931 infinite-scroll and #3928 group cursors). The lock **constrains the row primitive to O(1)-insert + collab-safe** (→ fractional/rational index) and makes the rebalance sub-design a completeness requirement of rung R1, with the exact RED mutation that proves O(1) is load-bearing.

## Guardrails
Gated ladder R0–R6, each with unblockedBy + model tier + named verification + the exact mutation that must go RED; every runtime rung is open-a-PR (never self-merge). Permission parity by citation (`ensureRecordWriteAllowed` etc.), central rbac/auth untouched, and a required golden tying row-order-key enumeration to the same post-deny discipline as the G5 finding. Full mutex ledger incl. `[MUTEX:BE]`. Seam to #3928 (grouping×order) and #3931 (kanban card order) declared, not re-litigated. Own-principles only — grep-verified zero external-product names.

Do not merge as ratification.


zensgit:
**✅ Now independently adversarial-gated: MERGE_CLEAN** (subagent capacity recovered; the earlier "weekly limit" block was stale). All four decisive claims verified in code at the pinned baseline: manual row order has no substrate (`meta_records` has no order column), field-reorder is integer shift-on-insert (O(n)/move — wrong for rows, confirming the §4 fork), kanban cross-column drag already works via `patch-cell` (only intra-column order is the gap), dashboard already has linear panel `order` (exclusion correct). Permission citations real, no invented gates, every rung mutation load-bearing, seam vs #3928/#3931 clean.

Applied the gate’s two P3 doc-accuracy notes in the latest push:
- **P3-1** — preempt the dormant `table_rows.row_order`: it is a *different* legacy table (`DefaultViewDataProvider`), has **no writer** (read/projection/type only), and is not a usable substrate. §1.1 now says so.
- **P3-2** — honest framing: the §4 O(1)+collab-safe hard constraint already **forces the primitive class to rational-index**, so ratifying commits to that class; only rebalance/keygen stays open. §4/§9 reworded.

Report: `/tmp/ordering-substrate-lock-gate-20260708.md`. Still PROPOSED — do not merge as ratification.

PR #3959, opened 2026-07-08
feat(multitable): formula engine — 13 standard scalar functions (cluster-C path-1 batch 1)

zensgit:
## Summary
Adds 13 standard scalar functions the formula engine was missing (cluster-C path-1, batch 1 of the formula benchmark goal pool):

`PRODUCT · SIGN · PI · RADIANS · DEGREES · GCD · LCM · VALUE · PROPER · TEXTJOIN · STDEVP · VARP · CHOOSE`

All 13 are pure-scalar or variadic-scalar — none take a range/array — so all are `{fld}`-reference expressible under the current engine model. This is exactly why COUNTIF/SUMIFS-style criteria functions are deliberately excluded from this batch (they need range semantics the `{fld}` model can't express).

Each mirrors an existing engine pattern rather than inventing a new convention:
- **PRODUCT** — mirrors `SUM`'s flatten + non-numeric→0 coercion (multiplicative identity seed = 1; empty-args special-cased to 0 to match `SUM()`'s own empty-args convention).
- **SIGN / RADIANS / DEGREES** — mirror `INT`/`LN`'s NaN-guard → `#VALUE!` sentinel shape.
- **PI** — mirrors `TRUE`/`FALSE`'s zero-arg convention.
- **GCD / LCM** — Excel-standard: truncate non-integers, reject negatives (`#NUM!`), non-numeric → `#VALUE!`. `GCD(0,0)=0`; `LCM` short-circuits to `0` the moment a `0` argument appears.
- **VALUE** — parses a number from text; unparseable/empty → `#VALUE!`.
- **PROPER** — capitalizes each letter-run (non-letters reset the word boundary), mirroring `UPPER`/`LOWER`'s `String(text)` coercion.
- **TEXTJOIN** — joins with a delimiter; `ignore_empty` drops `''` entries post-coercion.
- **STDEVP / VARP** — population variants of the existing sample `STDEV`/`VAR` (divide by N, not N-1 — verified the existing ones are sample/N-1 first). A single value is a *valid* population (variance 0), unlike the sample variant which is undefined for N<2 — only the true 0-arg case is degenerate, and it mirrors `STDEV()`'s own 0-arg behavior (propagates as a thrown error, caught by `calculate()` as `'#ERROR!'`) rather than inventing a new sentinel.
- **CHOOSE** — 1-based; truncates a fractional index toward zero (Excel convention); out-of-range/non-numeric index → `#VALUE!`.

Catalog + i18n exposure so the editor advertises these: `formula-docs.ts` (`FORMULA_FUNCTION_DOCS` + `FORMULA_FUNCTION_ARITY`, alphabetically inserted per existing convention) and `meta-formula-labels.ts` (`FUNCTION_DESCRIPTIONS`, en+zh for all 13, the sole i18n extension point).

This is behavior/runtime, not presentation-only — **owner merges, not self-merge/auto-merge.**

## Collision note (#3934, OPEN)
`claude/multitable-formula-catalog-parity` (#3934) also touches `formula-docs.ts` + `meta-formula-labels.ts` + `FORMULA_FUNCTION_ARITY`, adding 12 *different* engine-supported functions (IFERROR/ISERROR/ISBLANK/ISNUMBER/IFS/XOR/ROUNDUP/ROUNDDOWN/WEEKDAY/SECOND/EDATE/DAYS). No function-name overlap — purely additive. Two structural notes for whoever lands second:
- `FORMULA_FUNCTION_DOCS` (array) and `FUNCTION_DESCRIPTIONS` (record): both PRs append a new block at the very end, matching every prior capability-expansion batch's convention — trivial union, same pattern already used repeatedly on this file.
- `FORMULA_FUNCTION_ARITY` (alphabetically-sorted map): this one is genuinely interleaved, not appended, so 3 of my 13 keys land on the *same anchor lines* as 3 of #3934's 12 keys — `DEGREES` (mine) next to `DAYS`/`EDATE` (theirs), `LCM` (mine) next to `ISBLANK`/`ISERROR`/`ISNUMBER` (theirs), and `SIGN` (mine) next to `ROUNDDOWN`/`ROUNDUP`/`SECOND` (theirs). Whichever PR lands second will likely see a textual merge conflict at those 3 spots — trivial to resolve (keep both key blocks; the key sets are disjoint, no semantic conflict) but flagging so it isn't a surprise.
- Did not add a drift-guard test (#3934 adds `formula-catalog-web-parity.test.ts`; a second one would collide — relying on #3934's post-merge, per instruction).

## Test plan
- [x] `pnpm --filter @metasheet/core-backend run type-check` — exit 0
- [x] `pnpm --filter @metasheet/web run type-check` — exit 0
- [x] New `packages/core-backend/tests/unit/formula-functions-batch1.test.ts` — 12/12 green (correct-value + edge-case + error-sentinel per function)
- [x] Existing backend formula suites green: `formula-engine.test.ts`, `formula-functions-expansion.test.ts`, `formula-1a-text-date-expansion.test.ts`, `multitable-formula-engine.test.ts`, `formula-value-substitution.test.ts`, `formula-dryrun.test.ts`, `multitable-formula-over-lookup.test.ts`, `multitable-formula-reference-guard.test.ts` — 201/201 green
- [x] Existing frontend formula suites green: `multitable-formula-editor.spec.ts`, `meta-formula-labels.spec.ts`, `multitable-formula-catalog-1a.spec.ts`, `multitable-formula-suggest-field-manager.spec.ts`, `multitable-formula-dryrun-panel.spec.ts` — 39/39 green
- [x] Mutation self-test: neutered the `SIGN` NaN-guard, `GCD`/`LCM`'s negative-→`#NUM!` guard, and `CHOOSE`'s out-of-range guard — all 4 corresponding assertions flipped red, everything else stayed green — then restored to the committed implementation (`git checkout --` against the pre-mutation commit, verified `git diff HEAD` empty)
- [x] `git diff --stat origin/main`: `engine.ts` + 1 new backend test + `formula-docs.ts` + `meta-formula-labels.ts` only — no workflow file, no permission/rbac file, no other runtime touched


zensgit:
**✅ Independently adversarial-gated: MERGE_CLEAN — safe to merge.** Every one of the 13 functions was semantically recomputed against the implementation (not the author’s asserted values). All correct. Highlights the gate verified:
- **STDEVP/VARP use the population divisor N** (distinct from the existing sample STDEV/VAR at N-1) — mutation-proven: flipping to N-1 turns STDEVP(1..5) into the sample value and the test goes red.
- **All three name-keyed maps** (`FORMULA_FUNCTION_DOCS`, `FUNCTION_DESCRIPTIONS` en+zh, `FORMULA_FUNCTION_ARITY`) complete with correct arities (PI 0/0, CHOOSE min2, TEXTJOIN min3, etc.) — confirmed by the catalog specs, not just a read.
- **No invented error sentinel** (only `#VALUE!`/`#NUM!`/`#ERROR!` from the authoritative set); scope is exactly the 4 declared files, zero permission/rbac/workflow drift.
- Full backend `formula` suite 277/277, web type-check clean.

**Three disclosed behaviors (owner-decidable, NOT changed — each is faithful to the engine’s own conventions and never returns a wrong number):**
1. `PRODUCT` zero-coerces a non-numeric/blank arg (mirrors the engine’s SUM/AVERAGE `parseFloat||0` convention).
2. `VALUE` uses JS `Number()` — accepts `0x1F`→31, rejects `50%`/`$1,000` with an honest `#VALUE!` (never a wrong number).
3. `PROPER` is ASCII-only (`[A-Za-z]`) — non-ASCII leading letters aren’t title-cased.

If you want any of these three changed (e.g. Unicode-aware PROPER, VALUE parsing percent/currency), that’s a product decision — say so and I’ll do a follow-up. Report `/tmp/formula-functions-batch1-gate-20260708.md`. Behavior PR → your merge (not self-merge).

PR #3965, opened 2026-07-09
fix(collab): fail closed Yjs sync for field-masked actors

zensgit:
## Summary
- Add a Yjs read-access check that disables record-wide realtime sync when the actor cannot read every field on the sheet.
- Wire the check into the production Yjs auth gate before a socket joins the record room or receives sync data.
- Add regression tests for the field-read guard and the socket-level fail-closed behavior.

## Verification
- pnpm --filter @metasheet/core-backend exec vitest run tests/unit/yjs-field-read-access.test.ts tests/unit/yjs-awareness.test.ts
- pnpm --filter @metasheet/core-backend exec vitest run tests/unit/yjs-field-read-access.test.ts tests/unit/yjs-awareness.test.ts tests/unit/yjs-sync-seed.test.ts tests/unit/yjs-hardening.test.ts tests/unit/yjs-rest-invalidation.test.ts
- git diff --check

## Notes
- Scoped to the Yjs realtime read surface. No REST route, write path, comment, attachment, or permission schema changes.
- The realtime path fails closed and lets clients fall back to REST reads, where response-scoped field masking already applies.

### DIFF
--- a/packages/core-backend/src/collab/yjs-field-read-access.ts
+++ b/packages/core-backend/src/collab/yjs-field-read-access.ts
@@ -1,0 +1,57 @@
+import {
+  deriveFieldPermissions,
+  type MultitableCapabilities as FieldPermissionCapabilities,
+} from '../multitable/permission-derivation'
+import {
+  loadFieldPermissionScopeMap,
+  type QueryFn,
+} from '../multitable/permission-service'
+
+type YjsFieldRow = {
+  id: string
+  type: string
+  property: unknown
+}
+
+function normalizeFieldRow(row: YjsFieldRow): {
+  id: string
+  type: string
+  property: Record<string, unknown>
+} {
+  const property = row.property && typeof row.property === 'object' && !Array.isArray(row.property)
+    ? row.property as Record<string, unknown>
+    : {}
+  return {
+    id: String(row.id),
+    type: String(row.type),
+    property,
+  }
+}
+
+/**
+ * Yjs sync sends a whole record-level Y.Doc, not a field-scoped payload. If the
+ * actor cannot read even one field on the sheet, serving that shared document
+ * would let denied field values travel over the realtime channel. In that case
+ * callers must fail closed and let the client fall back to REST reads, where
+ * field-level masking is already applied per response.
+ */
+export async function canReadEveryYjsFieldForUser(
+  query: QueryFn,
+  sheetId: string,
+  userId: string,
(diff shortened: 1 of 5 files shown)


zensgit:
**✅ Post-merge independent adversarial gate: MERGE_CLEAN — the fail-closed fix is sound, no residual leak on main.** (Reviewed at merged head; auto-merge landed before the gate, no race remained.) This closes the F0 Yjs field-mask bypass (realtime read-sync leaking field-denied values via `writeSyncStep1`).

The fix is a **coarse fail-closed**: if the actor cannot read EVERY field on the sheet, `/yjs` subscribe is refused (`FORBIDDEN`, client falls back to REST) — never subscribing a masked actor, which sidesteps the "must mask both initial snapshot AND every incremental update" trap. Verified:
- **Both initial-sync AND incremental egress blocked** — the `canReadAllFields===false` early-return (adapter L207-212) fires before `getOrCreateDoc`/`writeSyncStep1`/`socket.join`/presence; no room join ⇒ no incremental broadcast; `socketPermissions` unset ⇒ client `yjs:message` rejected "Not subscribed"; presence never joins ⇒ no awareness leak.
- **Fail-closed default** — DB error/missing user/`permissionHidden` all → block.
- **Per-actor** — recomputed each subscribe via `loadFieldPermissionScopeMap(…, userId)`, never cached across actors.
- **Full-access actor unbroken** — `canReadAllFields=true` ⇒ full sync; presence tests green.
- **Tests are load-bearing** — mutating the mask fn (`return true`) → 3 unit fail-closed tests RED; mutating the adapter gate (`if(false)`) → the awareness "rejects record-wide sync" test RED (it asserts the payload is ABSENT from the wire, not just that the fn ran). 9/9 pass unmutated.
- **Parity** — gate block-set ⊇ REST mask-set (both use the same `deriveFieldPermissions`); reuses the canonical resolver, does not touch central rbac/auth.

**P3 (non-blocking, optional follow-up):** (1) adapter L210 `access.canReadAllFields === false` fails-OPEN on `undefined` — safe today (the sole production `setAuthChecker` at index.ts:2359 always sets a boolean), but `!== true` would be defense-in-depth against a future authChecker that omits the field; (2) the fail-closed is whole-channel per sheet (one masked field disables realtime for that actor on every record, with REST fallback) — a correct safe-side UX tradeoff. Report `/tmp/yjs-field-mask-3965-gate-20260708.md`.

PR #3972, opened 2026-07-09
feat(directory): batch admit synced DingTalk users

zensgit:
## Summary

- Add a batch directory admission backend path for synced DingTalk accounts that need local users created in bulk.
- Default the batch path to DingTalk login grant OFF, so missing web-OAuth openId does not block creating local accounts.
- Add a directory review-queue action for manual pending items: batch create local users, bind them, and show temporary credentials for secure distribution.

## Behavior

- `POST /api/admin/directory/accounts/batch-admit-users` accepts `accountIds` and returns committed accounts, users, onboarding packets, and per-item failures.
- Each item is isolated like the existing batch bind/unbind paths: successes are preserved and audited even if a later item fails.
- The UI only enables the batch-create action for selected pending review items that have no local user and no direct recommendation candidate, so recommended matches still use the existing confirmation flow.
- The generated local username reuses the existing directory auto-admission username builder; scan login can be opened later after openId is available.

## Verification

- `pnpm --filter @metasheet/core-backend exec vitest run tests/unit/admin-directory-routes.test.ts tests/unit/directory-sync-bind-account.test.ts --watch=false` (53 passed)
- `pnpm --filter @metasheet/web exec vitest run tests/directoryManagementView.spec.ts --watch=false` (51 passed)
- `pnpm --filter @metasheet/web exec vue-tsc -b`

## Notes

This is the bulk local-account admission cut for directory-synced members. It does not bulk-enable DingTalk scan login; that remains a separate step after web-OAuth openId enrichment.

### DIFF
--- a/apps/web/src/views/DirectoryManagementView.vue
+++ b/apps/web/src/views/DirectoryManagementView.vue
@@ -140,6 +140,60 @@
       </article>
     </article>
 
+    <article v-if="batchAdmissionOnboardingPackets.length > 0" class="directory-admin__progress-card">
+      <div class="directory-admin__section-head">
+        <div>
+          <h2>批量创建临时凭据</h2>
+          <p class="directory-admin__hint">以下目录成员已批量创建本地用户并完成绑定。扫码登录授权默认未开启，请先通过安全渠道下发登录账号与临时密码。</p>
+        </div>
+        <div class="directory-admin__actions">
+          <button class="directory-admin__button directory-admin__button--secondary" type="button" @click="clearBatchAdmissionOnboardingPackets()">
+            关闭结果
+          </button>
+        </div>
+      </div>
+
+      <article
+        v-for="packet in batchAdmissionOnboardingPackets"
+        :key="packet.userId"
+        class="directory-admin__review-admission"
+      >
+        <div class="directory-admin__account-grid">
+          <div>
+            <strong>本地用户</strong>
+            <div>{{ packet.name || packet.email || packet.username || packet.mobile || packet.userId }}</div>
+          </div>
+          <div>
+            <strong>登录账号</strong>
+            <div>{{ packet.email || packet.username || packet.mobile || packet.userId }}</div>
+          </div>
+          <div>
+            <strong>用户 ID</strong>
+            <div>{{ packet.userId }}</div>
+          </div>
+          <div>
+            <strong>手机号</strong>
+            <div>{{ packet.mobile || '未填写' }}</div>
+          </div>
+        </div>
+        <p class="directory-admin__status">
+          临时密码：{{ packet.temporaryPassword }}
+        </p>
+        <p v-if="packet.onboarding?.acceptInviteUrl" class="directory-admin__hint">
+          邀请链接：
+          <a :href="packet.onboarding.acceptInviteUrl" target="_blank" rel="noreferrer">
+            {{ packet.onboarding.acceptInviteUrl }}
+          </a>
+        </p>
+        <p v-else class="directory-admin__hint">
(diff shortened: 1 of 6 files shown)


zensgit:

Head `05bfc5ed6` (matches `gh pr view`). Reviewed against the DingTalk hardening line's trap catalogue, with real-Postgres proofs. Full report: `/tmp/pr3972-review-claude-20260709.md`.

### What holds (proven, not assumed)
- **Orphan-safe by construction**: the batch reuses `admitDirectoryAccountUser` — one `transaction()` per item + the #3896 SAVEPOINT inside `createDirectoryAdmittedUserInTransaction`. A bind throw after the users INSERT rolls back; proven on real PG (Test B). Base already includes the SAVEPOINT; merge-tree vs current main is clean.
- **Grant defaults OFF at both route and service** — a corp account without openId admits fine, no post-INSERT throw.
- Standard `ensurePlatformAdmin` authz; `accountIds` Set-deduped; deterministic username + uniqueness backstops intra-batch collisions; **onboarding packets never persisted, audited, or logged**.
- **Mutation 1 (make the batch fail-fast) → RED** — per-item isolation is genuinely covered (the #3902 hole does not repeat here).

### P2-1 (should fix before/with merge) — server-side eligibility not enforced; case-sensitivity gap → DUPLICATE user (proven on real DB)
The "no local user / no recommendation candidate" gate lives **only in the Vue computed** (`DirectoryManagementView.vue:2638`); the server never consults recommendation logic. Its only defense is `WHERE email = $1` — **case-sensitive** (`directory-sync.ts:3566`) — while sync matching is case-insensitive (`LOWER(email)`, `:1467`) and `AuthService.register` stores raw emails. Real-DB Test A: a mixed-case stored email + a curl submission bypassing the UI gate → **users 1→2**, the account bound to the new empty duplicate. Bounded today (fail-closed in the same-case flow; 0 mixed-case emails in the live DB), but it is the preview/apply-drift class: the eligibility decision must live server-side.
**Fix**: make the create-time uniqueness checks case-insensitive (`lower(email) = lower($1)`, same for the username check if applicable) **and/or** enforce the no-candidate eligibility server-side before admitting.

### P3
1. No real-DB test in the PR — both added tests mock (route→service, service→DB), and the mock isolation test fails its item at LOAD, so the throw-AFTER-INSERT rollback is unproven in the suite (proven manually here; recommend porting the harness as a `.db.test.ts` with the two-point CI wiring).
2. A post-commit throw (`getDirectoryAccountSummary` / `recordInvite`) reports a committed user as `failed` and unaudited — pre-existing, amplified by batch.
3. Unbounded `accountIds` (bcrypt per item amplifies; consider a cap like the deprovision breaker's 25).
4. Admission-scope/exclude-list silently bypassed — matches single-admit semantics, but say so in the PR body (admin-driven bypass as intent, not omission).
5. The web spec (with the new UI eligibility tests) runs in **no CI gate** — web-guards use explicit run-lists that don't include `directoryManagementView`.
6. Audit-packet-leak is unguarded (Mutation 2 GREEN — `objectContaining`): current code is correct, but given the OPS-02 precedent, add a negative assertion that the audit payload NEVER contains `temporaryPassword`.

### Train interaction note
This PR touches `directory-sync.ts` while the merge train is draining (#3915 in window; #3905/#3910/#3884 queued, #3905 heavily in this file). Recommend slotting its merge **after** the train drains to avoid re-DIRTYing the queue — or merge earlier and I'll absorb the re-rebase.

Verification: backend unit 53/53, web spec 51/51, tsc clean; real-DB harness 3/3. Nothing on the branch was modified.

PR #3973, opened 2026-07-09
fix(multitable): enforce read gates on attachment downloads

zensgit:
## Summary
- extend multitable attachment metadata with owning record/field ids for download authorization
- require sheet read, row-read visibility, and field-read visibility before serving attachment bytes
- add focused attachment API coverage proving denied field/row cases do not touch storage

## Verification
- pnpm install --frozen-lockfile
- pnpm --filter @metasheet/core-backend exec vitest run tests/unit/multitable-attachment-service.test.ts
- pnpm --filter @metasheet/core-backend exec vitest --config vitest.integration.config.ts run tests/integration/multitable-attachments.api.test.ts
- git diff --check

## Notes
- scoped to attachment download read gates only
- no attachment upload/delete behavior changes intended

### DIFF
--- a/packages/core-backend/src/multitable/attachment-service.ts
+++ b/packages/core-backend/src/multitable/attachment-service.ts
@@ -95,6 +95,8 @@ export type ReadAttachmentMetadataInput = {
 export type ReadAttachmentMetadataResult = {
   id: string
   sheetId: string
+  recordId: string | null
+  fieldId: string | null
   storageFileId: string
   filename: string | null
   originalName: string | null
@@ -469,7 +471,7 @@ export async function readAttachmentMetadata(
 ): Promise<ReadAttachmentMetadataResult | null> {
   const { query, attachmentId } = input
   const result = await query(
-    `SELECT id, sheet_id, storage_file_id, filename, original_name, mime_type, size
+    `SELECT id, sheet_id, record_id, field_id, storage_file_id, filename, original_name, mime_type, size
      FROM multitable_attachments
      WHERE id = $1 AND deleted_at IS NULL`,
     [attachmentId],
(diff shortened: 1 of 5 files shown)


zensgit:
**✅ Post-merge independent adversarial gate: MERGE_CLEAN — F2 is fully closed.** (Read-only; no code touched.) The download endpoint now enforces BOTH halves of the mask, fail-closed, live per-request, via canonical resolvers:
- **record_id + field_id now SELECTed** (`attachment-service.ts:474`, mapped `:486-487`) — the missing data that made a record/field check impossible pre-fix.
- **Row-deny** → `loadDeniedRecordIds` (gated on `loadRowLevelReadDenyEnabled` + `!isAdminRole`) → **404** (`univer-meta.ts:15354-15360`). **Field-mask** → `loadAllowedFieldIds` → **403 FIELD_FORBIDDEN** (`:15362-15367`). Both halves, not one. Mutation-probe: neutering EACH gate turns exactly its test RED (12/12 restored).
- **Live per-request + fail-closed** — resolvers run on every GET, nothing cached at upload (closes the retained-UUID-after-revocation vector); any throw → route catch → file never served.
- **Canonical mechanism reused** (same resolvers as REST `/view`); K3 untouched. Sole client byte-path — no `storage_file_id` exposure.

**P2 (should-fix — CI coverage):** the F2 regression tests (`multitable-attachments.api.test.ts`) run in NO CI gate (excluded at `vitest.config.ts:135`, absent from `plugin-tests.yml`) → a future regression neutering either gate stays green. I can open a CI-coverage PR (behavior-neutral, open-PR). **P3:** upload permits null record_id/field_id (bypass needs write access, matches DELETE sibling); `files.ts` download is authenticate-only (pre-existing, separate namespace). Report `/tmp/attachment-readgate-3973-gate-20260708.md`.

zensgit:
**Correction to my CI-coverage note (verified the real situation — it is not a one-line un-exclude).** `tests/integration/multitable-attachments.api.test.ts` is excluded at `vitest.config.ts` **not merely for convenience** — the comment there records that **8 of its tests have a pre-existing real-wire failure** (`CommentService.mapRowToComment` drops `containerId/targetId/targetFieldId`), tracked under its own opt-in fix. And it is **not** in the `Run multitable real-DB integration` run-list (`plugin-tests.yml:179+`) either (that "now wired in" comment refers to `multitable-context.api.test.ts`, a different file). So:
- **Naively un-excluding it → CI goes RED on the 8 pre-existing failures**, unrelated to F2. Not safe as-is.
- The F2 row-deny/field-mask download tests themselves DO pass (I ran 12/12 green in the gate).

**Two clean paths to actually protect the F2 fix in CI (your call):**
1. **Extract** the F2 row-deny/field-mask download assertions into a new self-contained real-DB test file (green, no `mapRowToComment` dependency) and add it to the `Run multitable real-DB integration` run-list — protects F2 now, independent of the tracked serialization bug. I can open this as a behavior-neutral PR if you want it.
2. **Fix `mapRowToComment`** (the tracked serialization bug) first, then un-exclude the whole file + add it to the real-DB run-list — heavier, closes the 8 debts too, but touches `CommentService` (your call under red-line-#3 adjacency).

Recommend (1) for immediate F2 protection; (2) as the eventual full-file rehab. Say the word and I open the (1) PR. The F2 fix itself remains verified-closed — this is only about regression protection.

PR #3976, opened 2026-07-09
test(ci): CI-cover the F2 attachment read-gate security regression (extract to green real-DB file)

zensgit:
## Summary

- #3973 (F2) added attachment-download row-deny (404) / field-mask (403 FIELD_FORBIDDEN) enforcement, but its regression tests live inside `tests/integration/multitable-attachments.api.test.ts`, which is on the `vitest.config.ts` exclude list and is **not** wired into any real-DB CI job — so a future regression re-opening F2 would not turn CI red.
- Naively un-excluding that parent file is not safe on its own merits: it sits in the exclude list directly under the `comments.api.test.ts` entry, whose comment documents "8 pre-existing real-wire failures (`CommentService.mapRowToComment` drops containerId/targetId/targetFieldId)". **On inspection that comment belongs to `comments.api.test.ts`, not `multitable-attachments.api.test.ts`** — the attachments file self-mocks its DB pool (`vi.doMock`/`vi.spyOn(poolManager, 'get')`) and all 12/12 of its own tests pass standalone with no live DB and no `mapRowToComment` involvement at all (verified locally, see Test plan). So the F2 tests were never blocked by that specific 8-test failure; they were simply never wired into any CI job because the whole parent file predates the real-DB lane convention.
- This PR extracts just the two F2 assertions into a new self-contained real-DB file, `tests/integration/multitable-attachment-readgate.security.test.ts` (real Postgres via `poolManager`, same fixture pattern as `multitable-rowlevel-readdeny-enforce.test.ts` / `multitable-records-read-field-mask.test.ts` — no SQL-string-matching mocks to drift), and wires it as a WHOLE FILE into the `Run multitable real-DB integration` step in `plugin-tests.yml`.
- The original `multitable-attachments.api.test.ts` is untouched (still mock-based, still excluded, still not wired) — this PR does not attempt to fix or un-exclude it; that remains separate, lower-priority cleanup (refs the file's own future conversion, not any specific tracked issue).
- Added a clarifying comment in `vitest.config.ts` next to the `multitable-attachments.api.test.ts` exclude entry so the misattribution doesn't recur, and pointing at the new file as F2's actual CI coverage.

## Why this shape

- New capability tests belong in an independent module/file, not stacked onto an already-excluded mock file.
- Real-DB fixtures (namespaced by `Date.now()`, cleaned up in `afterAll`) instead of SQL-string mocks, per this repo's stated preference to avoid mock/wire drift.
- Zero permission/runtime code changes — test + CI wiring only.

## Mutation proof (both F2 gates independently)

Ran against a fresh scratch Postgres DB (`metasheet_f2ci_test`, migrated from a clean create) with `vitest.integration.config.ts`:

1. Baseline: `multitable-attachment-readgate.security.test.ts` → **3/3 green** (sentinel + row-deny 404 + field-mask 403).
2. Neutered the row-deny gate in `univer-meta.ts` (`ensureAttachmentDownloadReadable`, the `if (metadata.recordId && !access.isAdminRole && ...)` block guarding the 404) by forcing its condition to `false` → the row-deny test went **RED** (`expected 500 to be 404`, storage-read now falls through to a real download attempt against a nonexistent file). Restored via `git checkout --` (file was untouched relative to `origin/main` before and after, confirmed by `git diff origin/main --stat` showing zero hits on `univer-meta.ts`).
3. Re-ran → back to 3/3 green.
4. Neutered the field-mask gate (the `if (metadata.fieldId) { ... }` block guarding the 403) the same way → the field-mask test went **RED** (`expected 500 to be 403`). Restored via `git checkout --`.
5. Re-ran → back to 3/3 green.

This proves the extracted file actually exercises both F2 branches (not vacuously green).

## Test plan

- [x] `DATABASE_URL=postgresql://postgres@localhost:5432/metasheet_f2ci_test npx vitest --config vitest.integration.config.ts run tests/integration/multitable-attachment-readgate.security.test.ts` → 3/3 passed.
- [x] Mutation proof on both gates (see above) — RED then restored to GREEN.
- [x] Confirmed the original `tests/integration/multitable-attachments.api.test.ts` runs 12/12 green standalone (temporarily un-excluded locally only, then reverted) — establishing that it is NOT blocked by the `mapRowToComment` issue attributed to a neighboring exclude-list entry.
- [x] `git diff --name-only origin/main` = exactly `.github/workflows/plugin-tests.yml`, `packages/core-backend/tests/integration/multitable-attachment-readgate.security.test.ts`, `packages/core-backend/vitest.config.ts` — zero permission/runtime code changed, original attachments test file byte-identical to `origin/main`.
- [ ] CI green on this PR (`Run multitable real-DB integration` step).

Not merging / not enabling auto-merge — behavior-neutral test/CI work for review.


zensgit:
## 独立对抗门禁 — MERGE_CLEAN ✅

独立 gate(报告 `/tmp/pr3976-pr3853-gate-20260709.md`)在新鲜迁移 Postgres 上实跑:baseline 3/3 绿;两个 F2 闸独立 mutation 证明(neuter row-deny 闸 → expected 500 to be 404;neuter field-mask 闸 → expected 500 to be 403;恢复后全绿)——断言与 #3973 落地实现逐字节对应(univer-meta.ts 404 NOT_FOUND / 403 FIELD_FORBIDDEN),非镜像 fixture。fixture 命名空间唯一(`*_argate_*`),与两个兄弟套件同 vitest 同 pg 合跑 24/24 无碰撞。workflow 改动 append-only,接进 required `test (20.x)` lane = 真 merge-blocking。1 NIT(文件注释 'fails-not-skips' 措辞不精确,wired step 硬编码 DATABASE_URL 故无害)。

PR #3998, opened 2026-07-09
fix(directory): case-insensitive admission uniqueness + server-side batch-admit eligibility (#3972 P2-1)

zensgit:
## Summary

The #3972 review found — and proved on real Postgres — a P2 correctness defect in DingTalk directory admission: the create-time uniqueness backstop was case-sensitive while directory sync's own matching (and login) is case-insensitive, so a stored `Alice@x.com` plus an admission of `alice@x.com` slips past the check and creates a **second** `users` row for the same person. The review drove `batchAdmitDirectoryAccountUsers` on real PG and watched the user count go 1→2. This is bounded today only because the live DB happens to have zero mixed-case emails — it is a live, unbounded defect, not a hypothetical.

The review also flagged that the "no local user / no recommendation candidate" eligibility gate for batch-admit lives only in the Vue computed (`selectedReviewAdmissionIds` in `DirectoryManagementView.vue`) and is never enforced server-side, so a direct `POST /api/admin/directory/accounts/batch-admit-users` can bypass it.

This PR fixes both, minimally and locally within `packages/core-backend/src/directory/directory-sync.ts`:

1. **Case-insensitive uniqueness** — `createDirectoryAdmittedUserInTransaction`'s email existence check now uses `lower(email) = lower($1::text)` (matching the existing case-insensitive username check right below it, and the `users_email_lower_idx` index already in place). The `idx_users_email` unique index in Postgres is *also* case-sensitive, so it does not backstop this at the DB layer — the app-level check is the only guard. Username was already case-insensitive; mobile is digits-only and needed no change.
2. **Server-side batch-admit eligibility** — a new `assertDirectoryAccountEligibleForBatchAdmission` helper rejects an account that (a) is already linked to a local user, or (b) has an email/mobile that (case-insensitively) already matches an existing active user — mirroring the UI's eligibility rule at the minimum-safe level, so a direct API call can no longer create a duplicate or silently repoint an existing link.

## Real-DB proof

New file `packages/core-backend/tests/integration/directory-admission-case-insensitive-uniqueness.db.test.ts` (`describeIfDatabase`, real Postgres) reproduces the review's exact scenario and both fixes, isolated from each other:

- `createDirectoryAdmittedUserInTransaction rejects a case-differently-cased email match (no duplicate)` — drives the create-time check directly (bypassing the batch-admit gate), seeding `Alice@x.com` then admitting `alice@x.com`. Asserts it throws and the user count for that email stays **1**, not 2.
- `batchAdmitDirectoryAccountUsers rejects an account whose email case-insensitively matches an existing user` — end-to-end through the actual vulnerable route's service function.
- `batchAdmitDirectoryAccountUsers rejects an account already linked to a local user` — isolates the new eligibility gate (no email/mobile at all, so only the "already linked" branch can be what rejects it).
- `admits a fresh, non-colliding account (control)` — proves the new gate isn't overly broad.

**Mutation-proof (both parts, isolated):**
- Reverted `lower(email) = lower($1::text)` → `email = $1::text` → only the isolated create-time test went RED (`expected false to be true` — the check no longer threw); the other 3 stayed green. Restored, all 4 green again.
- Neutered the "already linked" branch of the new eligibility gate (`if (false && ...)`) → only the isolated eligibility test went RED, and it caught the actual vulnerability directly: `batchAdmitDirectoryAccountUsers` created a **second** `users` row for the already-linked account instead of rejecting it. The other 3 tests stayed green. Restored, all 4 green again.

## CI wiring (two-point, no skip-green)

- `packages/core-backend/vitest.config.ts` — added to the no-DB default job's `exclude` list (DATABASE_URL-gated, would otherwise skip-green).
- `.github/workflows/plugin-tests.yml` — wired as a **whole file** into the existing `Run approval real-DB integration` step (line ~399, alongside `directory-sync-admission-orphan-guard.db.test.ts`), so it runs against real Postgres every PR.

## Test plan

- [x] `npx tsc --noEmit -p tsconfig.json` — exit 0
- [x] `npx vitest run tests/unit` — 338 files / 4520 tests passed
- [x] Real-DB: `directory-admission-case-insensitive-uniqueness.db.test.ts` — 4/4 passed (and alongside the other directory `.db.test.ts` files — 48/48 passed together)
- [x] Updated `directory-sync-bind-account.test.ts` (mock-based) for the two new query calls in the batch-admit success path, plus two new negative-path unit tests for the eligibility gate — 18/18 passed
- [x] `admin-directory-routes.test.ts` (mocks the whole `directory-sync` module) — unaffected, 44/44 passed

Refs the #3972 review's real-DB finding. Not opened against #3972 (already merged) — this is a follow-up fix.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -396,6 +396,7 @@ jobs:
             tests/integration/approval-node-signature-policy.api.test.ts \
             tests/integration/dingtalk-approval-card-deliveries.db.test.ts \
             tests/integration/dingtalk-container-login.db.test.ts \
+            tests/integration/directory-admission-case-insensitive-uniqueness.db.test.ts \
             tests/integration/directory-primary-department-write.db.test.ts \
             tests/integration/directory-primary-department-backfill.db.test.ts \
             tests/integration/directory-sync-preview-apply-parity.db.test.ts \
--- a/packages/core-backend/src/directory/directory-sync.ts
+++ b/packages/core-backend/src/directory/directory-sync.ts
@@ -3951,6 +3951,52 @@ export async function batchBindDirectoryAccounts(
   return outcome
 }
 
+/**
+ * P2-1 (post-#3972 review): "only admit pending items with no local user and no
+ * recommendation candidate" is enforced ONLY in the Vue computed
+ * (`selectedReviewAdmissionIds` in DirectoryManagementView.vue) — a direct POST to
+ * batch-admit-users bypasses it entirely. This mirrors that rule at the minimum-safe
+ * level for the server: reject an account that is already linked to a local user, or
+ * whose email/mobile (case-insensitively — recommendations themselves are built from a
+ * case-insensitive email match, see loadDirectoryReviewRecommendations) already maps to
+ * an existing active user, instead of silently admitting a duplicate or overwriting an
+ * existing link.
+ */
+async function assertDirectoryAccountEligibleForBatchAdmission(
+  account: Pick<DirectoryBindingTargetAccountRow, 'id' | 'email' | 'mobile'>,
+): Promise<void> {
+  const linkResult = await query<{ link_status: string | null; local_user_id: string | null }>(
+    `SELECT link_status, local_user_id
+     FROM directory_account_links
+     WHERE directory_account_id = $1::uuid
+     LIMIT 1`,
+    [account.id],
+  )
+  const link = linkResult.rows[0]
+  if (link?.link_status === 'linked' && normalizeText(link.local_user_id)) {
+    throw new Error('Directory account is already linked to a local user')
+  }
+
+  const normalizedEmail = normalizeText(account.email).toLowerCase()
+  const normalizedMobile = normalizeText(account.mobile)
+  if (!normalizedEmail && !normalizedMobile) return
+
+  const matchResult = await query<{ id: string }>(
+    `SELECT id
+     FROM users
+     WHERE COALESCE(is_active, TRUE) = TRUE
+       AND (
+         ($1::text <> '' AND lower(email) = $1::text)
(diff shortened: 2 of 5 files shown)


zensgit:
## Gate: ACCEPT — closes #3972 P2-1 (the case-sensitivity duplicate-user defect)

Verified the fix against the review's proven scenario. Two localized changes on `directory-sync.ts` (58 lines, no stray edits to the merge-train hot regions):

1. **Create-time uniqueness is now case-insensitive** — `lower(email) = lower($1)`, matching the already-CI username check and the `users_email_lower_idx`. Confirmed independently that Postgres's own `idx_users_email` is case-sensitive too, so the app-level check was the only backstop.
2. **Server-side batch-admit eligibility** (`assertDirectoryAccountEligibleForBatchAdmission`) — rejects already-linked accounts and any whose email/mobile (case-insensitively) matches an active user, closing the curl-bypass of the UI-only gate. **Not over-broad**: it early-outs for accounts with no email/mobile, and a genuinely new account (no link, no match) still admits — so legitimate admissions are unaffected.

Mutation proof (both isolated): revert `lower()` → the create-time real-DB test RED; neuter the already-linked branch → the eligibility test RED (users 1→2 reproduced). Real-DB test 4/4 + 48/48 alongside siblings; two-point wired (`vitest.config.ts` exclude + whole-file in the approval real-DB step). `tsc` 0; full unit 338/4520.

**Hold recommendation**: this PR touches `directory-sync.ts`, which #3911 (the last hardening-train car) is trying to land into. Keep it **unarmed until #3911 merges**, then rebase — otherwise it re-DIRTYs #3911's window. One residual, non-blocking: the route-layer status-code mapping for the new eligibility errors wasn't driven end-to-end via HTTP (relies on the existing regex catch-block mapping) — worth a follow-up route test but not a blocker.

PR #3975, opened 2026-07-09
feat(multitable): 4c-3 — record-undelete inbound-edge replay（Option A，flag 默认 off）

zensgit:
## 4c-3 record-undelete inbound-edge replay（design-lock owner-ratified 2026-07-09，语义 = **Option A** 邻居同意）

**主循环实现**（指派的 Sonnet agent 两度死于 stalled-response / session 限额，未写代码；锁文即 spec，逐条照做）。commit 切分照锁 §10。

## 逐 commit
1. **c1 锚**：`meta_records_trash.delete_revision_id`（forward-only、nullable、无回填；禁启发式反推——错 vintage 风险）；`deleteRecord` 复用已预生成的 revision id 写入；42703 部署窗降级为旧 INSERT 形。
2. **c2 重放 helper**：六道前置 + NOT EXISTS 全部 SQL 谓词化（FK 23503 会整单回滚）；锚查询恒 `source_revision_id + reason`，**永不**按 `sheet_id`（存的是被删记录的 sheet，跨 sheet 陷阱）或 `foreign_record_id` 过滤；诊断遍与 INSERT 谓词步调一致，回传逐原因 skip 计数；`// lock-exempt: … index convergence to the neighbour's own data; no new information`（§5）。
3. **c3 接线**：restore 事务内 trash 行 `FOR UPDATE` 重读（C4）；inbound 重放排在 outbound **之后**（NOT EXISTS 须看见 outbound 行，self-link 不重）；NULL 锚/零 tombstone ⇒ 零重放 + `recoverable=false`（C1 诚实）；PIT resurrect **复用同一 helper**（§7 无语义分叉；锚 = 该记录最新 delete revision——记录此刻不存在，最近一次删除即销毁现场者，确定性非猜测；未捕获的删除→零 tombstone→诚实零重放）；trash 列表逐行 `inboundEdgesRecoverable`（§6，批量存在性查询，omitted-when-false + flag-gated 保形）。
4. **c4 D-3**：PIT-reset 内联删除补捕获（预生成 revision id → 捕获在清 link **之前** → 锚进 revision+trash；超 cap 抛错→整个 reset 回滚 fail-closed，复用既有 409 映射）。
[...]

PR #4004, opened 2026-07-09
docs(multitable): D-2 side-door delete recoverability design-lock — RATIFIED (owner 2026-07-12) + errata

zensgit:
**Docs-only, gate-front** — design-lock for the owner-gated D-2 item (R9 round, /goal 接续多维表历史记录与版本恢复线). No runtime authorized until ratified.

### What
The recoverability half of the destruction-path gap audit: plugin-SDK / automation `delete_record` deletes emit a D-1 delete revision but write **no `meta_records_trash` row and no inbound tombstones** — unrecoverable, unlike UI deletes.

- **Scope finding (narrows the audit)**: PIT-reset inline delete is ALREADY at full parity as-built (4c-3 D-3, #3975: pre-generated anchor + capture + trash w/ degradation guards, univer-meta.ts:10450-10506) ⇒ D-2 = paths 3+4 only. The gap-audit §1 rows saying otherwise are stale (flip timing = OD-4).
- **Invariants**: anchor parity (one pre-generated uuid = revision id = trash.delete_revision_id = tombstone source_revision_id, no heuristics), same-txn atomicity pinned by the D1-5b BEFORE-DELETE-trigger technique (revision-INSERT injection was proven fake-green in #3992), cap fail-closed, capture-before-links-DELETE ordering, restore-side zero delta (trash rows are source-agnostic; retention floor auto-covers), flag-off byte-identity.
- **G1-G10 goldens** incl. per-path mutation proofs, PIT-resurrect equivalence, own-only NULL-created_by visibility honesty.
- **OD-1..OD-7 owner decisions** with recommendations: full parity (a) made safe by a **new default-off flag** `MULTITABLE_SIDE_DOOR_DELETE_TRASH_ENABLED` (+1 O-2 rung); trash-growth acknowledged, retention rung stays a separate menu item; txn contract pinned in prose + CI-guarded by G3.
- **Honest consequences recorded**: machine deletes become user-restorable (side-effect re-execution), write-own actors can't see NULL-created_by trash rows, plugin cap-refusal is a flag-on API behavior change, trash gets its first machine-rate writer with no sweep.

### Verification
Every file:line was read primary-source by the audit lane against origin/main (records.ts:547-606 / automation-executor.ts:2262-2333 / univer-meta.ts:10450-10506 / record-service.ts:806-898 reference chain / meta-revision-retention.ts:229-235 floor / d1-parity golden :202-210 that must flip in the impl PR).

No code, no flags changed, zero runtime.

---

## ⚠️ RATIFIED 2026-07-12 (owner) — status flipped, + two errata

**All eight open decisions and the §1.11 truth table are resolved.** §4 records each ruling; the §8 checklist is ticked.

**Rulings:** OD-1 **(a) full parity** · OD-2 **new flag, DEFAULT OFF** (enablement is a separate O-2 rung; the implementation enables **no** env flag) · OD-3 **trash retention as a single column** · OD-4 **scope = paths 3+4 ONLY** (paths 1/2 NO CHANGE) · OD-5 **`actorId ?? null`** (plugin = `null`) · OD-6 **typed SDK error, exported from the package root** · OD-7 **txn contract — see errata** · OD-8 **accept target-base semantics**. HOT-CORE implementation authorized (strong model + independent adversarial review + txn-boundary proof).

### Errata — two defects in this lock, both found *while implementing it*

**1. The "`:10183` latest-delete heuristic" is STALE.** **R11 A′ (#4117, 2026-07-11 — after this lock was drafted)** replaced it with the vintage-exact anchor: the FIRST delete revision with **`created_at > T`**, `ORDER BY created_at ASC, version ASC, id ASC LIMIT 1`. Crucially it applies **no `source` filter** — which is *exactly why* a side-door (plugin/automation) delete revision is eligible to **be** the anchor. G5 is restated against the anchor as it actually is, and now pins both load-bearing properties: (i) no `source` filter; (ii) D-2 **pre-generates** the revision id, so trash and tombstones have something to anchor to.

**2. OD-7's CI-guard claim was FALSE as written.** It said *"the flag-on reordered path's atomicity is then CI-guarded by G3 (a non-txn regression turns G3 red)."* **It is not** — G3 supplies its **own** transaction, so unwrapping the **real** wiring (`index.ts` SDK factory; `automation-service.ts` `deps.transaction`) leaves G3 **green**. Atomicity now requires **two independent layers of evidence**:
- **Layer 1 — "atomic GIVEN a transaction":** G3a/G3b (link intact + 0 revisions + 0 tombstones + 0 trash on failure — not merely "the record still exists").
- **Layer 2 — "the real entry points actually SUPPLY a transaction":** a **production-wiring guard** driving the **actual** SDK factory and automation deps, not a test re-implementation. Acceptance: unwrapping the txn at the real call sites must turn **this** guard **red**.
- **Belt-and-braces:** the reordered path **refuses to run outside a transaction at runtime** (typed throw) — a runtime precondition cannot be silently refactored away; a test can.

**Landing order:** this lock lands **first**; the implementation (**#4168**) lands after, and must satisfy both OD-7 evidence layers.


zensgit:
**Owner P2 + adversarial-review findings folded in (still PROPOSED — ratify separately when ready).**

- **§1.11 dual-flag × schema truth table** (normative, every row golden-pinned): SIDE_DOOR-off ⇒ byte-identical D-1 even with CAPTURE on (capture is NESTED under SIDE_DOOR on paths 3+4 — review P3-1's operator-copy hazard closed); SIDE_DOOR-on+CAPTURE-off ⇒ trash+anchor, zero tombstones; both-on ⇒ 4/4 parity; **SIDE_DOOR-on + schema missing (42P01/42703) ⇒ FAIL-CLOSED refuse** (owner P2 ruling — supersedes the original mirror-the-UI-degradation stance; asymmetry vs record-service's never-fail degradation is called out as part of what gets ratified).
- New goldens: **G11** per-lane schema-missing fail-closed · **G12** PIT read-path invariance (§1.7 pinned) · G6a nesting · G9 un-nest mutation.
- **OD-8** added (review P3-4): cross-base automation deletes → target-base trash/restore semantics, recommended accept.
- §1.9/§OD-7 ordering descriptors corrected per lane (automation = emit-revision-then-DELETE inside withTransaction — review P3-2).
- Review verdict: **APPROVE 0 P1 / 0 P2 / 4 P3 / 2 NIT**, all folded; MD = /tmp/pr4004-review-claude-20260709.md.

zensgit:
Execution note: holding this design-lock, not merging it as a routine docs PR. The lock still declares PROPOSED / owner-ratification-required with OD-1..OD-8 choices, and the linked runtime line (#4168) is currently red on test (20.x). I will treat #4004 + #4168 as one D-2 gated line: ratify/merge the lock only after the owner decisions are explicit and the runtime gate is not red, rather than silently converting PROPOSED into ratified by merge.

PR #4006, opened 2026-07-10
ci(multitable): wire HistoryBatchChangesList + 3 real-mount history-center specs into web-guard (L1c)

zensgit:
**CI/test wiring only — zero product code.** R9 audit residual **L1c** (new finding, previously unrecorded).

### Problem
`apps/web/src/multitable/components/HistoryBatchChangesList.vue` is the shared renderer for the Global History Center's per-batch change lists (imported by HistoryCenterModal for BOTH the pinned-deeplink banner and normal detail), but:
- it appears in **no** path-filter trigger of `multitable-web-guard.yml`;
- its only real behavioral coverage — `multitable-history-center-inline-diff`, `multitable-history-center-ai-shortcut-label`, `multitable-history-center-pinned-batch-deeplink` (all mount the real component tree via HistoryCenterModal) — ran in **no CI gate at all** (not the guard, not the required web-tests script).

A PR touching only HistoryBatchChangesList.vue previously triggered no web-guard and ran no specs.

### Fix (mirrors the R5c ResetConfirmDialog precedent)
- Path filters: component + 3 spec files added to **both** duplicate blocks.
- Guard job vitest filter: +3 spec-name substrings.
- `apps/web/scripts/run-required-web-tests.sh` (hand-kept filter): +3 substrings, alphabetical position preserved.

### Verification
- Local full guard command: **59 files / 821 tests passed** (previously 56/810 without the 3 specs).
- YAML parse-validated.

### Not covered
- These specs still run in non-required guards only (house convention: required `test(20.x)` only builds apps/web); no required-check change here.
- Lane provenance: implemented by the Sonnet impl lane, verified green locally, landed by the main session after the lane was killed by a session limit (work was complete; only commit/push/PR remained).


zensgit:
**Owner P1 fixed**: `multitable-workbench-history-field-scope-wiring` (the real-workbench security golden for #4007/#4018) is now wired into BOTH duplicate path-filter blocks + the guard's vitest filter + `run-required-web-tests.sh`.

Landing-order safety (this PR merges before #4007 lands the spec file): verified in this worktree (spec absent, as on main today) that vitest tolerates an unmatched positional filter while others match — probe run 8/8 green, exit 0. The filter becomes live automatically when #4007 lands.

PR #4007, opened 2026-07-10
fix(multitable): History Center field list uses RBAC-scoped fields (field-NAME leak)

zensgit:
**One-line prop swap + real-mount regression spec.** R9 UX-parity audit finding (g).

### Problem
`MultitableWorkbench.vue` fed HistoryCenterModal's `fields` prop from `propertyVisibleGridFields` (property-hidden filtering only). A field the actor is **field_permissions-denied** on (but not property-hidden) still had its **name** enumerated in the History Center's field-filter `<select>` and used for diff labels. The backend already withholds that field's values/counts (LOCK-3, history-projection masking) — the option label itself was a pure FE-side name leak. The sibling TrashModal, wired three lines above, already uses the RBAC-aware `scopedAllFields`.

### Fix
Swap the prop to `scopedAllFields` (same pattern as TrashModal). One line.

### Verification
- New spec `multitable-workbench-history-field-scope-wiring.spec.ts` mounts the **real** workbench + real HistoryCenterModal/HistoryBatchChangesList (heavy siblings stubbed, same pattern as the permission-wiring spec): with `fld_2` denied via fieldPermissions, its name is absent from the filter options while a normal field's name is present.
- **Mutation-verified**: reverting the prop swap → spec red on exactly `expected [...] to not include 'SecretSalary'`.
- `vue-tsc` clean; spec green.

### Not covered
- In all-tables mode, batches from NON-active sheets have no field list threaded in at all (labels fall back to field ids) — pre-existing, unchanged, recorded in the R9 audit as a separate candidate.
- Denied fields appearing inside a batch's `changedFieldIds` now render as id-fallback (no name) — the values were already masked server-side.
- Lane provenance: implemented by the Sonnet FE lane (killed by session limit before running tests); main session completed the spec's missing mocks (presence/sheet-realtime/comments members), mutation-verified, and landed.

### DIFF
--- a/apps/web/src/multitable/views/MultitableWorkbench.vue
+++ b/apps/web/src/multitable/views/MultitableWorkbench.vue
@@ -534,7 +534,7 @@
       :open="showHistory"
       :base-id="activeBaseId || ''"
       :sheet-id="workbench.activeSheetId.value"
-      :fields="propertyVisibleGridFields"
+      :fields="scopedAllFields"
       :initial-batch-id="historyDeepLinkBatchId"
       @close="closeHistory"
     />
--- a/apps/web/src/multitable/views/MultitableWorkbench.vue
+++ b/apps/web/src/multitable/views/MultitableWorkbench.vue
@@ -534,7 +534,7 @@
       :open="showHistory"
       :base-id="activeBaseId || ''"
       :sheet-id="workbench.activeSheetId.value"
-      :fields="scopedAllFields"
+      :fields="historyVisibleFields"
       :initial-batch-id="historyDeepLinkBatchId"
       @close="closeHistory"
     />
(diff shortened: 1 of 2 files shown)


zensgit:
**Owner P1 fixed (layer-2 ∩ layer-3) + amended into this PR.**

- `historyVisibleFields = filterPropertyVisibleFields(scopedAllFields)` — BOTH visibility layers now filter the History field list (the original swap to RBAC-only `scopedAllFields` re-leaked property-hidden names, exactly as the owner's deep review found).
- Spec extended with the property-hidden case (`fld_3 HiddenNotes`, `property.hidden: true`) alongside the RBAC-denied case; header comment rewritten to describe the two-layer model.
- **Mutation-verified both directions**: `scopedAllFields` alone → red on exactly `not include 'HiddenNotes'`; `propertyVisibleGridFields` alone → red on exactly `not include 'SecretSalary'`; intersection → green.

PR #4009, opened 2026-07-10
test(dingtalk): fault-inject ledger mark-failed write to prove deliveryStuckPending surfaces (#3900 P2)

zensgit:
## What / why

#3900 P2 recipe: the DT-HARDEN-06 guard in `automation-executor.ts`'s
`executeSendDingTalkApprovalCard` catch block — surfacing
`deliveryStuckPending`/`deliveryLedgerError` in the returned step output
when the DingTalk send fails AND the ledger `markDingTalkApprovalCardDeliverySendFailed`
bookkeeping write ALSO fails — had zero test coverage. Reverting the catch
to `.catch(() => null)` and dropping the `ledgerError` output spread kept
every existing test in `automation-dingtalk-approval-card-action.test.ts`
green.

This PR is test-only. Product code on `main` is already correct (per
prior #3900/74114d742 landing) — this adds the missing falsifiability.

## Design

Reuses the already-whitelisted real-DB suite
`tests/integration/automation-dingtalk-approval-card-action.test.ts`
(already in `plugin-tests.yml`'s real-DB whitelist and already excluded
from the no-DB `vitest.config.ts` job — no new two-point wiring needed).

- Wraps the `AutomationService`'s injected `queryFn` with a one-shot
  `failNextLedgerMarkFailed` flag. When set, it matches the mark-failed
  UPDATE by SQL shape (`dingtalk_approval_card_deliveries` + `send_status
  = 'failed'`) and throws `simulated ledger outage` once; every other
  query passes through to the real pool unchanged.
- New test reuses the existing bound-recipient fixture (DingTalk
  directory link established by the "bound recipient" test earlier in
  the same `describe` block), sets `failNextSend` + `failNextLedgerMarkFailed`,
  triggers `approvals.createApproval(...)`, then polls
  `multitable_automation_executions.steps` (the event-driven execution
  record — the non-brittle oracle) for the `send_dingtalk_approval_card`
  step and asserts:
  - `status === 'failed'`
  - `output.deliveryId` is set
  - `output.deliveryStuckPending === true`
  - `output.deliveryLedgerError` contains `'simulated ledger outage'`
  - the ledger row itself is still `send_status='pending'` (the actual
    stuck condition the flag surfaces)
- Also asserts `ledgerMarkFailedInjectorFired === true` — a matcher-rot
  guard: if the mark-failed helper's SQL shape is later refactored, the
  injector could silently stop firing and the test would only be
  exercising the ordinary (never-fails) happy send-fail path without
  this check.

## Test evidence

Ran against a fresh local Postgres DB (migrated from scratch), using the
same `vitest.integration.config.ts` invocation `plugin-tests.yml` uses
for this file (bypasses the no-DB job's exclude list):

```
✓ sentinel: DATABASE_URL set
✓ placement gate: the card action saves ONLY on approval.task_created (create + update)
✓ unbound recipient: skipped person-delivery telemetry, NO card row, mapping never guessed
✓ bound recipient: ledger-first send — sent card, task_id, signed values-free deep link
✓ send failure is traceable: send_status=failed + send_error on the ledger row
✓ DT-HARDEN-06: send fails AND the ledger mark-failed write ALSO fails — deliveryStuckPending/deliveryLedgerError surface in the persisted execution record, not swallowed

Test Files  1 passed (1)
     Tests  6 passed (6)
```

`npx tsc --noEmit` clean.

## Mutation table

| Mutation | Result |
|---|---|
| Revert `automation-executor.ts:2756-2773` catch block to `await markDingTalkApprovalCardDeliverySendFailed(...).catch(() => null)`, output `{ deliveryId }` only (dropping the `ledgerError` spread) | New DT-HARDEN-06 test goes **RED** (`expected undefined to be true` on `deliveryStuckPending`); the other 5 tests in the file stay green — confirms the new test is the ONLY one covering this property and the mutation is precisely targeted. Mutation applied, run, then restored via `git checkout --` (implementation was committed first per lane protocol; final diff on `automation-executor.ts` is empty). |
| Matcher-rot guard (`ledgerMarkFailedInjectorFired` assertion) | Not separately mutation-tested (would require actually drifting the helper SQL, out of scope for a test-only PR) — documented as a design decision per the brief's risk note: protects against the mark-failed helper's SQL shape changing under the fault injector without anyone noticing the fault injection quietly stopped firing. |

## Honest gaps

- The matcher-rot guard itself isn't mutation-proven in this PR (see table above) — it's a defensive assertion, not a property with its own red/green cycle here.
- Per the brief's `testStrategy`, a second "cheap mutant" (hardcoding `deliveryStuckPending: true` unconditionally on the happy send-fail path) would be caught by adding a one-line assertion to the *existing* "send failure is traceable" test. Per lane discipline ("another lane is adding a different test case to the same file; keep yours a self-contained new test case"), this PR does NOT touch that existing test — flagging this as a follow-up for the reviewer/owner rather than doing it here to avoid cross-lane file conflicts.

## Notes for reviewer

- Product code (`automation-executor.ts`) is untouched in the final diff — the mutation was applied, run, and reverted per the lane's mutation-testing protocol, never committed.
- Branch is test-only per lane brief. Not armed for auto-merge.

### DIFF
--- a/packages/core-backend/tests/integration/automation-dingtalk-approval-card-action.test.ts
+++ b/packages/core-backend/tests/integration/automation-dingtalk-approval-card-action.test.ts
@@ -268,6 +268,34 @@ describeIfDatabase('A-2b send_dingtalk_approval_card action (real DB)', () => {
     expect(rows).toHaveLength(1)
     expect(rows[0].send_status).toBe('failed')
     expect(String(rows[0].send_error)).toContain('Failed to send DingTalk action-card work notification')
+
+    // Negative direction of DT-HARDEN-06 (gate hardening, #4009 P2): here the send fails but the
+    // ledger mark-failed write SUCCEEDS (no fault injection — failNextLedgerMarkFailed stays
+    // false) — send_status='failed' landed on the row above via the ordinary path. deliveryStuckPending
+    // /deliveryLedgerError are an exceptional signal reserved for when the bookkeeping write ITSELF
+    // failed, not a generic "send failed" flag. Without this assertion, hardcoding
+    // `deliveryStuckPending: true` unconditionally on every send-fail output would survive the whole
+    // suite: it would make every ordinary, correctly-ledgered failure scream "stuck pending" and kill
+    // the flag's signal value for operators chasing real stuck rows.
+    const executionRows = await waitFor(async () => {
+      const r = await q(
+        `SELECT steps FROM multitable_automation_executions WHERE rule_id = $1 ORDER BY created_at DESC`,
+        [rule.id],
+      )
+      return r.rows
+    })
+    expect(executionRows.length).toBeGreaterThan(0)
+    const rawSteps = (executionRows[0] as { steps: unknown }).steps
+    const steps = (typeof rawSteps === 'string' ? JSON.parse(rawSteps) : rawSteps) as Array<Record<string, unknown>>
+    const cardStep = steps.find((s) => s.actionType === 'send_dingtalk_approval_card') as
+      | { status: string; output?: { deliveryId?: string; deliveryStuckPending?: boolean; deliveryLedgerError?: string } }
+      | undefined
+    expect(cardStep).toBeDefined()
+    expect(cardStep!.status).toBe('failed')
+    expect(cardStep!.output?.deliveryId).toBeTruthy()
+    expect(cardStep!.output?.deliveryStuckPending).toBeUndefined()
+    expect(cardStep!.output?.deliveryLedgerError).toBeUndefined()
+
     await svc.deleteRule(rule.id, SHEET_ID)
   })
 
--- a/packages/core-backend/tests/integration/automation-dingtalk-approval-card-action.test.ts
+++ b/packages/core-backend/tests/integration/automation-dingtalk-approval-card-action.test.ts
@@ -30,7 +30,27 @@ const DD_ACCOUNT = randomUUID()
 const DD_USER_ID = `dd_ext_${TS}`
 
 const q = (sqlText: string, params?: unknown[]) => poolManager.get().query(sqlText, params)
-const queryFn = ((sqlText: string, params?: unknown[]) => poolManager.get().query(sqlText, params)) as never
+
+// DT-HARDEN-06 fault injector: a one-shot flag that makes the NEXT `dingtalk_approval_card_deliveries`
+// mark-failed write (the ledger UPDATE ... SET send_status = 'failed' issued from the executor's
+// catch block) throw instead of hitting the real pool. Everything else passes through unchanged, so
+// this is the fault-injection seam only — assertions read the persisted execution record, not this
+// wrapper. `ledgerMarkFailedInjectorFired` proves the matcher actually fired (guards against the
+// helper SQL shape drifting and the injector silently going quiet).
+let failNextLedgerMarkFailed = false
+let ledgerMarkFailedInjectorFired = false
+const queryFn = (async (sqlText: string, params?: unknown[]) => {
+  if (
+    failNextLedgerMarkFailed &&
+    sqlText.includes('dingtalk_approval_card_deliveries') &&
+    sqlText.includes("send_status = 'failed'")
+  ) {
+    failNextLedgerMarkFailed = false
+    ledgerMarkFailedInjectorFired = true
+    throw new Error('simulated ledger outage')
+  }
+  return poolManager.get().query(sqlText, params)
+}) as never
 
 let svc: AutomationService
 let approvals: ApprovalProductService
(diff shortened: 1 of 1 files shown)


zensgit:
# Adversarial review — PR #4009

- **PR**: #4009 `test(dingtalk): fault-inject ledger mark-failed write to prove deliveryStuckPending surfaces (#3900 P2)`
- **Head reviewed**: `6f5f1e9cc3a93b35df97a71719849025ded3466e` (branch `claude/dt-ledger-stuckpending-test`, base `main`, re-pinned after review — unchanged)
- **Scope of diff**: 1 file, `packages/core-backend/tests/integration/automation-dingtalk-approval-card-action.test.ts` (+69/−1). `git diff origin/main refs/reviews/pr4009 --stat` confirms **zero product-code changes** — the PR-body claim "test-only, executor untouched" is true.
- **Reviewer**: Claude (independent adversarial gate), 2026-07-09
- **Method**: own disposable worktree at the PR head → `pnpm install --frozen-lockfile` → fresh Postgres DB (`rev4009_1783643292` on :5435), migrated from scratch → real test runs → reviewer-fired mutations (not trusting the author's) → CI-gate wiring check → cleanup (DB dropped, worktree removed).

## Verdict: **APPROVE with hardening** (1×P2, 1×P3, 1×NIT — no P1)

The test is real, non-tautological, wired into a gate that actually runs on every PR, and the load-bearing mutation goes RED under my own hand. The one substantive gap (false-positive direction of `deliveryStuckPending` unguarded) is author-disclosed, but the stated reason for deferring it — a parallel lane touching the same file — is **moot**: no other open PR touches this file, so the one-block fix can land now (in this PR or an immediate follow-up).

## What I actually ran (evidence, not reading)

| Run | Result |
|---|---|
| Clean PR head, fresh migrated DB: `npx vitest --config vitest.integration.config.ts run tests/integration/automation-dingtalk-approval-card-action.test.ts` | **6/6 passed**. Logged stack proves the fault propagates through the real path: `automation-executor.ts:2758 → markDingTalkApprovalCardDeliverySendFailed (approval-card-deliveries.ts:139) → test queryFn throw` — assertions read the persisted `multitable_automation_executions` row, not the injector. |
| **Mutation 1 (reviewer-fired)**: reverted the executor catch (automation-executor.ts:2756–2773) to `await markDingTalkApprovalCardDeliverySendFailed(...).catch(() => null)` + `output: { deliveryId }` only | **RED** — DT-HARDEN-06 fails `expected undefined to be true` on `deliveryStuckPending`; other 5 stay green. Matches the author's claim exactly. Restored via `git checkout --`. |
| **Mutation 2 (reviewer-fired, the disclosed gap)**: hardcode `deliveryStuckPending: true` unconditionally on the send-fail output | **ALL 6 GREEN** — gap is real; the flag's *selectivity* (absent when the ledger write succeeds) has zero coverage. Restored. |
| **Matcher-rot drift (reviewer-fired)**: changed the helper SQL to `send_status='failed'` (no spaces) so the injector's substring match goes quiet | **RED** — DT-HARDEN-06 fails on the `deliveryStuckPending` assertion (injector never fires → mark-failed succeeds → flag absent). So the test **cannot silently go green under SQL-shape drift**; the `ledgerMarkFailedInjectorFired` assertion is belt-and-braces diagnostics, not the only protection. Restored. |
| `npx tsc --noEmit -p tsconfig.json` in `packages/core-backend` | Clean. |
| Neighbor suites (env health): `automation-approval-task-created-trigger.test.ts` + `dingtalk-approval-card-deliveries.db.test.ts` on the same fresh DB | **2/2 files passed**. |
| Final worktree state | `git status --short` empty — all mutations reverted, nothing committed. DB dropped, worktree removed. |

## Gate wiring (two-point check)

- `packages/core-backend/vitest.config.ts:86` — file in the no-DB job's exclude list (can't skip-green). Pre-existing.
- `.github/workflows/plugin-tests.yml:331` — file in the `Run multitable real-DB integration` run-list inside the `test` (18.x/20.x matrix) job. Workflow triggers on `pull_request: branches: [main]` with **no paths filter**, so it runs on every PR. On this PR, `test (18.x)` already **passed** (i.e. the new test ran green in CI); `test (20.x)` (required) was still pending at review time — landing must wait for it, as usual.
- No new wiring was needed or added — the PR extends an already-whitelisted suite, per the brief.

## Adversarial checks that came back clean

- **Injector selectivity**: `git grep "send_status = 'failed'"` over PR-head `src/` hits exactly one place — the mark-failed UPDATE (`approval-card-deliveries.ts:141`). Mark-sent, decision-callback, and ledger INSERT SQL don't match the predicate; no cross-fire. Cleanup/assertion queries use the un-wrapped `q`, not the injected `queryFn`.
- **No fixture shadowing / tautology**: the oracle is the persisted execution record written by `automation-log-service.ts:35` through the kysely `db` handle (single INSERT after step completion — also no row-without-steps flake window), plus the real ledger row still `pending`. The injector only decides *when* the real helper throws; it doesn't fabricate any asserted value.
- **Redaction**: `deliveryLedgerError` survives the log-service `redactValue` pass (asserted `toContain('simulated ledger outage')`, green on real DB).
- **Flag leakage**: `failNextLedgerMarkFailed` is one-shot and DT-HARDEN-06 is the last test in the file; `afterAll` uses `q`, so even a non-fired flag can't poison cleanup.
- **Scope discipline**: diff touches exactly the one file the lane brief names. No routes/index/runtime smuggling. PR body has no bare closing keywords (won't auto-close #3900).
- **queryFn async wrapper**: original was a sync fn returning `pool.query(...)`'s promise; the new `async` wrapper is call-compatible; `tsc` clean.

## Findings

### P2 — untested guard direction: unconditional `deliveryStuckPending: true` survives all 6 tests
`packages/core-backend/tests/integration/automation-dingtalk-approval-card-action.test.ts:256` (the `send failure is traceable` test is the natural home for the missing negative assertion).
**Failure scenario (proven by my mutation run)**: change `automation-executor.ts:2772` to always emit `deliveryStuckPending: true` on any send failure → all 6 tests green. In production that mutant makes every ordinary, correctly-ledgered send failure scream "stuck pending" — operators chase phantom stuck rows and the flag's signal value dies, which is precisely the property this hardening line (#3900 P2 recipe's second cheap mutant) exists to pin.
**On the deferral**: the author deferred this citing "another lane is touching the same file". I checked all open PRs — **#4009 is the only open PR touching this file**, so the collision rationale no longer holds. Fix now (here or as an immediate follow-up commit on this branch): in the `send failure is traceable` test, after the ledger-row assertions, read the execution record for that rule (same `waitFor`/steps-parse pattern as DT-HARDEN-06) and assert `cardStep.output?.deliveryStuckPending` is falsy/absent and `deliveryLedgerError` absent. ~8 lines; kills the surviving mutant.

### P3 — DT-HARDEN-06 is order-coupled to the `bound recipient` test's directory-link fixture
`packages/core-backend/tests/integration/automation-dingtalk-approval-card-action.test.ts:274-276`.
**Failure scenario**: `vitest run -t 'DT-HARDEN-06' tests/integration/automation-dingtalk-approval-card-action.test.ts` (or any future test-shuffling config) → APPROVER is unbound → card action takes the skipped-telemetry path, no delivery row, `output.deliveryId` undefined → the test fails for a reason unrelated to the property under test. Mitigating: the pre-existing `send failure is traceable` test has exactly the same coupling, the file always runs as a whole in both gates, and vitest runs tests in-file sequentially — so this is house style, not a regression introduced here. Worth a comment or a `beforeAll`-hoisted binding if the file is ever restructured; not a land-blocker.

### NIT — matcher-rot guard is redundant (keep it anyway)
`.../automation-dingtalk-approval-card-action.test.ts:307` — my SQL-drift experiment shows the `deliveryStuckPending` assertion already goes RED when the injector rots, so `ledgerMarkFailedInjectorFired` never fires alone. It still earns its keep as a diagnostic (points the debugger at the injector instead of the executor), so keep it — just don't credit it as load-bearing coverage.

## Author-report cross-check

- Claimed 6/6 on fresh DB — **reproduced**.
- Claimed mutation 1 RED with the other 5 green — **reproduced under my own hand**.
- Claimed matcher-rot guard "not mutation-proven" — I proved the stronger fact: the test fails safely under actual SQL drift even without that guard.
- Claimed second mutant unguarded — **confirmed** (mutant survives), and I additionally established the deferral's premise (parallel lane on this file) is not currently true.

*Review artifacts: worktree `/…/scratchpad/ms2-rev-pr4009` (removed), DB `rev4009_1783643292` (dropped). Nothing pushed; no product or test files modified by the reviewer.*

zensgit:
## Gate P2 hardening applied — `73d9bae2c`

Per the gate review's P2 finding ("untested guard direction: unconditional `deliveryStuckPending: true` survives all 6 tests"), extended the pre-existing `send failure is traceable` test with the missing negative case: send fails but the ledger mark-failed write **succeeds** (no fault injection — `failNextLedgerMarkFailed` stays `false`), so `send_status='failed'` lands on the ledger row via the ordinary path. Alongside the existing ledger-row assertions, the test now reads the persisted `multitable_automation_executions` step output (same `waitFor`/steps-parse pattern as DT-HARDEN-06) and asserts `cardStep.output?.deliveryStuckPending` and `deliveryLedgerError` are both `undefined` — i.e. genuinely absent from the output, matching production shape, not merely falsy.

Ran on a fresh, migrated local Postgres DB: **6/6 passed** (both directions of the property now covered), `tsc --noEmit` clean, and the two real-DB neighbor suites (`automation-approval-task-created-trigger.test.ts`, `dingtalk-approval-card-deliveries.db.test.ts`) stayed green (12/12) as an environment sanity check.

### Mutation table

| Mutation | Result |
|---|---|
| Hardcode `deliveryStuckPending: true` + inject a fake `deliveryLedgerError` unconditionally in `automation-executor.ts`'s send-fail catch-block output (the exact gap the P2 finding named) | **RED** — new assertion in `send failure is traceable` fails: `expected true to be undefined` on `deliveryStuckPending`. This is the load-bearing signal the P2 hardening exists to pin. `DT-HARDEN-06` also failed on the same run (`expected 'success' to be 'failed'`), but that's a cascade artifact, not a second independent catch: the mid-test assertion throw in `send failure is traceable` skips that test's `svc.deleteRule` cleanup, leaving a stale rule active that cross-fires on `DT-HARDEN-06`'s approval event and consumes the shared `failNextSend` flag first. That test-order coupling is pre-existing house style in this file (every test's cleanup runs only after its assertions — same shape the gate already flagged as P3, non-blocking) and is not present in the committed green state (6/6). Mutation applied, run, then reverted via `git checkout --` per lane protocol — real implementation (`automation-executor.ts`) is committed first and untouched in the final diff. |

Not armed for auto-merge. Test-only change, no product code touched.

zensgit:
Window protocol executed: rebased onto post-#3898 main (`37ad93fdd`, clean, both commits survived), fault-injection suite 6/6 on a fresh migrated DB, load-bearing mutation replayed on the rebased head (unconditional deliveryStuckPending:true hardcode ⇒ negative-direction test RED ×2, restored). Armed alone per single-window discipline.

PR #4010, opened 2026-07-10
fix(web): batch-bind directory review no longer toasts success on partial failure

zensgit:
## What / why

`#3902` P3-2 residual: `DirectoryManagementView`'s batch-bind review flow always showed a blanket success toast/progress-chip on any 2xx response from `POST /api/admin/directory/accounts/batch-bind`, even when the server committed only some of the items. The backend (DT-HARDEN-04) already returns `{ items, updatedCount, failedCount, failed }` with per-item isolation, but the frontend consumer only read `data.items` (and inflated the count to `bindings.length` when `items` was empty), discarding `failedCount`/`failed` entirely. `submitReviewBindings` then unconditionally called `setStatus(successMessage)`.

This made batch-bind inconsistent with the sibling `batchAdmitReviewItems` handler in the same file, which already reads `failedCount`/`failed`, previews failed identities, and sets error-level status (landed in #3972 for batch-admit-users).

## Fix

- `postReviewBindings` now returns the full outcome `{ appliedCount, failedCount, failed }` instead of a bare (sometimes inflated) count.
- `submitReviewBindings` builds a partial-failure message (`已绑定 N 个目录成员，另有 M 个失败：<up to 3 failed identities>`) and sets `statusTone = 'error'` when `failedCount > 0`, mirroring the batch-admit pattern; the progress chip's `applied` count and completion message now reflect the real applied count instead of the caller's selection size.
- Extracted the failed-item preview formatting (previously duplicated inline in `batchAdmitReviewItems`) into a shared `readFailedBindingsPreview(failed, limit)` helper, used by both handlers, to avoid a third copy.

Single-item callers (`handleReviewBind`, `backfillUserMobileAndBindReviewItem`, `confirmRecommendedReviewBinding`) are unaffected in practice: a 1-item batch either fully succeeds or the backend throws before reaching the success path (`succeeded.length === 0 && failed.length > 0` → throw), so `failedCount > 0` is unreachable for them.

## CI wiring

`apps/web/tests/directoryManagementView.spec.ts` (5k+ lines) previously ran in **no CI workflow** — absent from `run-required-web-tests.sh`'s hand-kept filter and from every guard workflow's path filter. Per the lane brief's explicit risk note, I ran the whole file on `main` first (51/51 green, no pre-existing red/flaky tests), then added `directoryManagementView` to `apps/web/scripts/run-required-web-tests.sh` (the required `web-tests.yml` job) in a separate commit, verified green with the fix + new test (52/52), and re-ran the full required-web-tests script end to end (108 files / 1623 tests, all green) to rule out collateral damage from the filter change.

## Test evidence

- `npx tsc --noEmit -p tsconfig.json` — clean.
- `CI=true npx vitest run tests/directoryManagementView.spec.ts` on `main` (pre-fix): 51/51 green.
- Same command on this branch (post-fix + new test): 52/52 green.
- `apps/web/scripts/run-required-web-tests.sh` (full required gate, post-wiring): 108 files / 1623 tests green, including `directoryManagementView.spec.ts (52 tests)`.

## Mutation table

| Mutation | Test result |
|---|---|
| Revert the `submitReviewBindings` message logic to unconditional success (`const message = successMessage; setStatus(message)`, dropping the `failedCount > 0` branch) | New test `reports a partial-failure status when batch-bind commits some items but not all` goes RED (`expected ... not to contain '已完成 2 个目录成员的批量绑定'` — fails because the mutated code still shows blanket success). All 51 pre-existing tests stay green under the mutation, confirming today's (pre-PR) guard for this path was non-falsifiable and the new test is the one actually pinning the behavior. Mutation reverted via `git checkout --` after recording red, then confirmed 52/52 green again on the real implementation. |

## Honest gaps

- Did not modify `batchConfirmRecommendedReviewItems` / `confirmRecommendedReviewBinding` call sites' message templates — they share `submitReviewBindings` and get the partial-failure branch "for free," but no new test exercises a partial failure specifically through the recommend-confirm path (only through `batchBindReviewItems`). Both flows call the same shared function so the fix is structurally the same, but the recommend path is unproven directly.
- No backend changes — the route/service response shape already carried everything needed (confirmed by reading `admin-directory.ts` and `batchBindDirectoryAccounts` in `directory-sync.ts`).
- Did not add this spec to `approval-web-guard.yml` / `multitable-web-guard.yml` (non-required, path-filtered guards) — only to the required `run-required-web-tests.sh` gate, per the brief's scope. Those two guards are for approval/multitable domains and don't obviously apply to a DingTalk-directory spec.

## Decisions a reviewer should scrutinize

- Chose to keep the progress chip's `phase` as `'completed'` (not `'failed'`) for a partial-failure batch, matching the existing `batchAdmitReviewItems` precedent exactly (same file) — the chip's `applied` count and message text carry the partial-failure signal instead of the phase enum.
- Refactored `batchAdmitReviewItems`'s previously-duplicated failed-item preview logic to call the new shared `readFailedBindingsPreview` helper (same `.slice(0, 2)` limit, byte-identical output) — this touches a working, tested code path outside the strict "batch-bind" defect, but was explicitly suggested by the brief's `proposedFix` ("avoid a third copy") and is covered by the unchanged existing test suite staying green.


zensgit:
# Adversarial Review — PR #4010

- **PR**: #4010 `fix(web): batch-bind directory review no longer toasts success on partial failure`
- **Branch / head**: `claude/dt-batchbind-partial-toast` @ `fb140d558dc58a006a6256f22d585736a51e0705` (verified identical to `refs/pull/4010/head` at review time)
- **Base**: `main`
- **Scope reviewed**: full diff (3 files), backend contract cross-check, CI-wiring reality check, own worktree test runs + re-fired mutations

## Verdict: APPROVE

No P1/P2. The defect is real, the fix is correct against the *actual* backend response shape, the new guard is proven load-bearing by mutations I ran myself, and the CI wiring claim (required gate) is verified true — `web-tests` is genuinely in main's required status-check contexts and the script has no path filter.

## What I verified myself (not taken from the report)

Disposable worktree `/tmp/ms2-rev-pr4010` @ PR head; second worktree `/tmp/ms2-rev-pr4010-main` @ origin/main for baseline. Both `pnpm install --frozen-lockfile --prefer-offline`.

| Check | Result |
|---|---|
| `npx tsc --noEmit -p tsconfig.json` (apps/web, PR head) | clean |
| `vitest run tests/directoryManagementView.spec.ts` @ PR head | **52/52 green** (ran 6×, deterministic) |
| Same file @ origin/main (baseline, pre-fix) | **51/51 green** — confirms no pre-existing red/flaky test was smuggled into a required gate |
| Full `apps/web/scripts/run-required-web-tests.sh` @ PR head | **108 files / 1623 tests, all green** (56.9s) — matches claim exactly |
| Filter-token accounting main→PR | 93→94 tokens, single insertion `directoryManagementView`, **no dropped tokens** (union preserved) |
| `directoryManagementView` filter uniqueness | matches exactly one spec file (`apps/web/tests/directoryManagementView.spec.ts`); 107→108 file count confirms |
| `web-tests` required? | `gh api .../required_status_checks/contexts` → `["contracts (strict)","contracts (dashboard)","pr-validate","test (20.x)","contracts (openapi)","web-tests"]` — **yes, required**; `web-tests.yml` runs the script unconditionally (no path filter) |
| Spec wired anywhere on main today? | `git grep directoryManagementView origin/main -- .github apps/web/scripts` → no hits. "Ran in no CI workflow" claim is true |
| PR CI | `web-tests` green (3m4s) on head `fb140d558` |
| PR body closing keywords | none (no accidental issue-close) |

### Backend contract cross-check (wire-vs-fixture)

Read `packages/core-backend/src/routes/admin-directory.ts:669-732` and `directory-sync.ts:3921-3953` at the PR ref:

- Route returns `jsonOk(res, { items: succeeded accounts, updatedCount: succeeded.length, failedCount: failed.length, failed })`; `jsonOk` wraps as `{ ok: true, data, meta }` — the test fixture's shape matches the wire exactly, including `failed: [{ accountId, error }]` (service pushes exactly that shape).
- Zero-success + any-failure → route **throws** (`admin-directory.ts` "Nothing committed → keep the historical error mapping") → non-2xx. So the claim "single-item callers can never see `failedCount > 0` on a 2xx" is **true against the real backend**, not just asserted.
- Corollary: on any 2xx, `items.length === updatedCount` always, and `items` is never empty (see mutation C below).

### Mutations I re-fired (evidence, not claims)

| # | Mutation (in `submitReviewBindings` / `postReviewBindings`) | Result |
|---|---|---|
| A | Revert to unconditional `setStatus(successMessage)` (drop partial-failure branch) | **RED**: new test fails on `not.toContain('已完成 2 个目录成员的批量绑定')`; other 51 stay green → pre-fix guard was non-falsifiable, new test is the sole falsifier |
| B | Keep partial message, force tone `'info'` (drop `'error'`) | **RED**: `directory-admin__status--error` class assertion fails → tone assertion is load-bearing, not decorative |
| C | Restore old count source `processedItems.length > 0 ? processedItems.length : bindings.length` | **GREEN (survives)** — analyzed: **equivalent mutant**, not a coverage gap. From the real route, `items.length === updatedCount` on every reachable 2xx and `items` is never empty on 2xx, so the mutant is behaviorally indistinguishable for any response this backend can produce |

All mutations restored via `git checkout --`; worktree verified clean, 52/52 re-green after restore.

## Lane-brief conformance

| Gate | Verdict | Evidence |
|---|---|---|
| Fix batch-bind blanket-success toast on partial failure | PASS | branch + tone + failed-preview at `DirectoryManagementView.vue:4967-4971`; mutations A/B red |
| Read real outcome (`updatedCount/failedCount/failed`), stop inflating count | PASS | `postReviewBindings` returns full outcome; chip `applied` = real `appliedCount` |
| Avoid third copy of failed-preview logic | PASS | `readFailedBindingsPreview` shared helper; admit-site refactor textually identical (`.slice(0,2)`, same map/filter/join) |
| Chip phase on partial failure: 'completed' acceptable option | PASS (option taken) | mirrors `batchAdmitReviewItems` precedent; message + error-tone status carry the signal |
| CI-wire the 5k-line spec, main-baseline first | PASS | separate commit `fb140d558`; my own main baseline 51/51; full required gate green locally + in CI |
| No backend changes | PASS | diff = 3 files, all apps/web |
| No out-of-scope files | PASS | `run-required-web-tests.sh` (+1 token), view, spec only |

## Findings

**P3-1 — Recommend-confirm and admit partial-failure paths have no direct test** (`apps/web/tests/directoryManagementView.spec.ts`)
`batchConfirmRecommendedReviewItems` / `confirmRecommendedReviewBinding` share `submitReviewBindings` (same code path, different strings), and `batchAdmitReviewItems`'s partial-failure preview (refactored to the shared helper) is exercised by **zero** tests with `failedCount > 0` — the only admit test uses `failedCount: 0` (spec ~line 5562). This was equally untested on main (the old inline block had no falsifier either), so it is a pre-existing gap the PR merely touched output-identically, honestly disclosed. Failure scenario if it ever regresses: admit partial-failure toast loses its failed-identity preview with all tests green. Cheap hardening: one admit-path test with `failedCount: 1` asserting `已创建并绑定 … 另有 1 个失败：…`.

**NIT-1 — Preview limits inconsistent between siblings** (`DirectoryManagementView.vue:4967` vs `:5191`)
Bind path previews 3 failed items, admit path 2. Each preserves its own precedent; harmless, but a shared constant would prevent drift.

**NIT-2 — New test relies on mock call-order, not endpoint assertion** (`directoryManagementView.spec.ts` new test)
The batch-bind fixture is consumed positionally; the test never asserts `apiFetchMock` was called with `/api/admin/directory/accounts/batch-bind`. Matches the file's established style, and any misrouting would fail the message assertions anyway.

## Report-vs-reality delta

One overstatement in the implementer's notes: "existing admit-flow tests stayed green, so output is provably unchanged there" — the admit tests never reach the refactored preview (all use `failedCount: 0`), so green proves nothing about it. The refactor is nonetheless safe by textual identity of the logic plus content-level coverage of the helper body through the bind path (mutation A asserts preview content `account-2: Directory account is already linked…`). Everything else in the report reproduced exactly.

## Cleanup

Review worktrees `/tmp/ms2-rev-pr4010` and `/tmp/ms2-rev-pr4010-main` removed after review; canonical checkout untouched.

zensgit:
## P3 hardening applied — commit `87e110dba`

Closes the gate review's three findings (`gh pr view 4010 --comments`, `/tmp/pr4010-review-claude-20260709.md`). No behavior change to the original fix; web-only, same branch.

### P3-1 — sibling partial-failure paths now have direct tests

Added two spec cases to `apps/web/tests/directoryManagementView.spec.ts`, mirroring the existing full-success test for each sibling exactly (same fixture/DOM flow) but with a partial `{updatedCount, failedCount, failed}` response from the mocked endpoint:

- **`batchAdmitReviewItems`** (own inline message logic, not `submitReviewBindings`): *"reports a partial-failure status when batch-admit-users commits some items but not all"* — asserts `已创建并绑定 1 个目录成员，另有 1 个失败：account-2: …` + `directory-admin__status--error`, and NOT the blanket `已创建并绑定 2 个目录成员`.
- **`batchConfirmRecommendedReviewItems`** (shares `submitReviewBindings` with batch-bind): *"reports a partial-failure status when batch-confirming recommended pending bindings commits some items but not all"* — asserts `已绑定 1 个目录成员，另有 1 个失败：…` + error tone, and NOT the blanket `已完成 2 个目录成员的推荐绑定确认`. Endpoint/body pinned to `/api/admin/directory/accounts/batch-bind` with the recommendation-derived `user-1`/`user-2` bindings.

**Not added:** a partial-failure test for the single-item `confirmRecommendedReviewBinding` (also named in the finding). Per the PR body's own analysis (confirmed by the adversarial review's mutation-C equivalence argument): a 1-item batch can never reach `failedCount > 0` against the real backend — zero successes + any failure makes the route throw before the success path, so `succeeded.length === 0 && failed.length > 0` is unreachable for a single-item caller. A test forcing that shape through the mock would be asserting behavior the real backend cannot produce, which conflicts with the lane's "fixtures must not shadow the thing under test" bar. Flagging this omission explicitly rather than silently dropping the third named path.

### NIT (preview limit) — unified to a shared constant

`DirectoryManagementView.vue`: added `FAILED_BINDINGS_PREVIEW_LIMIT = 3`, used by both the bind/recommend-confirm call site (previously `3`) and the admit call site (previously `2`). **Behavior change, flagged for ops awareness:** admit's partial-failure toast can now show up to 3 failed identities instead of 2 (harmless, cosmetic — no test asserted the old boundary, so nothing needed updating).

### NIT (fixture pinning) — endpoint/body now asserted

Added the missing `apiFetchMock` call assertion (path `/api/admin/directory/accounts/batch-bind`, `method: 'POST'`, exact `bindings` body) to the bind partial-failure test, matching the pattern already used by every other batch test in the file. Fixture reordering can no longer silently retarget it.

### Test evidence

- `npx tsc --noEmit -p tsconfig.json` (apps/web) — clean.
- `CI=true npx vitest run tests/directoryManagementView.spec.ts` — **54/54 green** (was 52/52; +2 new cases).
- `apps/web/scripts/run-required-web-tests.sh` (full required `web-tests` gate) — **108 files / 1625 tests green** (was 1623).

### Mutation table

| # | Mutation | Result |
|---|---|---|
| 1 | `batchAdmitReviewItems`: drop the `failedCount > 0` branch, force unconditional `已创建并绑定 ${appliedCount} 个目录成员` + `'info'` tone | **RED** — new admit-path test fails on `not.toContain`/`toContain` for the partial message; all other 53 tests stay green |
| 2 | `submitReviewBindings` (shared by bind + recommend-confirm): drop the `outcome.failedCount > 0` branch, force unconditional `setStatus(successMessage)` | **RED on both** — the original bind partial-failure test *and* the new recommend-confirm test fail simultaneously; 52 others stay green. This is the expected signature of genuinely shared plumbing: one mutation, two falsifiers. There is no path-isolated mutation for recommend-confirm specifically, because the partial-failure message logic *is* the same code as bind by design — the new test's distinct value is pinning the call site (endpoint/body + recommendation-derived `localUserRef`s), not an independent guard. |

Mutations applied on top of the committed real implementation, run, reverted via `git checkout --` (verified clean before push).

### Not armed, not merged

CI queued as of this push (`87e110dba`). Leaving auto-merge off per lane discipline — landing is gated by the main loop.

zensgit:
## Conflict resolved: rebased onto `origin/main` (28 commits ahead), UNION applied to `run-required-web-tests.sh`

New head: `5e4bec0b7`

### Conflict
Both sides touched the single `exec npx vitest run ...` line in `apps/web/scripts/run-required-web-tests.sh`:
- **main** added 4 new entries (`multitable-history-center-ai-shortcut-label`, `multitable-history-center-inline-diff`, `multitable-history-center-pinned-batch-deeplink`, `multitable-workbench-history-field-scope-wiring`, via #4006) and **removed** `approval-inbox-auth-guard` (via #4034 — `ApprovalInboxView.vue` and its only spec `approval-inbox-auth-guard.spec.ts` were deleted, the view was never routed).
- **this PR** added `directoryManagementView` (its only change to this file).

### Resolution rule applied: union of additions, minus one deliberate upstream deletion
Resolved list = `origin/main`'s current line (already carries main's 4 new entries) + this PR's `directoryManagementView`, inserted after `AutomationExecutionsView` (same position PR originally used).

`approval-inbox-auth-guard` is **not** a lost union entry — keeping it would point the required gate at a spec file that no longer exists anywhere in the tree (verified: `find . -iname "*approval-inbox-auth-guard*"` → no matches, post-#4034). Main's maintainer already decided this deletion in the same commit that removed the dead view; re-adding the token here would silently no-op (or worse, mask a future accidental re-creation) rather than honor that decision.

**Proof — every entry from both parents is present in the resolved list except that one legitimate deletion:**
```
main tokens missing from resolved:  (none)
PR tokens missing from resolved:    approval-inbox-auth-guard   (deleted spec file, confirmed gone)
resolved tokens beyond union(main,PR): (none)
```

### Full required gate + PR's own spec, re-run on rebased head
- `CI=true npx vitest run tests/directoryManagementView.spec.ts` → **54/54 green** (grew from the PR's original 52/52 — main landed 2 more cases upstream in the interim).
- `apps/web/scripts/run-required-web-tests.sh` (full required gate, post-rebase) → **111 files / 1649 tests, all green** (grew from 108/1623 at PR-open time, consistent with main's 4 new entries + this PR's addition). Confirmed `directoryManagementView.spec.ts` (54 tests) is actually collected and green inside that full run, not just runnable standalone.

### Mutation replay on rebased head (re-fired, not just re-cited from the original PR body)
Reverted `submitReviewBindings`'s partial-failure branch (`DirectoryManagementView.vue` ~L4970-4974) to unconditional success:
```diff
-    const failedPreview = readFailedBindingsPreview(outcome.failed, FAILED_BINDINGS_PREVIEW_LIMIT)
-    const message = outcome.failedCount > 0
-      ? `已绑定 ${appliedCount} 个目录成员，另有 ${outcome.failedCount} 个失败${failedPreview ? `：${failedPreview}` : '，请重试'}`
-      : successMessage
-    setStatus(message, outcome.failedCount > 0 ? 'error' : 'info')
+    const message = successMessage
+    setStatus(message)
```
Result: **both target tests went RED**, all other 52 stayed green (54 total: 2 failed | 52 passed):
- `reports a partial-failure status when batch-bind commits some items but not all` (L3492) — RED, exactly as originally reported.
- `reports a partial-failure status when batch-confirming recommended pending bindings commits some items but not all` (L3866) — RED. This closes the PR's original "honest gap" (the recommend-confirm path was previously unproven directly through the shared `submitReviewBindings` function; the third commit's dedicated test now demonstrably pins it).

Mutation reverted via `git checkout --`, re-confirmed 54/54 green on the real implementation before pushing.

### Not armed
Per landing order, this PR's merge window comes after #4014 — not enabling auto-merge.

zensgit:
Window protocol executed: rebased onto post-#4014 main; web-test list conflict resolved as PROVEN UNION (set-diff check: 7 main-only entries preserved — approvalDetailPolish/asyncStateBlock/approvalMetricsTopnReport/approvalMetricsView/approvalTemplateGovernance/templateArchiveConfirm/approvalTemplateVersionHistory — plus this PR's directoryManagementView; resolved set == exact union); full required web gate 118 files / 1708 tests green; load-bearing mutation replayed (partial-failure branch forced to success ⇒ RED ×2: bind + recommend-confirm), restored. Armed alone per single-window discipline.

PR #4011, opened 2026-07-10
feat(attendance): R8 — quiet-hours window for DingTalk notification delivery

zensgit:
## What

Adds a quiet-hours window that defers the C5-2 attendance notification delivery
worker's dispatch rather than lose or misdate events.

- `ATTENDANCE_NOTIFICATION_QUIET_HOURS="HH:MM-HH:MM"` — 24h clock, cross-midnight
  aware (e.g. `22:00-08:00`). Unset/empty = off (unchanged behavior). Malformed →
  off + one `warn` log, never a crash.
- `ATTENDANCE_NOTIFICATION_QUIET_HOURS_TZ` — IANA zone, defaults to `Asia/Shanghai`.
  Invalid zone → off + one `warn` log.
- Gate sits in `AttendanceNotificationDeliveryWorker.runBatch()` **before**
  `claimDueDeliveries()`: an in-window tick returns the zero-result shape
  (`{claimed:0,sent:0,retrying:0,failed:0,skipped:0}`) without touching the DB.
  Due outbox rows stay `pending`/`retrying` with `next_attempt_at` unchanged, no
  lease taken, `attempt_count` not incremented — no event loss, no retry/lease
  interaction. Once the window ends, the existing `ORDER BY next_attempt_at`
  drains them oldest-first.
- Local `HH:MM` is computed via `Intl.DateTimeFormat` in the configured
  timezone (**not** `toTimeString()`), unlike the pre-existing
  `NotificationService.isInQuietHours` precedent (a separate, per-subscription
  in-app path) whose wrap-logic *shape* this borrows but whose TZ bug
  (ignoring its own `timezone` field, using server-local time) it does not repeat.
- Wired through `AttendanceScheduler.resolveAttendanceNotificationDeliveryJob`
  alongside the other `ATTENDANCE_NOTIFICATION_*` knobs (batch size, lease,
  max attempts) for consistency.
- `.env.example` documents both new vars next to the existing
  `ATTENDANCE_NOTIFICATION_DEEP_LINK_ENABLED` block.

## Why

Attendance DingTalk (and email) outbox notifications currently have no
time-of-day gate — producers insert rows with `next_attempt_at` defaulting to
`now()`, and the worker claims/sends any due row 24/7 whenever the scheduler
job is enabled. This adds an opt-in nighttime quiet window.

## Scope decision (flag for review)

**v1 is deployment-wide env config**, consistent with every other
`ATTENDANCE_NOTIFICATION_*` knob (batch size, lease ms, max attempts, default
channel) — the worker is org-agnostic and claims rows across all orgs in one
query, so per-org gating would need a per-row org lookup at claim time (a
bigger surface change). It also defers **all channels**, including
`email_smtp` — a single worker-level gate, not a per-channel one.

**Per-org and per-channel quiet hours are named follow-ups, not v1.**

## Test evidence

New pure-unit suite `packages/core-backend/tests/unit/attendance-notification-quiet-hours.test.ts`
(13 tests, all green) — the worker already injects `query` and `now()`, no
real-DB wiring needed:

- Inside a non-wrapping window → zero-result shape, `query` never called.
- Outside the window → claims normally, real claim SQL (`FOR UPDATE SKIP
  LOCKED`) actually issued.
- Cross-midnight `22:00-08:00`: `23:30` and `07:30` both defer, `12:00` sends.
- Deferral is logged once per window entry, not on every tick.
- `resolveAttendanceNotificationQuietHours`: unset/empty → off no warn;
  malformed shape / out-of-range HH:MM / unrecognized IANA TZ → off + exactly
  one warn each; valid window with TZ unset → defaults to `Asia/Shanghai`;
  valid window + explicit TZ → both honored.
- `isAttendanceNotificationQuietHours` timezone correctness: same UTC instant
  evaluated against `Asia/Shanghai` vs `UTC` configs disagrees as expected
  (proves the formatter actually consults `config.timeZone`).
- Worker-level env resolution at construction: malformed/unset env → gate off,
  worker claims normally.

Ran full existing `attendance-scheduler.test.ts` +
`attendance-email-delivery-channel.test.ts` alongside — 53/53 pass, no
regressions. `tsc --noEmit` clean.

### Mutation table

| Mutation | Result |
|---|---|
| Delete the pre-claim quiet-hours guard in `runBatch()` | 3 tests RED (zero-SQL assertion, cross-midnight defer, once-per-entry log) |
| Break the wrap logic (drop the `start<=end` branch, always use the non-wrapping comparison) | 2 tests RED (cross-midnight `23:30`/`07:30` no longer defer) |

Both mutations applied against the committed implementation, observed RED,
then reverted via `git checkout --` before pushing (implementation in this
PR is unmodified from what was tested green).

## Honest gaps / not proven

- No real-DB integration test asserting a pending row's `attempt_count`
  literally stays `0` across a quiet-window `runBatch()` against real
  Postgres — the unit-level "query never called" assertion is the
  intentionally chosen proof surface per the brief (this needs no
  `plugin-tests.yml` wiring as a result).
- Rows claimed seconds before the window starts still send — lease is
  already taken by claim time; inherent to a pre-claim gate, not a bug.
- Morning drain after a window ends is bounded by `batchSize` (default 50)
  per scheduler tick — unchanged from existing behavior, fine at current
  volumes.

## Ops note

Nothing changes for existing deployments unless
`ATTENDANCE_NOTIFICATION_QUIET_HOURS` is explicitly set — default behavior
(env unset) is byte-identical to before this PR.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -66,6 +66,10 @@ JWT_EXPIRES_IN=2h
 #   notifications simply wait (no event loss, no retry/lease interaction) until the
 #   window ends. Unset/empty = off (unchanged behavior). Malformed value = off + one
 #   warn log, never a crash. Supports cross-midnight windows, e.g. "22:00-08:00".
+#   Both endpoints are INCLUSIVE of the whole minute (e.g. "22:00-08:00" still defers at
+#   08:00:59). start === end (e.g. "22:00-22:00") is INVALID, not a zero-length or
+#   full-day window — it is rejected the same as a malformed value (off + one warn); use
+#   two distinct HH:MM values.
 #   v1 is deployment-wide (all orgs) AND channel-agnostic: the gate sits BEFORE the worker
 #   claims any outbox row, so it applies to EVERY channel this worker dispatches — DingTalk,
 #   WeCom, and email_smtp alike — with no per-channel opt-out. Per-org quiet hours is a
--- a/.env.example
+++ b/.env.example
@@ -60,6 +60,20 @@ JWT_EXPIRES_IN=2h
 # WECOM_AGENT_ID=1000002
 # Optional override (defaults to https://qyapi.weixin.qq.com):
 # WECOM_BASE_URL=https://qyapi.weixin.qq.com
+#
+# ATTENDANCE_NOTIFICATION_QUIET_HOURS (R8) — "HH:MM-HH:MM" (24h clock, local to the TZ
+#   below); gates the C5-2 delivery worker BEFORE it claims outbox rows, so due
+#   notifications simply wait (no event loss, no retry/lease interaction) until the
+#   window ends. Unset/empty = off (unchanged behavior). Malformed value = off + one
+#   warn log, never a crash. Supports cross-midnight windows, e.g. "22:00-08:00".
+#   v1 is deployment-wide (all orgs) AND channel-agnostic: the gate sits BEFORE the worker
+#   claims any outbox row, so it applies to EVERY channel this worker dispatches — DingTalk,
+#   WeCom, and email_smtp alike — with no per-channel opt-out. Per-org quiet hours is a
+#   follow-up, not v1.
+# ATTENDANCE_NOTIFICATION_QUIET_HOURS=22:00-08:00
+# ATTENDANCE_NOTIFICATION_QUIET_HOURS_TZ — IANA zone for the window above. Defaults to
+#   Asia/Shanghai when ATTENDANCE_NOTIFICATION_QUIET_HOURS is set but this is not.
+# ATTENDANCE_NOTIFICATION_QUIET_HOURS_TZ=Asia/Shanghai
 
 # =============================================================================
 # REDIS CACHE (Optional)
--- a/packages/core-backend/src/services/AttendanceNotificationDeliveryWorker.ts
+++ b/packages/core-backend/src/services/AttendanceNotificationDeliveryWorker.ts
@@ -243,6 +243,13 @@ const DEFAULT_QUIET_HOURS_TZ = 'Asia/Shanghai'
  * Malformed input (bad "HH:MM-HH:MM" shape, or a TZ Intl does not recognize) never throws: it logs a
  * single warn and disables the gate (off), exactly like unset — a typo in ops config must not crash the
  * scheduler tick or start silently dropping notifications outside a window nobody intended.
+ *
+ * Channel-agnostic by construction: this config is consumed only by AttendanceNotificationDeliveryWorker
+ * .runBatch()'s pre-claim gate (see the R8 comment there), which runs BEFORE claimDueDeliveries() and
+ * therefore BEFORE any per-row channel dispatch. It applies to every channel this worker dispatches —
+ * DingTalk, WeCom, and email alike — with no per-channel opt-out; there is no separate claim/send path
+ * for any registered channel that could bypass it (verified for WeCom's S4 adapter, which routes through
+ * the same channelsByName / deliver() call site as DingTalk and email).
  */
 export function resolveAttendanceNotificationQuietHours(
   env: NodeJS.ProcessEnv = process.env,
@@ -257,6 +264,19 @@ export function resolveAttendanceNotificationQuietHours(
     )
     return null
   }
+  const start = `${match[1]}:${match[2]}`
+  const end = `${match[3]}:${match[4]}`
+  if (start === end) {
+    // start === end (e.g. "22:00-22:00") is treated as INVALID config, same tier as a malformed shape —
+    // NOT as a valid zero-length or full-day window. A zero-length window is almost certainly a config
+    // typo, not an intentional one-minute quiet period; and "fixing" it to mean full-day-quiet instead
+    // would let a typo silently stop ALL attendance notifications for 24h with no visible signal beyond
+    // this warn. Least-surprise is to refuse the config outright, exactly like a malformed shape.
+    logger.warn(
+      `Ignoring ATTENDANCE_NOTIFICATION_QUIET_HOURS="${raw}" (start equals end — a zero-length window is not a valid config; use two distinct HH:MM values); quiet hours disabled`,
+    )
+    return null
+  }
   const timeZone = (env.ATTENDANCE_NOTIFICATION_QUIET_HOURS_TZ ?? '').trim() || DEFAULT_QUIET_HOURS_TZ
   try {
     // Throws RangeError for a zone Intl does not recognize; this is the validation, the format result
@@ -266,7 +286,7 @@ export function resolveAttendanceNotificationQuietHours(
     logger.warn(`Ignoring invalid ATTENDANCE_NOTIFICATION_QUIET_HOURS_TZ="${timeZone}"; quiet hours disabled`)
     return null
   }
-  return { start: `${match[1]}:${match[2]}`, end: `${match[3]}:${match[4]}`, timeZone }
+  return { start, end, timeZone }
 }
 
 /**
@@ -275,10 +295,23 @@ export function resolveAttendanceNotificationQuietHours(
  * computed via Intl.DateTimeFormat in the CONFIGURED timezone — NotificationService uses
  * `toTimeString()` (server-local time) and silently ignores its own `timezone` field; that bug is not
  * repeated here.
+ *
+ * Both endpoints are INCLUSIVE: `current <= end` means the window covers the entire end minute (e.g.
+ * "22:00-08:00" still defers at 08:00:59), not up-to-but-excluding it. Intentional, documented here and
+ * in .env.example — not changed by this pass.
+ *
+ * Fail-OPEN on a broken Intl result: if formatToParts unexpectedly omits the hour/minute parts (should
+ * not happen given the options passed above, but a broken runtime ICU data set is not impossible), this
+ * returns `false` (outside the window) rather than defaulting to "00:00". Defaulting to midnight would
+ * silently land inside any cross-midnight window and could make notifications stop flowing permanently
+ * and invisibly on every tick. A late/deferred notification is recoverable (the next in-window tick
+ * sends it); a silent permanent quiet window is not — so an unparseable local time must not block
+ * delivery.
  */
 export function isAttendanceNotificationQuietHours(
   now: Date,
   config: AttendanceNotificationQuietHoursConfig,
+  logger: Pick<Logger, 'warn'> = new Logger('AttendanceNotificationDeliveryWorker'),
 ): boolean {
   const parts = new Intl.DateTimeFormat('en-US', {
     timeZone: config.timeZone,
@@ -286,8 +319,14 @@ export function isAttendanceNotificationQuietHours(
     minute: '2-digit',
     hourCycle: 'h23',
   }).formatToParts(now)
-  const hour = parts.find((p) => p.type === 'hour')?.value ?? '00'
-  const minute = parts.find((p) => p.type === 'minute')?.value ?? '00'
+  const hour = parts.find((p) => p.type === 'hour')?.value
+  const minute = parts.find((p) => p.type === 'minute')?.value
+  if (hour === undefined || minute === undefined) {
+    logger.warn(
+      `Could not determine local time for quiet-hours check (timeZone="${config.timeZone}"); treating as OUTSIDE the quiet window so notifications are not blocked`,
+    )
+    return false
+  }
   const current = `${hour}:${minute}`
   if (config.start <= config.end) {
     return current >= config.start && current <= config.end
@@ -824,7 +863,14 @@ export class AttendanceNotificationDeliveryWorker {
       // runBatch() call drains them oldest-first via claimDueDeliveries' ORDER BY. This must NOT be
       // expressed as a channel-level retryable failure — that would burn attempt_count toward
       // maxAttempts and could dead-letter a row via computeDeliveryBackoffMs's up-to-6h backoff.
-      if (this.quietHours && isAttendanceNotificationQuietHours(this.now(), this.quietHours)) {
+      //
+      // This gate fences the WHOLE batch, not any one channel: claimDueDeliveries() is the ONLY place
+      // outbox rows get claimed, and deliver() (the only caller of channel.send()) only runs over rows
+      // that claimDueDeliveries() returned. Every registered AttendanceDeliveryChannel — DingTalk, WeCom
+      // (S4), and email — is dispatched exclusively through that one claim -> deliver loop below, so
+      // quiet hours applies to every channel this worker dispatches with no per-channel opt-out. Ratified
+      // product decision (owner, 2026-07-10): "don't disturb at night" is channel-independent.
+      if (this.quietHours && isAttendanceNotificationQuietHours(this.now(), this.quietHours, this.logger)) {
         if (!this.quietHoursLogged) {
           this.logger.info(
             `Attendance notification dispatch deferred: quiet hours ${this.quietHours.start}-${this.quietHours.end} (${this.quietHours.timeZone})`,
--- a/packages/core-backend/src/services/AttendanceNotificationDeliveryWorker.ts
+++ b/packages/core-backend/src/services/AttendanceNotificationDeliveryWorker.ts
@@ -93,6 +93,13 @@ export interface AttendanceNotificationDeliveryWorkerOptions {
   workerId?: string
   now?: () => Date
   logger?: Logger
+  /**
+   * R8 quiet hours gate. `undefined` (the default when the worker is constructed by
+   * resolveAttendanceNotificationDeliveryJob) resolves from ATTENDANCE_NOTIFICATION_QUIET_HOURS /
+   * ATTENDANCE_NOTIFICATION_QUIET_HOURS_TZ at construction time; pass `null` explicitly to force the
+   * gate off (e.g. tests), or a concrete config to bypass env parsing entirely (also tests).
+   */
+  quietHours?: AttendanceNotificationQuietHoursConfig | null
 }
 
 export interface DingTalkAttendanceDeliveryChannelOptions {
@@ -212,6 +219,80 @@ export function computeDeliveryBackoffMs(attemptCount: number): number {
   if (attemptCount === 3) return 15 * 60_000
   if (attemptCount === 4) return 60 * 60_000
   return 6 * 60 * 60_000
+}
+
+export interface AttendanceNotificationQuietHoursConfig {
+  /** 24h "HH:MM", local to `timeZone`. */
+  start: string
+  /** 24h "HH:MM", local to `timeZone`; may be < start (cross-midnight window, e.g. 22:00-08:00). */
+  end: string
+  /** IANA zone name, e.g. "Asia/Shanghai". */
+  timeZone: string
+}
+
+const QUIET_HOURS_ENV_PATTERN = /^([01]\d|2[0-3]):([0-5]\d)-([01]\d|2[0-3]):([0-5]\d)$/
+const DEFAULT_QUIET_HOURS_TZ = 'Asia/Shanghai'
+
+/**
+ * R8 quiet-hours window (design decision: deployment-wide v1, consistent with every existing
+ * ATTENDANCE_NOTIFICATION_* knob — per-org config is a named follow-up, not v1).
+ *
+ * ATTENDANCE_NOTIFICATION_QUIET_HOURS="HH:MM-HH:MM" — unset/empty = off (unchanged behavior).
+ * ATTENDANCE_NOTIFICATION_QUIET_HOURS_TZ — IANA zone; defaults to Asia/Shanghai.
+ *
+ * Malformed input (bad "HH:MM-HH:MM" shape, or a TZ Intl does not recognize) never throws: it logs a
+ * single warn and disables the gate (off), exactly like unset — a typo in ops config must not crash the
+ * scheduler tick or start silently dropping notifications outside a window nobody intended.
+ */
+export function resolveAttendanceNotificationQuietHours(
+  env: NodeJS.ProcessEnv = process.env,
+  logger: Pick<Logger, 'warn'> = new Logger('AttendanceNotificationDeliveryWorker'),
+): AttendanceNotificationQuietHoursConfig | null {
+  const raw = (env.ATTENDANCE_NOTIFICATION_QUIET_HOURS ?? '').trim()
+  if (!raw) return null
+  const match = QUIET_HOURS_ENV_PATTERN.exec(raw)
+  if (!match) {
+    logger.warn(
+      `Ignoring malformed ATTENDANCE_NOTIFICATION_QUIET_HOURS="${raw}" (expected "HH:MM-HH:MM", 24h clock); quiet hours disabled`,
+    )
+    return null
+  }
+  const timeZone = (env.ATTENDANCE_NOTIFICATION_QUIET_HOURS_TZ ?? '').trim() || DEFAULT_QUIET_HOURS_TZ
+  try {
+    // Throws RangeError for a zone Intl does not recognize; this is the validation, the format result
+    // itself is discarded.
+    new Intl.DateTimeFormat('en-US', { timeZone }).format(new Date())
+  } catch {
+    logger.warn(`Ignoring invalid ATTENDANCE_NOTIFICATION_QUIET_HOURS_TZ="${timeZone}"; quiet hours disabled`)
+    return null
+  }
+  return { start: `${match[1]}:${match[2]}`, end: `${match[3]}:${match[4]}`, timeZone }
+}
+
+/**
+ * Same cross-midnight wrap shape as NotificationService.isInQuietHours (start<=end → contiguous
+ * window; start>end → wraps past midnight, "inside" means >= start OR <= end), but the local HH:MM is
+ * computed via Intl.DateTimeFormat in the CONFIGURED timezone — NotificationService uses
+ * `toTimeString()` (server-local time) and silently ignores its own `timezone` field; that bug is not
+ * repeated here.
+ */
+export function isAttendanceNotificationQuietHours(
+  now: Date,
+  config: AttendanceNotificationQuietHoursConfig,
+): boolean {
+  const parts = new Intl.DateTimeFormat('en-US', {
+    timeZone: config.timeZone,
+    hour: '2-digit',
+    minute: '2-digit',
+    hourCycle: 'h23',
+  }).formatToParts(now)
+  const hour = parts.find((p) => p.type === 'hour')?.value ?? '00'
+  const minute = parts.find((p) => p.type === 'minute')?.value ?? '00'
+  const current = `${hour}:${minute}`
+  if (config.start <= config.end) {
+    return current >= config.start && current <= config.end
+  }
+  return current >= config.start || current <= config.end
 }
 
 export class DeterministicFakeAttendanceDeliveryChannel implements AttendanceDeliveryChannel {
@@ -664,6 +745,8 @@ export class AttendanceNotificationDeliveryWorker {
   private readonly workerId: string
   private readonly now: () => Date
   private readonly logger: Logger
+  private readonly quietHours: AttendanceNotificationQuietHoursConfig | null
+  private quietHoursLogged = false
   private running = false
 
   constructor(options: AttendanceNotificationDeliveryWorkerOptions = {}) {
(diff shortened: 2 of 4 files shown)


zensgit:
# PR #4011 独立对抗审阅 — R8 attendance 通知 quiet-hours 窗口

- **PR**: zensgit/metasheet2#4011 `feat(attendance): R8 — quiet-hours window for DingTalk notification delivery`
- **Head 钉死**: `832216bdd5be3bf71215a04d6ec6095709f36e65`（`refs/pull/4011/head` fetch 后逐一核对，审阅全程针对该 SHA）
- **Base**: merge-base == 当前 `origin/main` tip (`a894d8d0a`)，无落后
- **审阅日期**: 2026-07-09 · 审阅者: Claude（独立对抗 gate，refute-first）
- **方法**: 一次性 worktree `/tmp/ms2-rev-pr4011`（detach 到 PR head）→ `pnpm install --frozen-lockfile` → 实跑测试 → 自主 6 组 mutation → CI 两点接线核查 → 用后即弃

## 判定：APPROVE（附 1 条 P3 hardening 建议）

无 P1/P2。作者报告的每一条声明都经独立复验属实，且我自己追加的 3 组 mutation 里 2 组 RED（守卫承重）、1 组 GREEN（仅日志去重 reset 未测 → P3）。范围纪律干净。

---

## 1. 实跑测试记录（全部在我自己的 worktree、PR head 上）

| 项目 | 结果 |
|---|---|
| `tests/unit/attendance-notification-quiet-hours.test.ts`（新增） | **13/13 PASS** |
| 邻居回归 `attendance-scheduler.test.ts` + `attendance-email-delivery-channel.test.ts` | **40/40 PASS**（作者报的 53/53 = 40 + 新增 13 之和，口径松但无实质出入） |
| 全量 `tests/unit/` 目录（对齐 required `test` job 的 unit 面） | **342 文件 / 4558 tests 全 PASS** |
| `tsc --noEmit -p tsconfig.json` | **clean（exit 0）** |

## 2. Mutation 复验（每组：改 → 跑 → 记录 → `git checkout --` 还原；终态 tree clean @ head）

| # | Mutation | 结果 | 结论 |
|---|---|---|---|
| A | 删除 `runBatch()` 里 pre-claim quiet-hours 守卫块 | **RED 3**（zero-SQL 断言 / 跨午夜 defer / once-per-entry log） | 与作者报告一致，守卫承重 |
| B | 砍掉 `start<=end` 分支，恒用非跨午夜比较 | **RED 2**（23:30 / 07:30 不再 defer） | 与作者报告一致 |
| C（自主） | `isAttendanceNotificationQuietHours` 里硬编码 `timeZone:'UTC'` 无视 config | **RED 1** | TZ 消费被真实钉住，且双 config 设计对 runner 本地时区鲁棒 |
| D（自主） | 放松 regex 为 `(\d\d):(\d\d)` 使 `25:00` 通过 | **RED 1** | 越界负例（房规 enum/格式严格性）真实承重 |
| E（自主） | 删除窗口外 tick 的 `this.quietHoursLogged = false` reset（:558） | **GREEN 13/13** | **未测守卫** → 见 Finding 1（仅日志，P3） |
| F（自主） | 删除 IANA TZ 的 `Intl` try/catch 校验 | **RED 1** | 该校验同时是防 runBatch 内 RangeError 崩溃的守卫，已被测试钉住 |

## 3. CI 两点接线核查（本仓雷区）

- 新测试位于 `tests/unit/`，默认 `vitest.config.ts` include、**不在 exclude 列表** → 跑在 plugin-tests.yml 的 **`test` matrix job（18.x/20.x，required check）**的 `pnpm --filter @metasheet/core-backend test` 步骤里。**不是**「绿 CI ≠ 跑过」的 plugin-integration-core 死角。
- PR 当前 checks：`test (18.x)` 已 PASS（4m33s，含新 suite + 同 job 内 attendance 真 DB integration 步骤跑了改动后的 worker 构造函数），`test (20.x)` 审阅时 pending，其余全绿。
- 既有 `attendance-notification-deliveries.test.ts`（10 处 `new AttendanceNotificationDeliveryWorker` 不传 `quietHours`）在 required job 的真 DB 步骤照跑——构造函数新默认（undefined → 读 env）在 CI env 未设该变量时行为 byte-identical，已由 18.x 绿灯实证。
- 无新依赖、无 lockfile 变更、无 `package.json` test-chain 改动 → 无 UNION 掉行风险。

## 4. 范围纪律 & 房规扫雷

- **4 个文件全部在 lane 内**：worker（+102）、scheduler 接线（+6）、`.env.example`（+11）、新测试（+212）。纯新增 0 删除；无路由/index/runtime 偷 wire。
- **PR body 无裸 closing keyword**（不会误关 issue）。
- **枚举/格式静默 fallback**：malformed shape、越界 HH:MM、非法 TZ 三类负例齐全且经 mutation D/F 证实承重（本仓房规达标）。
- **默认行为不变**：env 未设 → `resolveAttendanceNotificationQuietHours` 返回 null → gate off；有全量 unit 目录 4558 绿 + 18.x 真 DB 步骤绿背书。
- **无 skip-when-unreachable / 无 tautology**：断言面是 query-mock 调用次数 + 真实 claim SQL 形状（`FOR UPDATE SKIP LOCKED`），测试没有复刻窗口数学。
- **代码注释里对 `NotificationService.isInQuietHours` 前例 TZ bug 的指控经核实为真**（NotificationService.ts:807 用 `toTimeString()`（server-local），其 `timezone` 字段确实被无视）——新实现用 `Intl.DateTimeFormat(config.timeZone)`，没有复刻该 bug。我在本机 Node 探针确认 `hourCycle:'h23'` 午夜输出 `00`（无 h24 的 `24:00` 怪癖）。
- **配置面决策（deployment-wide、含 email_smtp 的全 channel 单 gate）**：PR body 有专节「Scope decision (flag for review)」明确陈述并列出 per-org/per-channel 为 named follow-up——满足 brief 要求的「flag 给 owner 点头」，非偷渡。
- **无独立 design-lock 文档管辖本项**（docs 内无 R8/quiet-hours lock）；governing spec = backlog brief 本身，逐项对照（pre-claim gate / 无 event loss / malformed 不 crash / 跨午夜 / TZ / 单元证明面）全 PASS。
- **无事件丢失路径复核**：gate 在 `claimDueDeliveries()` 之前、零 SQL（unit 实证）；due 行不动 `next_attempt_at`/`attempt_count`/lease。窗口内过期的 `sending` lease 行同样顺延到窗口后由既有 reclaim 分支收回（`status='sending' AND claim_expires_at<=now`），无泄漏。

## 5. Findings

### P3 — 窗口重入日志去重的 reset 是未测守卫
`packages/core-backend/src/services/AttendanceNotificationDeliveryWorker.ts:558`（`this.quietHoursLogged = false`）。Mutation E（删除该行）13/13 全绿。失败场景：未来重构误删 reset → 第二次及以后进入静默窗口不再打「dispatch deferred」info 日志（首窗仍打），投递行为完全不受影响——纯可观测性，非 correctness/security-shaped，故 P3 而非房规 P2。**修法（顺手 hardening）**：加一个单实例状态迁移测试——可变 `now`，tick 窗口内→窗口外→再窗口内，断言 info 恰好 2 次且中间 tick 真实发起 claim。该测试同时把「同一实例出窗后恢复 claim」这条迁移路径钉住（现有测试只用不同实例分别测内/外）。

### NIT — `start === end` 语义未定义
`AttendanceNotificationDeliveryWorker.ts:271-272`：如 `22:00-22:00` 走非跨午夜分支 → 恰好 1 分钟的窗口；有的运维可能预期 24h。`.env.example:52-58` 未说明。可在注释里补一句。

### NIT — 窗口端点双闭区间未文档化
`AttendanceNotificationDeliveryWorker.ts:272,274`：`current <= end` 使 `22:00-08:00` 在 08:00:00–08:00:59 仍 defer（end 那一整分钟含在窗口内）。无害，但含入性未在 env 文档里写明。

### NIT — Intl parts 缺失时 fail-defer
`AttendanceNotificationDeliveryWorker.ts:268-269` 的 `?? '00'`：若 formatToParts 异常缺 hour/minute（现实中给定这些 options 不会发生），跨午夜窗口会把 00:00 视为在窗内 → 静默 defer。方向是 fail-safe（宁可晚发不误发），仅备注。

## 6. 结论

作者报告诚实：测试数、mutation 结果、gaps 自述（无真 DB attempt_count 字面断言 / 窗口前已 claim 行仍发 / 晨间 drain 受 batchSize 限）逐条属实且定级合理。实现小而对，负例覆盖是本仓少见的齐全。**APPROVE**；P3 迁移测试属 nice-to-have hardening，不阻落地。owner 对 deployment-wide 配置面的点头以 PR body 专节为准。

zensgit:
## Gate-finding closure — 4 findings closed (P3 + 3×NIT)

Commit: `67183d33f` (fast-forward on top of `832216bdd`, pushed to `claude/dt-attendance-quiet-hours`).

Closes all four findings from the independent adversarial review (`/tmp/pr4011-review-claude-20260709.md`):

### 1. P3 — window-exit `quietHoursLogged` reset was an untested guard
Added a single-instance state-migration test: enter window (defer + 1st log) → exit window (claims normally, real SQL issued) → re-enter window (asserts a **2nd** deferral log). This is a genuinely new scenario — the existing "logs once per entry" test never leaves the window, so it couldn't have caught this.

**Mutation**: delete `this.quietHoursLogged = false` (the reset on the non-deferral path) → **RED** — `info` call count stays at 1 instead of reaching 2, because the flag latches `true` forever after the first deferral.

### 2. NIT — `start === end` semantics
Decided: treat as **INVALID config** (off + one warn), same tier as a malformed shape — not a valid one-minute window, and explicitly *not* reinterpreted as full-day-quiet. A zero-length window is almost certainly a typo; silently mapping it to "block everything for 24h" would be far more dangerous than refusing the config. Implemented in `resolveAttendanceNotificationQuietHours` (checked after shape-match, before TZ validation) and documented in `.env.example`.

**Mutation**: remove the `start === end` guard → **RED** — both new tests (`"22:00-22:00"` and `"00:00-00:00"`) fail because the config now resolves instead of being rejected.

### 3. NIT — end-minute inclusivity
No behavior change (as directed). Documented in the `isAttendanceNotificationQuietHours` JSDoc and in `.env.example`: both endpoints are inclusive of the whole end minute (e.g. `"22:00-08:00"` still defers through `08:00:59`). Doc-only — no mutation applies here.

### 4. NIT — `?? '00'` Intl fallback failed toward midnight
Flipped the failure direction: if `formatToParts` doesn't return `hour`/`minute` parts, `isAttendanceNotificationQuietHours` now logs a warn and returns `false` (**outside** the window — notifications keep flowing) instead of defaulting to `"00:00"` (which would read as **inside** any cross-midnight window and could silently and permanently stop all attendance notifications on a broken TZ database). `runBatch()` now passes `this.logger` through so the warn is observable at the worker level.

**Mutation**: revert to the old `?? '00'` fallback → **RED** on both new tests — the pure-function test flips from `false` to `true` (proving the direction reversal), and the `runBatch` test shows `query` is never called (proving the old behavior would silently defer forever instead of claiming).

### Mutation table

| # | Mutation | Result | Finding |
|---|---|---|---|
| 1 | Delete `this.quietHoursLogged = false` reset (non-deferral path) | **RED** — 1 test (`info` stuck at 1 instead of 2) | P3 |
| 2 | Remove `start === end` invalid-config guard | **RED** — 2 tests (both now resolve instead of `null`) | NIT (start===end) |
| 3 | — (doc-only, no code path to mutate) | n/a | NIT (inclusivity) |
| 4 | Revert `?? '00'` fallback (undo fail-open) | **RED** — 2 tests (direction flips `false→true`; `runBatch` never claims) | NIT (Intl fallback) |

Each mutation was applied against the **committed** implementation (`67183d33f`), run, recorded, then reverted via `git checkout --` before the next one — tree stayed clean at HEAD throughout.

### Test evidence
- `attendance-notification-quiet-hours.test.ts`: **18/18 pass** (13 original + 5 new: 1 for Finding 1, 2 for Finding 2, 2 for Finding 4).
- Neighbor regression (`attendance-scheduler.test.ts` + `attendance-email-delivery-channel.test.ts`): **40/40 pass**.
- Full `tests/unit/`: **342 files / 4563 tests pass** (4558 baseline + 5 new), no regressions.
- `tsc --noEmit -p tsconfig.json`: clean.

### One decision to flag for the reviewer
Finding 4's warn is **not** deduped across ticks the way the existing `quietHoursLogged` deferral log is — it fires once per `runBatch()` call in which Intl parts are missing, not once-ever. Given the scheduler ticks on a short interval, a genuinely broken ICU/TZ dataset would warn every tick. I judged this acceptable (the condition should not occur given the fixed `formatToParts` options passed, per the original NIT text) rather than adding a second dedup-state field, but flagging it explicitly rather than leaving it silent.

Files touched (unchanged from the original PR's scope — worker, its tests, `.env.example`):
- `packages/core-backend/src/services/AttendanceNotificationDeliveryWorker.ts`
- `packages/core-backend/tests/unit/attendance-notification-quiet-hours.test.ts`
- `.env.example`

Not arming auto-merge — per standing discipline, landing stays gated/serialized by the owner.

zensgit:
## DIRTY-state resolution: rebased onto main (WeCom S4 channel landed) + re-verified the combination

Main landed the WeCom (企业微信) delivery channel (#4028, `fd242899c`) after this PR's branch point, and it touches the **same file** this PR's quiet-hours gate lives in (`AttendanceNotificationDeliveryWorker.ts`). This comment documents the rebase, the combination analysis, and the re-verification.

### Combination analysis — does WeCom bypass the gate?

**No bypass path.** Verified by reading the S4 diff and grepping the whole `packages/core-backend/src` tree on `origin/main`:

- `WeComAttendanceDeliveryChannel` is registered into `channelsByName` by `createAttendanceDeliveryChannelsFromEnv()` exactly like `DingTalkAttendanceDeliveryChannel` and `EmailAttendanceDeliveryChannel` — one more entry in the same `Map`.
- The **only** caller of `channel.send()` for any channel is `AttendanceNotificationDeliveryWorker.deliver(row)`, and `deliver()` is only ever called from the loop inside `runBatch()`, over rows returned by `claimDueDeliveries()`.
- `claimDueDeliveries()` is the **only** place outbox rows get claimed — there is no second claim/send path anywhere in the tree (`git grep` for `sendWeComTextMessage`, `sendWeComTextCardMessage`, `WeComAttendanceDeliveryChannel`, `fetchWeComAppAccessToken` across `packages/core-backend/src` on `origin/main` turns up only the WeCom transport client and its one consumer, the channel class itself).
- `AttendanceScheduler.ts`'s `resolveAttendanceNotificationDeliveryJob()` is the only scheduler entry point that constructs the worker; there's no separate WeCom-specific scheduler job.

Since the R8 quiet-hours gate sits **before** `claimDueDeliveries()` inside `runBatch()`, and WeCom rows can only ever be claimed and dispatched through that same call, the pre-existing gate already fences WeCom (and would fence any future channel added the same way) with **zero code changes required** to the gate itself.

### Ratified decision, now documented in code

Per the owner's product decision (main-loop ratified): **quiet hours applies to every channel this worker dispatches — DingTalk, WeCom, and email alike — with no per-channel opt-out.** Updated:
- The `runBatch()` gate's inline comment — spells out why this fences the whole batch, not one channel, and names WeCom explicitly.
- The `isAttendanceNotificationQuietHours`/`resolveAttendanceNotificationQuietHours` doc-comment block — added a paragraph stating the channel-agnostic guarantee and that it was verified for WeCom's S4 adapter.
- `.env.example` — the `ATTENDANCE_NOTIFICATION_QUIET_HOURS` block now says explicitly: "applies to EVERY channel this worker dispatches — DingTalk, WeCom, and email alike — with no per-channel opt-out."

### Rebase

`git rebase origin/main` (onto `accfb01e4`, main's tip at time of rebase). Two conflicts, both in the two quiet-hours commits' hunks vs. the WeCom commit's hunks in `.env.example` (comment-block ordering only — resolved by keeping both blocks, WeCom block first, quiet-hours block after, merged with the channel-agnostic wording above). `AttendanceNotificationDeliveryWorker.ts` **auto-merged cleanly** both times — WeCom's additions (imports, channel class, error classification) and quiet-hours' additions (config type, resolver, gate in `runBatch()`) don't touch overlapping lines. No `AttendanceScheduler.ts` conflict.

New head: **`671488299c87861b7ef36dd524baabd437ae8822`**
(`14e01d5ed` R8 quiet-hours feat → `04fa5173d` R8 hardening fix → `671488299` new combination-test commit, replayed on top of main's `accfb01e4`, which already includes WeCom's `fd242899c`.)

### Re-verification

**Full core-backend unit suite** (collateral-damage check, since this file is shared broadly): `CI=true npx vitest run tests/unit` → **346 files / 4663 tests, all passing.**

**Targeted suites:**
| Suite | Tests | Result |
|---|---|---|
| `attendance-notification-quiet-hours.test.ts` (18 prior + 3 new combination) | 21 | ✅ all pass |
| `attendance-wecom-delivery-channel.test.ts` | 51 | ✅ all pass |
| `wecom-client.test.ts` | — (included below) | ✅ all pass |
| `attendance-email-delivery-channel.test.ts` | 18 | ✅ all pass |
| `attendance-scheduler.test.ts` | 19 | ✅ all pass |
| **Combined (wecom-channel + wecom-client + email + scheduler)** | **105** | ✅ all pass |

**New combination tests** (extends `attendance-notification-quiet-hours.test.ts`, real `DingTalkAttendanceDeliveryChannel` + `WeComAttendanceDeliveryChannel` + `EmailAttendanceDeliveryChannel` instances with DI-injected send seams, driven through one real `AttendanceNotificationDeliveryWorker`):
1. **Inside window** → `runBatch()` returns the zero-result shape, claim SQL never issued, **and** every channel's underlying send seam (DingTalk `sendWorkNotification`/`sendWorkNotificationActionCard`, WeCom `sendTextMessage`/`sendTextCardMessage`, email `sendMail`) has **zero calls**. This is the test that would catch a hypothetical WeCom bypass path.
2. **Outside window** → all three channels claim and dispatch, `claimed: 3, sent: 3`, one send call per channel.
3. **Quiet hours off** (`quietHours: null`) → same as outside-window, channel dispatch is completely unaffected when the gate isn't configured.

### Mutation replays (on the rebased head)

Protocol: commit real implementation first (already the head), apply mutation, run `attendance-notification-quiet-hours.test.ts`, record red/green, `git checkout --` to restore, re-verify clean.

| Mutation | Result |
|---|---|
| Delete the pre-claim guard (`if (this.quietHours && isAttendanceNotificationQuietHours(...))` short-circuited to `if (false && ...)`) | **RED** — 6/21 tests fail, including the new "inside window: ZERO claims AND ZERO sends on every channel" combination test (claimed jumped 0→3, sent 0→3) |
| Break the cross-midnight wrap (`current >= config.start || current <= config.end` → `&&`) | **RED** — 4/21 tests fail, including the same new combination test (the 23:30 inside-window case incorrectly reads as outside under `&&`) |

Both restored; full targeted suite (126 tests across the 5 files above) re-confirmed green on the clean rebased head afterward.

### Not armed

Per the owner's landing order, this PR's window comes **after #4010** — not arming auto-merge. A fresh adversarial re-gate will run at this PR's actual landing window (the gate that reviewed the pre-rebase head does not cover this combination and should not be treated as still current).

zensgit:
# PR #4011 FRESH adversarial re-gate — R8 quiet-hours × full channel roster (post-WeCom-landing combination)

- **PR**: zensgit/metasheet2 #4011 — `feat(attendance): R8 — quiet-hours window for DingTalk notification delivery`
- **Head reviewed**: `13f13fefa4210d3833b7ea035bb3ff85ff16b3e4` (`claude/dt-attendance-quiet-hours`) — matches the owner-mandated SHA exactly; verified before any other step.
- **Base**: `main` @ `2331a33d7480ccad79ca4e89f688636efaeaece4` — the PR's merge-base **equals the current main tip** (branch is exactly 3 commits ahead of live main, fully rebased).
- **Reviewer**: Claude (Fable 5), independent adversarial re-gate, refute-first. Own disposable worktree `/tmp/ms2-rev-pr4011-regate` at the exact head, `pnpm install --frozen-lockfile`, all mutations fired and reverted there. Canonical checkout untouched.
- **Scope**: the COMBINATION of R8 quiet-hours with main-side WeCom channel (#4028 S4) + H1 fix (#4045) now living in the same worker file — not just the quiet-hours diff.
- **Date**: 2026-07-10

## Verdict: **APPROVE-with-hardening**

No P1. The combination claim ("pre-claim gate fences all channels; no bypass path") is **UPHELD** by independent structural verification and by live mutation testing. One P2 coverage hole found by a compound mutation the author's suite does not see (details below); it does not change shipped behavior and is a one-test fix.

---

## Probe 1 — THE BYPASS CLAIM: independently verified, UPHELD

Every consumer of the WeCom client/channel symbols on THIS head, enumerated (non-test):

| Symbol | Consumers outside `integrations/wecom/client.ts` |
|---|---|
| `sendWeComTextMessage` / `sendWeComTextCardMessage` / `fetchWeComAppAccessToken` / `invalidateWeComAppAccessTokenCache` | **only** `AttendanceNotificationDeliveryWorker.ts` |
| `WeComAttendanceDeliveryChannel` | constructed **only** in `createAttendanceDeliveryChannelsFromEnv` (worker file :386) |
| `createAttendanceDeliveryChannelsFromEnv` | **only** `AttendanceScheduler.ts:406` |
| `AttendanceNotificationDeliveryWorker` (construction) | **only** `AttendanceScheduler.ts:401` (`resolveAttendanceNotificationDeliveryJob`, with `quietHours` wired :405) |
| `worker.runBatch` | **only** `AttendanceScheduler.ts:410` (the scheduled job) |
| `attendance_notification_deliveries` writers besides the worker | `UnscheduledReminderService.ts:210` and `CompTimeExpiryReminderService.ts:160` — **INSERT-only producers** (plus a SELECT dedupe at UnscheduledReminderService.ts:154) |

No WeCom test-send endpoint exists (no route/API file references WeCom at all). Direct DingTalk senders outside the worker: `testDingTalkWorkNotificationAgentId` (work-notification-settings.ts:322 — admin test-send, explicitly allowed) and `multitable/automation-executor.ts` (:1829/:2745/:3092 — multitable automation actions, a different product surface, not attendance outbox notifications). Email outside the worker: `NotificationService` serves SLOService / ChangeManagementService / multitable `send_email` automation — **nothing attendance-shaped**; no attendance digest/summary sender exists outside this worker. No second worker, no second scheduler job, no "flush now" endpoint.

Main-side landings since the original verdict that touch this territory (`#4028` S4 WeCom adapter, `#4045` H1 org-integration-inactive fix, `#3920` skip semantics) are all **inside the worker file** and all present intact on this head.

## Probe 2 — Combination tests honesty: HONEST, and proven load-bearing

The 3 roster tests (`attendance-notification-quiet-hours.test.ts:333-503`) construct **real** `DingTalkAttendanceDeliveryChannel` / `WeComAttendanceDeliveryChannel` / `EmailAttendanceDeliveryChannel` instances (`makeRoster()`, :357-394) with spies injected at the **transport seams** (`sendWorkNotification`, `sendTextMessage`, `sendMail`) — not at `worker.deliver`. Not tautological in the trap sense: `makeWorkerQuery` (:412-418) is armed to RETURN 3 claimable rows (one per channel) if the claim SQL ever fires, so removing or weakening the gate flips the test red at both the claim assert and the per-channel send-seam asserts — confirmed live by mutations A and C below.

Internal-queue hypothetical assessed: **no channel on this head has an internal queue** — all three send synchronously inside `send()`; the email channel caches only the transport object (worker file :1284-1285), not messages. A future channel with its own flush timer would evade this unit test — nothing to fix today, just the known boundary of the seam.

## Probe 3 — Suite + mutations on THIS head (all runs in the disposable worktree)

Clean baseline: `attendance-notification-quiet-hours` (21) + `attendance-wecom-delivery-channel` (40) + `attendance-email-delivery-channel` (18) + `attendance-scheduler` (24) = **103/103 PASS** (4 files, vitest, ~540ms).

| Mutation | Result |
|---|---|
| **A. Delete the pre-claim quiet-hours guard** (runBatch :873-882) | **RED — 6 failed**, including the all-channels zero-claims/zero-sends combination test |
| **B. Break cross-midnight wrap** (:334 `\|\|` → `&&`) | **RED — 4 failed** (cross-midnight defer, once-per-entry log, window-exit reset, combination test) |
| **C. (extra, reviewer's own) Weaken gate to DingTalk-only, per-row after claim** | **RED — 5 failed**, including the combination test — the all-channels semantics is genuinely pinned, not just the gate's existence |
| **D. (extra, compound) Neuter constructor env-fallback (:802-804 → `options.quietHours ?? null`) AND delete scheduler wiring (:405)** | **GREEN — 103/103 pass** ← finding P2-1 |

All mutations reverted; worktree verified byte-identical to the PR head afterwards.

CI gating: this suite is picked up by `pnpm --filter @metasheet/core-backend test` in `plugin-tests.yml`'s `test` matrix job = the **required** `test (18.x)` / `test (20.x)` checks (not excluded in `vitest.config.ts`). `test (18.x)` already green at this head.

## Probe 4 — Rebase reversion check: CLEAN

Merge-base equals live main tip `2331a33d7`; the full diff vs main is **675 insertions, 0 deletions** (verified: zero `^-` lines). No main-side WeCom/email hunk was dropped or reverted; the H1 disambiguation blocks are present in both DingTalk (:491-529) and WeCom (:717-753) recipient resolution on the head.

## Probe 5 — `.env.example`: all three semantics survive the rebase

Present at .env.example:63-77: all-channels wording ("applies to EVERY channel this worker dispatches — DingTalk, WeCom, and email_smtp alike — with no per-channel opt-out"), end-minute inclusivity ("still defers at 08:00:59"), and `start === end` rejection ("INVALID, not a zero-length or full-day window").

---

## Findings

### P2-1 — Untested guard: the env→gate-ON production wiring is invisible to the whole suite
**Evidence**: mutation D above. Neutering BOTH anchors of the env path — the constructor fallback (`AttendanceNotificationDeliveryWorker.ts:802-804`) and the scheduler wiring (`AttendanceScheduler.ts:405`) — leaves **all 103 tests green**, yet in production `ATTENDANCE_NOTIFICATION_QUIET_HOURS=22:00-08:00` would silently do nothing and 3am sends resume with no signal.
**Why the suite misses it**: every gate-ON test injects a `quietHours` config object directly; the "env resolution at construction" describe (test file :295-331) asserts only the **OFF** outcomes (malformed → claims, unset → claims). There is no test that a *valid* env value, with no `quietHours` option passed, actually defers. Each anchor alone is redundant with the other (single-deletion mutations are behavior-preserving), which is exactly why only the compound kill shows the hole — and a refactor "simplifying" both sides at once is a realistic single-PR event.
**Fix (one test)**: in the construction describe, set `process.env.ATTENDANCE_NOTIFICATION_QUIET_HOURS='22:00-08:00'` / `_TZ='UTC'`, construct the worker **without** a `quietHours` option, `now` inside the window → expect `ZERO_RESULT` and `query` never called. This kills mutation D and pins the production path end-to-end.

### P3-1 — Scheduler wiring is behaviorally redundant with the constructor default
`AttendanceScheduler.ts:405` passes `quietHours: resolveAttendanceNotificationQuietHours()`, but the constructor already resolves the same envs, at the same moment, with an equivalent default logger, when the option is `undefined` (:802-804). Harmless belt-and-braces today — but it is what degraded P2-1 to compound-only visibility. Either drop the scheduler line and let the tested (after P2-1's fix) constructor path be the single source, or keep it deliberately and say so in the option doc-comment (it currently documents the fallback, not the redundancy).

### NIT-1 — Cross-midnight × non-default TZ never composed in one test
Wrap windows are tested in UTC (:66-96); TZ correctness is tested with a non-wrapping window (:227-237). One case with `22:00-08:00` in `Asia/Shanghai` at a UTC timestamp that flips the verdict across zones would close the composition cell. Low risk (both dimensions share the same string-compare core), hence NIT.

### Observations (no grade — consistent with the ratified scope)
- Out-of-quiet-hours senders that exist by design: admin test-send (`testDingTalkWorkNotificationAgentId`), multitable automation DingTalk/email actions, `NotificationService` (SLO/change-mgmt; has its own per-subscription quiet-hours with the known server-local-TZ bug this PR pointedly does not repeat). None is a scheduled attendance notification; the ratified decision covers "channels this worker dispatches" and that is exactly what is fenced.
- PR body contains no bare issue-closing keywords (checked).

## Verdict restated

**APPROVE-with-hardening**: mergeable as-is on correctness; land the P2-1 one-test hardening (in this PR or an immediate follow-up) so the env-configured production path can never be silently disconnected.

zensgit:
Re-gate window closed: fresh adversarial re-gate on exact head 13f13fefa returned APPROVE-with-hardening (P1 none; all five probes upheld — no WeCom bypass on this head, combination tests spy at transport seams and are armed against tautology, rebase carried 0 deletions of main-side hunks). The single P2-1 (env→gate-ON production wiring invisible: compound mutation left 103/103 green while a production env var silently did nothing) is closed on this head — new test drives the gate ON purely from process.env through the real constructor default, in-window runBatch issues ZERO SQL; compound mutation (constructor fallback→null + scheduler wiring deleted) now goes RED on exactly that test. One shape fix en route (zero-result gained skipped:0 from #3920). Armed alone per single-window discipline.

PR #4012, opened 2026-07-10
feat(multitable): History Center record titles + type-aware link/person diff values (PR-B)

zensgit:
**Stacked on #4007 (PR-A)** — retarget to main after #4007 lands, land bottom-up per stacked-PR discipline.

R9 UX-parity audit findings (b)(h)(i): the History Center rendered only short raw record ids and JSON.stringify'd link/person diff values, even though everything needed is already on the wire.

### What
1. **Record titles** — `pickRecordTitle` (the TrashModal heuristic: first field in column order whose value renders non-empty; masked fields render '—' and are skipped) over the change's OWN permission-masked payload: `after` = full post-change snapshot; `before` = pre-delete snapshot, so **deleted records title from `before`**. Fallback = short id; full id kept in the `title` attr. Leak-safe by construction: no fetch, no un-masking, masked-everything payloads fall back to the id.
2. **Type-aware diff values** — `fields` prop widened (optional `type/order/property`); values route through the shared `formatFieldDisplay` (exact same renderer as MetaRecordDrawer's history diffs) fed by the grid's `linkSummaries`/`personSummaries` caches plumbed workbench→modal→list. Cache miss (historical/deleted/off-page rows) → formatFieldDisplay's own count/id fallbacks. Unknown-type fields keep the legacy plain-text path — the 6 pre-existing spec cases passing unchanged is the regression proof.

### Verification
- 6 new spec cases (titles ×3, link ×2, person ×1) in the CI-wired `multitable-history-center-inline-diff.spec.ts` (gate wiring = #4006).
- **Mutation-verified**: neuter title path → exactly 2 red; force legacy value path → exactly 3 red.
- Neighbor sweep: 7 files / 48 tests green (all 3 history-center specs + history-fe + field-scope wiring + restore-wiring + trash-fe). `vue-tsc` clean.

### Not covered / honest limitations
- All-tables mode: batches from NON-active sheets still have no field metadata threaded in → id-fallback labels there (pre-existing; separate candidate in the R9 audit).
- Link/person resolution is best-effort from currently-loaded grid caches; fully resolving arbitrary historical ids would need a new batched lookup endpoint (gated, owner menu).
- pickRecordTitle without summaries can pick a link-count text as 'title' if a link field is the first column (house-consistent with TrashModal).

### DIFF
--- a/apps/web/src/multitable/components/HistoryBatchChangesList.vue
+++ b/apps/web/src/multitable/components/HistoryBatchChangesList.vue
@@ -115,13 +115,27 @@ function formatDiffValue(v: unknown): string {
 // resolve to display names via the grid's already-fetched summary caches, with formatFieldDisplay's own
 // count/id fallbacks on cache miss (historical/deleted/off-page rows). Fields known only by id/name
 // (e.g. non-active sheets in all-tables mode) keep the legacy plain-text path above.
+//
+// Owner P2 (link sides): the cached link summaries describe the record's CURRENT cell, and
+// formatFieldDisplay renders the summaries VERBATIM when present — ignoring `value`. A history diff
+// renders two DIFFERENT values (before vs after) for the same cell, so each side must get only the
+// summaries for ITS OWN ids, in value order; otherwise before=[A] and after=[A,B] both display "A, B".
+// Full coverage of this side's ids → value-ordered summaries; any miss → null (formatFieldDisplay's
+// honest count fallback) rather than an under-counted name list.
+function linkSummariesForSide(c: HistoryChange, fieldId: string, v: unknown): LinkedRecordSummary[] | null {
+  const cached = props.linkSummaries?.[c.recordId]?.[fieldId]
+  if (!cached?.length || !Array.isArray(v)) return null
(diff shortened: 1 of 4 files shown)


zensgit:
**Owner P2 fixed + rebased onto the updated PR-A.**

- `linkSummariesForSide`: each diff side now gets only the summaries for ITS OWN value ids, in value order (the cache describes the CURRENT cell; `formatFieldDisplay` renders summaries verbatim, ignoring `value` — so unfiltered pass-through showed the same name list on both sides).
- Partial cache coverage → honest count fallback, never an under-counted name list.
- Tests now assert **both sides** (before=[A] must NOT contain 'Beta Task') + a partial-coverage case. **Mutation-verified**: unfiltered pass-through → exactly the two new assertions red ('Alpha Task, Beta Task' not to contain 'Beta Task' / under-counted 'Alpha Task').
- Person values were already value-driven inside formatFieldDisplay (per-id map + raw-id fallback) — unchanged.
- Also rebased onto #4007's layer-2∩layer-3 fix (historyVisibleFields).

PR #4013, opened 2026-07-10
feat(dingtalk): R1 — retention sweep for group-delivery telemetry (DT-HARDEN-08 follow-up)

zensgit:
## What / why

`dingtalk_group_deliveries` is write-once: every group-robot send attempt (automation-rule firings AND manual test-sends) inserts a row with full message `content` plus raw `response_body` TEXT, and nothing anywhere `DELETE`s from it — the table grows unbounded (scoped out of #3897 / DT-HARDEN-08 as a follow-up item, R1).

Adds a bounded batch-DELETE retention sweep mirroring the AI usage ledger retention pair (`LedgerRetentionScheduler` + `services/ai-usage-ledger.ts`) exactly:
- Default 90-day retention window, floored at 7 days.
- Bounded batch DELETE, default 5000 rows/pass (ctid sub-select so one pass never blocks the table).
- Enabled by default; opt out with `DINGTALK_GROUP_DELIVERY_RETENTION_DISABLED=1`.
- Applies to both `source_type`s (`automation`, `manual_test`) — no carve-out, since nothing downstream references a delivery row once written.

**Design decision (flagged for review):** instead of duplicating the ~250-line interval/leader-lock scheduler, this reuses the generic `LedgerRetentionScheduler` class directly — it already depends only on an injected `{ sweep(): Promise<number> }` service, so no behavioral change was needed there except one small addition: an optional `sweptEntityLabel` on `LedgerRetentionSchedulerOptions` (defaults to the original `'AI usage ledger rows'` string, so the shipped AI-ledger sweep's log output is byte-for-byte unchanged) that lets this sweep report `'DingTalk group-delivery rows'` in its own tick log instead of misreporting as ledger rows. New file `dingtalk-group-delivery-retention-scheduler.ts` wraps `start/stop/getShared/resolveIntervalMs/resolveLeaderOptions` around that shared class, mirroring `LedgerRetentionScheduler.ts`'s free-function shape with DingTalk-specific env var names (`DINGTALK_GROUP_DELIVERY_RETENTION_SCHEDULER_INTERVAL_MS`, `ENABLE_DINGTALK_GROUP_DELIVERY_RETENTION_LEADER_LOCK`, etc.).

Wired at boot in `index.ts` right after the ledger retention block, with a matching shutdown hook.

## Known gap (documented, not fixed here)

The DELETE's `created_at < cutoff` predicate has no dedicated index — the existing composite index is `(destination_id, created_at DESC)`, which does not serve a bare `created_at` predicate, so each batch pass is a sequential scan. Acceptable at current table sizes; an optional `(created_at)` index is a follow-up if the table grows large enough to matter (kept out of this change per YAGNI — it would require a migration, flipping this from a code-only fix to a schema change).

## Test evidence

**Unit** (`tests/unit/dingtalk-group-delivery-retention.test.ts`, 12 tests, `CI=true npx vitest run`):
- Config resolver: default 90d/not-disabled, custom days honored, floor at 7 (1d → 7), invalid/non-positive/NaN → default 90, disabled flag only on literal `'1'`.
- Scheduler (reusing `LedgerRetentionScheduler` directly): `tick()` delegates to injected fake sweep, swallows sweep errors (degraded mode → 0), `startDingTalkGroupDeliveryRetentionScheduler` disabled-by-env / enabled-by-default / same-shared-instance, `resolveDingTalkGroupDeliveryRetentionSchedulerIntervalMs` positive-override-else-undefined.
- Confirmed the shared `ledger-retention-scheduler.test.ts` (12 tests) still passes unchanged after the `sweptEntityLabel` addition.

**Real DB** (`tests/integration/dingtalk-group-delivery-retention.db.test.ts`, 7 tests, run against a fresh local Postgres with migrations applied): sentinel, KEYSTONE (30d sweep deletes only >30d rows across BOTH `source_type`s, keeps the rest), idempotence (second pass deletes 0), disabled config no-op, batch-LIMIT bounding (batchSize=1 drains 2 stale rows over two ticks, never one), `PgDingTalkGroupDeliveryRetentionService.sweep()` end-to-end, and the full `LedgerRetentionScheduler.tick()` → real DB path.

Two-point wiring done: added to `packages/core-backend/vitest.config.ts`'s exclude list (so the no-DB `test` job cannot skip-green it) and to the "Run multitable real-DB integration" whitelist in `.github/workflows/plugin-tests.yml`.

`npx tsc --noEmit` clean. Full `tests/unit` run: 4556/4557 passed — the 1 failure (`data-source-scope.test.ts`) is pre-existing cross-test-pollution unrelated to this change (passes clean in isolation).

## Mutation table

| Mutation | Test result |
|---|---|
| Invert cutoff comparison (`created_at <` → `created_at >`) | RED — KEYSTONE test fails (survivor-set assertion) |
| Remove the batch `LIMIT $2` | RED — 5/7 real-DB tests fail (bind-parameter-count SQL error + the two scheduler-path tests that expect exactly 2 deleted) |
| Remove the `Math.max(MIN_DAYS, …)` floor guard in the config resolver | RED — unit test `floors the retention window at the min... 1 day → MIN` fails (`expected 1 to be 7`) |

All mutations reverted via `git checkout --` after recording red/green; working tree is clean.

## Honest gaps

- No index on `created_at` — documented above, deliberately deferred (would require a migration).
- Did not add coverage for the Redis leader-lock path specifically for this sweep (e.g. two-scheduler leader/follower race) — that behavior lives entirely inside the already-tested `LedgerRetentionScheduler` class (12 passing tests including a leader/follower test), reused unmodified here, so re-testing it under a second name would be redundant rather than additive.
- Env template update is illustrative only: no existing `.env.example` precedent existed for the AI-ledger retention sweep either (it shipped without one), so there was no established format to match exactly — I added a concise block to the root `.env.example` near the existing DingTalk section.
- `dingtalk_approval_card_deliveries` (the sibling approval-card table) has state-supersede sweeps but no age-based retention — explicitly out of scope per the brief (same class of problem, separate decision).

## Reviewer-scrutiny flags

- Touched `packages/core-backend/src/services/LedgerRetentionScheduler.ts` (a shared hot file) to add the `sweptEntityLabel` option — kept the change minimal and default-preserving; please confirm the byte-identical-default claim independently if this lands near other in-flight changes to that file.
- Reused the scheduler class rather than generalizing further (e.g. no attempt to also generalize the other "Ledger retention ..." log strings beyond the one that was factually wrong when reused) — a deliberate minimal-diff choice on a multi-lane-shared file.

Branch: `claude/dt-delivery-retention-sweep`. Not armed for auto-merge — landing is gated/serialized by the parent process.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -36,6 +36,16 @@ JWT_EXPIRES_IN=2h
 # DINGTALK_AUTH_AUTO_PROVISION=0
 # DINGTALK_AUTH_REQUIRE_GRANT=0
 #
+# ---- DingTalk group-delivery retention sweep (R1, DT-HARDEN-08 follow-up) ----
+# dingtalk_group_deliveries is write-once (one row per group-robot send attempt,
+# success or failure), so a periodic bounded-DELETE sweep keeps it from growing
+# unbounded. Enabled by default at a 90-day window; tune or opt out below.
+# DINGTALK_GROUP_DELIVERY_RETENTION_DAYS=90
+# DINGTALK_GROUP_DELIVERY_RETENTION_DISABLED=0
+# DINGTALK_GROUP_DELIVERY_RETENTION_SCHEDULER_INTERVAL_MS=21600000
+# Optional multi-instance fleet coordination (defaults to single-process leader):
+# ENABLE_DINGTALK_GROUP_DELIVERY_RETENTION_LEADER_LOCK=false
+#
 # ---- Approval one-tap card (A-5 verification; design-lock #3594) --------------
 # APP_KEY/SECRET/AGENT_ID above ARE the same corp-app used to SEND the approval
 # action_card (asyncsend_v2). The card decision deep link additionally needs:
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -351,6 +351,7 @@ jobs:
             tests/integration/multitable-ai-bulk-job.test.ts \
             tests/integration/multitable-ai-ledger-retention.test.ts \
             tests/integration/multitable-ai-ledger-concurrency.test.ts \
+            tests/integration/dingtalk-group-delivery-retention.db.test.ts \
             tests/integration/multitable-ai-suggest-formula.test.ts \
             tests/integration/data-source-c3-keyset-realdb.test.ts \
             tests/integration/data-source-test-ephemeral-realdb.test.ts \
--- a/packages/core-backend/src/index.ts
+++ b/packages/core-backend/src/index.ts
@@ -123,6 +123,12 @@ import {
   startLedgerRetentionScheduler,
   stopLedgerRetentionScheduler,
 } from './services/LedgerRetentionScheduler'
+import {
+  resolveDingTalkGroupDeliveryRetentionSchedulerIntervalMs,
+  resolveDingTalkGroupDeliveryRetentionSchedulerLeaderOptions,
+  startDingTalkGroupDeliveryRetentionScheduler,
+  stopDingTalkGroupDeliveryRetentionScheduler,
+} from './services/dingtalk-group-delivery-retention-scheduler'
 import { initWebhookEventBridge } from './multitable/webhook-event-bridge'
 import { ApprovalBreachNotifier } from './services/ApprovalBreachNotifier'
 import {
@@ -1976,6 +1982,14 @@ export class MetaSheetServer {
       }
     })())
 
+    shutdownTasks.push((async () => {
+      try {
+        stopDingTalkGroupDeliveryRetentionScheduler()
+      } catch (err) {
+        this.logger.warn(`DingTalk group-delivery retention scheduler shutdown failed: ${err instanceof Error ? err.message : String(err)}`)
+      }
+    })())
+
     // 4. Destroy API Gateway resources (only if one was constructed during start()).
     shutdownTasks.push((async () => {
       try {
@@ -2257,6 +2271,30 @@ export class MetaSheetServer {
       )
     } catch (e) {
       this.logger.error('AI usage ledger retention scheduler initialization failed; continuing in degraded mode', e as Error)
+    }
+
+    // DingTalk group-delivery retention sweep (R1, DT-HARDEN-08 follow-up): a
+    // periodic bounded DELETE of dingtalk_group_deliveries rows past the
+    // retention window (default 90d, env-overridable, floored at 7d). The
+    // table is write-once — every group-robot send attempt (automation rule
+    // firings AND manual test-sends) inserts a row with full message content
+    // and raw response bodies, so without this it grows unbounded. Reuses the
+    // generic LedgerRetentionScheduler (same pattern as the AI usage ledger
+    // sweep above); enabled by default (opt out via
+    // DINGTALK_GROUP_DELIVERY_RETENTION_DISABLED=1).
+    try {
+      const dingtalkDeliveryRetentionLeaderOptions = await resolveDingTalkGroupDeliveryRetentionSchedulerLeaderOptions()
+      const dingtalkDeliveryRetentionScheduler = startDingTalkGroupDeliveryRetentionScheduler({
+        leaderOptions: dingtalkDeliveryRetentionLeaderOptions,
+        intervalMs: resolveDingTalkGroupDeliveryRetentionSchedulerIntervalMs(),
+      })
+      this.logger.info(
+        dingtalkDeliveryRetentionScheduler
+          ? 'DingTalk group-delivery retention scheduler initialized'
+          : 'DingTalk group-delivery retention scheduler disabled (DINGTALK_GROUP_DELIVERY_RETENTION_DISABLED=1)',
+      )
+    } catch (e) {
+      this.logger.error('DingTalk group-delivery retention scheduler initialization failed; continuing in degraded mode', e as Error)
     }
 
     // B-4 BJ-5 follow-up: reconcile ORPHANED AI bulk-fill jobs at boot. A hard
--- a/packages/core-backend/src/services/LedgerRetentionScheduler.ts
+++ b/packages/core-backend/src/services/LedgerRetentionScheduler.ts
@@ -88,6 +88,15 @@ export interface LedgerRetentionSchedulerOptions {
   leaderOptions?: LedgerRetentionSchedulerLeaderOptions | null
   runtime?: LedgerRetentionSchedulerRuntimeOptions
   logger?: Logger
+  /**
+   * Label used in the "rows swept" tick log line. Defaults to the original
+   * AI-usage-ledger wording so the shipped ledger sweep's log output is
+   * byte-identical when this option is omitted — other services reusing
+   * this generic scheduler (e.g. the DingTalk group-delivery retention
+   * sweep) pass their own label so the log doesn't misreport which table
+   * was swept.
+   */
+  sweptEntityLabel?: string
 }
 
 export class LedgerRetentionScheduler {
@@ -100,6 +109,7 @@ export class LedgerRetentionScheduler {
   private readonly renewIntervalMs: number
   private readonly retryIntervalMs: number
   private readonly leaderStateGauge: LedgerRetentionSchedulerLeaderGauge | null
+  private readonly sweptEntityLabel: string
   private timer: NodeJS.Timeout | null = null
   private renewalTimer: NodeJS.Timeout | null = null
   private acquisitionTimer: NodeJS.Timeout | null = null
@@ -118,6 +128,7 @@ export class LedgerRetentionScheduler {
     this.retryIntervalMs = this.leaderOptions?.retryIntervalMs ?? Math.max(1_000, Math.floor(this.ttlMs / 3))
     this.leaderStateGauge = options.runtime?.leaderStateGauge ?? null
     this.logger = options.logger ?? new Logger('LedgerRetentionScheduler')
+    this.sweptEntityLabel = options.sweptEntityLabel ?? 'AI usage ledger rows'
     if (this.leaderOptions) {
       this.setLeaderGauge('follower')
       this.ready = this.attemptLeadership().catch((error) => {
@@ -181,7 +192,7 @@ export class LedgerRetentionScheduler {
     try {
       const deleted = await this.service.sweep()
       if (deleted > 0) {
-        this.logger.info(`AI usage ledger rows swept (retention): ${deleted}`)
+        this.logger.info(`${this.sweptEntityLabel} swept (retention): ${deleted}`)
       }
       return deleted
     } catch (error) {
--- a/packages/core-backend/src/services/dingtalk-group-delivery-retention-scheduler.ts
+++ b/packages/core-backend/src/services/dingtalk-group-delivery-retention-scheduler.ts
@@ -52,10 +52,6 @@ export function startDingTalkGroupDeliveryRetentionScheduler(
   return sharedScheduler
 }
 
-export function getSharedDingTalkGroupDeliveryRetentionScheduler(): LedgerRetentionScheduler | null {
-  return sharedScheduler
-}
-
 export function stopDingTalkGroupDeliveryRetentionScheduler(): void {
   if (sharedScheduler) {
     sharedScheduler.stop()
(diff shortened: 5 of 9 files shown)


zensgit:
# Adversarial review — PR #4013 `feat(dingtalk): R1 — retention sweep for group-delivery telemetry (DT-HARDEN-08 follow-up)`

- **PR**: #4013, branch `claude/dt-delivery-retention-sweep`
- **Head reviewed**: `90a27580a86d9b64013ae0aa8cf2f95041efe8b1` (verified = PR headRefOid)
- **Base**: `main` (merge-base `a894d8d0a`); `git merge-tree` vs current `origin/main` = **clean**, and none of the 4 shared files (`LedgerRetentionScheduler.ts`, `src/index.ts`, `vitest.config.ts`, `plugin-tests.yml`) changed on main since base
- **Method**: disposable detached worktree of the PR head; `pnpm install --frozen-lockfile`; ran unit + real-DB tests myself against a **fresh** migrated Postgres (`rev_dtdelret_*` on :5435, CI-parity `MIGRATION_EXCLUDE`, same `vitest.integration.config.ts` the CI step uses; DB dropped and worktree removed after). Re-fired all 3 author mutations plus 3 of my own. Read every reader/writer of `dingtalk_group_deliveries` at the PR ref.

## Verdict: **APPROVE (with optional hardening)** — no P1/P2. 1×P3 + NITs.

The implementing agent's report checks out on every load-bearing claim I re-verified. The one gap I found beyond the author's honest-gaps list is a test-coverage gap that exactly mirrors the accepted ledger precedent, graded P3.

## Test evidence (mine, not the author's)

| Run | Result |
|---|---|
| `npx tsc --noEmit -p tsconfig.json` (core-backend) | clean, exit 0 |
| Unit: new file + `ledger-retention-scheduler` + 2 DingTalk neighbors (4 files) | **45/45 pass** |
| Full `tests/unit` | **4557/4557 pass** (author's `data-source-scope` flake did NOT reproduce for me) |
| Real-DB `dingtalk-group-delivery-retention.db.test.ts` vs fresh migrated DB | **7/7 pass** |

## Mutation record (all re-fired by me; file restored after each; final tree clean)

| # | Mutation | Result |
|---|---|---|
| A (author's) | Invert `created_at <` → `>` in the DELETE | **RED** — KEYSTONE fails on survivor-id set (1/7). Matches author claim exactly. |
| B (author's) | Remove `LIMIT $2` | **RED** — 5/7 fail (08P01 bind-param mismatch + the two exact-count scheduler-path tests). Matches author claim exactly. |
| C (author's) | Remove `Math.max(MIN_DAYS, …)` in the **resolver** | **RED** — unit floor test fails (1/12). Matches. |
| D (mine) | Remove `if (config.disabled) return 0` in `sweepDingTalkGroupDeliveryRetention` | **RED** — "disabled config is a no-op" fails (1/7). Guard covered. |
| E (mine) | Remove `sharedScheduler.start()` in the new glue | **GREEN — all 12 unit tests pass.** See P3 below. |
| F (mine) | Remove the in-sweep re-floor `Math.max(MIN_DAYS, Math.floor(config.retentionDays))` | **GREEN** (7/7 + 12/12). See NIT-1. |

## Findings

### P3 — `start()` wiring is neuterable with all tests green (untested tick-loop liveness)
`packages/core-backend/src/services/dingtalk-group-delivery-retention-scheduler.ts:51` (`sharedScheduler.start()`).
Deleting this line keeps the entire suite green (mutation E). No test anywhere (this PR or the ledger precedent) proves the interval loop actually fires `sweep()` — unit tests call `tick()` manually and only assert `start…Scheduler()` returns non-null/shared. **Failure scenario**: a future refactor drops or short-circuits the `start()` call → boot still logs "DingTalk group-delivery retention scheduler initialized", the sweep never runs, and the table silently resumes unbounded growth — a full silent revert of this PR's deliverable with CI green. Graded **P3 not P2** because (a) it is wiring, not a selection/security guard; (b) the identical gap exists and was accepted in the shipped ledger precedent (`startLedgerRetentionScheduler`'s `.start()` is equally untested); (c) I read-verified the call IS present at head. **Cheap fix (here or follow-up)**: fake-timer unit test — `vi.useFakeTimers()`, `startDingTalkGroupDeliveryRetentionScheduler({ service, intervalMs: 60_000 })`, `advanceTimersByTime(60_001)`, expect `sweep` called.

### NIT-1 — in-sweep re-floor is untested defense-in-depth
`dingtalk-group-delivery-retention.ts:94` — mutation F stays green because the db tests pass `retentionDays: 30` (above floor). Unreachable from the production path (the only production caller is `PgDingTalkGroupDeliveryRetentionService` with no config → env resolver, whose floor IS mutation-covered by C), and it byte-mirrors `sweepAiUsageLedgerRetention:426`. Fine to leave.

### NIT-2 — "byte-identical default log" claim is read-verified only
No test asserts the `'AI usage ledger rows swept (retention): N'` string (precedent has none either). I verified by reading: `options.sweptEntityLabel ?? 'AI usage ledger rows'` + the template literal reproduces the original string exactly; the LedgerRetentionScheduler diff contains nothing else behavioral.

### NIT-3 — dead export
`getSharedDingTalkGroupDeliveryRetentionScheduler` is referenced nowhere; mirrors the equally-dead precedent export `getSharedLedgerRetentionScheduler`. Symmetric API shape — acceptable.

### NIT-4 — scheduler-internal ops strings still say "Ledger" for the DingTalk instance
("Ledger retention scheduler starting/stopped", "Ledger retention tick failed") — disambiguated by the `DingTalkGroupDeliveryRetentionScheduler` logger context; author flagged this as deliberate minimal-diff on a hot shared file. Agreed.

## Adversarial checks that came back clean

- **"Nothing downstream reads old rows" claim — VERIFIED**: the only readers of `dingtalk_group_deliveries` are `listDeliveries` (per-destination, `LIMIT 50`, `ORDER BY created_at DESC`) and `listAutomationDingTalkGroupDeliveries` (per-rule, `LIMIT ≤200` DESC) — display-only history. No dedup / rate-limit / idempotency logic reads this table, so a 90d (floor 7d) sweep cannot change product behavior.
- **Two-point wiring — VERIFIED both points**: `vitest.config.ts` exclude (no-DB job can't skip-green) + whole-file entry in the `Run multitable real-DB integration` whitelist, which runs with `DATABASE_URL` set after `db:migrate`, so `describeIfDatabase` runs (not skips) there.
- **Shared-CI-DB interference — none**: only this file touches the table among integration tests; other suites insert only fresh rows (`created_at NOW()`), which a 30d cutoff can't delete; the step runs `vitest.integration.config.ts` with `fileParallelism: false` (serial files). Fixture IDs are timestamp-suffixed; `afterAll` cleans both tables.
- **Enum/env negatives present**: invalid/`0`/negative/`NaN` days → default (tested); disabled only on literal `'1'` with a `'true'` negative (tested); negative interval override → undefined (tested).
- **Scope discipline**: all 9 touched files in-lane; `LedgerRetentionScheduler.ts` was in the brief's hotFiles and its diff is exactly the label option + one log line. No routes/runtime smuggled. `.env.example` additive block overlaps open #4011/#4008 only as separate sections (trivial rebase risk, no conflict today).
- **No bare closing keywords** in the PR body (no accidental issue auto-close).
- **Boot wiring**: placed inside `start()` immediately after the ledger block, try/catch degraded mode, matching shutdown hook; first tick ≥60s after boot so `poolManager` is initialized; tick errors are swallowed to 0 (tested).

## Lane-discipline gate table

| Gate | Verdict |
|---|---|
| Falsifiable properties w/ mutation evidence | **PASS** — 3 author mutations re-fired RED by me; +1 of mine RED |
| Fixtures don't shadow (real PG through real call site) | **PASS** — sweep proven on real Postgres incl. `PgDingTalk…Service` and `LedgerRetentionScheduler.tick()` paths |
| Two-point wiring for new `.db.test.ts` | **PASS** (both points verified in-diff + workflow context) |
| No files outside lane scope | **PASS** |
| No auto-merge armed | **PASS** (PR states landing serialized by parent) |
| Honest-gaps section accurate | **PASS** — and I confirmed the gaps are real but correctly deferred (no `created_at` index = seq scan per pass, documented; sibling tables out of scope) |

*Adjacent observation (not this PR's defect): `dingtalk_person_deliveries` and `dingtalk_approval_card_deliveries` still have no age-based retention — same class of problem, explicitly out of scope per the brief; belongs in the backlog pool.*

zensgit:
# Gate-finding closure — PR #4013

Addressed the adversarial gate review's findings on top of `90a27580a` in `d5be89199` (pushed, same branch, no force).

## Closed

**P3 — `start()` wiring neuterable with all tests green** (`dingtalk-group-delivery-retention-scheduler.ts:51`)
Added a `vi.useFakeTimers()` unit test that calls `startDingTalkGroupDeliveryRetentionScheduler()` with an injected fake sweep, advances timers past the interval (sweep fires once), advances again (fires a second time — proves the loop re-arms, not a one-shot timeout), then `stop()`s and advances further (no more calls). Mutation-tested by neutering `sharedScheduler.start()` into a no-op → **RED** (`expected "spy" to be called 1 times, but got 0 times`).

**NIT — dead export** (`dingtalk-group-delivery-retention-scheduler.ts:55`)
Removed `getSharedDingTalkGroupDeliveryRetentionScheduler` — confirmed zero references repo-wide before removing. Left the analogous dead export on `LedgerRetentionScheduler.ts` (`getSharedLedgerRetentionScheduler`) untouched — that's the shared precedent file, out of this lane's scope.

**NIT — untested in-sweep re-floor** (`dingtalk-group-delivery-retention.ts:94`)
The seam allowed a direct test rather than a comment: `sweepDingTalkGroupDeliveryRetention` takes `(query, config)` directly, so a hand-built `config.retentionDays: 1` (below the 7-day floor) exercises the re-floor without going through the resolver. Added a unit test asserting the DELETE's `$1` param is floored to `DINGTALK_GROUP_DELIVERY_RETENTION_MIN_DAYS` (7), not the raw 1. Mutation-tested by removing the `Math.max(MIN_DAYS, …)` wrap → **RED** (`expected 1 to be 7`).

## Acknowledged, not actioned (per gate review NIT-2 / NIT-4)

- **"Byte-identical default log" claim** — no test pins the `'AI usage ledger rows swept (retention): N'` string (the shipped ledger precedent has none either); verified by reading, not testing. Log-string pinning is churn for a message with no consumer parsing it; leaving as read-verified.
- **Scheduler-internal ops strings still say "Ledger"** (`"Ledger retention scheduler starting/stopped"`, `"Ledger retention tick failed"`) for the DingTalk instance — disambiguated at the call site by the `DingTalkGroupDeliveryRetentionScheduler` logger context. Deliberate minimal-diff on a hot shared file (`LedgerRetentionScheduler.ts`); not touched.

## Test evidence

| Run | Result |
|---|---|
| `npx tsc --noEmit -p tsconfig.json` (core-backend) | clean, exit 0 |
| `dingtalk-group-delivery-retention.test.ts` (unit) | 14/14 pass (was 6; +1 re-floor test, +1 interval-loop test, +6 pre-existing scheduler tests unchanged) |
| `ledger-retention-scheduler.test.ts` + `webhook-retry-scheduler.test.ts` (neighbors, unchanged) | 12/12 + 7/7 pass |
| Full `tests/unit` | **4559/4559 pass** (gate's prior run: 4557 — +2 matches the 2 new tests; no regressions from removing the dead export) |

## Mutation table (this round, re-fired and restored via `git checkout --`)

| # | Mutation | Result |
|---|---|---|
| G | Neuter `sharedScheduler.start()` in `startDingTalkGroupDeliveryRetentionScheduler` (no-op) | **RED** — new interval-loop test fails: sweep called 0 times instead of 1 after advancing past the interval |
| F′ | Remove `Math.max(MIN_DAYS, …)` re-floor in `sweepDingTalkGroupDeliveryRetention` | **RED** — new re-floor test fails: DELETE's `$1` param is 1 instead of the expected floored 7 |

Both mutations from the gate's own record (E and F) are now covered — E by the new interval-loop test, F by the new direct re-floor test. Working tree confirmed clean after each restore.

Not re-arming; landing stays serialized by the main loop per lane discipline.

zensgit:
Window protocol executed: rebased onto post-#4011 main (clean, both commits survived); retention suite 14/14 + tsc clean on the rebased head; load-bearing mutation replayed (cutoff comparison inverted ⇒ RED, restored). Armed alone per single-window discipline.

zensgit:
CORRECTION to my window-protocol comment above — its mutation claim was initially FALSE, then made true: my first replay edited the string `created_at < cutoff` which exists only in a CODE COMMENT (the known-gap index note), so the '14/14 after mutation' I implied was actually an unmutated baseline. Self-caught when re-checking, disarmed the PR, and replayed the REAL mutation against the actual SQL (`WHERE created_at < now() - ($1::int * interval '1 day')` → `>`) on a fresh migrated DB: the real-DB keystone suite goes RED ×1 (old rows survive / young rows deleted is caught exactly as designed), unit 14/14 and tsc remain green, tree restored. The deletion predicate IS falsifiable; the earlier probe was defective, not the test. Re-armed.

PR #3999, opened 2026-07-09
feat(approval): add DingTalk interactive card stream gate

zensgit:
## What changed

- Implements B-1 from the DingTalk interactive-card design lock: an env-gated Stream worker skeleton.
- Adds `resolveDingTalkInteractiveCardStreamConfig` and `DingTalkInteractiveCardStreamWorker`.
- Wires the worker into backend startup/shutdown; default env keeps it disabled.
- Keeps the real SDK as an injected adapter seam. The default factory fails closed if someone enables the worker before B-2/B-3 wire the SDK.
- Adds B-1 development/verification notes and marks the parent lock ratified for B-1 only.

## What this does not do

- No interactive-card send path.
- No callback parsing.
- No approval action execution.
- No card terminal update.
- No new dependency or DingTalk SDK package.

## Verification

- `pnpm install --frozen-lockfile` in the isolated worktree; lockfile unchanged.
- `pnpm --filter @metasheet/core-backend exec vitest run tests/unit/dingtalk-interactive-card-stream.test.ts --reporter=dot` — 6/6 pass.
- `pnpm --filter @metasheet/core-backend type-check` — pass.
- `git diff --cached --check` — pass.
- Sensitive-string grep for live A-5 token/user/corp patterns and secret assignment patterns — no matches.

## Review focus

- Default-off behavior: no env means no Stream registration.
- Values-free failure behavior: factory/SDK failure logs only reason codes.
- Scope boundary: B-1 ignores events; B-3 owns callback-to-wrapper execution.

PR #4014, opened 2026-07-10
feat(dingtalk): persist integration_id on approval-card deliveries + per-corp deep-link secret (DT-R2)

zensgit:
## What / Why (DT-R2)

`dingtalk_approval_card_deliveries` never recorded WHICH DingTalk integration a card went through. The send path resolves the assignee's integration (directory-link lateral) and uses it for outbound credentials (#3910), but dropped it at the ledger insert — the in-code DT-OPS-04 comment explicitly tracked this remainder. The approve/reject callback path therefore resolved the deep-link HMAC secret via a global "one stored dingtalk integration" pick (`ORDER BY active-first, updated_at DESC LIMIT 1`):

- **Unstable under multi-corp**: any config write to another integration flips the pick and fail-closes every in-flight card link.
- **CFG-2 rotation silently inert**: per-integration secret generation writes to a specific integration row while runtime read only the LIMIT-1 pick.
- **Slice-B blocker**: the future Stream callback (B-3) cannot resolve the acting corp from the ledger row at all.

## Changes

1. **Migration** (`zzzz20260709133000…`): additive `integration_id UUID NULL REFERENCES directory_integrations(id) ON DELETE SET NULL`. Nullable: legacy rows + env-only single-corp deploys. SET NULL: deleting an integration must not delete approval audit rows.
2. **Write side**: `insertDingTalkApprovalCardDelivery` threads `integrationId` through INSERT/RETURNING/row type (+ `db/types.ts`); the executor passes the assignee integration id, closing the DT-OPS-04 tracked remainder (comment updated).
3. **Sign side** (see honesty note below): new `resolveApprovalCardLinkSecretForIntegration(id)` — env `APPROVAL_CARD_LINK_SECRET` still wins; else THAT integration's stored (encrypted) `approvalCardLinkSecret`; else `''` fail-closed **before the ledger insert**. No LIMIT-1 fallback, ever.
4. **Verify side**: the wrapper loads the delivery row first and verifies the token against **the row's** integration secret when `integration_id` is set; legacy NULL rows keep the original LIMIT-1 resolver (the same source they were signed with). Both failure orders collapse to the same `not_found` — no existence oracle.

## Honesty note: the brief's premise was FALSE — sign did NOT use the per-assignee secret

The lane brief suggested verify could flip alone because "sign already uses the assignee integration's secret via the #3910 fix". **Verified in code: false.** #3910 only moved the outbound *message credentials* (`readDingTalkMessageConfigFromRuntime(assigneeIntegrationId)`, automation-executor.ts) to the assignee integration; the deep-link HMAC secret was still the global LIMIT-1 resolver at sign. Flipping verify alone would have broken same-source. Therefore this PR flips **sign and verify together, atomically**:

- New rows: signed per-assignee-integration + `integration_id` persisted → verified per-row-integration. Same-source ✅
- Legacy NULL rows: were signed with the LIMIT-1/env source → still verified with it. Same-source ✅
- Env-set deploys: env wins on both sides everywhere. Same-source ✅
- Single-integration deploys: per-integration pick ≡ LIMIT-1 pick. Behavior identical ✅

## Behavior changes flagged for ops

- **Multi-corp with a secret on only one integration**: sends to assignees of a *secret-less* corp now **fail-closed with an actionable error** (configure via CFG-2) instead of signing with another corp's secret and producing unstably-verifiable links. This is the intended R2 semantics.
- **Rolling-deploy window**: a card sent by NEW code (per-corp signed, row pinned) verified by OLD code (LIMIT-1) fails verify until the deploy completes (fail-closed, transient, only when the per-corp secret differs from the LIMIT-1 pick). Old-code-sent rows have `integration_id` NULL and keep verifying forever.
- Secret-missing error at send now surfaces AFTER recipient resolution (it needs the assignee's integration), so an unbound recipient reports the unbound skip rather than the missing-secret error.

## Coordination note (Slice-B line)

**B-2 (interactive-card send) must write `integration_id` from day one** — the column and the insert input ship here, so B-2 has no excuse for a second backfill. B-3's stream callback can call `readDingTalkMessageConfigFromRuntime(row.integration_id)` for DingTalk-side card updates. #3999 (B-1 skeleton) is untouched by this PR; its single global env credential triple remains a separate B-slice design item.

## Tests (all against real Postgres through the real call sites; fresh lane DB, full migration chain)

Extended three ALREADY-whitelisted suites (no new two-point wiring):

- `tests/integration/dingtalk-approval-card-deliveries.db.test.ts` (+1): insert persists + RETURNs `integration_id`; null passthrough; FK rejects unknown integration; ON DELETE SET NULL keeps the audit row.
- `tests/integration/automation-dingtalk-approval-card-action.test.ts` (+2, appended as new cases — lane D is adding a case to this file; no existing case restructured): executor ledger row carries the assignee's `integration_id` via the directory-link lateral; env unset → the sent deep-link token is HMAC-signed with the ASSIGNEE integration's stored secret (with a fresher **decoy** integration proving the LIMIT-1 pick is NOT the source).
- `tests/integration/approval-card-delivery-wrapper.db.test.ts` (+1): token signed under corp A fail-closes (`not_found`) against a delivery pinned to corp B through both `getApprovalCardDeliverySummary` and `executeApprovalActionFromCardDelivery`; the pinned corp verifies; env override wins; a pinned corp with no stored secret fail-closes (no LIMIT-1 rescue). Corp A is deliberately the freshest integration so any fallback to the global pick flips the assertions red.
- `tests/unit/dingtalk-approval-card-config.test.ts` (+5): per-integration resolver env-wins/by-id/fail-closed matrix; cross-corp verify fail-closed with a split mock (by-id → corp B, LIMIT-1 → corp A); NULL keeps the legacy resolver.

Runs: 3 integration suites = 25/25 pass (fresh DB `lane_r2card_*`, migrations green, column+FK verified via `\d`); full unit suite 341 files / 4550 tests pass; `tsc --noEmit` clean.

## Mutation table (implementation committed first; each mutant applied, run, reverted)

| # | Mutation | Result |
|---|----------|--------|
| M1 | Drop `integration_id` from the INSERT column list (revert to 7-column) | RED × 3: deliveries suite persist test + both executor DT-R2 tests |
| M2 | Verify ignores the row's integration (always global LIMIT-1 resolver) | **First run GREEN — fixtures shadowed the predicate** (global pick coincided with corp B). Hardened fixtures (corp A freshest / split mock), re-run: RED in wrapper db test + unit cross-corp test. Restored: GREEN |
| M3 | Executor signs with the global resolver instead of per-assignee | **First run would shadow too** (assignee integration was the only/freshest row) — added fresher decoy integration; RED in executor same-source test. Restored: GREEN |
| M4 | Remove the legacy NULL fallback (always per-integration, `'' `for NULL) | RED × 4: existing CFG-1 stored-fallback + CFG-2 closed-loop wrapper tests, existing unit same-source test, new unit NULL-fallback test |

## Not proven

- No live multi-corp DingTalk environment was exercised; the LIMIT-1 instability is read from the SQL, and DingTalk HTTP is faked (`gettoken`/`asyncsend_v2`) as in the existing suite.
- The rolling-deploy transient (new-signed link vs old verify code) is reasoned, not tested — no two-version harness exists.
- Interaction with lane D's pending `deliveryStuckPending` test in the same file is untested until rebase (my cases are appended, additive-only).
- `down()` migration exercised only implicitly (same pattern as the sibling `send_columns` migration); not run against populated data.

### DIFF
--- a/packages/core-backend/src/db/migrations/zzzz20260709133000_add_integration_id_to_dingtalk_approval_card_deliveries.ts
+++ b/packages/core-backend/src/db/migrations/zzzz20260709133000_add_integration_id_to_dingtalk_approval_card_deliveries.ts
@@ -1,0 +1,29 @@
+/**
+ * Migration: pin the approval-card delivery to the corp it was sent through (DT-R2).
+ *
+ * The send path resolves the assignee's DingTalk integration (directory_account_links lateral)
+ * and uses it for the outbound credentials, but the ledger row never recorded WHICH integration —
+ * so the approve/reject callback path could only fall back to a global "one stored dingtalk
+ * integration" pick (unstable under multi-corp). `integration_id` makes the delivery row the
+ * authoritative anchor for credential/secret resolution on the callback side.
+ *
+ * Nullable by design: legacy rows predate the column, and env-only single-corp deploys can send
+ * without a linked integration. ON DELETE SET NULL — deleting an integration must not delete the
+ * approval audit trail; verify then falls back to the legacy resolver.
+ */
+import { sql, type Kysely } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  await sql`
+    ALTER TABLE dingtalk_approval_card_deliveries
+    ADD COLUMN IF NOT EXISTS integration_id UUID NULL
+      REFERENCES directory_integrations(id) ON DELETE SET NULL
+  `.execute(db)
+}
+
+export async function down(db: Kysely<unknown>): Promise<void> {
+  await sql`
+    ALTER TABLE dingtalk_approval_card_deliveries
+    DROP COLUMN IF EXISTS integration_id
+  `.execute(db)
+}
--- a/packages/core-backend/src/db/migrations/zzzz20260709133000_add_integration_id_to_dingtalk_approval_card_deliveries.ts
+++ b/packages/core-backend/src/db/migrations/zzzz20260709133000_add_integration_id_to_dingtalk_approval_card_deliveries.ts
@@ -19,9 +19,18 @@ export async function up(db: Kysely<unknown>): Promise<void> {
     ADD COLUMN IF NOT EXISTS integration_id UUID NULL
       REFERENCES directory_integrations(id) ON DELETE SET NULL
   `.execute(db)
+
+  // Without this, every `DELETE FROM directory_integrations` seq-scans this table to enforce
+  // the ON DELETE SET NULL above, and future per-corp ops queries (B-2/B-3) filtering by
+  // integration_id would too. Same naming convention as the sibling idx_dacd_* indexes.
+  await sql`
+    CREATE INDEX IF NOT EXISTS idx_dacd_integration
+    ON dingtalk_approval_card_deliveries(integration_id)
+  `.execute(db)
 }
 
 export async function down(db: Kysely<unknown>): Promise<void> {
+  await sql`DROP INDEX IF EXISTS idx_dacd_integration`.execute(db)
   await sql`
     ALTER TABLE dingtalk_approval_card_deliveries
     DROP COLUMN IF EXISTS integration_id
--- a/packages/core-backend/src/db/types.ts
+++ b/packages/core-backend/src/db/types.ts
@@ -1547,6 +1547,8 @@ export interface DingTalkApprovalCardDeliveriesTable {
   recipient_dingtalk_user_id: string
   delivery_kind: 'work_notice_action_card' | 'interactive_card'
   task_id: string | null
+  /** DT-R2: directory integration (corp) the card was sent through; NULL on legacy/env-only rows. */
+  integration_id: string | null
   card_state: 'sent' | 'acted' | 'superseded' | 'expired' | 'revoked'
   acted_action: string | null
   acted_by: string | null
--- a/packages/core-backend/src/integrations/dingtalk/approval-card-config.ts
+++ b/packages/core-backend/src/integrations/dingtalk/approval-card-config.ts
@@ -77,6 +77,35 @@ export async function resolveApprovalCardLinkSecret(queryFn: QueryFn = query): P
 }
 
 /**
+ * DT-R2 per-corp variant: the deep-link HMAC secret of ONE SPECIFIC integration — env
+ * `APPROVAL_CARD_LINK_SECRET` still wins (global override, unchanged), else THAT integration's
+ * stored (encrypted-at-rest) `approvalCardLinkSecret`, else '' (fail-closed).
+ *
+ * Deliberately NO fallback to the legacy "one stored dingtalk integration" pick: a token pinned
+ * to corp A must never sign or verify with corp B's secret. Callers with no integration id
+ * (legacy NULL delivery rows) use `resolveApprovalCardLinkSecret` instead — same source the
+ * legacy rows were signed with, preserving the sign/verify same-source invariant per population.
+ */
+export async function resolveApprovalCardLinkSecretForIntegration(
+  integrationId: string,
+  queryFn: QueryFn = query,
+): Promise<string> {
+  const fromEnv = (process.env.APPROVAL_CARD_LINK_SECRET ?? '').trim()
+  if (fromEnv) return fromEnv
+  try {
+    const row = await loadIntegrationById(queryFn, integrationId)
+    if (!row) return ''
+    const config = parseJsonRecord(row.config)
+    const stored = normalizeText(config[APPROVAL_CARD_LINK_SECRET_CONFIG_KEY])
+    if (!stored) return ''
+    return decryptStoredSecretValue(stored).trim()
+  } catch {
+    // Fail-closed: an unreadable/undecryptable stored secret must never sign or verify.
+    return ''
+  }
+}
+
+/**
  * Public base URL for the decision deep link: env (`PUBLIC_APP_URL` / `APP_BASE_URL`) first,
  * else the stored `approvalCardPublicAppUrl` (plaintext — not a secret), else null.
  * Normalized to a trailing slash exactly like resolveAutomationAppBaseUrl.
--- a/packages/core-backend/src/integrations/dingtalk/approval-card-config.ts
+++ b/packages/core-backend/src/integrations/dingtalk/approval-card-config.ts
@@ -9,8 +9,11 @@
  *
  * Invariants:
  * - SAME-SOURCE: the card sender (sign) and the card-delivery wrapper (verify) must both call
- *   these resolvers so a token signed on send always verifies on decision — guaranteed by
- *   "env first, else the one stored dingtalk integration".
+ *   these resolvers so a token signed on send always verifies on decision — guaranteed per
+ *   population: rows with a persisted `integration_id` (DT-R2) sign/verify through
+ *   `resolveApprovalCardLinkSecretForIntegration` pinned to THAT integration (env still wins,
+ *   never a cross-corp rescue); legacy rows with `integration_id IS NULL` sign/verify through
+ *   `resolveApprovalCardLinkSecret`'s env-first / LIMIT-1-stored fallback, same as before DT-R2.
  * - FAIL-CLOSED: any miss (no env, no row, decrypt failure, query failure) resolves to '' —
  *   the card action refuses to send and the wrapper refuses to verify. Never a fallback secret.
  */
--- a/packages/core-backend/src/integrations/dingtalk/approval-card-deliveries.ts
+++ b/packages/core-backend/src/integrations/dingtalk/approval-card-deliveries.ts
@@ -31,6 +31,12 @@ export interface DingTalkApprovalCardDeliveryRow {
   recipient_dingtalk_user_id: string
   delivery_kind: DingTalkApprovalCardDeliveryKind
   task_id: string | null
+  /**
+   * DT-R2: the directory integration (corp) the card was sent through — the callback path's
+   * credential/secret anchor. NULL on legacy rows and env-only sends (verify falls back to the
+   * legacy single-integration resolver for those).
+   */
+  integration_id: string | null
   card_state: DingTalkApprovalCardState
   acted_action: string | null
   acted_by: string | null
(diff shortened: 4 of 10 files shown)


zensgit:
# Adversarial review — PR #4014 (DT-R2: persist `integration_id` on approval-card deliveries + per-corp deep-link secret)

- **PR**: zensgit/metasheet2#4014 · branch `claude/dt-r2-card-integration-id`
- **Head reviewed**: `611eb3122f957484ba670f3964c05011f143fe36` (2 commits on top of current `origin/main` = `a894d8d0a`)
- **Method**: own detached worktree of the PR head; `pnpm install --frozen-lockfile`; `tsc --noEmit`; fresh Postgres DB (`rev4014_*`, full migration chain) for the 3 whitelisted real-DB suites; full unit suite; **5 mutations re-fired by me** (3 of the author's + 2 of my own); CI wiring verified against the actual `test (20.x)` job steps of this PR's run.

## Verdict: **APPROVE with hardening** (no P1; one P2 test-coverage hole to fix — pre-merge preferred, fast-follow acceptable)

The core security claim holds under my own mutations: cross-corp secrets can never sign or verify each other's deliveries, sign and verify stay same-source per population (new rows per-corp / legacy NULL rows LIMIT-1 / env wins both sides), and the ledger pin is load-bearing. The one real gap: the PR's **headline ops behavior change** — "send to an assignee of a secret-less corp now fail-closes with an actionable error" — is an **untested guard**: I neutered it entirely and all 47 touched-suite tests plus 4550 unit tests stayed green.

---

## What I verified myself (not taken from the report)

### Environment / claims audit
- Head SHA on GitHub == ref I reviewed == worktree HEAD (`611eb3122`).
- `tsc --noEmit` clean in `packages/core-backend`.
- Fresh DB: full migration chain green, `zzzz20260709133000_add_integration_id_…` last; `\d` shows `integration_id uuid` + FK `… REFERENCES directory_integrations(id) ON DELETE SET NULL`.
- **3 real-DB suites: 25/25 pass** (deliveries 8, wrapper 10, executor 7) — matches the report exactly.
- **Full unit suite: 4550/4550 pass** (matches).
- **"Honesty note" premise-falsification is TRUE**: `origin/main` `automation-executor.ts:2641` signs with the global `resolveApprovalCardLinkSecret` (LIMIT-1) — #3910 only moved message credentials. The author's atomic sign+verify flip was the correct contingency; verify-only would have broken same-source.

### Two-point wiring (real, not claimed)
- All three integration suites are (a) in the `vitest.config.ts` exclude list (lines 75/86/92) so the no-DB job can't skip-green them, and (b) whole-file whitelisted in `.github/workflows/plugin-tests.yml` (executor+wrapper in the multitable real-DB step, deliveries in the approval real-DB step).
- Those steps run **inside the required `test (20.x)` matrix job** — I pulled this PR's actual job (`86261188686`) step list: both `Run multitable real-DB integration` and `Run approval real-DB integration` steps `conclusion: success`. So the new tests are CI-live on every PR, not invisible debt. New unit tests are auto-collected.
- `vitest.integration.config.ts` has `fileParallelism: false` + `maxConcurrency: 1` → the temporary `active`/`now()`-fresh integrations the tests insert into the shared CI DB cannot shadow concurrently-running files; all are deleted in `finally` (and the wrapper's corp B is backdated 1 min as crash insurance).

### Mutations I re-fired (evidence, not claims)
| # | Mutation (by me, in my worktree) | Result |
|---|---|---|
| MUT-A (=M2) | `verifyApprovalCardLinkToken` ignores `integrationId`, always global resolver | **RED×2**: wrapper db cross-corp test + unit cross-corp test. Restored → green |
| MUT-B (=M3) | Executor signs with global resolver instead of per-assignee | **RED×1**: executor "same-source signing" (decoy caught the LIMIT-1 pick: expected `deecf1…` got `84d6a6…`). Restored → green |
| MUT-C (=M1) | Drop `integration_id` from the INSERT column list | **RED×3**: deliveries persist test + both executor DT-R2 tests. Restored → green |
| MUT-D (mine) | `resolveApprovalCardLinkSecretForIntegration` falls back to the legacy LIMIT-1 pick when row/stored missing (cross-corp rescue) | **RED×1**: wrapper db test ("no stored secret fail-closes" assertions). The no-rescue guard is real-DB load-bearing. Restored → green |
| MUT-E (mine) | **Neuter the executor's missing-secret fail-close** (`if (!linkSecret && false)`) | **ALL GREEN** — 25/25 integration + 22/22 unit. Untested guard → finding F1. Restored → green |

The anti-shadowing fixtures (corp A freshest in the wrapper test; fresher decoy in the executor test) are genuinely load-bearing — MUT-A/B go red precisely because of them. The author's M2/M3 "first run green → hardened" narrative is consistent with what I observed.

### Known-landmine sweep (repo history)
- **Fixture shadowing**: addressed (above), verified by mutation.
- **Selection predicate coverage**: per-corp by-id predicate has a real second-tenant negative (corp A token vs corp B row) against real PG through the real wrapper call sites.
- **Tautology**: the executor same-source test recomputes the expected HMAC from the *stored corp secret it planted*, and asserts against the token extracted from the **fake DingTalk HTTP body** — independent oracle, not a re-implementation.
- **Wire-vs-fixture drift**: real-DB through real call sites; DingTalk HTTP faked (`gettoken`/`asyncsend_v2`) same as the existing suite — acceptable, flagged by the author.
- **Skip-when-unreachable**: `describeIfDatabase`-gated files are excluded from no-DB config and whole-file wired into the required job (verified executed).
- **package.json test chains / lockfile**: untouched.
- **PR body closing keywords**: none (`#3910`/`#3999`/`#3594` are plain mentions).
- **Scope**: 10 files, all in DT-R2 lane scope; no route/index/runtime smuggling. `automation-executor.ts` is the hottest DT file — landing must stay serialized with other lanes (author flagged; auto-merge NOT armed, verified `autoMergeRequest: null`).
- **Kysely insert sites**: none for this table (raw SQL only) — the non-optional `integration_id` addition to `DingTalkApprovalCardDeliveriesTable` can't break a typed insert.
- **Other sign/verify paths**: `verifyApprovalCardLinkToken` has exactly 2 src callers (both wrapper functions, both now pass `delivery.integration_id`); `insertDingTalkApprovalCardDelivery` has exactly 1 src caller (the executor, passes the id). No `interactive_card` writer exists yet — the B-2 "write it from day one" coordination note is accurate.

### Design-lock conformance (CFG lock 20260706 + one-tap lock #3594)
| Gate | Verdict | Evidence |
|---|---|---|
| Sign/verify must resolve the SAME secret (lock line 43) | **PASS (strengthened)** | Per-population same-source; MUT-A/B prove any divergence goes red |
| Secret encrypted-at-rest, decrypt server-side only, never echoed | PASS | New resolver reuses `decryptStoredSecretValue`; no new egress |
| Fail-closed on any miss, never a fallback secret | PASS | MUT-D proves no-LIMIT-1-rescue is load-bearing |
| Ledger-only resolution, zero engine bypass, no existence oracle | PASS | Both failure orders return identical `not_found`; row-before-verify costs one extra indexed row read on invalid tokens — no data egress difference |
| Migration additive/nullable, audit rows survive integration deletion | PASS | Real-DB SET-NULL test; `migration-replay` CI green |

---

## Findings

### F1 · P2 · Untested guard: the flagged fail-closed send semantics can be neutered with all tests green
`packages/core-backend/src/multitable/automation-executor.ts:2706-2714`
The PR body's headline ops change — *"multi-corp deploys where the assignee's corp has no stored `approvalCardLinkSecret` (and no env) now fail the SEND fail-closed with an actionable error"* — has zero test coverage. MUT-E (`if (!linkSecret && false)`) keeps 25/25 integration + 22/22 targeted unit + 4550/4550 full-unit green. Failure scenario: a future refactor of this hot file (B-2 interactive-card send is imminent in this very function family) drops or reorders the guard; the executor then signs with `''`, the card **sends successfully** (ledger `send_status='sent'`), and the assignee's tap dead-ends in `not_found` forever — the exact silent-degradation this PR exists to kill, with no failed action result to alert ops. (The guard predates this PR in global form and was untested then too, but this PR relocates it, re-scopes it per-corp, and advertises it as the behavior change — it should land proven.)
**Fix**: one appended executor-suite case: env unset, assignee corp with NO stored secret, plus a fresher decoy corp WITH a secret (no-rescue proof) → assert action `failed` with the actionable per-integration error, zero delivery rows for the instance, zero DingTalk HTTP calls. ~20 lines, same fixtures as the existing DT-R2 cases.

### F2 · P3 · New FK column has no index
`packages/core-backend/src/db/migrations/zzzz20260709133000_add_integration_id_to_dingtalk_approval_card_deliveries.ts:17-21`
`integration_id` gets a FK with `ON DELETE SET NULL` but no index (verified via `\d`: only pkey + two instance indexes). Every `DELETE FROM directory_integrations` seq-scans the deliveries table per deleted row to enforce SET NULL, and B-3's Stream callback / future per-corp ops queries will filter by this column. Low frequency today (admin op, modest table), so P3 — add `idx_dacd_integration` here or in the B-2/B-3 rung that first queries by it.

### F3 · NIT · Stale module-header invariant comment
`packages/core-backend/src/integrations/dingtalk/approval-card-config.ts:10-14`
The file header still states same-source is "guaranteed by 'env first, else the one stored dingtalk integration'" — no longer the whole story (new rows are per-integration). The new function's own docstring is correct; update the header's SAME-SOURCE bullet to the per-population wording to keep the next reader from re-deriving the old model.

### Noted, not findings (correctly deferred / bounded)
- Legacy `integration_id=NULL` rows remain LIMIT-1-unstable under multi-corp until they drain — inherent to the no-backfill design (a backfill can't know the sign-time corp; guessing would be worse). Correct call.
- Rolling-deploy transient (new-signed link vs old verify code) fail-closes; reasoned, untestable without a two-version harness.
- Verify-order change (row load before verify) costs one indexed row read on invalid tokens; identical `not_found` both orders; no new oracle.
- `down()` not run against populated data — trivial `DROP COLUMN IF EXISTS`, same pattern as the sibling migration; `migration-replay` CI green.

## Disposition
- No P1. F1 is the only substantive ask; it is additive-test-only, cannot regress runtime, and sits in the file lane D is also appending to — **fix pre-merge in this PR** (preferred, keeps the flagged semantics proven at birth) or as an immediate fast-follow if landing order with lane D forces it.
- Cleanup done on my side: mutation edits all restored (`git status` clean at `611eb3122`), review DB dropped, review worktree removed.

zensgit:
Double-mutation window protocol executed (owner-mandated for this PR's combination risk): rebased onto post-#4009 main (clean, 3 commits survived); fresh-DB baselines — deliveries+action suites 17/17, wrapper verify suite 10/10, tsc clean; **both load-bearing mutations replayed RED on the rebased head**: (1) H03 executor decrypt→passthrough ⇒ RED ×1; (2) R2 verify ignores the delivery row's integration (forced LIMIT-1 fallback) ⇒ wrapper suite RED ×1. Honest note: the first M-R2 probe against the deliveries suite stayed green — the verify-preference guard lives in approval-card-delivery-wrapper.db.test.ts, where it is load-bearing. Armed alone per single-window discipline.

PR #4018, opened 2026-07-10
feat(multitable): History Center batch-detail click-through to the record drawer (PR-C)

zensgit:
**Stacked on #4012 (PR-B) → #4007 (PR-A).** Land bottom-up with base retargeting per stacked-PR discipline.

R9 UX-parity audit finding (d): no way to jump from a History Center change row to the record itself.

### What
Record labels on non-delete change rows become buttons → `('open-record', {sheetId, recordId})` emitted by `HistoryBatchChangesList`, relayed verbatim by `HistoryCenterModal`, handled by the workbench **exactly like the Notification Center's click-to-locate** (`onNotificationNavigate` shape): `onSelectSheet` first for cross-sheet batches (declined unsaved-changes discard aborts — never locates in the wrong sheet), close the modal, then `resolveDeepLink` — loaded row → drawer; off-page → `getRecord` fetch; gone/denied → the existing `Record not found` toast. Delete rows stay plain text. **Zero new endpoints, zero new i18n labels** — reuses shipped deep-link machinery.

### Verification
- Modal-level: BUTTON chip emits exactly once with payload; delete row renders SPAN and cannot emit.
- Workbench-level (real mount): in-page hit (no fetch, no toast) · off-page `getRecord` success (no toast) · `getRecord` failure → not-found toast · cross-sheet → `selectSheet('sheet_2')` before locating.
- **Mutation-verified**: neutering the emit → exactly 5 red across both layers.
- Neighbors 7 files / 54 tests green; `vue-tsc` clean.
- Spec-infra fix folded in: jsdom shares one window per spec file, and the workbench mirrors selection into `#recordId=…` and deep-link-bootstraps FROM it on mount — a prior test's hash auto-fired `resolveDeepLink` on the next mount (phantom double-toast). `beforeEach` now clears the hash.

### Not covered
- Restored-after-delete records still render as plain text in old delete rows (chip enablement is action-shaped, not liveness-checked).
- The pinned-banner list relays the same emit (shared component) but has no dedicated spec case.

### DIFF
--- a/apps/web/src/multitable/components/HistoryBatchChangesList.vue
+++ b/apps/web/src/multitable/components/HistoryBatchChangesList.vue
@@ -11,7 +11,15 @@
     <li v-for="(c, i) in changes" :key="`${c.recordId}-${i}`" class="meta-hist__change" data-test="hist-change">
       <div class="meta-hist__change-summary">
         <span class="meta-hist__change-action" :data-action="c.action">{{ actionLabel(c.action) }}</span>
-        <span class="meta-hist__change-rec" :title="c.recordId" data-test="hist-rec-label">{{ recordDisplay(c) }}</span>
+        <button
+          v-if="c.action !== 'delete'"
+          type="button"
+          class="meta-hist__change-rec meta-hist__change-rec--link"
+          :title="c.recordId"
+          data-test="hist-rec-label"
+          @click="emit('open-record', { sheetId: c.sheetId, recordId: c.recordId })"
+        >{{ recordDisplay(c) }}</button>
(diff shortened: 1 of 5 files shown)


zensgit:
Rebased onto the updated PR-B (owner P1 layer-2∩layer-3 + P2 link per-side fixes inherited). Full neighbor sweep 7 files green + vue-tsc clean on the rebased stack. No PR-C-side changes — per the owner's review, click-through wiring itself had no new blockers. Will retarget base to main after upstream lands so full CI runs here.

PR #4046, opened 2026-07-10
feat(dingtalk): unified transport resilience — tiered retry/backoff/rate-limit at the client seam (roadmap §7.2)

zensgit:
## Phase A — What / why

Roadmap §7.2 (docs/development/dingtalk-sync-integrated-roadmap-20260708.md, never ticketed — section text is the spec): `packages/core-backend/src/integrations/dingtalk/client.ts` had ZERO retry/backoff/rate-limit handling. This adds ONE shared transport seam — `packages/core-backend/src/integrations/dingtalk/transport.ts` — that every DingTalk outbound call in client.ts funnels through. No per-call-site retry sprinkling.

**Mid-lane design revision (owner, supersedes the brief's constraint #2):** classification is by BUSINESS SEMANTICS (`kind: 'read' | 'exchange' | 'send'`), never by HTTP verb, because a fetch timeout/disconnect only proves the client saw no response — not that DingTalk didn't execute the request. The first commit (0bf550944) implemented the original two-class matrix; the second (906e88a9a) reworks it to the ratified tiering.

### The tier matrix (explicit at all 10 call sites in client.ts)

| Tier | Call sites | Retries |
|---|---|---|
| `'read'` | gettoken, contact/users/me, department listsub/get, user list/get | network error, HTTP 429, selected transient 5xx (500/502/503/504), flow-control errcodes in HTTP-200 envelopes |
| `'exchange'` | OAuth code exchange, E1 container authCode | **never** — single-use codes; ambiguous retry burns/double-redeems them |
| `'send'` | asyncsend_v2 markdown + action_card | **never auto-resent**; network/timeout/5xx failures rethrown marked `outcomeUnknown: true` (+ exported `isDingTalkOutcomeUnknown` guard) so ledger writers can record `outcome_unknown`. A lost response has no `task_id`, so DingTalk's async-send result query cannot serve as an idempotency key (stated in the code comment at the classification). 429-on-send auto-retry deliberately NOT enabled in this slice — needs endpoint-contract confirmation first |
| (omitted) | — | defaults to `'exchange'`, the most conservative tier |

### Mechanics

- Exponential backoff + **full jitter** (`random() * min(cap, base * 2^(attempt-1))`); `Retry-After` honored (seconds or HTTP-date), clamped to the cap so a huge header can't stall a sync worker.
- Env knobs `DINGTALK_TRANSPORT_MAX_ATTEMPTS` / `_BACKOFF_BASE_MS` / `_BACKOFF_CAP_MS` (defaults 3 / 300ms / 5000ms); invalid values fall back to defaults (negative tests included). Documented in `.env.example`.
- **DT-HARDEN-06 preserved and composed per attempt**: each attempt gets its OWN fresh `AbortSignal.timeout`; a caller-supplied overall signal (`options.signal` / `init.signal`, composed via `AbortSignal.any`) aborts the in-flight attempt AND any backoff sleep immediately. Per-attempt timeouts are never retried (wall-time bound preserved).
- Flow-control errcode set anchored on DingTalk's documented `90018` family / `-1` system-busy PLUS parity with what the repo already treats as retryable (`AttendanceNotificationDeliveryWorker.isRetryableDingTalkErrorCode`: 408/429/5xx-echo/50001) — not invented.
- **Zero error-shape change**: exhausted/non-retryable failures rethrow the ORIGINAL `DingTalkRequestError` / `DingTalkBusinessError` / `DingTalkTimeoutError` / raw network error. Callers pattern-match these (routes/auth.ts:1295ff, automation-executor.ts:1850/3125, AttendanceNotificationDeliveryWorker classifiers) — verified by reading them; error classes physically moved to transport.ts and re-exported from client.ts (single class definition, `instanceof` identity preserved).
- Observability: exactly one `warn` per retry (attempt count + reason + tier + delay); **no URLs logged** (gettoken query carries the app secret). No new tables.
- **In-process rate limiting is BEST-EFFORT only** — replicas do not share a quota; multi-replica shared throttling (per app/corp, Redis or equivalent) is out of scope, listed as follow-up.

## Test evidence

- New `tests/unit/dingtalk-transport-resilience.test.ts`: 28 tests (read retry/backoff/Retry-After/flow-control/501-boundary/exhaustion shape; send no-resend + outcome-unknown markers; exchange no-retry; conservative default; abort composition ×3; env knobs ×4; pure helpers ×4). Mock fetch at the seam; real timing asserted with stubbed `Math.random` (backoff ≥280ms, Retry-After ≥950ms, abort exits <1500ms of a 3000ms backoff).
- H06 suite `tests/unit/dingtalk-request-timeout.test.ts`: 5/5 green unchanged (one stale implementation comment updated to point at the transport seam).
- Full core-backend unit suite: **346 files / 4670 tests, all green** (`CI=true npx vitest run tests/unit`).
- `npx tsc --noEmit -p tsconfig.json`: clean.

## Mutation table (protocol: real impl committed first; each mutation applied → run → recorded → restored via git checkout)

| # | Mutation | Result |
|---|---|---|
| M1 | Retry loop removed (always throw on first failure) | **RED ×8**: 429-retry, Retry-After, flow-control-retry, network-retry, exhausted-attempts(3), abort-mid-backoff, per-attempt freshness, maxAttempts=2 bound |
| M2 | Sends classified like reads (auto-resend flip) | **RED ×4**: send network-error / 5xx / 429 / flow-control all detect extra fetch calls |
| M3 | Retry-After ignored | **RED ×2**: end-to-end honor test (elapsed <950ms) + pure-helper test |
| M4 | Per-attempt signal hoisted out of the loop (composed once, reused) | **RED ×1**: freshness test (`seenAborted` `[false,true]`); **H06 suite stayed GREEN 5/5** — proves the pre-existing tests alone would NOT catch this and the new test is load-bearing |
| M5 | Backoff sleep ignores the abort signal | **RED ×1**: abort-mid-backoff (elapsed ~3000ms > 1500ms bound) |
| M6 | Flow-control errcode not retried on reads | **RED ×1** |
| M7 | Exchange tier retries like reads | **RED ×4**: 3 exchange tests + conservative-default test |
| M8 | Unclassified default flips `'exchange'` → `'read'` | **RED ×1**: conservative-default test |
| M9 | Outcome-unknown marking dropped | **RED ×3**: send network / 5xx / timeout marker assertions |

## Honest gaps (not proven / deliberately out of scope)

1. ~~**Ledger writers are not wired to `isDingTalkOutcomeUnknown` in this PR**~~ — **CLOSED by Phase B below**: all three send ledgers (approval-card, person-delivery telemetry, attendance outbox) now consume the marker and record `outcome_unknown` without resending.
2. §7.2 acceptance "repeated flow-control failures produce clear run stats and alerts" is only partially served here (warn logs + preserved error shapes feeding existing run stats); run-stat/alert enrichment belongs to §7.4 / the directory-sync lane.
3. 429-on-send auto-retry NOT enabled — pending endpoint-contract confirmation (owner instruction).
4. Retry-After beyond the backoff cap is clamped (bounded stall) — a deviation from a strict "honor" reading, documented in code.
5. The 408/429/5xx/50001 errcode-echo portion of the flow-control set is codebase-parity-anchored (delivery worker), not DingTalk-doc-anchored; only 90018/-1 are the documented codes.
6. `AbortSignal.any` pre-Node-20.3 fallback (caller signal wins, no per-attempt timeout) is structurally present but untested — CI runs Node 20.x where `any` exists.
7. Timeouts are never retried on ANY tier (including idempotent reads) — conservative reading of the spec; preserves the H06 wall-time bound. A read-tier timeout retry would be safe but multiplies hung-endpoint wall time 3×.
8. Unit tests mock fetch at the seam; no real DingTalk endpoint or end-to-end flaky-sync exercise. Directory sync gets read-tier retries transparently via `fetchDingTalkAppAccessToken`/list/get calls (directory-sync.ts untouched per lane boundary).

## Behavior changes flagged for ops

- Reads now retried up to 3× with jittered backoff (env-tunable): a persistently failing endpoint costs up to ~2 extra attempts + ≤10s backoff per call in sync paths.
- Send failures on network/timeout/5xx now carry an enumerable `outcomeUnknown: true` property on the (otherwise unchanged) error — visible in serialized error dumps.
- A caller-supplied signal now COMPOSES with the per-attempt timeout via `AbortSignal.any` (previously a caller signal replaced the timeout). No in-repo caller passes one (verified; noted in the H06 test).

## Follow-up candidates: raw `fetch` to DingTalk OUTSIDE client.ts (NOT converted here, per lane boundary)

- `packages/core-backend/src/multitable/dingtalk-group-destination-service.ts` (~L321) — group-robot webhook send, own AbortController/timeout.
- `packages/core-backend/src/directory/directory-sync-alert-delivery.ts` (~L128) — sync-alert robot webhook delivery.
- `packages/core-backend/src/services/NotificationService.ts` (~L135) — robot webhook recipients path.
- (Generic, not DingTalk-specific: automation-executor `send_webhook` has its own retry loop.)

---

## Phase B — the outcome-unknown marker is CONSUMED by the ledgers (owner blockers closed)

Phase A defined the send-tier semantics (uncertain outcomes throw marked `isDingTalkOutcomeUnknown`, never resent at the transport). **Phase B closes the owner's blocker: the marker is now consumed by every send ledger, so the system stops auto-resending on ambiguity — and a malformed HTTP-2xx response can no longer masquerade as success.** Together, Phase A + Phase B close both owner blockers on this PR.

**What reconciliation IS (stated plainly):** a delivery recorded `outcome_unknown` is resolved **manually / by ops** via the distinct state + persisted error text. DingTalk's async-send result query (by `task_id`) can confirm a send **only when a task_id exists** — a lost response carries none, so the result query can never serve as an idempotency key for automatic retry. Everything in-process (retry budgets, backoff, rate limiting) remains **best-effort**; replicas do not share state.

### B1. Malformed-2xx guard (P1, same seam)

Pre-fix, `readJson` swallowed a body `SyntaxError` into `null` and the ok-path normalized `null → {} → SUCCESS`. Now an HTTP-2xx whose body is HTML / empty / truncated JSON / a bare primitive / a top-level array — or, for `envelope: 'oapi'`, a JSON object **missing the `errcode`/`code` envelope discriminator** — throws the new typed `DingTalkMalformedResponseError` (message carries **no body content**). Tier behaviors kept separate (owner guardrail), each pinned by its own test:

| Tier | Malformed 2xx |
|---|---|
| `read` | retryable **within the existing attempt budget** (like 5xx-on-read); recovery on a later good attempt pinned |
| `exchange` | **never** retried — exactly one fetch |
| `send` | **never** resent — exactly one fetch, error marked `isDingTalkOutcomeUnknown` |

`envelope: 'none'` (v1.0) endpoints have no discriminator requirement — object-without-errcode is still success there (pinned; the real v1.0 success shapes carry no errcode). Non-ok statuses keep their original `DingTalkRequestError` shape (`responseBody: null`) — pinned.

### B2. Three persistence surfaces (status-vocabulary decision: `outcome_unknown` in all three)

New migration `zzzz20260710150000_add_outcome_unknown_delivery_status.ts` widens all three CHECK constraints (each guarded by `checkTableExists` for partial-schema test harnesses). **down() maps live `outcome_unknown` rows → `'failed'` BEFORE re-tightening** — verified end-to-end against real Postgres (seed rows on all 3 tables → down() → all `'failed'`, constraint rejects `outcome_unknown` again (23514) → up() re-accepts).

1. **`dingtalk_approval_card_deliveries.send_status`** (was CHECK `pending|sent|failed`, merged on main via #3647 → new migration required): the card action's catch consumes the marker → new helper `markDingTalkApprovalCardDeliverySendOutcomeUnknown` (same pending-only guard) records the DISTINCT state + `send_error`; execution-record output carries `deliveryOutcomeUnknown: true`. Definite rejections (errcode envelopes, 4xx) still record `failed` — pinned.
2. **`dingtalk_person_deliveries.status`** (was CHECK `success|failed|skipped`): BOTH writers consume the marker — the batch person-message action records `outcome_unknown` for **exactly the in-flight batch** (batches never attempted have a KNOWN outcome → stay `failed`; earlier successful batches keep `success`), and the rule-creator failure-alert path records it for the alert send. Step output carries `deliveryOutcomeUnknown: true` + `outcomeUnknownRecipientCount`.
3. **`attendance_notification_deliveries.status`** (was CHECK `pending|sending|sent|retrying|failed|skipped`; `'skipped'` precedent from #3920): `classifyDingTalkSendError` checks the marker FIRST → `{ retryable: false, outcomeUnknown: true }` → the worker terminates the row as `outcome_unknown` on the first such attempt. **No retry burn, never re-claimed (claim predicate only takes `pending`/`retrying`/expired `sending`), never resent.** Quiet-hours gate and lease/CAS semantics untouched (the change sits strictly in the per-row post-send classification; `markOutcomeUnknown` uses the identical CAS guard as `markFailed`). Ordinary transient errors still retry — pinned. **WeCom/email classifiers untouched** (their transports never emit the marker).

### B3. Card `outcome_unknown` × HMAC callback (fail-open for valid tokens — surgically)

An `outcome_unknown` card MAY in fact have been delivered — and the deep-link HMAC token only ever existed inside the delivered card, so a valid callback is itself proof of delivery. The possibly-delivered set is widened by **exactly** `('sent', 'outcome_unknown')` in two places: `buildSummary().actionable` and the atomic acted-claim SQL. Nothing else changed — HMAC verify, instance-state checks, supersede handling, one-shot claim semantics all untouched. Pinned both directions: `outcome_unknown` + valid token + pending instance → approve proceeds and claims (send-time truth preserved on the audit row); `pending`/`failed` stay stale; forged token still `not_found`.

### Reader-compatibility audit (three consumer classes, file:line in the PR comment)

- **#4013 retention sweep** (landed mid-lane; rebased onto it): targets `dingtalk_group_deliveries` ONLY, deletes purely by `created_at < cutoff` — no status filter, touches none of the three Phase B tables.
- **Stats/aggregations**: no SQL aggregation on any of the three status columns exists in `src/` (worker `runBatch` counters extended with an optional `outcomeUnknown` count, present only when > 0 — same pattern as `lostLease`, so existing `toEqual` result assertions stay exact).
- **List/UI readers**: person-delivery endpoint passes the state through verbatim (folding it to `failed` is now a RED test); viewer gets a distinct badge + filter option + typed i18n label (`结果未知`/`outcome unknown`); the mobile decision page keys on server `actionable` (buttons render for actionable outcome_unknown; the "未成功投递" stale message is now reserved for pending/failed, where non-delivery is KNOWN).

### Phase B test evidence (fresh DB `lane_dtphaseb_*`, full migration stack)

- core-backend unit: **358 files / 4866 tests green** (`CI=true npx vitest run tests/unit`), incl. 8 new malformed-2xx transport tests, 5 channel-classification tests, 3 executor tests, reader pass-through pin.
- Real-DB (6 suites, 52 tests green): attendance-notification-deliveries (+outcome_unknown terminal/no-reclaim/no-resend), automation-dingtalk-approval-card-action (+transport-level marker → ledger + output flag + single asyncsend; person-deliveries CHECK both directions), approval-card-delivery-wrapper.db (+actionable/claim widening both directions), dingtalk-approval-card-deliveries.db (+helper guard + claimability + CHECK), dingtalk-person-message-integration-scoping.db, automation-approval-task-created-trigger.
- Web: 232 tests green across the 6 touched-surface spec files; `tsc` + `vue-tsc` clean.
- gettoken unit fixtures aligned to the real oapi wire contract (`errcode: 0`) — wire-vs-fixture drift surfaced by the new missing-errcode guard.

### Phase B mutation table (protocol: impl committed first; each mutation applied → run → restored)

| # | Mutation | Result |
|---|---|---|
| B-M1 | Restore `null → {}` ok-path normalization | **RED ×6** (send/read/exchange malformed suites + array + recovery) |
| B-M2 | Drop ONLY the oapi missing-errcode requirement | **RED ×3** (the JSON-without-errcode case per tier) |
| B-M3 | Exchange tier retries malformed responses | **RED ×2** (exchange single-fetch + 'none'-envelope test) |
| B-M4 | Read tier gives up immediately on malformed | **RED ×2** (budget-retry + recovery tests) |
| B-M5 | Executor maps outcome-unknown → plain `failed` (no flag) | **RED ×4** (card real-DB) |
| B-M6 | Send-tier network error retryable ("make it retry") | **RED ×4** card real-DB (single-asyncsend pin) **+ RED ×1** transport unit |
| B-M7 | Worker classifier treats outcome-unknown as retryable (the resend hazard) | **RED ×1** worker real-DB **+ RED ×2** channel unit |
| B-M8 | Worker maps outcome_unknown terminal → `failed` | **RED ×1** worker real-DB (distinct-state + counter pins) |
| B-M9 | Claim SQL back to `send_status = 'sent'` | **RED ×2** (wrapper + ledger claimability) |
| B-M10 | `actionable` back to `send_status === 'sent'` | **RED ×1** (wrapper actionable test) |
| B-M11 | Person-message marks ALL unsent recipients outcome_unknown (in-flight scoping dropped) | **RED ×1** (3-batch test: batch-3 must stay `failed`) |
| B-M12 | Endpoint reader folds outcome_unknown → `failed` | **RED ×1** (pass-through pin) |

### Behavior changes flagged for ops (Phase B)

- A 200-with-garbage-body DingTalk response now FAILS (typed `DingTalkMalformedResponseError`) instead of silently succeeding — directory-sync reads retry it up to the budget then surface a failed run (previously: silent `{}` success with confusing downstream field errors); asyncsend "successes" without a parseable envelope no longer record `sent` with no task_id.
- New terminal state `outcome_unknown` appears in the three delivery tables + the person-delivery viewer/endpoint + worker log line (`outcomeUnknown=N`). These rows are NOT retried; resolution is manual (resend by hand only after checking with the recipient / the DingTalk backend, or via the task_id result query when a task_id exists).
- Approval cards recorded `outcome_unknown` remain operable via their deep link (valid HMAC = delivery proof).

### Honest gaps (Phase B)

1. No automated reconciliation job (deliberate — owner doctrine): `outcome_unknown` rows stay until ops resolves them; no retention/aging policy for them yet (the #4013 sweep covers group deliveries only).
2. The worker terminates outcome-unknown on the FIRST such attempt even if earlier attempts were ordinary retryable failures — attempt history is preserved in `attempt_count`/`last_error`, but there is no per-attempt error trail.
3. `getDingTalkUserInfoByAuthCode` (oapi exchange) malformed-2xx is covered; the v1.0 `exchangeCodeForUserAccessToken` unparseable-body path is covered via the raw-transport 'none'-envelope test, not through that specific client function.
4. The decision page's outcome_unknown behavior is pinned at component level (server `actionable` drives it); no e2e over a real DingTalk callback.
5. Migration down() is verified by a scripted real-DB run (documented in the PR comment), not a committed test — the repo has no migration-down test convention to extend.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -51,7 +51,9 @@ JWT_EXPIRES_IN=2h
 # business semantics: safe reads retry on network errors / HTTP 429 / transient
 # 5xx / flow-control errcodes; one-shot code exchanges NEVER retry (single-use
 # codes); side-effect sends NEVER auto-resend (uncertain outcomes are surfaced
-# as outcome-unknown instead — no duplicate notifications).
+# as outcome-unknown instead — no duplicates introduced at the transport layer;
+# the attendance outbox may still resend on ambiguous failures until it consumes
+# the outcome-unknown marker — follow-up).
 # Exponential backoff + full jitter, honoring Retry-After (clamped to the cap).
 # In-process/best-effort only — replicas do not share a quota.
 # Invalid values fall back to the defaults shown.
--- a/.env.example
+++ b/.env.example
@@ -46,6 +46,19 @@ JWT_EXPIRES_IN=2h
 # Optional multi-instance fleet coordination (defaults to single-process leader):
 # ENABLE_DINGTALK_GROUP_DELIVERY_RETENTION_LEADER_LOCK=false
 #
+# ---- DingTalk transport resilience (roadmap §7.2) -----------------------------
+# Shared retry/backoff for every DingTalk API call in the core client: idempotent
+# reads retry on network errors / HTTP 429 / HTTP 5xx / flow-control errcodes;
+# sends and one-shot code exchanges retry ONLY on network-error-before-response
+# and HTTP 429 (never on an ambiguous 5xx — no duplicate notifications).
+# Exponential backoff + full jitter, honoring Retry-After (clamped to the cap).
+# Invalid values fall back to the defaults shown.
+# DINGTALK_TRANSPORT_MAX_ATTEMPTS=3
+# DINGTALK_TRANSPORT_BACKOFF_BASE_MS=300
+# DINGTALK_TRANSPORT_BACKOFF_CAP_MS=5000
+# Per-attempt request timeout in ms (DT-HARDEN-06; default 10000):
+# DINGTALK_REQUEST_TIMEOUT_MS=10000
+#
 # ---- Approval one-tap card (A-5 verification; design-lock #3594) --------------
 # APP_KEY/SECRET/AGENT_ID above ARE the same corp-app used to SEND the approval
 # action_card (asyncsend_v2). The card decision deep link additionally needs:
--- a/.env.example
+++ b/.env.example
@@ -52,8 +52,8 @@ JWT_EXPIRES_IN=2h
 # 5xx / flow-control errcodes; one-shot code exchanges NEVER retry (single-use
 # codes); side-effect sends NEVER auto-resend (uncertain outcomes are surfaced
 # as outcome-unknown instead — no duplicates introduced at the transport layer;
-# the attendance outbox may still resend on ambiguous failures until it consumes
-# the outcome-unknown marker — follow-up).
+# Phase B: the approval-card ledger, person-delivery telemetry, and attendance
+# outbox all consume the marker and record 'outcome_unknown' without resending).
 # Exponential backoff + full jitter, honoring Retry-After (clamped to the cap).
 # In-process/best-effort only — replicas do not share a quota.
 # Invalid values fall back to the defaults shown.
--- a/.env.example
+++ b/.env.example
@@ -47,11 +47,13 @@ JWT_EXPIRES_IN=2h
 # ENABLE_DINGTALK_GROUP_DELIVERY_RETENTION_LEADER_LOCK=false
 #
 # ---- DingTalk transport resilience (roadmap §7.2) -----------------------------
-# Shared retry/backoff for every DingTalk API call in the core client: idempotent
-# reads retry on network errors / HTTP 429 / HTTP 5xx / flow-control errcodes;
-# sends and one-shot code exchanges retry ONLY on network-error-before-response
-# and HTTP 429 (never on an ambiguous 5xx — no duplicate notifications).
+# Shared retry/backoff for every DingTalk API call in the core client, tiered by
+# business semantics: safe reads retry on network errors / HTTP 429 / transient
+# 5xx / flow-control errcodes; one-shot code exchanges NEVER retry (single-use
+# codes); side-effect sends NEVER auto-resend (uncertain outcomes are surfaced
+# as outcome-unknown instead — no duplicate notifications).
 # Exponential backoff + full jitter, honoring Retry-After (clamped to the cap).
+# In-process/best-effort only — replicas do not share a quota.
 # Invalid values fall back to the defaults shown.
 # DINGTALK_TRANSPORT_MAX_ATTEMPTS=3
 # DINGTALK_TRANSPORT_BACKOFF_BASE_MS=300
--- a/apps/web/src/approvals/cardDecision.ts
+++ b/apps/web/src/approvals/cardDecision.ts
@@ -10,7 +10,8 @@ import { apiFetch } from '../utils/api'
 export interface ApprovalCardSummary {
   deliveryId: string
   cardState: 'sent' | 'acted' | 'superseded' | 'expired' | 'revoked'
-  sendStatus: 'pending' | 'sent' | 'failed'
+  /** 'outcome_unknown' (PR #4046 Phase B): send attempted, response lost — the card MAY be delivered; the server keeps it actionable. */
+  sendStatus: 'pending' | 'sent' | 'failed' | 'outcome_unknown'
   nodeKey: string
   recipientUserId: string
   viewerIsRecipient: boolean
--- a/apps/web/src/multitable/components/MetaAutomationPersonDeliveryViewer.vue
+++ b/apps/web/src/multitable/components/MetaAutomationPersonDeliveryViewer.vue
@@ -13,6 +13,7 @@
             <option value="success">{{ automationStatusLabel('success', isZh) }}</option>
             <option value="failed">{{ automationStatusLabel('failed', isZh) }}</option>
             <option value="skipped">{{ automationLabel('delivery.statusSkippedUnbound', isZh) }}</option>
+            <option value="outcome_unknown">{{ automationStatusLabel('outcome_unknown', isZh) }}</option>
           </select>
           <MtButton class="meta-person-delivery__btn" data-action="refresh" :disabled="loading" @click="loadData">{{ automationLabel('log.refresh', isZh) }}</MtButton>
         </div>
@@ -89,12 +90,15 @@ const filteredDeliveries = computed(() => {
   return deliveries.value.filter((delivery) => deliveryStatus(delivery) === statusFilter.value)
 })
 
-type DeliveryStatus = 'success' | 'failed' | 'skipped'
+// 'outcome_unknown' (PR #4046 Phase B): the send was attempted and the response lost — the
+// message MAY have been delivered. Distinct badge/filter value; folding it into 'failed' would
+// erase the reconciliation signal.
+type DeliveryStatus = 'success' | 'failed' | 'skipped' | 'outcome_unknown'
 
 const UNBOUND_DINGTALK_REASON = 'DingTalk account is not linked or user is inactive'
 
 function deliveryStatus(delivery: DingTalkPersonDelivery): DeliveryStatus {
-  if (delivery.status === 'success' || delivery.status === 'failed' || delivery.status === 'skipped') return delivery.status
+  if (delivery.status === 'success' || delivery.status === 'failed' || delivery.status === 'skipped' || delivery.status === 'outcome_unknown') return delivery.status
   if (delivery.success) return 'success'
   if (!delivery.dingtalkUserId && delivery.errorMessage === UNBOUND_DINGTALK_REASON) return 'skipped'
   return 'failed'
@@ -285,6 +289,11 @@ watch(
 .meta-person-delivery__badge--skipped {
   background: #fef3c7;
   color: #b45309;
+}
+
+.meta-person-delivery__badge--outcome_unknown {
+  background: #e2e8f0;
+  color: #475569;
 }
 
 .meta-person-delivery__time {
--- a/apps/web/src/multitable/types.ts
+++ b/apps/web/src/multitable/types.ts
@@ -1507,7 +1507,8 @@ export interface DingTalkPersonDelivery {
   subject: string
   content: string
   success: boolean
-  status?: 'success' | 'failed' | 'skipped'
+  /** `outcome_unknown` (PR #4046 Phase B): send attempted, response lost — maybe delivered, never auto-resent. */
+  status?: 'success' | 'failed' | 'skipped' | 'outcome_unknown'
   httpStatus?: number
   responseBody?: string
   errorMessage?: string
--- a/apps/web/src/multitable/utils/meta-automation-labels.ts
+++ b/apps/web/src/multitable/utils/meta-automation-labels.ts
@@ -288,6 +288,7 @@ export type AutomationLabelKey =
   | 'status.success'
   | 'status.failed'
   | 'status.skipped'
+  | 'status.outcomeUnknown'
   | 'status.running'
   | 'status.resolved'
   | 'status.queued'
@@ -560,6 +561,7 @@ export const AUTOMATION_LABEL_KEYS: readonly AutomationLabelKey[] = [
   'status.success',
   'status.failed',
   'status.skipped',
+  'status.outcomeUnknown',
   'status.running',
   'status.resolved',
   'status.queued',
@@ -881,6 +883,7 @@ const LABELS: Record<AutomationLabelKey, { en: string; zh: string }> = {
   'status.success': { en: 'success', zh: '成功' },
   'status.failed': { en: 'failed', zh: '失败' },
   'status.skipped': { en: 'skipped', zh: '跳过' },
+  'status.outcomeUnknown': { en: 'outcome unknown', zh: '结果未知' },
   'status.running': { en: 'running', zh: '运行中' },
   'status.resolved': { en: 'resolved', zh: '已完成' },
   'status.queued': { en: 'queued', zh: '排队中' },
@@ -936,6 +939,7 @@ export function automationStatusLabel(status: AutomationStatus | UnknownAutomati
   if (status === 'success') return automationLabel('status.success', isZh)
   if (status === 'failed') return automationLabel('status.failed', isZh)
   if (status === 'skipped') return automationLabel('status.skipped', isZh)
+  if (status === 'outcome_unknown') return automationLabel('status.outcomeUnknown', isZh)
   if (status === 'running') return automationLabel('status.running', isZh)
   if (status === 'resolved') return automationLabel('status.resolved', isZh)
   if (status === 'queued') return automationLabel('status.queued', isZh)
--- a/apps/web/src/views/approval/ApprovalCardDecisionView.vue
+++ b/apps/web/src/views/approval/ApprovalCardDecisionView.vue
@@ -136,7 +136,11 @@ const staleTitle = computed(() => {
     return `该待办已处理（${s.actedAction === 'approve' ? '同意' : s.actedAction === 'reject' ? '驳回' : s.actedAction ?? ''}）。`
   }
   if (s.approval.status !== 'pending') return '该审批已办结，无需处理。'
-  if (s.sendStatus !== 'sent') return '该卡片未成功投递，无法在此处理。'
+  // outcome_unknown deliberately NOT here (PR #4046 Phase B): such a card MAY have been
+  // delivered, so while its instance is pending the server marks it actionable and this stale
+  // branch never renders; if it is non-actionable for another reason the generic 已流转 message
+  // below is the accurate one — not "未成功投递".
+  if (s.sendStatus === 'pending' || s.sendStatus === 'failed') return '该卡片未成功投递，无法在此处理。'
   return '该待办已流转（转办/新一轮），此卡片不再有效。'
 })
 
--- a/apps/web/tests/approvalCardDecisionView.spec.ts
+++ b/apps/web/tests/approvalCardDecisionView.spec.ts
@@ -132,6 +132,25 @@ describe('ApprovalCardDecisionView (A-3)', () => {
     expect((container!.querySelector('[data-testid="card-decision-reject"]') as HTMLButtonElement).disabled).toBe(false)
   })
 
+  it('PR #4046 Phase B: an actionable outcome_unknown card renders live buttons; a non-actionable one gets 已流转 — never 未成功投递', async () => {
+    // The valid deep-link token only ever existed inside the delivered card, so an
+    // outcome_unknown delivery whose instance is pending stays actionable server-side.
+    apiFetchMock.mockResolvedValueOnce(jsonResponse({ ok: true, data: summaryFixture({ sendStatus: 'outcome_unknown' }) }))
+    await mountView()
+    expect(container!.querySelector('[data-testid="card-decision-approve"]')).toBeTruthy()
+    expect(container!.querySelector('[data-testid="card-decision-stale"]')).toBeNull()
+    app!.unmount(); container!.remove()
+
+    // Non-actionable for ANOTHER reason (superseded): the accurate message is 已流转 —
+    // "未成功投递" is reserved for pending/failed sends, where non-delivery is KNOWN.
+    apiFetchMock.mockReset()
+    apiFetchMock.mockResolvedValueOnce(jsonResponse({ ok: true, data: summaryFixture({ sendStatus: 'outcome_unknown', cardState: 'superseded', actionable: false }) }))
+    await mountView()
+    const stale = container!.querySelector('[data-testid="card-decision-stale"]')
+    expect(stale?.getAttribute('data-title')).toContain('已流转')
+    expect(stale?.getAttribute('data-title')).not.toContain('未成功投递')
+  })
+
   it('approve submits to the card-delivery endpoint and renders the terminal state', async () => {
     apiFetchMock.mockResolvedValueOnce(jsonResponse({ ok: true, data: summaryFixture() }))
     await mountView()
--- a/apps/web/tests/meta-person-delivery-viewer-migration.spec.ts
+++ b/apps/web/tests/meta-person-delivery-viewer-migration.spec.ts
@@ -40,4 +40,30 @@ describe('MetaAutomationPersonDeliveryViewer — MtButton migration (UI-P2-1c)',
     expect(get).toHaveBeenCalledTimes(1)
     expect(get).toHaveBeenCalledWith('s1', 'r1', 50)
   })
+
+  it('PR #4046 Phase B: an outcome_unknown delivery renders its own badge + reason (not folded into failed) and the filter offers the state', async () => {
+    const delivery = {
+      id: 'd1',
+      localUserId: 'u1',
+      dingtalkUserId: 'dd1',
+      sourceType: 'automation',
+      subject: 'S',
+      content: 'C',
+      success: false,
+      status: 'outcome_unknown',
+      errorMessage: 'dingtalk_send_outcome_unknown: fetch failed',
+      createdAt: new Date().toISOString(),
+      localUserIsActive: true,
+    }
+    mount({ getAutomationDingTalkPersonDeliveries: vi.fn().mockResolvedValue([delivery]) })
+    await flush()
+    const badge = container!.querySelector('.meta-person-delivery__badge') as HTMLElement
+    expect(badge.getAttribute('data-status')).toBe('outcome_unknown')
+    expect(badge.textContent?.trim()).toBe('outcome unknown') // en locale label, NOT 'failed'
+    // the send-uncertainty reason stays visible (status !== success renders the error line)
+    expect(container!.querySelector('.meta-person-delivery__error')?.textContent).toContain('outcome_unknown')
+    // admin-side list filter: the new state is selectable, not silently excluded by the IN-list
+    const options = Array.from(container!.querySelectorAll('[data-field="statusFilter"] option')).map((o) => (o as HTMLOptionElement).value)
+    expect(options).toContain('outcome_unknown')
+  })
 })
--- a/packages/core-backend/src/db/migrations/zzzz20260710150000_add_outcome_unknown_delivery_status.ts
+++ b/packages/core-backend/src/db/migrations/zzzz20260710150000_add_outcome_unknown_delivery_status.ts
@@ -1,0 +1,133 @@
+/**
+ * Migration: `outcome_unknown` status for the three DingTalk send ledgers (PR #4046 Phase B,
+ * roadmap §7.2 send-tier semantics).
+ *
+ * Owner-ratified doctrine: a side-effect send whose outcome is unknowable (network error /
+ * timeout / 5xx / malformed 2xx — the client saw no usable response, DingTalk may well have
+ * delivered) is recorded as `outcome_unknown` and NEVER auto-resent. It is a DISTINCT terminal
+ * state, not a plain failure: reconciliation is manual/ops (DingTalk's async-send result query
+ * by task_id only helps when a task_id exists — a lost response has none, so it cannot be an
+ * idempotency key).
+ *
+ * Three persistence surfaces get the widened vocabulary:
+ *  - dingtalk_approval_card_deliveries.send_status  (approval-card ledger, one-tap lock #3594)
+ *  - dingtalk_person_deliveries.status              (automation person-message + rule-creator telemetry)
+ *  - attendance_notification_deliveries.status      (C5 outbox — worker terminal state, not re-claimed)
+ *
+ * Each table is guarded by checkTableExists so partial-schema test harnesses (isolated-schema
+ * suites that run only the migrations they need) can apply this migration too. down() maps
+ * existing outcome_unknown rows to 'failed' BEFORE re-tightening each constraint — a down()
+ * that fails on live rows is broken.
+ */
+import { sql, type Kysely } from 'kysely'
+import { checkTableExists } from './_patterns'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  if (await checkTableExists(db, 'dingtalk_approval_card_deliveries')) {
+    // The original send_status CHECK was added inline by
+    // zzzz20260705150100_add_send_columns_to_dingtalk_approval_card_deliveries (auto-named by
+    // Postgres). Drop whatever CHECK constrains send_status, then add a named, widened one.
+    await sql`
+      DO $$
+      DECLARE con RECORD;
+      BEGIN
+        FOR con IN
+          SELECT conname
+            FROM pg_constraint
+           WHERE conrelid = 'dingtalk_approval_card_deliveries'::regclass
+             AND contype = 'c'
+             AND pg_get_constraintdef(oid) LIKE '%send_status%'
+        LOOP
+          EXECUTE format('ALTER TABLE dingtalk_approval_card_deliveries DROP CONSTRAINT %I', con.conname);
+        END LOOP;
+      END $$;
+    `.execute(db)
+    await sql`
+      ALTER TABLE dingtalk_approval_card_deliveries
+      ADD CONSTRAINT chk_dingtalk_approval_card_deliveries_send_status
+      CHECK (send_status IN ('pending', 'sent', 'failed', 'outcome_unknown'))
+    `.execute(db)
+  }
+
+  if (await checkTableExists(db, 'dingtalk_person_deliveries')) {
+    await sql`
+      ALTER TABLE dingtalk_person_deliveries
+      DROP CONSTRAINT IF EXISTS chk_dingtalk_person_deliveries_status
+    `.execute(db)
+    await sql`
+      ALTER TABLE dingtalk_person_deliveries
+      ADD CONSTRAINT chk_dingtalk_person_deliveries_status
+      CHECK (status IN ('success', 'failed', 'skipped', 'outcome_unknown'))
+    `.execute(db)
+  }
+
+  if (await checkTableExists(db, 'attendance_notification_deliveries')) {
+    await sql`
+      ALTER TABLE attendance_notification_deliveries
+      DROP CONSTRAINT IF EXISTS chk_attendance_notification_deliveries_status
+    `.execute(db)
+    // NOTE: chk_attendance_notification_deliveries_delivered_status (delivered_at only when
+    // status='sent') is untouched — outcome_unknown rows never carry delivered_at.
+    await sql`
+      ALTER TABLE attendance_notification_deliveries
+      ADD CONSTRAINT chk_attendance_notification_deliveries_status
+      CHECK (status IN ('pending', 'sending', 'sent', 'retrying', 'failed', 'skipped', 'outcome_unknown'))
+    `.execute(db)
+  }
+}
+
+export async function down(db: Kysely<unknown>): Promise<void> {
+  if (await checkTableExists(db, 'dingtalk_approval_card_deliveries')) {
+    // Map live outcome_unknown rows to 'failed' BEFORE tightening the constraint back. This is
+    // lossy by design (the down path has no outcome_unknown vocabulary to keep); send_error
+    // retains the original failure text.
+    await sql`
+      UPDATE dingtalk_approval_card_deliveries
+         SET send_status = 'failed'
+       WHERE send_status = 'outcome_unknown'
+    `.execute(db)
+    await sql`
+      ALTER TABLE dingtalk_approval_card_deliveries
+      DROP CONSTRAINT IF EXISTS chk_dingtalk_approval_card_deliveries_send_status
+    `.execute(db)
+    await sql`
+      ALTER TABLE dingtalk_approval_card_deliveries
+      ADD CONSTRAINT chk_dingtalk_approval_card_deliveries_send_status
+      CHECK (send_status IN ('pending', 'sent', 'failed'))
+    `.execute(db)
+  }
+
+  if (await checkTableExists(db, 'dingtalk_person_deliveries')) {
+    await sql`
+      UPDATE dingtalk_person_deliveries
+         SET status = 'failed'
+       WHERE status = 'outcome_unknown'
+    `.execute(db)
+    await sql`
+      ALTER TABLE dingtalk_person_deliveries
+      DROP CONSTRAINT IF EXISTS chk_dingtalk_person_deliveries_status
+    `.execute(db)
+    await sql`
+      ALTER TABLE dingtalk_person_deliveries
+      ADD CONSTRAINT chk_dingtalk_person_deliveries_status
+      CHECK (status IN ('success', 'failed', 'skipped'))
+    `.execute(db)
+  }
+
+  if (await checkTableExists(db, 'attendance_notification_deliveries')) {
+    await sql`
+      UPDATE attendance_notification_deliveries
+         SET status = 'failed'
+       WHERE status = 'outcome_unknown'
+    `.execute(db)
+    await sql`
+      ALTER TABLE attendance_notification_deliveries
+      DROP CONSTRAINT IF EXISTS chk_attendance_notification_deliveries_status
+    `.execute(db)
+    await sql`
+      ALTER TABLE attendance_notification_deliveries
+      ADD CONSTRAINT chk_attendance_notification_deliveries_status
+      CHECK (status IN ('pending', 'sending', 'sent', 'retrying', 'failed', 'skipped'))
+    `.execute(db)
+  }
+}
--- a/packages/core-backend/src/db/types.ts
+++ b/packages/core-backend/src/db/types.ts
@@ -1553,7 +1553,7 @@ export interface DingTalkApprovalCardDeliveriesTable {
   acted_action: string | null
   acted_by: string | null
   acted_at: NullableTimestamp
-  send_status: 'pending' | 'sent' | 'failed'
+  send_status: 'pending' | 'sent' | 'failed' | 'outcome_unknown'
   send_error: string | null
   created_at: CreatedAt
   updated_at: CreatedAt
@@ -1567,7 +1567,7 @@ export interface DingTalkPersonDeliveriesTable {
   subject: string
   content: string
   success: boolean
-  status: 'success' | 'failed' | 'skipped'
+  status: 'success' | 'failed' | 'skipped' | 'outcome_unknown'
   http_status: number | null
   response_body: string | null
   error_message: string | null
--- a/packages/core-backend/src/integrations/dingtalk/approval-card-deliveries.ts
+++ b/packages/core-backend/src/integrations/dingtalk/approval-card-deliveries.ts
@@ -41,7 +41,14 @@ export interface DingTalkApprovalCardDeliveryRow {
   acted_action: string | null
   acted_by: string | null
   acted_at: Date | string | null
-  send_status: 'pending' | 'sent' | 'failed'
+  /**
+   * `outcome_unknown` (PR #4046 Phase B): the send threw with the transport's
+   * isDingTalkOutcomeUnknown marker — DingTalk may well have delivered the card even though the
+   * client saw no usable response. NEVER auto-resent (owner doctrine); reconciliation is
+   * manual/ops, and a valid HMAC callback (the token only exists inside the delivered card's
+   * deep link) is itself proof of delivery, so outcome_unknown rows stay claimable.
+   */
+  send_status: 'pending' | 'sent' | 'failed' | 'outcome_unknown'
   send_error: string | null
   created_at: Date | string
   updated_at: Date | string
@@ -107,10 +114,13 @@ export async function findDingTalkApprovalCardDeliveryById(
 
 /**
  * Atomic acted-claim: succeeds for exactly one caller while the card is still `sent` AND was
- * actually delivered (`send_status='sent'` — review P2: a pending/failed send never produced a
- * live button, so nothing may claim it). Returns the updated row for the winner, `null` for
- * everyone else (duplicate callback, double tap, undelivered card, or a card already
- * superseded/expired) — losers re-read via `find…ById` to render the real terminal state.
+ * possibly delivered (`send_status IN ('sent','outcome_unknown')` — review P2: a pending/failed
+ * send never produced a live button, so nothing may claim it; PR #4046 Phase B widened the set by
+ * exactly `outcome_unknown`, whose card MAY in fact have been delivered — a valid HMAC callback
+ * proves it was, so the ledger's send-time uncertainty must not make a delivered card inoperable).
+ * Returns the updated row for the winner, `null` for everyone else (duplicate callback, double
+ * tap, undelivered card, or a card already superseded/expired) — losers re-read via `find…ById`
+ * to render the real terminal state.
  */
 export async function claimDingTalkApprovalCardDeliveryActed(
   query: QueryFn,
@@ -120,7 +130,7 @@ export async function claimDingTalkApprovalCardDeliveryActed(
   const result = await query(
     `UPDATE dingtalk_approval_card_deliveries
      SET card_state = 'acted', acted_action = $2, acted_by = $3, acted_at = NOW(), updated_at = NOW()
-     WHERE id = $1 AND card_state = 'sent' AND send_status = 'sent'
+     WHERE id = $1 AND card_state = 'sent' AND send_status IN ('sent', 'outcome_unknown')
      RETURNING ${RETURNING_COLUMNS}`,
     [id, input.action, input.actedBy],
   )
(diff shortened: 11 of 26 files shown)


zensgit:
# Adversarial review — PR #4046 unified DingTalk transport resilience (roadmap §7.2)

- **PR**: zensgit/metasheet2 #4046, branch `claude/dt-transport-resilience` → `main`
- **Head pinned**: `906e88a9a3d47fe5b63fa31d9a109a81af6dfdb6` (verified identical before and after review)
- **Scope**: `.env.example`, `packages/core-backend/src/integrations/dingtalk/{transport.ts,client.ts}`, `packages/core-backend/tests/unit/{dingtalk-transport-resilience,dingtalk-request-timeout}.test.ts` — 5 files, no route/index/runtime wiring beyond the client seam (verified: `transport.ts` is imported ONLY by `client.ts` and its own test)
- **Method**: refute-first. Disposable worktree `/tmp/ms2-rev-pr4046` @ head, `pnpm install --frozen-lockfile`; ran new suite + H06 suite + neighbors + FULL `tests/unit`; re-fired mutations M2/M7/M8/M4 + M9 myself; audited all 10 call-site classifications against endpoint semantics; traced error-shape identity through every importer; traced the outbox double-retry layering end-to-end; checked flow-control codes against the repo's pre-existing discrimination. Reviewer: Claude (adversarial gate), 2026-07-10.

## Verdict: **APPROVE** (no P1, no P2; three P3 hardening notes, four NITs)

The implementer's report checked out on every load-bearing claim I could attack. The doctrine (read/exchange/send by business semantics, no verb inference; no resend on ambiguity; timeout ≠ proof of non-execution; task_id not an idempotency key) is faithfully implemented and mutation-proven.

## Test & mutation evidence (mine, not the author's)

| What | Result |
|---|---|
| `dingtalk-transport-resilience.test.ts` | 28/28 green |
| H06 `dingtalk-request-timeout.test.ts` | 5/5 green (only a comment changed vs main — verified via diff) |
| Neighbors (`dingtalk-client`, `attendance-scheduler`, `dingtalk-work-notification`, `dingtalk-work-notification-settings`, `notification-service-dingtalk`) | all green |
| Full `CI=true npx vitest run tests/unit` | **346 files / 4670 tests green** (matches claim exactly) |
| `tsc --noEmit -p tsconfig.json` | clean |
| **M2** sends classified like reads (`isRead ||= kind==='send'`) | **RED ×4** (send network / 5xx / 429 / flow-control) — matches claimed ×4 |
| **M7** exchange retries like reads | **RED ×4** (3 exchange tests + conservative-default) — matches |
| **M8** default tier `'exchange'`→`'read'` | **RED ×1** (conservative-default) — matches |
| **M4** per-attempt signal hoisted out of the loop (composed once, passed down) | **RED ×1** (freshness `seenAborted [false,true]`) while **H06 suite stayed 5/5 GREEN** — confirms the author's claim that the pre-existing suite alone would NOT catch this and the new test is load-bearing |
| **M9** (my extra) `markOutcomeUnknown` neutered | **RED ×3** (send network / 5xx / timeout marker asserts) |

Each mutation restored via `git checkout --`; worktree verified clean (`git status --porcelain` = 0) and removed after review.

## The hunt list — all cleared

**1. Classification audit (all 10 call sites, none missed).** Counted exactly 10: `requestDingTalkJson` ×2 (OAuth code exchange → `'exchange'` ✓; `/contact/users/me` GET → `'read'` ✓) + `requestDingTalkDirectoryJson` ×8 (gettoken → `'read'` ✓ idempotent within TTL; listsub / department/get / user/list / user/get → `'read'` ✓ query-shaped POSTs that only read; `getuserinfo` authCode → `'exchange'` ✓ single-use code; asyncsend_v2 ×2 → `'send'` ✓). No query-shaped POST that actually mutates. No naked `fetch` in client.ts bypassing the seam (grep: `fetchFn` appears only as option threading). The helpers take `kind` as a **required positional param** → classification is compile-enforced at every future call site. The raw transport's optional `kind` defaults to `'exchange'` (no retry) — mutation-proven (M8).

**2. Error-shape contract.** Classes physically moved to transport.ts, re-exported from client.ts — single definition, so `instanceof` identity holds for all three pattern-matching importers, each read in full: `routes/auth.ts:38` → 1295ff (callback: RequestError→502) and 1393-1411 (container login: BusinessError→401 `invalid_auth_code`, RequestError→502); `multitable/automation-executor.ts:19-20` → 1850ff/3125ff (httpStatus/responseBody ledger extraction); `AttendanceNotificationDeliveryWorker.ts:5-6` → 621/628 classifiers. Messages byte-identical (same `normalizeErrorMessage` + same fallback strings). Pre-existing `dingtalk-client.test.ts` (which pins envelope negative shapes incl. errcode 40078/40001) passes **unchanged**.

**3. Double-retry layering (the highest-value question).** Traced end-to-end:
- Send tier: transport makes exactly 1 attempt (M2 proves the guard). The outbox (`AttendanceNotificationDeliveryWorker`) remains the ONLY retry authority for sends.
- The outbox **does** resend after ambiguous failures: `classifyDingTalkSendError` (worker:620-640) routes `DingTalkTimeoutError` and raw network errors through the generic branch → `retryable: true` → `markRetrying` → resend on `computeDeliveryBackoffMs` schedule (60s/5m/15m/1h). **This duplicate hazard is pre-existing and byte-for-byte unchanged by this PR** — the classifier is untouched and the thrown error shapes are identical, so no classification flips. The PR neither adds nor removes a duplicate source; it surfaces the `isDingTalkOutcomeUnknown` marker the worker lane must consume to close the hazard (declared honest gap #1). See P3-1/P3-2.
- Read-tier transport retry composes under the outbox only via `fetchDingTalkAppAccessToken` (safe, idempotent, backoff-bounded). No multiplication of side effects.
- automation-executor DingTalk sends (1824/2725/3072): no retry loop of their own — ledger-record on failure. ✓

**4. Envelope errcode moved inside the retry loop.** For non-retryable tiers the `DingTalkBusinessError` is thrown on attempt 1 with identical shape/message (same `normalizeDingTalkApiPayload`, same fallbackError string) — nothing new bubbles, nothing is swallowed. The only behavior delta is read-tier flow-control errcodes now get backoff+retry, which is the point.

**5. Flow-control errcode set.** `90018` (documented 频率超限) and `-1` (系统繁忙) are DingTalk-documented; `408/429/5xx/50001` exactly mirrors the repo's pre-existing `isRetryableDingTalkErrorCode` (worker:616-618). No invented codes. Cross-layer conflict impossible in practice: transport consults the set for retry only on reads; the worker only classifies sends.

**6. Signal precedence change** (main: caller `init.signal` **replaced** the timeout; PR: composed via `AbortSignal.any` on Node ≥20.3). No in-repo caller passes a signal (verified by grep); author flagged it. Bonus: CI's `test (18.x)` lane (green) actually executes the pre-`AbortSignal.any` fallback for the composition tests, so honest gap #6 ("fallback untested") is slightly over-pessimistic — the fallback path runs under Node 18 CI.

## Roadmap §7.2 gate table

| Gate | Verdict | Evidence |
|---|---|---|
| Centralize timeout/retry/rate-limit/error classification/token-cache | **PASS** | one seam; token cache (pre-existing lock) sits above it; group-robot webhook explicitly out of scope + listed as follow-up |
| Retry only safe/idempotent API classes | **PASS** | tier matrix, M2/M7/M8 mutation-proven |
| Recognize flow-control errors and back off | **PASS** | HTTP 429 + envelope errcodes, Retry-After honored (clamped), full jitter |
| Keep work-notification sends bounded and classified | **PASS** | 1 attempt, 10s per-attempt timeout, typed outcome-unknown marker |
| Acceptance: directory sync survives transient 5xx/rate-limit within bounds | **PASS** (transport layer) | all directory calls are `'read'`; directory-sync.ts untouched per lane boundary |
| Acceptance: repeated flow-control failures → clear run stats and alerts | **PARTIAL — declared** | warn-per-retry only; deferred to §7.4 (honest gap #2). Legitimate defer, not a silent miss |

Owner doctrine conformance: explicit `kind` at every call site ✓; never verb-inferred ✓; exchange never retries ✓; send never resends on network/timeout/5xx ✓; 429-on-send NOT whitelisted ✓ (test pins it); timeouts never retried on any tier ✓; task_id-not-idempotency-key rationale in code ✓.

## Findings

### P3-1 — `.env.example` over-claims "no duplicate notifications" end-to-end
`.env.example` (new hunk): "side-effect sends NEVER auto-resend (uncertain outcomes are surfaced as outcome-unknown instead — **no duplicate notifications**)". True at the transport layer only. The attendance outbox (`AttendanceNotificationDeliveryWorker.ts:620-640` + `markRetrying`) still classifies timeout/network send failures as retryable and **resends** — a timeout where DingTalk actually delivered still yields a duplicate message today. Concrete scenario: ops sees a user get the same attendance reminder twice after a network blip, reads `.env.example`, concludes duplicates are impossible, and misdirects the investigation. Fix (one line): qualify to "no duplicate sends from the transport; the attendance outbox retains its own retry policy until it consumes the outcome-unknown marker".

### P3-2 — Outcome-unknown marker unconsumed (deferred wiring; needs a tracker)
Both real ledger writers (delivery worker; automation-executor person-delivery) record plain failures and the worker resends. Declared honest gap #1, other lanes' files — correctly deferred, but the doctrine's user-visible guarantee is NOT delivered until the worker lane consumes `isDingTalkOutcomeUnknown` (record `outcome_unknown`, decide resend policy at ledger level). Recommend opening a tracking issue at merge so it doesn't evaporate.

### P3-3 — Envelope-code vs HTTP-status 5xx asymmetry
An HTTP-200 envelope `code: 501` retries on reads (`isDingTalkFlowControlErrcode` spans all 5xx for worker parity, transport.ts:162-169) while a real HTTP 501 deliberately does not (`RETRYABLE_READ_5XX` = {500,502,503,504}, transport.ts:175). Declared (honest gap #5), parity-anchored; acceptable, noting for the record.

### NITs
1. transport.ts:160 comment "Only consulted for `'read'` calls" is imprecise — `classifyDingTalkFailure` consults the set for every tier (retry gated on read; for send it shapes the reason string / not-outcome-unknown decision).
2. `DingTalkRequestOptions.signal` (client.ts:94) has zero in-repo consumers and is untested through the client-function surface (covered only via `requestDingTalkTransportJson` directly). Harmless plumbing; watch the dead-option.
3. Root `package.json` engines allows `node >=18.0.0`, where `AbortSignal.any` is absent — fallback matches pre-PR semantics and the 18.x CI lane exercises it, but per-attempt timeout is silently absent there when a caller passes a signal (no caller does).
4. Enumerable `outcomeUnknown: true` on thrown errors appears in serialized dumps (author flagged); verified no ledger/shape consumer is affected (worker stores message + responseBody; automation-executor stringifies responseBody only).

## CI at review time
All checks passing incl. required set (`test (20.x)`, `test (18.x)`, contracts ×3, pr-validate); "Strict E2E with Enhanced Gates" in its normal skipping state.

— Claude adversarial gate, 2026-07-10. Review-only: no pushes, no merges, no fixes applied.

zensgit:
# Phase A closure — owner CHANGES-REQUESTED on #4046 (commit `dcf506faf`)

Addresses owner blockers 1 and 3 plus the gate's P3-1 and the transport.ts:160 NIT, **in transport.ts/tests only**. Phase B (ledger consumption of `outcomeUnknown` in the delivery worker / automation-executor) is explicitly **sequenced behind #4014/#4011 landings** — it touches their files. An **exact-head adversarial gate will run after Phase B**; this comment is the implementer's closure, not a gate verdict.

## Blocker 1 — what the body-timeout leak actually was

The owner's suspicion was correct, and the failure mode is **worse than a swallowed retry**: it was a **silent SUCCESS**.

Verified empirically first (standalone probe + failing tests against a real localhost server that sends headers then stalls the body):

1. undici **does** tie the response body stream to the request signal — `res.json()` rejects at the per-attempt timeout with the abort reason (`TimeoutError` DOMException; caller abort → `AbortError`). The body read does NOT escape the timeout at the stream level. Part (a) of the owner's question: no leak there.
2. Part (b) is the leak: `readJson()`'s catch-all (`catch { return null }`) ate that rejection. On the ok-path the `oapi` envelope then normalized `null → {}` (no errcode) and the transport **returned `{}` as success** — on the send tier that is a phantom "delivered" with no task_id. On the non-ok path the null payload produced `DingTalkRequestError(status)` — a stalled-body 500 was **misclassified as retryable `http_500`** on the read tier (observed: 3 attempts) instead of a never-retried timeout.

**Failing-test-first proof** (tests written and run against the unfixed transport at `906e88a9a`, real undici fetch + real stalling server, 4 RED):

```
× read tier … → expected undefined to be an instance of DingTalkTimeoutError   (transport RESOLVED)
× exchange tier … → promise resolved "{}" instead of rejecting                 (the silent success, verbatim)
× send tier … → resolved; isDingTalkOutcomeUnknown never surfaced
× body stall on NON-ok response → expected DingTalkRequestError: failed … to be an instance of DingTalkTimeoutError (retried 3×)
```

**Fix** (`readJson`, transport.ts): the tolerant `null` is now reserved for `SyntaxError` only (genuinely non-JSON body — pre-existing behavior preserved); abort/timeout rejections — including Node 18 undici's `TypeError: terminated` with the reason as `cause` — map to `DingTalkTimeoutError` with the **same per-tier classification as a request-phase timeout** (no retry on any tier; `outcomeUnknown` marker on send); any other mid-body stream failure (connection reset) rethrows raw → network-error classification (reads may retry, sends marked outcome-unknown, exchanges fail). `isAbortError` is now structural (name-based) because abort reasons are DOMExceptions, which only inherit from `Error` in recent Node lines (floor is >=18.0.0).

## Blocker 3 — the pre-20.3 fallback

The owner's instinct was right here too: the old fallback (`return callerSignal`) didn't compose at all — the per-attempt timeout was **silently dropped** whenever a caller passed a signal on Node <20.3 (and there was no listener to clean up because it never listened). It now composes manually (first signal to fire aborts the attempt, same abort-reason surface as `AbortSignal.any`) and **detaches its listeners when the attempt settles**, so a long-lived caller signal doesn't accumulate one listener per attempt. Three new unit tests force the path by stubbing `AbortSignal.any = undefined`:

- per-attempt timeout still fires with a live caller signal present → `DingTalkTimeoutError` (RED pre-fix: 1.5s guard fired instead)
- caller abort still cancels the in-flight attempt
- `getEventListeners(callerSignal, 'abort')` is empty after a 2-attempt retry run (leak pin)

## Mutation table (each applied to the committed fix, suite run, then restored)

| Mutation | Result |
|---|---|
| MU-1 revert `readJson` to catch-all swallow (owner's mutation) | **RED ×4** (all body-stall tests: 3 tiers + non-ok stall) |
| MU-2 fallback returns caller signal alone (owner's mutation) | **RED ×1** (fallback timeout test — guard fired at ~1.5s) |
| MU-3 fallback cleanup neutered (`dispose` → noop) | **RED ×1** (`expected […onCallerAbort…] to have a length of +0 but got 2`) |
| MU-4 body abort rethrown raw instead of `DingTalkTimeoutError` | **RED ×4** (instanceof fails; read tier also re-retried 3×, caught by request-count assert) |

## Also in this commit

- **P3-1**: `.env.example` claim qualified — "no duplicates introduced at the transport layer; the attendance outbox may still resend on ambiguous failures until it consumes the outcome-unknown marker — follow-up" (that follow-up is Phase B).
- **NIT (transport.ts:160)**: flow-control-set comment now says the classifier consults it for every tier with retry gated on `read`.
- **Two test fixtures repaired** (`dingtalk-transport-resilience.test.ts` MAX_ATTEMPTS=2 test; `automation-v1.test.ts` DT-OPS-04 two-corp split): both reused ONE consumed `Response` body across multiple attempts/calls — a real fetch never does that, and their old green **depended on exactly the swallow under repair** (the double-read `TypeError: Body is unusable` was eaten into a null payload / phantom send success). Fixed to mint a fresh `Response` per call; assertions unchanged.

## Test evidence

- `dingtalk-transport-resilience.test.ts`: 35/35 (28 pre-existing + 4 body-stall + 3 fallback)
- H06 `dingtalk-request-timeout.test.ts`: 5/5 (unchanged)
- Full `CI=true npx vitest run tests/unit`: **346 files / 4677 tests green** (one interim run had 3 ECONNRESET flakes in unrelated route tests; green on isolated rerun and on full rerun)
- `tsc --noEmit`: clean

## Honest gaps

- The `cause`-branch of `isBodyAbortError` (Node 18 undici's `terminated` wrap) is not killable on local Node 25 (the reason is thrown directly); it is exercised for real only by the **test(18.x) CI lane** running the new real-server body-stall tests.
- Phase B unshipped: ledger writers still record body/request timeouts as plain failures and the outbox still resends on ambiguous send failures — that is the declared follow-up behind #4014/#4011, not a Phase A regression.

zensgit:
## Phase B complete — the outcome-unknown marker is consumed; the system no longer auto-resends on ambiguity

Head: `61a9d5d7d` (rebased onto `origin/main` @ `31c2bbb7e`, which includes both prerequisites #4014 and #4011 **and** the mid-lane-landed #4013). Phase A (send-tier semantics) + Phase B (ledger consumption + malformed-2xx guard) together close both owner blockers. Full details in the updated PR body; this comment records the decisions and the audit.

### Status-vocabulary decision (exact)

**`outcome_unknown`** — the same literal in all three ledgers, one migration (`zzzz20260710150000_add_outcome_unknown_delivery_status.ts`, table-existence-guarded, `down()` maps live rows → `'failed'` before re-tightening; down/up verified against real Postgres with seeded rows on all three tables):

| Surface | Was (CHECK) | Now |
|---|---|---|
| `dingtalk_approval_card_deliveries.send_status` | `pending\|sent\|failed` (merged on main #3647 → new migration, no unmerged one existed on this branch) | + `outcome_unknown` |
| `dingtalk_person_deliveries.status` | `success\|failed\|skipped` | + `outcome_unknown` |
| `attendance_notification_deliveries.status` | `pending\|sending\|sent\|retrying\|failed\|skipped` ('skipped' precedent #3920) | + `outcome_unknown` |

Reconciliation, plainly: **manual/ops** via the distinct state + `send_error`/`last_error`. The DingTalk async-send result query (`task_id`) applies only when a task_id exists; a lost response has none, so it is never an idempotency key. In-process retry/backoff/rate-limiting remains best-effort.

### Reader-compatibility audit (every reader of the three status columns, file:line @ `61a9d5d7d`)

**`send_status` (approval-card ledger)**
- `packages/core-backend/src/integrations/dingtalk/approval-card-deliveries.ts:133` — acted-claim `WHERE … send_status IN ('sent','outcome_unknown')` (the ONLY widened predicate); `:149/:165/:187` mark-transitions keep the `send_status='pending'` guard (outcome_unknown is terminal for the send lifecycle).
- `packages/core-backend/src/services/ApprovalCardDeliveryAction.ts:130` — `actionable` widened by exactly `('sent'|'outcome_unknown')`; HMAC verify / instance-state / supersede / one-shot claim untouched.
- `packages/core-backend/src/db/types.ts:1556` — type union widened.
- `apps/web/src/approvals/cardDecision.ts` — `sendStatus` type union widened.
- `apps/web/src/views/approval/ApprovalCardDecisionView.vue:143` — the "未成功投递" stale message now fires only for `pending|failed` (KNOWN non-delivery); actionable outcome_unknown renders live buttons (server `actionable` drives the template — pinned in `apps/web/tests/approvalCardDecisionView.spec.ts`). UI shows: live 同意/驳回 buttons while pending; 已流转 if non-actionable for another reason.

**`dingtalk_person_deliveries.status`**
- `packages/core-backend/src/multitable/dingtalk-person-delivery-service.ts:48` — `normalizeDeliveryStatus` passes `outcome_unknown` through unfolded (folding it back is a RED test); unknown junk still folds by `success`.
- `packages/core-backend/src/routes/univer-meta.ts:16882` — list endpoint has no status filter; passes through.
- `apps/web/src/multitable/components/MetaAutomationPersonDeliveryViewer.vue:100` — distinct badge (`data-status="outcome_unknown"`, neutral grey), error line stays visible, and the status filter `<option value="outcome_unknown">` makes the state selectable (not IN-list-excluded). Label added at the sanctioned i18n extension point (`meta-automation-labels.ts`: `status.outcomeUnknown` = `outcome unknown` / `结果未知`).
- `apps/web/src/multitable/types.ts` — type union widened.
- Writers: card action skip-telemetry unchanged; batch action `automation-executor.ts:3189-3215` (in-flight batch only); rule-creator alert path (same file, `notifyRuleCreatorOfDingTalkGroupFailure` catch).

**`attendance_notification_deliveries.status`**
- `packages/core-backend/src/services/AttendanceNotificationDeliveryWorker.ts:826/:830` — claim predicate takes only `pending|retrying|expired sending` → `outcome_unknown` is never re-claimed; `:988/:1008/:1026` terminal writes keep the identical lease/CAS guard (`status='sending' AND claim_worker_id AND attempt_count`); quiet-hours gate untouched (sits before claim; my change is per-row post-send classification at `:1159`).
- Producers: `CompTimeExpiryReminderService.ts:187` + `UnscheduledReminderService.ts:152` dedupe by `(org_id, source_key)` existence / `ON CONFLICT DO NOTHING` — **status-independent, so an outcome_unknown row is never re-produced**.
- `runBatch` counters: optional `outcomeUnknown` count (present only when > 0, like `lostLease`) + log line — existing exact `toEqual` assertions unaffected.

**Three specific consumer classes (owner guardrail #3):** (a) **#4013 retention sweep** — `dingtalk-group-delivery-retention.ts:97-100` deletes from `dingtalk_group_deliveries` only, by `created_at < cutoff`, no status filter, touches none of the three tables; (b) **stats/aggregations** — no SQL aggregation over any of the three status columns exists in `src/` (checked `COUNT/SUM/GROUP BY/status IN` across the repo); (c) **admin list filters** — the only one is the person-delivery viewer's client-side filter, extended above.

### Mutation table (12 mutations, protocol: impl committed → mutate → run → restore)

| # | Mutation | Result |
|---|---|---|
| B-M1 | restore `null→{}` 2xx normalization | **RED ×6** (transport malformed suites) |
| B-M2 | drop oapi missing-errcode requirement only | **RED ×3** |
| B-M3 | exchange retries malformed | **RED ×2** |
| B-M4 | read gives up immediately on malformed | **RED ×2** |
| B-M5 | executor maps outcome-unknown → plain `failed` | **RED ×4** (card real-DB) |
| B-M6 | send-tier network error retryable (**"make it retry"**) | **RED ×4** card real-DB (single-asyncsend pin) + **RED ×1** transport unit |
| B-M7 | worker classifies outcome-unknown retryable (**the resend hazard**) | **RED ×1** worker real-DB + **RED ×2** channel unit |
| B-M8 | worker terminal state → `failed` instead of distinct | **RED ×1** worker real-DB |
| B-M9 | claim SQL back to `= 'sent'` | **RED ×2** (wrapper + ledger) |
| B-M10 | `actionable` back to `=== 'sent'` | **RED ×1** (wrapper) |
| B-M11 | in-flight batch scoping dropped (all unsent → outcome_unknown) | **RED ×1** (3-batch: never-attempted batch must stay `failed`) |
| B-M12 | endpoint reader folds outcome_unknown → `failed` | **RED ×1** (pass-through pin) |

### Test evidence @ `61a9d5d7d` (fresh DB, full migration stack)

- core-backend unit: **358 files / 4866 tests green**; `tsc` clean.
- Real-DB (fresh `lane_dtphaseb_*` DB): **6 suites / 52 tests green** (attendance-notification-deliveries, automation-dingtalk-approval-card-action, approval-card-delivery-wrapper.db, dingtalk-approval-card-deliveries.db, dingtalk-person-message-integration-scoping.db, automation-approval-task-created-trigger).
- Web: **232 tests green** across the 6 touched-surface spec files; `vue-tsc` clean.
- Migration `down()`: scripted real-DB verification — seeded outcome_unknown rows on all 3 tables → down() maps all → `'failed'`, constraint rejects the value again (23514) → up() re-accepts.

### Honest gaps

1. No automated reconciliation job / no aging policy for `outcome_unknown` rows (deliberate; #4013 covers group deliveries only).
2. Worker terminates on the FIRST outcome-unknown attempt; per-attempt error history is only `attempt_count` + latest `last_error`.
3. v1.0 `exchangeCodeForUserAccessToken`'s malformed-body path is pinned via the raw-transport 'none'-envelope test, not through that client function specifically.
4. Decision-page outcome_unknown behavior pinned at component level; no e2e over a real DingTalk callback.
5. down() verification is scripted, not a committed test (no migration-down test convention in the repo).

Not arming — exact-head gate to follow per the owner's sequencing.

zensgit:
# PR #4046 EXACT-HEAD adversarial gate — independent review (Claude, 2026-07-10)

- **PR**: #4046 `feat(dingtalk): unified transport resilience — tiered retry/backoff/rate-limit at the client seam (roadmap §7.2)` → `main`
- **Head reviewed**: `61a9d5d7dd8997daeac6d9db0f0bc59bb76f9f0c` — **exactly** the owner-specified SHA (verified against `refs/pull/4046/head` after force-fetch; my stale local review ref had pointed at the pre-rework 906e88a9a — the live branch is 61a9d5d7d)
- **Merge-base**: `31c2bbb7e` · 7 commits · 26 files (+2536 −179)
- **Method**: fresh disposable worktree `/tmp/ms2-rev-pr4046` (own build, not the implementer's lane worktree); full unit suite + tsc + vue-tsc re-run; own fresh-PG migration up/seed/down/up script; 6 real-DB suites on my own fresh DB (`ms2_rev_pr4046_it`, full migration stack); 11 mutations independently re-fired (incl. one of my own devising); reader audit by repo-wide grep with file:line evidence. Refute-first.

## VERDICT: **APPROVE**

All four owner items verified with independent evidence. Every mutation I fired went RED where claimed. No P1/P2 findings. Two P3s and two NITs, none merge-blocking. Landing stays owner-sequenced (not armed).

---

## Owner item ① — three-tier malformed-2xx classification

**PASS.** Code: `transport.ts` — `readJson` (L367) returns tolerant `null` only for genuine non-JSON/non-object bodies; `normalizeDingTalkApiPayload` (L389) throws `DingTalkMalformedResponseError` on `null` payload for EVERY envelope, and additionally on missing `errcode`/`code` for `envelope:'oapi'`. Tier split lives in ONE classification point (`classifyDingTalkFailure` L568-576): `retryable: isRead`, `outcomeUnknown: kind==='send'` — the tiers are not merged; exchange/send can never inherit read's retry.

All 5 malformed shapes (200+HTML / empty / truncated JSON / bare primitive / JSON-missing-errcode) are exercised **per tier** in `tests/unit/dingtalk-transport-resilience.test.ts` L318-463, plus top-level array, plus non-ok-status guard-the-guard, plus no-body-content-in-message.

Mutations I re-fired (each: mutate → run → record → `git checkout` restore):

| Mutation | My result | Claimed |
|---|---|---|
| B-M1 `null → {}` success restoration | **RED ×6** | RED ×6 ✓ |
| B-M2 drop oapi missing-errcode requirement | **RED ×3** | RED ×3 ✓ |
| B-M3 exchange retries malformed (`retryable: kind !== 'send'`) | **RED ×2** | RED ×2 ✓ |
| B-M4 read gives up immediately (`retryable: false`) | **RED ×2** | RED ×2 ✓ |
| B-M6 send-tier network error retryable | **RED ×1 unit + RED ×4 card real-DB** | ×1+×4 ✓ |

**Envelope classification audit** (misclassification-hole check): the only two `envelope:'none'` call sites are v1.0 `api.dingtalk.com` endpoints — `oauth2/userAccessToken` (client.ts:312) and `contact/users/me` (client.ts:358). Both are HTTP-status-erroring v1.0 endpoints whose success bodies carry **no** errcode — requiring one there would break every healthy call; not requiring it is correct. All 8 `oapi.dingtalk.com`/`topapi` call sites (gettoken, listsub, department/get, user/list, user/get, getuserinfo, asyncsend_v2 ×2) declare `envelope:'oapi'` and all genuinely return errcode envelopes. **No misclassification hole either direction.**

## Owner item ② — three ledgers, migration up/down on fresh PG

**PASS.** I wrote and ran my own script against a fresh local Postgres DB (`ms2_rev_pr4046_mig`), applying the REAL creating migrations in order (person 20260419214000+20260422225000, attendance 20260611120000, card 20260705120000+150100+20260709133000) then `zzzz20260710150000`. All checks passed:

- pre-up: `outcome_unknown` **rejected 23514 in all three tables** (widening is load-bearing, not pre-satisfied);
- up: exactly ONE `send_status` CHECK remains on the card table and it is the named `chk_dingtalk_approval_card_deliveries_send_status` — the DO-block correctly found and dropped the **auto-named inline** CHECK from 20260705150100;
- seed `outcome_unknown` in all three → accepted; bogus enum value still 23514 (strictness preserved);
- **down(): all three seeded rows mapped to `'failed'`; unrelated `success`/`sent` rows untouched; re-insert of `outcome_unknown` → 23514 in all three**;
- up again: re-accepted in all three; up() is re-runnable (idempotent).

Third-ledger in-flight scoping: `automation-executor.ts` L3090-3117 sets `inFlightSendBatch` strictly around the `sendDingTalkWorkNotification` await and nulls it on return, so token/config failures and never-attempted batches mark nobody unknown. **B-M11 re-fired** (scoping dropped → all unsent marked unknown): **RED ×1** — the 250-recipient/3-batch unit test (`automation-v1.test.ts` L1552) fails exactly on "batch-3 must stay failed".

Attendance claim exclusion: claim SQL takes only `('pending','retrying')` + expired `'sending'` (`AttendanceNotificationDeliveryWorker.ts` L826-832). I fired **my own mutation** widening the claim IN-list by `'outcome_unknown'` (the resend-hazard resurrection): **RED ×1** — real-DB test "zero retries, never re-claimed, never resent" catches it. Also re-fired **B-M7** (classifier says retryable): **RED ×2 unit + RED ×1 real-DB**. Card-claim revert **B-M9**: **RED ×2** (wrapper + ledger claimability). Reader-fold **B-M12**: **RED ×1**.

Producers cannot re-enqueue an unknown row: both attendance producers dedup via `ON CONFLICT (org_id, source_key) DO NOTHING` (CompTimeExpiryReminderService.ts:187, UnscheduledReminderService.ts:236) and `findDispatchesMissingDeliveries` uses status-less NOT EXISTS — an `outcome_unknown` row keeps occupying its source_key. No resend path via producers.

## Owner item ③ — surgical card-callback widening

**PASS.** Full-file diff vs merge-base of `approval-card-deliveries.ts` + `ApprovalCardDeliveryAction.ts`: the only semantic changes are (a) claim `WHERE … send_status IN ('sent','outcome_unknown')` (was `= 'sent'`), (b) `actionable: … (send_status === 'sent' || === 'outcome_unknown')`, (c) new `markDingTalkApprovalCardDeliverySendOutcomeUnknown` with the **identical pending-only guard** as mark-sent/mark-failed, (d) type union + comments. `verifyApprovalCardLinkToken` (HMAC), instance-status gate, `supersedeDingTalkApprovalCardDeliveriesForInstance`, and the one-shot claim structure are **byte-unchanged**.

Negative pins exist in the real-DB wrapper suite: `send_status=pending` stale (L289), `send_status=failed` stays stale — "widening is EXACTLY (sent, outcome_unknown), nothing else" (L301), outcome_unknown + valid HMAC actionable and claims (L320), invalid token still `not_found` (L359). **B-M10 re-fired: RED ×1. B-M9 re-fired: RED ×2.**

Flagged decision (1) (acted claim keeps `send_status='outcome_unknown'` under `card_state='acted'`): I grepped every `send_status`/`sendStatus` reader repo-wide. Backend: only the claim/actionable/mark helpers (this PR). Web: only the type in `cardDecision.ts:14` and the stale-title branch in `ApprovalCardDecisionView.vue:143` — and that computed checks `cardState === 'acted'` FIRST, so an acted-but-unknown card renders 该待办已处理, never a delivery misreport. No stats, no retention, no list surface treats send_status alone as delivery truth. **The decision is safe as shipped.**

## Owner item ④ — reader audit spot-check (my own file:line)

- **#4013 retention sweep**: `src/services/dingtalk-group-delivery-retention.ts:97-100` — `DELETE FROM dingtalk_group_deliveries WHERE created_at < now() - interval` (ctid-bounded). Group table only, age-only, **no status filter, none of the three Phase B tables**. ✓
- **Stats/aggregations**: repo-wide grep for COUNT/GROUP BY over the three tables — the only GROUP BYs are producer INSERT CTEs keyed on recipient identity (CompTime:158, Unscheduled:208), zero aggregation over any of the three status columns. Worker `runBatch` counter is additive-optional (`outcomeUnknown` present only when >0, same pattern as `lostLease`) so existing exact `toEqual` result assertions keep passing — verified by the green suites. ✓
- **IN-lists repo-wide**: the only IN-lists over these columns are the two correct ones (card claim widened; worker claim deliberately excluding). `directory-sync-alert-delivery.ts:64` and `ai-bulk-job-service.ts` hits are different tables. **No silent exclusion found.** ✓
- **Person viewer**: filter `<option value="outcome_unknown">`, distinct badge CSS, typed i18n `status.outcomeUnknown` (结果未知/outcome unknown) wired through the parity-checked label module. Endpoint pass-through pinned (B-M12 RED). Web specs run by me: `approvalCardDecisionView` + `meta-person-delivery-viewer-migration` + `meta-automation-labels` + `multitable-automation-rule-editor` (147 tests) + `meta-automation-delivery-viewers-i18n` (4) — **all green**; `vue-tsc` clean. ✓

## Extra items

- **Full suites (my runs)**: core-backend unit **358 files / 4866 tests green**; `tsc --noEmit` clean; real-DB **6 suites / 52 tests green** on my own fresh DB with the full migration stack. Matches the implementer's numbers exactly.
- **CI wiring**: all 6 real-DB suites are in `plugin-tests.yml` run-lists (L336-339, 405, 493) — the real-DB pins ARE CI-gated. ✓
- **gettoken fixture drift**: fixtures gained `errcode: 0` — that IS the real oapi `/gettoken` wire contract (`{"errcode":0,"errmsg":"ok","access_token":…,"expires_in":7200}`); the OLD fixtures were the drift, surfaced by the new missing-errcode guard. Correct direction, not fixture-bending.
- **`failedRecipientCount` (decision 3)**: repo-wide, the ONLY consumers are the executor itself and its own tests — it is step-output JSON, no programmatic downstream. `failed + outcomeUnknown = unsent` remains reconstructible from the same output object. No semantic breakage.
- **Send-tier caller inventory**: three callers of the send functions — automation-executor (both paths consume the marker), AttendanceNotificationDeliveryWorker (classifier consumes it first), and `work-notification-settings.ts:322` `testDingTalkWorkNotificationAgentId` — an interactive admin test-send with no ledger; the error surfaces directly to the caller (see P3-2).
- **Migration ordering**: `zzzz20260710150000` sorts after every existing migration (latest sibling `zzzz20260710130000`); FileMigrationProvider auto-discovers it. ✓
- **PR body**: no bare closing keywords — no issue auto-close hazard. ✓
- **package.json / lockfile**: untouched — no test-chain merge-union hazard, no dep changes. ✓
- **§7.2 lock conformance**: centralize ✓ · retry-only-safe-classes ✓ (mutation-proven) · flow-control recognition+backoff ✓ · sends bounded/classified ✓ · acceptance-1 (directory sync survives transient 5xx/429) ✓ via read-tier at the token/list/get seam · acceptance-2 (flow-control run stats/alerts) PARTIAL — explicitly deferred to §7.4 in the PR body; consistent with lock discipline, not a finding.

## Findings

**P1**: none. **P2**: none.

- **P3-1 — touched web specs are CI-ungated**: `approvalCardDecisionView.spec.ts` and `meta-person-delivery-viewer-migration.spec.ts` appear in NO workflow run-list (checked approval-web-guard / multitable-web-guard / attendance-web-guard / web-tests; `test(20.x)` only builds apps/web). The new FE pins (badge/filter not folded to failed; 已流转-not-未成功投递) therefore regress silently in CI. Pre-existing repo-wide gap, merely extended here → P3. Fix: add both files (and `meta-automation-labels.spec.ts` if not present) to the appropriate guard run-lists in a follow-up.
- **P3-2 — admin test-send drops the marker**: `work-notification-settings.ts:322` (`testDingTalkWorkNotificationAgentId`) can throw a marked outcome-unknown error that surfaces as a plain failure to the admin. Interactive surface, no ledger, no resend loop — acceptable; a "可能已送达" hint would improve ops accuracy. Follow-up nicety.
- **NIT-1**: `DINGTALK_REQUEST_TIMEOUT_MS` is module-load-frozen while the three `DINGTALK_TRANSPORT_*` knobs are read per call — inconsistent hot-override behavior; both are documented, but a one-line comment in `.env.example` noting the timeout needs a restart would prevent ops confusion.
- **NIT-2**: `attendance_notification_deliveries.outcome_unknown` rows have no aging/retention policy (PR body honest-gap #1, owner doctrine — reconciliation is manual). Fine now; if unknown rows accumulate, the manual queue needs an ops surface eventually.

## Mutation log (all restored via `git checkout --`; worktree verified pristine at 61a9d5d7d)

| # | Target | Result |
|---|---|---|
| B-M1 | transport `null→{}` | RED ×6 |
| B-M2 | transport drop missing-errcode | RED ×3 |
| B-M3 | transport exchange retries malformed | RED ×2 |
| B-M4 | transport read no-retry malformed | RED ×2 |
| B-M6 | transport send retries network err | RED ×1 unit + ×4 real-DB |
| B-M7 | worker classifier retryable | RED ×2 unit + ×1 real-DB |
| B-M9 | card claim back to `'sent'` | RED ×2 real-DB |
| B-M10 | actionable back to `==='sent'` | RED ×1 real-DB |
| B-M11 | executor in-flight scoping dropped | RED ×1 unit |
| B-M12 | endpoint folds unknown→failed | RED ×1 unit |
| own | attendance claim IN-list + `'outcome_unknown'` | RED ×1 real-DB |

*Reviewer: Claude (Fable 5) — independent adversarial gate; review-only, nothing pushed to the branch; landing not armed.*

zensgit:
**Landing window executed — with one semantic integration seam a reviewer should see.** Rebase onto post-#4057 main hit Slice-B's newly-landed `sendDingTalkInteractiveApprovalCard` (written against the pre-transport client: unbound re-exported error class + old `requestDingTalkJson` signature). Resolution (`871beeaf7`, on top of exact-head-gated `61a9d5d7d`): bind the class locally and **classify the B-2 call `kind='send'`** — the interactive card is a side-effect send, so ambiguous outcomes now surface as outcome-unknown and flow into the SHARED executor catch that Phase B already wired (verified: both card branches share one try/catch → the new branch inherits the distinct `outcome_unknown` ledger state with zero extra code). Post-adaptation verification: tsc clean · transport+timeout+quiet-hours units green · card real-DB suites 33/33 on fresh migrated DB · full unit 363 files / 4998 green. Armed alone — final code window of the pool.

PR #4047, opened 2026-07-10
fix(multitable): inbound-replay total contract (owner P2) + RB16 comment + R9 ledger fix-forward

zensgit:
**Owner strict follow-up review of #4025/#4042 → 1 P2 + 2 P3, all fixed here** (single small fix-forward per owner's recommendation).

### P2 — `total` contract violation (behavior fix)
`InboundReplayResult.total` documents `replayed + sum(skipped)`, but was computed BEFORE the drift-tripwire fold — in the drift window it under-counted to 0, RB15 pinned that inconsistency (`total===0`), and downstream `recoverable: replay.total > 0` (record-service.ts:1119) reported false despite a real captured tombstone. **Fix: compute `total` after the fold.** RB15 now expects `1` + asserts the contract identity outright.
- **Mutation-verified**: moving the computation back pre-fold → exactly RB15 red (`expected +0 to be 1`), 15 green; restored → **16/16** + PIT-neighbor **4/4** (fresh `postgres:16-alpine`, plugin-tests MIGRATION_EXCLUDE parity); `tsc --noEmit` clean.
- Consumer effect: drift-window restore responses now honestly report `recoverable: true` (a tombstone existed; only its skip LABEL is advisory). Non-drift paths byte-identical (fold is a no-op there).

### P3 — comment self-contradiction
Tripwire comment claimed NOT EXISTS 'still prevents any duplicate' while RB16 proves same-statement blindness — narrowed to: rejects duplicates **already visible in the statement's snapshot**; same-statement sibling inserts are invisible (RB16).

### P3 — R9 ledger staleness
`multitable-timemachine-r9-recovery-round-design-verification-20260710.md`: #4025 row corrected to merge SHA `ea45dde6a` (final head `7f48983a`); new **§2b** records this owner follow-up honestly — the round report had presented #4025/#4042 alongside the content-reviewed six when they had only been merge-verified.

### DIFF
--- a/docs/development/multitable-timemachine-r9-recovery-round-design-verification-20260710.md
+++ b/docs/development/multitable-timemachine-r9-recovery-round-design-verification-20260710.md
@@ -31,7 +31,10 @@
 
 轮末汇报曾把 #4025/#4042 与前六项并述——实际上二者当时**仅做了合并核验，未做 owner 内容深审**。补审结论：1 P2 + 2 P3，由本 fix-forward PR 同修：
 
-- **P2（行为修正，owner 裁决）**：`InboundReplayResult.total` 承诺 `total === replayed + sum(skipped)`，但实现先算 `total` 再做 tripwire fold（`skipped.alreadyPresent` 后增）——drift 窗口内 total 少计且下游 `recoverable: replay.total > 0`（record-service.ts）随之失真；RB15 还把这一不一致固化为断言（`total===0`）。修：**fold 完再算 total**；RB15 改期望 `total===1` 并新增契约恒等断言。
+- **P2（行为修正，owner 裁决，两阶段）**：`InboundReplayResult.total` 原实现先算 `total` 再做 tripwire fold——DOWNWARD drift 窗口内 total 少计到 0，下游 `recoverable: replay.total > 0`（record-service.ts）随之失真，RB15 还把这一不一致固化为断言（`total===0`）。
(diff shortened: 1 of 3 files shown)


zensgit:
**Owner P2 blocker (upward drift) fixed — head `fa885f20c`, auto-merge re-armed per「修完再放行」.**

- `total` is now **locked from the diagnostic snapshot** (`diagnosticTotal = replayable + sum(diag buckets)`; every anchor row lands in exactly one CASE bucket, and the anchor's tombstone set is stable in-txn — capture can't run for a deleted record; the retention floor keeps any group whose anchor still has the FOR-UPDATE-held trash row). Never derived from replayed/skipped arithmetic — exact in BOTH drift directions.
- The over-strong sum contract is dropped exactly as ruled: upward drift's stale diagnostic label cannot be attributed away, so the `skipped` breakdown is documented advisory in a drift window (interface doc updated); `total`/`replayed` always exact. Downward keeps the fold + sum identity (bonus, pinned in RB15, not promised).
- **NEW RB17** (owner-prescribed): diag-time declined → re-consent injected between the two statements ⇒ `replayed=1`, `edgeCount=1`, **`total=1`** (not the stale-label-inflated 2); the sum-exceeds-total boundary pinned honestly at 2.
- **Discriminating mutation** (fresh postgres:16-alpine, plugin-tests MIGRATION_EXCLUDE parity): reverting `total` to the derived post-fold sum → **exactly RB17 red (`expected 2 to be 1`), RB15 green** — proving RB17 pins snapshot-truth, not arithmetic. Restored: 17/17 + PIT-neighbor 4/4; tsc clean.

PR #3903, opened 2026-07-08
fix(dingtalk): claim a sync run lease before the DingTalk API pull (DT-HARDEN-05)

zensgit:
`syncDirectoryIntegration` inserts its run row and then walks the **entire DingTalk directory** before opening its apply transaction. Nothing stopped a manual sync, the scheduler, and a second replica from doing that at once: double consumption of the shared DingTalk API quota, two large apply transactions lock-waiting on the same integration, and a `last_seen_at < syncTimestamp` sweep that can deactivate rows the other run just wrote. **A transaction-scoped advisory lock around the apply would not protect the pull** — the expensive part happens before it. (A session advisory lock is also unsafe here: `query()` draws from a pool, so lock and unlock could land on different connections.)

The lease therefore lives on the run row, which already predates the first outbound call:
- a **partial unique index** on `(integration_id) WHERE status='running'` makes "at most one running run per integration" a *database* invariant — the second claim is a unique violation, not a lost race under READ COMMITTED;
- the loser gets `DirectorySyncInProgressError` → the route answers **409 with the active runId on BOTH trigger paths (sync and `{async:true}`)**, the scheduler logs a **skip**;
- **liveness is a heartbeat, not run age**: the sync beats `last_heartbeat_at` every `DIRECTORY_SYNC_HEARTBEAT_INTERVAL_MS` (default 60s; invalid/sub-5s values fall back), and reclaim fires only when `COALESCE(last_heartbeat_at, started_at)` is older than `DIRECTORY_SYNC_LEASE_STALE_MINUTES` (default 10, clamped to ≥5× the beat interval). A live long pull on a large tenant is never reclaimed; a crashed run is reclaimed within minutes — before each claim and once at scheduler boot;
- completion **and** failure release the lease, in **one transaction** (lease first), and the completion `UPDATE` is guarded `AND status='running'`.

Two deliberate behavior changes (ops-visible):
- a **reclaimed zombie that later finishes rolls back its entire apply** (`DirectorySyncLeaseLostError` inside the apply transaction) instead of silently completing over the new claimant's writes;
- `error_message` distinguishes *reclaimed-while-presumed-crashed* (`orphaned: sync lease heartbeat went stale…`) from a run's own failure.

[...]

PR #4049, opened 2026-07-10
fix(dingtalk): preview integrity — same-batch overcount + lease-aware refusal

zensgit:
## What / why

Two scoped correctness fixes to `previewDirectorySyncIntegration` (DT-OPS-02), from the design-verif MD residual tracker.

**FIX 1 — same-batch preview overcount.** Preview counted two brand-new pulled users who share an email/mobile as 2 separate `autoAdmissionCandidateCount` entries, because it walks `resolveDirectoryIdentityMatch` against `identityMatchMaps` built once up front. Apply, however, mutates those same maps as it admits users *within a single run* — so the second duplicate resolves `matched:'email'`/`'mobile'` against the user apply just created for the first, and only 1 user is actually created. Preview now mirrors apply's post-admit map mutations (`emailMap.set` / `mobileMap.set` / ambiguous-key deletes / identity-map sets) with a sentinel userId (`'preview-would-admit'`) right after counting an in-scope candidate, keyed with the exact same `normalizeText`/`normalizeMobileIdentifier` helpers `resolveDirectoryIdentityMatch` uses internally — so a later duplicate in the same pull resolves against it exactly as it would resolve against the real created user.

**FIX 2 (R3) — preview ignored the sync lease.** `previewDirectorySyncIntegration` pulls the FULL DingTalk directory (same API quota as a real sync) with no lease check at all. Running a preview while a real sync is in flight for the same integration doubles the shared API quota consumption the DT-HARDEN-05 lease exists to bound. Preview now calls the existing `findActiveDirectorySyncRunId` **read-only** (it never claims a lease of its own — unlike apply, no `claimDirectorySyncRun`/`INSERT INTO directory_sync_runs`) before the pull, and throws `DirectorySyncInProgressError` when a run is active. The preview route (`POST /integrations/:id/sync/preview`) now maps that error to 409 + `activeRunId`, exactly like the two sync-trigger branches already do.

## Honest gaps (not proven / by design)

- **FIX 1 residual** (documented in the doc-comment): preview cannot foresee an apply-side creation **failure**. If `createDirectoryAdmittedUserInTransaction` throws mid-run for the first half of a duplicate pair, apply's catch block skips its own map mutation, so apply's own second-duplicate resolution differs from a clean run too — this is an apply-side divergence unrelated to preview, and preview (which cannot know a future apply run will fail) always mutates the maps. Preview and apply only diverge here when apply itself fails partway through a run — not a preview bug.
- FIX 1's new test only proves the same-email and same-mobile duplicate shapes; it does not add a case for two new users sharing only an `external_key`/`union_id`/`open_id` (the sentinel mutation covers that map too, per the brief's `proposedFix`, but it isn't independently exercised by a new fixture — a real DingTalk pull cannot produce two accounts with the same union/open id in practice, so this is lower-risk than the email/mobile paths).
- FIX 2's real-DB coverage is the existing `directory-sync-run-lease.db.test.ts` (untouched, still green) proving the underlying lease/reclaim mechanics against real Postgres; the NEW preview-refusal assertions are unit-level (mocked pg), following the same precedent the existing `syncDirectoryIntegration` lease tests already use in that file. I did not add a dedicated real-DB preview-refusal test — the lease semantics themselves are already real-DB-proven, and mocked-pg is the "cheaper honest seam" the brief pointed at.

## Tests

All run against a fresh `lane_dtpreview_*` Postgres DB (dropped after) + `tsc --noEmit` (clean) + full `tests/unit` (**4667/4667 passed, 347 files**).

- `tests/integration/directory-sync-preview-apply-parity.db.test.ts` (already whitelisted in `plugin-tests.yml` — no new wiring): extended with a new `describe` block — a separate integration fixture with two users sharing an email, two sharing a mobile, and one solo new user. **6/6 passed** (4 pre-existing + 2 new):
  - `preview counts exactly 3 candidates — one per duplicate pair plus the solo user, not 5`
  - `apply creates exactly 3 users for the batch — preview/apply parity holds for intra-batch duplicates`
- `tests/unit/directory-sync-run-lease.test.ts` (mocked-pg precedent, matches the existing `syncDirectoryIntegration` lease pins in the same file): new `describe` block, **3 new tests, 16/16 passed** in the file:
  - refuses a preview while a sync is running and never touches the DingTalk API
  - carries the active run id, and never claims a lease of its own (asserts no `INSERT INTO directory_sync_runs`)
  - proceeds with the pull when no run is active (happy-path smoke)
- `tests/unit/admin-directory-routes.test.ts`: **1 new test, all 60 tests in the file passed**:
  - `maps a lease conflict to 409 with the active runId on the preview path`

## Mutation table

| Mutation | Test that goes RED | Confirmed |
|---|---|---|
| Remove the sentinel map-mutation block in the preview loop (FIX 1) | Both new cases in `directory-sync-preview-apply-parity.db.test.ts` (`preview counts exactly 3...` expected 5 got 3→**5**; `apply creates exactly 3...` expected 5 got **3**) | ✅ RED, confirmed |
| Remove the `findActiveDirectorySyncRunId` lease check from `previewDirectorySyncIntegration` (FIX 2) | Both new `directory-sync-run-lease.test.ts` R3 tests (`TypeError: children is not iterable` instead of `DirectorySyncInProgressError` — the pull proceeded and hit the mocked DingTalk client with a stale mock queue) | ✅ RED, confirmed |
| Remove the `instanceof DirectorySyncInProgressError` branch from the preview route's catch (FIX 2, route side) | `maps a lease conflict to 409 with the active runId on the preview path` (500 instead of 409) | ✅ RED, confirmed |

Each mutation was applied on top of the committed implementation, run to confirm RED, then restored via `git checkout --` per the mutation-testing protocol (working tree is clean — `git diff --stat 685643cfc..HEAD` shows exactly the 5 intended files, 338 insertions / 5 deletions).

## Base / scope

Base is `origin/main` at merge `685643cfc` (DT-HARDEN-05, #3903) — confirmed an ancestor of current `origin/main`, and no intervening main commit touches any of the 5 files this PR changes, so no conflict risk.

Scope: `packages/core-backend/src/directory/directory-sync.ts`, `packages/core-backend/src/routes/admin-directory.ts`, and their test files only. Per lane discipline, **not merging/arming this PR** — landing is gated and serialized by the main loop (Wave 2, lands after the Wave 1 queue drains).

### DIFF
--- a/packages/core-backend/src/directory/directory-sync.ts
+++ b/packages/core-backend/src/directory/directory-sync.ts
@@ -3476,12 +3476,31 @@ export async function syncDirectoryIntegration(
  * eligibility check alone, so preview no longer counts accounts that are already linked or
  * would merely be linked (not created).
  *
- * Residual known gap (fail-safe, over-counts only): apply mutates its match maps as it
- * auto-admits users within a single run, so two brand-new in-scope pulled users who share an
- * email/mobile count as 1 apply-created account but 2 preview candidates. Preview does not
- * simulate creation order. This does not affect the already-linked / matched / out-of-scope
- * cases this function was built to fix.
+ * Same-batch duplicates: apply mutates its match maps as it auto-admits users within a
+ * single run, so a second brand-new in-scope pulled account that shares an email/mobile/
+ * external-identity with one already counted resolves matched:'email'/'mobile'/
+ * 'external_identity' against the FIRST account's freshly-created user, not matched:'none'
+ * again. Preview mirrors that below with a sentinel userId (`'preview-would-admit'`)
+ * written into `identityMatchMaps` right after counting an in-scope candidate, using the
+ * same normalizeText/normalizeMobileIdentifier keying `resolveDirectoryIdentityMatch`
+ * itself uses — so the next iteration's cascade call sees exactly what apply's would see.
+ *
+ * Residual known gap (fail-safe, over-counts only): this cannot foresee an apply-side
+ * creation FAILURE (e.g. `createDirectoryAdmittedUserInTransaction` throwing mid-run) —
+ * apply's catch block skips its map mutation for a failed account, so a later same-batch
+ * duplicate would still resolve against it as a candidate on the apply side too, while
+ * preview (which cannot know a future apply run will fail) always mutates the maps. Preview
+ * and apply only diverge here when apply itself fails partway through a run.
  */
+
+/**
+ * Placeholder written into `identityMatchMaps` in place of a real `created.userId` (preview
+ * never creates anything). Only ever compared as a map VALUE by `matched-kind` branching —
+ * never read back as a real user id or returned in any preview response field. Distinctive on
+ * purpose so it is unmistakable in a debugger or log if that invariant is ever violated.
+ */
+const PREVIEW_ADMIT_SENTINEL_USER_ID = 'preview-would-admit'
+
 export type DirectorySyncPreview = {
   integrationId: string
   integrationName: string
@@ -3509,6 +3528,15 @@ type ExistingAccountPreviewRow = {
 export async function previewDirectorySyncIntegration(integrationId: string): Promise<DirectorySyncPreview> {
   const integration = await getIntegrationRow(integrationId)
   if (!integration) throw new Error('Directory integration not found')
+
+  // DT-OPS-02 / R3: refuse a preview while a real sync holds the lease for this
+  // integration. Preview pulls the FULL DingTalk directory (same API quota `syncDirectoryIntegration`
+  // consumes — see the doc-comment above) with no lease check at all; running one during an
+  // in-flight sync doubles the shared quota consumption the lease exists to bound. This is a
+  // READ-ONLY refusal: unlike apply, preview never calls `claimDirectorySyncRun` — it only asks
+  // whether a run is currently `running` and, if so, declines before making any outbound call.
+  const activeRunId = await findActiveDirectorySyncRunId(integrationId)
+  if (activeRunId) throw new DirectorySyncInProgressError(activeRunId)
 
   const config = parseIntegrationConfig(integration)
   const departments = await fetchAllDepartments(config, integration.name)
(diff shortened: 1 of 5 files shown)


zensgit:
# PR #4049 adversarial gate — preview integrity (main-loop takeover after gate agent died on quota)

Head reviewed: `e514ef573`. Verdict: **APPROVE** (1 P3, 1 NIT — nothing blocking).

## Primary probe: normalizer-vs-sanitizer divergence (implementer self-flagged)
Read line-by-line. Lookup keys (directory-sync.ts:1129-1131): `normalizeText(email).toLowerCase()` / `normalizeMobileIdentifier(mobile)`. Apply's post-admit mutation (:3231-3232) keys on `sanitize*` output = same normalizers **+ `.slice(0,255)` / `.slice(0,32)` truncation**. The PR's sentinel (:3636-3649) uses the LOOKUP keys — the correct choice: it must anticipate the next iteration's lookup, which is untruncated.
- Divergence surface = emails >255 chars / mobiles >32 chars only. Unreachable for real DingTalk payloads, and the truncated-key map mutation is PRE-EXISTING apply behavior (apply's own second-duplicate lookup would miss its own map entry for such inputs — an apply-side edge, not introduced here). **P3, documented, no action.**
- Sentinel mirroring is COMPLETE: emailMap + mobileMap + both ambiguous-key deletes + externalIdentityMap + scopedOpen/scopedUnion maps — the full set apply mutates.

## Other probes — all pass
- **Lease refusal placement**: `findActiveDirectorySyncRunId` check sits before `parseIntegrationConfig`/`fetchAllDepartments`/`fetchAllUsers` — no outbound call happens on refusal. Read-only (single SELECT; no claim, no reclaim). Verified no write crept into the preview path.
- **Sentinel leak**: `PREVIEW_ADMIT_SENTINEL_USER_ID` referenced at exactly 6 sites (1 const + 5 map.set). Zero flow into preview output/response/counts (`localUserId` never emitted inside previewDirectorySyncIntegration).
- **Route 409 mapping**: test rides the mock factory that re-exports the REAL `DirectorySyncInProgressError` via importOriginal (the #3903 trap fix) — instanceof discrimination is genuinely exercised.
- **Stuck-run blocks preview until external reclaim**: acceptable — sync trigger reclaims pre-claim, scheduler boot sweeps, staleness default 10min. Admin path exists (trigger a sync). NIT.

## Dynamic verification (mine, fresh DB `gate4049_*`, migrated, dropped)
- parity `.db` suite 6/6; lease unit + route unit 63/63.
- **M1 re-fired**: sentinel block removed → parity RED ×2 (real DB).
- **M2 re-fired**: preview lease check removed → RED ×2.
Tree restored clean after each.

## Findings
- **P3** directory-sync.ts:3231 — apply's post-admit map key truncates (sanitize slice) while lookups don't; unrealistic inputs only; pre-existing. Recommend: none now; if DingTalk ever relaxes field caps, unify on one key fn.
- **NIT** preview refusal offers no self-heal for a stale-lease block (deliberate read-only choice; sync trigger heals).

zensgit:
Window protocol executed: rebased onto post-#4013 main (`8e9d88f6f`, clean); fresh-DB parity 6/6 + lease/route units 63/63 + tsc clean; load-bearing mutation replayed on the rebased head (sentinel block removed ⇒ parity RED ×2, restored). Armed alone per single-window discipline — Wave 2 window 1.

PR #4053, opened 2026-07-10
feat(directory): validate schedule_cron at save time (roadmap §7.8)

zensgit:
## What / why

Roadmap §7.8 (`docs/development/dingtalk-sync-integrated-roadmap-20260708.md`) targets "Validate cron at save time" for the directory sync scheduler. Before this PR, `POST /integrations` and `PUT /integrations/:integrationId` in `routes/admin-directory.ts` accepted **any** `schedule_cron` string with zero validation. An invalid or unschedulable expression was written straight into `directory_integrations.schedule_cron`, and the failure only ever surfaced later, silently, inside `directory-sync-scheduler.ts`'s `applySchedule`:

```ts
} catch (error) {
  logger.warn(`Failed to apply directory sync schedule for ${row.id}: ...`)
  if (existingJob) await scheduler.unschedule(jobName)
}
```

`scheduler.schedule()` / `.reschedule()` (`SchedulerServiceImpl` → `SimpleCronExpression` in `services/SchedulerService.ts`) throws on a bad cron; the catch block just logs a warning. The admin believes they scheduled a sync and nothing ever runs, with no error surfaced anywhere in the product.

## What changed

- `packages/core-backend/src/routes/admin-directory.ts`: both `POST /integrations` and `PUT /integrations/:integrationId` now validate `scheduleCron` before calling `createDirectoryIntegration` / `updateDirectoryIntegration`. An invalid/unschedulable expression returns `400 DIRECTORY_SCHEDULE_CRON_INVALID` with a message documenting the expected grammar and the UTC semantics (see below). Empty / null / whitespace-only stays allowed (= no schedule), matching the existing `normalizeOptionalText` semantics in `directory-sync.ts`.
- `docs/development/dingtalk-sync-integrated-roadmap-20260708.md` §7.8: marked the "Validate cron at save time" bullet landed with a pointer to this change; called out "Add timezone support" as the still-open follow-up.

## Validator choice — corrected mid-review (see commit history)

The lane brief pointed at the multitable automation scheduler's cron code (`packages/core-backend/src/multitable/automation-scheduler.ts`, `parseCronExpression` / `cronHasNoMatchingDay`) as "the" cron validator to reuse. **My first implementation did exactly that, and it was wrong** — caught by a second review pass before this PR settled, not by CI:

`parseCronExpression`/`cronHasNoMatchingDay` uses standard cron **OR-semantics** for day-of-month + day-of-week (either restriction can match). But the directory sync scheduler doesn't run on that parser at all — it runs on **`SimpleCronExpression`** (`services/SchedulerService.ts`, via `directory-sync-scheduler.ts` → `SchedulerServiceImpl`), which **ANDs** every field. Verified empirically:

```
0 0 30 2 1   ("any Monday in February")
  multitable parser (OR-semantics):  schedulable  ✅ accepted
  SimpleCronExpression (AND-semantics): NEVER matches (no Feb 30th exists) → hasNext() = false
```

Validating against the multitable parser would have let an admin save `0 0 30 2 1` as "valid" while the real scheduler silently dropped it forever — reproducing the exact bug this gate exists to close, just with a false sense of security. I switched the validator to `SimpleCronExpression` itself (already the underlying lib `directory-sync-scheduler.ts` depends on — still zero new dependencies), giving byte-for-byte parity with what will actually be scheduled, at the cost of a bounded scan (up to ~366 days of minutes; ~40ms worst case, measured) instead of an O(1) field check.

This is flagged as the top thing for a reviewer to scrutinize: **is there any other case where `SimpleCronExpression` diverges from the "obviously correct" grammar?** I did not attempt to formally prove equivalence beyond the specific AND/OR divergence found and the boundary cases in the test suite below.

## Timezone (explicitly NOT built here)

The directory sync scheduler hardcodes `timezone: 'UTC'` at `applySchedule` (`directory-sync-scheduler.ts`) — there is no per-integration timezone yet. This PR documents that in the route's error message and code comment; it does **not** add timezone config (that's a separate design decision — roadmap §7.8's "Add timezone support" bullet, still open). A `schedule_cron` accepted here always runs on UTC wall-clock time, validated with the same fixed `'UTC'`.

**Follow-up (not in this PR):** per-integration timezone, mirroring the multitable automation `schedule.cron` trigger's already-shipped `trigger_config.timezone` (T2-5).

## Bonus finding (not code — just confirmed, no FE change needed)

The directory admin FE (`apps/web/src/views/DirectoryManagementView.vue` `saveIntegration()`) already generically surfaces `error.message` from a non-2xx response via `readApiError` + `setStatus(..., 'error')`. So the new 400 error message renders in the UI with **zero frontend changes** — the roadmap's "Invalid cron cannot silently unschedule without clear UI feedback" acceptance criterion is satisfied end-to-end by this backend-only change.

## Tests

`packages/core-backend/tests/unit/admin-directory-routes.test.ts` (extended existing file, respecting its `vi.mock('../../src/directory/directory-sync', async (importOriginal) => ...)` factory that re-exports the real `DirectorySyncInProgressError` — untouched):

- `schedule_cron save-time validation (roadmap §7.8)` — 9 new tests: invalid cron on create → 400 + `createDirectoryIntegration` never called; day-impossible cron (Feb 30th) on create → 400; the AND/OR-divergence case (`0 0 30 2 1`) on create → 400, proven end-to-end through the actual route handler; invalid cron on update → 400 + `updateDirectoryIntegration` never called; valid cron passes through unmodified; `undefined` / `null` / `''` / `'   '` all allowed on create; `''` allowed on update (clears the schedule).
- `reused cron validator boundary semantics (SimpleCronExpression / SchedulerService)` — direct pin of `SimpleCronExpression`'s real behavior (imported un-mocked, empirically verified case-by-case before writing assertions): 5-field wildcard, `*/5` step syntax, comma lists, range+step, Sunday-as-7, "every Monday" (DOM wildcard + DOW restricted — reachable), 6-field-with-seconds **rejected**, 4-field **rejected**, nonsense text, out-of-range minute/hour/day/month, negative values, zero step, Feb 30th, the AND-semantics-impossible combo, empty/whitespace — 19 cases — plus one test that directly contrasts the multitable parser's OR-semantics against `SimpleCronExpression`'s AND-semantics on the same input, documenting why the validator choice changed.

Full run: `CI=true npx vitest run tests/unit/admin-directory-routes.test.ts` → **77/77 passed**. Full `tests/unit`: **4733/4733 passed** (348 files). `npx tsc --noEmit -p tsconfig.json`: clean.

## Mutation table (commit real impl → mutate → run → record → revert)

| # | Mutation | Test(s) that went RED |
|---|---|---|
| 1 | Delete the validation guard block from `POST /integrations` | `rejects an invalid schedule_cron on create...`, `rejects a day-impossible schedule_cron on create (Feb 30th...)`, `rejects a day-of-month + day-of-week combo...(0 0 30 2 1)` |
| 2 | Delete the validation guard block from `PUT /integrations/:integrationId` | `rejects an invalid schedule_cron on update...` |
| 3 | `isDirectoryScheduleCronValid`: drop the `.hasNext()` check (accept anything that merely parses without throwing) | `rejects a day-impossible schedule_cron on create (Feb 30th...)` |
| 4 | Guard condition `scheduleCronInput && !isValid(...)` → `!isValid(...)` (validates empty string too) | all 4 "allows ... schedule_cron on create" tests |

All mutations reverted via `git checkout -- packages/core-backend/src/routes/admin-directory.ts` after recording red; working tree confirmed clean before pushing.

## Honest gaps / reviewer notes

- **The validator-choice reversal above is the main thing to scrutinize.** I did not exhaustively prove `SimpleCronExpression` matches "true" cron semantics beyond the specific divergence found and the pinned boundary cases — it's a pre-existing, independently-implemented parser with its own quirks (e.g. its range parsing doesn't bounds-check before `hasNext()` catches it, which I confirmed behaves correctly but via a different code path than the constructor's explicit-value checks).
- **Deviated from the literal lane brief** ("reuse the multitable automation cron validator") after empirically proving that doing so would accept a cron the real scheduler silently drops. Chose fidelity to the actual runtime behavior over the letter of the instruction; flagging explicitly per lane discipline.
- **No FE changes**: confirmed (not modified) that the existing generic error-surfacing path in `DirectoryManagementView.vue` already renders this new 400 message; did not add a schedule_cron-specific UI hint or client-side pre-validation.
- **Timezone support itself is out of scope**, per the lane brief — this PR only documents the current UTC-only behavior; it does not add a timezone field/config surface.
- Did not touch the sync (`/integrations/:integrationId/sync`) or preview (`/integrations/:integrationId/sync/preview`) routes — kept the diff strictly to the create/update handlers per the brief's coordination note (landed #3903 sync 409 mapping, unlanded #4049 preview route catch, both in the same file but different regions). Verified via `git log bdf37fb08..origin/main -- <file>` that no other commit touched either of the two files I changed since my branch point.
- A validated cron with a very sparse legitimate schedule (e.g., an unusual leap-day-only pattern) could, in principle, validate differently depending on the exact moment `hasNext()` runs (it scans forward from "now" for up to 366 days) — this is inherent to reusing the real runtime scheduler for validation rather than a pure structural check, and is the same behavior the scheduler itself would exhibit, so it's considered correct-by-construction rather than a gap. Not covered by a dedicated test because asserting on it would be time-of-test-run-dependent; the AND/OR divergence case used for regression coverage instead is deterministic (always false, not date-dependent).


zensgit:
# Adversarial Review — PR #4053 `feat(directory): validate schedule_cron at save time (roadmap §7.8)`

- **PR**: zensgit/metasheet2 #4053, branch `claude/dt-directory-cron-validation`, base `main`
- **Head reviewed**: `549d40f2b7c2868e45efefb3cfbadb70967b8265` (verified == PR head via `git fetch origin pull/4053/head`)
- **Branch point**: `bdf37fb08`; `git log bdf37fb08..origin/main -- <both code files>` = empty (no drift; implementer's claim verified)
- **Scope**: 3 files — `packages/core-backend/src/routes/admin-directory.ts` (+66), `packages/core-backend/tests/unit/admin-directory-routes.test.ts` (+182), `docs/development/dingtalk-sync-integrated-roadmap-20260708.md` (±4)
- **Method**: refute-first; full-file reads of `SchedulerService.ts`, `directory-sync-scheduler.ts`, `directory-sync.ts` (normalization + create/update SQL), `DirectoryManagementView.vue` payload path at PR head; disposable worktree `/tmp/ms2-rev-pr4053` @ head, `pnpm install --frozen-lockfile --prefer-offline`; real test runs; own boundary/timing probe suite; 4 mutations re-fired; bypass-write sweep; CI-gate confirmation. Reviewer: Claude (Fable 5), 2026-07-10.

## Verdict: **APPROVE with hardening** — one P2 (doc-only) should be fixed before/at merge; runtime code is correct, guard fully tested, mutation-hardened, no scope drift.

The implementer's headline claim — that the lane brief's "reuse the multitable automation parser" was wrong and validation must use `SimpleCronExpression` (the directory scheduler's actual runtime parser) — **is correct and I verified it independently**. Ironically, the one defect I found is that the PR's own roadmap-doc edit still documents the abandoned multitable-parser choice (P2-1 below).

---

## Probe 1 — Validator-runtime identity: **PASS**

- `directory-sync-scheduler.ts` `applySchedule` → `SchedulerServiceImpl.schedule/reschedule` (`services/SchedulerService.ts`). Both validate via `new SimpleCronExpression(cronExpression, options.timezone)` + `hasNext()` (schedule ~L358-363, reschedule ~L432-437), and `JobScheduler.scheduleCronJob` re-parses with the same class for `next()`. **No fallback parser, no cron library anywhere on the path** (full file read; the only cron code in `SchedulerService.ts` is `SimpleCronExpression` itself).
- Route validation (`admin-directory.ts` `isDirectoryScheduleCronValid`): `new SimpleCronExpression(cron, 'UTC').hasNext()` in try/catch — **the same class, constructed with the same `'UTC'` arg `applySchedule` passes**. Not a reimplementation.
- Empirical parity (my probe, run in-worktree): `SchedulerServiceImpl.schedule('probe', '0 0 30 2 *', …, {timezone:'UTC'})` **rejects** (throws) exactly where the route 400s; `'0 2 * * *'` schedules fine. Identity holds end-to-end.
- Normalization parity: route's `normalizeScheduleCronInput` ≡ `directory-sync.ts` private `normalizeText` (byte-identical logic); persistence stores `normalizeOptionalText(input.scheduleCron)` = trimmed-or-null, i.e. exactly the string that was validated. `shouldSchedule`/`applySchedule` normalize the DB value the same way. No validate-X-store-Y drift.
- Bypass sweep: no other caller of `createDirectoryIntegration`/`updateDirectoryIntegration`; no other `INSERT/UPDATE directory_integrations` touches `schedule_cron` (approval-card-config / work-notification-settings / status-update SQL all confirmed schedule_cron-free). The gated route pair is the sole write surface for the column.

## Probe 2 — `hasNext()` bound / DoS / rare-valid crons: **PASS (bounded), with two P3 notes**

- `next()` is hard-bounded at `366 * 24 * 60` = 527,040 iterations (`SchedulerService.ts:97`); `hasNext()` calls it once on a fresh per-request instance. **Measured wall-time** (my probe at head):

  | expression | verdict | ms |
  |---|---|---|
  | `0 0 30 2 *` (never fires) | reject | 40.7 |
  | `* * 30 2 *` (worst case: min+hour always match) | reject | 55.0 |
  | `0 0 30 2 1` (AND-combo) | reject | 38.0 |
  | `0 0 29 2 *` (Feb-29 yearly) | reject | 37.8 |
  | `0 2 * * *` (valid) | accept | 0.1 |

  PR's "~40ms worst case" claim confirmed. No unbounded scan; validation runs **after** `ensurePlatformAdmin`, so only platform admins can trigger the ~50ms CPU burn. **No DoS.**
- **Rare-but-valid crons**: `0 0 29 2 *` gets 400 today (next leap day 2028-02-29 is beyond the 366-day horizon from 2026-07-10) — a "false 400" vs textbook cron but a **true 400 vs this runtime**: `SchedulerServiceImpl.schedule` performs the identical `hasNext()` check and throws, and `applySchedule` would warn-and-drop — the job would genuinely never fire. The error message documents the one-year horizon. Save-date-dependence (same expression accepted if saved within 366 days of a leap day) is inherent to runtime-faithful validation → P3-2 note, not a defect.

## Probe 3 — Update-path parity / legacy rows: **behavior verified, acceptable (P3-1 note)**

- PUT omitting `scheduleCron` → normalizes to `''` → validation skipped. Consistent with persistence: `updateDirectoryIntegration` is a **full-replace** (`directory-sync.ts:2018` `scheduleCron: normalizeOptionalText(input.scheduleCron)` — no `?? current` carry-through, unlike `baseUrl`), so an omitted field was already "clear the schedule", not "keep". Pre-existing semantics, unchanged by this PR.
- **Legacy row with stored-invalid cron**: FE (`DirectoryManagementView.vue:2800` loads stored value; `:4062` always sends `scheduleCron: draft.scheduleCron.trim()`) round-trips it, so editing *any* unrelated field now 400s until the admin fixes/empties the cron field. **Not a brick** — the cron input is on the same form and the 400 message says exactly what to do — and any such stored cron was **never running anyway** (scheduler warn-and-drop), so the 400 surfaces a real latent misconfiguration instead of preserving the silent lie. Right behavior. Gap: no sweep/log of existing invalid rows to find them proactively → P3-1.

## Probe 4 — Route-factory / test integrity: **PASS**

- The `vi.mock('../../src/directory/directory-sync', async (importOriginal) => …)` factory is untouched and still re-exports the **real** `DirectorySyncInProgressError` via `importOriginal` (DT-HARDEN-05 comment block intact).
- `services/SchedulerService` is **not** in the test file's `vi.mock` list (verified: mocks are rbac, audit, directory-sync, directory-sync-scheduler, alert-delivery, work-notification-settings, approval-card-config) → the route tests exercise the **real validator**, and the 19-case boundary suite imports `SimpleCronExpression` directly un-mocked. The OR-vs-AND contrast test imports the real multitable `parseCronExpression`/`cronHasNoMatchingDay` and pins the divergence on `0 0 30 2 1`.

## Probe 5 — Scope audit: **PASS**

- Exactly 3 files. Route hunks confined to: import block, two module-top helpers + error constant, and the POST/PUT create-update handlers. `sync` / `sync/preview` route regions untouched (no collision with #4049/#4054 territory). No `origin/main` drift on either code file since branch point. Zero new deps (no lockfile change needed). No closing keywords in the PR body (no issue auto-close risk).
- CI gate: the file runs in the **required** `test (20.x)` job (`plugin-tests.yml` → "Run core-backend tests" → `pnpm --filter @metasheet/core-backend test`). At review time: `test (18.x)` pass, `test (20.x)` pending (in-flight run).

## Test + mutation record (all in disposable worktree at head, reverted after; tree clean)

- `tests/unit/admin-directory-routes.test.ts`: **77/77 pass** (matches PR claim).
- Neighbors (env health): `directory-sync-scheduler.test.ts` 5/5, `directory-deprovision-policy.test.ts` 18/18.
- Own probe suite (9 tests): timings above + runtime-parity + timezone-ignored proof, all green.
- **Mutations re-fired — 4/4 kill:**

  | # | Mutation | Result |
  |---|---|---|
  | M1 | `isDirectoryScheduleCronValid`: drop `.hasNext()` (parse-only) | **2 red** (route-level Feb-30 + AND-combo pins) |
  | M2 | Delete POST guard block | **3 red** |
  | M3 | Delete PUT guard block | **1 red** (update-path pin) |
  | M4 | Drop `scheduleCronInput &&` (validate empty too) | **5 red** (empty/whitespace-allowed pins) |

  Guard is fully load-bearing and fully tested. (M1's kill count is 2, not the PR table's 1, because the last commit added the route-level AND-combo pin — strictly better.)

---

## Findings

### P2-1 — Roadmap doc §7.8 documents the ABANDONED validator choice (contradicts the shipped code)
`docs/development/dingtalk-sync-integrated-roadmap-20260708.md` §7.8, the landed bullet added by this PR, says: *"Reuses the multitable automation scheduler's cron grammar — `parseCronExpression` + `cronHasNoMatchingDay` in `packages/core-backend/src/multitable/automation-scheduler.ts` — rather than adding a new cron-parsing dependency."* That is the first-draft implementation the PR itself proved wrong and reverted; the shipped code uses `SimpleCronExpression` (`services/SchedulerService.ts`). **Failure scenario**: this roadmap is the coordination substrate for the DingTalk lanes — the very next §7.8 bullet ("Add timezone support") will be implemented by someone reading this sentence, who would naturally wire the multitable parser and re-open the exact `0 0 30 2 1` accept-but-never-fire hole this PR closed. One-sentence doc fix; fix before/at merge. (Doc-only → P2, not P1.)

### P3-1 — No sweep/telemetry for pre-existing invalid `schedule_cron` rows
Legacy rows saved before this PR keep their invalid cron silently (still warn-and-drop at scheduler level) until an admin next opens the form — at which point any edit 400s until the field is fixed/cleared (FE round-trips the stored value, `DirectoryManagementView.vue:4062`). Behavior is defensible (surfaces reality, error message self-explanatory), but a one-off log line or admin-surface flag for rows failing `isDirectoryScheduleCronValid` would close the loop. Optional follow-up.

### P3-2 — "UTC wall-clock" claim is stronger than the code: `SimpleCronExpression` ignores its timezone argument
`SchedulerService.ts` `matches()` (~L131-137) uses `date.getMinutes()/getHours()/getDate()/getMonth()/getDay()` — **local-time** getters; the `timezone` constructor arg is stored and never read. Empirically proven: `next()` identical for `'UTC'` vs `'Pacific/Kiritimati'` (+14). Validator-runtime identity is unaffected (both sides pass the same ignored arg — they are wrong together, which is the PR's actual guarantee), and prod containers run TZ=UTC so the claim coincides operationally. But the 400 message / comments / roadmap say "always runs on UTC wall-clock time", which is really "server-process-local time". Suggest softening the wording now or folding into the timezone follow-up. Pre-existing parser behavior → P3.

### P3-3 — (Adjacent, pre-existing, NOT this PR) `setTimeout` overflow makes far-future crons fire immediately in a loop
`JobScheduler.scheduleCronJob` passes `delayMs = nextRun - now` straight to `setTimeout`. Node clamps delays > 2^31-1 ms (~24.86 days) to 1ms — a **validated, legitimate** cron whose next fire is >24.86 days out (e.g. `0 0 1 1 *` saved in July: ~175 days ≈ 1.5e10 ms) would execute immediately, re-schedule, and hot-loop sync attempts gated only by the sync lease / `isRunning`. This PR's validation correctly accepts such crons; the defect is in the runtime scheduler and predates the PR. Flag to the #4049/#4054 scheduler lane; do not fix here.

### NIT-1 — `normalizeScheduleCronInput` duplicates `directory-sync.ts`'s private `normalizeText`
Byte-identical logic, comment-pinned as a deliberate mirror. Exporting the original (or a shared util) would remove the drift risk. Cosmetic.

### NIT-2 — Lenient range parsing accepted-and-consistent
`50-70 * * * *` (upper bound past field max) is accepted by both validator and runtime (fires at minutes 50-59). Divergent from strict cron but identity holds; the PR's honest-gaps section already discloses the range-parsing quirk. No action.

---

## Lock/brief gate table

| Gate (lane brief / roadmap §7.8) | Verdict | Evidence |
|---|---|---|
| Save-time validation on POST + PUT | **PASS** | Both handlers gated before service call; M2/M3 mutations red |
| Validate with the directory scheduler's real parser | **PASS** | Same class, same args; empirical schedule() parity; brief's "multitable parser" instruction correctly overridden (deviation disclosed) |
| 400 + stable error code on invalid/never-fires | **PASS** | `DIRECTORY_SCHEDULE_CRON_INVALID`; never-fires proven via hasNext, M1 red |
| Empty/null/whitespace = no schedule, allowed | **PASS** | 4 create cases + 1 update case pinned; M4 red |
| No FE change needed for error surfacing | **PASS (verified)** | `readApiError` + `setStatus` generic path in `saveIntegration()` |
| Timezone support NOT built | **PASS** | No timezone config added; doc marks bullet open (but see P3-2 wording) |
| No drift into sync/preview regions (#4049/#4054) | **PASS** | Hunk audit; no main drift on touched files |
| Doc updated accurately | **FAIL → P2-1** | §7.8 bullet names the abandoned multitable parser |

**Bottom line**: runtime change is correct, minimal, mutation-hardened, and the validator-choice reversal is genuinely the right call. Fix the one stale sentence in the roadmap doc (P2-1) and this is clean to land.

*Reviewed by Claude (Fable 5) adversarial gate, 2026-07-10. Worktree `/tmp/ms2-rev-pr4053` removed after use; no changes pushed.*

zensgit:
Gate P2-1 + P3-2 closed on this head: §7.8 now names `SimpleCronExpression` identity as the load-bearing property (with an explicit do-not-swap warning recording the abandoned multitable-parser draft and the `0 0 30 2 1` divergence), and the UTC wall-clock claim is corrected to host-local semantics shared by validator and runtime. P3-1 (legacy invalid-row sweep) and the adjacent pre-existing P3-3 (setTimeout >2^31-1ms clamp → far-future cron hot-loops behind the lease) go to the pool backlog, not this PR.

zensgit:
Window protocol executed: rebased onto post-#4054 main (`4b7c4cb61`, clean, 5 commits survived); routes suite 78/78 + tsc clean; load-bearing mutation replayed at the real code line (hasNext() check dropped ⇒ RED, restored). Armed alone — Wave 2 window 3.

PR #4054, opened 2026-07-10
feat(directory): per-run sync change summary persisted into directory_sync_runs.stats (R5)

zensgit:
## What / why

R5 from the DingTalk backlog pool: a completed directory sync run's `stats` jsonb previously conflated "0 new" and "500 new" (`accountsSynced` is just the pull size), never recorded what THIS run deactivated, and manager-binding coverage (#3914) was live-only with no per-run trend. This persists a complete, honest per-run change summary — **all-additive, no schema change** (`directory_sync_runs.stats` is already jsonb), **no route change** (`GET /integrations/:id/runs` passes `stats` through `summarizeRun` verbatim — verified by test), **no FE change** (the run panel reads defensively via `readNumericStat`; rendering the new keys is a separate FE slice).

New stats keys written at the single completion write point, merged into the existing stats object with every pre-existing key kept:

- `accountsCreatedCount` / `accountsUpdatedCount`, `departmentsCreatedCount` / `departmentsUpdatedCount` — via `RETURNING (xmax = 0) AS created` on both upserts (freshly INSERTed tuple has xmax 0; ON CONFLICT DO UPDATE carries the updating xid — standard PostgreSQL discriminator, pg-only repo). Read defensively (`rows[0]?.created === true`) so unit-suite fake clients answering `{ rows: [] }` don't throw.
- `accountsDeactivatedCount` — from the DT-OPS-01 transition sweep's ids already in memory.
- `departmentsDeactivatedCount` — via `RETURNING id` on the departments sweep.
- `managerCount` / `linkedManagerCount` / `managerCoverage` — `getDirectoryManagerBindingCoverage` run through the **transaction client** (queryFn was already injectable), so the snapshot sees this run's final (uncommitted) link state; end-of-run placement after link loop + member groups + deprovision.
- `durationMs` — lease claim → completion write, app-side clock only (no app/DB clock-skew arithmetic).

## Behavior change flagged for ops (the one non-additive edit)

The departments deactivation sweep now carries `AND is_active = true`, mirroring the DT-OPS-01 account sweep: it is a TRANSITION (what departed this run), not a re-stamp of every long-departed department on every sync. End state is identical for already-inactive rows; the only observable difference is `directory_departments.updated_at` is no longer re-stamped on them — grepped the backend: nothing reads that column for logic. Without this the per-run count would report the same departed department forever (run-4 test pins this).

## Tests (real DB required — xmax + ON CONFLICT cannot be mocked honestly)

Extended the already-whitelisted `directory-sync-alert-coverage.db.test.ts` (both wiring points already exist: vitest.config.ts exclude + plugin-tests.yml whole-file entry — **zero new CI wiring**). Chose it over the parity suite deliberately: #4049 (Wave 2, lands before this) touches `directory-sync-preview-apply-parity.db.test.ts`, so extending the alert-coverage file keeps the eventual 3-way merge clean. All source edits sit in the `syncDirectoryIntegration` region (ends ~line 3400); #4049's hunks start at the preview doc-comment (~3476) — no overlap.

New block drives the REAL `syncDirectoryIntegration` 4 times (mocked DingTalk pull, real Postgres apply) over an evolving directory:

- run 1 (seeded: 1 pre-existing linked manager account + 2 new pulled users): `created=2, updated=1`, `managerCount=2 / linkedManagerCount=1 / managerCoverage=0.5` equal to the live #3914 function's own numbers, `durationMs >= 0`, pre-existing keys (`linkedCount`, `unmatchedCount`, `deprovisionApplied`, …) survive the merge, and the new keys arrive through `listDirectorySyncRuns` (the GET /runs surface) with no route change
- run 2 (one user departs): `created=0, updated=2, accountsDeactivatedCount=1` + DB cross-check that the account really is inactive
- run 3 (one department departs): `departmentsDeactivatedCount=1`, already-inactive account NOT re-counted
- run 4 (steady state): both deactivation counts 0 — transition semantics pinned

Local evidence on a fresh migrated DB (`lane_perrunsummary_*`): extended suite 11/11; all 8 whitelisted directory real-DB suites 64/64 (gate-corrected arithmetic across the same 8 whitelisted files); full `tests/unit` 4702/4702; `tsc --noEmit` clean.

## Mutation table (each applied to the committed impl, suite re-run, then restored)

| # | Mutation | Result |
|---|---|---|
| M1 | Drop xmax discriminator on account upsert (count every upsert as created) | RED — run 1 (`created 3≠2`), run 2 + run 4 (`created 2≠0`) |
| M2 | Skip the stats merge write (new keys never persisted) | RED — all 4 run tests |
| M3 | Swap coverage numerator/denominator in `getDirectoryManagerBindingCoverage` | RED — existing DT-OPS-03 test + run 1 + run 4 |
| M4 | Drop `AND is_active = true` from departments sweep | RED — run 4 (`departmentsDeactivatedCount 1≠0`) |
| M5 | Hardcode `accountsDeactivatedCount: 0` | RED — run 2 |
| M6 | Miswire coverage call to global pool `query` instead of the transaction client | RED — run 1 (`managerCount 0≠2`: snapshot must see uncommitted in-transaction state) |

## Not proven / honest gaps

- **Failed runs** still leave `stats = '{}'` (only `error_message`) — pre-existing `markSyncFailure` behavior, unchanged and untested here; accepted by the brief.
- `durationMs` excludes the completion UPDATE itself and all post-commit work (invite ledger, deprovision audit) — it measures claim → just-before-completion-write.
- FE rendering of the new keys is **not** in this PR; old run rows lack the keys and `readNumericStat` renders them as 0 — the future FE slice should gate on key presence to distinguish "absent (pre-R5 run)" from a true 0.
- The manager-coverage query adds one heavier CTE query per run inside the write transaction (marginal lengthening; it is the point of the snapshot).
- xmax discriminator is PostgreSQL-specific — fine for this pg-only repo, flagged for the record.
- The new tests' `it` blocks are order-dependent (run 1 → 4), matching the file's existing style.

## Sequencing

Wave 2 of the directory-sync serial queue: lands **after the Wave 1 queue and after #4049** in the Wave 2 serial order. Do not arm auto-merge; landing is gated and serialized by the main loop.

### DIFF
--- a/packages/core-backend/src/directory/directory-sync.ts
+++ b/packages/core-backend/src/directory/directory-sync.ts
@@ -31,7 +31,7 @@ import { invalidateUserPerms } from '../rbac/service'
 import { decryptStoredSecretValue, normalizeStoredSecretValue } from '../security/encrypted-secrets'
 import { getBcryptSaltRounds } from '../security/auth-runtime-config'
 import { SimpleCronExpression } from '../services/SchedulerService'
-import { deliverDirectorySyncFailureAlert } from './directory-sync-alert-delivery'
+import { deliverDirectorySyncFailureAlert, getDirectoryManagerBindingCoverage } from './directory-sync-alert-delivery'
 
 const logger = new Logger('DirectorySync')
 const DEFAULT_ORG_ID = 'default'
@@ -2856,6 +2856,10 @@ export async function syncDirectoryIntegration(
   await reclaimStaleDirectorySyncRuns(integrationId)
   const runResult = await claimDirectorySyncRun(integrationId, triggeredBy, triggerSource)
   const runId = runResult.rows[0].id
+  // R5: wall-clock anchor for stats.durationMs. Measured app-side from the lease claim to
+  // just before the completion UPDATE (stats is serialized inside the transaction), so it
+  // is immune to app/DB clock skew that started_at/finished_at arithmetic would inherit.
+  const runStartedAtMs = Date.now()
   hooks.onRunStarted?.(runId)
 
   // DT-HARDEN-05 heartbeat: prove this run is alive for as long as it holds the lease.
@@ -2919,8 +2923,16 @@ export async function syncDirectoryIntegration(
     const autoAdmissionOnboardingPackets: DirectoryAutoAdmissionOnboardingPacket[] = []
 
     await transaction(async (client) => {
+      // R5: created-vs-updated split for the run summary. `departmentsSynced`/`accountsSynced`
+      // conflate "0 new" and "500 new"; the discriminator is `(xmax = 0)` on the upserted row —
+      // a freshly INSERTed tuple has xmax 0, an ON CONFLICT DO UPDATE tuple carries the updating
+      // transaction's xid (pg-only repo, standard PostgreSQL trick). Read defensively
+      // (`rows[0]?.created === true`) so a fake client answering empty rows in unit suites
+      // falls into the "updated" bucket instead of throwing.
+      let departmentsCreatedCount = 0
+      let departmentsUpdatedCount = 0
       for (const department of departments.values()) {
-        await client.query(
+        const departmentUpsert = await client.query(
           `INSERT INTO directory_departments (
              integration_id, provider, external_department_id, external_parent_department_id, name,
              full_path, order_index, is_active, raw, last_seen_at, created_at, updated_at
@@ -2935,7 +2947,8 @@ export async function syncDirectoryIntegration(
              is_active = true,
              raw = EXCLUDED.raw,
              last_seen_at = EXCLUDED.last_seen_at,
-             updated_at = NOW()`,
+             updated_at = NOW()
+           RETURNING (xmax = 0) AS created`,
           [
             integrationId,
             DEFAULT_PROVIDER,
@@ -2953,12 +2966,16 @@ export async function syncDirectoryIntegration(
             syncTimestamp,
           ],
         )
+        if (departmentUpsert.rows[0]?.created === true) departmentsCreatedCount += 1
+        else departmentsUpdatedCount += 1
       }
 
+      let accountsCreatedCount = 0
+      let accountsUpdatedCount = 0
       const userList = Array.from(users.values())
       for (const user of userList) {
         const externalKey = normalizeText(user.unionId || user.openId || user.userId)
-        await client.query(
+        const accountUpsert = await client.query(
           `INSERT INTO directory_accounts (
              integration_id, provider, corp_id, external_user_id, union_id, open_id, external_key,
              name, nick, email, mobile, job_number, title, avatar_url, is_active, raw, last_seen_at, created_at, updated_at
@@ -2979,7 +2996,8 @@ export async function syncDirectoryIntegration(
              is_active = true,
              raw = EXCLUDED.raw,
              last_seen_at = EXCLUDED.last_seen_at,
-             updated_at = NOW()`,
+             updated_at = NOW()
+           RETURNING (xmax = 0) AS created`,
           [
             integrationId,
             DEFAULT_PROVIDER,
@@ -2999,14 +3017,24 @@ export async function syncDirectoryIntegration(
             syncTimestamp,
           ],
         )
+        if (accountUpsert.rows[0]?.created === true) accountsCreatedCount += 1
+        else accountsUpdatedCount += 1
       }
 
-      await client.query(
+      // R5, mirroring the DT-OPS-01 account sweep below: `AND is_active = true` makes this a
+      // TRANSITION (departments that departed in THIS run), not a re-stamp of every
+      // long-gone department on every sync — without it the per-run count would report the
+      // same departed department forever. The resulting state is identical for already-
+      // inactive rows (nothing reads directory_departments.updated_at); only the honest
+      // count is new.
+      const deactivatedDepartmentsResult = await client.query(
         `UPDATE directory_departments
          SET is_active = false, updated_at = NOW()
-         WHERE integration_id = $1 AND last_seen_at < $2`,
+         WHERE integration_id = $1 AND last_seen_at < $2 AND is_active = true
+         RETURNING id`,
         [integrationId, syncTimestamp],
       )
+      const departmentsDeactivatedCount = deactivatedDepartmentsResult.rows.length
       // DT-OPS-01: `AND is_active = true` makes this a TRANSITION, not a re-stamp of the
       // whole backlog. Without it the deprovision executor re-processed every account ever
       // deactivated on every sync — audit spam, and it would stomp a reactivation.
@@ -3314,9 +3342,32 @@ export async function syncDirectoryIntegration(
         enabled: isDirectoryDeprovisionEnabled(),
       })
 
+      // R5: per-run manager-binding snapshot. The live GET /manager-coverage endpoint
+      // (#3914) answers "what is coverage NOW"; persisting the same numbers here answers
+      // "what did THIS run leave it at", so admins can see the trend across runs. Runs
+      // through the transaction client on purpose: it must see this run's (uncommitted)
+      // final link state, and `getDirectoryManagerBindingCoverage`'s queryFn is injectable
+      // for exactly this. Placed after the link loop, member-group projection and
+      // deprovision so it is the end-of-run snapshot.
+      const managerBindingCoverage = await getDirectoryManagerBindingCoverage(
+        integrationId,
+        (sql, params) => client.query(sql, params),
+      )
+
       const stats = {
         departmentsSynced: departments.size,
         accountsSynced: users.size,
+        // R5 per-run change summary — all additive; every pre-existing key above/below is kept.
+        departmentsCreatedCount,
+        departmentsUpdatedCount,
+        departmentsDeactivatedCount,
+        accountsCreatedCount,
+        accountsUpdatedCount,
+        accountsDeactivatedCount: deactivatedAccountIds.length,
+        managerCount: managerBindingCoverage.managerCount,
+        linkedManagerCount: managerBindingCoverage.linkedManagerCount,
+        managerCoverage: managerBindingCoverage.coverage,
+        durationMs: Date.now() - runStartedAtMs,
         deprovisionApplied: deprovisionOutcome.applied,
         deprovisionCandidateCount: deprovisionOutcome.candidateCount,
         deprovisionManualReviewCount: deprovisionOutcome.manualReviewCount,
--- a/packages/core-backend/tests/integration/directory-sync-alert-coverage.db.test.ts
+++ b/packages/core-backend/tests/integration/directory-sync-alert-coverage.db.test.ts
@@ -1,9 +1,34 @@
-import { afterAll, beforeAll, describe, expect, it } from 'vitest'
+import { afterAll, beforeAll, describe, expect, it, vi } from 'vitest'
+
+// R5 (per-run summary block below): the DingTalk HTTP client is mocked so real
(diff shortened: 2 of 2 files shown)


zensgit:
# Adversarial review — PR #4054 (R5: per-run sync summary persisted into directory_sync_runs.stats)

- **PR**: zensgit/metasheet2#4054 — `claude/dt-per-run-summary`
- **Head reviewed**: `be3525364e40c11e12a4009eaa8308271e837e19` (matches reported head; 1 commit over merge-base `bdf37fb08`)
- **Reviewer**: Claude (Fable 5), independent adversarial gate, 2026-07-10
- **Method**: refute-first. Own disposable worktree at PR head, `pnpm install --frozen-lockfile`, fresh migrated PG per mutation run (4 fresh databases total), real vitest runs via `vitest.integration.config.ts`, 2 re-fired mutations + 1 reviewer-original mutation, full unit suite, tsc, repo-wide reader sweeps for every touched column/table, hunk-level overlap audit vs unlanded sibling #4049.

## Verdict: **APPROVE** (no P1, no P2; 3×P3 + 2×NIT, none merge-blocking)

Scope is exactly the R5 lane brief's proposedFix (a)–(d): xmax created/updated discriminators on both upserts, per-run deactivation transition counts, in-transaction manager-coverage snapshot, durationMs — all written at the single completion write point. The FE run-card extension from the brief is openly deferred. Every load-bearing guard I attacked went RED.

## Probe results (parent's 6 probes)

### Probe 1 — the flagged non-additive edit (`AND is_active = true` on the departments sweep)
- **Transition-only semantics**: TRUE. Verified against the real DB: run 3 counts the departing department exactly once; run 4 (steady state) counts 0. My own mutation (predicate dropped) → run 4 RED (`expected 1 to be +0`).
- **End-state identical**: TRUE for `is_active`; **FALSE for `updated_at`** on already-inactive rows — they are no longer re-stamped each sync (frozen at real deactivation time).
- **"Nothing reads directory_departments.updated_at"**: **FALSE as stated.** Repo-wide grep: no backend logic branches on it, but the admin FE *displays* it in three places: `apps/web/src/views/DirectoryManagementView.vue:876` (per-department "最近同步：{{ formatDateTime(department.updatedAt) }}"), `:527` ("最近更新" chip via `departmentMirrorLatestUpdatedAt`), `:3151` (diagnostics copy text). All display-only, zero branching (`isDepartmentInLatestMirrorBatch` and the stale-batch warning logic use `lastSeenAt`, not `updatedAt`). The frozen timestamp on departed departments is arguably *more* truthful, and mirrors the accounts precedent (DT-OPS-01 made the identical trade for accounts, whose `updatedAt` is likewise displayed at `:1634`). → **P3-1** (claim inaccuracy, benign observable UI change).
- **rowCount readers beyond the new stats**: none. On main the sweep's result was discarded (`main:directory-sync.ts:3005-3009`); no other consumer exists.

### Probe 2 — xmax discriminator correctness
- Both discriminators sit on `ON CONFLICT … DO UPDATE` upserts (PR `directory-sync.ts:2941/3000` departments, `:2984/3000-area` accounts). DO UPDATE always writes (no conditional WHERE on the conflict action), so RETURNING always yields one row and updated tuples carry xid → `xmax ≠ 0`. **No `DO NOTHING` on these tables anywhere** — the only `ON CONFLICT DO NOTHING` statements in the module are `user_roles` (`:4586`) and `platform_member_group_members` (`:4743`), neither feeds created/updated counts.
- Defensive read verified: `rows[0]?.created === true` else the updated bucket — unit fake clients answering `{ rows: [] }` don't throw; full unit suite 4702/4702 green in the worktree confirms no module-graph or fake-client breakage.
- Real-DB proof: run 1 splits pre-seeded conflicting account (updated=1) from fresh inserts (created=2). Mutation M1 re-fired (discriminator → `RETURNING true AS created`): **RED** — run 1 `expected 3 to be 2`, run 2 & run 4 `expected 2 to be +0`. Matches implementer's table.

### Probe 3 — coverage-in-transaction semantics
- `getDirectoryManagerBindingCoverage` (`directory-sync-alert-delivery.ts:175-227`) has an injectable `queryFn` (default = global pool); PR passes `(sql, params) => client.query(sql, params)`. Placement after link loop + member-group projection + deprovision = end-of-run state; under READ COMMITTED this equals the post-commit state the live endpoint would serve (barring unrelated concurrent admin writes, which is inherent to any snapshot).
- `GET /manager-coverage` (`admin-directory.ts:500`) calls the same function with the default pool — the test's `snapshot === live` assertion is a genuine equivalence check against the endpoint's exact computation. Ran it: green.
- Mutation M6 re-fired (drop the injected queryFn → global pool inside the open transaction): **RED** — run 1 `managerCount expected +0 to be 2`. This proves the in-tx wiring is load-bearing: the global pool cannot see the transaction's uncommitted departments/accounts, so a miswire zeroes the snapshot. Exactly the failure the implementer claimed.
- Noted (P3-3): the coverage SQL now executes inside the apply transaction, so a pathological error there fails the whole sync run — a new (tiny) failure channel for a stats-only feature. The SQL is type-guarded (`jsonb_typeof` checks) and its inputs are produced by our own sync; risk accepted, note only.

### Probe 4 — stats merge cannot clobber
- Exhaustive writer sweep of `directory_sync_runs.stats`: exactly two writers in the entire repo — the lease-claim INSERT (`'{}'::jsonb`, `:2820`) and the single completion UPDATE (`stats = $2::jsonb`, `:3416-3428`). All other references (alert-delivery `:63`, list/summary queries `:3689-3910`) are reads.
- The completion write serializes ONE object literal containing every pre-existing key (departmentsSynced … memberGroup* counters, `...directoryDiagnosticStats`). New R5 key names collide with nothing, including the diagnostics keys (`rootDepartment*`, `diagnostic*`, `sampled*`). Test pins survival of `linkedCount/pendingCount/unmatchedCount/autoAdmittedCount/deprovisionApplied` — green, and implementer's M2 (skip merge) is credible (all 4 run tests assert new keys).
- NIT-1: the `...directoryDiagnosticStats` spread sits *after* the R5 keys, so a future diagnostics key named e.g. `durationMs` would silently win. Not a today-bug.

### Probe 5 — failed runs / lease interaction
- `markSyncFailure` (`:2541-2568`) and `reclaimStaleDirectorySyncRuns` (`:2767-2791`) touch `status/finished_at/error_message/updated_at` only — stats stays `'{}'` on failure, untouched by this PR (diff confirms zero hunks in those functions). Heartbeat UPDATE (`:2872-2878`) writes `last_heartbeat_at/updated_at` only. The completion write keeps its `AND status = 'running'` lease-loss guard (`:3421-3427`) — unchanged. `durationMs` is app-side (`Date.now()` at claim → stats construction), no interaction with lease timestamps. Clean.

### Probe 6 — overlap vs unlanded #4049
- #4049's `directory-sync.ts` hunks: base lines 3476 (preview doc-comment + sentinel const), 3510, 3592 — all inside `previewDirectorySyncIntegration`. #4054's last hunk ends ~base 3345 (inside `syncDirectoryIntegration`). **>130 lines of untouched context between them; zero hunk overlap, zero semantic interaction** (apply-path stats vs preview-path candidate simulation). Test files disjoint (#4049 → parity suite; #4054 → alert-coverage suite — the PR body says this file choice was deliberate for exactly this reason, and it holds). Merges clean in either order. Also: origin/main is 2 commits ahead of the PR base with **zero diff** in either touched file.

## Lane-brief (governing spec) gate table

| Brief item | Verdict | Evidence |
|---|---|---|
| (a) xmax discriminator on account upsert, "same for department if cheap" | PASS | Both upserts, `:2941/:2984`; M1 RED |
| (b) accountsDeactivatedCount + departmentsDeactivatedCount via RETURNING | PASS | `:3027-3037`; my predicate mutation RED at run 4 |
| (c) coverage via injectable queryFn inside the transaction, after link loop | PASS | `:3350-3354`; M6 RED |
| (d) durationMs | PASS | `:2859/:3373`; asserted `>= 0` (weak but harmless — a stronger bound isn't reliably assertable) |
| No schema change / no route change | PASS | jsonb reused; `GET /runs` passthrough pinned by the `listDirectorySyncRuns` assertion |
| FE run-card extension (新增/更新/停用 + coverage lines, absent-vs-0 rendering) | **DEFERRED** (declared) | PR body: "separate FE slice" → P3-2 |
| Test strategy: extend already-whitelisted real-DB suite, falsifiable mutations | PASS | Extended `directory-sync-alert-coverage.db.test.ts`; two-point wiring pre-exists (vitest.config.ts:118 exclude + plugin-tests.yml:403 whitelist); the whitelist step runs inside `test (20.x)` of Plugin System Tests, which is green on this PR — the new tests are genuinely CI-gated |
| Risk: serialize vs other R-items on the hot file | PASS | #4049 audit above |

## Test & mutation evidence (my own runs, fresh DBs on :5435)

| Run | Result |
|---|---|
| Extended suite `directory-sync-alert-coverage.db.test.ts` | **11/11** green |
| Other 7 whitelisted directory/dingtalk real-DB suites | **53/53** green (env healthy; module-level DingTalk mock does not bleed — vi.mock is per-file) |
| Full `tests/unit` in worktree | **4702/4702** green |
| `tsc --noEmit` (core-backend) | clean |
| **M1 re-fired**: accounts discriminator → `true AS created` | **RED** ×3 (run 1/2/4) |
| **M6 re-fired**: coverage → global pool | **RED** ×1 (run 1, managerCount 0≠2) |
| **MX (reviewer-original)**: drop `AND is_active = true` from departments sweep | **RED** ×1 (run 4, deactivated 1≠0) |

All mutations restored via `git checkout --` in the disposable worktree; final `git status` clean.

## Findings

- **P3-1** — `packages/core-backend/src/directory/directory-sync.ts:3021-3037` + PR body: the ops flag says "nothing reads directory_departments.updated_at" — the admin UI displays it (`DirectoryManagementView.vue:876/:527/:3151`). Observable change: a departed department's "最近同步" freezes at deactivation time instead of advancing every sync. Display-only, no branching, more truthful, mirrors the accounts precedent — no code change needed; correct the ops-note wording wherever this ships.
- **P3-2** — R5 brief scope: "exposed to admins" (FE run-card rendering of the new keys, incl. the brief's absent-vs-0 requirement for pre-fix runs) is deferred. Keep the FE slice in the backlog pool or R5 is only half-delivered.
- **P3-3** — `directory-sync.ts:3350-3354`: coverage SQL inside the apply transaction is a new sync-failure channel (previously endpoint-only). Type-guarded and producer-controlled inputs; accept, no action.
- **NIT-1** — `directory-sync.ts:3399`: `...directoryDiagnosticStats` spread after the R5 keys; a future colliding diagnostics key would silently clobber. Consider a one-line comment or moving the spread first in a later slice.
- **NIT-2** — PR body claims "49/49 across all 8 whitelisted directory real-DB suites"; my run of the same 8 files at the same head totals **64/64** (11+53). Everything passes — the discrepancy is evidence arithmetic, not behavior.

## Not proven / accepted residuals
- `(xmax = 0)` is undocumented-but-universal PostgreSQL behavior (pg-only repo, acknowledged in code comment); proven empirically here on the repo's PG 5435.
- Failed-run stats staying `'{}'` is verified structurally (no code change on those paths), not by a new test.
- `durationMs > 0` tightness not asserted (only `>= 0`) — fine.

zensgit:
Window protocol executed: rebased onto post-#4049 main (clean); fresh-DB alert-coverage suite 11/11 + tsc clean; load-bearing mutation replayed at the REAL call site (coverage queryFn miswired to global `query` instead of the transaction client ⇒ RED ×1, restored — first probe hit the import line and was discarded as a dud, per this line's probe-must-hit-real-code discipline). Armed alone — Wave 2 window 2. NOTE for #4057's later window: it collides with this PR in the same import block + EOF of the alert-coverage test; resolve as UNION.

PR #4057, opened 2026-07-10
feat(directory): §7.1 inactive-linked-for-N-days metric + post-run alert

zensgit:
## Summary

Roadmap §7.1 (Offboarding Policy Executor) target bullet 3: *"Add inactive linked for N days metric and alert."* Only the `inactive_linked` review-item **filter** existed (`directory-sync.ts` `buildReviewFilterSql` + the review-items admin endpoint/UI filter) — no age dimension and no alert. Because `DIRECTORY_DEPROVISION_ENABLED` (DT-OPS-01) is default-off by owner decision, a directory account going `is_active = false` never touches the linked local user, so this backlog was invisible. This PR adds the metric and a gated alert, read-only, no deprovision writes.

- **`getDirectoryInactiveLinkedMetric(integrationId, thresholdDays)`** (`packages/core-backend/src/directory/directory-sync-alert-delivery.ts`, module scope near `getDirectoryManagerBindingCoverage`): count + oldest-first sample (capped 20) of directory accounts that are `is_active = false`, linked (`directory_account_links.local_user_id IS NOT NULL` — matches the review-item's own definition, independent of `link_status`), local user still `is_active = true`, and inactive at least N days.
- **GET `/integrations/:integrationId/inactive-linked?days=N`** admin route in `admin-directory.ts` (default 30, clamped to [1, 3650]), mirroring the existing `/manager-coverage` route exactly (same admin gate, same error-mapping shape).
- **`deliverDirectoryInactiveLinkedAlert`**: best-effort DingTalk webhook digest, reusing the `DIRECTORY_SYNC_ALERT_WEBHOOK` channel and gated by a new **`DIRECTORY_INACTIVE_LINKED_ALERT_DAYS`** threshold env (unset = off, following the existing env-gate side-channel convention). Never throws; only sends when the backlog is nonzero.
- Wired as a **post-run hook in `directory-sync-scheduler.ts`'s `runScheduledSync`** (scheduler-triggered/nightly syncs only — see "Decisions" below).

## Deactivation-time anchor (read before reviewing the SQL)

There is no dedicated deactivation-transition column in this schema. I grepped every write site that can touch `directory_accounts` across `packages/core-backend/src` — there are exactly two, both in `directory-sync.ts`:

1. The re-sync upsert (`INSERT ... ON CONFLICT DO UPDATE SET is_active = true, ..., updated_at = NOW()`) — only fires for accounts currently present in the DingTalk pull, and always sets `is_active` back to `true` (which removes the row from this metric's target set).
2. The deactivation sweep (`UPDATE directory_accounts SET is_active = false, updated_at = NOW() WHERE ... AND is_active = true`) — guarded by `AND is_active = true`, i.e. a genuine one-way transition, not a re-stamp. Once inactive, this statement's WHERE clause no longer matches the row, so it can't bump `updated_at` again while the row stays inactive.

No other module issues an UPDATE/INSERT against `directory_accounts` (every other reference is a read-only JOIN). So for a row currently `is_active = false`, `directory_accounts.updated_at` is exactly the deactivation timestamp — an honest anchor given the current schema, not an approximation. **No migration added or needed.**

## Coordination / scope decisions

- `directory-sync.ts` is being touched by unlanded #4049 (preview region) and #4054 (apply region ~2860-3370). The new metric function lives entirely in `directory-sync-alert-delivery.ts`; `directory-sync.ts` itself is **untouched** by this PR.
- The post-run alert hook is wired in `directory-sync-scheduler.ts`'s `runScheduledSync`, **not** in the manual-sync route (`admin-directory.ts` `/integrations/:id/sync`) or inside `syncDirectoryIntegration`'s apply/preview bodies. This means the digest only fires after **scheduler-triggered (nightly) syncs**, not manual ones. Manual-sync visibility still exists via the new on-demand endpoint. Flagging this narrowing for reviewer scrutiny — it was a deliberate collision-avoidance choice given #4049/#4054 are in flight on the same file.
- `admin-directory.ts` is touched by unlanded #4053 on the create/update handlers; this PR only adds a new GET route and does not touch any existing handler.
- The alert is a pure outbound webhook POST — no new row written to `directory_sync_alerts` (no new persistence semantics/migration introduced).

## Tests

- `packages/core-backend/tests/unit/directory-sync-alert-delivery.test.ts` — new `§7.1 directory inactive-linked backlog digest` describe block: env-gate unit tests (unset/non-integer/negative/valid), message-builder cap-at-5 test, delivery wiring tests (both gates required, empty-backlog no-send, successful send, webhook-failure swallow, DingTalk business-error swallow) using a fake `queryFn` (SQL-shape discriminated, matching the existing `getDirectoryManagerBindingCoverage` unit-test style — real SQL is proven in the `.db.test.ts` below).
- `packages/core-backend/tests/integration/directory-sync-alert-coverage.db.test.ts` — new self-contained describe block **appended at the end of the file** (per lane brief, since #4054 also extends this file — keeps the 3-way merge trivial). Real-Postgres fixtures: past-threshold (20d), exact boundary (10d, inclusive `<=`), short-of-threshold (5d, 1d), unlinked, linked-but-locally-inactive-user, linked-but-account-still-active, and a second integration to prove scoping. Both new tests pass on a fresh migrated DB.
- `packages/core-backend/tests/unit/admin-directory-routes.test.ts` — new route tests (200 with metric shape, day-param default/invalid-fallback, 403 non-admin, 500 on failure), plus wired `getDirectoryInactiveLinkedMetric` into the existing whole-module mock factory (would otherwise return `undefined` at call time once the route imports it).
- `packages/core-backend/tests/unit/directory-sync-scheduler.test.ts` — new assertions: digest called with `{integrationId, integrationName}` after a successful scheduled sync; **not** called on lease-conflict skip or on a real sync failure.
- Full `tests/unit` suite: **353 files / 4767 tests passed**. `tsc --noEmit`: clean. Real-DB whitelisted file (`directory-sync-alert-coverage.db.test.ts`) + two sibling directory `.db.test.ts` files: all green on a fresh migrated database (dropped after each run).
- No new two-point wiring needed: `directory-sync-alert-coverage.db.test.ts` was already excluded in `vitest.config.ts` and whitelisted in `plugin-tests.yml`'s real-DB step; this PR only extends that file.

## Mutation table

| Mutation | Result |
|---|---|
| Invert age comparison (`<=` → `>=` in both count/sample queries) | RED — both new `.db.test.ts` assertions fail |
| Drop the linked-join (query `directory_accounts` alone, no join to links/users) | RED — count assertion fails (unlinked fixture leaks in) |
| Drop `AND u.is_active = TRUE` | RED — count assertion fails (locally-inactive-user fixture leaks in) |
| Remove the `if (!thresholdDays) return false` env-gate in `deliverDirectoryInactiveLinkedAlert` | RED — exactly the targeted unit test ("sends nothing when the threshold env is unset") fails; used a fake `queryFn` that reports a nonzero backlog regardless of args, so the guard removal is provably load-bearing (not masked by a redundant inner guard) |

All four mutations applied one at a time (real implementation committed first), confirmed RED, then reverted (`git checkout --`) before moving to the next.

## Honest gaps

- The digest only fires after scheduler-triggered syncs, not manual ones (see Coordination above) — a manual-sync operator only sees the backlog via the new GET endpoint, not a push notification.
- No persisted `directory_sync_alerts` row for this alert type — it's delivery-only, so there's nothing to acknowledge/list in the existing alerts admin UI (only the webhook digest and the on-demand metric endpoint exist).
- `getDirectoryInactiveLinkedMetric`'s sample is capped at 20 rows; the message builder further caps at 5 examples per digest. Not configurable in this PR.
- No frontend UI changes — brief scoped this to backend metric + endpoint + alert only.

### DIFF
--- a/packages/core-backend/src/directory/directory-sync-alert-delivery.ts
+++ b/packages/core-backend/src/directory/directory-sync-alert-delivery.ts
@@ -367,12 +367,16 @@ export function buildDirectoryInactiveLinkedAlertMessage(options: {
     `- 阈值：${metric.thresholdDays} 天`,
     `- 数量：${metric.count}`,
   ]
-  for (const item of metric.sample.slice(0, 5)) {
+  const shown = Math.min(metric.sample.length, 5)
+  for (const item of metric.sample.slice(0, shown)) {
     const label = item.localUserName || item.localUserEmail || item.localUserId
     lines.push(`  - ${item.accountName || item.externalUserId} → ${label}（已停用 ${item.inactiveDays} 天）`)
   }
-  if (metric.count > metric.sample.length) {
-    lines.push(`  - ……以及其他 ${metric.count - Math.min(metric.sample.length, 5)} 个账号`)
+  // `metric.sample` is already capped (INACTIVE_LINKED_SAMPLE_LIMIT) upstream, so
+  // `count` can exceed `sample.length` even before this message's own 5-item cap —
+  // compare against what was actually SHOWN here, not the upstream sample size.
+  if (metric.count > shown) {
+    lines.push(`  - ……以及其他 ${metric.count - shown} 个账号`)
   }
   return { subject, content: lines.join('\n') }
 }
--- a/packages/core-backend/src/directory/directory-sync-alert-delivery.ts
+++ b/packages/core-backend/src/directory/directory-sync-alert-delivery.ts
@@ -220,3 +220,228 @@ export async function getDirectoryManagerBindingCoverage(
     coverage: managerCount === 0 ? 1 : linkedManagerCount / managerCount,
   }
 }
+
+export type DirectoryInactiveLinkedSample = {
+  directoryAccountId: string
+  externalUserId: string
+  accountName: string | null
+  localUserId: string
+  localUserEmail: string | null
+  localUserName: string | null
+  inactiveSinceAt: string
+  inactiveDays: number
+}
+
+export type DirectoryInactiveLinkedMetric = {
+  thresholdDays: number
+  count: number
+  sample: DirectoryInactiveLinkedSample[]
+}
+
+const INACTIVE_LINKED_SAMPLE_LIMIT = 20
+
+/**
+ * roadmap §7.1 — the offboarding blind spot. `DIRECTORY_DEPROVISION_ENABLED` (DT-OPS-01) is
+ * default-OFF by owner decision, so a directory account going `is_active = false` (DingTalk
+ * removal / the deactivation sweep) does NOT touch the linked local user at all: password
+ * login keeps working indefinitely. The `inactive_linked` review-item filter
+ * (`directory-sync.ts` `buildReviewFilterSql`) surfaces these as a REVIEW QUEUE, but has no
+ * notion of how long an account has sat that way — this is the missing age dimension.
+ *
+ * "Linked" here matches the `inactive_linked` review-item's own definition exactly:
+ * `directory_account_links.local_user_id IS NOT NULL` (link_status is NOT required to be
+ * `'linked'` — a recorded candidate match is enough to mean "this local user inherits
+ * whatever the directory account grants"). We ALSO require the local user to still be
+ * `is_active = true`: a local user already deactivated by some other path is not part of
+ * this specific blind spot (there is nothing left to revoke).
+ *
+ * Anchor for "since when": `directory_accounts.updated_at`. There is no dedicated
+ * deactivation-transition column in this schema. Verified by grepping every write site
+ * that can touch `directory_accounts` in this codebase — there are exactly two, both in
+ * `directory-sync.ts`:
+ *   1. The re-sync upsert (`INSERT ... ON CONFLICT DO UPDATE SET is_active = true, ...,
+ *      updated_at = NOW()`) — only fires for accounts CURRENTLY present in the DingTalk
+ *      pull, and always sets is_active back to true (which takes the row OUT of this
+ *      metric's target set).
+ *   2. The deactivation sweep (`UPDATE directory_accounts SET is_active = false,
+ *      updated_at = NOW() WHERE ... AND is_active = true`) — guarded by `AND is_active =
+ *      true`, i.e. a genuine one-way transition, not a re-stamp. Once a row is inactive,
+ *      this statement's WHERE clause no longer matches it, so it cannot bump `updated_at`
+ *      again while the row stays inactive.
+ * No other module in `packages/core-backend/src` issues an UPDATE or INSERT against
+ * `directory_accounts` (every other reference is a read-only JOIN). So for a row that is
+ * CURRENTLY `is_active = false`, `updated_at` is exactly the timestamp of the transition
+ * into inactivity — an honest anchor, not an approximation, given the current schema.
+ */
+export async function getDirectoryInactiveLinkedMetric(
+  integrationId: string,
+  thresholdDays: number,
+  queryFn: typeof query = query,
+  sampleLimit: number = INACTIVE_LINKED_SAMPLE_LIMIT,
+): Promise<DirectoryInactiveLinkedMetric> {
+  const normalizedThresholdDays = Math.floor(thresholdDays)
+
+  const [countResult, sampleResult] = await Promise.all([
+    queryFn<{ total: number }>(
+      `SELECT COUNT(*)::int AS total
+         FROM directory_accounts a
+         JOIN directory_account_links l ON l.directory_account_id = a.id
+         JOIN users u ON u.id = l.local_user_id
+        WHERE a.integration_id = $1
+          AND a.is_active = FALSE
+          AND u.is_active = TRUE
+          AND a.updated_at <= NOW() - ($2 || ' days')::interval`,
+      [integrationId, normalizedThresholdDays],
+    ),
+    queryFn<{
+      directory_account_id: string
+      external_user_id: string
+      account_name: string | null
+      local_user_id: string
+      local_user_email: string | null
+      local_user_name: string | null
+      inactive_since_at: string
+      inactive_days: number
+    }>(
+      `SELECT a.id AS directory_account_id,
+              a.external_user_id,
+              a.name AS account_name,
+              u.id AS local_user_id,
+              u.email AS local_user_email,
+              u.name AS local_user_name,
+              a.updated_at AS inactive_since_at,
+              FLOOR(EXTRACT(EPOCH FROM (NOW() - a.updated_at)) / 86400)::int AS inactive_days
+         FROM directory_accounts a
+         JOIN directory_account_links l ON l.directory_account_id = a.id
+         JOIN users u ON u.id = l.local_user_id
+        WHERE a.integration_id = $1
+          AND a.is_active = FALSE
+          AND u.is_active = TRUE
+          AND a.updated_at <= NOW() - ($2 || ' days')::interval
+        ORDER BY a.updated_at ASC
+        LIMIT $3`,
+      [integrationId, normalizedThresholdDays, sampleLimit],
+    ),
+  ])
+
+  return {
+    thresholdDays: normalizedThresholdDays,
+    count: Number(countResult.rows[0]?.total ?? 0),
+    sample: sampleResult.rows.map((row) => ({
+      directoryAccountId: row.directory_account_id,
+      externalUserId: row.external_user_id,
+      accountName: row.account_name,
+      localUserId: row.local_user_id,
+      localUserEmail: row.local_user_email,
+      localUserName: row.local_user_name,
+      inactiveSinceAt: row.inactive_since_at,
+      inactiveDays: Number(row.inactive_days),
+    })),
+  }
+}
+
+/**
+ * `unset = off`, matching `DIRECTORY_SYNC_ALERT_WEBHOOK`'s own channel-gating discipline. A
+ * non-positive or non-integer value is treated as unset (logged once) rather than silently
+ * clamped, so a typo in ops config fails closed instead of alerting on a bogus threshold.
+ */
+export function readDirectoryInactiveLinkedAlertThresholdDays(): number | null {
+  const raw = String(process.env.DIRECTORY_INACTIVE_LINKED_ALERT_DAYS ?? '').trim()
+  if (!raw) return null
+  const parsed = Number(raw)
+  if (!Number.isInteger(parsed) || parsed <= 0) {
+    logger.warn(`DIRECTORY_INACTIVE_LINKED_ALERT_DAYS is set but not a positive integer: ${raw}`)
+    return null
+  }
+  return parsed
+}
+
+export function buildDirectoryInactiveLinkedAlertMessage(options: {
+  integrationName: string
+  metric: DirectoryInactiveLinkedMetric
+}): { subject: string; content: string } {
+  const { integrationName, metric } = options
+  const subject = `钉钉目录存在停用超 ${metric.thresholdDays} 天但本地账号仍激活的绑定`
+  const lines = [
+    `- 集成：${integrationName}`,
+    `- 阈值：${metric.thresholdDays} 天`,
+    `- 数量：${metric.count}`,
+  ]
+  for (const item of metric.sample.slice(0, 5)) {
+    const label = item.localUserName || item.localUserEmail || item.localUserId
+    lines.push(`  - ${item.accountName || item.externalUserId} → ${label}（已停用 ${item.inactiveDays} 天）`)
+  }
+  if (metric.count > metric.sample.length) {
+    lines.push(`  - ……以及其他 ${metric.count - Math.min(metric.sample.length, 5)} 个账号`)
+  }
+  return { subject, content: lines.join('\n') }
+}
+
+/**
+ * §7.1 post-run informational alert. Best-effort and NEVER throws — this is a backlog
+ * digest, not a failure signal, and must not make a healthy sync look broken.
+ *
+ * Two independent env gates, both required, matching the `breach-channels` /
+ * `DIRECTORY_SYNC_ALERT_WEBHOOK` convention that a side channel exists only when configured:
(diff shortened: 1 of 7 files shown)


zensgit:
Follow-up commit (c09ddb22f) addresses a self-review pass on the initial PR:

- Added the 401 unauthenticated case for `GET /inactive-linked` (the brief called for 401/403/shape; only 403 existed).
- Fixed `buildDirectoryInactiveLinkedAlertMessage`'s "…and N more" remainder guard: it compared `count` against the upstream sample length (capped at 20), which `getDirectoryInactiveLinkedMetric` can never make less than `count` for `count <= 20` — so the remainder line was only reachable via a fixture the real query cannot produce. Now compares against what the message actually displays (`min(sample.length, 5)`), plus a test for the "small backlog, no remainder line" case.
- Dropped an unused `accountIds` array in the new db-test fixture setup.

Full unit suite (353 files / 4769 tests) and the whitelisted real-DB file both green on the rebased tip; `tsc --noEmit` clean.

zensgit:
# Adversarial Review — PR #4057 (§7.1 inactive-linked-for-N-days metric + post-run alert)

- **PR**: zensgit/metasheet2 #4057 — `feat(directory): §7.1 inactive-linked-for-N-days metric + post-run alert`
- **Branch / head**: `claude/dt-inactive-linked-metric` @ **c09ddb22f3f7a3a534e095312cda10d402f79f91** (verified = PR head at review time)
- **Base**: main (merge-base 7a4bd826b)
- **Reviewer**: Claude (independent adversarial gate, refute-first), 2026-07-10
- **Method**: full-file reads off `refs/reviews/pr4057`; repo-wide write-site sweep for the anchor claim; own disposable worktree `/tmp/ms2-rev-pr4057` + frozen install; fresh migrated Postgres (`pr4057_review`, CI's `MIGRATION_EXCLUDE`); re-fired implementer mutations 1 & 4 + 3 mutations/probes of my own; overlap audit vs in-flight #4049/#4053/#4054.

## Verdict: **APPROVE**

No P1/P2. The load-bearing claims — the `updated_at` anchor, the never-throws alert hook, the env gate, the admin gate, zero overlap with in-flight hunks — all survived independent refutation. Two P3s (one is a landing-mechanics hazard vs #4054, not a code defect) and NITs below.

---

## Probe results (per gate brief)

### Probe 1 — THE ANCHOR CLAIM: **UPHELD**

Independent repo-wide sweep (not just `packages/core-backend/src`) for any INSERT/UPDATE touching `directory_accounts`:

- **Exactly two runtime write sites**, both in `directory-sync.ts`:
  - upsert @ `directory-sync.ts:2962` — `ON CONFLICT ... DO UPDATE SET ... is_active = true, updated_at = NOW()`. Fires only for accounts present in the current DingTalk pull and **unconditionally sets `is_active = true`**, which removes the row from the metric's target set. Cannot re-stamp a row *while it stays inactive*.
  - deactivation sweep @ `directory-sync.ts:3015` — `UPDATE ... SET is_active = false, updated_at = NOW() WHERE ... AND is_active = true`. The `AND is_active = true` guard (DT-OPS-01) makes it a one-way transition; the WHERE no longer matches already-inactive rows.
- **All other references are read-only JOINs** — checked each: `multitable/automation-executor.ts` (3× LATERAL JOIN), `multitable/permission-service.ts:660`, `routes/admin-users.ts` (6× JOIN), `routes/auth.ts:374`, `routes/univer-meta.ts:814`, `services/ApprovalDirectoryOrg.ts` (5× SELECT), `services/AttendanceNotificationDeliveryWorker.ts` (2× JOIN), plus the admin bind/unbind/admit/review paths — those write `directory_account_links` / `users` / `user_external_identities` only, never `directory_accounts` (full write-statement inventory of `directory-sync.ts` checked line-by-line).
- **#3911's batched unnest seam** targets `directory_account_departments` (`directory-sync.ts:1036`), not `directory_accounts`.
- **The deprovision executor** writes `users` (`directory-sync.ts:1355`) and grants — never `directory_accounts`. Interaction is actually correct: a `mark_inactive`-deprovisioned local user drops out of the metric via `u.is_active = TRUE` (nothing left to revoke).
- **No triggers** on `directory_accounts` (migration `zzzz20260324150000_create_directory_sync_tables.ts` creates only indexes; the repo's `update_updated_at` triggers exist only on views/tables/kanban/calendar tables).
- One extra hit outside src: `scripts/ops/staging-attendance-c5-delivery-smoke.ts:295` — a staging smoke script that INSERTs *fresh* accounts with `is_active=true` (random UUID ids, no ON CONFLICT). Cannot touch an inactive row's clock. Non-issue.
- `directory_account_links` has a **UNIQUE index on `directory_account_id`** (migration line ~147), so the metric's JOIN cannot double-count an account.

Anchor verdict: for a row currently `is_active = false`, `updated_at` = the latest transition into inactivity. Claim holds. (See NIT-1 for the pre-DT-OPS-01 history caveat.)

### Probe 2 — Alert-hook fail-stop safety: **UPHELD** (verified in composition, not just structurally)

- Structural: everything that can throw inside `deliverDirectoryInactiveLinkedAlert` (metric query, signing, fetch, response validation) is inside its try/catch. The two gate reads *outside* the try are throw-safe: `readDirectoryInactiveLinkedAlertThresholdDays` is pure env parsing; `readDirectorySyncAlertWebhookConfig` catches internally and returns null (`directory-sync-alert-delivery.ts:36-51`).
- Runtime probe (my own, scratch composition test in the worktree — real `deliverDirectoryInactiveLinkedAlert`, NOT mocked, inside the real `runScheduledSync` via `startDirectorySyncScheduler`'s captured handler, both env gates open):
  - metric query **throws** → handler **resolves**; sync completed. GREEN.
  - metric returns nonzero backlog, webhook layer **fails** → handler **resolves**. GREEN.
- Lease-conflict skip and real-sync-failure paths never reach the digest (scheduler unit tests pin both; verified in `directory-sync-scheduler.ts:56-76`).

No #3910-style fail-stop hazard.

### Probe 3 — Env gate: **UPHELD**

- Unset ⇒ off; `'0'/'-3'/'2.5'/'abc'` ⇒ treated as unset (fail-closed, logged), pinned by unit tests.
- Re-fired the implementer's mutation 4 myself: deleted `if (!thresholdDays) return false` (`directory-sync-alert-delivery.ts:419`) → **exactly** the targeted unit test fails ("sends nothing when the threshold env is unset"), with a fake queryFn reporting a nonzero backlog — the guard is provably load-bearing, not masked by a redundant inner gate. RED confirmed, reverted.

### Probe 4 — Endpoint: **UPHELD**

- `admin-directory.ts:535-548`: `ensurePlatformAdmin` first (401 unauthenticated / 403 non-admin; both pinned by new route tests, metric fn not called).
- `normalizeInactiveLinkedDays` (`admin-directory.ts:302-308`): non-string/array ⇒ 30; NaN/non-integer/`<=0` ⇒ 30; huge ⇒ clamped 3650. Garbage `:integrationId` (non-uuid) surfaces as PG "invalid input syntax" ⇒ mapped 400 by the sibling-mirrored error shape.
- Response: `{metric: {thresholdDays, count, sample[]}}` — ids, external ids, account name, local user name/email. No tokens, no secrets, no `raw` jsonb. Admin-gated. Acceptable.

### Probe 5 — Days semantics / SQL: **UPHELD**

- `AND a.updated_at <= NOW() - ($2 || ' days')::interval` — fully parameterized ($2 is a server-side `Math.floor`ed integer); no user-string interpolation. Scheduler runs UTC; day-granularity threshold makes DST drift immaterial. Inclusive `<=` boundary pinned on real PG in both directions (20d in / exactly-10d in / 5d out / 1d out).

### Probe 6 — Overlap audit: **UPHELD, one landing hazard**

- `directory-sync.ts`: **zero hunks** in this PR (diffstat-verified). ✓
- `admin-directory.ts`: this PR's hunks @26 (import), @65 (helper), @520 (new GET) vs #4053's @39/@70/@172/@185 and #4049's @359 — disjoint regions, no create/update-handler contact. ✓
- `admin-directory-routes.test.ts`: @93/@213/@939 vs #4049 @627, #4053 @4/@470/@1646 — disjoint. ✓
- `directory-sync-alert-coverage.db.test.ts`: this PR appends @216→EOF **and so does #4054** (@216,3 +241,236), and **both also edit the same import block** (lines 1-9). ⇒ See P3-1.

---

## Findings

### P1 — none
### P2 — none

### P3-1 — "trivial 3-way merge" claim vs #4054 is factually wrong; landing order needs a union-resolve
`tests/integration/directory-sync-alert-coverage.db.test.ts` — #4057 and #4054 both (a) modify the same import statement (lines 2-8) and (b) append a new describe block at the same EOF anchor (after line 218). Git cannot auto-order two different additions at the same position: **whichever PR lands second gets a real merge conflict at both spots.** Failure scenario if resolved carelessly: dropping the other PR's import or describe block makes those tests silently vanish — and because this file only runs in the plugin-tests real-DB whitelist step, a dropped block produces no missing-check signal on the PR that dropped it (house rule: conflict resolution in test files must take the UNION). Not a defect in this PR's code; a required landing note. Fix: whoever merges second resolves both conflicts as union (keep both imports, keep both trailing describe blocks, either order).

### P3-2 — schedule closure captures a mutable `row.name`; renamed integrations alert under the stale name
`directory-sync-scheduler.ts:96-100`: `scheduler.schedule(..., async () => { await runScheduledSync(row.id, row.name) })` captures the display name at job-creation time, and the `existingJob` path calls `reschedule(jobName, cronExpression)` which keeps the **old callback**. Failure scenario: admin renames the integration (name is editable via the update route); every nightly digest keeps saying `集成：<old name>` until the job is unscheduled/recreated or the process restarts. Prior to this PR the closure captured only the immutable `row.id`, so this staleness class is newly introduced. Blast radius: cosmetic (alert text only — id, thresholds, counts all correct), hence P3 not P2. Fix (post-merge fine): have `runScheduledSync` take only `integrationId` and let the digest resolve the current name (e.g. via `getScheduleRow`, itself inside deliver's try/catch), or refresh the callback on rename.

### NIT-1 — "updated_at is exactly the deactivation timestamp" is over-strong for pre-DT-OPS-01 history
The doc-comment (`directory-sync-alert-delivery.ts:258-273`) is correct for rows deactivated after the sweep gained `AND is_active = true` (DT-OPS-01). Rows already inactive *before* that guard landed were re-stamped by every nightly sweep, so their `updated_at` ≈ the last pre-guard sync, not the true departure time — the metric **underestimates** their age (conservative direction: alerts later, never falsely early). Worth one sentence in the comment; no code change needed.

### NIT-2 — digest content pushes local user email/name to the ops webhook
`buildDirectoryInactiveLinkedAlertMessage` (`directory-sync-alert-delivery.ts:381-383`) includes `localUserName || localUserEmail || localUserId` per sample line. Same PII class and channel discipline as the existing sync-failure alert (ops-configured DingTalk group robot, HMAC-signed, masked in logs); no secrets. Fine as-is; noting for the record.

### NIT-3 — nightly repeat while backlog nonzero
The digest re-fires after every scheduled sync until the backlog clears (no dedup/ack/quiet-hours). By design for a digest, and quiet-hours is separately tracked in the DingTalk backlog pool; no change requested.

---

## Test & mutation record (all in my own worktree, `/tmp/ms2-rev-pr4057` @ c09ddb22f, frozen install)

| Run | Result |
|---|---|
| 3 touched unit files + 2 neighbors (`directory-sync-review-items`, `directory-deprovision-policy`) | **104/104 passed** |
| Full `tests/unit` suite | **353 files / 4769 tests passed** |
| `tsc --noEmit` (core-backend) | clean (exit 0) |
| Fresh DB `pr4057_review` (full migrations, CI's `MIGRATION_EXCLUDE`) → `directory-sync-alert-coverage.db.test.ts` | **9/9 passed** (2 new §7.1 tests included) |
| Same DB → neighbors `directory-deprovision-selection.db.test.ts` + `directory-sync-preview-apply-parity.db.test.ts` | **14/14 passed** |

| Mutation (applied → run → reverted) | Result |
|---|---|
| M1 (re-fired): `<=` → `>=` in both metric queries | **RED** — 2 db tests fail |
| M4 (re-fired): delete `if (!thresholdDays) return false` | **RED** — exactly the targeted env-gate unit test fails |
| Mine: neuter `a.integration_id = $1` scoping | **RED** — 3 tests fail (second-integration fixture is load-bearing) |
| Mine: drop `AND a.is_active = FALSE` | **RED** — 1 db test fails (still-active fixture leaks in) |
| Mine (probe 2): metric query throws / webhook fails inside REAL `runScheduledSync` composition | scheduled sync handler **resolves** both times — no fail-stop |

Two-point wiring verified: file excluded at `vitest.config.ts:123` and whitelisted at `plugin-tests.yml:403` — the new db tests **do** run in CI (not silent-green).

## Design-lock / scope table (roadmap §7.1, `dingtalk-sync-integrated-roadmap-20260708.md` ~line 333)

| Gate | Verdict | Evidence |
|---|---|---|
| "Add 'inactive linked for N days' metric and alert" | **PASS** | metric + GET endpoint + gated post-run digest; read-only, no deprovision writes |
| No deprovision behavior change (policy modes = prior DT-OPS work) | **PASS** | zero hunks in `directory-sync.ts`; no writes anywhere in PR |
| Env-gate side-channel convention (unset = off) | **PASS** | mutation-verified load-bearing guard |
| Coordination: `directory-sync.ts` untouched; no #4053 handler contact | **PASS** | hunk-region audit above |
| Claimed out-of-scope (no UI, no persisted alert row, no migration) | **PASS** | diff = 7 files, none of those surfaces |

Cleanup: review worktree and `pr4057_review` DB dropped after this review; no other sessions' worktrees/containers touched.

zensgit:
UNION window executed (the gate-forewarned double-collision with #4054 in directory-sync-alert-coverage.db.test.ts): resolved via net-delta terminal-tree construction — main's file (R5 block intact) + this PR's self-contained describe block appended + import union, with a set-diff union proof (every main line present, PR block whole). Fresh-DB merged suite 14/14 (R5's 4 + this PR's tests + priors) + tsc clean; load-bearing mutation replayed at both REAL SQL sites (age comparison inverted ⇒ RED ×2, restored). Armed alone — Wave 2 window 4.

PR #4058, opened 2026-07-10
feat(dingtalk): add interactive approval-card delivery

zensgit:
## Summary

- harden the B-1 Stream worker before a real SDK adapter lands: single-flight initialization, best-effort close after a half-start, and a distinct values-free `sdk_unwired` status
- add the B-2 interactive-card send branch to `send_dingtalk_approval_card`, selected only by the complete explicit Stream env gate
- preserve the flag-off OA ActionCard path, per-corp `integration_id`, and same-source signed decision link from current main
- call DingTalk `createAndDeliver` with `outTrackId` equal to the delivery ledger id, store the returned carrier id, and fail closed on HTTP-200 business/delivery failures
- keep cards values-free; reject remains the signed Slice-A decision-page URL

## Operational gate

This PR is send-only. Keep `DINGTALK_INTERACTIVE_CARD_STREAM_ENABLED` off until B-3 callback execution, B-4 terminal card updates, the real Stream SDK adapter, and real DingTalk UAT are complete. The default worker factory intentionally remains `sdk_unwired`.

## Verification

- unit: 16/16 (`dingtalk-interactive-card-stream` + `dingtalk-work-notification`)
- real DB: 10/10 (`automation-dingtalk-approval-card-action`) on an ephemeral database migrated with the CI exclusion list
- focused real-DB test also passes independently (`-t 'B-2 opt-in'`)
- mutation: forcing the runtime branch back to OA makes the B-2 golden fail on `work_notice_action_card` vs `interactive_card`
- backend ESLint clean for all changed runtime modules

## Docs

- updates the ratified Slice-B lock through B-2
- reconciles the B-1 verification note with the hardened status behavior
- adds the B-2 development/verification record

No migration and no new dependency.


zensgit:
P2-1 is closed on the rebased head `0c18f3526`.

- Added the discriminating response shape: top-level `success:false` with a misleading nested `deliverResults[0].success:true` and carrier id.
- The test requires rejection with the top-level provider message.
- Mutation proof: neutralizing only the top-level `payload.success !== true` guard makes this case resolve successfully and the focused test goes RED; restoring the guard returns green.
- Updated the B-2 verification MD to 17/17 unit tests and recorded the four non-blocking B-3/UAT follow-ups.

Post-rebase verification: unit 17/17 and backend type-check clean. Runtime code is unchanged by this review round. CI is re-running on the current head.

PR #4059, opened 2026-07-10
docs(multitable): R10 gate-front — 4 owner-menu decision one-pagers (PROPOSED)

zensgit:
**Docs-only, zero runtime** — R10 round gate-front artifacts (owner /goal「这条线余下开发」, owner away ⇒ gated items pushed to decision-ready only, per the line's standing R4 discipline).

Four per-item decision units, each with options + recommendation + inline unlock word:

| # | Item | Recommendation |
|---|---|---|
| 1 | **restore 批次回链** mini-lock | `restored_from_version` col + 2 execute-path threading + projection/FE + G1-G5 |
| 2 | **field-value tombstone 留存地板** | **A: accept the documented boundary**(re-decision trigger tied to any 4c capture/revert flag going prod+retention-on); B=weak floor rejected reasoning included |
| 3 | **PIT-resurrect 锚精确化** | **A: opt-in explicit `deleteRevisionId`**(fail-closed validation + preview-identity hash binding), backward compatible, zero schema |
| 4 | **all-tables 字段元数据** | **B: server-side masked `fieldNames` map** reusing the projection's per-sheet allow-sets — field NAMES are two-layer-maskable (R9 #4007 lesson); client-side re-derivation = two-truth-sources hazard |

No implementation is authorized by this PR; each item unlocks independently on its own owner nod.

### DIFF
--- a/docs/development/multitable-global-history-alltables-field-metadata-decision-20260710.md
+++ b/docs/development/multitable-global-history-alltables-field-metadata-decision-20260710.md
@@ -1,0 +1,22 @@
+# Global History — all-tables 模式跨表字段元数据 — 决策 one-pager (PROPOSED)
+
+- **Status**: PROPOSED 决策文档；owner ratify 前零实现授权。R10 gate-front 工件（R9 UX 审计携带项）。
+- **现状**：History Center 的 `fields` prop 只承载**活动表**（workbench 的 `historyVisibleFields`，layer-2∩layer-3 已修）；all-tables 模式下其它表的批次 diff 标签/筛选回退为原始 field id。R9 已如实保留该回退。
+
+## 约束（不可绕）
+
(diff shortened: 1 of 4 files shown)


zensgit:
**Review: APPROVE-with-hardening (0 P1/0 P2, 2 P3 — both folded, head updated).**

- P3-1: legacy `POST …/records/:recordId/restore`(:9362, no active FE caller) now named in the 回链 lock's §5 out-of-scope with its NULL intent pinned by G2.
- P3-2: 锚 doc corrected — rec-A's 'client already knows candidate anchors' was imprecise (preview response doesn't surface them; A needs a response-shape addition + identity-hash binding), AND the reviewer surfaced a strictly cheaper option the table missed: **A′ = server-side anchor derivation from the `asOf` T already on the wire**(zero wire change, fixes the exact multi-vintage under-replay). A′ is now the recommendation; A demoted with the corrected cost analysis.

Review MD: /tmp/pr4059-review-claude-20260710.md. Arming auto-merge (docs-only, PROPOSED, zero runtime — nothing here authorizes implementation).

PR #4060, opened 2026-07-10
fix(multitable): wire TrashModal to the two-layer-visible field list (R10)

zensgit:
## What

Field VISIBILITY has two independent display layers:
- **layer-2 property-hidden** (`field.property.hidden === true || property.visible === false`) — `filterPropertyVisibleFields`
- **layer-3 per-subject RBAC** (`effectiveFieldPermissions[f.id]?.visible !== false`) — `scopedAllFields`

`TrashModal` was wired `:fields="scopedAllFields"` — layer-3 only. It uses `fields` (via `pickRecordTitle`) to derive a trash-row **title** from field **values**, so a property-hidden (but RBAC-readable) field's value could seed a visible title. This is a display-consistency gap, not a permission bypass: the acting user already has read access to that field's value (RBAC layer-3 is unaffected and untouched by this PR) — the fix only stops a display-hidden field from surfacing in a derived title where the plain field itself wouldn't be shown.

Fix: reuse the existing `filterPropertyVisibleFields(scopedAllFields)` computed added by #4007 for History Center (`historyVisibleFields`), and wire it into `TrashModal` too. Since it now serves two consumers, renamed to `twoLayerVisibleFields` (both usages + the one spec comment reference updated — grepped the whole repo, no other reference exists). No RBAC/permission logic changed.

Also tagged the trash-open toolbar button with `data-action="open-trash"`, matching the existing `open-history` / `open-config-history` / `open-template-library` convention, so it's clickable from a wiring spec without a text-content query.

## Verification

Extended the existing `multitable-workbench-history-field-scope-wiring.spec.ts` (neither `multitable-trash-fe.spec.ts` nor `multitable-workbench-restore-wiring.spec.ts` mounts `TrashModal` through the real workbench with it open — this file already mounts the real workbench + real `TrashModal`/`HistoryCenterModal`) with a new `describe` block that opens the real Trash modal and asserts the derived row title.

```
pnpm --filter @metasheet/web exec vitest run \
  tests/multitable-trash-fe.spec.ts \
  tests/multitable-history-fe.spec.ts \
  tests/multitable-history-center-ai-shortcut-label.spec.ts \
  tests/multitable-history-center-inline-diff.spec.ts \
  tests/multitable-history-center-pinned-batch-deeplink.spec.ts \
  tests/multitable-workbench-history-field-scope-wiring.spec.ts \
  tests/multitable-workbench-restore-wiring.spec.ts \
  --watch=false
# Test Files  7 passed (7)
#      Tests  56 passed (56)
```

`multitable-workbench-history-field-scope-wiring.spec.ts` alone: 6/6 passed (5 pre-existing History Center cases + 1 new TrashModal case).

`pnpm --filter @metasheet/web exec vue-tsc -b` — clean, exit 0, no output.

Required CI (`test 20.x`) only builds `apps/web`; it does not execute this vitest suite. The commands above are the reproducible evidence.

## Mutation evidence

Checkpoint-committed the fix first, then used the Edit tool to swap `TrashModal`'s `:fields` back to `scopedAllFields` (the pre-fix wiring), re-ran the same spec file, and reverted with the Edit tool back to the exact committed diff (`git diff --stat` empty after revert):

```
❯ tests/multitable-workbench-history-field-scope-wiring.spec.ts (6 tests | 1 failed)
  ❯ MultitableWorkbench -> TrashModal field-scope wiring > the trash-row title falls through
    a property-hidden field (layer-2) to the record-id fallback, and never shows the
    RBAC-denied field (layer-3) either
    → expected 'Hidden Value' to be '#del_1'

Test Files  1 failed (1)
     Tests  1 failed | 5 passed (6)
```

Exactly the property-hidden (layer-2) assertion reds — the fld_3 value leaks into the title instead of falling through to the record-id fallback (`#del_1`). The RBAC-denied (layer-3, fld_2) assertion and all 5 pre-existing History Center cases stay green, confirming layer-3 already held via `scopedAllFields` and this fix is precisely load-bearing for layer-2.

## Not covered

- No RBAC/permission-layer logic was touched — `scopedAllFields` (the RBAC filter) is reused as-is; this PR only layers the existing property-hidden filter on top for `TrashModal`, same as #4007 did for History Center.
- The property-hidden field's value still legitimately lives in the record's `data` payload and in any other surface that renders it directly (e.g. the record drawer, when the field itself isn't display-hidden there) — this change only touches trash-row *title derivation* via `pickRecordTitle`.
- Did not rename the spec file itself (still named `...history-field-scope-wiring.spec.ts` though it now also covers TrashModal) — kept the diff minimal per the task's neighbor-sweep list, which names this file explicitly.

### DIFF
--- a/apps/web/src/multitable/views/MultitableWorkbench.vue
+++ b/apps/web/src/multitable/views/MultitableWorkbench.vue
@@ -61,7 +61,7 @@
       <button class="mt-workbench__mgr-btn" :class="{ 'mt-workbench__mgr-btn--active': showDashboardView }" @click="showDashboardView = !showDashboardView" data-action="toggle-dashboard"><el-icon class="mt-workbench__mgr-btn-icon"><component :is="ICON.dashboard" /></el-icon> {{ wb('toolbar.dashboard', isZh) }}</button>
       <button v-if="activeViewType === 'form'" class="mt-workbench__mgr-btn" @click="showFormShareManager = true"><el-icon class="mt-workbench__mgr-btn-icon"><component :is="ICON.shareForm" /></el-icon> {{ wb('toolbar.shareForm', isZh) }}</button>
       <button class="mt-workbench__mgr-btn" @click="showApiTokenManager = true"><el-icon class="mt-workbench__mgr-btn-icon"><component :is="ICON.apiWebhooks" /></el-icon> {{ wb('toolbar.apiWebhooks', isZh) }}</button>
-      <button v-if="caps.canDeleteRecord.value" class="mt-workbench__mgr-btn" @click="showTrash = true"><el-icon class="mt-workbench__mgr-btn-icon"><component :is="ICON.trash" /></el-icon> {{ wb('toolbar.trash', isZh) }}</button>
+      <button v-if="caps.canDeleteRecord.value" class="mt-workbench__mgr-btn" data-action="open-trash" @click="showTrash = true"><el-icon class="mt-workbench__mgr-btn-icon"><component :is="ICON.trash" /></el-icon> {{ wb('toolbar.trash', isZh) }}</button>
       <button v-if="activeBaseId" class="mt-workbench__mgr-btn" data-action="open-history" @click="historyDeepLinkBatchId = null; showHistory = true"><el-icon class="mt-workbench__mgr-btn-icon"><component :is="ICON.history" /></el-icon> {{ isZh ? '历史' : 'History' }}</button>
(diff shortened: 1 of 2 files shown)


zensgit:
**Adversarial review: APPROVE — 0 P1 / 0 P2 / 0 P3**(independent reviewer, isolated worktree, mutation self-reproduced: swap back to scopedAllFields → exactly 1 red `expected 'Hidden Value' to be '#del_1'`; 7-file sweep 56/56 + vue-tsc clean re-run, not taken on faith). Sibling sweep confirms TrashModal was the LAST layer-3-only pickRecordTitle feed — no second leak surface. 2 non-blocking NITs (spec filename scope, belt-and-suspenders assertion) — no change required. Review MD: /tmp/pr4060-review-claude-20260710.md. Arming auto-merge.

PR #4062, opened 2026-07-10
fix(onprem): route pnpm through corepack version-addressed passthrough (entity-machine PNPM_VERSION_MISMATCH)

zensgit:
## What

Implements the operator-requested correction from the #3751 corrective-rerun report (`failureClass=PNPM_VERSION_MISMATCH`, blocked before dependency preflight):

> Requested correction: after Corepack activation, resolve/invoke the pinned Corepack shim directly (or otherwise exclude shadowing user/system-profile pnpm binaries) before enforcing the exact-version check.

`Initialize-PinnedPnpm` in `scripts/ops/multitable-onprem-apply-package.ps1` now:

1. after a successful `corepack prepare pnpm@<pin> --activate`, probes the pnpm shims **co-located with the corepack binary** (where `--activate` writes them) and selects the candidate whose `--version` equals the pin;
2. falls back to the existing generic PATH walk **only when no co-located shim matches**;
3. keeps the exact-version check **fail-closed on either path** — a shadowing profile pnpm can delay but never bypass the pin — and the `PNPM_VERSION_MISMATCH` error now carries the probed shim list for diagnosability.

`onprem-windows-system-hardening.test.mjs` locks the ordering (shim probe precedes the PATH fallback) so the correction can't silently regress.

## Verification

- `node --test scripts/ops/onprem-windows-system-hardening.test.mjs` → 12/12
- `node --test scripts/ops/multitable-onprem-package-no-node-modules.test.mjs` → 8/8
- `pwsh` tokenizer parse of the patched ps1 → clean
- `PNPM_VERSION_MISMATCH` guard string retained (package-verify.sh contract intact)

After merge: re-cut the on-prem package via `multitable-onprem-package-build.yml` and post the new release pointer to #3751 for the rerun.


zensgit:
Superseded by #4061, which landed the same correction first via a generated corepack-dispatch cmd wrapper (`corepack "pnpm@<pin>" %*` as the resolved pnpm path) — mechanism-equivalent to this PR's version-addressed passthrough, with its own hardening tests. Per the adversarial review of this PR (/tmp/pr4062-review-claude-20260710.md), the version-addressed dispatch structurally defeats PATH shadowing and overrides ambient project specs (corepack source :23005), so no residual delta from this PR is load-bearing. Closing; branch will be deleted.

PR #4069, opened 2026-07-10
feat(directory-web): async sync polling + dry-run preview + R5 run-card stats (Wave 2 final, R7)

zensgit:
## What / why

R7 from the DingTalk backlog pool (Wave 2, final): #3915 shipped opt-in async directory sync + a dry-run preview endpoint, #4049 added the preview-route 409 lease refusal, and #4054 (armed) persists per-run change stats — but the admin UI still only did a blocking synchronous sync with no preview surface and no run polling. **Web-only** change in `DirectoryManagementView.vue` + its required-gate spec; zero backend edits.

- **预览同步 (dry-run)**: POSTs `.../sync/preview`, renders the `DirectorySyncPreview` counts (seen/would-create/would-deactivate/linked-deactivations + auto-admission triage) and the sampled new/deactivated account lists in a card mirroring the 自动同步观测 card. Own `previewBusy` flag — a preview performs a **full DingTalk pull** and must not piggyback the global `busy`.
- **后台同步**: POSTs `.../sync` with `{async:true}` → 202 `{runId}` → polls `GET .../runs?page=1&pageSize=10` every 5 s, matching by runId until terminal, then setStatus with the run summary (or `errorMessage`) and the same panel refresh the synchronous path does. Poll is a self-rescheduling `setTimeout` + token (the `useAiBulkFill` pattern) so a slow poll never overlaps the next and a stale continuation can't resurrect the loop.
- **MAIN-LOOP-RATIFIED FAIL-SAFE (implemented exactly)**: when auto-admission is enabled, the async button warn-and-confirms first — the dialog states that async discards the one-time temp-password onboarding packets (never persisted, unrecoverable) and recommends synchronous 手动同步 when admitting new users.
- **409 `DIRECTORY_SYNC_IN_PROGRESS` (#4049)**: preview, async **and** the pre-existing manual path now show `已有同步在进行中（运行 <activeRunId>）` instead of a generic failure (manual path included for consistency — flagged below).
- **R5 run-card stats (#4054 gate P3-2)**: run cards render the new per-run keys (`accounts/departmentsCreated/Updated/DeactivatedCount`, `managerCoverage` as `%（linked/total）`, `durationMs` humanized) via the defensive `readNumericStat` pattern, **gated on key presence** so pre-R5 runs don't render a misleading all-zero "this run changed nothing" line. Key names only — no dependency on #4054's code (additive jsonb).
- **Timer hygiene**: poll cleared on unmount, on terminal/error, and when switching integration; it survives same-integration refreshes (e.g. saving Agent ID mid-run).

## Test evidence

`directoryManagementView.spec.ts` is IN the required web gate (`run-required-web-tests.sh`, since #4010) — no new CI wiring needed.

- Spec: 54 pre-existing + **10 new = 64/64 green**
- Full required web gate: `bash apps/web/scripts/run-required-web-tests.sh` → **141 files / 1811 tests, all green**
- `vue-tsc -b` (apps/web): clean

New tests: exact-path + exact-body asserts for both POSTs; 409 rendering on both paths; fake-timer poll lifecycle (continues on running, stops on completed AND failed, renders summary/errorMessage, refreshes panels, **zero fetches of any kind after terminal**); unmount timer-leak tripwire; confirm-gate decline/confirm/manual_only-skips; R5 stats present/absent gating.

## Mutation table (each applied to the committed impl, spec re-run, then restored)

| # | Mutation | Result |
|---|---|---|
| M1 | Bend preview path `/sync/preview` → `/preview-sync` | RED — preview exact-path test |
| M2 | Drop `body: JSON.stringify({async:true})` from the async POST | RED — 2 tests (exact-body + confirm-proceeds) |
| M3 | Remove preview's 409 branch | RED — preview 409 test |
| M4 | Remove async path's 409 branch | RED — async 409 test |
| M5 | Remove the auto-admission confirm gate | RED — 2 FAIL-SAFE tests |
| M6a | Neuter terminal check (`&& false`) — poll never stops | RED — both poll-lifecycle tests (timer-leak tripwire) |
| M6b | Drop `stopRunPolling()` in the terminal branch only | RED — completed-poll test (lingering in-progress state) |
| M7 | Drop `stopRunPolling()` from `onUnmounted` | RED — unmount leak test |
| M8 | Un-gate the R5 change-summary line (`v-if="true"`) | RED — pre-R5 absent-keys test |
| M9 | `readActiveRunId` always returns `''` | RED — both 409 tests (runId must be surfaced) |

## Not proven / honest gaps

- **Run paged out of page 1**: polling scans `runs?page=1&pageSize=10` (no per-run GET endpoint exists); if >10 newer runs appear between polls the run temporarily vanishes and polling simply continues (commented in code). Not tested; unbounded polling if the run never reappears — an admin can navigate away (unmount cleans up) and the run row itself still records the outcome.
- **Poll survives same-integration refresh** is by design (see `selectIntegration` comment) but has no dedicated test; the switching-integration stop path is likewise untested directly (covered indirectly by unmount + terminal stops).
- The completed-run status line, preview card copy, and confirm-dialog text are pinned by substring asserts, not full-string snapshots.
- No browser-level verification (jsdom only); ElMessageBox is spied, not rendered.
- #4054 is NOT merged yet — this renders its **key names** defensively; if that PR renames keys before landing, the run-card lines silently stay hidden (fail-safe direction: hidden, not wrong).

## Flagged for review

- I also added the 409 branch to the **pre-existing synchronous 手动同步 path** (3 lines, same helper). Strictly speaking the brief listed 409 handling for the two new paths; leaving manual sync showing the raw backend English message felt like a consistency bug. Say the word and I'll drop it.
- `syncIntegration`'s post-sync `Promise.all` refresh was extracted into `refreshAfterDirectorySync()` and shared with the poll's terminal branch — behavior identical, pinned by the pre-existing manual-sync test.
- Manual sync + preview + async buttons all disable while this tab tracks a background run (`pollingRunId`); cross-tab/scheduler collisions still land on the 409 message.

## Sequencing

Wave 2 (final) of the directory-sync serial queue — lands after #4049 (already on main; base verified to include #4010 + #4049). **Do not arm auto-merge; landing is gated and serialized by the main loop.**

### DIFF
--- a/apps/web/src/views/DirectoryManagementView.vue
+++ b/apps/web/src/views/DirectoryManagementView.vue
@@ -4525,6 +4525,12 @@ function stopRunPolling(): void {
 }
 
 const DIRECTORY_RUN_POLL_INTERVAL_MS = 5000
+// P3 hardening: a crashed worker (scheduler disabled) leaves the run row 'running' forever, and
+// this tab locks its own sync/preview buttons on pollingRunId — the very actions that would
+// reclaim the lease. Bound the poll so the tab gives up and unlocks instead of hammering /runs
+// indefinitely; the run itself is untouched server-side.
+const DIRECTORY_RUN_POLL_MAX_DURATION_MS = 10 * 60 * 1000
+let runPollStartedAtMs = 0
 
 async function pollAsyncRun(integrationId: string, runId: string, token: number): Promise<void> {
   let body: unknown = null
@@ -4555,6 +4561,11 @@ async function pollAsyncRun(integrationId: string, runId: string, token: number)
       setStatus(buildAsyncRunSummary(run))
     }
     await refreshAfterDirectorySync(integrationId)
+    return
+  }
+  if (Date.now() - runPollStartedAtMs >= DIRECTORY_RUN_POLL_MAX_DURATION_MS) {
+    stopRunPolling()
+    setStatus('已停止自动轮询（超过 10 分钟），运行可能仍在进行，请前往下方运行记录手动刷新查看最新状态', 'error')
     return
   }
   // Still running — or paged out of the first page (>10 newer runs between polls; the run is
@@ -4612,6 +4623,7 @@ async function startAsyncSync() {
     stopRunPolling()
     pollingRunId.value = runId
     pollingIntegrationId = integration.id
+    runPollStartedAtMs = Date.now()
     const token = runPollToken
     setStatus(`后台同步已启动（运行 ${runId}），正在轮询运行状态...`)
     void pollAsyncRun(integration.id, runId, token)
--- a/apps/web/src/views/DirectoryManagementView.vue
+++ b/apps/web/src/views/DirectoryManagementView.vue
@@ -432,10 +432,61 @@
           <button v-if="selectedIntegration" class="directory-admin__button directory-admin__button--secondary" type="button" :disabled="approvalCardBusy" @click="void saveApprovalCardPublicAppUrl()">
             保存对外访问地址
           </button>
-          <button class="directory-admin__button directory-admin__button--secondary" type="button" :disabled="busy || !selectedIntegration" @click="void syncIntegration()">
+          <button class="directory-admin__button directory-admin__button--secondary" type="button" :disabled="busy || pollingRunId !== '' || !selectedIntegration" @click="void syncIntegration()">
             手动同步
           </button>
+          <button class="directory-admin__button directory-admin__button--secondary" type="button" :disabled="busy || previewBusy || pollingRunId !== '' || !selectedIntegration" @click="void previewSync()">
+            {{ previewBusy ? '预览中...' : '预览同步' }}
+          </button>
+          <button class="directory-admin__button directory-admin__button--secondary" type="button" :disabled="busy || asyncSyncBusy || pollingRunId !== '' || !selectedIntegration" @click="void startAsyncSync()">
+            {{ asyncSyncBusy ? '启动中...' : pollingRunId ? '后台同步中...' : '后台同步' }}
+          </button>
         </footer>
+        <p v-if="pollingRunId" class="directory-admin__hint">
+          后台同步进行中（运行 {{ pollingRunId }}），每 5 秒自动刷新运行记录。后台模式不返回一次性临时凭据。
+        </p>
+
+        <section v-if="selectedIntegration && (previewBusy || previewResult)" class="directory-admin__section">
+          <div class="directory-admin__section-head">
+            <div>
+              <h3>同步预览（dry-run）</h3>
+              <p class="directory-admin__hint">只读预演：按同步的方式拉取钉钉通讯录并计算将发生的变更，不写入任何数据。</p>
+            </div>
+            <button
+              class="directory-admin__button directory-admin__button--secondary"
+              type="button"
+              :disabled="busy || previewBusy"
+              @click="void previewSync()"
+            >
+              {{ previewBusy ? '预览中...' : '重新预览' }}
+            </button>
+          </div>
+          <div v-if="previewBusy" class="directory-admin__empty">同步预览生成中（将完整拉取一次钉钉通讯录）...</div>
+          <article v-else-if="previewResult" class="directory-admin__schedule-card">
+            <div class="directory-admin__chips">
+              <span class="directory-admin__chip">部门 {{ previewResult.departmentsSeen }}</span>
+              <span class="directory-admin__chip">账号 {{ previewResult.accountsSeen }}</span>
+              <span class="directory-admin__chip">将新增账号 {{ previewResult.wouldCreateAccounts }}</span>
+              <span class="directory-admin__chip" :class="{ 'directory-admin__chip--warning': previewResult.wouldDeactivateAccounts > 0 }">
+                将停用 {{ previewResult.wouldDeactivateAccounts }}
+              </span>
+              <span class="directory-admin__chip" :class="{ 'directory-admin__chip--danger': previewResult.wouldDeactivateLinkedAccounts > 0 }">
+                其中已绑定 {{ previewResult.wouldDeactivateLinkedAccounts }}
+              </span>
+            </div>
+            <p class="directory-admin__hint">
+              自动准入：{{ previewResult.autoAdmissionMode === 'auto_for_scoped_departments'
+                ? `符合条件 ${previewResult.autoAdmissionCandidateCount} 人 · 缺少邮箱跳过 ${previewResult.autoAdmissionSkippedMissingEmailCount} 人 · 命中排除部门 ${previewResult.autoAdmissionExcludedCount} 人`
+                : '已关闭' }}
+            </p>
+            <p class="directory-admin__hint">
+              将新增账号：{{ formatPreviewSample(previewResult.sampledNewAccounts, previewResult.wouldCreateAccounts) }}
+            </p>
+            <p class="directory-admin__hint">
+              将停用账号：{{ formatPreviewDeactivationSample(previewResult.sampledDeactivations, previewResult.wouldDeactivateAccounts) }}
+            </p>
+          </article>
+        </section>
 
         <section v-if="selectedIntegration" ref="accountsSectionRef" class="directory-admin__section">
           <h3>当前概览</h3>
@@ -1883,6 +1934,23 @@
               账号 {{ readNumericStat(run.stats, 'accountsSynced') }} / 部门 {{ readNumericStat(run.stats, 'departmentsSynced') }} /
               待确认 {{ readNumericStat(run.stats, 'pendingCount') }} / 已链接 {{ readNumericStat(run.stats, 'linkedCount') }}
             </p>
+            <!-- R5 per-run change summary (#4054): keys are additive jsonb — gate each line on key
+                 presence so pre-R5 runs (keys absent) don't render a misleading all-zero summary. -->
+            <p
+              v-if="hasRunStat(run.stats, 'accountsCreatedCount') || hasRunStat(run.stats, 'accountsUpdatedCount') || hasRunStat(run.stats, 'departmentsCreatedCount')"
+              class="directory-admin__hint"
+            >
+              本次变更：账号 新增 {{ readNumericStat(run.stats, 'accountsCreatedCount') }} · 更新 {{ readNumericStat(run.stats, 'accountsUpdatedCount') }} · 停用 {{ readNumericStat(run.stats, 'accountsDeactivatedCount') }}
+              / 部门 新增 {{ readNumericStat(run.stats, 'departmentsCreatedCount') }} · 更新 {{ readNumericStat(run.stats, 'departmentsUpdatedCount') }} · 停用 {{ readNumericStat(run.stats, 'departmentsDeactivatedCount') }}
+            </p>
+            <p
+              v-if="hasRunStat(run.stats, 'managerCoverage') || hasRunStat(run.stats, 'durationMs')"
+              class="directory-admin__hint"
+            >
+              <template v-if="hasRunStat(run.stats, 'managerCoverage')">主管绑定覆盖 {{ formatManagerCoverage(run.stats) }}</template>
+              <template v-if="hasRunStat(run.stats, 'managerCoverage') && hasRunStat(run.stats, 'durationMs')"> · </template>
+              <template v-if="hasRunStat(run.stats, 'durationMs')">耗时 {{ formatRunDurationMs(readNumericStat(run.stats, 'durationMs')) }}</template>
+            </p>
             <p v-if="readStringStat(run.stats, 'diagnosticTitle')" class="directory-admin__hint">
               诊断：{{ readStringStat(run.stats, 'diagnosticTitle') }}
               · 子部门 {{ readNumericStat(run.stats, 'rootDepartmentChildCount') }}
@@ -1957,6 +2025,24 @@ type DirectoryRun = {
   finishedAt: string | null
   stats: Record<string, unknown>
   errorMessage: string | null
+}
+
+// Mirrors DirectorySyncPreview from packages/core-backend/src/directory/directory-sync.ts
+// (DT-OPS-02 preview endpoint) — read-only dry-run counts + samples, nothing written.
+type DirectorySyncPreview = {
+  integrationId: string
+  integrationName: string
+  departmentsSeen: number
+  accountsSeen: number
+  wouldCreateAccounts: number
+  wouldDeactivateAccounts: number
+  wouldDeactivateLinkedAccounts: number
+  autoAdmissionMode: 'manual_only' | 'auto_for_scoped_departments'
+  autoAdmissionCandidateCount: number
+  autoAdmissionSkippedMissingEmailCount: number
+  autoAdmissionExcludedCount: number
+  sampledNewAccounts: Array<{ externalUserId: string; name: string }>
+  sampledDeactivations: Array<{ externalUserId: string; name: string; linked: boolean }>
 }
 
 type DirectoryObservationStatus =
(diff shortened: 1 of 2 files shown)


zensgit:
# Adversarial review — PR #4069 (directory-web async sync polling + dry-run preview + R5 run-card stats)

- **PR**: #4069 `claude/dt-directory-async-ui` → main
- **Head reviewed**: `b5266989c785dd34127bdc1a7607d383fc6e1f84` (verified == PR head at review time)
- **Base**: merge-base = `0e4d4f416` (#4049 preview-integrity fix — the PR's claimed prerequisite is on base, verified)
- **Scope**: exactly 2 files — `apps/web/src/views/DirectoryManagementView.vue` (+361/−11), `apps/web/tests/directoryManagementView.spec.ts` (+411/−0). Zero backend edits, zero routing/CI wiring. Verified against full diff.
- **Method**: refute-first. Disposable worktree `/tmp/ms2-rev-pr4069` at the PR head; all wire-shape claims checked against `origin/main` backend source (not the diff); implementer's mutations M5/M6a/M9 re-fired; 4 additional adversarial mutations of my own; full required web gate + vue-tsc run locally.

## Verdict: **APPROVE with hardening**

No P1. Every wire-shape, gate, and lifecycle-correctness claim I attacked held up — including the two spots most likely to be mock-not-contract (the 409 body and the #4054 R5 key names/semantics). The residual findings are test-coverage gaps around the poll lifecycle guards (house rule: neuter-able guard with green tests = P2) plus small hardenings. None require blocking the merge if the P2 tests land as an immediate follow-up; landing them in this PR would be better.

---

## Findings

### P2-1 — Poll zombie-continuation guard (`runPollToken` check) is untested: deleting it leaves 64/64 green
`DirectoryManagementView.vue` ~L4534 (`if (token !== runPollToken) return` in `pollAsyncRun`).

My mutation **MX1** deleted the token check entirely → **all 64 spec tests still pass**. This line is the PR body's central lifecycle claim ("a stale continuation can't resurrect the loop") and it is load-bearing: without it, a poll GET that is *in flight* when `stopRunPolling()` runs (integration switch, unmount) will, on resolve, (a) write the **old** integration's runs into `runs.value` — now rendering under the **new** integration's panel, (b) re-arm the 5s timer, resurrecting the loop forever (fetches continue after unmount), and (c) on a terminal run, fire `refreshAfterDirectorySync(oldIntegrationId)`, repainting accounts/alerts/review/schedule panels with the wrong integration's data. The existing unmount test doesn't catch this because `flushUi(8)` settles the fetch *before* unmount — the in-flight race is never exercised.

**Fix**: one fake-timer test with a manually-deferred runs-GET promise: start async sync → advance 5s so a poll is in-flight → unmount (or `selectIntegration` to a second integration) → resolve the GET with a terminal run → assert zero state writes and zero further fetches. ~30 lines, kills MX1.

### P2-2 — Integration-switch hygiene in `selectIntegration` is untested: both new lines individually deletable, 64/64 stay green
`DirectoryManagementView.vue` ~L3642–3646 (the two lines added at the top of `selectIntegration`).

- **MX2** deleted `if (pollingRunId.value && pollingIntegrationId !== integrationId) stopRunPolling()` → green. Failure it guards (with the token check intact): switch integration mid-run → poll keeps hammering the OLD integration's `/runs` endpoint every 5 s and overwrites the 运行记录 panel of the newly selected integration with the old integration's runs — mislabeled data on an admin surface, plus terminal-branch `refreshAfterDirectorySync(oldId)` repaint. Realistic for multi-corp tenants.
- **MX4** deleted `if (previewResult.value && previewResult.value.integrationId !== integrationId) previewResult.value = null` → green. Failure: integration A's dry-run counts/sample names render under integration B's header.

The PR body honestly flags the switch path as untested; honesty doesn't change the house rule. **Fix**: one test — mount with two integrations, start poll (or render a preview) on A, click B, assert poll stopped (no further `/runs` fetches after 5s advance) and preview card gone. Kills MX2+MX4.

### P3-1 — Unbounded polling when the run never reaches terminal on page 1 (author-flagged); a max-poll-duration bound is a ~6-line harden
`pollAsyncRun` reschedules unconditionally when the run is missing or `running`. Two ways this never terminates: (a) run permanently paged out of `page=1&pageSize=10` (needs sustained run churn — unlikely); (b) **crashed worker with the scheduler disabled**: the run row stays `running` until some *new claim* reclaims the stale lease (`directory-sync.ts` L2777 reclaim runs only on claim), and this tab's 手动同步/预览/后台同步 buttons are all locked on `pollingRunId` — so the very action that would reclaim the lease is disabled in this tab, while it polls every 5 s indefinitely. Escape hatches exist (navigate away, other tab, scheduler if enabled). Suggest: stop after ~30 min (`已停止自动轮询，运行仍在进行，请手动刷新` + unlock buttons). Severity honestly assessed: operational annoyance, not corruption — P3.

### P3-2 — FAIL-SAFE confirm gate fails OPEN under stale config (cross-admin TOCTOU)
`startAsyncSync` reads `integration.config.admissionMode` from the integrations list loaded at mount/last refresh. Verified the list serializer (`summarizeIntegration`, directory-sync.ts L1539) always emits a normalized `admissionMode`, so the gate is real on fresh data. But if another admin flips `manual_only → auto_for_scoped_departments` after this tab loaded, the click shows **no warning** and the one-time temp-password packets are silently discarded — fail-open. The owner-ratified design is a client-side warn-confirm, so full fail-closed (backend refusing `async:true` + auto-admission without an explicit ack flag) is out of scope for a web-only PR; recording it as the follow-up direction. Cheap client narrow: re-fetch the single integration's config immediately before evaluating the gate. P3.

### P3-3 — The flagged scope addition (manual-path 409 branch) is itself untested
`syncIntegration` ~L4364–4371. My mutation **MX3** removed the whole branch → 64/64 green. It's ~6 code lines (not 3 as the PR body says), verified behavior-neutral otherwise (early return before throw; `finally` still clears `busy`; catch parity). The PR body's "409 rendering on both paths" test claim is accurate for the two *new* paths only. Keep the addition (it's the right consistency call), add the one-mock test.

### NIT-1 — Poll-error stop message loses the "已停止轮询" guidance on structured HTTP errors
`pollAsyncRun` error branch: `setStatus(readApiError(body, '后台同步状态查询失败，已停止轮询，请手动刷新运行记录'), 'error')`. On a real 500 the server's English `Failed to load sync runs` wins and the user is never told polling stopped (the informative message shows only for network/parse failures). Behavior itself (stop on first error, buttons unlock, re-click yields the 409-with-runId recovery path) is defensible and matches the claim; just append the guidance to whatever message is shown.

### NIT-2 — Button-disable matrix inconsistencies (all server-guarded by the lease, so cosmetic)
In-card 重新预览 button (`:disabled="busy || previewBusy"`) is not disabled while `pollingRunId` is active, unlike the footer 预览同步 button; 手动同步/后台同步 are not disabled while `previewBusy`. Every one of these dead-ends in the benign 409 message, so no correctness impact.

### NIT-3 — `startAsyncSync` entry guard checks `pollingRunId` only
Not `asyncSyncBusy`; during the `ElMessageBox.confirm` await the footer button is technically clickable (real browsers: the modal mask blocks; jsdom-only exposure). Even the pathological double-202 resolves safely — the second `stopRunPolling()` token-bump kills the first poll, so exactly one loop survives. Belt-and-braces only.

---

## Probe table (parent's gate spec)

| # | Probe | Verdict | Evidence |
|---|---|---|---|
| 1 | Timer/poll lifecycle | **PASS (code) / PARTIAL (coverage)** | Token check placed after every await and **before any state write** (`runs.value` write is post-check); self-rescheduling `setTimeout` makes overlap structurally impossible; `stopRunPolling` bumps token before clearing state; `startAsyncSync` stop-then-capture-token ordering correct. Error path = stop-with-message (honest, see NIT-1). Terminal semantics verified against backend: run row is INSERTed `status='running'` (claim = lease, directory-sync.ts L2816) and only ever moves to `completed`/`failed` — no `pending` state exists, so `status !== 'running'` = terminal is sound. Coverage gaps → P2-1/P2-2. Paged-out gap → P3-1. |
| 2 | FAIL-SAFE completeness | **PASS** | Gate reads `config.admissionMode` from the integrations-list payload; `summarizeIntegration` always serializes it, normalized via `normalizeAdmissionMode` (invalid → `manual_only`, i.e. the gate never throws and never sees garbage). `selectedIntegration` null → early return (fail-closed). Residual fail-open = cross-admin staleness only → P3-2. Dialog copy states discard + unrecoverable + recommends 手动同步, matching the ratified wording. |
| 3 | 409 wire shape (mock-is-not-the-contract) | **PASS — byte-for-byte** | Backend: `jsonError(res, 409, 'DIRECTORY_SYNC_IN_PROGRESS', message, { activeRunId })` → `{ ok:false, error:{ code, message, details:{ activeRunId } } }` (util/response.ts L3-5; emission sites admin-directory.ts L327/343/365, all three lease paths). `apiFetch` returns the raw Response on non-ok (api.ts — only 401/403 get special handling), so the view's own `readJson` + `error.details.activeRunId` path is the real wire path, not an apiFetch-surface assumption. Spec mock (`createSyncInProgressResponse`) reproduces the shape exactly incl. status 409 and the real English message, and asserts the raw message does NOT leak. `activeRunId: null` (possible server-side) handled: falls back to the no-runId message. M9 re-fired RED. |
| 4 | R5 stats keys vs #4054 | **PASS — 10/10 keys + semantics** | Checked against `gh pr diff 4054` (head `9ccdedf3f`): `accountsCreatedCount/UpdatedCount/DeactivatedCount`, `departmentsCreatedCount/UpdatedCount/DeactivatedCount`, `managerCount`, `linkedManagerCount`, `managerCoverage`, `durationMs` — all exact. **Semantics verified, not just names**: #4054 tests pin `managerCoverage` = 0.5 for 1/2 (fraction, 0–1) → view's `Math.round(coverage*100)%` is correct (spec pins `83%（5/6）` for 0.8333); `durationMs = Date.now() - runStartedAtMs` (milliseconds) → humanizer correct. Key-presence gating direction confirmed fail-safe-hidden (pre-R5 absent-keys negative test exists; M8 claimed RED, consistent with the v-if structure). #4054 still OPEN — a rename there hides lines, never lies. |
| 5 | Required-gate reality | **PASS — ran it** | `directoryManagementView` is in `run-required-web-tests.sh`'s filter list at the PR head; the script runs in `web-tests.yml` **unconditionally** (no path filter, PR + merge_group + push). Local runs at head `b5266989c`: spec alone **64/64**; full gate **141 files / 1811 tests, all green**; `vue-tsc -b` exit 0. All three implementer numbers reproduced exactly. |
| 6 | Scope audit | **PASS** | Exactly 2 files, nothing else in the three-dot diff. Manual-path 409 = ~6 lines + comments, behavior-neutral otherwise (verified: early-return skips only the throw; `finally` intact). `refreshAfterDirectorySync` extraction is a faithful 1:1 move of the seven-load `Promise.all` incl. the conditional departments force-reload; the mid-refresh switch race it inherits is **pre-existing** on the manual path (parity, not a regression). PR body has no bare closing keywords (refs #3915/#4049/#4054 only). Spec diff is pure addition — no sibling coverage removed. |

## Mutation record (all applied to committed impl in the disposable worktree, spec re-run, file restored via `git checkout --` each time)

| Mutation | What | Result |
|---|---|---|
| M5 (re-fired) | Remove auto-admission confirm gate | **RED — 2 failed** (matches claim) |
| M6a (re-fired) | `status !== 'running' && false` (poll never terminal) | **RED — 2 failed** (matches claim) |
| M9 (re-fired) | `readActiveRunId` → always `''` | **RED — 2 failed** (matches claim) |
| MX1 (mine) | Delete `if (token !== runPollToken) return` | **GREEN 64/64 — survivor → P2-1** |
| MX2 (mine) | Delete switch-stop line in `selectIntegration` | **GREEN 64/64 — survivor → P2-2** |
| MX3 (mine) | Remove manual-path 409 branch | **GREEN 64/64 — survivor → P3-3** |
| MX4 (mine) | Delete preview clear-on-switch | **GREEN 64/64 — survivor → P2-2** |

## Bottom line

The code is right; the two highest-value guards in it are unproven. Ship after (or with) the two ~30-line lifecycle tests in P2-1/P2-2 — they are cheap, they kill three of my four surviving mutations, and they convert the PR body's strongest claim from prose into a tripwire. P3s and NITs at author's discretion.

*Reviewed by Claude (adversarial gate), 2026-07-10. Worktree: /tmp/ms2-rev-pr4069 (removed after review).*

zensgit:
# Gate hardening applied — P2-1 + P2-2 + P3-1 closed at `bffe7fec0`

Follow-up to the adversarial gate review above (head reviewed: `b5266989c`). All three actioned findings land in one commit on this branch; no product-behavior changes beyond the review's own P3-1 suggestion.

## What landed

**P2-1 — poll zombie-continuation race is now a tripwire.** New test: the poll's `/runs` GET is left as a manually-deferred promise, the view switches from integration A to integration B WHILE that GET is in flight, then the stale GET resolves with a terminal run for A. Asserts: A's runs are NOT written into B's panel, `refreshAfterDirectorySync(A)` does NOT fire (call count pinned), and the timer is NOT re-armed (zero further fetches after +30s fake time). Deleting the `if (token !== runPollToken) return` check now goes RED.

**P2-2 — both integration-switch hygiene lines are now individually covered.** New two-integration test: A has a rendered dry-run preview card AND an active background poll; click B. Asserts: A's preview numbers/card do not render under B's header (kills the preview-clear deletion), and zero further fetches after +10s fake time (kills the poll-stop deletion). Each deleted line trips a different assertion.

**P3-1 — max-poll-duration bound (12 lines).** `pollAsyncRun` now stops after 10 minutes of the run staying non-terminal (`DIRECTORY_RUN_POLL_MAX_DURATION_MS`, clock started at the 202), sets `已停止自动轮询（超过 10 分钟），运行可能仍在进行，请前往下方运行记录手动刷新查看最新状态` as an error status, and unlocks the sync/preview buttons via the normal `stopRunPolling()` path — so the crashed-worker scenario (review P3-1b) no longer wedges the tab into polling forever with its own lease-reclaiming buttons locked. Covered by a fake-timer test that crosses the cap and pins the message, button re-enable, and zero further fetches.

Not actioned in this pass (per gate "author's discretion"): P3-2 (stale-config TOCTOU narrow), P3-3 (manual-path 409 mock test), NIT-1/2/3.

## Mutation table (protocol: real impl committed first, mutation applied, spec re-run, `git checkout --` restore each time)

| Mutation | What was deleted | Result |
|---|---|---|
| MX1 (gate's P2-1 survivor) | `if (token !== runPollToken) return` in `pollAsyncRun` | **RED — 1 failed** (new in-flight-race test: stale continuation fired `refreshAfterDirectorySync(A)`, call count 19≠13) |
| MX2 (gate's P2-2 survivor) | switch-stop line in `selectIntegration` | **RED — 2 failed** (new switch test: extra `/runs` fetch after +10s; new race test also caught the polling banner surviving the switch) |
| MX4 (gate's P2-2 survivor) | preview clear-on-switch line in `selectIntegration` | **RED — 1 failed** (A's `将新增账号 3` + dry-run card rendered under B's header) |
| MX-P3 (new guard, self-fired) | the whole 10-min cap block in `pollAsyncRun` | **RED — 1 failed** (poll still alive past the cap; no 已停止自动轮询 message) |

All 4 gate survivors relevant to this PR's lifecycle guards are now killed (MX3 was P3-3, left open by design). After each restore: spec back to green on the committed impl.

## Test evidence at `bffe7fec0`

- Spec alone: `tests/directoryManagementView.spec.ts` **67/67** (was 64; +3 new)
- Full required web gate (`scripts/run-required-web-tests.sh`): **141 files / 1814 tests, all green** (was 1811)
- `vue-tsc -b`: exit 0

Diff: +186 lines total — 12 impl lines (P3-1 bound) in `DirectoryManagementView.vue`, rest is spec. Not armed; landing stays gated on the main loop.

zensgit:
Final window protocol executed: rebased onto post-#4046 main (`b94b46bf3`, clean, 3 commits survived); spec 67/67 + FULL required web gate 144 files / 1828 tests green on the rebased head; load-bearing mutation replayed (runPollToken staleness check dropped ⇒ RED, restored). Armed alone — the pool's last code window.

PR #4072, opened 2026-07-10
fix(files): F5 blob compensating-delete — deleteByKey + delete-route root-cause fix + sweep

zensgit:
## Summary

Implements the RATIFIED F5 design lock (`f5-design-lock.md`, 2026-07-10): compensating deletion for orphaned physical blobs left behind by F2's tombstone-first delete, plus the root-cause fix in the delete route itself. refs #3925. Non-permission-layer (storage mechanism + background reconciliation) — ACL gates are untouched.

Base: `main` (F3 `#4063` @ `7d4cb8027` already merged, required per the design lock's dependency ordering).

## Per-gate self-certification (GF5-1 .. GF5-8)

- **GF5-1 (consume marker)**: new migration `zzzz20260710150000_add_files_blob_purged_at.ts` adds `files.blob_purged_at timestamptz null`, additive, `checkColumnExists`/`checkTableExists`-guarded, no backfill (NULL = not yet confirmed purged). Sweep query filters `blob_purged_at IS NULL`.
- **GF5-2 (deleteByKey)**: added to the `StorageProvider` interface, `LocalStorageProvider` (`resolveWithinBase` containment, `fs.unlink`, ENOENT swallowed as success, other errors rethrown), `S3StorageProvider` (delegates to `delete`, mirroring `downloadByKey`'s G6 note), and `StorageServiceImpl` (delegate + `file:deleted` emit, mirroring `downloadByKey`).
- **GF5-3 (poison/null skip)**: sweep imports `isPoisonedStorageKey` from `filesStorageKey.ts` directly (no reimplementation); a poisoned key is skipped (never resolved/deleted/stamped). Null `storage_key` rows are excluded at the SQL level (`storage_key IS NOT NULL`) — the sweep does not repeat the delete route's url-fallback.
- **GF5-3b (shared-key defense in depth)**: before every physical delete, `SELECT count(*) FROM files WHERE storage_key=$key AND blob_purged_at IS NULL AND id<>$id` must be 0; `>0` skips (both rows, no keep-one-winner).
- **GF5-4 (grace period)**: `deleted_at < now() - make_interval(hours => $1)`, default 24h, clamp `[1h, 30d]` via `FILES_ORPHAN_BLOB_RETENTION_GRACE_HOURS`.
- **GF5-5 (failure grading)**: ENOENT is swallowed inside `deleteByKey` itself (single source of truth) so callers just see resolve/reject; non-ENOENT failures leave `blob_purged_at` NULL and warn for next-tick retry. No attempt cap (per design lock, kept simple for v1).
- **GF5-6 (sweep + registration)**: new `services/files-orphan-blob-retention.ts`, skeleton copied from `multitable/attachment-orphan-retention.ts` (`setInterval` + `unref` + batch/clamp + schema-not-ready tolerance + stop closure). **Default OFF** regardless of `NODE_ENV` (`FILES_ORPHAN_BLOB_RETENTION_ENABLED=true` to opt in) — stricter than attachment-cleanup's prod-default-on. Registered in `index.ts` alongside the other three retention schedulers (import, `stopFilesOrphanBlobRetention` field, registration block, shutdown call — four spots). No leader lock (idempotent + ENOENT-tolerant ⇒ at-least-once safe, per design lock).
- **GF5-7 (delete-route root-cause fix)**: `files.ts`'s `DELETE /api/files/:id` active branch now resolves the physical key via `resolveActiveStorageKey(detail)` — the SAME resolution `download`/`info` already use (persisted `storage_key`, poison-rejected, else `deriveStorageKeyFromUrl` fallback) — and calls `storage.deleteByKey(resolved.key)` instead of the old `storage.delete(id)`, which located the object through the id-keyed in-memory disk-scan index (drifts to `md5(relpath)` after any rebuild — the same root cause F3 fixed on the read side). Success (incl. ENOENT) stamps `blob_purged_at = now()` synchronously; failure leaves it NULL for the sweep to retry. A poisoned/unresolvable key skips the physical delete entirely (never guessed at). Tombstone-first ordering and the ACL gate are **untouched**; the `missing` branch still uses `storage.delete(id)` unchanged, exactly as specified.
- **GF5-8 (test matrix + mutation)**: see below.

## Files changed

- `packages/core-backend/src/db/migrations/zzzz20260710150000_add_files_blob_purged_at.ts` (new)
- `packages/core-backend/src/services/StorageService.ts` (`deleteByKey` on provider interface + Local/S3/Impl)
- `packages/core-backend/src/routes/files.ts` (delete route active branch)
- `packages/core-backend/src/services/files-orphan-blob-retention.ts` (new sweep module)
- `packages/core-backend/src/index.ts` (registration, 4 spots)
- Tests: `tests/unit/storageServiceLocalProvider.test.ts` (deleteByKey, real fs), `tests/unit/files-orphan-blob-retention.test.ts` (new, mocked), `tests/integration/files-orphan-blob-retention.db.test.ts` (new, real DB isolated schema), `tests/unit/files-delete-tombstone-order.test.ts` (updated mocks for the new deleteByKey call + 2 new skip-path cases), `tests/integration/attendance-files-acl.test.ts` (2 new real-DB route-level GF5-7 tests)
- CI wiring: `vitest.config.ts` (exclude the new `.db.test.ts` from the no-DB job) + `.github/workflows/plugin-tests.yml` (wire it into the existing "Run attendance integration tests" real-DB step, alongside `files-storage-key-migration.db.test.ts`)

## Test plan (all actually run, not just written)

**Real-DB runs** used a fresh, isolated local database (`metasheet_f5_test`, migrated via `pnpm --filter @metasheet/core-backend migrate`), never the shared CI `metasheet_test`.

- [x] GF5-8 matrix 1–8, all real-DB (isolated per-test Postgres schema, real local-filesystem `StorageServiceImpl` in a temp dir — not mocked):
  1. delete + stamp ✓ 2. idempotent second run selects nothing (no re-delete, spy asserts `deleteByKey` not called) ✓ 3. ENOENT (never-written key) treated as success, stamped ✓ 4. within grace → not scanned ✓ 5. active row (`deleted_at NULL`) → never scanned ✓ 6. poisoned key → skipped, untouched ✓ 7. null key → excluded at query level ✓ 8. non-ENOENT failure (unlink on a directory → EPERM) → not stamped, retried next tick ✓
  - plus GF5-3b (two un-purged rows sharing one key → both skipped) and a mixed-batch case.
- [x] Migration test: column added additively, idempotent re-run, `down()` removes it (no-op if absent), no backfill on pre-existing rows.
- [x] `deleteByKey` containment: real fs, traversal key rejected, out-of-base file never touched (unit, `storageServiceLocalProvider.test.ts`).
- [x] `deleteByKey` idempotency at the provider level (real fs, ENOENT swallowed, double-delete never throws).
- [x] GF5-7 route-level, real-DB, real running server (`attendance-files-acl.test.ts`): active delete → blob physically gone + `blob_purged_at` stamped synchronously in the same request; poisoned-key row → delete succeeds (200) but blob left untouched, `blob_purged_at` stays NULL.
- [x] `files-delete-tombstone-order.test.ts` (mocked-db unit): updated for the new `deleteByKey`-based call order (`tombstone-update → storage-deleteByKey → blob-purged-update`); added poison-skip and null-key-skip cases.

**Mutation testing (≥4 required, all confirmed red then reverted clean)**:
1. Neutered `blob_purged_at IS NULL` filter → real-DB test **#2 (idempotency) went red** (selected + re-purged a row that should have been excluded). Reverted, confirmed clean (`diff` against pre-mutation backup empty).
2. Neutered grace check (`< now() - grace` → `< now() + grace`, always-true, keeps parameter typing intact) → real-DB test **#4 went red** (plus the mixed-batch test, which also exercises grace) — everything else stayed green. Reverted clean.
3. Neutered `isPoisonedStorageKey` skip (`if (false && ...)`) → real-DB test **#6 went red**, everything else green. Reverted clean.
4. Neutered ENOENT-as-success swallow in `LocalStorageProvider.deleteByKey` (`if (false && code === 'ENOENT')`) → **both** the unit-level mutation tripwire test and the idempotent-double-delete test went red in `storageServiceLocalProvider.test.ts`, AND real-DB test **#3 went red** — everything else green. Reverted clean.

(Note: my first attempt at mutation 2 — deleting the WHERE clause text outright while still passing 2 SQL params — accidentally broke Postgres parameter-type inference for the now-unreferenced `$1`, which the sweep's outer catch silently swallowed, producing a much broader false-red across most tests. Re-did it as an always-true comparison that keeps `$1` referenced, isolating the mutation to grace semantics only. Documented here for transparency, not left in the diff.)

- [x] `tsc --noEmit`: 0 errors (checked both mid-implementation and after full mutation-testing revert).
- [x] Full no-DB suite: 438 test files / 5950 tests passed, 0 failures, 1566 skipped (pre-existing `describeDb`/`describeIfDatabase` self-skips), both before and after mutation testing.

## Deviations / notes for review

- `deleteByKey` was **not** added to the public `StorageService` interface in `types/plugin.ts` — mirroring F3's own precedent: `downloadByKey` isn't there either, since `files.ts` always uses the concrete `StorageServiceImpl` type, never the plugin-facing interface.
- The sweep's default storage instance mirrors `files.ts`'s `getStorageService()` construction exactly (same `STORAGE_PATH`/`STORAGE_BASE_URL` env vars) so it points at the same physical root — but is always local-only (S3 isn't wired into `files.ts` today either; `S3StorageProvider.deleteByKey` exists per the design lock's ask but is currently unreachable in practice, matching `downloadByKey`'s status quo).
- Left uncommitted: local `packages/core-backend/uploads/*` test-run artifacts (pre-existing repo gap — no `.gitignore` entry for `uploads/`; out of scope for this design lock, cleaned up before every commit).
- Not touched: tracker/verification/closeout docs, F4/F6 files, multitable attachment logic (read-only reference only), any ACL/auth judgment code, the tombstone-first ordering, or the `missing`-branch delete path.

## Reviewer focus (per design lock's "opus 对抗审" brief)

- Idempotency's real guard (mutation #1/#2 above) — not just a passing assertion.
- Poison-skip sharing the exact same `isPoisonedStorageKey` source as F3 (no reimplementation/drift).
- Grace-period race safety.
- `deleteByKey` containment (no traversal escape).

Not merged — for review only, per design lock instructions.


zensgit:
Follow-up verification after the initial PR body (per self-review):

- **Lint**: ran `eslint` scoped to every non-test file this PR touches (migration, `StorageService.ts`, `files.ts`, `files-orphan-blob-retention.ts`, `index.ts`) — 0 errors, 0 warnings on any line this PR added. `index.ts` shows pre-existing warnings (`no-explicit-any`, unused `ILogger` type import) at lines far from my 4 insertion points (82, 256, 1861-1867, 2762) — confirmed via diff, not introduced here. Test files are eslint-ignored (matches the existing `tests/` ignore pattern, same as every other test file in the repo).
- **`file:deleted` event payload**: `deleteByKey`'s emit uses `{ fileId: storageKey }` (the `<uuid>/<basename>` key) instead of the uuid alone that `delete(id)` emits. Checked `grep -rn "file:deleted" packages/core-backend/src` — the only two emit sites are `delete()` and the new `deleteByKey()` in `StorageService.ts`; there are no consumers anywhere in `src`. Moot today, but flagging in case a future listener assumes `fileId` is always the bare uuid.

**Design-scoped residual tail (informational, not a gap — matches the ratified lock as-is)**: a legacy row with `storage_key IS NULL` whose route-time `deleteByKey` attempt fails non-ENOENT leaks its blob and is **not** sweep-reclaimable — the sweep query filters `storage_key IS NOT NULL` and deliberately does not repeat the delete route's one-shot `deriveStorageKeyFromUrl` fallback (GF5-3's stated boundary: the sweep acts only on a persisted key, never guesses). This is what the design lock specifies, not something to fix here — noting it so the residual-orphan tail after this PR is understood to be bounded to persisted-key rows only.

PR #4075, opened 2026-07-10
feat(multitable): formula engine — batch 2, 26 standard scalar functions (cluster-C path-1)

zensgit:
## Summary

Cluster-C path-1 **batch 2**: registers the 26 standard scalar functions left over after #3959 (batch-1, MERGED). Same zero-design-discretion, audit-checklist pattern as batch-1: pure-scalar/variadic-scalar only (no range/array semantics — SUMIFS/COUNTIFS-style criteria functions stay excluded by design), reusing the engine's existing error-sentinel set, no new sentinels invented.

**Pre-flight check**: grepped origin/main's `engine.ts` for all 26 names before starting — zero were already registered. `functions.set` count was 91 (78 baseline + batch-1's 13); it is now **117** (91 + 26).

## Function list (26)

- **Trig/math (13):** SIN, COS, TAN, ASIN, ACOS, ATAN, ATAN2, LOG10, EVEN, ODD, FACT, QUOTIENT, MROUND
- **Numeric predicates (2):** ISEVEN, ISODD
- **Text (5):** EXACT, CHAR, CODE, CLEAN, ISTEXT
- **Date/time (6):** DATEVALUE, TIMEVALUE, TIME, ISOWEEKNUM, YEARFRAC, NETWORKDAYS

## Semantic decision table (every non-obvious call, one row each)

| Function | Input → Output | Basis |
|---|---|---|
| ATAN2(x, y) | Excel arg order is **(x_num, y_num)**, NOT JS's `atan2(y, x)` — result = `atan(y_num / x_num)` = `Math.atan2(y_num, x_num)` | Locked by Microsoft's own worked example: `ATAN2(1, -1) = -0.785398163` (angle to point (1,-1) is -45°). Verified via MS docs before implementing. |
| ATAN2(0, 0) | → `#DIV/0!` | Documented Excel behavior; reuses the engine's existing division sentinel. |
| ASIN/ACOS out of [-1,1] | → `#NUM!` | Matches the engine's existing LN/LOG10 domain-error convention (distinct from `#VALUE!` = non-numeric input). |
| LOG10(n) for n≤0 | → `#NUM!` | Same LN-domain-error convention. |
| FACT(negative) | → `#NUM!` | Standard domain error. |
| FACT(non-integer, e.g. 3.5) | → `#NUM!` | **Stricter than real Excel** (which silently truncates a fractional input to its integer part) — a deliberate divergence, chosen per the task's explicit "FACT 负数或非整" error-path instruction, and deliberately *inconsistent* with batch-1's GCD/LCM (which truncate non-integers instead of rejecting). Called out here because it's the one function in this batch that isn't Excel-parity. |
| EVEN/ODD | Round **away from zero** (magnitude grows, sign preserved); `EVEN(0)=0`, `ODD(0)=1` | Excel convention. |
| ISEVEN/ISODD(non-integer) | Truncates toward zero before testing parity | Excel convention, consistent with batch-1 GCD/LCM's own truncation (unlike FACT above). |
| QUOTIENT(n, 0) | → `#DIV/0!` | Reuses engine's division sentinel. |
| MROUND(n, m) opposite signs | → `#NUM!`; `MROUND(n, 0) = 0` | Excel convention. |
| CHAR(n) outside 1..255 | → `#VALUE!` | Documented single-byte scalar subset (not full Unicode/UNICHAR); this is the audited "CHAR 越界" error path. |
| CODE("") | → `#VALUE!` | Empty text has no first character. |
| DATEVALUE/TIME/TIMEVALUE | Return a real JS `Date` anchored at the Unix epoch (1970-01-01 local midnight) for time-only values | This engine represents every date/time as a real `Date` object, never an Excel-style fractional serial number — the epoch anchor is composable with the existing HOUR/MINUTE/SECOND extractors, matching every other date function in the file. |
| ISOWEEKNUM | Delegates to the engine's existing `WEEKNUM` | `WEEKNUM` (added pre-batch-1) already implements ISO-8601 week numbering per its own comment — reused instead of duplicating the Thursday-anchor math. |
| YEARFRAC basis | Only **basis 0** (US/NASD 30/360) implemented; any other explicit basis → `#VALUE!` | Task requirement: reject rather than silently compute a wrong day-count convention. |
| YEARFRAC + end-of-February | **No** EOM special-case (the classic "30/360 US EOM" rule is NOT applied) | Microsoft's own documentation admits basis-0 "may return an incorrect result ... when start_date is the last day in February" — this does not attempt to out-guess an admittedly-inconsistent host convention. Verified via MS Q&A + Wikipedia's Day Count Convention algorithm before implementing. |
| YEARFRAC(start, end) reversed | Order-independent, always non-negative | Matches Excel: `YEARFRAC(a,b) == YEARFRAC(b,a)`. |
| NETWORKDAYS | 2-arg **scalar core only** — inclusive Mon-Fri count | A holidays-range 3rd argument is explicitly **OUT OF SCOPE** for this batch: the `{fld}` single-value-reference model has no range semantics to express it. |
| NETWORKDAYS(start > end) | Negative count | Matches Excel's documented sign behavior (verified via MS docs — distinct from YEARFRAC, which is always non-negative). |

## Deliverables

1. `packages/core-backend/src/formula/engine.ts` — 26 `functions.set(...)` registrations + a `days360US` private helper (YEARFRAC), reusing existing helpers (`coerceDateValue`, `flattenValues`) where applicable.
2. `packages/core-backend/tests/unit/formula-functions-batch2.test.ts` — 22 tests (per-function correct-value + edge-case + error-sentinel goldens, incl. NETWORKDAYS weekend-crossing/same-day, ISOWEEKNUM year-boundary via verified ISO calendar dates, YEARFRAC non-default-basis reject).
3. `apps/web/src/multitable/utils/formula-docs.ts` — 26 catalog entries + 26 alphabetized `FORMULA_FUNCTION_ARITY` keys.
4. `apps/web/src/multitable/utils/meta-formula-labels.ts` — en+zh `FUNCTION_DESCRIPTIONS` for all 26.
5. Existing live drift guard (`formula-catalog-web-parity.test.ts`) and the 4 required web specs pass **unchanged** — no wiring updates needed.

## Mutation-sampling evidence (3 cuts, each independently reverted)

| # | Mutation | Result |
|---|---|---|
| 1 | `ATAN2`: swapped `Math.atan2(yn, xn)` → `Math.atan2(xn, yn)` | `ATAN2(1,-1)` test failed: expected `-0.785398163`, got `2.356194490192345` |
| 2 | `NETWORKDAYS`: removed the `dow !== 0 && dow !== 6` weekend filter | Same-Saturday test failed: expected `0`, got `1` |
| 3 | `FACT`: removed the `!Number.isInteger(n)` reject | `FACT(3.5)` test failed: expected `'#NUM!'`, got `6` |

All three reverted clean (verified via `diff -q` against a pre-mutation backup).

## Local verification

- Backend formula suite: `vitest run formula` → **302 passed / 52 skipped** (skipped = real-DB integration tests, expected in a sandboxed worktree).
- Web specs: `vitest run multitable-formula-catalog-parity multitable-formula-catalog-1a meta-formula-labels multitable-formula-editor` → **22/22 passed**.
- `node --experimental-strip-types --check` on both touched TS files: clean.
- `tsc --noEmit` for `@metasheet/core-backend`: clean.
- Post-implementation set-diff of all 26 names across engine registration ∩ catalog doc ∩ arity table ∩ i18n labels: zero gaps.

Not merging — leaving for review per standing convention.


zensgit:
## 独立对抗门禁 — MERGE_CLEAN(P2 硬化已落) ✅

26 函数逐一独立重算全对(报告 `/tmp/pr4075-gate-20260710.md`):error sentinel 全复用既有集合、四面 set-diff(engine∩docs∩arity∩labels en+zh)零豁口零碰撞、required 全绿且 test(20.x) head-scoped 确认真跑新测试+漂移守卫。mutation 台账:K1 QUOTIENT trunc→floor RED / K2 EXACT 大小写 RED;**两个幸存者**=P2(days360US 的 31 日调整载重零覆盖:neuter 后 22/22 仍绿,而 YEARFRAC(2026-01-31,2026-06-30) 正确 0.416667 vs 变异 0.413889)+P3(TIMEVALUE 分秒界未测)。

**硬化已按处方落地**(本 PR 内):4 条 goldens(YEARFRAC 31 日两条 + TIMEVALUE 12:75/12:30:99),刀验新 golden 真红(1 failed)后复原 22/22 绿。2 NIT(YEARFRAC EOM-二月刻意不实现已披露;Number('')===0 引擎级既有非回归)。

P1/P2 清零,落地。

PR #4085, opened 2026-07-11
fix(dingtalk): distinct outcome-unknown response for work-notification test-send

zensgit:
## What / why

Roadmap #4046 gave the DingTalk transport a `'send'` tier that marks network-error / timeout / 5xx / malformed-2xx outcomes with `isDingTalkOutcomeUnknown` — DingTalk may have actually delivered the message even though the client saw no usable response. Every other send path in the codebase (approval-card deliveries, attendance delivery worker, automation-executor alerts) already discriminates this marker into a distinct `outcome_unknown` state.

The interactive admin **test-send** endpoint (`POST /api/admin/directory/dingtalk/work-notification/test`, `packages/core-backend/src/routes/admin-directory.ts`) did not: it caught every failure — including outcome-unknown ones — into the same generic `DINGTALK_WORK_NOTIFICATION_TEST_FAILED` (400) response. An admin verifying their Agent ID could see a plain "failed" for a test message that DingTalk actually delivered.

## Fix

In the test-send route's catch block, discriminate `isDingTalkOutcomeUnknown(error)` (imported from `../integrations/dingtalk/client`, same real function every other send-tier caller uses — not re-implemented) before falling through to the generic branch:

- **Outcome-unknown**: `502 DINGTALK_TEST_SEND_OUTCOME_UNKNOWN`, message = the underlying reason + explicit guidance that the message may still have been delivered and to check the device before retrying.
- **Everything else**: unchanged — `400 DINGTALK_WORK_NOTIFICATION_TEST_FAILED` with the original error message.

No ledger involved — this is the interactive test-send path, which by design has no ledger row to reconcile (kept that way per the lane brief).

### FE consumer checked

`apps/web/src/views/DirectoryManagementView.vue`'s `testWorkNotificationAgentId()` calls `readApiError(body, fallback)` and renders `error.message` directly via `setStatus(...)`. It does **not** switch on `error.code` for this route (unlike the directory-sync routes, which do use `readApiErrorCode(...) === 'DIRECTORY_SYNC_IN_PROGRESS'` for special UI). So no FE change was needed — the new message text alone conveys the distinct outcome. Noted as a candidate follow-up if product wants a visually distinct (non-error-red) treatment for this ambiguous state.

## Tests

Extended `packages/core-backend/tests/unit/admin-directory-routes.test.ts` (existing whitelisted `tests/unit` auto-collected file, no new wiring needed):

- `reports a distinct outcome-unknown code+message when the test send outcome is ambiguous` — mocks `testDingTalkWorkNotificationAgentId` rejecting with `Object.assign(new Error(...), { outcomeUnknown: true })` (the real property `isDingTalkOutcomeUnknown` checks — not a stand-in), asserts `502` + `DINGTALK_TEST_SEND_OUTCOME_UNKNOWN` + message contains both the original reason and the "may still have been delivered" / "check the test message on the device" guidance.
- `keeps the generic failure shape for a plain (non-outcome-unknown) test-send error` — pins the unchanged direction: `400` + `DINGTALK_WORK_NOTIFICATION_TEST_FAILED` + original message, unmodified.

`isDingTalkOutcomeUnknown` itself is NOT mocked in this test file (`dingtalk/client` is unmocked), so the route runs against the real discriminator.

Run results:
- `npx tsc --noEmit -p tsconfig.json` (packages/core-backend): clean
- `CI=true npx vitest run tests/unit/admin-directory-routes.test.ts`: 85/85 passed
- `CI=true npx vitest run tests/unit` (full core-backend unit suite): 368 files / 5064 tests passed

## Mutation table

| # | Mutation | Result |
|---|----------|--------|
| 1 | Delete the `isDingTalkOutcomeUnknown` branch entirely (fold back into generic failure) | RED — `reports a distinct outcome-unknown code+message...` fails (`expected 400 to be 502`); generic test stays green |
| 2 | Keep the branch but make it call the exact same `jsonError(res, 400, 'DINGTALK_WORK_NOTIFICATION_TEST_FAILED', ...)` as the generic path | RED — same test fails (`expected 400 to be 502`) |
| 3 | Keep code/status distinct but drop the "may still have been delivered / check the test message on the device" guidance from the message | RED — same test fails on the message-content assertion |

After each mutation the implementation file was restored via `git checkout --` to the committed state (verified via `git diff HEAD` empty before/after).

## Honest gaps

- No FE visual differentiation (e.g. warning vs. error styling) for the outcome-unknown case — deliberately out of scope per "if it renders the message directly, no FE change needed", but flagged above as a candidate follow-up.
- HTTP status `502` was chosen by analogy to existing upstream-provider-error usage in this codebase (`multitable-ai.ts`, `plm-workbench.ts`); there's no prior DingTalk-route precedent for this exact status pinning outcome-unknown specifically, so a reviewer should sanity-check that choice.
- Did not add integration/e2e coverage exercising the real transport → real DingTalk send path; the test mocks `testDingTalkWorkNotificationAgentId` at the settings-module boundary (consistent with the rest of this test file) and exercises the real `isDingTalkOutcomeUnknown` predicate against a hand-built error, not a live retry-exhausted transport call.

Base: `origin/main` (includes #4046 `b94b46bf3` and #4082 `be6f3fde1`). Not armed for auto-merge — landing is gated per the lane brief.

### DIFF
--- a/packages/core-backend/src/routes/admin-directory.ts
+++ b/packages/core-backend/src/routes/admin-directory.ts
@@ -27,6 +27,7 @@ import {
   updateDirectoryIntegration,
 } from '../directory/directory-sync'
 import { getDirectoryInactiveLinkedMetric, getDirectoryManagerBindingCoverage } from '../directory/directory-sync-alert-delivery'
+import { isDingTalkOutcomeUnknown } from '../integrations/dingtalk/client'
 import {
   getDingTalkWorkNotificationRuntimeStatusFromStore,
   saveDingTalkWorkNotificationAgentId,
@@ -190,6 +191,22 @@ export function adminDirectoryRouter(): Router {
       const result = await testDingTalkWorkNotificationAgentId(req.body as never)
       jsonOk(res, { result })
     } catch (error) {
+      // #4046 send-tier: network error/timeout/5xx/malformed-2xx on the actual send carries the
+      // transport's isDingTalkOutcomeUnknown marker — DingTalk may well have delivered the test
+      // message even though this request saw no (usable) response. Folding that into the generic
+      // "failed" code below would tell an admin their Agent ID is broken when the message may in
+      // fact have arrived. This is the interactive test-send path — there is no ledger row to mark
+      // outcome_unknown on (by design), so the ambiguity is surfaced directly to the caller instead.
+      if (isDingTalkOutcomeUnknown(error)) {
+        const reason = readErrorMessage(error, 'DingTalk did not confirm the outcome')
+        jsonError(
+          res,
+          502,
+          'DINGTALK_TEST_SEND_OUTCOME_UNKNOWN',
+          `${reason}. The test message may still have been delivered — check the test message on the device before retrying.`,
+        )
+        return
+      }
       jsonError(res, 400, 'DINGTALK_WORK_NOTIFICATION_TEST_FAILED', readErrorMessage(error, 'Failed to test DingTalk work notification Agent ID'))
     }
   })
--- a/packages/core-backend/tests/unit/admin-directory-routes.test.ts
+++ b/packages/core-backend/tests/unit/admin-directory-routes.test.ts
@@ -397,6 +397,57 @@ describe('adminDirectoryRouter', () => {
     })
   })
 
+  // #4046 P3-2: send-tier failures (network error/timeout/5xx/malformed-2xx) on the actual test
+  // send carry the transport's REAL isDingTalkOutcomeUnknown marker (imported unmocked from
+  // dingtalk/client — this is the same discriminator production code runs, not a stand-in). Before
+  // this fix both branches below folded into the same 400 DINGTALK_WORK_NOTIFICATION_TEST_FAILED
+  // code, telling an admin their Agent ID is broken for a message that may have actually arrived.
+  it('reports a distinct outcome-unknown code+message when the test send outcome is ambiguous', async () => {
+    const sendError = Object.assign(new Error('DingTalk request timed out after 10000ms'), {
+      outcomeUnknown: true,
+    })
+    workNotificationMocks.testDingTalkWorkNotificationAgentId.mockRejectedValue(sendError)
+
+    const payload = { integrationId: 'dir-1', agentId: '123456789', recipientUserId: 'user-1' }
+    const response = await invokeRoute('post', '/dingtalk/work-notification/test', {
+      body: payload,
+      user: { id: 'admin-1', role: 'admin' },
+    })
+
+    expect(response.statusCode).toBe(502)
+    expect(response.body).toMatchObject({
+      ok: false,
+      error: {
+        code: 'DINGTALK_TEST_SEND_OUTCOME_UNKNOWN',
+      },
+    })
+    const message = (response.body as { error: { message: string } }).error.message
+    expect(message).toContain('DingTalk request timed out after 10000ms')
+    expect(message.toLowerCase()).toContain('may still have been delivered')
+    expect(message.toLowerCase()).toContain('check the test message on the device')
+  })
+
+  it('keeps the generic failure shape for a plain (non-outcome-unknown) test-send error', async () => {
+    workNotificationMocks.testDingTalkWorkNotificationAgentId.mockRejectedValue(
+      new Error('DingTalk appSecret is required'),
+    )
+
+    const payload = { integrationId: 'dir-1', agentId: '123456789' }
+    const response = await invokeRoute('post', '/dingtalk/work-notification/test', {
+      body: payload,
+      user: { id: 'admin-1', role: 'admin' },
+    })
+
+    expect(response.statusCode).toBe(400)
+    expect(response.body).toMatchObject({
+      ok: false,
+      error: {
+        code: 'DINGTALK_WORK_NOTIFICATION_TEST_FAILED',
+        message: 'DingTalk appSecret is required',
+      },
+    })
+  })
+
   it('saves DingTalk work notification Agent ID and writes a redacted audit entry', async () => {
     workNotificationMocks.saveDingTalkWorkNotificationAgentId.mockResolvedValue({
       integration: { id: 'dir-1', name: 'DingTalk CN', status: 'active' },


zensgit:
# Adversarial review — PR #4085 `fix(dingtalk): distinct outcome-unknown response for work-notification test-send`

- **PR**: zensgit/metasheet2#4085, branch `claude/dt-testsend-outcome-unknown` → `main`
- **Head reviewed**: `f031467b7149aa252f3fc0386db573f4145e7c1a` (matches gate brief; verified via `refs/reviews/pr4085`)
- **Scope**: 2 files, +68/-0 — `packages/core-backend/src/routes/admin-directory.ts`, `packages/core-backend/tests/unit/admin-directory-routes.test.ts`
- **Method**: refute-first; disposable worktree `/tmp/ms2-rev-pr4085` at PR head, frozen install; real predicate traced to `transport.ts`; 4 live probes driving the REAL transport (mocked `fetchFn`) through the REAL route; 3 mutations re-fired against the PR's own tests; full unit suite + tsc; CI-gate + alerting-surface + FE-consumer verification.
- **Reviewer**: Claude adversarial gate, 2026-07-11. Review-only — no fixes pushed.

## Verdict: **APPROVE** (with one landing note: rerun the flaked required check before merge)

The change does exactly what it claims, the guard is load-bearing in both directions under mutation, the test exercises the real discriminator contract, 502 is a defensible honest shape that trips no in-repo alerting differently than the old 400, and scope is exactly as declared. No P1/P2.

---

## Probe 1 — Predicate fidelity (the implementer's own flagged gap)

**Real predicate** (`packages/core-backend/src/integrations/dingtalk/transport.ts:226-230`):

```ts
export function isDingTalkOutcomeUnknown(error: unknown): boolean {
  if (typeof error !== 'object' || error === null) return false
  if (outcomeUnknownErrors.has(error)) return true          // WeakSet (frozen/exotic errors)
  return (error as { outcomeUnknown?: unknown }).outcomeUnknown === true  // additive property
}
```

`markOutcomeUnknown` (transport.ts:232-239) attaches **both** channels: WeakSet always, plus `defineProperty(outcomeUnknown: true)` unless the error is frozen. The predicate accepts **either**.

- The PR test's `Object.assign(new Error(...), { outcomeUnknown: true })` therefore satisfies the real predicate via its documented **primary** branch — it passes for the right reason, not accidentally. `dingtalk/client` is **not** among the test file's `vi.mock` targets (verified: mocks cover rbac/audit/directory-sync/scheduler/alert-delivery/work-notification-settings/approval-card-config only), so the route runs the real discriminator.
- **Forged-property acceptance**: yes, any error carrying `outcomeUnknown: true` triggers the 502 path. Acceptable here and by design — the property is the predicate's documented contract channel; the only error producer on this path is internal (`testDingTalkWorkNotificationAgentId` → transport), not user input (`req.body` never lands properties on thrown errors), and the worst case is an over-cautious "may have been delivered" message on an interactive admin route. Judged fine; stated explicitly per the gate brief.
- **Real-transport drive (my probes, all green at head)**:
  - **Probe A** — symbol identity: `client.ts:14-21` re-exports the transport symbol; `expect(fromClient).toBe(fromTransport)` passes (same function object; route import at `admin-directory.ts:30` is the transport's predicate).
  - **Probe B** — real send-tier network error: `sendDingTalkWorkNotification` with `fetchFn` rejecting `TypeError('fetch failed')` → real marked error → route answers **502 DINGTALK_TEST_SEND_OUTCOME_UNKNOWN**.
  - **Probe C** — **frozen** error (WeakSet-only marking; `defineProperty` throws, property verifiably absent) → route still answers **502**. The WeakSet branch is live end-to-end.
  - **Probe D** — real **pre-send** failure (`fetchDingTalkAppAccessToken` network error, auth tier, transport does NOT mark) → route answers **400 DINGTALK_WORK_NOTIFICATION_TEST_FAILED**. Negative direction holds with a real transport error too.
- **Producer wrap check**: `work-notification-settings.ts:297-356` has no try/catch around the send — the marked error propagates unwrapped to the route. The branch is not dead code. Pre-send validation errors (`appKey/appSecret/agentId required`, integration-not-found) and token failures stay 400, correctly.

## Probe 2 — 502 vs 400: alarm surface + repo precedent

- **No in-repo pager distinguishes this 502 from the old 400**:
  - `ops/prometheus/phase5-recording-rules.yml:6` computes availability as `2xx / total` — a 400 and a 502 depress it identically; the status change is a **no-op** for that rule.
  - `ops/prometheus/phase5-alerts.yml:5` keys on `metasheet_http_requests_total{result="success"}` — that metric is **not emitted by any current code** (only a stale doc artifact in `claudedocs/`); the alert can't fire from this route.
  - `ops/prometheus/attendance-alerts.yml:5` keys on `attendance_api_errors_total` — attendance-only counter, untouched.
  - `AuditService.ts:332` (`>=500 → 'ERROR'` severity) belongs to a middleware that is **exported but never mounted** in `index.ts` — not on this path.
  - `APIGateway.ts:764` circuit-breaker 5xx trip only wraps gateway-registered endpoints; `adminDirectoryRouter` is mounted directly (`index.ts:1248`), bypassing it.
  - Remaining effect: `http_requests_total`/histogram gain a `status="502"` label for a **manual, one-shot admin click** — negligible rate, not #3903-style recurring background noise.
- **Precedent**: 502 for "upstream gave no usable answer" is established — `multitable-ai.ts:312` (`AI_PROVIDER_ERROR`, the cited precedent — real), `auth.ts:1296/1400` (DingTalk identity path), `plm-workbench.ts:1104/1197`, `ApprovalBridgeService.ts` (×5), `federation.ts:434`. Semantically, Bad Gateway is the *most honest* code for "DingTalk did not confirm": it is a server-side upstream condition, not a client error — a 400 would (and did) tell the admin *their input* is wrong. A 200-with-status-payload would misreport a non-success and break the FE's uniform `!response.ok` error path. **502 upheld; no downgrade.**
- **FE consumer** (`apps/web/src/views/DirectoryManagementView.vue:4234-4246`): status-agnostic — `if (!response.ok) throw new Error(readApiError(body, fallback))`, message rendered directly. No `error.code` switching on this route; nothing breaks. See P3-2 for the styling gap.

## Probe 3 — Both directions pinned (mutations re-fired at head, PR's own tests only)

| # | Mutation (route file) | Result |
|---|---|---|
| M1 | Delete the whole `isDingTalkOutcomeUnknown` branch | RED — PR's `reports a distinct outcome-unknown code+message...` fails (`expected 400 to be 502`); generic test stays green. (My probes B/C also red.) |
| M2 | Same-shape fold: keep the predicate branch, respond with the generic `jsonError(res, 400, 'DINGTALK_WORK_NOTIFICATION_TEST_FAILED', ...)` | RED — same PR test fails |
| M3 | Reverse pin: `if (true)` — every error takes the 502 distinct path | RED — PR's `keeps the generic failure shape...` fails (`expected 502 to be 400`). (My probe D also red.) |

All mutations restored via `git checkout --`; `git status --porcelain` empty at head before final runs. The PR body's own mutation table (its #1/#2/#3) is consistent with what I reproduced; its #3 (message-guidance drop) is additionally covered by the message-content assertions I confirmed present at `admin-directory-routes.test.ts:425-427`.

## Probe 4 — Scope & wiring

- Exactly 2 files (verified against full diff `origin/main...refs/reviews/pr4085`); no route/index/runtime wiring beyond the one import + catch-branch; no ledger writes (matches the "no ledger row by design" claim); no FE files; `package.json`/lockfile untouched.
- Import `isDingTalkOutcomeUnknown` from `../integrations/dingtalk/client` (`admin-directory.ts:30`) — verified **same symbol** as the transport export (Probe A, `toBe` identity), consistent with how every other send-tier caller imports it.
- CI gate reality: the touched test file runs in the **required** `test (18.x)`/`test (20.x)` jobs ("Run core-backend tests" step = `tests/unit`, observed executing in this PR's job logs). Not a phantom-gated test.
- PR body: no bare closing keywords (#4046 referenced as context only — no issue mis-close risk). Lineage: this is the interactive-route slice of the #4046 gate's P3-2 follow-up thread (marker consumption); ledger-writer consumption in the worker lanes remains the separately tracked remainder.

## Test / typecheck record (worktree `/tmp/ms2-rev-pr4085`, frozen install, head `f031467b7`)

- `vitest run` PR file + 2 neighbors (`admin-directory-routes` 85, `dingtalk-transport-resilience` 43, `attendance-dingtalk-outcome-unknown` 5): **133/133 pass**
- 4 review probes (added temporarily, then reverted): **4/4 pass**
- Full core-backend unit suite (`npm run test:unit`): **368 files / 5064 tests pass** (matches PR body's claim exactly)
- `npx tsc --noEmit`: **clean** (exit 0)

## Findings

### P3-1 — Cross-layer marker contract pinned per-layer, not end-to-end (hardening suggestion)
`tests/unit/admin-directory-routes.test.ts:404-406` forges the marker property at the settings-module boundary. If the transport ever stopped attaching the marker (or a wrapper landed between settings and transport), this route test would stay green while the 502 branch went dead. Today that drift is caught — `dingtalk-transport-resilience.test.ts:204/221/237` pins the marking in the **same required CI job**, and `work-notification-settings.ts:297-356` provably doesn't wrap — but the two pins live in different files with no single test crossing the seam. Cheap hardening: adopt my Probe-B/C shape (build the rejected error via the real `sendDingTalkWorkNotification` with a mocked `fetchFn`; the frozen-error variant also locks the WeakSet-only path through the route). The implementer's "honest gaps" already names this; graded P3, not P2, because both layers are independently pinned in required CI and I verified the live seam end-to-end this round.

### P3-2 — FE renders the ambiguous state as a generic red error with raw English text
`DirectoryManagementView.vue:4245-4246` funnels the 502 into `setStatus(error.message, 'error')`: an admin in the zh-CN UI sees error-red English ("... may still have been delivered — check the test message on the device before retrying."). The state is semantically a *warning*, and every neighboring string on that surface is Chinese. Not a regression (the old 400 path surfaced raw English messages identically), and the implementer flagged it as a follow-up candidate — but the user-facing payoff of this PR is materially blunted until the FE differentiates (warning styling keyed on `error.code === 'DINGTALK_TEST_SEND_OUTCOME_UNKNOWN'` + localized copy). Defer is acceptable; recommend a named follow-up so it doesn't evaporate.

### NIT-1 — Message wording repetition
`admin-directory.ts:206`: "The **test message** may still have been delivered — check the **test message** on the device before retrying." Second occurrence could be "check the device" / "check for it on the device".

### NIT-2 — Forged-property acceptance is by-design (documented judgment, no action)
Any internal error object carrying `outcomeUnknown: true` reaches the 502 path (predicate property branch). Acceptable on this interactive admin route: producers are internal, user input can't plant the property, and the failure mode is a benign over-cautious message.

## Landing note (not a code finding)

Required check **`test (20.x)` is red** on this PR — the failure is `tests/integration/multitable-oapi2a-ratelimit-realdb.test.ts:82` (`rows.some(outcome === 'denied' ...)` audit-row assertion in the OAPI-2a real-DB rate-limit hammer), inside the multitable real-DB step. Zero overlap with this diff (admin-directory route + unit tests only; `test (18.x)` green, my local full unit run green). The same workflow is green on main for the 8 most recent runs (latest at `11a2a29d8`). Assessment: run-specific flake (shared-DB timing on audit-row readback). **Rerun the job before merge** — the check is in the required set, and per house rules a red required check blocks landing regardless of attribution.

## Design-gate table

| Gate / claim | Verdict | Evidence |
|---|---|---|
| Distinct 502 + code only when send-tier marker present | PASS | route:200-209; probes B/C; M1/M2 red |
| Generic failures unchanged at 400 (regression pin) | PASS | test:430-449; probe D; M3 red |
| Real predicate, not a stand-in / not mocked | PASS | client unmocked in test file; Probe A symbol identity |
| No ledger involvement (lane brief) | PASS | diff has zero ledger code |
| Scope = 2 files, no extra wiring | PASS | full-diff sweep |
| FE needs no change to stay functional | PASS | DirectoryManagementView.vue:4245 status-agnostic (styling gap → P3-2) |
| 502 doesn't page (#3903 lesson) | PASS | recording rule 2xx-ratio invariant; dead `metasheet_http_requests_total`; AuditService middleware unmounted; APIGateway CB not on mount path |
| Claimed test/tsc runs reproducible | PASS | 85/85, 5064/5064, tsc clean — reproduced exactly |

— Worktree `/tmp/ms2-rev-pr4085` removed after review.

zensgit:
Window protocol executed (round-2 window 1): rebased onto current main (clean); load-bearing mutation replayed on the rebased head (outcome-unknown discrimination neutered ⇒ 502 test RED, restored); unrelated test(20.x) rate-limit flake rerun requested per gate's landing note. Armed alone.

PR #4086, opened 2026-07-11
test(onprem): require superseded audit marker in packages

zensgit:
## Scope after #4084

#4084 has merged the core migration-provider no-op for `20250926_create_audit_tables` into `main`. This PR is rebased and narrowed to the remaining package-assurance delta only:

- require the compiled attendance on-prem package to contain the superseded audit migration marker;
- require the multitable on-prem package to contain the same marker;
- extend both verifier contract suites so a package cannot silently omit that upgraded-database safeguard.

No migration-provider implementation or provider unit test is duplicated here.

## Verification

```text
attendancePackageVerifierTests=3/3 PASS
multitablePackageVerifierTests=10/10 PASS
packageVerifierTests=13/13 PASS
shellSyntax=PASS
head=70457d226
```

## Boundary

```text
migrationProviderFix=#4084 already on main
operatorLedgerEdit=false
operatorSchemaEdit=false
broadDuplicateSuppression=false
deploymentPerformed=false
releasePublished=false
externalPlmK3ErpWrite=false
```

#3751 remains open for a newly built corrective package and values-free entity-machine acceptance. This PR does not claim deployment, restart, health, or smoke completion.


zensgit:
Adversarial review (opus): **APPROVE**, no P1/P2. Verified against real code + real runs:
- Both verifiers query the compiled `dist/src/db/migration-provider.js` (correct target, matches package layout).
- Fail-closed `search_fixed_string … || die`; substring-false-green ruled out structurally — the literal `20250926_create_audit_tables` appears ONLY in `SUPERSEDED_LEGACY_SQL_MIGRATIONS` (all other names resolve dynamically), so the marker ships in `.js` **iff** the #4084 no-op is compiled in.
- Symmetric across attendance + multitable; negative teeth (real missing-marker `.tgz` fixture) + mutation M1/M2 (delete either check → test RED).
- Marker byte-exact across all 5 loci; no existing assertion moved. attendance 4/4, multitable 10/10, `bash -n` clean.
- Non-blocking: P3-1 (multitable test is source-grep-only; deletion still caught by M2, CI-gated per-PR) · NIT-1 (PR body verification block stale: actual head `da69b47c8`, attendance 4/4).

Per owner authorization to land #4086 in parallel before final accounting — arming.

PR #4087, opened 2026-07-11
fix(scheduler): far-future cron delay clamp + invalid-cron telemetry + stale-name closure (R1-L1)

zensgit:
## What / why

Three scheduler-hygiene fixes in the directory sync scheduler family, all catalogued as P3 findings by adversarial gates during the just-closed pool round:

- **FIX 1** (P3-3 from #4053's adversarial-gate comment) — `JobScheduler.scheduleCronJob` (`packages/core-backend/src/services/SchedulerService.ts`) passed the raw `delayMs` to `setTimeout`. Node silently clamps delays above `2^31-1` ms (~24.86 days) to 1ms, so a validated far-future cron (e.g. a yearly schedule saved mid-year, ~175 days out) would fire immediately and hot-loop, gated only by whatever the handler itself does (for directory sync, the sync lease).
- **FIX 2** (P3-1 from #4053's adversarial-gate comment) — Directory integrations saved before #4053's save-time cron validation landed may hold a `schedule_cron` the runtime can never fire. `applySchedule` already unschedules/skips these safely, but the failure was only a per-row `logger.warn`. Boot now also logs ONE structured summary of every such row so ops can find and fix them.
- **FIX 3** (P3-2 from #4057's adversarial-gate comment) — The schedule callback in `directory-sync-scheduler.ts` closed over the mutable `row.name` captured at job-creation time. `reschedule()` (the `existingJob` branch of `applySchedule`) keeps the same job/handler and only swaps the cron expression, so a renamed integration alerted/logged under its stale name until the job was unscheduled and recreated (or the process restarted).

(#4053 = "validate schedule_cron at save time"; #4057 = "§7.1 inactive-linked-for-N-days metric + post-run alert" — both merged into main earlier in this pool round.)

## Changes

- `packages/core-backend/src/services/SchedulerService.ts` — new `armCronTimeout` helper chunks any delay above a safe ceiling (`2^31-1-1_000_000` ms, headroom under Node's clamp point) into repeated max-size waits, recomputing the remaining delay against the fixed target time at each re-arm. Only once the remaining delay fits under the ceiling does the real fire-the-job timeout get set. `removeJob`/`destroy` always cancel whatever timeout is currently in `cronJobs` — since every re-arm updates that map entry, unscheduling mid-chain still kills the job before it fires. Zero behavior change for delays under the ceiling (still a single `setTimeout` call, same computed delay).
- `packages/core-backend/src/directory/directory-sync-scheduler.ts` —
  - `applySchedule` now accepts an optional `onInvalidCron` hook, invoked from its existing catch block (unchanged failure semantics — still unschedules/skips). `startDirectorySyncScheduler` collects every invalid-cron row hit during the boot sweep and, if any, logs one structured `logger.warn` with `{ invalidCronIntegrations: [{id, name, cron, error}, ...] }`. Never mutates/clears the offending rows.
  - New `resolveScheduledIntegrationName(integrationId)` re-reads the row via the existing `getScheduleRow` helper at fire time and falls back to the id on lookup failure — mirroring the existing `readIntegrationNameForAlert` convention in `directory-sync.ts`. The schedule-creation callback now carries only `row.id` and resolves the display name at fire time instead of closing over `row.name`.
- Tests: new `packages/core-backend/tests/unit/scheduler-service.test.ts` (FIX 1); extended `packages/core-backend/tests/unit/directory-sync-scheduler.test.ts` (FIX 2, FIX 3, plus one extra mocked `query()` response in the pre-existing "registers a job..." test to account for FIX 3's new fire-time lookup).

## Design decision flagged for review (FIX 3)

Brief asked to pick the smaller diff between (a) resolving the name at fire time vs (b) recreating the job on every reschedule, and justify. Chose (a): it doesn't touch `reschedule()`'s in-place semantics, so job `runCount`/`lastRun` survive a rename exactly like they survive any other cron-only edit. (b) would reset those stats on every rename since `schedule()` always constructs a fresh `ScheduledJob`. Cost of (a): one extra `SELECT` on `directory_integrations` per triggered sync (not per scheduler tick) — negligible relative to the DingTalk API sync work that follows, but flagging as a genuine behavior change for ops. On DB failure at fire time, the alert degrades to showing the integration id instead of the (possibly now-stale) name — strictly safer than the prior "always stale, never expires" behavior, but different from status quo.

## Test evidence

- `packages/core-backend/tests/unit/scheduler-service.test.ts` (new, 4 tests) — pass
- `packages/core-backend/tests/unit/directory-sync-scheduler.test.ts` (extended, 8 tests, was 5) — pass
- Scheduler-family suites run explicitly by name (`scheduler-service`, `directory-sync-scheduler`, `attendance-scheduler`, `ledger-retention-scheduler`, `webhook-retry-scheduler`, `approval-sla-scheduler`, `automation-scheduler-date-reminder`, `automation-scheduler-leader`, `automation-scheduler-metrics`): **9 files / 84 tests pass**
- Full `tests/unit`: **369 files / 5069 tests pass**
- `tsc --noEmit -p tsconfig.json` (core-backend): clean

## Mutation table

| # | Fix | Mutation | Result |
|---|---|---|---|
| M1 | FIX 1 | Remove the `armCronTimeout` chunking guard (`if (remainingMs > MAX_SAFE_TIMEOUT_MS)` branch) — always pass the raw remaining delay to `setTimeout` | **RED** — the armed-delay invariant test fails cleanly (`expected 15120000000 to be less than or equal to 2147483647`). This is the load-bearing detector for the clamp property: per the brief's own note, fake timers don't apply Node's real clamp, so a delay-based advance-and-assert test can't distinguish chunked-vs-raw by itself — only inspecting the actual armed `setTimeout` delay can. The other 3 tests in the file also failed under this mutation (via 30s timeouts from fake-timer/setTimeout interaction instability with an unchunked ~1.5e10ms delay) — treat those as corroborating noise, not as clean falsification; the invariant test is the one to trust here. |
| M2 | FIX 1 | Drop `this.cronJobs.set(jobName, { expression, timeout })` from the chunk-wait branch (leaves `removeJob` unable to find/cancel the pending chunk timer) | **RED**, cleanly — exactly the "kills a far-future cron unscheduled mid-chain" test fails (`expected "spy" to not be called at all, but actually been called 1 times`) |
| M3 | FIX 2 | Delete the `if (invalidCronRows.length > 0) { logger.warn(...) }` boot summary block entirely | **RED**, cleanly — exactly the "logs one structured summary..." test fails (`expected [] to have a length of 1 but got +0`); the "does not log a summary when all rows are valid" test still passes (as expected — nothing to falsify there) |
| M4 | FIX 3 | Revert the schedule callback to `runScheduledSync(row.id, row.name)` (the pre-fix stale closure) | **RED**, cleanly — exactly the "resolves the CURRENT integration name at fire time..." test fails (expected `'DingTalk CN Renamed'`, got `'DingTalk CN'`) |

All four mutations applied to a committed real implementation, one at a time, then reverted via `git checkout --` after recording red/green; working tree confirmed clean (`git status --short` / `git diff --stat` empty) and full suite re-verified green before opening this PR.

## Honest gaps / not proven

- FIX 1's early-fire sanity test ("does not fire a far-future cron early, and fires once the full delay elapses") is a behavior pin, not a mutation-falsifiable guard under fake timers (per the brief's own caveat — Node's real `setTimeout` clamp isn't simulated by `vi.useFakeTimers()`). It documents intended behavior and catches gross regressions in the re-arm loop, but the armed-delay invariant test (M1's detector) is the actual falsifiable proof for the clamp property.
- FIX 2's test triggers the catch path via a **mocked scheduler rejection** (`scheduler.schedule` rejects with the real `SimpleCronExpression` error string), not by driving a genuinely-invalid cron string through the real `SchedulerServiceImpl` — consistent with how this test file already mocks the injected scheduler throughout (`createSchedulerMock()`), but noting it as a boundary: this proves the collect-and-summarize wiring, not that `SimpleCronExpression` itself classifies any particular cron as invalid (that's covered elsewhere, in #4053's route-validation tests).
- Not verified against a live DingTalk sync run or production log volume — this is a logging/timer-mechanics change with no route/UI surface, verified at the unit level only, consistent with the rest of this scheduler family's test coverage.
- Landing-coordination check: `git log origin/main --oneline` shows no drift on the three touched files since branch point; searched all currently-open PRs' diffs for `SchedulerService.ts` / `directory-sync-scheduler.ts` — none touch either file. No known collision, but flagging per house convention since this file family has had cross-PR collisions before (#4057 vs #4054).

## Behavior changes for ops

- Boot now emits one additional structured `warn` log line when any directory integration has an unschedulable `schedule_cron` (read-only; no row mutation).
- Every scheduler-triggered sync now issues one extra `SELECT` against `directory_integrations` (to resolve the current display name) before the §7.1 offboarding-digest alert fires — negligible relative to the sync itself.

### DIFF
--- a/packages/core-backend/src/directory/directory-sync-scheduler.ts
+++ b/packages/core-backend/src/directory/directory-sync-scheduler.ts
@@ -53,6 +53,23 @@ async function getScheduleRow(integrationId: string): Promise<DirectoryScheduleR
   return result.rows[0] ?? null
 }
 
+// P3-2 (#4057's gate): the schedule callback used to close over the mutable `row.name`
+// captured at job-CREATION time. reschedule() (the `existingJob` branch of applySchedule)
+// keeps the same job/handler and only swaps the cron expression, so a renamed integration
+// kept alerting/logging under its old display name until the job was unscheduled and
+// recreated (or the process restarted). Resolving the name at fire time — carrying only
+// the immutable `row.id` in the closure — fixes that without changing reschedule-in-place
+// semantics (job stats/runCount survive a rename, matching every other cron-only edit).
+// Mirrors the existing `readIntegrationNameForAlert` fallback convention in directory-sync.ts.
+async function resolveScheduledIntegrationName(integrationId: string): Promise<string> {
+  try {
+    const currentRow = await getScheduleRow(integrationId)
+    return currentRow?.name || integrationId
+  } catch {
+    return integrationId
+  }
+}
+
 async function runScheduledSync(integrationId: string, integrationName: string): Promise<void> {
   try {
     await syncDirectoryIntegration(integrationId, SYSTEM_TRIGGERED_BY, 'scheduler')
@@ -75,7 +92,18 @@ async function runScheduledSync(integrationId: string, integrationName: string):
   await deliverDirectoryInactiveLinkedAlert({ integrationId, integrationName })
 }
 
-async function applySchedule(row: DirectoryScheduleRow | null): Promise<void> {
+type InvalidCronReport = { id: string; name: string; cron: string; error: string }
+
+type ApplyScheduleOptions = {
+  // §7.8 follow-up (P3-1 from #4053's gate): rows saved before save-time cron validation
+  // landed may hold a schedule_cron the runtime can never fire. applySchedule already
+  // unschedules/skips them safely; this hook lets the boot sweep additionally collect
+  // them for one aggregate telemetry line so ops can find and fix the rows. Never used
+  // to mutate/auto-clear anything.
+  onInvalidCron?: (report: InvalidCronReport) => void
+}
+
+async function applySchedule(row: DirectoryScheduleRow | null, options: ApplyScheduleOptions = {}): Promise<void> {
   if (!scheduler) return
 
   const jobName = buildJobName(normalizeText(row?.id))
@@ -96,14 +124,20 @@ async function applySchedule(row: DirectoryScheduleRow | null): Promise<void> {
       logger.info(`Rescheduled directory sync job: ${jobName} (${cronExpression})`)
     } else {
       await scheduler.schedule(jobName, cronExpression, async () => {
-        await runScheduledSync(row.id, row.name)
+        await runScheduledSync(row.id, await resolveScheduledIntegrationName(row.id))
       }, {
         timezone: 'UTC',
       })
       logger.info(`Scheduled directory sync job: ${jobName} (${cronExpression})`)
     }
   } catch (error) {
-    logger.warn(`Failed to apply directory sync schedule for ${row.id}: ${error instanceof Error ? error.message : String(error)}`)
+    const message = error instanceof Error ? error.message : String(error)
+    logger.warn(`Failed to apply directory sync schedule for ${row.id}: ${message}`, {
+      integrationId: row.id,
+      integrationName: row.name,
+      scheduleCron: cronExpression,
+    })
+    options.onInvalidCron?.({ id: row.id, name: row.name, cron: cronExpression, error: message })
     if (existingJob) {
       await scheduler.unschedule(jobName)
     }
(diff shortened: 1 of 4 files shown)


zensgit:
CODEX_REVIEW
lane_id: MS-DINGTALK-SCHEDULER-HYGIENE
reviewed_head_sha: 4d96e7c7e7bc974b895ea845a3cd3b580b6cdb86
scope_checked: far-future cron timer chunking; cancel/reschedule behavior; boot invalid-cron telemetry; fire-time integration-name resolution
tests_and_ci:
- focused scheduler suites: 2 files / 12 tests PASS
- diff check: PASS
- current-main merge simulation: no conflicts and no main-side changes in the four touched files
verdict: RATIFIED@SHA
next_authorized_action: owner review/merge decision. GitHub reports MERGEABLE/BEHIND; if the head changes, this ratification no longer applies and the new SHA requires review.

Notes:
- The boot aggregate hook is active only while constructing a fresh scheduler and therefore reflects schedule-time cron validation failures; runtime refreshes do not feed that aggregate.

PR #4088, opened 2026-07-11
test(directory): syncDirectoryIntegration orchestration harness — heartbeat lifecycle, H02 admission wiring, OPS-01 deprovision wiring, reclaim composition (real DB)

zensgit:
## What

R1-L4 (remainder round, test-infra): the `syncDirectoryIntegration` **orchestration harness** — the repo-wide systemic gap named in `docs/development/dingtalk-sync-hardening-design-and-verification-20260708.md` §5.2. Until now no test drove the sync orchestration end-to-end, so a regression at a **call site** (H02 admission wiring, OPS-01 deprovision executor wiring, heartbeat lifecycle, reclaim composition) turned no test red even though every helper was individually proven.

One new whitelisted real-DB suite, `packages/core-backend/tests/integration/directory-sync-orchestration.db.test.ts` (12 tests, ~14s), building on the R5 fixture pattern from `directory-sync-alert-coverage.db.test.ts` (#4054): mocked DingTalk client, REAL `syncDirectoryIntegration`, real Postgres apply on a fresh migrated DB.

**Test-only change.** No production source is touched. The heartbeat is timed by pinning `DIRECTORY_SYNC_HEARTBEAT_INTERVAL_MS` to its existing 5000ms env floor via `vi.hoisted` before module import — no DI seam was added.

## Coverage

1. **Heartbeat lifecycle (DT-HARDEN-05 composition)** — a controllable gate holds the mocked pull open: `last_heartbeat_at` advances at least once mid-run; a concurrent second trigger gets `DirectorySyncInProgressError` (409) carrying the active `runId` without disturbing the live run's row; after completion the interval is proven **cleared** via a flipped-row probe (the beat's own `status='running'` predicate would mask a leaked interval from any passive observation — this was the previously-unproven claim).
2. **Admission wiring (H02 call site)** — one real run carries three accounts through the real loop: eligible account → user created + linked `auto_admit` + identity bound (`corp:openId`) + grant enabled, all in the SAME run; corp account without openId → admitted with the grant **computed off** at the call site (the exact line the original bug hardcoded); poisoned account (no identity key, bind throws AFTER the users INSERT) → per-account SAVEPOINT rolls the user back — **no orphan row**, account `unmatched`, run still `completed`, stats split honest (2 admitted / 1 failed / 1 without-grant).
3. **Deprovision executor wiring (OPS-01 call site)** — integration-row `default_deprovision_policy='mark_inactive'` + env flag: a departed linked account's user is deactivated and its grant revoked **inside the run**, fed by the per-run transition sweep, `deprovisionAffected` names the identity; default-off arm pins that **nothing is written** while candidate/would-deactivate counts stay preview-honest.
4. **Reclaim + completion-guard composition** — a seeded 2h-stale `running` zombie is reclaimed by the next real trigger, which claims its own run and completes; the zombie stays `failed` (`orphaned…` message) and no `running` row survives; the scheduler **boot sweep** (`startDirectorySyncScheduler` with a stub scheduler) independently reclaims a stale zombie.

## Two-point wiring (§9)

- `packages/core-backend/vitest.config.ts` — added to `exclude` so the no-DB job cannot skip-green it.
- `.github/workflows/plugin-tests.yml` — added as a WHOLE FILE to the `Run approval real-DB integration` step (alongside the sibling directory-sync goldens).

## Test evidence (local, fresh createdb + full migrate, PG 5435)

- New suite: **12/12 passed**, 13.8–14.0s.
- `tsc --noEmit` (core-backend): clean.
- Full `tests/unit`: **368 files / 5062 tests passed**.

## Mutation table (each applied to committed source, run, restored)

| # | Mutation (source, then reverted) | Result |
|---|---|---|
| M1 | Delete the heartbeat `setInterval` body (the `UPDATE … last_heartbeat_at` call) | RED — `beats last_heartbeat_at at least once…` (11/12 pass) |
| M1b | Remove `clearInterval(heartbeat)` from the sync's `finally` | RED — `…heartbeat interval is genuinely cleared (no zombie beats)` |
| M2a | Revert H02 call site to `enableDingTalkGrant: true` (the original bug) | RED — `admits a corp account without openId with the grant computed OFF` + stats split test (2 red) |
| M2b | Strip the per-account SAVEPOINT wrapper (SAVEPOINT/ROLLBACK/RELEASE) from `createDirectoryAdmittedUserInTransaction` | RED — `rolls a poisoned account back per-account (SAVEPOINT): no orphan users row…` (the committed orphan reappears) |
| M3 | Skip the `applyDirectoryDeprovisionPolicies` call at the sync call site (substitute inert outcome) | RED — both deprovision arms (2 red) |
| M4a | Drop `reclaimStaleDirectorySyncRuns(integrationId)` from the trigger path | RED — trigger-path reclaim test (`DirectorySyncInProgressError` instead of completing); boot-sweep test also red by cascade (unreclaimed zombie blocks the partial unique index) |
| M4b | Drop `reclaimStaleDirectorySyncRuns()` from `startDirectorySyncScheduler` | RED — boot-sweep test only (`expected 'running' to be 'failed'`) |

## Not proven / honest gaps

- **Lease-lost mid-apply race** (`DirectorySyncLeaseLostError`: a reclaim stealing the lease WHILE a live run is inside its apply transaction, completion guard `AND status='running'` aborting the zombie's commit): not driven end-to-end. Engineering that race needs a second controllable gate between claim and completion **inside the transaction**, i.e. a DI seam in production code, which this test-only lane deliberately avoids. Mutating the completion guard would NOT redden this suite; it stays pinned at unit level (`tests/unit/directory-sync-run-lease.test.ts` pins the SQL text and the `DirectorySyncLeaseLostError` rejection).
- **"No further beats" is probed, not exhaustively proven**: the flipped-row probe watches 1.5 heartbeat periods of silence after completion. A leaked timer firing slower than the configured interval would escape it (cannot happen with `setInterval` unless the interval constant itself is corrupted, which M1 covers separately).
- **Failure path lease release** (`markSyncFailure` flips the run to `failed` and releases the lease): not asserted here; covered by the existing lease golden (completion/failure → re-claim OK) and unit pins.
- **HTTP surface**: the 409 route mapping of `DirectorySyncInProgressError` and the async-202 hook are not driven (service boundary only, matching the R5 precedent).
- **Heartbeat env normalization** (sub-5s values fall back to 60s etc.) is unit-level territory; this suite relies on the documented 5000ms floor and would fail loudly (beat never observed in 15s) if that floor changed upward.

## Ops notes

No behavior change. The suite sets `DIRECTORY_DEPROVISION_ENABLED` only inside one test's try/finally and unsets it after; the boot-sweep test performs the production-shaped UNSCOPED stale sweep, which is safe in the serial real-DB CI lane (no sibling suite holds a live running fixture concurrently; >10min-stale leftovers are exactly what the sweep exists to close).

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -424,6 +424,7 @@ jobs:
             tests/integration/directory-sync-alert-coverage.db.test.ts \
             tests/integration/directory-deprovision-selection.db.test.ts \
             tests/integration/directory-sync-run-lease.db.test.ts \
+            tests/integration/directory-sync-orchestration.db.test.ts \
             tests/integration/attendance-csv-export-bom.test.ts \
             --reporter=dot
 
--- a/packages/core-backend/tests/integration/directory-sync-orchestration.db.test.ts
+++ b/packages/core-backend/tests/integration/directory-sync-orchestration.db.test.ts
@@ -1,0 +1,712 @@
+import { afterAll, beforeAll, describe, expect, it, vi } from 'vitest'
+
+// R1-L4 — the syncDirectoryIntegration ORCHESTRATION HARNESS.
+//
+// dingtalk-sync-hardening-design-and-verification-20260708.md §5.2 names the repo-wide
+// systemic gap this file closes: no test drove the sync orchestration end-to-end, so a
+// regression at a CALL SITE (DT-HARDEN-02 admission wiring, DT-OPS-01 deprovision executor
+// wiring, DT-HARDEN-05 heartbeat lifecycle / reclaim composition) turned no test red even
+// though each helper was individually proven. This suite follows the R5 fixture pattern in
+// directory-sync-alert-coverage.db.test.ts (#4054): the DingTalk HTTP client is mocked so
+// the REAL `syncDirectoryIntegration` pulls a synthetic directory and applies it against
+// real Postgres — nothing between the test and the orchestration except the network client.
+//
+// Heartbeat timing: DIRECTORY_SYNC_HEARTBEAT_INTERVAL_MS is a module-load constant with a
+// hard 5000ms floor (sub-5s values fall back to 60s by design), so it is pinned to the
+// minimum via vi.hoisted BEFORE the module import — no production seam was added for this.
+// The lease-stale floor stays max(10min, 5x beats) = 10min, so live runs here (beating
+// every 5s) are never reclaimed while seeded 2h-old zombies always are.
+const hoistedEnv = vi.hoisted(() => {
+  process.env.DIRECTORY_SYNC_HEARTBEAT_INTERVAL_MS = '5000'
+  return { heartbeatMs: 5000 }
+})
+
+const clientMocks = vi.hoisted(() => ({
+  fetchDingTalkAppAccessToken: vi.fn(),
+  listDingTalkDepartments: vi.fn(),
+  getDingTalkDepartmentDetail: vi.fn(),
+  listDingTalkDepartmentUsers: vi.fn(),
+  getDingTalkUserDetail: vi.fn(),
+}))
+
+vi.mock('../../src/integrations/dingtalk/client', () => ({
+  fetchDingTalkAppAccessToken: clientMocks.fetchDingTalkAppAccessToken,
+  listDingTalkDepartments: clientMocks.listDingTalkDepartments,
+  getDingTalkDepartmentDetail: clientMocks.getDingTalkDepartmentDetail,
+  listDingTalkDepartmentUsers: clientMocks.listDingTalkDepartmentUsers,
+  getDingTalkUserDetail: clientMocks.getDingTalkUserDetail,
+}))
+
+import { query } from '../../src/db/pg'
+import {
+  createDirectoryIntegration,
+  DirectorySyncInProgressError,
+  syncDirectoryIntegration,
+} from '../../src/directory/directory-sync'
+import {
+  resetDirectorySyncSchedulerForTests,
+  startDirectorySyncScheduler,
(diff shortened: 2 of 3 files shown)


zensgit:
CODEX_REVIEW
lane_id: MS-DINGTALK-SYNC-ORCHESTRATION
reviewed_head_sha: afce5f120089337b095aa478a71c03bce4cdbf6c
scope_checked: real-DB orchestration harness; heartbeat lifecycle and concurrent-trigger rejection; H02 admission call-site wiring; OPS-01 deprovision call-site wiring; trigger/boot stale-lease reclaim; two-point CI wiring

tests_and_ci:
- focused real-DB suite on a migrated local PostgreSQL database: 1 file / 12 tests PASS
- diff check: PASS
- current-main merge simulation: clean; the merged tree retains both this PR's directory-sync entries and main's newer multitable-attachment entries in plugin-tests.yml and vitest.config.ts

verdict: RATIFIED@SHA
next_authorized_action: owner review/merge decision. GitHub reports MERGEABLE/BEHIND; if the head changes, this ratification no longer applies and the new SHA requires review.

Notes:
- The first local run failed only because the pre-existing local `metasheet_test` database lacked the `last_heartbeat_at` migration; after applying the branch's migrations, the exact suite passed 12/12. The fresh-database GitHub lane was already green.
- Non-blocking test-hygiene debt: the suite directly overwrites `DIRECTORY_SYNC_HEARTBEAT_INTERVAL_MS`, and its enabled deprovision arm deletes rather than restores any prior `DIRECTORY_DEPROVISION_ENABLED` value. The current CI invocation is serial and the following file is unrelated, so this does not contaminate the present gate; preserve/restore prior values if this suite is later rewired or reordered.

PR #4093, opened 2026-07-11
feat(stock-preparation): add readonly source intake feeders

zensgit:
## Summary

Approved readonly source configs feed the existing pure stock-preparation intake contract, behind admin-gated dry-run endpoints whose HTTP projection is values-free shape evidence only.

- feed mapped PLM project/BOM and ERP/K3 material rows through the existing pure readonly-intake contract;
- expose admin-gated PLM and ERP material source-run endpoints (dry-run; no internal or external write);
- support PLM wrapper, K3 WebAPI / SQL Server, configured SQL readonly, and Bridge Agent readonly source kinds;
- **completeness is proven or the run fails closed** — see below.

## Completeness: proven, never assumed

A partial snapshot reported as `ready` is the worst failure this module can have: the rows are real, the counts look clean, and nothing downstream can tell. A run reaches `ready` only when completeness is **proven**, one of exactly two ways:

| proof | meaning |
|---|---|
| `short_page` | the adapter returned FEWER rows than the page bound **it applied** — it had nothing more to give |
| `declared_total` | the rows we collected match the row count the **source itself** declared |

Anything else fails closed. Three things are never taken on trust:

- **an adapter's `done: true`** — the Bridge Agent and the K3 SQL-Server executor return it unconditionally, including on a page they just clamped;
- **the page bound** — "we asked for 1000, so it must have been 1000" is how a clamped 20-row page passed as a complete 5,000-row source. Each kind declares how the *applied* bound is learned (`adapter_reported` for the clamping bridge, else `honours_request`, asserted against its adapter source). A clamping adapter that reports nothing is unprovable, not assumed;
- **that paging happened at all** — a source must echo back the page it served, and no page may repeat one already served. A K3 endpoint that declares `ROWCOUNT: 100` but ignores `PageIndex` served page 1 ten times; the duplicate rows added up to the declared total and satisfied `declared_total`, the strongest proof we have — 100 "provably complete" rows holding 10 real materials.

## One plane, one contract

The feeder reads the adapter's **record plane**, while the configured-read route defaults to the raw plane. A fieldMap that resolves perfectly where an operator checks it could therefore resolve *nowhere* where it runs — and the projection is per-field fail-soft, so the result was a 2,500-line BOM snapshot with `designQty` and `designUnit` null on **every** row, reported `ready` with `rowErrors: 0`.

Three rules now make a silently-null column impossible:

- **a configured field that resolves on NOT ONE row fails the run** (`SOURCE_RUN_FIELD_MAP_UNRESOLVED`, 422 — values-free: target names and counts, never the source path, which names the external system's schema). A field that resolves on *some* rows is a sparse column and still passes; an explicitly-null source value is *resolved* (present-but-empty is a faithful representation).
- **a fieldMap may not write the same target twice** (`duplicate_target`, rejected by the config validator). The mapping is a sequence of writes, so two entries on one column are not "try both spellings, whichever resolves" — the last one wins, and an entry that resolves nowhere writes null OVER a real value the earlier one read. That is data loss, and it is order-dependent.
- **the projection never erases a value it already read**, and its resolution tally is derived from the row actually emitted — so the counter cannot drift from the value written, whatever the entry order.

**On plane parity, stated exactly:** `POST /read-source-configs/:id/read` accepts a `rowSource` (closed enum), so an operator *can* preview the exact plane the feeder ingests — including for `data-source:sql-readonly`, which emits no raw payload at all and so had **no** working authoring surface before (the raw plane can only ever answer `RESPONSE_UNRECOGNIZED` for it). It is **opt-in, and the default is unchanged** — deliberately: that route serves every mode, and the record plane cannot express `detail_with_lines` (header/lines) or `resolver_lookup` at all, so it cannot be the default without breaking those configs outright; flipping it would also silently change the mapped values that already-approved configs return on a shipped runtime route. Parity is therefore available, not automatic — and the feeder's own guard is what makes a raw-authored config fail loudly instead of ingesting nulls.

## The rows come from the adapter, not from its raw payload

The feeder reads the adapter's **record plane** — the rows it normalized, flattened and paged — not the unprocessed upstream payload. Walking the raw payload silently bypasses that entire layer:

- the PLM wrapper flattens a BOM **tree** and pages the flattened lines; `raw` stays the tree ROOT, so a real 2,500-line BOM presented as **three rows**, `status: ready`, `rowErrors: 0`;
- `data-source:sql-readonly` never sets a raw payload at all — its rows exist only in `records`.

Backed by a cross-check: the adapter's records array and the row count its metadata *claims* must agree, or the run fails closed. (There is no third comparison against the mapped row count — on the record plane those are equal by construction, and a check that cannot fail is not a defence.) A PLM BOM config addresses the wrapper's normalized fields; the untouched upstream row stays reachable under `rawPayload.*`.

## Real row envelopes per source kind

Measured end-to-end against the shipped adapters. **A source above its envelope fails closed — never silently truncated.**

| kind | continuation | provable rows |
|---|---|---|
| `plm:yuantus-wrapper`, client that pages | offset cursor | **up to 9,999** (10 × 1000); BOM **trees** included |
| `plm:yuantus-wrapper`, client that does NOT page | — | **up to 999**. At exactly 1000 the source replays its page → `PAGE_NOT_ADVANCING` after **2** reads (it used to be read **ten** times before failing); above 1000 → `RESULT_TOO_LARGE` on the first read |
| `data-source:sql-readonly` | offset cursor on every full page | up to 9,999 |
| `erp:k3-wise-webapi` (LIST) | bounded `listPageIndex` 1..10, 10 rows/page | **up to 100** — and only when K3 both echoes `Data.PAGEINDEX` (matching the page we asked for) and declares `ROWCOUNT`. An endpoint that pages correctly but does not echo the page index, or echoes `PAGEINDEX: 0`, is **unusable through this feeder** and fails closed on the first read: paging that cannot be proven is not paging |
| `bridge:legacy-sql-readonly` | **none** — the query body has no cursor field | **`config.maxLimit - 1`** (default **19**, at most 499). At the clamp it cannot prove there is nothing more, so it fails closed |
| `erp:k3-wise-sqlserver` | **none** — always `nextCursor: null, done: true` | one page (1000) |

## Safety boundary

```text
mode=dry_run
internalWriteExecuted=false
externalWriteExecuted=false
productionWrite=false
k3SaveSubmitAudit=false
plmExternalWrite=false
rawSql=false
sourcePayloadRowsInEvidence=false
privateConfigIdInEvidence=false
credentialsInEvidence=false
autoApply=false
```

Read-only against external systems; **no** external write / apply-writer / K3 Save-Submit-Audit / raw SQL. Approved-config-only (through the store's `getForRuntime`), tenant/workspace scoped, admin-gated (403 before any IO). `inputs` is a closed sub-allowlist — only the preset-declared key, never a filter, path or page control. The normalized intake rows stay an internal backend data plane: the HTTP projector never returns them, raw adapter payloads, config/system ids, credentials, or cursors.

## Verification

- full `plugins/plugin-integration-core` chain: **PASS** (EXIT=0)
- both silent-truncation defects were reproduced end-to-end against the **real** adapters before the fix and re-run after it: the 2,500-line BOM tree (3 rows `ready` → 2,500 rows `ready`), and the K3 endpoint that ignores `PageIndex` (100 "complete" rows from 10 materials → fails closed on page 1).
- guards are mutation-verified — delete the guard, a test goes red (each mutation checksum-verified as actually written to disk): row plane, row-count cross-check, page echo, page-advance, effective-page-size contract, overflow, project scope, mode allowlist, trusted-execution allowlist, route `inputs` sub-allowlist, route kind mismatch, admin gate on both routes, cursor loop, page budget, `done:true` distrust, values-free projection, intake rowErrors.

## Known limitations

- **the `duplicate_target` rule is retroactive.** The execution path re-validates a stored config (`prepareConfiguredRead` → `assertS1NormalizedConfig` → `validateReadSourceConfig`), so **any already-approved config whose fieldMap writes the same target twice now fails with a 400 (`config_invalid`)** — on the shipped `/read-source-configs/:id/read` route and in composition, not just in this feeder. That is deliberate and fail-closed: such a config was *already* losing data (the last entry wins, and an entry that resolves nowhere wrote null over a real value), so it fails loudly instead of quietly. A sweep of the repo's templates/presets finds no such config, but an operator-authored one would break on its next run and must drop the duplicate entry. (I argue below against flipping the `rowSource` default because it would silently change what approved configs return; the same pen owes you this: *this* rule changes what a narrow class of approved configs does — loudly, by design.)

- an approved config is authored against **one** plane; a config written for the raw plane and run by the feeder is now rejected by name rather than silently mapped to nulls, but it must be re-authored against the record plane (`rowSource: 'adapter_records'` on the read route shows exactly what it will ingest).
- three checks are **structurally unreachable** from the feeder and are documented as such rather than counted as coverage: the row-plane integer check (the executor already fails closed when the adapter returns no row array); the record-plane mode fence in the runtime (the feeder refuses those modes first); and the "page has no verifiable identity" check (the fingerprint is always produced). The projection's no-overwrite rule is kept as a deliberate second line behind the validator — it *is* the write semantics, and it is unit-tested directly.
- the field guard's prototype safety is **deliberately redundant**: the resolution tally has a null prototype AND the criterion requires a positive count. Either alone closes the hole, so each survives its own mutation while the other holds; the mutation that reproduces the actual defect (bare truthy lookup on a plain-`{}` tally) is red.
- on the raw plane the resolution tally is shared across containers, so a target resolving in *any* container counts. The feeder is unaffected (the record plane has one container), but a future caller running the guard over the raw plane would need this split per container.

- `detail_with_lines` is refused: its header/lines split exists only in the raw payload, and no usable PLM config can reach the mode anyway (the probe builder only ever sends the K3 `bomKey` dialect for it).
- a PLM client that does not paginate caps at 999 lines (above, the feeder cannot prove the source is not withholding rows).
- the ERP body accepts `projectId` but the ERP material run does not use it.
- these admin routes follow the existing platform model in which an admin's tenant is not pinned by `resolveTenantId` — shared with the other admin routes in this file, not introduced here.

Implements the code scope of #3889. Keep #3889 open until this PR is reviewed and merged.

#3751 remains open for its separate on-prem deployment acceptance gate.


zensgit:
TARGET_RECONCILIATION_REQUIRED

state: OWNER_HOLD
exact_head_sha: a277f8171935cfde6b657001ae33b3b72e9a7e78

The PR body still claims issue #3889 as its implementation boundary, but that issue now returns 404 and is no longer present in the live open/closed issue ledger. The branch and its green checks remain intact, but there is no current issue authorization to reinterpret this large readonly-feeder feature as part of #2343 or another surviving tracker.

Owner decision required:
- preserve and retarget this PR to an explicit live task/issue with the intended acceptance boundary; or
- close it as retired/superseded.

No close, rebase, merge, or scope transfer was performed by this sweep.

zensgit:
## Adversarial review (opus): **CHANGES-REQUESTED**

Security boundary holds and survives mutation — readonly external / no external write / no internal write / values-free / approved-only / tenant isolation / admin-gate placement all verified. But the PR's two headline claims — **bounded pagination** and the **fail-closed matrix** — are falsified, both reproduced end-to-end against the real adapters (not inferred from the diff).

### P1 — `bridge:legacy-sql-readonly` silently truncates

A 5000-row source yields **20 rows with `status: ready`, `rowErrors: 0`, zero errors**. The adapter clamps to `DEFAULT_MAX_LIMIT = 20` (`bridge-agent-readonly-adapter.cjs:46,270-275`), returns `done: true`, and `/query/:object` only sends `{limit, filters}` — **`request.cursor` is never emitted, so this kind structurally cannot page**. The runner's full-page guard compares `pageRows.length >= pageSize` (20 >= 500 → false), i.e. it tests the **requested** page size, not the adapter's **effective** limit — so the guard never fires. This kind is in both `PLM_SOURCE_KINDS` and `ERP_SOURCE_KINDS` and is explicitly claimed as supported. **Silent under-read presented as a complete, ready intake is a data-integrity P1.**

### P2

- **PLM BOM feeder has a 499-row hard ceiling.** Real PLM wrapper + `single_record`: 50 ✅, 499 ✅, **exactly 500 → 502 `SOURCE_RUN_PAGINATION_AMBIGUOUS`**, 600 → 502, with no config escape. `executeConfiguredRead` slices the raw container to `rowCap` (500), so `pageRows.length` is always exactly 500 while PLM returns everything at once with `done: true` and no cursor — the multi-page path never activates. Verified that "just send a limit" does NOT fix it: a fully well-behaved adapter also 502s at exactly 500 rows. The advertised 5000-row bound is unreachable.
- **12 apply-verified mutations survive**, including the PR's own promised **cursor-loop** and **page-cap** guards, and **M25: deleting `requireAccess(req, 'admin')` from BOTH new routes leaves `http-routes.test.cjs` fully green** → the new admin endpoints have **zero 403 coverage**.
- **ERP/K3 material channel is effectively capped at 100 rows** (10×10), and exactly 100 also 422s.

### P3 / NIT

`sourceTotalKnown` is a dead value (evidence cannot reveal truncation); approved-only has no route-level negative (the real gate is in the store, and it is there); `detail_with_lines` is unusable for the PLM wrapper (`buildReadSourceProbeRequest` only emits the K3 dialect `options.bomKey`; PLM needs `filters.productId`), which structurally makes the `SOURCE_RUN_PROJECT_SCOPE_MISMATCH` guard unreachable on the PLM path.

### Runs

91/93 CJS suites pass; the 2 failures are `MODULE_NOT_FOUND` environment artifacts that fail identically on `origin/main`. package.json test-chain is a clean UNION (85→86, zero loss). Note this chain runs only in `integration-guard.yml` — **not a required check**.

Full report with the mutation matrix and repro scripts: `/tmp/pr4093-review-claude-20260711.md`.

I'm taking the fixes on this branch — the core correction is that **completeness must be PROVEN, and an unprovable read must fail closed** rather than reporting `ready`.

zensgit:
## Round-2 独立对抗审阅 @ `078988d2a` — **CHANGES-REQUESTED**(2×P1 + 3×P2)

**上一轮的修复是真的。** 用真适配器复核:bridge 静默截断已消失(5000 行源 → 502 `COMPLETENESS_UNPROVABLE`);M25 亲手删掉两个新路由的 `requireAccess(req,'admin')` → `http-routes.test.cjs` **RED**(`actual: 200, expected: 403`);12 个存活 mutation 杀掉 9 个;`completenessProof` 取代了死值 `sourceTotalKnown`。「**永不信 `done:true`**」是本 PR 最有价值的设计决断。

**但同一类缺陷——把"看起来完整"当成"已证完整"——在另外两条通道上仍然活着。**两条都是真适配器 end-to-end 复现,不是读代码推断的。

### P1-A · 真实 BOM **树**:2500 行 → 只吃进 3 行,`status: ready`、`rowErrors: 0`

`executeConfiguredRead` 的数据面是从 `raced.raw`(适配器**未加工**的上游载荷)建的,而 wrapper 的**展平 + offset 翻页 + 字段别名归一化**全都在 `raced.records` 里。探针实测:同一次读里 `page.returnedRecordCount = 1000`(适配器正确展平出的 BOM 行)紧挨着 `mapped recordCount = 1`(树根)——**运行时把检测材料算好了摆在那儿,feeder 两个都不读。** 配置绕不开:展平后的行只存在于 `records`,任何 `containerPaths` 都够不到。

可达性不是假设:树形是 shipped 适配器的**一等受测形状**(`plm-yuantus-wrapper.test.cjs:223-246` 有专门的防环断言——没有真实树形载荷,没人会去写它),而**本 PR 是全仓唯一存在 PLM read-source config 的地方**(preset 注册表里只有 K3)。作者自己的 PLM 测试只覆盖了扁平 + 服务端翻页的 client,恰好绕开树形。

### P1-B · K3 不认 PageIndex 时,重复页被 `declared_total` 背书为"可证完整"

真 k3-wise-webapi 适配器:K3 声明 `ROWCOUNT: 100`、对 10 个页码**都返回第 1 页** → `ready` / `completenessProof: declared_total` / `sourceRows: 100`,而 intake 实际只持有 **100 行里 10 个不同物料**,`rowErrors: 0`。适配器**明确暴露了** K3 回显的 PAGEINDEX(注释原文:"what K3 reports it applied, vs what we requested"),`buildInternalPage` 也把它算进了 `page.pageIndex`——feeder 从不读。

本 PR 拒绝相信 `done: true`,却**无条件相信"翻页真的发生了"**。一个把第 1 页读十遍就能满足的证明,不是证明。

### P2

- **P2-A**:`effectivePageSize` 在适配器**不自报** `metadata.limit` 时**回落到"请求的上限"**(fail-open)→ 上一轮的 P1 原样复活。逐个查过 5 个 shipped kind:**今日不可达**(故未判 P1),但它依赖一条**未写进契约、无人强制、零测试**的隐含不变量「凡夹取者必自报」,而 5 个 kind 里有 2 个根本不报。
- **P2-B**:4 个存活 mutation——M12 project-scope(且在 `list_page` 下**恒不触发**)、M16、M20 路由 `inputs` 子 allowlist、M21 system-kind 不匹配。
- **P2-C**:非翻页 PLM client → **999 行硬天花板**;恰好 1000 行时会对外部 PLM 发起 **10 次冗余读**才 422。

### 站得住(返工时别弄丢)

只读边界 · values-free 投影(往公共投影里注入 `intake` → 2 个 suite RED)· admin 403 门 · approved-only · test-chain UNION —— 全部 PASS **且有测**。**intake 契约的必填字段校验是有效兜底**(真实 PLM 字段别名 → 422 fail-closed,这也是 P1-A 没有变成脏数据落库的原因)。

### 两条工程纪律记录

- 审阅自纠:第一次跑「公共投影泄露 intake」报 GREEN,事后查出是 harness bug(多行 pattern 在 `perl -pe` 逐行模式下**从未落盘**)——改成单行 + md5 校验落盘后变 RED。**存活 mutation 必须先验证变异真的落盘**,否则是假阳性。
- CI 如实记录:`npm test` EXIT=0(88 suite 全过;上一轮那 2 个 MODULE_NOT_FOUND 确系未装依赖,核实为真)。但该 CJS 链**只在 `integration-guard.yml` 跑,不是 required check**。

未合并、未 arm。返工中。

zensgit:
## Round-3 独立对抗审阅 @ `863ee5974` — **CHANGES-REQUESTED**(1 P1 + 2 P2)

**Round-2 的 5 条 finding 全部真修**,逐条用真适配器独立复现,无一采信自证:BOM 树 2500 → **2500 行**;不认 PageIndex 的 K3 在**第 1 页**即拒(**1 次**外部读,非 10 次);fail-open 回落已堵;M12/M16/M20/M21 四条守卫补测(逐个 neuter 复验全变红);10 次冗余读 → 2 次;R2 认定站得住的 7 道承重守卫**一条没丢**。**页指纹是本轮最好的决断**——它还抓住了审阅另造的「**回显页码但仍服务第 1 页**」的欺骗源,补上了 `echoedPageIndex` 单独做不到的那一半。

**但修法本身开了新的静默通道:同一类缺陷从「行丢失」变成「字段清空」。**

### P1 · 探针面与 feeder 面分叉 → 业务字段静默变 null,仍报 `ready`

feeder 吃 `adapter_records`(`source-run.cjs:15`),而 configured-read **路由**吃 `raw_containers`(`http-routes.cjs:2610` 不传第 3 参 —— 已核实调用点,不是采信注释)。**两面的字段名不同**;`mapRecord` 对解析不到的 source **静默填 null**(`read-source-read-runtime.cjs:141-148`),而 intake **只强制 `pathKey`**(`intake.cjs:177-206`)—— `designQty` / `designUnit` / `childDrawingNo` 全部接受 null **且不报 rowError**。

**决定性证据(不需要任何构造)**:真 PLM wrapper + 一个**在探针里 100% 干净**的 config(四个字段全部解析出真实值)→ feeder **`ready` / 2500 行 / `rowErrors: 0` / 2500 行的 `designQty` 与 `designUnit` 全是 null**。

**操作员正是对着探针写配置的。探针验过的东西,不是 feeder 实际执行的东西。**

更硬的一刀:`data-source:sql-readonly` 适配器**从不产生 raw**(`grep raw` 零命中)→ 探针对它**恒返回 `RESPONSE_UNRECOGNIZED`**,而 feeder 现在能跑。**该 kind 没有任何可用的 config 编写/验证手段**;盲写错一个可选字段名 → `ready` / `rowErrors: 0` / `erpMaterialName: null`。(顺带核实:「sql-readonly 从来不工作」属实,**本 PR 新启用了它**。)

### P2-1 · `assertSourceKind` 跨 run-type 白名单 = **未测守卫**

neuter 后全绿。唯一那条测试用的 kind 是 `'http'`,会被下游 `sourceCapability()` 先拦下,**根本走不到这道守卫**。真正的门(K3 config 走 PLM 路由)**零覆盖**。下游是 fail-closed,所以不是实弹,但这道守卫本身没有测。

### P2-2 · 被当作 BOM-树防线卖的「结构性交叉校验」在记录面**恒真**

`source-run.cjs:383` 的 mapped-vs-adapter 半 mutation **SURVIVED** —— `records.length > pageSize` 已在 :369 抛 TOO_LARGE,两者**由构造相等**,该校验**永不可能触发**。注释声称的纵深防御是假的,**会误导下一个作者**。

### 存活 mutation 4/23

S6(P2-1)· D1(P2-2)· N6 + N9(惰性、结构不可达 —— 如实记 P3,不假装有覆盖)。

### P3 / NIT

页指纹算在**有损的 fieldMap 投影**上(实测误杀:两页仅在**未映射字段**上不同 → 502)· K3 必须回显 `PAGEINDEX` 否则该端点不可用 · ERP 路由仍收一个用不上的 `projectId`(**连续 3 轮未修**)。

### CI 如实记录

该 CJS 链 EXIT=0(54 suites),但**只**在 `integration-guard.yml` 跑,**不是 required check** → 它变红不阻塞 admin-merge,以本地实跑为准。只读边界成立(唯一的适配器调用 = `adapter.read()`)。

未合并、未 arm。返工中。

zensgit:
## Round-4 独立对抗审阅 @ `0fd55df08` — **CHANGES-REQUESTED**(1 P1 + 1 P2 + 3 P3)

**关于「每轮修好、每轮又开一条新通道」的模式 —— 诚实结论:这轮没重演。** R3 的 3 条 finding **全部真修**(逐条独立证伪,非采信自证);**16 道旧守卫逐个 mutation 复验,零丢失**。

**专门攻过、确认关着的**:`rowSource` **不能**流进 feeder(两条 source-run 路由的 allowlist 都没有它,runner 里是硬编码)→ **R2 的原始 P1(BOM 树 2500 → 3 行)无法经 config / 请求体复活**。values-free 红线守住:新加的 `fieldResolution` / `rowFingerprints` 挂在非枚举的 `page` 上,实测不进公共投影 / evidence / error details。

**但 R4 新加的字段守卫在一条路径上有缺口,而那条缺口把 R3 的 P1 原样放了回来。**

### P1 · 重复的 fieldMap target 绕过 `SOURCE_RUN_FIELD_MAP_UNRESOLVED`

三个事实叠加:

1. `read-source-read-runtime.cjs:152` 的 `mapRecord` **无条件赋值** —— 后一条 entry 会用 null **覆盖**前一条已解析出的真实值;
2. `:153-155` 的计数器**只在解析成功时递增,且键在 target 上**;
3. `read-source-config.cjs:107` **不查重复 target**。

于是「**两个拼写都写上**」这种最自然的写法 —— `[{quantity→designQty}, {qty→designQty}]` —— 第一条读到真实值、喂饱了计数器,第二条把它**覆盖成 null**,守卫放行。

真校验器 + 真适配器,**两条路径都跑通**:

- PLM 2500 行 BOM → `ready` / `short_page` / `rowErrors: 0` / **`designQty` null 2500-of-2500**
- K3 `ROWCOUNT=20` → `ready` / `declared_total` / `rowErrors: 0` / **`erpMaterialName` null 20-of-20**
- 对照组:单条不可解析 entry → **422**(守卫按设计工作);把两条 entry **顺序调换** → 值正确(证明这是 last-write-wins 的**顺序相关**行为)

性质是**数据丢失**,不是稀疏列:那条解析成功的 entry **确实读到了非 null 值**,然后被覆盖掉。

(审阅措辞校正:显式的 null 属于 `resolved: true`,所以单条 entry 出现「计数 = N 且值全 null」是**正确行为**。区别在上面这条。)

### P2 · 守卫的 `declared_total` 出口**未测**

mutation `F3`(只删这一个出口的守卫)→ **3 个 suite 全绿**。守卫**是接上的**(实测它在那儿会触发 422),属**未测守卫**;但 `erp:k3-wise-webapi` **只能**靠 `declared_total` 证明完整性 —— **ERP feeder 走的正是这条没测的出口**。

### P3

1. **双平面 parity 是 opt-in**:探针**默认**仍是 `raw_containers`。PR body 里那句 "what is verified is what executes" **比代码强** —— 不传参的操作员,验证面和执行面仍然是两个。
2. **PR body 第 41 行已过时**:仍在卖 R4 **正确删掉**的那个三方 cross-check(正是 R3 说的「误导下一个作者」)。
3. body 的 unreachable 清单少列 2 条(`F7` / `F2`)。

### mutation 台账

**27 个 md5-apply-verified mutation:24 KILLED / 3 SURVIVED**。作者声称的 24/24 **在它自己选的那一组上属实**;审阅另行构造的 3 个存活,其中 `F3` 是真 P2。

### CI 如实记录

`npm test` **EXIT=0**(87 条链式命令,零 failure;不报未经核实的 suite 数 —— 上一轮在 grep 口径上栽过)。但该链只在 `integration-guard.yml:240` 跑,**不是 required check** —— **CI 全绿 ≠ 它过了**。

未合并、未 arm。返工中。

zensgit:
## Round-5 独立对抗审阅 @ `1f5269303` — **APPROVE**(0 P1 · 0 P2 · 2 P3 · 2 NIT),硬化已落 @ `2be6f2698`

### 收敛曲线(五轮,每轮都是真适配器 end-to-end 挖出来的,不是读代码推断)

| 轮 | 判定 |
|---|---|
| R1 | 1 P1 + 3 P2 —— bridge 静默截断(5000 行 → 20 行报 `ready`) |
| R2 | 2 P1 + 3 P2 —— BOM 树 2500 行 → 3 行;不认 PageIndex 的 K3 重复页被背书为完整 |
| R3 | 1 P1 + 2 P2 —— 探针面/feeder 面分叉 → 2500 行字段全 null 仍报 `ready` |
| R4 | 1 P1 + 1 P2 —— 重复 fieldMap target,last-write-wins 用 null 覆盖已读到的真实值 |
| **R5** | **0 P1 · 0 P2** —— APPROVE |

**四轮返工里 20+ 道旧守卫逐个 mutation 复验,零丢失。**

### R5 审阅独立确认(逐条证伪,非采信自证)

- **R4-P1 三层封死**:写入校验拒 `duplicate_target` · **执行时再校验**(所以「**存量已 approved 的重复 target 旧 config**」也 fail-closed 成 400 —— 审阅直接造了这个态)· 把重复 fieldMap **强行偷渡过两道校验器**后,真 PLM + 2500 行 BOM 树 → `designQty=7`、**null 0/2500**(R4 时是 2500/2500 全 null)。
- **R4-P2**:`declared_total` 出口(**ERP/K3 feeder 唯一的完整性证明出口**)补了负例,`F3` 现在 KILLED,且断言 `reads.length === 2` —— 证明是字段守卫拒的,不是提前失败的假绿。
- **探针默认面**:审阅**明确背书作者的相反决定,并撤销了自己上一轮那条 finding**。`read-source-read-runtime.cjs:207` 的 mode 围栏是硬证据:记录面**在构造上表达不了** `detail_with_lines` / `resolver_lookup`;翻默认会打断已 approved 的 config,并**静默改变** shipped `#1709 S3-2` 路由的返回值。**parity 是「可用,非自动」,body 现在与代码一致。**

### R5 硬化 @ `2be6f2698`

- **prototype-key 绕过(本轮唯一新洞)** —— 清理「不可达的 `< 1` 臂」时把**承重的 `Number.isInteger` 类型守卫一起删了**。校验器接受 `constructor` / `toString` / `valueOf` / `hasOwnProperty` 作字段名,而计数器是裸 `{}` → **一个在 0 行上解析到的字段 → `ready` / 2500 行**。修:三个计数器全部 `Object.create(null)`(含交给受信调用方的冻结副本),守卫改用 **own-property + 正计数**(非计数值一律 fail-closed,不论类型)。五个 prototype 成员全部覆盖。
- 记录面 `page` 非枚举补测(mutation 现在变红)· `duplicate_target` 的**追溯性**写进 Known limitations(执行时再校验 ⇒ 已 approved 的重复 target 旧 config 从此在 shipped 路由 400 —— **大声失败,因为它本来就在丢数据**)。

### mutation 台账 —— 作者**没有硬凑 24/24**,这点值得记

24 个 mutation:22 KILLED / **2 SURVIVED by design**。两个存活项是 prototype 防护的**冗余双层**(单独删任一层,另一层仍独立堵住这个洞),而**精确复现真实缺陷的那个组合 mutation(裸查找 + 普通 `{}`)是 KILLED 的**。这正是 mutation testing 的已知边界:**它证明守卫承重,不证明守卫正确**。作者如实写进了 body 而不是报一个漂亮的 24/24。

### 我的独立复验 @ `2be6f2698`

净测 PASS · **组合 mutation(md5 验证真落盘)→ RED**(`Missing expected rejection` @ `testPrototypeNamedTargetsCannotSatisfyTheFieldGuard`)· 变异已还原(`git diff` 干净)· 全链 `npm test` **EXIT=0**(87 个 `&&` 串联的 node 调用)。

### 如实记录

该 CJS 测试链**只在 `integration-guard.yml` 跑,不是 required check** —— **CI 全绿 ≠ 它过了**,以上均为本地实跑。

PR #4100, opened 2026-07-11
feat(directory): make directory-sync N+1 (user/get fan-out) ops-verifiable (§7.7)

zensgit:
## What / why

Roadmap **§7.7 Sync Performance** targets the directory-sync N+1: `fetchAllUsers` issues one `topapi/v2/user/get` **detail call per unique user**, so a 2,000-employee tenant does ~2,000 sequential HTTP round trips inside the pull.

The actual §7.7 fix ("prefer `user/list` as the primary field source") is **staging-gated** — it cannot be verified here that `user/list` returns every field the sync reads (no real DingTalk tenant). So this PR does **not** change WHICH fields are read. It makes the N+1 **ops-verifiable**: instrument the external call pattern and surface it on the run record, so operators can measure the cost and later **prove the staging fix helped** via before/after on `externalUserDetailCalls`.

## What this instruments

New pure module `directory-sync-api-telemetry.ts` (per-run counters + a `summarizeDirectorySyncApiCalls` fold), incremented at each external directory-pull call site in `fetchAllDepartments` / `fetchAllUsers`:

| stats key | meaning |
|---|---|
| `externalDepartmentListCalls` | `department/listsub` pages |
| `externalDepartmentDetailCalls` | `department/get` (dept-manager enrichment) |
| `externalUserListPageCalls` | `user/listsub` pages |
| **`externalUserDetailCalls`** | **`user/get` — THE N+1** (one per UNIQUE user; deduped) |
| `externalDirectoryPullCalls` | sum of the four above |

- `userDetailCalls` increments only on first sight of a userId, so a user in two departments is counted once — the counter equals the unique-user total, which is the N+1 signature.
- Folded into the run's `stats` JSONB next to `accountsSynced`. `summarizeRun` already returns the whole `stats` blob, so `GET /integrations/:id/runs` surfaces the new keys with **no route/UI change**. When `externalUserDetailCalls` ≈ `accountsSynced`, the per-user fan-out is fully in play — the exact number the staging fix should drive down.
- `externalDirectoryPullCalls` is deliberately **not** named "total": it excludes the (cached) app-access-token fetch and the manual-run diagnostic sample, whose logical-vs-HTTP counts are ambiguous. It is the honest sum of the four pull categories.
- Counters are threaded only through the sync (which persists a run row); the preview path (no run row) passes no counter — the param is optional.

## Test evidence

Both suites run locally against the repo toolchain (vitest 1.6.1); real-DB against a fresh migrated Postgres.

- **Unit** `tests/unit/directory-sync-api-telemetry.test.ts` (no DB, runs in required `test(20.x)`): drives the real `fetchAllDepartments`/`fetchAllUsers` with only the DingTalk client mocked. Fixture shares one user across two departments + paginates one department. **4 passed.**
- **Real-DB**: extends the already-whitelisted `directory-sync-preview-apply-parity.db.test.ts` (two-point-wired: vitest.config exclude + plugin-tests.yml whitelist — no new wiring) with one focused case that asserts the counts land in the persisted run `stats`, deriving expected from the run itself. **7 passed** (`... --config vitest.integration.config.ts`).
- Full unit suite: **369 files / 5078 tests passed**; `tsc --noEmit` clean.

### Mutation table (impl committed first, reverted, run, restored)

| mutation | test that went RED |
|---|---|
| remove `userDetailCalls += 1` | unit: `expected +0 to be 4` (counter 0 vs mock call count 4 — divergence proves non-tautology) |
| remove `userListPageCalls += 1` | unit: `expected +0 to be 4` |
| remove `...summarizeDirectorySyncApiCalls(...)` fold | real-DB: `expected undefined to be 5` (stat not persisted) |

## Not done here (design-only / out of scope)

- **`user/list` primary field-source flip** — the actual perf fix. Stays **design-only, staging-gated**. No scaffold or dead code added; it needs a real tenant to verify `user/list` field completeness before the flip is safe.
- **Batch upsert of account-department links** — already landed (DT-PERF-01, `unnest`).
- **Batch upsert of departments/accounts** — still per-row loops in the apply transaction. Batchable via `unnest`, but touches the hot transaction and the per-department `dept_manager_userid_list` carry-forward; left as a design note to keep this PR to the instrumentation.
- **bcrypt outside the transaction** — `bcrypt.hash` runs inside the apply transaction in the auto-admission loop. Moving it out entangles with the identity-match cascade that DECIDES who is admitted (that decision is inside the txn); left as a design note.

## Ops note

Additive-only: new keys appear in `directory_sync_runs.stats` and on the runs API. No behavior change to what the sync reads or writes.

## Honest gaps / CI coverage

- The **increment logic** is covered by the unit test in the required `test(20.x)` job. The **fold-into-persisted-stats** is covered only by the `.db` case, which runs in **`plugin-tests.yml` (non-required)**, not required CI — consistent with this line's model (all directory-sync stats live in `.db` tests). Green required CI does not by itself mean that `.db` case passed.
- `externalUserDetailCalls === accountsSynced` is pinned as a **current-N+1 canary** and is **expected to go RED when the `user/list` field-source flip lands** (detail count drops below account count = the fix working). The assertion carries a comment saying to relax it to `<=` at that point. Flagged so a future failure reads as a signpost, not a regression.

### DIFF
--- a/packages/core-backend/src/directory/directory-sync-api-telemetry.ts
+++ b/packages/core-backend/src/directory/directory-sync-api-telemetry.ts
@@ -1,0 +1,73 @@
+/**
+ * DT-PERF (roadmap §7.7 "Sync Performance") — make the directory-sync external API call
+ * pattern OPS-VERIFIABLE.
+ *
+ * The pull walks the department tree and, for every UNIQUE user it sees, issues one
+ * `topapi/v2/user/get` detail call (directory-sync.ts `fetchAllUsers`). On a 2,000-employee
+ * tenant that is ~2,000 sequential HTTP round trips — the N+1 the roadmap targets. The fix
+ * itself (prefer `user/list` fields, detail only when missing) is STAGING-GATED: it cannot be
+ * proven here that `user/list` returns every field the sync reads, so this module does NOT
+ * change WHICH fields are read. It only instruments the call pattern so an operator can SEE
+ * the cost on the run record and later prove the staging fix helped (before/after
+ * `externalUserDetailCalls`).
+ *
+ * Counters are per DIRECTORY-PULL client invocation — one increment per real call at the
+ * `fetchAllDepartments` / `fetchAllUsers` call sites. They are folded into the sync run's
+ * `stats` JSONB (see `summarizeDirectorySyncApiCalls`), which `summarizeRun` already returns
+ * verbatim, so the counts reach the runs API/UI with no new surface.
+ */
+
+/** One counter per external directory-pull client call issued during a sync run. */
+export interface DirectorySyncApiCallCounters {
+  /** `department/listsub` pages (one per department visited in the BFS). */
+  departmentListCalls: number
+  /** `department/get` detail calls (dept-manager enrichment; one per department, best-effort). */
+  departmentDetailCalls: number
+  /** `user/listsub` pages (one per page per department; more pages = larger tenant). */
+  userListPageCalls: number
+  /**
+   * `user/get` detail calls — THE N+1. One per UNIQUE user (deduped: a user in two
+   * departments is fetched once). Compare against `accountsSynced`: when they are ~equal the
+   * per-user detail fan-out is fully in play.
+   */
+  userDetailCalls: number
+}
+
+/** A fresh zeroed counter set for one sync run. */
+export function createDirectorySyncApiCallCounters(): DirectorySyncApiCallCounters {
+  return {
+    departmentListCalls: 0,
+    departmentDetailCalls: 0,
+    userListPageCalls: 0,
+    userDetailCalls: 0,
+  }
+}
+
+/**
+ * The stats-JSONB projection of the counters. Keys are prefixed `external*` so ops can see at
+ * a glance these are network cost, not row counts.
+ *
+ * `externalDirectoryPullCalls` is the SUM OF THE FOUR PULL CATEGORIES above — deliberately NOT
+ * named "total": it excludes the (cached) app-access-token fetch and the manual-run diagnostic
+ * sample, which are not part of the per-user fan-out and whose logical-vs-HTTP count is
+ * ambiguous. It is the honest aggregate of what the directory pull itself costs.
+ */
+export function summarizeDirectorySyncApiCalls(counters: DirectorySyncApiCallCounters): {
+  externalDepartmentListCalls: number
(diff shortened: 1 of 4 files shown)


zensgit:
CODEX_REVIEW
lane_id: MS-DINGTALK-SYNC-NPLUS1-TELEMETRY
reviewed_head_sha: 155d8d5e1e1901a74f2177f9c31158d7794fade8
scope_checked: external-call counter placement; unique-user dedupe; pagination accounting; persisted run-stats fold; current-main compatibility

tests_and_ci:
- focused unit telemetry suite: 4/4 PASS
- real PostgreSQL preview/apply parity suite: 7/7 PASS
- current-main merge simulation at `5e5673ab8`: no conflicts
- diff check: PASS

verdict: RATIFIED@SHA
next_authorized_action: owner review/merge decision. If the head changes, this ratification no longer applies and the replacement SHA requires review.

Notes:
- The counters measure issued directory-pull client invocations and are persisted on completed runs. A pull that fails before completion still follows the existing failed-run path and does not preserve partial call counters; that is consistent with this PR's successful-run baseline scope.
- The `externalUserDetailCalls === accountsSynced` assertion is intentionally a current-N+1 canary. The later staging-gated `user/list` primary-source fix should replace that equality with a proven drop, as documented in the test.

zensgit:
CODEX_EXACT_HEAD_RATIFICATION

reviewed_final_head_sha: 169441496134d2cc45c2b016f5dde9f64797d48b
prior_ratified_head_sha: 155d8d5e1e1901a74f2177f9c31158d7794fade8
merged_commit_sha: 29582328f4308baf872994a787ce012f015eadcd
verdict: RATIFIED_AT_FINAL_SHA

The final head was a pure rebase of the already-reviewed three-commit series:
- git range-diff reports all three commits as exact "=" matches.
- The four PR files are byte-identical between the prior ratified head and the final head (empty restricted tree diff).
- Final-head GitHub checks: 15 PASS + 1 expected SKIPPED, including Node 18/20, web-tests, migration replay, contracts, and E2E.

No implementation delta was introduced after the prior review, so its focused unit, real-PostgreSQL parity, TypeScript, and counter-placement findings carry forward unchanged to the final merged SHA.

This closes the exact-head evidence gap only. This same-account comment is a technical verdict, not an independent GitHub approval or merge authorization.

PR #4104, opened 2026-07-11
test(multitable): add permission-list postdeploy closeout smoke

zensgit:
## What changed

- add a dependency-free postdeploy smoke for the three permission-list endpoints tracked by #3408
- create a temporary sheet/field/view/record, probe all three GET contracts, and cascade-delete the sheet in `finally`
- emit only statuses, booleans, counts, and bounded error codes; leak-scan token, base/resource IDs, fixture names, and values before writing artifacts
- add a PR contract job plus a manual deployed-environment job that keeps host/base identifiers out of dispatch metadata
- ship the smoke in the multitable on-prem package and require it in package verification

## Safety boundaries

- automatic base discovery considers only token-visible bases with explicit `perf`, `smoke`, `test`, `staging`, `baseline`, or `benchmark` name tokens; otherwise it fails before mutation
- an explicit `MULTITABLE_PERF_BASE_ID` repo variable remains the preferred path when configured
- a client-selected sheet ID lets cleanup run even if a successful create response is malformed
- no raw response body, host, token, permission value, business value, or resource ID is written to logs/artifacts
- this PR does not claim the private on-prem target has passed; #3408 stays open until an authorized entity-machine rerun reports 3/3 HTTP 200

## Verification

- Node 20 smoke contracts: 7/7 PASS
- Windows/on-prem hardening contracts: 14/14 PASS
- on-prem package contracts: 9/9 PASS
- Node syntax, shell syntax, YAML parse, and `git show --check`: PASS


zensgit:
TARGET_RECONCILIATION_REQUIRED

state: OWNER_HOLD
exact_head_sha: 82fab58128b9a9f733d0caa9ae50f3872384e440

The PR body still claims issue #3408 as its postdeploy acceptance boundary, but that issue now returns 404 and is absent from the live issue ledger. The smoke/package hardening remains available and its prior checks are green, but a deleted tracker cannot supply the authorization or closure criteria for a deployed mutation smoke.

Owner decision required:
- preserve this as standalone operational hardening under a new explicit live tracker and acceptance boundary; or
- close it as retired/superseded.

No close, rebase, merge, deployed smoke, or scope transfer was performed by this sweep.

zensgit:
Review result (same-account evidence, not a formal approval): PASS. I reviewed the workflow, smoke script, packaging hooks, and contract tests. The deployed probe is workflow_dispatch-only, values-free, fails closed before mutation when no safe reusable base exists, cleans up the temporary sheet, and the package verifier now requires the smoke script. Required checks were green on the stale head; current blocker is only main moving.

PR #4118, opened 2026-07-11
fix(dingtalk): P2 wire-contract — IM_ROBOT casing + callback reads official params.action

zensgit:
Two P2 wire-contract deviations from the owner REQUEST-CHANGES review. Interactive-card path stays flag-OFF.

**P2-1 openSpaceId casing (`im_robot` → `IM_ROBOT`)** — `client.ts` sent lowercase `dtv1.card//im_robot.<userId>` while the same createAndDeliver body already sends `imRobotOpenDeliverModel.spaceType: 'IM_ROBOT'`. DingTalk's official card-callback contract + sample library use UPPERCASE; lowercase was internally inconsistent and unverified. Flipped the two fixtures that pinned the lowercase guess (a guess, not evidence). UAT-confirmable.

**P2-2 callback parser reads the official `params.action`** — the parser read ONLY `cardPrivateData.actionIds`, which depends on the operator naming the component's actionId literally `approve` (fragile); DingTalk's official approval-template callback fills `cardPrivateData.params.action`. Now: `params.action` = PRIMARY (official), `actionIds` = SECONDARY (backward compat); both present must AGREE (disagreement → fail-closed no-op); either alone works; neither readable → existing no-op. approve-only invariant kept; P3-1 unreadable-content fail-closed preserved for genuinely-unreadable content, but a readable frame filling `params.action` is no longer rejected.

**Verify:** full backend unit 370 files / 5094 pass; tsc clean. New goldens cover params-only / agree / disagree / reject / backward-compat / unreadable, with RED-before (5 fail on the old parser) and a mutation check proving the `params.action` path is load-bearing (neuter → 5 fail). No DB migration.

### DIFF
--- a/packages/core-backend/src/integrations/dingtalk/client.ts
+++ b/packages/core-backend/src/integrations/dingtalk/client.ts
@@ -861,7 +861,11 @@ export async function sendDingTalkInteractiveApprovalCard(
   if (!rejectUrl) throw new Error('DingTalk interactive-card reject url is required')
   if (!callbackRouteKey) throw new Error('DingTalk interactive-card callback route key is required')
 
-  const openSpaceId = `dtv1.card//im_robot.${userId}`
+  // openSpaceId spaceType segment casing per DingTalk's official card-callback contract: the
+  // documented value is UPPERCASE `IM_ROBOT` (the sample library + `imRobotOpenDeliverModel.spaceType`
+  // below both use `IM_ROBOT`; a lowercase segment here was internally inconsistent). UAT-confirmable
+  // against the real createAndDeliver API.
+  const openSpaceId = `dtv1.card//IM_ROBOT.${userId}`
   const payload = await requestDingTalkJson(
     `${normalizeDingTalkOpenApiBaseUrl(config.openApiBaseUrl)}/v1.0/card/instances/createAndDeliver`,
     {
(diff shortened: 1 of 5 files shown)


zensgit:
Rebased onto main (now carrying #4112's strict node+epoch binding and #4116's cross-corp gate).

**Conflict resolved in `interactive-card-callback.ts`** — the two changes are semantically independent concerns in one file, so the resolution keeps **both**: #4116's corp gate (`readCallbackCorpId` + `resolveIntegrationCorpId` + `corp_mismatch`) is intact, and #4118's new wire-action parser (`readWireAction`/`WireActionRead`, `params.action` primary + `actionIds` secondary) replaces the old `readWireActionIds`. Verified by anchored greps in both directions — the corp gate survived, and the old parser token is gone (not resurrected).

**Combined regression on the rebased tree (the point of landing this last):**
- real-DB **61/61** — `dingtalk-approval-card-callback.db` + `approval-card-delivery-wrapper.db` + `dingtalk-approval-card-deliveries.db` + `automation-dingtalk-approval-card-action`
- unit **71/71** — callback parser + Stream SDK adapter + work-notification (`IM_ROBOT` casing)
- `tsc --noEmit` clean

The three changes compose: epoch binding (#4112) + corp gate (#4116) + wire contract (#4118).

Stream flag stays **OFF**; U1–U13 UAT and production enablement untouched.

PR #4135, opened 2026-07-12
fix(attendance): guard empty regression worker queue

zensgit:
## What

- guard the final active-worker array expansion when the queue has drained completely
- preserve the existing wait behavior whenever worker PIDs remain

## Why

macOS Bash 3.2 treats an empty array expansion as an unbound variable under set -u. With PROFILE=contracts and MAX_PARALLEL=1, each worker is drained inside the launch loop, so the final wait loop expanded an empty ACTIVE_PIDS array and exited before writing summary.json.

The same focused test failed 3/4 on current main before this patch.

## Verification

- attendance-fast-parallel-regression.test.mjs: 4/4 PASS (baseline was 3/4)
- all scripts/ops/attendance-*.test.mjs: 122/122 PASS
- bash -n: PASS
- git diff --check: PASS

No runtime deployment or workflow dispatch was performed.

### DIFF
--- a/scripts/ops/attendance-fast-parallel-regression.sh
+++ b/scripts/ops/attendance-fast-parallel-regression.sh
@@ -133,11 +133,14 @@ for index in "${!CHECK_NAMES[@]}"; do
   fi
 done
 
-for pid in "${ACTIVE_PIDS[@]}"; do
-  if [[ -n "$pid" ]]; then
-    wait "$pid" || true
-  fi
-done
+# Bash 3.2 treats an empty array expansion as unbound under `set -u`.
+if (( ${#ACTIVE_PIDS[@]} > 0 )); then
+  for pid in "${ACTIVE_PIDS[@]}"; do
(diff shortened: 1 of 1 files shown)


zensgit:
Rebased onto current main ec6ac3af52c9c37f7e032e29f1c0bfbed47a74fc.

exact_head_sha=7de1d35e426ff8b56659914420406d75f261f8aa
focused_regression=4/4 PASS
bash_syntax=PASS
diff_check=PASS

The intervening main commit is approval-backend-only and does not overlap this one-file shell fix.

zensgit:
FINAL_CI: exact head 7de1d35e426ff8b56659914420406d75f261f8aa now has 11/11 GitHub checks SUCCESS, including Node 18/20 and coverage. Local focused 4/4 and full attendance-prefixed ops 122/122 remain the exact-head evidence. Holding merge for owner authorization.

PR #4126, opened 2026-07-11
fix(core-backend): uuid runtime dependency + production-install startup contract (entity-machine crash)

zensgit:
## What

corrective-6 for the #4101 entity-machine acceptance, plus a fail-open security fix the guard surfaced.

### 1. The boot crash (entity machine)

corrective-5 migrated cleanly (42P07 gone, `migration066=applied`) but the backend crashed on boot — `/api/health` 502:

```text
failureClass=RUNTIME_DEPENDENCY_DECLARED_AS_DEV_ONLY
missingRuntimeModule=uuid
```

`uuid` was declared only in core-backend `devDependencies` while `WorkflowDesigner` / `BPMNWorkflowEngine` / `DelayService` import it at module load; a `--prod` install skips devDependencies. **Fix**: `uuid` → `dependencies` (`@types/uuid` stays dev); lockfile reclassified.

### 2. Production-install startup contract (guard)

`tests/unit/runtime-dependency-classification.test.ts` — an **AST** BFS of the startup graph from `src/index.ts`. Per file it collects the HARD-eager specifiers whose absence crashes module evaluation under `--prod`:

- `import … from` / side-effect `import '…'` / `export … from` (declarations; type-only excluded)
- **any top-level** `require('…')` / `import('…')` — in **any** try/catch/finally position (only a swallowing catch would be safe, and that is not inferable structurally: try-finally propagates, a require in catch/finally is unguarded, a rethrowing catch does not swallow)

Lazy `require()`/`import()` inside function/class bodies are excluded structurally. **Every** AST occurrence is recorded as a `(module, source file)` pair — nothing is deduplicated — and each must be a production dependency of core-backend or the workspace root, **or** an entry on the `OPTIONAL_SOFT_DEPENDENCIES` exemption list (`@opentelemetry/api` @ `src/core/logger.ts`, `js-yaml` @ `src/services/ConfigService.ts` — each a degraded-path-only optional with no security role).

An exemption is bound by **three** independent locks, not just a module name:
1. **site** — the module may be eager only at that exact file;
2. **occurrence count** — exactly one occurrence, so a *second* eager import of the same module **in the same file** or anywhere else fails;
3. **loader shape** — that sole occurrence must sit inside the `try` block of a **swallowing** `catch`, verified structurally: a rethrowing `catch`, a `try…finally` with no catch, a call in `catch`/`finally`, or a `throw` in the `finally` all classify as **unguarded** and fail.

[...]

PR #4139, opened 2026-07-12
test(core-backend): harden optional startup dependency exemptions

zensgit:
## What

Stacked corrective follow-up for parent PR #4126 at exact base 27b4d4e292998b448a2ce87c50d2e23d28c9d9eb.

- retain every eager external occurrence instead of deduplicating by module+file
- require each optional-loader exemption to have exactly one top-level dynamic loader
- structurally verify that loader remains inside a swallowing try/catch
- reject loaders moved outside try, placed in catch/finally, rethrown, duplicated in the same file, or added elsewhere
- correct stale guard comments and add a synthetic guarded-loader shape matrix

Only the parent PR's startup-contract test changes. No runtime/package code is modified.

## Why

The parent guard's claimed occurrence lock still collapsed imports to module@@file. Adding an unguarded import 'js-yaml' in ConfigService.ts beside the existing guarded require therefore remained 6/6 green. Moving the existing require outside try also remained exempt by file.

## Verification

Clean tree:
- focused runtime dependency guard: 7/7 PASS
- core-backend tsc --noEmit: PASS
- full core-backend unit suite: 372/372 files, 5097/5097 tests PASS
- git diff --check: PASS

Mutation teeth:
- same-file unguarded import beside the guarded require: RED on duplicate occurrence
- existing js-yaml require moved outside try/catch: RED on guardedCalls=0
- both mutations were restored before commit

## Integration boundary

This PR targets the parent feature branch, not main. Parent #4126 remains the merge/package unit for #4101. The parent owner may merge or cherry-pick this one commit, then request exact-head re-review. No package build, entity-machine action, deploy, or self-merge is performed here.

### DIFF
--- a/packages/core-backend/package.json
+++ b/packages/core-backend/package.json
@@ -65,6 +65,7 @@
     "sanitize-html": "^2.13.0",
     "semver": "^7.7.3",
     "socket.io": "^4.6.1",
+    "uuid": "^9.0.1",
     "winston": "^3.17.0",
     "xlsx": "https://cdn.sheetjs.com/xlsx-0.20.3/xlsx-0.20.3.tgz",
     "xml2js": "^0.6.2",
@@ -88,7 +89,6 @@
     "supertest": "^7.2.2",
     "tsx": "^4.7.0",
     "typescript": "^5.3.3",
-    "uuid": "^9.0.1",
     "vitest": "^1.1.0"
   }
 }
--- a/packages/core-backend/package.json
+++ b/packages/core-backend/package.json
@@ -44,6 +44,7 @@
     "dingtalk-stream": "^2.1.5",
     "eventemitter3": "^5.0.1",
     "express": "^4.18.2",
+    "express-validator": "^7.2.1",
     "geoip-lite": "^1.4.10",
     "glob": "^10.3.10",
     "ioredis": "^5.3.0",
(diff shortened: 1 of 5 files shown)


zensgit:
Folded into #4126 @ `2a5871d75` (cherry-picked, Codex retained as commit author; branch rebased on main). Independently verified before folding: the same-file unguarded-import escape reproduces GREEN on the old head `27b4d4e29`, and all three mutations (same-file unguarded import / loader moved out of try / rethrowing catch) go RED on the new head — run against the real `ConfigService.ts`, not a fixture. Clean: guard 7/7, tsc PASS, 372 files / 5097 tests PASS.

zensgit:
EXACT_HEAD_UPDATE

exact_head_sha=4b7e71eb50391ef9250acd8552ebf430609c5d51
supersedes=67061896a3cf4b8c398e86a6248346fcbe40b9e3

Self-review tightened the exemption shape once more: a catch is now accepted only when it is empty or performs simple local-identifier constant fallback assignments; calls such as process.exit(1), explicit rethrows, and non-empty finally blocks are rejected.

Verification on the replacement head:
- focused guard 7/7 PASS
- core-backend typecheck PASS
- full unit retry 372/372 files, 5097/5097 tests PASS
- same-file extra import mutation RED
- loader moved outside try/catch mutation RED
- synthetic exit/rethrow/finally cases RED
- diff check PASS

Residual test note: one intermediate full-suite run produced an unrelated approvals-bridge route 405-vs-400 failure; that file immediately passed 27/27 focused, and the next full-suite run passed 5097/5097 on the same tree. No approval runtime/test files are changed here.

PR #4141, opened 2026-07-12
test(core-backend): prove optional loader degradation

zensgit:
## What

Stacked corrective on parent #4126 exact head `2a5871d75ba98cd7cdd0b11ad6e9cb37448960d2`.

The one-file startup-contract change now makes every soft-dependency exemption prove both structure and behavior:

- require one synchronous top-level `require()` assigned directly to one module-local fallback binding;
- require an empty catch or literal fallback assignment to that same binding;
- reject calls, rethrows, other catch targets, dynamic `import()`, nested/control-flow loader shapes, and non-empty finally blocks;
- reject any later top-level read of the loader binding;
- spawn an isolated module-load probe for every exemption with that dependency forced to `MODULE_NOT_FOUND`, using a minimal environment allowlist and values-free failures.

Only `packages/core-backend/tests/unit/runtime-dependency-classification.test.ts` changes. No runtime or package files change.

## Why

The parent basic shape lock still accepted `catch { process.exit(1) }`. The first #4141 head closed that path, but self-review found a second real-file escape: keep the catch empty and add `if (yaml === null) process.exit(1)` immediately after it. Exact head `76ba1c185` remained 7/7 green even though missing `js-yaml` terminated module evaluation.

The replacement closes that escape twice: the static binding-use lock rejects it, and the actual missing-module child-process probe exits non-zero. The runtime probe also covers future indirection that a catch-only AST predicate cannot prove safe.

## Verification at exact head `0903a22535a16c5123060e447babceb388eef35d`

- clean focused guard: 8/8 PASS on Node 24 and Node 20.20.2
- core-backend `tsc --noEmit`: PASS
- full core-backend unit suite: 372/372 files, 5098/5098 tests PASS
- real `ConfigService.ts` post-catch exit mutation:
  - superseded head `76ba1c185`: 7/7 GREEN
  - replacement: RED in both shape lock and missing-module runtime probe
- original in-catch `process.exit(1)` mutation: RED
- mutation restored; `git diff --check`: PASS

## Integration boundary

This PR targets the parent feature branch. Parent #4126 remains the package/merge unit for #4101 and needs exact-head review after integration. No package build, entity action, deploy, or self-merge is performed here.

### DIFF
--- a/packages/core-backend/package.json
+++ b/packages/core-backend/package.json
@@ -65,6 +65,7 @@
     "sanitize-html": "^2.13.0",
     "semver": "^7.7.3",
     "socket.io": "^4.6.1",
+    "uuid": "^9.0.1",
     "winston": "^3.17.0",
     "xlsx": "https://cdn.sheetjs.com/xlsx-0.20.3/xlsx-0.20.3.tgz",
     "xml2js": "^0.6.2",
@@ -88,7 +89,6 @@
     "supertest": "^7.2.2",
     "tsx": "^4.7.0",
     "typescript": "^5.3.3",
-    "uuid": "^9.0.1",
     "vitest": "^1.1.0"
   }
 }
--- a/packages/core-backend/package.json
+++ b/packages/core-backend/package.json
@@ -44,6 +44,7 @@
     "dingtalk-stream": "^2.1.5",
     "eventemitter3": "^5.0.1",
     "express": "^4.18.2",
+    "express-validator": "^7.2.1",
     "geoip-lite": "^1.4.10",
     "glob": "^10.3.10",
     "ioredis": "^5.3.0",
--- a/packages/core-backend/src/middleware/validation.ts
+++ b/packages/core-backend/src/middleware/validation.ts
@@ -19,27 +19,30 @@ interface Result {
   array(): ValidationError[]
 }
 
-// Dynamic import for optional express-validator
-let validationResult: ((req: Request) => Result) | null = null
-
+// express-validator is a declared PRODUCTION dependency and request validation is a security
+// control — load it FAIL-CLOSED (#4126 review). The previous try/catch fallback left
+// `validationResult` null when the module was absent and `validate` then called next() unchecked,
+// so every declarative chain on the workflow / workflow-designer / PLM routes was silently
+// disabled. A missing module is now a loud boot failure, never a silent open door.
+let validationResult: (req: Request) => Result
 try {
   // eslint-disable-next-line @typescript-eslint/no-require-imports, @typescript-eslint/no-var-requires
   const expressValidator = require('express-validator')
   validationResult = expressValidator.validationResult
-} catch {
-  // express-validator not installed - validation will be skipped
+  if (typeof validationResult !== 'function') {
+    throw new Error('express-validator did not expose validationResult')
+  }
+} catch (error) {
+  throw new Error(
+    'express-validator is a required production dependency (request validation must never silently ' +
+      `degrade to pass-through). Install dependencies and retry. Underlying: ${(error as Error)?.message ?? 'unknown'}`
+  )
 }
 
 /**
  * Validation middleware that checks for express-validator errors
  */
 export const validate = (req: Request, res: Response, next: NextFunction) => {
-  if (!validationResult) {
-    // express-validator not installed, skip validation
-    next()
-    return
-  }
-
   const errors = validationResult(req)
   if (!errors.isEmpty()) {
     return res.status(400).json({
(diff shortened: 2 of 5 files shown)


zensgit:
EXACT_HEAD_VERIFICATION

exact_head_sha=76ba1c1858ecddaccb9e0e5a8f6686863e67c821
parent_base_sha=2a5871d75ba98cd7cdd0b11ad6e9cb37448960d2

- focused guard 7/7 PASS
- core-backend typecheck PASS
- full unit 372/372 files, 5097/5097 tests PASS
- current-parent real ConfigService process.exit(1) catch mutation: GREEN before / RED after
- mutation restored; diff check PASS
- GitHub pr-validate PASS

This is a one-file test-only stacked delta. Holding integration for the parent owner; no self-merge or package/runtime action.

zensgit:
BOT_REVIEW_RESPONSE

Both suggestions are intentionally not applied on this corrective head. This guard is an explicit, fail-closed allowlist for the two current production loader shapes, not a general-purpose catch-effect analyzer.

- Neither current exemption uses a signed numeric fallback; rejecting a future safe-but-unknown shape creates a review-time false positive, not a production escape. The allowlist can be extended with a focused regression when such a real loader is introduced.
- Property assignment is deliberately excluded: `object.property = null` may execute a setter/proxy and `this.property` requires additional receiver and execution-boundary reasoning. Accepting it generically would widen the exact side-effect gap this corrective closes.

Current accepted shapes remain only an empty catch or simple local-identifier assignment to a constant literal fallback. The real `ConfigService` `process.exit(1)` mutation is GREEN on parent `2a5871d75` and RED on exact head `76ba1c1858ecddaccb9e0e5a8f6686863e67c821`; clean focused/full verification remains green. No scope expansion is warranted for #4126 corrective-6.

zensgit:
REPLACEMENT_EXACT_HEAD

exactHead=1e9e61ea53173005e791614f3f04a78424e93fe6
supersedes=76ba1c1858ecddaccb9e0e5a8f6686863e67c821
parentBase=2a5871d75ba98cd7cdd0b11ad6e9cb37448960d2
changedFiles=1 test-only

Self-review reproduced a second escape on the superseded head: keep the real `ConfigService` catch empty and add `if (yaml === null) process.exit(1)` after it. The old guard remained 7/7 GREEN although missing `js-yaml` terminated module evaluation.

The replacement head binds an exemption to one synchronous top-level require assignment and its module-local fallback binding, rejects later top-level use/dynamic import/other catch targets, and adds an isolated missing-module load probe for every exemption. The exact mutation is now RED in both the AST shape lock and runtime probe.

Verification: focused 8/8 PASS; typecheck PASS; full unit 372/372 files and 5098/5098 tests PASS; pr-validate PASS; mutation restored; diff check PASS.

This exact head is CLEAN / MERGEABLE against the parent. No self-merge, package build, entity action, or runtime mutation was performed.

zensgit:
NODE20_COMPATIBILITY

exactHead=1e9e61ea53173005e791614f3f04a78424e93fe6
nodeVersion=v20.20.2
focusedGuard=8/8 PASS
missingModuleSubprocessProbe=PASS for both exemptions

This directly verifies the new `node --import tsx` child probe under the required Node 20 runtime, not only the local default Node 24 runtime. Node 18 remains covered when the parent integrates this commit and reruns its required matrix.

zensgit:
FINAL_HEAD_CORRECTION

finalExactHead=0903a22535a16c5123060e447babceb388eef35d
supersedes=1e9e61ea53173005e791614f3f04a78424e93fe6

The only delta is test-harness confidentiality hardening: the missing-module child process now receives a minimal HOME/NODE_ENV/PATH plus probe inputs instead of inheriting the full environment, and failure assertions report only status/signal rather than raw stderr. Guard semantics and production scope are unchanged.

Reverification on the final head: focused 8/8 PASS on Node 24 and Node 20.20.2; typecheck PASS; full unit 372/372 files and 5098/5098 tests PASS; diff check PASS. Fresh pr-validate is running. Use `0903a22535a16c5123060e447babceb388eef35d` for integration.

zensgit:
FINAL_CI_COMPLETE

exactHead=0903a22535a16c5123060e447babceb388eef35d
prValidate=SUCCESS
mergeable=MERGEABLE
mergeState=CLEAN

This supersedes the prior comment wording that the fresh check was still running.

zensgit:
队列:备料线收账后的**可选**质量硬化项(corrective-6 guard governance lane,不阻塞 runtime PASS)。作为质量债单独处理。

zensgit:
**Correction (口径修正):** record this as **BLOCKED / DIRTY**, not ordinary-queued. mergeStateStatus=DIRTY (merge conflict) and the live diff is **5 files** (package.json, validation.ts, validator.ts, the unit test, pnpm-lock.yaml) — the body's "single test file" is stale. Needs a **branch reconstruction** before it can re-queue; it is not a clean pick-up.

PR #4168, opened 2026-07-12
feat(multitable): D-2 side-door delete recoverability — trash + tombstone parity for plugin-SDK / automation deletes (flag default OFF)

zensgit:
Implements the **D-2 side-door delete recoverability** design-lock (#4004, owner-ratified). Every owner decision (OD-1=(a), default-off flag, scope=paths 3+4 only, `deleted_by` split, typed SDK error, OD-7 txn contract, target-base semantics, §1.11 truth table) is implemented as ratified.

> **Sequencing:** #4004 (the lock, with its errata) lands **first**; this PR lands after. **OD-3 carries NO retention implementation** — D-2 ships zero retention code (`git diff --name-only` touches no retention file). The trash-growth risk is accepted and documented, not mitigated: see *Honest consequences*.

> **Post-review hardening.** This PR was independently reviewed (`APPROVE-with-hardening`) and then the owner rejected the 0-P1 verdict and found a **real P1** the reviewer missed. All of it is fixed below: a **plugin lock TOCTOU**, the **OD-7 guard that guarded nothing**, the typed error's export, and O-2 docs that stated the *inverse* of the code.

## The problem

Records deleted via the **plugin-SDK** (`records.ts` path 3) or an **automation `delete_record` step** (path 4) got a D-1 delete revision but **no `meta_records_trash` row and no inbound-edge tombstones** — invisible to trash-restore and outside 4c-3's inbound-replay, i.e. **permanently irrecoverable**, while a UI-deleted record was fully recoverable. Paths 1 (UI) and 2 (PIT-reset) are **NO CHANGE** (path 2 was already at full parity via 4c-3 D-3).

## Flag posture

`MULTITABLE_SIDE_DOOR_DELETE_TRASH_ENABLED`, **default OFF**. **No env flag is enabled anywhere in this PR.** Flag-off is byte-identical to today's D-1 behavior on both lanes, each keeping its **existing** ordering (plugin: delete-then-emit fail-safe; automation: emit-then-delete inside `withTransaction`). On the plugin lane byte-identity is guaranteed **by construction**: the flag-off branch is the D-1 code *verbatim*, and the recoverability path is a separate function.

## §1.11 truth table — every row golden-pinned, for BOTH lanes

| `SIDE_DOOR` | `CAPTURE` | trash schema | Behavior on paths 3+4 | Golden (each × plugin **and** automation) |
|---|---|---|---|---|
| off | off | any | byte-identical D-1 (revision-only) | `G6a-1` |
| off | **on** | any | byte-identical D-1 — **zero** side-door tombstones | `G6a-2` ← the nesting golden |
| **on** | off | present | trash + `delete_revision_id` anchor, zero tombstones; restore ⇒ `recoverable=false`, zero replay | `G6b/G1`, `G6b-restore` |
| **on** | **on** | present | full UI parity: trash + anchor + cap-checked inbound tombstones | `G1/G2`, `G4` |
| **on** | any | **missing (42P01/42703)** | **fail-closed: delete refused**, record/links/revisions intact | `G11` (×2 error codes) |

Plus: `G11-inert` (flag-off never touches the schema hole ⇒ the pre-migration window is inert **by construction**), `G3a/G3b` (txn), `G5` (PIT-resurrect — below), `G7` (cap), `G8` (retention floor), `G12` (PIT invariance), `OD-8` (cross-base), `G10` (write-own visibility), plus the flag-ON missing-record contract.

**31 goldens in the new real-DB suite + 8 in the flipped D-1 suite = 39, all green.**

### G5 — the second resurrection surface (§2 same-PR obligation)

`G4` covers `restoreRecord` (the recycle-bin surface, anchored on `meta_records_trash.delete_revision_id`). The **PIT revert route is a second, independent resurrection surface**: it has no trash row to read, so it derives the anchor from the revert's `asOf` T. The lock demanded this be **proven, not assumed** — and the reason turned out to be concrete: that anchor query selects on `action='delete' AND created_at > T` and **does not filter on `source`**, so a side-door delete revision is eligible; and because D-2 **pre-generates** that revision's id, the tombstones it captured hang off exactly that id. Had either half been untrue (a `source` filter, or a self-generated id), the resurrect surface would have silently replayed **nothing** for machine-deleted records while reporting success. `G5` drives the real route end-to-end (automation delete → `revert-preview` → `revert-execute confirm:'undelete'`) and asserts `undeleteInbound.replayed: 1` + the edge is back. Mutation-verified (reds under anchor-break and capture-disable).

**Correction to the lock:** §3 describes this as "the `:10183` latest-delete-revision heuristic". That description is **stale** — the heuristic was replaced by the T-derived anchor (R11 A′, ratified 2026-07-11) *after* the lock was drafted. The obligation is unchanged; it is discharged against the anchor as it actually is today.

Two real traps G5 had to clear (both recorded in-file, and worth knowing):
- `AutomationExecutor.execute` **overrides `context.sheetId` with `rule.sheetId`** — a same-base delete on a non-default sheet otherwise returns the lenient 0-row `success` with **no revision at all**.
- `revert-preview` round-trips `asOf` through `new Date().toISOString()` (**millisecond truncation**), so a microsecond-precision cutoff rounds *down* and the record reads as absent at T.

### §1.5 nesting (the trap)
Side-door capture requires **both** flags. The UI path's CAPTURE-only gating was deliberately **not** copied: doing so would silently start writing tombstones for every operator already running with capture on, the moment this deploys — breaking §1.9. The rule lives in exactly one place (`isSideDoorTombstoneCaptureEnabled`), and `G6a-2` reds if it's un-nested (mutation M4/M10).

### §1.8 fail-closed (the asymmetry)
The trash INSERT carries **no** 42P01/42703 try/catch — *the absence of the guard is the feature*. An operator who opted into recoverability must never get a silently-unrecoverable delete. This is deliberately **asymmetric** to paths 1/2, which keep their never-fail degradation (they predate the flag and must not start failing on deploy ordering). Ratified in the lock; pinned by `G11`.

## P1 (owner-found) — plugin lock TOCTOU, closed

`guardRecordNotLockedForPlugin` (`records.ts`) pre-checks the lock with a plain `SELECT locked, locked_by, created_by` — **no `FOR UPDATE`, so no row lock is held**. My flag-on `FOR UPDATE` selected only `data, version, created_by, created_at, updated_at` and **never re-read lock state**. A concurrent txn committing a lock in that window would have had its **locked record deleted** — a rank-8 violation.

**Fix (plugin lane only):** the `FOR UPDATE` now also selects `locked, locked_by`, and `ensureRecordNotLocked` re-verifies **on that same row-locked result**, before any capture / links DELETE / revision / trash / record DELETE. The pre-check remains a fast path; it is no longer the authority. This mirrors the automation lane's existing shape — **no new mechanism, and the automation lane's source is unchanged** (it already read the lock columns from its own `FOR UPDATE` and re-checked; building a second check there would have been divergent logic on a destructive path).

- **`G14` [plugin]** reproduces the real race: a second connection **commits** `locked = true` in the window between the pre-check and the `FOR UPDATE`. Asserts the delete is refused and record + links + revisions are intact (and that the interleave actually happened — not a vacuous pass).
- **`G17` [automation]** pins the already-correct lane under real concurrency: connection A holds the row lock, sets `locked = true`, and commits only **after** the executor is already blocked on that row — so the executor must re-read the freshly committed state and refuse.

## OD-7 transaction contract — now three layers, all pinned

The lock's OD-7 claimed *"the flag-on reordered path's atomicity is CI-guarded by G3."* **That was false.** G3 doesn't drive the production wiring — it **re-implements** it (the goldens wrapped the call in their own transaction and supplied their own `transaction` dep). The reviewer proved it by mutating **production source**, with all 39 goldens still green.

| Layer | What it proves | Pinned by |
|---|---|---|
| **1** | The sequence is atomic **given** a transaction | `G3a/G3b` (D1-5b BEFORE-DELETE trigger; asserts link intact + 0 revisions + 0 tombstones + 0 trash — not merely "record exists") |
| **2** | The **real entry points actually supply** one | `multitable-d2-sidedoor-txn-wiring.guard.test.ts` (structural, on production source) **+ `G15`**, a real-DB golden that drives `new AutomationService(...).exec` — production's own deps, nothing re-implemented |
| **3** | Everything else **refuses to run** without one | `assertTransactionalQuery` — and it is itself pinned by **`G16`** (both lanes) |

**OD-6 public contract (`G19`):** both typed errors (`MultitableRecordDeleteCapExceededError`, `MultitableSideDoorDeleteNonTransactionalError`) are now imported **from the package root** — the way a plugin author would — and thrown for real (a real cap breach; a real non-transactional call). Previously deleting both root exports left `tsc` clean and every test green.

**Layer 3** is a runtime refusal: `txid_current()` twice — same id iff one transaction. It is **not** "un-removable" (any guard can be deleted); its value is that it is **independent of the entry wiring** — it holds for callers that don't exist yet, and for a caller whose transaction a distant refactor dropped. `G16` invokes the reordered path with a genuinely non-transactional query (the bare pool) and asserts it **throws and writes nothing**; deleting the guard reds it.

### Acceptance: production-source probes — all RED

| Probe | Production-source mutation | Guarded by | Result |
|---|---|---|---|
| **PROBE-1** | `index.ts` — unwrap the SDK `deleteRecord`'s transaction | **G18** (behavioral) | **RED** |
| **DEFEAT-1** | PROBE-1 **+ a transactional sibling key after `deleteRecord`** (`restoreRecord`) | **G18** (behavioral) | **RED** |
| **PROBE-2** | `automation-service.ts` — drop `deps.transaction` | **G15** (behavioral) + exact-string tripwire | **RED** |

**G18 drives the ACTUAL SDK factory** — `new MetaSheetServer({ port: 0, host: '127.0.0.1', pluginDirs: [] })` → `createCoreAPI().multitable.records.deleteRecord` → asserts the record is trashed, the anchor matches, the tombstone is captured. Nothing is re-implemented.

**Two corrections to what I previously reported, stated plainly:**

1. **My structural PROBE-1 guard was defeatable, and its comment was false.** It sliced a fixed 1200-char window and claimed it stopped "at the next sibling key". It passed only by *positional accident* — `deleteRecord` is currently the last key in the `records` object, so the window spilled into `auth:`, which contains no `transaction(`. Adding any transactional sibling after it (`restoreRecord` — literally the next thing this recoverability line would add) made it go **green with the destructive SDK path in no transaction at all**. It has been **deleted**, not patched; a behavioral golden cannot be fooled by source layout.
2. **"Booting the server hangs in-process" was wrong.** It hung because my throwaway probe **omitted `pluginDirs: []`**, so it scanned the real plugin directories. Three existing unit tests construct the server exactly this way and pass in CI. The lane is now driven for real.

## Mutation results (23/23 — every guard made red)

| # | Mutation | Reds |
|---|---|---|
| M1 | plugin: trash INSERT removed | G1/G6b, G4, G6b-restore, G3b, G11×2, G8, G10 `[plugin]` |
| M2 | automation: trash INSERT removed | same set `[automation]` + OD-8 |
| M3 | revision id self-generated (anchor broken) | G6b/G1 + G1/G2, **both lanes** |
| M4 | capture un-nested from `SIDE_DOOR` | G6a-2 `[automation]` |
| M5 | cap assert neutered | G7 **both lanes** |
| M6 | trash INSERT swallows 42P01/42703 (fail-closed defeated) | G11 ×4 |
| M7 | plugin: capture moved AFTER the links DELETE | G1/G2, G4, G7, G8 `[plugin]` |
| M8 | plugin: recoverability path taken with flag OFF (§1.9 broken) | G6a-1, G6a-2, G11-inert `[plugin]` + **both flipped D-1 goldens** |
| M9 | automation: trash written with flag OFF (§1.9 broken) | G6a-1, G6a-2, G11-inert `[automation]` + flipped D-1 automation golden |
| M10 | capture gated on `CAPTURE` alone (UI path's gating, copied) | G6a-2 `[automation]` |
| M11 | automation: `deps.transaction` removed (non-txn `queryFn`) | G3a, G3b, G11×2 `[automation]` |
| M12 | plugin: SDK call not wrapped in a txn (the OD-7 regression) | G3a, G3b, G11×2 `[plugin]` |
| M-G5a | automation revision id self-generated (resurrect anchor broken) | **G5** |
| M-G5b | side-door capture disabled | **G5** |
| M-P1a | plugin: the under-row-lock `ensureRecordNotLocked` re-check removed | **G14** (the locked record gets deleted) |
| M-P1b | plugin: lock columns dropped from the `FOR UPDATE` | **G14** |
| M-L3a | plugin: `assertTransactionalQuery` call removed | **G16 [plugin]** |
| M-L3b | automation: `assertTransactionalQuery` call removed | **G16 [automation]** |
| M-L3c | automation: its **existing** `ensureRecordNotLocked` removed (non-vacuity of G17) | **G17** |
| PROBE-1 | production `index.ts` txn wrapper unwrapped | **G18** (behavioral) |
| DEFEAT-1 | PROBE-1 + a transactional sibling key after `deleteRecord` | **G18** — the case that defeated the old structural guard |
| PROBE-2 | production `automation-service.ts` `deps.transaction` dropped | **G15** + tripwire |
| M-OD6 | both typed errors removed from the package-root exports | **G19** |

M4 reds only the automation lane — **by design**: the plugin lane's byte-identity comes from a *hard branch* (flag-off never reaches the capture call at all), so M8 is the mutation that proves its flag-off goldens are load-bearing. Both are covered.

## The flipped golden (same-PR obligation)

`multitable-d1-delete-revision-parity-realdb.test.ts` asserted the **opposite** of D-2 (no-trash/no-tombstone). Its name said "neither path" but its **body exercised the plugin lane only**. It is now a per-lane flag-off byte-identity guard: explicit **plugin** + **automation** coverage, a **nesting** case (capture-on alone ⇒ still zero tombstones), and it now also asserts the D-1 revision **is** still written ("no trash" must never quietly become "no revision"). Flag-on parity lives in the new D-2 suite.

## Regression evidence

- **Unit: 5206/5206 pass (376 files).** Two structural guards legitimately caught the new code and were fixed properly, not silenced:
  - `multitable-record-lock-guard.guard.test.ts` — the new `DELETE FROM meta_records` needed a lock disposition within 3 lines. It **is** guarded (the same `guardRecordNotLockedForPlugin` rejects before either branch dispatches); marker added.
  - `multitable-richtext-longtext-write-sink.guard.test.ts` — the new module is caught by `writesRecordData` only because its sole INSERT targets `meta_records_TRASH`, whose name shares the `meta_records` prefix the regex matches. Added to the allowlist as **SAFE** with that exact reasoning (it writes no `meta_records` row and adds no new user-content ingress).
- **Multitable real-DB CI allowlist (187 files incl. the new suite): 186/187.** The single failure is a **pre-existing shared-DB batch flake**, not a D-2 regression: it reproduces on `origin/main` in a clean baseline worktree, it is a **different file each run** (`conditional-rule-enforce` once, `ai-bulk-job` on both main's baseline and a re-run), and every such file **passes in isolation**. The D-2 suite passed in every run.
- **Neighbouring surfaces re-run green:** `multitable-undelete-inbound-resurrect-realdb` + `multitable-undelete-inbound-replay-realdb` alongside the D-2 and D-1 suites — **63/63** (D-2 touches the anchor those surfaces consume, so they are the ones most exposed to it).
- **All `automation-*` / `multitable-*` integration suites pass.** `attendance-plugin.test.ts` fails identically on `origin/main` (14 failures in a clean baseline) — pre-existing, unrelated, swept in only by a `grep plugin` glob.
- `npx tsc --noEmit` **clean** (verified `tsc` is real — 5.8.3 from installed `node_modules`, proven by a decoy-probe that it does report errors; a bare `npx tsc` in a worktree without `node_modules` silently compiles nothing).
- **Two-point wiring** verified: `vitest.config.ts` exclude **and** `plugin-tests.yml` whitelist. The no-DB lane prints **`No test files found`** (not "skipped") — a `describeIfDatabase` alone would skip-green and never run.

## Scope discipline

- **No migration** (§7: `delete_revision_id` and both tombstone tables already exist).
- **Restore-side zero delta** — no change to `listDeletedRecords` / `restoreRecord` / `replayInboundLinks` / retention code. D-2 trash rows are source-agnostic and flow through the existing machinery (`G4` proves it end-to-end; `G8` proves the retention floor already protects side-door anchors).
- **No PIT/read-path delta** (`G12`: identical `reconstructRecordsAtT` output flag-off vs flag-on).
- **4d red line untouched.** No trash-retention sweep (OD-3 stays a separate future rung).

## Two windows that remain OPEN — inherited, not regressions of this PR (owner decision)

Re-review surfaced **two pre-existing robustness gaps** that this PR deliberately does **not** close:

- **P3-1** — a concurrency window on the **legacy (flag-off) plugin path**. Not introduced by D-2; **§1.9
  forbids touching that path here** (any change breaks the byte-identity guarantee this PR is built on).
  It wants its own rung.
- **P3-2** — a durability gap in inbound-edge capture that **already exists on the ratified UI path** (path 1).
  D-2 inherits it rather than introducing it; closing it is a design change beyond this lock.

Neither is a regression of this PR and neither is buried. **Technical detail is withheld from this public
thread and has been reported to the maintainer directly**, per the repo's coordinated-disclosure posture —
each will be tracked on its own rung.

## Honest consequences (declared, not hidden)

- **Product-semantics change, flag-on only:** automation/plugin deletes become restorable by anyone with `canDeleteRecord`; restore re-fires events and can re-trigger automations. A plugin that deleted a record for policy reasons can have it resurrected behind its back.
- **A previously-succeeding SDK delete can now fail** on the cap (fail-closed) — surfaced as the typed `MultitableRecordDeleteCapExceededError` (OD-6), reachable only with both flags on.
- **Write-own visibility asymmetry:** machine-created records carry `created_by = NULL`, so their trash rows are invisible/unrestorable to write-own actors. `G10` pins this; per the lock it is **documented, not fixed**, in D-2.
- **Trash growth is ACCEPTED, not mitigated (OD-3).** With the flag ON, `meta_records_trash` becomes machine-written and **grows monotonically — no sweep removes it** (its only DELETE is a successful restore), and each live trash row **floor-pins its tombstone group indefinitely**. D-2 ships **zero retention code**; retention is a separate future rung. Recorded in the O-2 ladder with a monitoring query (`SELECT count(*) FROM meta_records_trash`, esp. `deleted_by IS NULL` plugin rows) and gated behind the L3.5 rung.

## O-2 / operator doc sweep — docs that stated the INVERSE of the code

Four docs still said side-door deletes have *"no capture"* and that D-2 is *"owner-gated"* / unimplemented — the #4074 class of defect (a doc that contradicts the code):

- **O-2 operator flag ladder §5** — corrected to as-built: the capability **exists**, nested under `MULTITABLE_SIDE_DOOR_DELETE_TRASH_ENABLED` (default OFF) **and** the capture flag, so **production behavior is unchanged** — a *flag state*, not a missing capability.
- **D-1 design-lock** (2 sentences) — marked as flag-state, with the flag-off byte-identity caveat.
- **Live time-machine tracker (2026-07-12)** — status updated to ratified + implemented, production OFF.
- **R11 round-closeout** — a **point-in-time historical record**; *not* rewritten. It gets a dated forward-pointer instead, so a reader can't be misled.

## Lock errata for #4004 (owner is folding these in; recorded here for the record)

1. **§3/G5's anchor description is stale.** It cites a *":10183 latest-delete-revision heuristic"*. The live anchor is `univer-meta.ts:10196-10201` (**R11 A′**, T-derived): first row with `created_at > T`, ordered `created_at ASC, version ASC, id ASC`, and **no `source` filter** — which is exactly *why* a side-door (`source:'automation'`) revision is eligible, and why pre-generating the revision id makes its captured tombstones reachable. G5 is discharged against the anchor **as it actually is**.
2. **OD-7's "G3 is the CI guard" claim was false** (proved by PROBE-1/PROBE-2). It is now true only because of the **three-layer** structure above — G3 alone never guarded the wiring.

## §6 doc sweep (ships here, as the lock requires)

4c-2 lock (C2 coverage 2/4 → 4/4, with the exact nesting caveat) · destruction-path gap audit (§1 table rows 2-4 + the D-2 menu entry marked landed) · 4c-3 lock (§0 reachable boundary widened) · **O-2 operator flag ladder** (the new flag, three new coupling constraints — nesting, schema fail-closed/deploy ordering, trash-growth warning — and a new **L3.5** rung gated on explicit owner confirmation before prod).

**The D-2 design-lock doc itself is NOT duplicated here** — it is #4004's payload and that PR is still open. Merging this PR does not require #4004 to land first, but the lock should land for the doc references to resolve.

---


zensgit:
## Independent adversarial review — **MERGE_CLEAN** (1× P2, comment-only)

Reviewed head **`27fec34d1`** against the ratified D-2 lock, refute-first. Disposable worktree + my own
Docker Postgres 15 (port 55771), fresh DB, full migration chain with CI's `MIGRATION_EXCLUDE`.
**I did not take the PR body's mutation table on faith — I re-ran every load-bearing mutation myself
against production source.**

### Empirical

- **Baseline**: D-2 (36) + flipped D-1 (8) + 2 neighbours = **68/68 real-DB green**; unit guards **18/18**.
  Neighbours green ⇒ environment healthy. Post-revert re-run: 68/68 + 18/18, tree pristine.
- **11 production-source mutations → 11 KILLED, 0 survivors**, including the ones that matter most:
  - **PROBE-1** (unwrap the *real* `poolManager.get().transaction` in `index.ts`) → wiring guard **RED**.
  - **PROBE-2** (drop `deps.transaction` in `automation-service.ts`) → wiring guard **RED** *and* **`G15` RED**
    (a real-DB golden driving `new AutomationService(...).exec` — behavioural, not source-text).
  - **M-L3a/M-L3b** (delete `assertTransactionalQuery`) → `G16` RED on the mutated lane. Layer 3 genuinely refuses.
  - **M-P1a** (delete the under-row-lock `ensureRecordNotLocked`) → `G14` RED. **`G14` is not vacuous**: I checked
    *how* it forces the race — a second connection **COMMITs** `locked=true` strictly between the pre-check and the
    `FOR UPDATE`. Constructed interleaving, not sequential argument. The owner-found P1 is genuinely closed.
  - **M8/M9** (take the recoverability path with the flag OFF) → **13 RED across both suites**. §1.9 is pinned, not asserted.
  - **M5** (cap) → `G7` RED both lanes. **M6** (swallow 42P01/42703) → `G11` RED ×4. **M1/M2** (trash INSERT) → **18 RED**.
- **Positive control present** — `G4`/`G15` prove the *legitimate* path SUCCEEDS (`inbound.replayed >= 1`), so this is
  not an all-fail-closed suite that would stay green if the guard refused everyone.

### Lock gates: OD-1..OD-8 all **PASS**

- **OD-2 (owner red line) — CLEAN.** The flag name appears **only** in comments/docs plus the single reader
  (`side-door-delete-trash.ts:117`). Nothing in `.env*`, `docker*`, or any `.github/*` env block enables it. The
  goldens set it in `process.env` per-test and **restore** in `afterEach`+`afterAll` — no leak into the flag-off legs.
- **OD-3 / OD-4** — `git diff --stat` over `*retention*`/`*sweep*` and over `record-service.ts`/`univer-meta.ts` is
  **empty**. No retention smuggled in; paths 1/2 untouched.
- **§1.9 byte-identity — verified by DIFF, not prose** (mutations cannot find a flag-off delta; only the diff can):
  `records.ts`'s flag-off branch changed **comments only**, and `guardRecordNotLockedForPlugin` is **pre-existing
  context**, so flag-off rejects exactly what it rejected before. Automation's only flag-off delta is two extra
  *columns* on an existing `FOR UPDATE` (zero extra queries).
- **Two-point CI wiring — COLLECTED AND RUN, not merely triggered.** `test (20.x)` *is* a required check and *is* the
  `plugin-tests.yml` job; the spec is in the real-DB step's file list **and** excluded from `vitest.config.ts` so the
  no-DB lane prints "no test files" instead of skip-green. Suite also carries a `DATABASE_URL` fail-not-skip sentinel.
- **No migration** ⇒ the `zzzz-` trap is N/A. Verified on a **fresh** DB that `delete_revision_id` is present after the
  full chain. Debug-residue removal in `27fec34d1` broke nothing (`git grep probe-server` = empty).

### Findings

**P1: none.** (I hunted the highest-probability one — a hidden flag-off behavioural delta in `records.ts` that would
have falsified the "byte-identical by construction" claim. The diff refutes it.)

**P2-1 — `records.ts:707` ships the claim the lock's own errata declares FALSE.**
The `TRANSACTION CONTRACT` doc-comment still says:
> *"Golden G3 is the CI guard — a non-transactional caller turns it red."*

The lock's OD-7 errata retracts exactly this (*"OD-7's CI-guard claim was FALSE as written"*), and this PR's body says
*"That was false."* The PR built three layers **because** it is false — then shipped the false claim as the normative
comment on the destructive hot function. Echoed at `side-door-delete-trash.ts:45-46`.

*Failure scenario*: a maintainer refactoring `index.ts` reads this comment, unwraps the transaction, sees **G3 green**
(it will be — G3 supplies its own txn), and ships. The comment names the one guard that does **not** catch this and
names **neither** of the two that do. It misdirects toward the exact regression that already happened once on this line
— the #4074 defect class this PR is fixing in four *other* docs.

*Fix (two lines)*: state the three layers truthfully, and refresh the stale `index.ts:634-653` refs (`records.ts:704`,
`side-door-delete-trash.ts:44`) — on this head the SDK `deleteRecord` is at **`index.ts:651-670`**.

Not blocking: zero runtime effect, and all three layers empirically bite.

**P3-1 — OD-7 Layer 2 on the *plugin* lane is a source-text regex, not a behavioural drive** (`…txn-wiring.guard.test.ts:58-75`).
The lock asks Layer 2 to *drive the ACTUAL factory*; the PR asserts a regex over a 1200-char window of `index.ts` and
honestly says why (booting `createCoreAPI()` hangs in-process). Per repo doctrine, source-text guards can be fooled —
`indexOf` takes the **first** `deleteRecord: async ({ sheetId, recordId })`, so a future sibling with the same signature
declared earlier would silently re-point it. **Not blocking**: it is not vacuous today (PROBE-1 reds it for real, which
*is* the lock's stated acceptance criterion), the automation lane's Layer 2 *is* behavioural (`G15`), and the plugin lane
is covered behaviourally by Layer 3 + `G16`. Worth an O-2 ladder note, not a change here.

**NIT-1** `automation-executor.ts:2330` — `randomUUID()` computed unconditionally on the flag-off path (dead work; no DB effect).
**NIT-2** PR body says 31+8=39 goldens; measured **36+8=44**. Correct the body to match what CI prints.
**NIT-3** The `richtext-longtext-write-sink` allowlist entry is **required collateral, not scope creep** — the guard's
regex matches `meta_records_TRASH` on the `meta_records` prefix. The reasoning is correct: the module writes **no**
`meta_records` row, and the `data` it copies is the deleted row's own snapshot read verbatim in-txn, i.e. content that
already passed the ingress chokepoint. Restore-side is unchanged. **No new user-content ingress.**

---

Full review (gate-by-gate table, mutation log, exact commands): `/tmp/pr4168-review-claude-20260712.md`

*Review only — nothing merged, no auto-merge armed, no push to the branch.*

zensgit:
## Re-gate on head `0bab16fbe` — **MERGE_CLEAN**

The branch moved twice after my first gate (`27fec34d1`). Re-gated from scratch: fresh worktree, fresh Docker
Postgres, fresh migrated DB. **Both new commits are improvements. No P1. No regression.** One new P3 (a
characterization, not a defect).

### Does my prior 11/11 mutation proof still hold? — proven, not assumed

- `843c9fd33` touches **zero** `src/` files (tests only).
- `0bab16fbe` — **every** changed line begins with ` * ` (JSDoc). Decisive check: strip all comments/blanks from
  `records.ts` + `side-door-delete-trash.ts` at both revs and diff the remaining code ⇒ **IDENTICAL CODE**.

⇒ Production runtime bytes are **identical** to the head I gated 11/11. The mutation proof and the §1.9 flag-off
byte-identity finding hold verbatim. Re-ran **M8** on the new head anyway: **7 RED**. ✅
**@coordinator: your "zero runtime bytes" claim is CONFIRMED — no smuggled byte.**

### G18 — verified, not trusted. And the "defeated" claim was TRUE.

Baseline on `0bab16fbe`: **70/70 real-DB** (D-2 now 38: +G18 +G19), **17/17 unit**, `tsc` clean.

| Mutation | Old structural guard | **New G18** |
|---|---|---|
| **PROBE-1** (unwrap the real `poolManager.get().transaction` at `index.ts:651`) | RED | **RED ✅** |
| **DEFEAT-1** (unwrap **+** transactional sibling key after `deleteRecord`) | **GREEN — DEFEATED** ❌ | **RED ✅** |

I constructed DEFEAT-1 myself (added a `restoreRecord` sibling carrying its own `poolManager.get().transaction(`)
and **replayed the deleted guard's exact regex against the mutated source**: it returns **GREEN** — the old guard
would have passed with the destructive SDK path in **no transaction at all**. So **my P3-1 was understated**: I
called it "not vacuous today" because unwrap-only reddened it; the real hole was the window spilling *forward*
past `deleteRecord` (the last key in `records`). G18 closes it behaviourally. Good catch.

**G18 is not self-satisfying** — it does not construct its own transaction (the exact trap G3 falls into). It builds
the real `MetaSheetServer({ pluginDirs: [] })` → real `createCoreAPI()` → real SDK surface, so the *production
factory* must supply the transaction, and it asserts the whole anchor chain end-to-end.

### G19 closes a gap **I missed**

Deleting both package-root re-exports from `src/index.ts` ⇒ **G19 RED** ✅. In pass 1 I marked OD-6 PASS on
structural grounds (export exists, package `main` resolves) — correct, but **untested**: every other test imported
the classes from their *defining module*, so deleting the root export left `tsc` clean and everything green. G19
imports them as a plugin author does and throws both for real. That's a real gate I'd passed without a test behind it.

### Coverage drop from the −23/+20 rewrite: **none silently dropped**

Only the structural PROBE-1 test was removed, and G18 *replaces* it. Retained: the frozen `AutomationExecutor`
construction-site allowlist, all 4 Layer-1 runtime tests, all 4 §1.5 nesting tests. PROBE-2 is **strengthened** —
from a positional `slice(start, start+1500)` window to a **full-file exact-production-string** regex, so it cannot be
defeated the way PROBE-1 was. The file is honest that PROBE-2's real evidence is G15 and the regex is a tripwire.

---

### NEW **P3-2** — G18's Layer-2 evidence is *mediated by* Layer 3, not independent

**Evidence (double mutation, run for real):** unwrap the SDK transaction **AND** delete `assertTransactionalQuery`
from `records.ts` ⇒ **G18 goes GREEN.** With no txn *and* no refusal, every statement autocommits: record deleted,
trash + revision + tombstone all land, anchor matches — all G18 assertions pass. G18's RED under PROBE-1 comes
**entirely from Layer 3 throwing**; it never proves *rollback* through the real factory.

**Not blocking — there is no green path to production:**
- unwrap alone ⇒ **G18 RED** · remove Layer 3 alone ⇒ **G16 RED + G19 RED** · do **both** ⇒ G18 green but **G16 + G19 still RED**.

Every route to a non-transactional destructive delete reds something in the required job. It is also disclosed
honestly in G18's own docstring. **Suggested hardening (follow-up):** make G18 independent by asserting *rollback*
through the real factory — D1-5b BEFORE-DELETE trigger + drive `sdk.deleteRecord()` via `createCoreAPI()` + assert
nothing committed. That would survive the double mutation.

### `0bab16fbe` — did you name the right guards? **Yes.**

G18 reds under PROBE-1 (verified this pass) and G15 reds under PROBE-2 (verified pass 1). *"G3 ALONE DOES NOT GUARD
THE WIRING"* matches the lock's errata and my own finding. `index.ts:651` is the correct line. The comment does **not**
overclaim — it says G18 "reds if this call site's `transaction` is unwrapped", which is exactly true, and it doesn't
assert independence from Layer 3. Its warning — *"Do not 'fix' a red layer-2/3 by relaxing the guard; re-wire the
caller"* — is precisely the right mitigation for P3-2. **P2-1 RESOLVED; you did not repeat the defect you were fixing.**

### NIT-1 (`randomUUID()` on the automation flag-off path) — declining it was **CORRECT**

`automation-executor.ts` is HOT-CORE, and the §1.9 byte-identity proof rests on its runtime being untouched. Trading a
re-earned byte-identity proof on a destructive path for deleting one unused UUID is a bad trade. Leave it; fold it into
the next PR that legitimately opens that file.

---

**Tally across both passes: 15 production-source mutations — 14 KILLED**, plus the one deliberate independence probe
whose green result *is* the evidence for P3-2 (and is itself covered by G16+G19).

Gates: **OD-1..OD-8 all PASS** (OD-6 upgraded from untested→behaviourally tested; OD-7 Layer-2 plugin lane upgraded
from source-text→behavioural). §1.9, §1.11, CI two-point wiring, migrations: unchanged and PASS — G18/G19 live inside
the already-wired D-2 suite, so no new CI wiring is needed.

Full review: `/tmp/pr4168-review-claude-20260712.md`

*Review only — nothing merged, no auto-merge armed, no push to the branch.*

PR #4099, opened 2026-07-11
feat(web): multitable UI P2-1c batch6 — migrate Calendar/Gallery/Kanban clean surfaces to MtButton/MtIconButton

zensgit:
## Scope

UI-P2-1c Lane-A **batch6** — presentation-only, `<button>` → `MtButton`/`MtIconButton`, zero behavior
change. Migrates the sharer-clean, ungated bespoke buttons on three multitable view components:

| Component | Migrated | Left bespoke (excluded) |
|---|---|---|
| `MetaCalendarView.vue` | prev/next nav (→ `MtIconButton`, ghost) · "today" (→ `MtButton`, ghost) · field "change" (→ `MtButton`, ghost) | `create-btn` (quick-create) · `day-create` (day-view create) |
| `MetaGalleryView.vue` | pagination prev/next (→ `MtButton`, ghost) | `create-btn` (toolbar) · `empty-action` |
| `MetaKanbanView.vue` | header "clear" grouping (→ `MtButton`, ghost) | `header-add` ×2 · `add-btn` (per-column) |

All "domain" buttons (row/field comment-affordance chips, card controls) are untouched — out of scope
per the migration doctrine (`multitable-ui-p2-1c-completion-verification-20260707.md` §3).

## Byte-equivalence

`git diff origin/main` shows only tag swaps (`<button>` → `<MtButton>`/`<MtIconButton>`) plus a
`variant="ghost"` prop. Every preserved attribute is unchanged byte-for-byte:
- `@click`: `goPrevious` / `goNext` / `goToday` / `onResetDateField` (Calendar); the two
  `emit('go-to-page', currentPage ± 1)` expressions (Gallery); `onClearGroupField` (Kanban).
- `:disabled="currentPage <= 1"` / `:disabled="currentPage >= totalPages"` (Gallery pagination).
- `v-if="dateField"` on Calendar's change-btn.
- Nav glyphs `&lsaquo;`/`&rsaquo;` pass through `MtIconButton`'s default-slot icon fallback unchanged
  (confirmed via mount test: `textContent` is exactly `‹`/`›`).

No new hex: `git diff '^+' | grep -oE '#[0-9a-fA-F]{3,6}'` on the three touched components is empty.
Removed bespoke CSS blocks are replaced with a one-line comment pointing at the new primitive (matches
existing precedent, e.g. `MetaChartLoadError`/`ResetToPointPicker`).

**Sanctioned NIT** (documented, not hidden): each migrated button previously carried its own accent
[...]

PR #4089, opened 2026-07-11
feat(web): multitable UI P2-1c batch4 — HistoryCenterModal/ResetConfirmDialog/MetaAutomationLogViewer to MtButton

zensgit:
## Scope
UI-P2-1c Lane-A **batch4** — presentation-only, bespoke `<button>` → `MtButton` migration on 3 components (fixed technique, zero behavior change). Follows the刀法 in `docs/development/multitable-ui-p2-1c-completion-verification-20260707.md` §4/§5.

| Component | Migrated (line refs from batch4 spec) | Variant | Deferred (out of scope) |
|---|---|---|---|
| `HistoryCenterModal.vue` | `__apply` Filter (52), `__pinned-dismiss` Clear (19), `__more` Load-more (76) | ghost (all 3) | `__close` (13, T1 glyph); `__summary` batch-row toggle (domain control) |
| `ResetConfirmDialog.vue` | `__btn` Revert-equivalent (28) + `__btn--destructive` Reset (41) — shared class, migrated together | ghost / `variant="danger"` | `__close` (12, T1); `reset-entry` (5, link-shaped T3) |
| `MetaAutomationLogViewer.vue` | `__btn` refresh (38) + `__btn--retry` retry (50) + `__btn--support` copy/download-support-packet (80/86) — all 4 sharers of the shared class | ghost (all 4) | `close-x` (6, T1) |

## Byte-equivalence
Every hunk is `<button>`→`<MtButton>` tag swap + bespoke CSS removal only. `@click`/`@click.stop`, `:disabled`, `v-if`, `data-test`/`data-action` attributes are unchanged byte-for-byte (verified via `git diff`). `type="button"` is dropped at each call site since `MtButton` hardcodes it — no `<form>` wraps any of these, so no functional change.

`@click.stop` on the two support-packet buttons still works correctly: `MtButton` re-emits the real native `MouseEvent` object it received internally, so Vue's compiled `withModifiers(handler, ['stop'])` wrapper still calls `stopPropagation()` on that same event before it continues bubbling — verified with a dedicated spec assertion (clicking copy/download does **not** also collapse the parent log-row via its own `@click="toggleExpand"`).

## Variant rationale
[...]

PR #4178, opened 2026-07-12
refactor(multitable): UI-P2-1c T1 — picker + scale-dialog close-× → MtIconButton

zensgit:
## What changed

T1 of `docs/development/multitable-ui-p2-1c-tail-resolution-designlock-20260707.md` §2-T1 (RATIFIED — owner directive 2026-07-11). Migrates the header close-× glyph button in 3 components from a bespoke `<button>&times;</button>` to the shared `MtIconButton` primitive:

| File | Class | Handler |
|---|---|---|
| `apps/web/src/multitable/components/MetaPersonPicker.vue` | `.meta-person-picker__close` | `@click="emit('close')"` |
| `apps/web/src/multitable/components/MetaLinkPicker.vue` | `.meta-link-picker__close` | `@click="emit('close')"` |
| `apps/web/src/multitable/components/ScaleFormattingDialog.vue` | `.scf-dlg__close` | `@click="close"` (local fn: `if (dirty.value && !window.confirm(...)) return; emit('close')`) |

Only the close-× glyph button in each was touched. All three imported `MtButton`/`MtIconButton`'s barrel (`../ui`) already — only the import line and the one `<button>`→`<MtIconButton>` tag swap changed per file.

## Scope correction — 2 files dropped mid-task

This task was originally assigned 5 components (adding `RestorePreviewDialog.vue` + `RestoreBatchDialog.vue`). Before making any edit, I checked for parallel work per repo discipline and found:
- An **open** PR #4174 ("T1 batch-4") already migrates the close-× in `RestorePreviewDialog.vue` and `RestoreBatchDialog.vue` (plus `ResetConfirmDialog.vue`), with matching class names and the same T1 approach.
- `origin/main` still had the bespoke buttons in both files (confirmed unmerged), so this is live, not stale.

I never touched either file — no edits, no spec changes — to avoid duplicating #4174's work and guaranteeing a merge conflict. **This PR's real scope is the 3 remaining components above.** If #4174 is ever abandoned or reverted, `RestorePreviewDialog.vue` / `RestoreBatchDialog.vue` still need their T1 close-× migration — that work is not done anywhere else once #4174's fate is unresolved.

## Left untouched — chip-remove is NOT a close button

[...]

PR #4180, opened 2026-07-12
refactor(multitable): UI-P2-1c T1 batch-4 — drawer/viewer header close-× → MtIconButton

zensgit:
## What changed

T1 batch-4 of `docs/development/multitable-ui-p2-1c-tail-resolution-designlock-20260707.md` §2-T1 (RATIFIED — owner directive 2026-07-11), covering the lock's **"各 manager header" remainder** — the 5 components explicitly assigned to this lane. Migrates each header close-× glyph button from a bespoke `<button>&times;</button>` to the shared `MtIconButton` primitive (`apps/web/src/multitable/ui`, same barrel import idiom as #4130/#4133):

| File | Class | Handler | aria-label pre-migration |
|---|---|---|---|
| `MetaCommentsDrawer.vue` | `.meta-comments-drawer__close` | `@click="emit('close')"` | none (none invented) |
| `MetaAutomationLogViewer.vue` | `.meta-log-viewer__close` | `@click="$emit('close')"` | none (none invented) |
| `MetaAutomationGroupDeliveryViewer.vue` | `.meta-group-delivery__close` | `@click="$emit('close')"` | none (none invented) |
| `MetaAutomationPersonDeliveryViewer.vue` | `.meta-person-delivery__close` | `@click="$emit('close')"` | none (none invented) |
| `HistoryCenterModal.vue` | `.meta-hist__close` | `@click="emit('close')"` | `t('关闭','Close')` — preserved verbatim |

Each source file also dropped the bespoke `type="button"` attribute where present (4 of 5) — redundant, not a behavior change: `MtButton`'s own template already hardcodes `type="button"` on its root native `<button>`, so the DOM type stays `"button"` either way regardless of whether the caller passes it through.

## Scope discipline — what was left untouched

**Only the header close-×.** Per-component button inventory, verified by handler before touching anything:

- **`MetaCommentsDrawer.vue`** has ~10 buttons total. Left untouched: reply / edit / delete / resolve (all mutate a specific comment thread), retry (re-fetches on error), reply-cancel / edit-cancel (dismiss a composer banner, not the drawer), the composer submit button, and the inbox `<RouterLink>` (navigation, not a button at all). None of these are a "close the whole drawer" dismiss — only `.meta-comments-drawer__close` is.
- **The 3 automation viewers** (log/group-delivery/person-delivery) each have Refresh (+ Retry/copy/download support-packet buttons on the log viewer) already migrated to `MtButton` in earlier, unrelated PRs (#3869/#3870/#4089) — those stay byte-identical; only the header close-× (previously untouched, "stays bespoke" per those PRs' own doc comments) moved in this PR.
- **`HistoryCenterModal.vue`**: Filter/Clear/Load-more were already `MtButton` (#4089); the per-row `.meta-hist__summary` batch-toggle button is a domain row control, not a dismiss — untouched.

No permission, deletion, or AI business logic was touched anywhere — buttons + CSS only, exactly as scoped.

## Byte-preservation

`@click` / `v-if` / `:disabled` / `aria-label` / `data-*` copied verbatim per file (diffed against `origin/main`: only the tag name, the `../ui` import line, and the CSS block for each `__close` class changed — nothing else in any of the 5 `.vue` files moved). 2 of 5 close-× buttons had no `aria-label` before migration (`MetaCommentsDrawer`, the 3 automation viewers) — none was invented, matching the "byte-preserve, don't improve" rule.

## Classes: kept, CSS removed

Each of the 5 `__close` classes is the **sole sharer** in its file (repo-wide grep confirmed no other `.vue`/`.ts` reference). Per file:
- class kept on `<MtIconButton class="...">` (Vue attr-fallthrough merges it onto the underlying native `<button>`) — for selector stability, matching existing behavior specs that query these classes directly (e.g. `multitable-comments-drawer.spec.ts`, `multitable-history-fe.spec.ts`).
- the bespoke hardcoded CSS rule was **deleted outright** and replaced with a one-line comment (same convention as #4130/#4133) — no orphaned rule, no double-styling risk.

No new hardcoded hex introduced (diff-grepped against the full component set — 0 hits). No new primitive variants/props/sizes — `MtIconButton` used with only `class` / `aria-label` / `@click` / default slot, all pre-existing on the barrel.

## Honest visual delta (token-normalized, NOT "zero visual change")

Same normalization already landed on `main` via #4099/#4130/#4133: `MtIconButton` renders the `×`/`&times;` glyph inside a fixed 32px (`--ms-control-height`) square control, flex-centered, with `--ms-*` token-driven hover/focus-visible states (a visible focus ring now appears on keyboard focus, which none of these 5 bespoke buttons had). Font-size (previously 18–22px, independently set per file) is no longer independently set — it inherits the shared icon-button sizing. Headers grow by a few px accordingly — expected, not a regression, per this repo's standing discipline against overclaiming pixel-identical migrations.

## Test evidence

A mount spec per component, anchored on the component's own `.X__close` class (never fuzzy `textContent`/optional-chained `querySelector(...)?.click()`), asserting: (1) it renders as a native `<button>` keeping the class + preserved glyph text, and (2) clicking it fires the exact same emit with the exact same (empty) payload. `MetaAutomationLogViewer` / `MetaAutomationGroupDeliveryViewer` / `MetaAutomationPersonDeliveryViewer` / `HistoryCenterModal` specs were extended (new `describe` blocks alongside their existing, unmodified button-migration coverage); `MetaCommentsDrawer` got a new file (none existed) using a real `vue-router` instance, matching the pre-existing `multitable-comments-drawer.spec.ts` mount convention (the drawer renders a `<RouterLink>` in its header) plus a 3rd test pinning that the inbox link and a per-comment reply button are NOT migrated (scope-discipline regression guard).

```
pnpm --filter @metasheet/web exec vitest run meta-comments-drawer-migration meta-automation-log-viewer-migration meta-group-delivery-viewer-migration meta-person-delivery-viewer-migration history-center-modal-migration --reporter=dot

 Test Files  5 passed (5)
      Tests  25 passed (25)
```

Also ran the broader pre-existing regression net around these 5 files (multitable-comments-drawer, meta-comments-drawer-i18n, multitable-history-fe, multitable-history-center-{ai-shortcut-label,inline-diff,pinned-batch-deeplink}, multitable-workbench-restore-wiring, multitable-config-history-modal, multitable-restore-{batch,preview}-dialog) — 15 files / 122 tests, all green.

`vue-tsc -b` (repo `type-check` script): clean, no errors.

**Pre-existing flake noted, not caused by this PR**: `multitable-workbench-history-field-scope-wiring.spec.ts` is flaky when run in the SAME vitest worker alongside several other spec files (2–4 of its 6 tests intermittently fail with timeouts/null-chip assertions) — reproduced identically on an unmodified `origin/main` checkout (stashed this branch's diff, same failure signature). In isolation it passes 6/6 reliably, both on `origin/main` and with this PR's diff applied (ran twice each way). Out of scope for this PR; flagging per this line's standing discipline against silently absorbing pre-existing flakes into an unrelated diff's blast radius.

## Mutation-red evidence (per component, done serially: mutate → red → `git checkout -- <file>` restore from the committed baseline)

1. **MetaCommentsDrawer** — dropped `@click="emit('close')"` → `clicking the header close-× emits 'close'` failed: `expected "spy" to be called 1 times, but got 0 times`. Restored.
2. **MetaAutomationLogViewer** — dropped `@click="$emit('close')"` → same failure shape. Restored.
3. **MetaAutomationGroupDeliveryViewer** — dropped `@click="$emit('close')"` → same failure shape. Restored.
4. **MetaAutomationPersonDeliveryViewer** — dropped `@click="$emit('close')"` → same failure shape. Restored.
5. **HistoryCenterModal** — dropped `@click="emit('close')"` (kept the aria-label binding, mutated ONLY the click handler) → same failure shape. Restored.

Each mutation was applied against the already-committed baseline (this branch's own commit) and restored via `git checkout -- <file>`; `git status --short` was clean after every restore. Full 5-spec / 25-test suite re-ran green after all 5 restores.

## CI wiring (two-point, verified — no edits needed)

- `.github/workflows/multitable-web-guard.yml`: the wildcard `apps/web/tests/*-migration.spec.ts` path trigger (present at both the `pull_request` and `push` blocks, pre-existing on `origin/main`) fires on all 5 touched/new spec files.
- `apps/web/scripts/run-required-web-tests.sh` (backs the REQUIRED `web-tests` check): the bare `migration` token (pre-existing) substring-matches every `*-migration.spec.ts` file, including the new `meta-comments-drawer-migration.spec.ts`.

Both were re-verified on the current `origin/main` tip after rebasing this branch (`grep` confirmed present at the expected lines). Same precedent as #4130/#4133 — neither touched these files either.

## Coordination note (dedup-check finding, for the record)

At the start of this lane I found `#4174` already open, self-titled "T1 batch-4," touching `ResetConfirmDialog.vue` / `RestoreBatchDialog.vue` / `RestorePreviewDialog.vue` — a disjoint file set from this PR's assignment (the "各 manager header" remainder). I did not author #4174 (read-only `gh pr view` during the dedup check only) and never touched any of its 3 files. This PR is branched fresh off `origin/main` and only touches the 5 files listed above.

refs docs/development/multitable-ui-p2-1c-tail-resolution-designlock-20260707.md §2-T1

### DIFF
--- a/apps/web/src/multitable/components/HistoryCenterModal.vue
+++ b/apps/web/src/multitable/components/HistoryCenterModal.vue
@@ -10,7 +10,7 @@
     <div class="meta-hist__modal" role="dialog" aria-modal="true" :aria-label="t('历史记录', 'History')">
       <header class="meta-hist__header">
         <h3 class="meta-hist__title">{{ t('历史记录', 'History') }}</h3>
-        <button class="meta-hist__close" type="button" :aria-label="t('关闭', 'Close')" @click="emit('close')">×</button>
+        <MtIconButton class="meta-hist__close" :aria-label="t('关闭', 'Close')" @click="emit('close')">×</MtIconButton>
       </header>
 
       <div v-if="initialBatchId && !pinnedDismissed" class="meta-hist__pinned" data-test="hist-pinned-batch">
@@ -90,7 +90,7 @@ import { useLocale } from '../../composables/useLocale'
 import { useHistoryCenter } from '../composables/useHistoryCenter'
 import { historyActor } from '../utils/meta-record-labels'
 import HistoryBatchChangesList from './HistoryBatchChangesList.vue'
-import { MtButton } from '../ui'
+import { MtButton, MtIconButton } from '../ui'
 import type { LinkedRecordSummary, PersonSummary } from '../types'
 
 const props = defineProps<{
@@ -224,7 +224,9 @@ function formatTime(iso: string): string {
 .meta-hist__modal { background: var(--meta-surface, #fff); color: var(--meta-text, #1f2329); border-radius: 8px; min-width: 420px; max-width: 640px; max-height: 76vh; overflow: auto; padding: 16px; box-shadow: 0 8px 32px rgba(0, 0, 0, 0.18); }
 .meta-hist__header { display: flex; align-items: center; justify-content: space-between; margin-bottom: 12px; }
 .meta-hist__title { margin: 0; font-size: 15px; }
-.meta-hist__close { background: none; border: none; font-size: 20px; line-height: 1; cursor: pointer; color: inherit; }
+/* .meta-hist__close: now <MtIconButton> (ghost, token-styled; the × glyph char passes through its
+   default-slot icon fallback (size token-normalized to the icon control)). Bespoke hardcoded CSS removed
+   (UI-P2-1c T1 batch-4). Class kept on the element only for selector stability. */
 .meta-hist__filters { display: flex; gap: 8px; margin-bottom: 12px; flex-wrap: wrap; }
 .meta-hist__filter { font-size: 12px; padding: 4px 6px; border: 1px solid var(--meta-border, #ddd); border-radius: 4px; }
 /* .meta-hist__apply / .meta-hist__more (below): the Filter and Load-more controls are now <MtButton>
--- a/apps/web/src/multitable/components/MetaAutomationGroupDeliveryViewer.vue
+++ b/apps/web/src/multitable/components/MetaAutomationGroupDeliveryViewer.vue
@@ -3,7 +3,7 @@
     <div class="meta-group-delivery">
       <div class="meta-group-delivery__header">
         <h4 class="meta-group-delivery__title">{{ automationLabel('delivery.groupTitle', isZh) }}</h4>
-        <button class="meta-group-delivery__close" type="button" @click="$emit('close')">&times;</button>
+        <MtIconButton class="meta-group-delivery__close" @click="$emit('close')">&times;</MtIconButton>
       </div>
 
       <div class="meta-group-delivery__body">
@@ -55,7 +55,7 @@ import { useLocale } from '../../composables/useLocale'
 import type { DingTalkGroupDelivery } from '../types'
 import type { MultitableApiClient } from '../api/client'
 import { automationLabel, automationStatusLabel } from '../utils/meta-automation-labels'
-import { MtButton } from '../ui'
+import { MtButton, MtIconButton } from '../ui'
 
 const props = defineProps<{
   visible: boolean
@@ -157,15 +157,9 @@ watch(
   color: #0f172a;
 }
 
-.meta-group-delivery__close {
-  border: none;
-  background: none;
-  font-size: 22px;
-  cursor: pointer;
-  color: #64748b;
-  line-height: 1;
-  padding: 0 4px;
-}
+/* .meta-group-delivery__close: now <MtIconButton> (ghost, token-styled; the &times; glyph char passes
+   through its default-slot icon fallback (size token-normalized to the icon control)). Bespoke hardcoded
+   CSS removed (UI-P2-1c T1 batch-4). Class kept on the element only for selector stability. */
 
 .meta-group-delivery__body {
   padding: 16px 20px 20px;
--- a/apps/web/src/multitable/components/MetaAutomationLogViewer.vue
+++ b/apps/web/src/multitable/components/MetaAutomationLogViewer.vue
@@ -3,7 +3,7 @@
     <div class="meta-log-viewer">
       <div class="meta-log-viewer__header">
         <h4 class="meta-log-viewer__title">{{ automationLabel('log.title', isZh) }}</h4>
-        <button class="meta-log-viewer__close" type="button" @click="$emit('close')">&times;</button>
+        <MtIconButton class="meta-log-viewer__close" @click="$emit('close')">&times;</MtIconButton>
       </div>
 
       <div class="meta-log-viewer__body">
@@ -144,7 +144,7 @@ import {
   renderAutomationLogSupportPacketJson,
   renderAutomationLogSupportPacketMarkdown,
 } from '../utils/automation-log-support-packet'
-import { MtButton } from '../ui'
+import { MtButton, MtIconButton } from '../ui'
 
 const props = defineProps<{
   sheetId: string
@@ -293,7 +293,9 @@ watch(
 }
 
 .meta-log-viewer__title { margin: 0; font-size: 16px; font-weight: 700; color: #0f172a; }
-.meta-log-viewer__close { border: none; background: none; font-size: 22px; cursor: pointer; color: #64748b; line-height: 1; padding: 0 4px; }
+/* .meta-log-viewer__close: now <MtIconButton> (ghost, token-styled; the &times; glyph char passes
+   through its default-slot icon fallback (size token-normalized to the icon control)). Bespoke hardcoded CSS
+   removed (UI-P2-1c T1 batch-4). Class kept on the element only for selector stability. */
 
 .meta-log-viewer__body {
   padding: 16px 20px 20px;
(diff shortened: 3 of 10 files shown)


zensgit:
## Adversarial gate — **MERGE_CLEAN** (binds head `dd1af89a232fe7380e5bab8aa20a3b946b72cf00`)

Reviewed refute-first in a fresh disposable worktree off `refs/reviews/pr4180` (canonical checkout untouched). **No P1, no P2.** Two P3s, neither a blocker.

### Owner red lines — all PASS
1. **Buttons + CSS only** — proven, not diff-read. These 5 sit next to exactly the logic the owner ring-fenced (comment permissions, delivery/retry, the audit surface), so I diffed the `<script>` block of each against `main`: the *only* change in **all five** files is the import line. Zero permission/deletion/AI/automation logic touched.
2. **Byte-preserve** — `@click` verbatim (`emit('close')` / `$emit('close')`); `HistoryCenterModal`'s `:aria-label="t('关闭','Close')"` preserved verbatim. **Two non-obvious cases, both cleared:**
   - **`type="button"` dropped on 4 of 5** — legitimate: `MtButton.vue:3` hardcodes `type="button"` on its root native `<button>`, so the DOM type is unchanged.
   - **`MetaCommentsDrawer` had *no* `type`** (implicit `type="submit"`) and now gets `type="button"`. Only a no-op if there's no `<form>` ancestor — verified: the only `<form>`s in `multitable/` are `MetaDashboardView` + `MetaFormView` (all `@submit.prevent`), neither of which renders the drawer. Behavior-identical.
   - **No aria-label invented**: `MtIconButton` binds `ariaLabel || title` → `undefined` for the 4 that had none, so Vue omits the attribute. Correct per byte-preserve.
   - **No double-fire**: `MtIconButton` declares `click` in `defineEmits`, so the caller's `@click` is stripped from `$attrs` and fires exactly once.
3. **No new hex** — 0 hits on added lines.
4. **No invented style semantics** — `multitable/ui/**` **untouched**.
5. **`×` ≠ close** — verified per file: exactly one `×` each, all true dismisses. The drawer's reply/edit/delete/resolve/retry/cancel and the inbox `RouterLink` are untouched (the spec even *pins* that the link and a reply button are NOT migrated). No chip/attachment/condition `×` was migrated.

### Orphaned/shared CSS — PASS
All 5 `__close` classes are sole sharers; after the diff each appears **only** as the element + its replacement comment — **no surviving rule anywhere**. No double-styling.

### Specs bite — mutation reproduced (I did not take the claim)
Dropped `@click="emit('close')"` from `MetaCommentsDrawer.vue:13` (the component closest to the red line) → the **exact** close-× test reds: `expected "spy" to be called 1 times, but got 0 times` (1 failed / 2 passed). Restored → green.
**No false-green anchors**: hard `as HTMLButtonElement` + direct `.click()` — no `querySelector(...)?.click()` anywhere. Exact `toBe('×')`, `toHaveBeenCalledTimes(1)`, `toHaveBeenCalledWith()`.

### Real runs
`7 files / 39 tests passed` (5 migration specs + 2 neighbor regression specs); `vue-tsc -b` exit 0.

### The flake — cleared by evidence, not by assertion (P3, informational)
This mattered: `multitable-workbench-history-field-scope-wiring.spec.ts` **mounts the real `HistoryCenterModal`**, which this PR modifies — so "pre-existing" needed proof. I ran the exact CI guard batch 3× on this ref and 3× on **unmodified `origin/main` (`b674dba8c`) in the identical environment**:

| Ref | run 1 | run 2 | run 3 |
|---|---|---|---|
| `origin/main` (unmodified) | 23 failed (field-scope 2/6) | 13 failed (field-scope **4/6**) | 1 failed |
| **#4180** | 7 failed (field-scope 2/6) | **0 failed — fully green** | 10 failed |

Main never produced a clean run in 3 attempts; this PR did. Same files, same signature. **Not caused or worsened by this PR** — and it did not manifest in CI (`multitable-web-guard` + `web-tests` both SUCCESS).

### P3-1 — CI path-trigger residual left unfixed (asymmetric with sibling #4178)
4 of your 5 sources (`MetaCommentsDrawer`, the 3 automation viewers) have **no** path trigger in `multitable-web-guard.yml` (only `HistoryCenterModal.vue` does). Sibling #4178 explicitly fixes exactly this residual for *its* files per the R5c precedent; this PR adds none. A *future* PR touching only e.g. `MetaCommentsDrawer.vue` would run no web-guard.
**Only P3, not a blocker**: the REQUIRED `web-tests` job is **unconditional** (no `paths:` filter) and its run-list already covers these components, so the specs still run on any such PR; the advisory guard is vitest-only. Your body's CI claims are, as written, **true** — you claim only that the wildcard + `migration` token cover your *specs*, and they do. Worth a one-line follow-up for consistency.

### Visual honesty — PASS
Body correctly says **token-normalized**, not "zero visual change."

**NIT**: the 4 unlabeled close-× buttons keep `×` as their accessible name. Correct under byte-preserve — flagging only so an a11y-labelling pass is understood as a separate deliberate slice, not an oversight.

Full report: `/tmp/pr4178-pr4180-review-claude-20260712.md`

PR #4181, opened 2026-07-12
feat(directory): make corp_id immutable once set on directory integrations (org-transfer Phase 1 §12.1)

zensgit:
## Why

`updateDirectoryIntegration` runs `UPDATE directory_integrations SET corp_id = $4` in place. Swapping `corp_id` to a **different** DingTalk corp on an integration that already has synced records is a tenant change disguised as an ordinary form save: the next sync's absence sweep marks every account **and** department still tagged with the old corp inactive (no longer "seen" under the new corp), **silently mass-deactivating the previous organization**. This is Phase 1 §12.1 of the provider-org-transfer plan (`docs/development/provider-org-transfer-development-plan-20260709.md`, #3944) — the highest-severity, most self-contained risk on that line.

## What

- Guard in `updateDirectoryIntegration`, **before any mutation**: block only a genuine tenant **swap** (current corp non-empty, next non-empty, different) when the integration already has synced records. The "already synced" probe checks **both** `directory_accounts` and `directory_departments` (the absence sweep deactivates both). Initial set, clear, and same-corp resend pass through.
- Escape hatch: env `DIRECTORY_ALLOW_ACTIVE_CORP_ID_CHANGE` (test/dev), default off.
- New `DirectoryTenantChangeBlockedError`; the admin PUT route maps it to **`409 DIRECTORY_TENANT_CHANGE_BLOCKED`** with a message pointing at an org-transfer workflow.
- Docs: line review + remaining-order plan + Phase 1 dev/verification MD.

## Verification

- tsc 0 · 6/6 unit tests (swap-blocked + UPDATE-never-issued, both-tables probe, allow paths) · neighbor directory suites green.
- Load-bearing mutations proven: neuter throw / flip predicate / drop the departments clause each redden a test.
- **Independent adversarial gate (Opus): APPROVE** (0 P1/P2). Bypass hunt confirmed `updateDirectoryIntegration` is the only in-place corp_id-swap path; no silent-clear hazard; the one P3 (probe missing departments) is fixed here.

## Scope note

This is the safe, unblocked first step. Phase 2+ (transfer schema/API, the **two-corp staging proof hard-gate**, user-identity rebind, group adapter, admin UI) is planned + sequenced in the MD, not implemented here — those touch live-synced schema, identity writes, or a staging proof that needs real infrastructure.


zensgit:
## Reworked per REQUEST_CHANGES — corp_id now IMMUTABLE once set (first-sync TOCTOU closed)

**Disarmed** (auto-merge off) and reworked. The first cut blocked a corp_id swap **only when the integration already had synced records**. Owner review found that rule leaves a **first-sync TOCTOU window**: `syncDirectoryIntegration` reads the corp config, claims the run lease, then pulls and writes rows — during the first sync (corp_id set, zero account/department rows yet) an interleaved PUT could change corp_id from under the in-flight sync, which then writes the OLD corp's data into the now-retagged integration, re-arming the exact mass-deactivation on the next sweep.

**New rule (absolute, race-free):**
- In an ordinary PUT, an **already-set `corp_id` can never change**. The guard sits in `updateDirectoryIntegration` after `normalized` is computed and **before the UPDATE**, and issues **no record probe** — so the block is independent of whether any rows have synced yet, which is what closes the window.
- Initial set (empty → value) and same-corp resend pass through. A "clear" is unreachable (`normalizeIntegrationInput` throws `'corpId is required'` earlier).
- **No production escape hatch** — the `DIRECTORY_ALLOW_ACTIVE_CORP_ID_CHANGE` env bypass was removed entirely (0 src occurrences). A mis-entered corp_id before first sync is fixed by delete-and-recreate; a genuine org change goes through the org-transfer workflow (Wave 4).
- New exported `DirectoryTenantChangeBlockedError`; admin PUT route maps it to **`409 DIRECTORY_TENANT_CHANGE_BLOCKED`**.

**Verification (rebased onto current `main`, +477/−0 additive):**
- `tsc --noEmit`: exit 0.
- Unit `directory-tenant-change-guard.test.ts` **4/4**: change-to-set blocked with **`toHaveBeenCalledTimes(1)`** (no probe, no UPDATE); env=true still blocked; same-corp resend applies; initial-set applies.
- Real-PG `directory-tenant-change-immutable.db.test.ts` **5/5** (two-point CI-wired): first-sync window (corp set, zero records) → swap rejected, row unchanged; blocked with records present; same-corp + initial-set succeed.
- Load-bearing mutation: neuter the `throw` → first-sync + records-exist tests redden; restore → green.
- Independent adversarial re-gate (Opus): **APPROVE** — `toctouClosed: true, bypassFound: false, mainRevertFound: false, overRestrictionFound: false`.

Left **UNARMED** pending owner re-review.

PR #4182, opened 2026-07-12
feat(web): multitable UI P2-1c T1 batch-6 — close-× to MtIconButton

zensgit:
## Summary

UI-P2-1c tail resolution, T1 (`docs/development/multitable-ui-p2-1c-tail-resolution-designlock-20260707.md` §2-T1, RATIFIED) — batch-6. Migrates the header close-× in 3 components to the shared `MtIconButton` primitive, matching the convention already landed in #4130 / #4133 / #4174 (open, same lane): keep the `&times;` glyph in the default slot, keep the bespoke class on the element for selector stability, delete only the now-dead CSS rule.

Presentation-only. No permission/delete/AI logic touched.

## Buttons migrated (exactly 3, per the batch table)

| File | Class | Handler | Change |
|---|---|---|---|
| `MetaCommentsDrawer.vue` (~L13) | `.meta-comments-drawer__close` | `emit('close')` | `<button>` → `<MtIconButton>`, CSS rule removed |
| `MetaSheetPermissionManager.vue` (~L9) | `.meta-sheet-perm__close` | `requestClose()` → `emit('close')` | `<button type="button">` → `<MtIconButton>` (type attr dropped, see below), CSS rule removed |
| `MetaRecordPermissionManager.vue` (~L10) | `.meta-record-perm__close` | `requestClose()` → `emit('close')` | `<button type="button">` → `<MtIconButton>` (type attr dropped, see below), CSS rule removed |

`emit()` set is byte-conserved: same event name (`close`), same payload (none), same call site, same condition (unconditional).

`type="button"` was dropped from the two permission-manager buttons during migration. This is DOM-behavior-equivalent, not a semantic change: `MtButton.vue` (which `MtIconButton` wraps) hardcodes `<button type="button" ...>` in its own template unconditionally — verified by reading the source, and pinned by a `getAttribute('type')).toBe('button')` assertion in both migration specs.

## Other `×` glyphs found and deliberately left (with handler)

Repo-wide grep for `×`, `✕`, `✖`, `&#215;`, and el-icon `Close`/`CloseBold` across all three files found **no other × glyph of any form** — the single `&times;` per file (the header close button) is the only one. Every other `<button>` in these files is text-labeled:

- **MetaCommentsDrawer.vue** (9 buttons total, 1 migrated): `.meta-comments-drawer__reply` → `emit('reply'|'edit'|'delete', id)`, `.meta-comments-drawer__resolve` → `emit('resolve', id)`, `.meta-comments-drawer__retry` → `emit('retry')`, `.meta-comments-drawer__reply-cancel` → `emit('cancel-edit'|'cancel-reply')`. None touched.
- **MetaSheetPermissionManager.vue** (18 buttons total, 1 migrated): `.meta-sheet-perm__tab` → local `activeTab` ref (tab switch), `.meta-sheet-perm__action` → `applyEntry`/`clearSubjectOverrides`/`copySubjectAclFromMemberGroup`/`removeEntry`/`grantCandidate`. None touched.
- **MetaRecordPermissionManager.vue** (5 buttons total, 1 migrated): `.meta-record-perm__action` → `saveEntry`/`removeEntry`/`grantCandidate`. None touched.

## Red-line confirmation

`MetaSheetPermissionManager.vue` and `MetaRecordPermissionManager.vue` are permission surfaces. Per lock: migration touches **only the close-button element + its CSS rule**. Confirmed by diff — each file's changeset is exactly: (1) the close `<button>` → `<MtIconButton>` element, (2) one new import line (`MtIconButton` from `../ui`), (3) the now-dead `__close` CSS rule replaced by a one-line comment. No permission/role/grant/ACL/delete logic line was touched in either file (verified via `git diff origin/main -- <file>`).

## Honest visual delta (not "zero visual change")

The `&times;` glyph shrinks from its original bespoke size (18px in MetaCommentsDrawer, 24px in both permission managers) to `MtIconButton`'s 14px icon-font size, and each control becomes a 32×32 box (`--ms-control-height`) — previously a borderless, unsized inline glyph with no hit-target beyond the character itself.

## Class-name collision check

`git grep` across the whole repo (excluding the 3 target files and the new spec files) for `meta-sheet-perm__close`, `meta-record-perm__close`, `meta-comments-drawer__close` — zero other consumers (no `:deep()`, no e2e selector, no other component). Each CSS rule was removed outright, matching the #4174 precedent for single-sharer classes.

## Zero new hardcoded hex

`git diff origin/main -- <the 3 files>` added lines contain no `#[0-9a-f]{3,6}` — token-only (`MtIconButton` consumes `--ms-*`/`--el-*` via `MtButton`).

## Tests

3 new files (`apps/web/tests/meta-comments-drawer-migration.spec.ts`, `meta-sheet-permission-manager-migration.spec.ts`, `meta-record-permission-manager-migration.spec.ts`), 9 tests total — 3 per component: (1) real mount + DOM shape assertion (native `<button>`, class retained, glyph text, `type="button"` reflects even though not authored), (2) real `.click()` + exact emit assertion (`toHaveBeenCalledTimes(1)` + `toHaveBeenCalledWith()`), (3) a POSITIVE CONTROL clicking an untouched button in the same component (`.meta-comments-drawer__retry`, `.meta-sheet-perm__tab`, `.meta-record-perm__action--danger`) to prove both that it still behaves exactly as before and that the harness can actually observe these controls.

Pre-existing specs for all 3 components re-run clean (`multitable-comments-drawer.spec.ts`, `meta-comments-drawer-i18n.spec.ts`, `multitable-sheet-permission-manager.spec.ts`, `multitable-record-permission-manager.spec.ts` — 40 tests, all pass).

### Mutation evidence (real output, then restored)

Deleted each migrated `@click` in turn (e.g. `<MtIconButton class="meta-comments-drawer__close">` with no `@click`), reran the corresponding migration spec:

```
✓ renders the header close-× as a native <button> ...
× clicking the header close-× emits `close` ...
  → expected "spy" to be called 1 times, but got 0 times
✓ POSITIVE CONTROL: the untouched retry button still emits `retry` exactly as before
```

Same red-then-restore pattern confirmed for `.meta-sheet-perm__close` (requestClose) and `.meta-record-perm__close` (requestClose) — the click-emit test goes red, the render-shape test and the positive control stay green. All three components restored via `git checkout -- <file>` after committing (working tree verified clean against HEAD before mutating).

### Type-check

`apps/web`: `npx vue-tsc -b` — clean, no output.

### CI wiring (verified, not assumed)

- `apps/web/scripts/run-required-web-tests.sh` (the **required**, unconditional `web-tests` job) filters `vitest run` with the bare token `migration` — ran it locally; all 3 new spec files execute inside that gate (179 files / 2104 tests, all green), confirmed by name in the output.
- `.github/workflows/multitable-web-guard.yml`'s `pull_request.paths` already lists `apps/web/tests/*-migration.spec.ts` (wildcard) plus explicit paths for `MetaSheetPermissionManager.vue`/`MetaRecordPermissionManager.vue` — its own `vitest run` command also contains the bare `migration` token; ran that filter set locally (114 files / 1114 tests, all green), confirmed by name.
- No workflow YAML was edited — matching the #4130/#4133/#4174 convention (new `*-migration.spec.ts` files ride the existing wildcard + bare-token wiring; adding explicit per-component paths would be a third, uninvited convention).

## Scope discipline

Per the design-lock's per-slice discipline, this batch touches exactly the 3 files / 3 buttons listed, and nothing enumerated for other tail tiers (T2 plain variant, T3 MtLink, T4 harness, T5 shared-class managers) was invented or pre-empted here.

Refs the tail-resolution design-lock `multitable-ui-p2-1c-tail-resolution-designlock-20260707.md` §2-T1 batch series (#4130, #4133, #4174).


zensgit:
## Independent adversarial gate: **APPROVE** — 0 P1 / 0 P2

Permission red line verified by **reconciling the diff's +/− counts** (perm managers 5+/9−, drawer 5+/2−), not merely by reading it — the counts leave no room for a hidden hunk. `applyEntry` / `clearSubjectOverrides` / `copySubjectAclFromMemberGroup` / `removeEntry` / `grantCandidate` / `saveEntry` all untouched. `requestClose()` is a bare `emit('close')`, so the "migrated button bypasses a guard" hypothesis has **no guard to bypass**.

### The `type="button"` question — and it turns out this PR is a net *hardening*

`MtButton` hardcodes `type="button"` unconditionally, so dropping the explicit attribute is DOM-equivalent. But note what that means for `MetaCommentsDrawer`: its original was a **bare `<button>` with no `type` at all** — i.e. an implicit **`submit`**. The migration makes it explicitly `type="button"`, *removing* a latent hazard rather than introducing one.

Better still: `multitable-web-guard.yml` carries a prior review's standing warning that **no gated spec pinned the primitive's `type="button"`**. This PR's specs now pin it **in the required gate** — deleting the hardcode from `MtButton` turns both permission specs RED. That closes part of the gap that comment describes.

### Follow-ups found — pre-existing, deliberately NOT fixed here

- **`MetaCommentsDrawer` runs in zero CI gate.** It is in no guard `paths:` list and its pre-existing specs are in no required filter — reply / edit / delete / resolve are entirely unguarded today. This PR *improves* that (its migration spec is the drawer's first spec in a required gate), but the underlying gap is its own ticket.
- **The "1 file / 2 tests" local flake baseline is optimistic.** Under a loaded 44-spec combined run, clean `origin/main` **with zero PR code** failed *worse* than this branch (22 and 13 failures across two runs, vs the branch's 14) — all `Test timed out in 5000ms`, non-deterministic. The tight baseline only holds when specs run alone. Worth knowing before anyone reads a loaded local run as a regression.

zensgit:
Narrowed to the two permission managers (head `96c33f5a2`). A parallel session's #4180 migrates MetaCommentsDrawer's header close-× (alongside the automation viewers), so I've ceded that file — MetaCommentsDrawer.vue is reverted byte-identical to main and its migration spec removed, to avoid a collision. What remains is the part #4180 does not touch: the close-× of MetaSheetPermissionManager and MetaRecordPermissionManager, whose migration was verified line-by-line to touch only the close button + CSS (no ACL logic). The adversarial APPROVE stands for these two files (they are byte-identical to the reviewed head; only MetaCommentsDrawer was removed).

PR #4187, opened 2026-07-12
docs(multitable): D-1c — form-submit EDIT writes no record revision (design-lock, PROPOSED)

zensgit:
**Docs-only. Zero runtime. Zero flags. Status: PROPOSED — ratifies nothing.**

Gate-front deliverable for a newly-found P1: a **proven reproduction** + a **ratify-ready design-lock**, taken to the furthest point short of the owner gate. No production source was touched. The owner must rule on **OD-1…OD-6** before any implementation begins.

## The defect

`packages/core-backend/src/routes/univer-meta.ts:14423` — the **authenticated** form-submit EDIT branch raw-`UPDATE`s `meta_records` and bumps `version` 1→2 while writing **no record revision**. `reconstructRecordsAtT` (`record-reconstructor.ts:34-70`) derives record data *purely* from `meta_record_revisions`, so history returns the **pre-edit** value at **every** T after the edit.

This is the same bug class D-1 was created to fix, on a path D-1 did not cover. D-1 §6/§7 deferred the **CREATE** half (a missing *tail*). This is the **EDIT** half — a *hole in the middle of the chain*, named in no document — which is why it needs a ruling rather than a self-driven fix.

## Reproduced — real Postgres 14, real routes, positive control

Isolated Docker PG on a non-default port; migrations with CI's exact `MIGRATION_EXCLUDE`. Every mutation driven through the **real Express route** (no hand-rolled SQL for the path under test). Positive control: a normal-path `PATCH` on the same sheet **does** emit a revision and reconstructs correctly — so the harness is trustworthy, not merely passing.

- live row = `v2-edited-via-form` @ version **2**; `reconstructRecordsAtT(T after edit)` = `v1-original` @ version **1**
- **T7 PIT view** shows the stale pre-edit value; the control record is correct
- **`revert-preview` at a T *after* the edit** — which should be a no-op — proposes reverting it (`visibleRevertCount: 1`)
- **`revert-execute` performs it** (`revertedCount: 1`): the member's edit is overwritten, and a scan of `snapshot` **and** `patch` across every revision finds it in **zero** rows → **irrecoverable**
- **Audit hole**: the edit is absent from the record history timeline; `restore-preview(targetVersion: 2)` → `404 VERSION_NOT_FOUND`

Neighbours green in the same run (env health): reconstructor + form-submit-trigger **11/11**, D-1 delete-revision parity **8/8**.

## It is a class of EIGHT, not one site (§7a)

The sibling audit enumerated every `meta_records` mutation site and read each enclosing function. **Most consequential finding: D-1 closed only the *delete* half of the automation/plugin side doors — those lanes still emit nothing for `create`/`update`.** Plugin-SDK `patchRecord` was **also reproduced**. Six further sites carry the identical fingerprint by source (labelled as source-verified, **not** measured).

This reframes the scope question (**OD-1**) and is the core of the case for a **revision-disposition guard** (**OD-6**) — the direct analogue of the existing rank-8 lock-disposition guard, whose own header already names *"form-submit EDIT"* as a path that burned this line once before, for the lock.

## Corrections to the source finding (verified, not inherited)

- The reconstructor's header attributes itself to *"restore-preview (T5-2)"*; the **record-level** restore-preview reads revisions **by version** directly (`:9225`) and does **not** call it. Its restore-family consumer is the **sheet-level** revert/reset.
- The finding cited `FormSubmitInput.recordId` at `types.ts:903`. **No such symbol exists in the repo.** The reachability claim is nonetheless correct — carried by the route's inline zod schema (`:14152-14153`).

## Open decisions left for the owner

**OD-1** scope (EDIT-only / +CREATE / full 8-site sweep) · **OD-2** `source` — note `RecordRevisionSource` **already declares a `'public-form'` slot that nothing emits**, yet this branch is *authenticated-only*, so the name is arguably wrong · **OD-3** `actorId` · **OD-4** `meta_links` side-effects (deferred, not absorbed) · **OD-5** existing polluted history (forward-only; no fabricated timestamps) · **OD-6** the systemic guard.

**Not merged, no auto-merge armed.** Awaiting owner ratification.


zensgit:
## ⛔ Owner verdict is REQUEST CHANGES — this lock is NOT ratifiable yet. Five corrections needed.

The owner ruled on D-1c today (2026-07-13) and was explicit: **「当前设计锁还不能直接 ratify，需要先把以下裁决和"既有污染历史预检"写进去，再复审一次。」** Verdict = **REQUEST CHANGES**; ratify only after the below are encoded and the lock is re-reviewed. The current header says `RATIFIED` — that is **premature and contradicts the owner's stated verdict**, and must be walked back to a pending/REQUEST-CHANGES status until re-review.

I'm recording the owner's verbatim rulings here so they're authoritative on the PR, then applying the corrections. Five gaps between the current §0.5 and what the owner actually ruled:

**1. OD-1 = 5 independent slices, NOT "3 cuts (lanes A/B/C)".** Owner: 「全量 bucket-A(A1–A8），但分 **5 个独立 runtime slice**。执行顺序：表单 A1+A6 → plugin A2+A5 → automation A3+A4 → **approval A7** → **attachment A8**。」 The owner deliberately separated **approval (A7)** and **attachment (A8)** into their OWN slices — §0.5 folded attachment into Lane A and approval into Lane C. Re-encode as five slices in that exact order; each slice independently verifies txn boundary, actor, source, revision shape, and PIT correctness — **not one hot-core PR**.

**2. OD-2 attachment source = `'attachment'`, NOT `'rest'`.** Owner: source describes the **write entry point**, not auth identity — 「表单=public-form、plugin=plugin、automation=automation、approval=approval、**attachment=attachment**」. §0.5 currently says "attachment-delete uses `'rest'`" — wrong per the ruling.

**3. OD-4 = link-edge history is a SEPARATE design-lock, deferred, marked UNSOLVED.** Owner: 「link-edge history 单独设计锁。**不混入 D-1c**，但必须明确仍未解决；不得把 link-only 修改宣称为已有完整历史。」 §0.5 currently says "IN SCOPE for lane A … Do NOT defer" — that is the opposite of the ruling. (The record's `data` snapshot naturally includes link-field ids; but **edge-level** `meta_links` history is its own lock and must be flagged as still-unsolved, not claimed complete.)

**4. The HISTORY_INCOMPLETE hard-lock is MISSING entirely** — and it is the owner's headline addition. Required, verbatim:
   - 当前活数据与最新 revision snapshot 不一致时返回统一 **`HISTORY_INCOMPLETE`**，**不得生成执行 token，不得写入**。
   - 预检比较**权威数据内容**，不能只比较 version；锁定/解锁等 **version-only 变化不得误报**。
   - **execute 必须重新检查**，防止 preview 后又出现无 revision 写入。
   - real-DB golden 必须覆盖「已污染记录 → preview/execute 都拒绝且**零写入**」和「仅 version 变化但数据一致 → **不误拒绝**」。

**5. §5 still carries the old advisory recommendations** (OD-1 "(b) as immediate rung", etc.) that now contradict the ruled §0.5 — a half-done fix-forward. §5 must be reconciled so no section still reads as "owner hasn't ruled."

OD-3 (known actor must be written; anonymous form CREATE → null; **no fabricated system actor**), OD-5 (forward-only, no speculative backfill), and OD-6 (revision-disposition guard as a separate slice, covering INSERT/UPDATE/DELETE) are correctly captured in spirit — I'll tighten wording to match.

I'll apply these now and then re-run an independent adversarial review (the owner's 「再复审一次」). **I am not flipping this to RATIFIED** — that stays the owner's call after re-review.

zensgit:
## Independent re-review — VERDICT: NOT-READY (advising only; owner ratifies)

Reviewed head `ff3a2e8b8` (OPEN), code verified primary-source vs `origin/main @ 087ffa47a`. Full report: `/tmp/pr4187-rereview-claude-20260713.md`.

The rulings OD-1…OD-6 are faithfully encoded, the status correctly reads REQUEST-CHANGES / not-ratified, the A1–A8 audit is accurate **8/8** (every file:line is a real `meta_records` version-bumping write with no revision), and the 5-slice grouping is internally consistent (`§0.5` = `§7a` = `§10`). But **§0.6 `HISTORY_INCOMPLETE` — the headline hard-lock this re-review exists to validate — is under-specified in both comparison directions, and as literally written its own two goldens certify a broken implementation.** Three blockers:

**P1-A — §0.6 false-POSITIVE (feature-bricking).** A full-`data` comparison false-positives `HISTORY_INCOMPLETE` on every healthy record with a formula/rollup/lookup/auto-number field. Proof: `formula-engine.ts:345` / relation-agg `univer-meta.ts:3855` (cross-record fan-out) / `auto-number-service.ts:101` all write `data` with **no version bump and no revision**; and the healthy REST create (`record-service.ts:706`) writes the revision `snapshot: patch` (user fields only) in-txn, then materializes formula keys into `data` post-commit. So `live.data ≠ latest snapshot` on any formula sheet immediately after a normal create/edit → revert/reset/PIT becomes DOA on most real sheets. The item-4b positive control only tests the *data-identical* lock/unlock case, so it does **not** catch a refuse-everything-on-drift precheck. **Fix:** comparator must project both sides to user-authored/writable field ids only (exclude field types {formula, rollup, lookup, relation-agg, autoNumber} via `loadFieldsForSheet`); add a **third golden**: formula/rollup sheet, healthy create+edit → precheck PASSES.

**P1-B — §0.6 false-NEGATIVE (fail-OPEN data loss).** The zero-revision (uncaptured-CREATE: A4/A5/A6) case has **no snapshot to compare**, and §0.6 never defines it. `computeSheetReset` (`univer-meta.ts:10024-10034`) pushes any live row absent from the revision-derived `stateMap` into the **delete-set** — so a precheck that iterates the *reconstructed* set (the natural implementation) never sees a zero-revision record, and it is **silently destroyed** on reset: exactly the worst outcome in §0.5 line 18, and precisely what §0.6 exists to prevent. Neither golden tests it. **Fix:** §0.6 must state the precheck enumerates **live rows**; a live record with zero revisions → `HISTORY_INCOMPLETE` (refuse, never silent-delete); add a **fourth golden**: uncaptured-create present before reset-T → refuse + zero writes + record NOT deleted.

**P2 — §5 OD-4 stale contradiction.** The §5 OD-4 cell (line 269) still asserts the *overruled* "IN SCOPE for lane A **(OD-4 ruling)** … `meta_links` restoration-consistency golden (reproduces … the `meta_links` **edges** consistently) … **Do NOT defer**" — which is exactly the edge-level history that corrected §0.5/§10 defer to a separate lock. The §5 preamble flags the draft was overruled, but the row body was never brought into line (half-done fix-forward). An implementer keying off §5/§6 goldens would build the deferred edge-history into the form slice. **Fix:** strike the meta_links-edge golden + the "(OD-4 ruling)" label; point to §0.5/§10.

Also: P3 site-count doesn't reconcile (§0.5 "37" / OD-6 "22+15" vs §7a buckets 8+12+11+2=33); NIT OD-2 union — `source` is plain `text` (no enum/CHECK) and the union is open, so neither `'approval'` nor `'attachment'` needs adding to work.

**PASS:** status ≠ ratified; "3 lanes" purged; attachment='attachment' corrected; OD-3 anonymous→`actorId=null` already in code; OD-4 snapshot-vs-edge coherent in §0.5/§10; §0.6 item-3 execute re-check is coherent **and necessary** (the existing `previewIdentity` scopeHash guard is computed from revision-derived state, so it is structurally blind to uncaptured writes).

Once P1-A, P1-B, P2 close in the lock text, the rulings/audit/slice-plan are — on this review — complete and consistent enough to ratify.

zensgit:
## Re-review blockers folded — `629ec746c`. The lock now awaits the owner's ratify word.

Process state, per the owner's own sequence (裁决已给 → 写入 → 复审 → 补齐 → ratify):

1. ✅ Owner rulings OD-1..OD-6 encoded (5 slices, entry-point sources, actor rules, link-edge deferral, forward-only, guard slice) — `ff3a2e8b8`
2. ✅ `HISTORY_INCOMPLETE` hard-lock encoded (§0.6)
3. ✅ Independent re-review ran (NOT-READY: 2 P1 + 1 P2)
4. ✅ All three blockers folded — `629ec746c`:
   - **§0.6 item 2** now projects the comparison onto **user-authored/writable fields only** (formula/rollup/lookup/auto-number materialization writes `data` with no revision *by design* — a full-`data` diff would brick revert/reset on every formula sheet), pinned by new **G-HI-3**;
   - **§0.6 item 3**: the precheck **enumerates LIVE rows**; a record with **zero revisions** ⇒ `HISTORY_INCOMPLETE` (the undefined case was fail-OPEN: `computeSheetReset` silently deletes live rows absent from the reconstruction), pinned by new **G-HI-4** with a zero-writes post-assertion;
   - **§5 OD-4 cell** reconciled to the ruling (edge-level `meta_links` history = separate deferred lock, marked UNSOLVED).

The re-review's remaining components all PASS: OD encodings faithful, A1–A8 audit verified **8/8**, 5-slice grouping consistent across §0.5/§7a/§10, execute-TOCTOU re-check confirmed necessary (`previewIdentity` scopeHash is revision-derived — structurally blind to uncaptured writes).

**Nothing here is ratified.** Per the re-review the lock is now complete enough to ratify; that word is the owner's. On ratify, implementation order is §10: **§0.6 precheck first**, then the five slices (form → plugin → automation → approval → attachment), then the OD-6 guard.

zensgit:
## ⚠ Coordination: FOUR runtime PRs now implement this lock — which is NOT yet ratified

For the record on the governance anchor. Runtime implementations **#4216 / #4219 / #4220 / #4227** (a parallel session's R13 lanes) are open against this lock while its status is **REQUEST CHANGES — NOT RATIFIED**. The owner's verdict was explicit: encode the rulings + the `HISTORY_INCOMPLETE` precheck, re-review, **then** ratify; implementation is not authorized before that word. (The session that opened them had earlier marked the lock RATIFIED in its header — that status was premature and has been walked back; they likely implemented in good faith against it.)

Two of the four also implement the **superseded 3-lane grouping** rather than the owner's ruled 5 slices:
- **#4219 (Lane A)** folds **attachment-delete (A8)** into the form slice — the owner ruled attachment is its **own slice ⑤**.
- **#4220 (Lane C)** folds **approval resultWriteback (A7)** into automation — the owner ruled approval is its **own slice ④**. The owner's stated reason for 5 slices: per-slice independent txn/actor/source/PIT verification, 不打成热核心大 PR.
- **#4216 (Lane B)** happens to coincide with ruled slice ② (plugin A2+A5).
- **#4227 (OD-6 guard)**: the owner sequenced the guard **last**, and §4.3 of the roadmap records why — landing it before the 8 sites are fixed turns CI red on every unfixed site (unless it exempts them, which would defeat it).

**None of these PRs is closed or armed by me** — they are another session's work and the content may be substantially salvageable (Lane B nearly as-is; Lane A minus attachment ≈ slice ①; Lane C minus writeback ≈ slice ③). The decision is the owner's:

- **(a)** ratify the lock, require re-slicing to the ruled 5 (split A8 out of #4219, A7 out of #4220; sequence #4227 last) — the conservative reading of the existing ruling;
- **(b)** ratify and explicitly accept the 3-lane grouping as delivered — the owner may amend their own ruling; that word has to be the owner's, not an implementer's;
- **(c)** hold all four until §0.6 (the precheck the owner ordered to land FIRST) exists.

Also noting: **no PR implements §0.6 `HISTORY_INCOMPLETE` yet** — the piece the owner ordered to land before the slices, because it protects existing polluted data during the backfill. Whatever else is decided, that gap is real.

PR #3964, opened 2026-07-09
fix(collab): require auth + sheet-read to join comment socket rooms

zensgit:
## 概要

把评论 socket 房间的加入对齐到 live-editing 的 `join-sheet` 认证范式:加入评论房现在需要**已验证身份 + 对目标 sheet 的读权限**(复用 join-sheet 用的同一注入式 `sheetRoomAuthChecker`),此前只有 join-sheet 有此门。

## 改动

**后端 `CollabService.ts`**
- 新增共享 `authorizeCommentSheetAccess(socket, spreadsheetId)`:解析已验证 user → 无则 deny `unauthorized` → `sheetRoomAuthChecker` 读权限 → 不允许 deny `forbidden`。接入 `join-comment-sheet` / `join-comment-record`。
- `join-comment-inbox`:要求已验证身份方可加入跨表活动流(per-user 活动路由为后续 follow-up,已在 handler 注释)。

**前端**(使登录用户仍可用)
- `comments-realtime.ts`(sheet+record 订阅)、`useMultitableCommentInboxRealtime.ts`(inbox):socket 握手带 `auth: { token }`(`useAuth().getToken()`),镜像 `useMultitableCommentRealtime` / sheet realtime 既有接线。

## 验证

- `rooms.basic` 集成测试 **11 pass**:4 个新测试(未认证 comment-sheet/record 加入被拒且收不到广播;已认证但无读权限 → `forbidden`;未认证 inbox 被拒)+ 3 个既有评论/presence 测试更新为携带 token。
- Mutation:`authorizeCommentSheetAccess→true` → 3 红(KILLED);inbox 认证门旁路 → 1 红(KILLED);restore 复绿。
- `tsc`(core-backend)+ `vue-tsc`(apps/web)0 error。
- 说明:`multitable-workbench-view` / `multitable-comment-inbox-realtime` 两 spec 在 origin/main 上因无关的 api mock 不全**本就 red**(非本 PR 回归),故无绿 FE spec 可扩;认证契约由后端集成测试证明。

## 不在本 PR(后续)

- inbox 的 per-user 活动路由(现仍为全局活动**元数据**流,无正文)——需广播收件人/订阅设计,单独改。

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4191, opened 2026-07-12
fix(web): repair 3 multitable specs red on main + wire them into CI (they ran in no workflow)

zensgit:
## Context

159 multitable `apps/web` specs run in **no CI workflow at all**. An independent audit (running all 159 locally) found 10 red files / 92 tests: 7 are a documented workbench-cluster quarantine, and 3 are fresh reds nobody had seen because nothing ever ran them. This PR repairs the 3 fresh reds and wires them into CI so a future regression is actually visible.

### 1. `apps/web/tests/multitable-client.spec.ts` (~L786, `loads and normalizes record history entries`)
**Case (b) — stale spec, superseded by an intentional server+client change.**
refs #2928 (2026-06-19, "resolve actor display names") added server-side `actorName` enrichment to `GET /sheets/:sheetId/records/:recordId/history` (`packages/core-backend/src/routes/univer-meta.ts` ~L9168: `resolveUserDisplayNames` + `{...it, actorName: ...}`) and a matching client-side field in `normalizeRecordHistoryEntry` (`apps/web/src/multitable/api/client.ts` ~L780). Verified both sides — this is a real, intentional, additive contract change, not a client-only guess (checked `packages/core-backend` per the FE-mock-is-not-the-contract rule). The spec's fixture response has no `actorName`, so the normalizer's fallback (`null`) is correct; the spec's *expected* object was just never updated when #2928 landed. Fixed by adding `actorName: null` to the expectation.

### 2. `apps/web/tests/multitable-comment-affordance.spec.ts` (~L137, ~L187)
**Case (b) — stale spec, superseded by an intentional presentation-only change.**
refs #3734 (2026-07-07, "multitable UI P1b — feature-row/view-tab/comment emoji→Element Plus SVG icons") replaced the inline `💬` emoji glyph in `MetaCommentAffordance.vue` with `<el-icon><ChatDotRound/></el-icon>`. Confirmed via a scratch mount probe: the icon element still renders (`<i class="el-icon meta-comment-affordance__icon" data-comment-affordance-icon="true"><svg>...</svg></i>`) but has **no text content** — an SVG has none. The two idle-icon assertions still checked `.textContent` for the emoji. Fixed by asserting the icon node exists, carries `.el-icon`, and contains an `<svg>` child instead.

### 3. `apps/web/tests/multitable-comment-inbox-realtime.spec.ts`
**Case (c) — broken/bit-rotted spec (fails at COLLECTION, 0 tests execute).** Two independent stale-mock gaps:
- refs #3964 (2026-07-09, "require auth + sheet-read to join comment socket rooms") made `useMultitableCommentInboxRealtime.ts` call `auth.getToken()` unconditionally before opening the socket. The spec's `useAuth` mock only stubbed `getCurrentUserId`, so `auth.getToken` was `undefined` — a `TypeError` inside the mounted composable.
- Independently, the spec's `../src/utils/api` mock only exported `getApiBase`. The composable imports `resolveMultitableCommentsRealtimeBaseUrl` from `comments-realtime.ts`, which imports normalize helpers from `client.ts`. `client.ts`'s module-level `export const multitableClient = new MultitableApiClient()` singleton resolves its default fetch fn via `apiFetch` at *import time* — vitest throws immediately when a mocked module is missing an export the graph touches, before any test body runs. This is why the file failed at collection, not inside a test.

Fixed by adding `getToken: vi.fn().mockReturnValue(...)` to the `useAuth` mock and `apiFetch: vi.fn()` to the `utils/api` mock.

**Not fixed (found, out of scope):** `apps/web/tests/multitable-comment-inbox-view.spec.ts` has the *same* `auth.getToken` mock gap as issue 3's first bullet above (confirmed failing identically on unmodified `origin/main`, 2/2 tests, `TypeError: auth.getToken is not a function`) and is not wired into any workflow either. It wasn't one of the 3 specs this PR was scoped to, so it's left untouched — flagging it as a 4th pre-existing red the original audit didn't catalog, for separate follow-up.

## CI wiring (two-point, both verified)

Added all 3 spec tokens (`multitable-client`, `multitable-comment-affordance`, `multitable-comment-inbox-realtime`) to:
- `apps/web/scripts/run-required-web-tests.sh` — the unconditional run-list backing the **required** `web-tests` check (`.github/workflows/web-tests.yml`). This is the primary fix for "ran in no workflow": this script runs regardless of which files changed.
- `.github/workflows/multitable-web-guard.yml` — both the `pull_request.paths` and `push.paths` triggers (spec files + their source deps: `MetaCommentAffordance.vue`, `MetaCommentActionChip.vue`, `comment-affordance.ts`, `useMultitableCommentInboxRealtime.ts`, `comments-realtime.ts`; `client.ts` was already a trigger) **and** this workflow's own `vitest run` filter line — closing the documented "path trigger fires but doesn't run the spec" hole for these 3 files specifically.

Verified each token resolves to exactly one file (`npx vitest run <token>` → 1 file matched, not a substring collision with sibling `multitable-comment-*` specs).

## Quarantined files — untouched

The 7 documented workbench-cluster red files are **not** part of this PR and remain red and unwired. This PR does not make "all multitable specs" green — only these 3.

## Test evidence

Local runs (this branch, `apps/web`):
```
npx vitest run tests/multitable-client.spec.ts tests/multitable-comment-affordance.spec.ts tests/multitable-comment-inbox-realtime.spec.ts
 ✓ tests/multitable-comment-inbox-realtime.spec.ts  (4 tests)
 ✓ tests/multitable-client.spec.ts  (35 tests)
 ✓ tests/multitable-comment-affordance.spec.ts  (10 tests)
 Test Files  3 passed (3)
      Tests  49 passed (49)
```

Ran the full `run-required-web-tests.sh` list on this branch — the 3 new specs execute and pass inside it. Note: a small set of unrelated tests (DingTalk automation-manager/rule-editor, approval-center, workbench field-scope) intermittently time out (`Test timed out in 5000ms`) under local sandbox CPU contention when running ~2100+ tests in one process; this reproduces identically on **unmodified `origin/main`** (22 failures, same file set, before this PR's changes even exist) and does not reproduce in `web-tests.yml` on GitHub Actions (`gh run list --workflow=web-tests.yml --branch main` — last 8 runs all `success`). Pre-existing local-sandbox flake, unrelated to this diff — not fixed here.

## Mutation proof (the wiring actually bites)

1. Token is in the required run-list: `run-required-web-tests.sh` diff (see commit).
2. Token selects exactly one file: `npx vitest run multitable-client` → matches only `multitable-client.spec.ts`.
3. Broke the `actorName` assertion in `multitable-client.spec.ts` (`null` → `'MUTATION-PROOF-should-not-pass'`), ran `npx vitest run multitable-client`:
   ```
   ❯ tests/multitable-client.spec.ts  (35 tests | 1 failed)
     → expected [...] to deeply equal [...]
   Test Files  1 failed (1)
   Tests  1 failed | 34 passed (35)
   EXIT_CODE=1
   ```
4. Restored the assertion via precise edit (not `git checkout`); re-ran → `35 passed (35)`, working tree clean (`git diff --stat` empty against the commit).

This confirms a regression in the now-wired spec would flip the required `web-tests` check from green to red.

## Not done / stopped on

- `multitable-comment-inbox-view.spec.ts` (4th pre-existing red, same root cause as spec 3 here) — intentionally left untouched, out of this PR's named scope.
- The 7 quarantined workbench-cluster files — intentionally left untouched per standing quarantine.
- Did not merge or arm auto-merge.


zensgit:
## Adversarial review — VERDICT: MERGE_CLEAN (approve-with-hardening) @ `7dd5393f5`

Re-derived all 3 diagnoses against real source; ran a mutation battery + 2 positive controls; reproduced red-on-main by restoring main's specs against identical product code; verified both-point CI wiring bites; type-check clean. **No P1, no BLOCK.** All three specs were genuinely red/uncollected on main (guarding nothing) — red→green + first-time CI wiring is net-positive, and the pinned `actorName: null` is verified value-correct against the server (`univer-meta.ts:8401-8402` emits `string|null`, never omitted/raw-id).

The three surviving mutants below are **coverage gaps left open, not wrong evidence** — recommend hardening but not blocking.

**P2 — `multitable-client.spec.ts:797` actorName enrichment unguarded.** Hardcoding `actorName: null` in the client normalizer (`api/client.ts:782`, killing #2928's enrichment) leaves the spec 35/35 green. This spec is the #2928 contract spec now in the REQUIRED gate, so a real regression stays invisible. Fix: add a second fixture row *with* `actorName` populated and pin it passes through.

**P3 — `multitable-comment-inbox-realtime.spec.ts` socket-token unguarded.** Removing `auth: { token }` from the handshake leaves 4/4 green. Mitigated: the server-side gate IS tested (`rooms.basic.test.ts:543` `SECURITY: rejects an unauthenticated join-comment-inbox`). Optional: assert `io()` was called with `objectContaining({ auth: { token } })`.

**P3 — `multitable-comment-affordance.spec.ts:137,192` icon identity unguarded.** Swapping `ChatDotRound`→`Delete` (trash icon) leaves 10/10 green; only "an svg exists" is asserted. Presentation-only. State distinction is still guarded. Optional: assert the SVG's 2-`<path>` shape or stub the icon by name.

**NIT** — PR body cites `univer-meta.ts ~L9168` for #2928; that's the audit-log endpoint. The record-history endpoint enriches at L8401-8402. Diagnosis substance is correct.

**Verified PASS:** required gate `web-tests.yml:73` runs the script; each token → exactly 1 file; wiring bites (broke an assertion → required-form vitest exits non-zero); 4th file `multitable-comment-inbox-view.spec.ts` is red-on-main with the same gap and in NO run-list (genuinely out of scope, falsifies no claim); flake claim true (named files byte-identical to main, pass in isolation; required gate green on merge-base).

Full report: `/tmp/pr4191-review-claude-20260712.md`. Review only — not merging, not arming auto-merge.

zensgit:
## P2 closed — `077880d24` (the enrichment positive control)

Took the gate's P2. The spec pinned only `actorName: null`, so killing #2928's client-side enrichment left it 35/35 green — and since this spec now sits in the REQUIRED `web-tests` gate as #2928's contract, that regression would have been invisible. A fail-closed suite with no positive leg.

Added a second fixture row (`rev_2`) carrying `actorName: 'Ada Lovelace'` and pinned it passes through.

**Mutation-proven the gap is now closed**: hardcode `actorName: null` at `api/client.ts:782` and the spec now REDS —
```
-     "actorName": "Ada Lovelace",
+     "actorName": null,
   Tests  1 failed | 34 passed (35)
```
It did not red before this commit. Enrichment is now guarded in both directions.

Leaving the two P3 hardenings (socket-token presentation, icon identity) as the gate recommended — both are presentation-only, both upstream-mitigated (the server auth gate and the affordance state distinction are each independently tested). Not blocking, and folding them in would widen this repair PR past its purpose.

Re-gate scoped to the new head before I arm — a verdict binds the SHA.

PR #4193, opened 2026-07-12
feat(integration): stock-prep T5 — DB-layer immutability trigger + CHECK↔store drift test for 066 audit table

zensgit:
## Summary

Stock-prep MVP remaining-development T5: **066 审计表不可变的 DB 层兜底 + store↔CHECK 断言测**.

### A. DB-layer immutability backstop — mechanism actually verified, not assumed

**App-connection-role finding (read `connection-pool.ts` + `db.ts` + env before choosing a mechanism):**
`packages/core-backend/src/integration/db/connection-pool.ts`'s `PoolManager` builds **one** `pg.Pool` from a single `DATABASE_URL`; `src/db/db.ts` hands that same pool to **both** the Kysely migration runner (`src/db/migrate.ts`) and app runtime queries. There is no migration-role vs. app-role split anywhere in this repo. Concretely: `docker-compose.dev.yml` sets `POSTGRES_USER=metasheet` for both DB and `DATABASE_URL`; `.github/workflows/plugin-tests.yml` uses `postgresql://postgres@localhost/...` for both. Whichever role runs `066_create_integration_stock_prep_audit.sql`'s `CREATE TABLE` becomes the table **owner** — the exact same role the app queries with.

**REVOKE tested and rejected as a fake backstop (the #4126 anti-pattern):** against a throwaway `postgres:16-alpine` container, connected AS the owner role: `REVOKE UPDATE, DELETE ON integration_stock_prep_audit FROM <owner>` executes with no error, and a subsequent `UPDATE`/`DELETE` by that same role **still succeeds** — Postgres owners retain implicit privileges that `REVOKE` cannot remove. A REVOKE-based "backstop" here would look like a control while doing nothing.

**Mechanism chosen: `BEFORE UPDATE OR DELETE ... RAISE EXCEPTION` trigger.** Row-level triggers fire for every role, including the table owner and superusers (unlike privilege GRANT/REVOKE, and unlike default RLS which owners/superusers bypass). New migration: `packages/core-backend/src/db/migrations/zzzz20260712120000_stock_prep_audit_immutable_trigger.ts` — zzzz-named so it always runs after 066 (name-sorted); idempotent (`CREATE OR REPLACE FUNCTION` + `DROP TRIGGER IF EXISTS` before `CREATE TRIGGER`); only touches `integration_stock_prep_audit`; append (INSERT) path is untouched — only UPDATE/DELETE are denied.

**Real-DB interception evidence** (both a dev-shaped container — custom role granted SUPERUSER — and a CI-shaped container — `postgres:14`, `postgres` role, `metasheet_test` DB, matching `plugin-tests.yml`'s exact setup):
```
current_user | owner
--------------+----------
 postgres     | postgres        -- app role IS the table owner (REVOKE would be a no-op)

INSERT 0 1                       -- append still works
ERROR:  integration_stock_prep_audit is append-only: UPDATE is not permitted (id=ci-check-1)
ERROR:  integration_stock_prep_audit is append-only: DELETE is not permitted (id=ci-check-1)
```
Full migration chain (`tsx src/db/migrate.ts --latest`, 264 migrations including 066 + the new zzzz) applies cleanly end-to-end against a fresh DB; `--rollback` cleanly removes the trigger, `--latest` re-applies it.

New real-DB test: `packages/core-backend/tests/integration/stock-prep-audit-immutable-trigger.db.test.ts` (isolated per-test schema, mirrors the F3 `files-storage-key-migration.db.test.ts` precedent; runs the REAL 066 SQL verbatim, not a hand-mirrored table) — 6/6 passing:
- precondition: connecting role owns the table (documents WHY REVOKE would be a no-op)
- INSERT unaffected before the trigger exists
- UPDATE/DELETE succeed *before* the trigger (proves the trigger — not something else — is the guard)
- after `up()`: INSERT still succeeds, UPDATE rejected, DELETE rejected, row untouched, append still works afterward
- idempotent: `up()` run twice → exactly one trigger, guard still holds
- `down()` removes trigger+function → UPDATE/DELETE succeed again (causality check)

Wired into `.github/workflows/plugin-tests.yml`'s real-DB allowlist (the "Run attendance integration tests" step, which is this repo's catch-all bucket for `*.db.test.ts` files) — per the house rule that core-backend integration specs outside that allowlist never execute in CI regardless of local green.

### B. store↔066 CHECK 8-action drift test — already exists on `main`, verified + mutation-tested

`plugins/plugin-integration-core/__tests__/stock-preparation-audit-migration.test.cjs` (merged in #4029, already on `main`) already does exactly this: parses the CHECK vocabulary out of 066's SQL text and asserts it's set-equal to the store's `STOCK_PREP_AUDIT_ACTIONS`. No new test needed for B — I verified it's real by mutation-testing it: temporarily added a 9th fake action to `stock-preparation-audit-store.cjs` only (066 left untouched), re-ran the test — **it went red** with exactly the expected diff (fake action listed under `expected`, missing from `actual`), then reverted the file (confirmed clean via `git diff`).

### Hard constraints honored
- Touches only: the 066 table (via a new trigger, not an edit to 066 itself), one new zzzz migration, one new test file, and one CI-wiring line. No changes to any of the 9 logical sheets, `http-routes.cjs`, or external-write paths.
- No fake defense-in-depth: A's mechanism was chosen only after empirically disproving REVOKE and empirically proving the trigger, against real Postgres in both dev-shaped and CI-shaped configurations.
- Append path verified unbroken in every real-DB run above.

## Test plan
- [x] Real Postgres (dev-shaped: custom role w/ SUPERUSER; CI-shaped: `postgres:14`, `postgres` role, `metasheet_test`) — full migration chain to latest, INSERT succeeds, UPDATE/DELETE rejected, rollback/re-apply clean
- [x] `packages/core-backend/tests/integration/stock-prep-audit-immutable-trigger.db.test.ts` — 6/6 passing against both container shapes
- [x] `packages/core-backend`: `tsc --noEmit` clean (0 errors)
- [x] `packages/core-backend`: `tests/unit` full run — 5250/5251 passing; the 1 failure (`plm-workbench-routes.test.ts`) is pre-existing/unrelated (passes in isolation; this PR touches only 2 new files + 1 CI-wiring line, none in that area) — not introduced by this change
- [x] `plugins/plugin-integration-core`: `npm test` (full ~90-file chain) — exit 0, includes the existing `stock-preparation-audit-migration.test.cjs` CHECK↔store parity test
- [x] Mutation test on the existing CHECK↔store parity test — confirmed red on a fake 9th action, confirmed clean revert


zensgit:
独立 opus 对抗审阅:**APPROVE(0 P1/0 P2)**。头号疑虑(触发器误杀合法 DELETE)**完全证伪**:全仓 grep `integration_stock_prep_audit` 唯一写者是 audit-store 的 `{append,list}`,**零** retention/cleanup/partition/TRUNCATE 生产者——没有合法 DELETE 被误杀。真库实证:264 迁移链净跑、066@47ms 早于 zzzz@49ms、INSERT/COPY 成、UPDATE/DELETE 连 superuser 也被拒(坐实 trigger 胜过 REVOKE-on-owner)、mutation neuter CREATE TRIGGER→3 测红。**可选 P3(非阻塞)**:行级触发器不拦 TRUNCATE(实测 TRUNCATE 能清空);建议后续加 `BEFORE TRUNCATE FOR EACH STATEMENT` 触发器 + 测——但当前**零 TRUNCATE 生产者**(store 只 INSERT),是 backstop 非 boundary,不阻塞合并。已记为 follow-up。

PR #4195, opened 2026-07-12
docs(approval): B3-07 attachment pipeline — fold in owner storage/MIME/auth decisions (PROPOSED)

zensgit:
## What

Rev 3 of the B3-07 approval-attachment design lock (`docs/development/approval-attachment-pipeline-design-lock-20260709.md`). **Docs only — zero runtime, zero migrations, zero flag changes. Status remains PROPOSED, awaiting owner ratify.**

## Owner architecture decisions (§0-bis, D1–D6)

- **D1** — object storage, NOT local FS (deploy-host disk pressure #159).
- **D2** — dedicated `approval_attachments` table.
- **D3** — initiator (发起人) uploads; approver/CC upload out of scope for v1.
- **D4** — auth-proxied download reusing the existing approval-visibility authorization; no public/signed URLs in v1; no new permission model.
- **D5** — default OFF (`APPROVAL_ATTACHMENTS_ENABLED=false`); B2-28 honest-disable byte-unchanged while OFF.
- **D6** — v1 MIME allowlist = PDF/JPEG/PNG/TXT/CSV; Office/ZIP deferred until AV integration.

## Updated in this revision (owner review round) — O1–O4 now all DECIDED

- **O1 — caps DECIDED:** 20 MB/file, 10 files/field, 50 MB/submission.
- **O2 — unbound-retention TTL DECIDED = 7 days** (was 24 h), plus a new **stale-reference detection on draft-restore** contract (a restored draft older than the TTL must detect and flag GC-swept attachment ids, never resolve to a deleted blob).
- **O3 — storage DECIDED:** production requires an S3-compatible object-store provider; a local-FS provider in production **fail-closes uploads (503)**.
- **O4 — DECIDED:** SVG/HTML/XML/executables are **permanently rejected** (415), even post-AV, on stored-XSS / execution-vector grounds AV does not mitigate.

## Two concurrency/cleanup contracts folded in (owner P1)

- **Bind↔GC race — claim-then-delete-blob with a symmetric guard.** The GC's conditional soft-delete and the create-time bind now both guard on `instance_id IS NULL AND deleted_at IS NULL`, so they serialize on the row and exactly one wins in **either** interleaving (bind-won → GC skips blob-delete; GC-won → bind matches 0 rows → create fails closed) — no "bound but blob gone." Proven by a constructed GC↔submit concurrency test exercising both interleavings (G11, §12 item 13).
- **Object-store cascade cleanup.** DB `ON DELETE CASCADE` deletes attachment rows on instance deletion but never the object-store blobs; instance deletion/purge now carries an explicit object-store blob-cleanup leg (G12, §12 item 14).

## Substrate correction

`StorageService` is local-FS only today (S3 skeleton removed in F6, refs #3925); an object-store `StorageProvider` (interface exported) is a named implementation prerequisite for D1.

**Do not merge on green alone — this is a design-lock revision for owner ratify review.**

### DIFF
--- a/docs/development/approval-attachment-pipeline-design-lock-20260709.md
+++ b/docs/development/approval-attachment-pipeline-design-lock-20260709.md
@@ -1,9 +1,75 @@
 # B3-07 — Approval attachment upload pipeline · DESIGN-LOCK
 
-**Status: PROPOSED** (doc-only; awaits owner ratify). No runtime code ships with this document.
+**Status: RATIFIED 2026-07-15 (owner).** The design is ratified — this lock's contracts are the implementation acceptance bar. Landing order: merges after #4196; once all four locks are on main, P2 durable-delivery runtime is authorized to start, and the attachment feature plus every related flag stay **OFF** until the full implementation + 8-scenario acceptance pass. **This document is still the design contract (no runtime code ships with it).** History: was PROPOSED (doc-only; awaiting owner ratify).
 Implementation is a **later, separately-opted-in** change and MUST land behind the default-OFF flag
 defined in §9. Until ratify + implementation, the B2-28 honest-disable stopgap (§1) stays exactly as
 it is.
+
+**Rev 2 (2026-07-12): owner architecture decisions folded in — see §0-bis.** Six owner decisions
+(object storage, dedicated table, initiator-upload, auth-proxied download, default-OFF, narrowed v1
+MIME allowlist) resolve five of the seven §13 ratify questions; the remainder were collected in
+§13.2 「待 owner 裁决」.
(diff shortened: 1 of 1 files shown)


zensgit:
OWNER RATIFICATION: RATIFIED at reviewed head 8df4d66b52c6f552d61edd8b07da2f880c3e10ac, sequenced after PR #4196. The approval-attachment design contract, including the bind/GC and object-storage decisions, is approved. Before merge: flip the document status to RATIFIED, rebase onto the main that contains #4196 without semantic drift, rerun required checks, and keep the feature default-OFF.

Issue #4162, opened 2026-07-12
[P2][planning] Untangle the excluded view-table migration cluster so the full snapshot-protection suite can run in CI


zensgit:
**Type:** planning-only (P2). Not on any current critical path. Investigation + design first; no code changes are requested by this ticket itself.

## Problem

`tests/integration/snapshot-protection.test.ts` boots a real `MetaSheetServer` against real Postgres and exercises snapshot labelling / protection / rules over HTTP. It currently runs in **no** CI workflow. Its silent-return false-green was removed in #4145 (setup now hard-fails; `expect.hasAssertions()`), so it can no longer report green without executing — but it still cannot be **wired into** CI, because the CI test database cannot be migrated to the schema it needs.

The `Run DB migrations` step in `.github/workflows/plugin-tests.yml` migrates `metasheet_test` with a `MIGRATION_EXCLUDE` list that omits the view-table migration cluster:

- `20250924120000_create_views_view_states.ts`
- `20250925_create_view_tables.sql`
- `042a_core_model_views.sql`

So the CI database has **no `views` table**. `SnapshotService.createSnapshot` → `captureViewState` does `selectFrom('views')`, so every assertion in the suite fails with `relation "views" does not exist`. (Empirically confirmed: wiring the suite in fails 21/21 on exactly that error.)

Re-enabling the two snapshot-label migrations alone is safe and additive, but insufficient — the `views` cluster is the real blocker, and it is almost certainly excluded because those migrations conflict with each other or with later ones.

## Goal

1. Determine **why** the view-table migration cluster is excluded (which migrations conflict, and how).
2. Design a resolution so the test database can be migrated to include `views` / `view_states` without breaking the other real-DB suites that share the migration lane.
3. Once resolved, wire the full `snapshot-protection.test.ts` suite into the required real-DB lane (two-point wiring: excluded from the no-DB default job so it cannot skip-green, run as a whole file in `test (20.x)`), with `--reporter=verbose` so an empty collection is visible.

## Non-goals / guard-rails

- Do **not** remove entries from `MIGRATION_EXCLUDE` piecemeal without understanding the conflicts — it feeds several real-DB steps, and a bad change would break unrelated suites.
- This is a migration-hygiene / CI change, independent of any feature work.

## Context

[...]

PR #4198, opened 2026-07-12
docs+ci(migration): T8 — document the 3 exclusion lists' rationale + harden warn-only guard to full coverage

zensgit:
## Scope

T8 (小头, non-planning-gated) from the database-audit P1-3 finding: **document, don't unify**,
the three independent migration exclusion/skip lists, and **harden the warn-only guard's
coverage without changing its semantics**. The big rework named in #4162 (fix
view_states/gantt/snapshot/protection migrations' `up()`, lift the CI exclusion, wire the two
zero-coverage integration tests into CI) is **explicitly out of scope here** and remains gated.

No actual exclude/skip set changed anywhere. No stock-prep code, `http-routes.cjs`, or migration
DDL touched. Verified via diff that every `MIGRATION_EXCLUDE:` value line and the
`SUPERSEDED_LEGACY_SQL_MIGRATIONS` array contents are byte-identical before/after.

## A. Three lists, now inline-documented

Confirmed (via git blame/log, not guessed) three independent mechanisms:

1. **CI per-PR gate's `MIGRATION_EXCLUDE`** — `plugin-tests.yml` (x2 jobs),
   `observability-strict.yml`, `observability-e2e.yml`, `safety-guard-e2e.yml`. Drops a
   migration entirely (no history marker).
2. **`migration-replay.yml`'s own `MIGRATION_EXCLUDE` subset** — narrower job (run
   `db:migrate` twice, assert idempotency), independently evolving.
3. **`SUPERSEDED_LEGACY_SQL_MIGRATIONS`** in `migration-provider.ts` — turns a migration into
   a no-op history marker (name stays, body doesn't run) instead of dropping it.

Every occurrence of #1/#2 (6 total) now carries per-item reason comments plus a shared header
explaining the 3-mechanism relationship. Two real divergences are documented with their actual
git history instead of guessed rationale:

- **`20250925_create_view_tables.sql`**: the owner_id FK bug it guards against was fixed by
  #3627, and `migration-replay.yml` already dropped this exclusion in #3632 (verified passing
  there). The other 4 occurrences were never re-verified/updated to match — flagged as
  **possibly stale**, not resolved here (that's the #4162 untangling).
- **`zzzz20260114110000_create_user_orgs_table.ts`**: `plugin-tests.yml`'s `test` job
  deliberately does **not** exclude it — attendance auto-absence needs `user_orgs` applied
  there (commit `b1fc1e19d1`, message quoted in the comment). The same commit also dropped it
  from the `after-sales-integration` job in the same file, which doesn't touch
  attendance/user_orgs — flagged as a **possible unverified side-effect** of a file-wide edit
  rather than asserted as intentional.

`migration-provider.ts`'s `SUPERSEDED_LEGACY_SQL_MIGRATIONS` (30 entries, array contents
unchanged) now has grouped rationale comments (zombie tables, two entries backing genuinely
dead code — `WorkflowRepository.ts` / `DataMaterializationService.ts` have zero importers
repo-wide, `035_create_files`'s gap closed by #4016, `users` column drift) plus the documented
3-item overlap with `MIGRATION_EXCLUDE` (042a/048/049, confirmed harmless by the 2026-07-10
audit) and the known asymmetry: **production/on-prem `db:migrate` runs with no exclude at
all**, so the view-table cluster excluded in CI has never had a per-PR green run of its `up()`
— that gap is #4162, gated.

`packages/core-backend/MIGRATION_EXCLUDE_TRACKING.md` gets a new section making it the single
source of truth for the 3-list relationship + the divergence table, plus a note that its own
Pre-Existing-Issues/Phase-2/History sections reference migration filenames that no longer exist
(2025-10-29 plan superseded by the 2026-05-12 no-op/twin design) — historical content kept, not
deleted. `packages/core-backend/migrations/claudedocs/MIGRATION_FIX_GUIDE.md` (2025-10-29,
progress frozen at 1/7) gets a stale-notice banner pointing at the current design doc and
tracking doc — content kept, not deleted.

## B. Guard hardened from 3/10 to full coverage

`scripts/ci/validate-migration-exclude.sh` previously checked whether a `MIGRATION_EXCLUDE`
env var contained 3 hardcoded filenames (and isn't wired into any workflow or npm script —
confirmed via repo-wide grep both before and after this change; it remains unwired). It now
reads the real workflow files and `migration-provider.ts` directly and checks:

- **A**: every item in every one of the 6 `MIGRATION_EXCLUDE` occurrences against the full
  known 11-item baseline (not 3) — anything extra is flagged as a possibly-undocumented new
  exclusion;
- **B**: every occurrence-to-occurrence divergence against a table of the two known, reviewed
  divergences above — anything else is flagged as undocumented drift;
- **C**: the overlap between `MIGRATION_EXCLUDE` and `SUPERSEDED_LEGACY_SQL_MIGRATIONS` against
  the known 3-item overlap, in both directions (new unexpected overlap, or an expected overlap
  item going missing);
- **D**: a size canary + design-doc-pointer presence check on `SUPERSEDED_LEGACY_SQL_MIGRATIONS`.

**Semantics unchanged: still warn-only, still exits 0 unconditionally, still not wired into any
workflow.** I did not flip it to fail-closed and did not wire it into CI — both would be
behavior changes.

**Recommendation for the owner** (not applied): now that the guard actually covers the full
list and cross-checks git-verified divergences instead of 3 arbitrary filenames, it may be
worth (a) wiring it as a non-blocking step somewhere (e.g. `plugin-tests.yml`) so its warnings
show up in CI logs instead of requiring a manual run, and/or (b) deciding whether categories A/C
(brand-new undocumented items) should eventually fail closed while B/D stay warn-only. Both are
your call, not mine to make unilaterally.

Refactored `main()` into a testable `run_checks()` (no `exit`) + thin `main()` wrapper, and made
`WORKFLOWS_DIR`/`PROVIDER_FILE` overridable via env vars. Added
`scripts/ci/__tests__/validate-migration-exclude.test.sh` (mirrors the existing
`ci-flake-classify.test.sh` convention) with `good`/`drift` fixture sets under
`scripts/ci/__fixtures__/migration-exclude/`.

## Test results

- `bash scripts/ci/validate-migration-exclude.sh` against the real repo: **0 warnings** (only
  informational asymmetry note) — confirms today's state is fully documented by the new
  KNOWN_* tables.
- `bash scripts/ci/__tests__/validate-migration-exclude.test.sh`: good fixture → 0 warnings;
  drift fixture → all 5 injected drift cases detected (undocumented exclusion, undocumented
  divergence, undocumented overlap, missing expected overlap, missing doc pointer). PASS.
- `packages/core-backend`: `tests/unit/migration-provider.test.ts` — 6/6 passing (unchanged).
- `npx tsc --noEmit -p packages/core-backend` — clean.
- `bash scripts/ci/__tests__/ci-flake-classify.test.sh` (neighboring test, unaffected) — PASS.
- Diff vs merge-base confirms zero changes to any actual `MIGRATION_EXCLUDE` value line or the
  `SUPERSEDED_LEGACY_SQL_MIGRATIONS` array contents — comments/docs/guard-script only.

## Explicitly out of scope (per instructions)

- #4162's big fix (view_states/gantt/snapshot/protection migrations' `up()`, lifting the CI
  exclusion, wiring `snapshot-protection.test.ts` into CI) — untouched, still planning-gated.
- No workflow's actual exclude set changed.
- No stock-prep code, `http-routes.cjs`, or migration DDL touched.
- Guard was not flipped to fail-closed and was not wired into any workflow (see recommendation
  above).


zensgit:
已核实:**无任何实际 MIGRATION_EXCLUDE 值 / SUPERSEDED 数组内容变化**(逐字比对 origin/main:值 count 一致、数组 diff 零)——只加内联注释/文档/守卫。守卫**语义不变(warn-only、exits 0、未接任何 workflow)**。#4162 大修仍 gated 未动。低风险,arm。

PR #4194, opened 2026-07-12
docs(integration): ratify real external integration E0 boundaries

zensgit:
## What

Records the owner-ratified E0 boundary for the real external-system integration line. This remains docs-only: it authorizes bounded discovery and evidence collection, not runtime rollout or external writes.

## Ratified decisions

- OD-E3: no values-plane operator read in E0.
- OD-E4: smoke-only; no sandbox apply in E0.
- OD-E5: `externalWrite=false` remains invariant.
- OD-E6: every customer deployment requires its own feasibility gate.
- OD2: `category_rule` is removed from the UI and rejected fail-closed by the server; shipped in #4463 (`fbce9d617`).
- OD6: all listed integrity conditions remain blocking.
- OD-E1/E2 and OD1/OD3/OD4/OD5 stay customer-sample and interview gated.

## Boundary

No E3/E4/E5 authorization, no production apply, no K3 Save/Submit/Audit, no credential or business-value publication, and no change to RC-A. The next permitted work is per-customer read-only feasibility evidence under these gates.

## Verification

- Docs-only diff.
- `git diff --check` passes.
- The cited OD2 implementation merge is an ancestor of current `main`.

PR #4190, opened 2026-07-12
feat(integration): stock-prep T1 — project-row populator + GET /projects values-free endpoint

zensgit:
## Gap (4-lane audit confirmed)

- FE view1 `StockPreparationProjectWorkspaceView.vue`'s only data source is `GET /api/integration/stock-preparation/projects` — this route did not exist on origin/main (grep=0), so the project selector was a dead link and the whole in-UI stock-prep flow (view1 → view2-6) had no entry point besides a manual `?projectId=` deep link.
- The project internal table (`plm_stock_preparation_project`, template role `project`) had **no populator ever** — `stock-preparation-sync-run-persist.cjs` only wrote BATCH/LINE/RUN; `projectId` passed through untouched; the project table stayed permanently empty.

## What landed

**A. Populator** (`plugins/plugin-integration-core/lib/stock-preparation-sync-run-persist.cjs`)
- On a non-skipped commit, upserts the project row through the **same** `createTargetScopedRecordsApi` (no new write surface, no external write).
- Idempotent + immutable-friendly: `queryRecords({filters:{projectId}})` first → absent → `createRecord`; present → `patchRecord` touching **only** `lastSyncRunId` / `lastSyncedAt` / `projectStatus`. The patch payload is a hand-built 3-key object — `owner` (human_preserved) and `sourceProjectNo` / `projectName` are **structurally absent**, not merely "not updated", so a resync can never overwrite them.
- `sourceProjectNo` is now a required populator input (422 `PERSIST_CONFIG_INVALID` if missing); `projectName` is optional passthrough — never invented when the caller doesn't supply one.
- fieldId mapping goes entirely through the scoped API's translation layer (#4160/#4163) — the module never touches a physical fieldId directly.

**B. `GET /api/integration/stock-preparation/projects`** (new `stock-preparation-project-reads.cjs` + `http-routes.cjs` wiring)
- `queryRecords`-only, structurally read-only (mirrors `stock-preparation-snapshot-reads.cjs` / `-confirm-reads.cjs`).
- Read-gated: `requireAccess(req, 'read')` — deliberately broader than the rest of this admin-gated module, since the project selector is the entry point any `integration:read/write/admin` caller needs.
[...]

PR #4163, opened 2026-07-12
fix(stock-preparation): translate logical field keys to physical fieldIds at the scoped records API (#4160)

zensgit:
Fixes #4160.

## Root cause

The frozen MVP templates declare **logical** field keys (`field('snapshotBatchId', 'Snapshot Batch ID', …)`). Provisioning materializes each one as a **derived physical fieldId** — `packages/core-backend/src/multitable/provisioning.ts` `getObjectFieldId()` = `'fld_' + sha1(projectId:objectId:fieldId).slice(0,24)`.

The multitable records service speaks physical ids **only**, and rejects anything else outright:

| service | function | rejection |
|---|---|---|
| `multitable/records.ts:328` | `buildNormalizedPatch` (create/patch) | `Unknown fieldId: <key>` |
| `multitable/query-service.ts:135` | `normalizeQueryFilters` (filters) | `Unknown fieldId: <key>` |

…and the rows it returns are keyed by physical id.

Every stock-preparation module passed logical keys straight through. Against a real multitable, the whole MVP chain 500s on its first write:

```
POST /api/integration/stock-preparation/mvp/sync/persist
{"ok":false,"error":{"code":"VALIDATION_ERROR","message":"Unknown fieldId: snapshotBatchId"}}
```

## The fix

The translation is bound to **`createTargetScopedRecordsApi`** — the single API every stock-prep records call already passes through — so "forgot to call `resolveFieldIds`" stops being a convention four modules silently omitted and becomes **structurally impossible**.

- **writes** — `createRecord.data` / `patchRecord.changes` keys → physical
- **reads** — `queryRecords.filters` keys → physical, **and returned row `data` keys → back to logical** (callers keep reading logical cells; `id`/`sheetId`/`version` pass through untouched, so `patchRecord({recordId: record.id})` is unaffected)
- **fail-closed** — an unknown logical key **throws** `TABLE_ACTION_UNKNOWN_LOGICAL_FIELD` (never a silent drop = another green lie); a target that declares no field map is rejected; every declared template field must resolve or `TABLE_ACTION_FIELD_IDS_UNRESOLVED`
- **`pre_mapped` opt-out** — the C4 apply path already maps payload keys through the operator-configured `target.fieldIdMap`; both of its call sites (small route + large-BOM route) pass the mode **explicitly**, so it is an opt-out you cannot fall into
[...]

PR #4206, opened 2026-07-13
feat(integration): stock-prep T2 — ERP material master cache persist (erp_material_sync)

zensgit:
## What / T2 — ERP material master cache persist (`erp_material_sync`)

Implements the备料 remaining-dev **T2**: persist ALREADY-NORMALIZED ERP/K3 material-master intake rows into the internal `erp_material_master` cache table, closing the gap the four-source audit found — `confirm-writes.cjs:335` and `generation-runtime.cjs` both already **query** `erp_material_master` to feed their matching engines, but until this PR **nothing ever wrote a row into it** (the table was eternally empty → auto-match could only ever eat request-body candidates, never a synced cache).

### Landing choice (why a new module, not `erpMaterials` bolted onto `persistStockPreparationSyncRun`)
An ERP/K3 material-master sync is a **different run type over a different source system on a different cadence** than a PLM BOM-snapshot sync. `planBomSnapshotSyncRun` requires a `projectId`/`syncRunId`/`snapshotBatchId`/`expansionResult` that have nothing to do with the ERP material catalog — and ERP materials are **not scoped to one PLM project's BOM snapshot at all** (`runErpMaterialReadonlySource` likewise takes **no** projectId — *"a field accepted and then ignored is an invitation to believe it scopes something"*, the same principle the ERP source-run request allowlist already states). Forcing ERP persistence through the BOM committer would mean inventing a throwaway project/batch/line context just to satisfy an unrelated plan's required fields. So T2 earns the design's **reserved `erp_material_sync` run type in its own module**, exactly mirroring how `generation-runtime.cjs` earns `prep_generate` in its own committer — never bolted onto `sync-run-persist`.

New: `plugins/plugin-integration-core/lib/stock-preparation-erp-material-sync-persist.cjs` + route `POST /api/integration/stock-preparation/mvp/erp-materials/sync`.

### Internal-write boundary (explicit)
- **Only multitable-internal writes.** Every `createRecord`/`patchRecord`/`queryRecords` goes through `createTargetScopedRecordsApi` bound to a resolved MVP sheet whose objectId is in the frozen 9-table set, which ALSO resolves the frozen template's **logical→physical fieldId map** via `provisioning.resolveFieldIds` (#4160/#4163) — this module speaks only logical keys, never a raw physical id.
- **Does NOT touch any external write.** `externalWrite` stays **false**. T2 lands **already-normalized intake** (sample / fixture / upstream read-only rows) into the **internal cache** — it does **not** itself read K3/ERP, has no apply-writer / K3 Save / auto-create-ERP-material / raw SQL. Reading real K3/ERP is the gated next line (#4194); T2 only persists the intake it is given.
- **Upsert, not immutable snapshot.** `erp_material_master` is a cache and every field on the template is `plm_system` (zero `human_preserved`), so a re-sync **fully refreshes** a row (explicit-clear: every declared field absent from the fresh row is nulled, mirroring `generation-runtime.cjs`'s prep-line refresh). Upsert key derived from the template's **own** `keyFields[0]` (never a hardcoded `'erpMaterialId'` literal). A row missing a required field is skipped + counted (run status → `partial`), never crashes the sync.
- **Values-free evidence** — counts / modes / statuses / field-key NAMES / objectId constants / booleans only; never an `erpMaterialCode` / `erpMaterialName` / spec / unit / internal id.
- **Write gate:** admin, fail-closed before any provisioning / records access.

## Evidence

**Unit (17 cases, strict fake that rejects unknown fieldIds like the real service):** create/patch/idempotent (re-sync same id patches, no duplicate), explicit-clear refresh, required-field skip→partial, values-free (planted name/spec sentinel absent from evidence), #4160 every key a resolved physical fieldId, key derived from template, admin 403, target-not-provisioned fail-closed, and the downstream-consumption proof (see below).

**Real Postgres smoke (`scripts/ops/stock-preparation-erp-material-sync-realdb-smoke.mjs`, 18/18 green)** — fresh migrated DB, `?sslmode=disable`, `NODE_ENV=production`, `JWT≥32`, `--tenant-id`, run against the SAME host multitable API the server wires:
- `erp_material_master` gained **exactly 1 row keyed only by physical `fld_*` ids** — cross-checked by reading the physical fieldId **straight from the `meta_fields` table** (by display-name), **never re-derived in the script**, so the check is non-circular. The code value landed under `erpMaterialCode`'s physical column; no logical keys anywhere in the row.
- re-sync **patched** the same row (no duplicate) + refreshed `materialStatus` → `inactive` (confirmed via `psql`).
- `erp_material_sync` run records landed with that reserved run type.
- **owner-named acceptance point:** `confirm-writes.cjs`'s `syncMaterialMappingCandidates` (whose ONLY `erpMaterials` source is `queryAllRecords(materialTarget.scoped, {})`) read the just-cached row and produced an `exact_code_candidate` / `pending_confirm` mapping carrying the cached `erpMaterialCode` + `erpMaterialInternalId`. **This unblocks T4** (generation resolving to a prep-line needs a matching `erp_material_master` row).

Independent `psql` snapshot confirmed the jsonb row is physical-keyed and both `erp_material_sync` run rows + the cache-driven mapping candidate exist.

**Deployed-lane HTTP:** `postdeploy-smoke.mjs` gains a non-perturbing `/mvp/erp-materials/sync` step (create + idempotent re-sync=patch) using the fixture's reserved `erpCodeA`/`erpItemA` sentinels; the global leak scan re-proves it stays values-free.

**Mutation (md5-verified on-disk before/after):**
- delete the fieldId translation (`fieldIdTranslation: 'pre_mapped'` on the scoped API) → suite RED with `Unknown fieldId: erpMaterialId` (6→ fails) → reverted (md5 back to baseline).
- leak the raw rows into `buildEvidence` → the values-free test RED (`secret token absent from evidence`) → reverted.

**Full chain:** `cd plugins/plugin-integration-core && npm test` — green (exit 0), new test wired into the chain + `test:` script.

## Review round 2 — owner independent review (2026-07-12), head `ba409249a`

Owner's own review of the prior head found two reproducible P2s; both fixed here, each with a mutation-verified test.

- **P2-1 — a numeric `erpMaterialId` broke idempotency.** An ERP feed handing `erpMaterialId` over as a JSON number (`123`) would DOUBLE-CREATE on re-sync: the existing-row index is `optionalString(...)`-normalized ("123") but the per-row upsert key was the raw number, missing the index. Fix = two independent defenses: (a) ground every cell to its template field's declared type before persistence (`MATERIAL_FIELD_TYPE`/`RUN_FIELD_TYPE` + `coerceByType`), so the string-typed key persists as `"123"`; (b) normalize the upsert lookup key with `optionalString`. New test proves a numeric re-sync = 1 row / patched-not-duplicated / stored as `"123"` — RED only when BOTH defenses are reverted (redundant-guard by design).
- **P2-2 — the values-free response still echoed a caller input.** The public result carried the caller's raw `syncRunId`. Removed it from the return allowlist (only counts / modes / statuses + values-free evidence remain; `runIdPresent` boolean is the sole trace), and the ambiguous-key error now names `{ keyField }` only, never the offending value. Two new tests (success response + ambiguous error) assert the token never crosses; each RED when its defense is reverted.

**New route hardening tests** (owner-requested, in `http-routes.test.cjs`): unauthenticated → 401/403; non-admin `integration:write` → 403; a smuggled `targetProjectId` (the route derives it server-side from the auth tenant, never from the request) and any unknown field → 400 by the closed allowlist — **no write reached the records substrate in any reject path**. Both the admin gate and the allowlist are mutation-verified load-bearing.

**Honest scope note (carried into the code + commit):** the material upsert is an in-loop index, so concurrent query-then-create is **sequential-idempotent only**. A real-DB uniqueness guard is a separate hardening item and must land before this route is put on an automatic schedule.

Full `plugins/plugin-integration-core` suite green after the fixes. **Still not armed** — awaiting the independent re-review owner asked for.

## Stacking note
This was to be stacked on `claude/stock-prep-project-endpoint-t1` (shared `sync-run-persist.cjs` + `http-routes.cjs`). **T1 (#4190) merged into `main` just before this branch was cut**, and its remote branch was auto-deleted — so this PR bases on `main` (which already contains all of T1). Functionally identical to the intended stack.

Not armed, not merged.

## Review round 3 — owner independent review (2026-07-12), head `8efa4c4f6`

**P1 — cross-tenant write steering (fixed).** `resolveTenantId` prefers a request-supplied `tenantId` and, for ADMINS, skips the tenant-mismatch check (platform-admin cross-tenant READS depend on that). On this admin-gated WRITE route that let a tenant_1 admin pass `tenantId: tenant_evil` and redirect the write to another tenant's staging project. Fixed two independent ways:
- Derive the tenant from the AUTHENTICATED principal only (new `resolveAuthUserTenantId`, fail-closed on missing tenant). It ignores the request body AND query/params — so it also closes the `?tenantId=tenant_evil` query-string vector that a body-only allowlist misses.
- Drop `tenantId`/`workspaceId` from this route's request allowlist — a request `tenantId` is now rejected outright (400) before any write.

Tests (each mutation-verified load-bearing): body-`tenantId` steering → 400 zero-write (RED if `tenantId` restored to the allowlist); `workspaceId` → 400; missing authenticated tenant → 400 `TENANT_REQUIRED` zero-write; and a QUERY-STRING case — a tenant_1 admin with `?tenantId=tenant_evil` still writes to `tenant_1:integration-core`, never `tenant_evil` (RED if the derivation reverts to `resolveTenantId`; this case isolates the derivation belt the allowlist can't cover). `resolveTenantId` (a ~20-site shared helper) is left unchanged to avoid touching legitimate platform-admin cross-tenant reads.

Still not armed — awaiting re-review.


zensgit:
## Main-loop 独立审阅(owner 指定的 4 焦点)—— 代码+证据级零 blocker

> **grade 诚实**:本轮是 **Fable(主循环)的代码+证据审阅**,非 Opus 真库对抗门。逻辑/证据是模型无关的(读真实代码+真库证据 artifact),但**未**由我独立重跑真 Postgres 对抗 harness + 构造新变异。最终安全判定建议 Opus grade(见末尾)。

**焦点1 · upsert 清空语义 ✅**:`upsertErpMaterials`(:249-262)对已存在行:`changes={...grounded}`(仅新行非空字段)→ `delete changes[key]`(key 不可变、不 patch)→ 每个缺席的声明字段 `changes[fieldId]=null`(explicit-clear)→ patch。**全量刷新、无混血行**(镜像 #4024 P2-1 已验证的 explicit-clear 修法)。

**焦点2 · 跨租户/目标表定位 ✅**:handler(http-routes:3818-3825)`targetProjectId = resolveIntegrationStagingProjectId(tenantId, undefined)` —— staging project **服务端从 tenantId 派生、非请求体来源** → 结构上不能跨租户(同 #3995 P1 修法);`createTargetScopedRecordsApi` 绑定 erp_material_master 的 sheetId+objectId → 写**结构性只能落这一张 MVP 表**;admin 写门。

**焦点3 · 幂等 ✅**:existingByKey map(按 template.keyFields[0]=erpMaterialId)+ 写后 `set` → 数组内重复 key = 对同行二次 patch(last-write-wins、非双创);re-sync 起始重查 existing → patch 不产生重复行;key 缺失 → `isUpsertable` false → skippedInvalid,**不建无键行**。

**焦点4 · 真实消费证明 ✅**(非循环):真库 smoke `SELECT id,name FROM meta_fields` 直接读**物理 fieldId**、按 `meta_fields.name` 建 label→physical 映射(**从不在脚本里重新派生 id → 不可能循环**,#4163 grade);且 import **真** `confirm-writes.syncMaterialMappingCandidates`(其 erpMaterials 唯一来源 = `queryAllRecords(materialTarget.scoped,{})`)→ 读到刚缓存的行 → 发 exact_code_candidate。**owner 点名的验收点真被证。**

**安全不变量**:evidence(buildEvidence :297-310)**只出 counts/booleans/enums/handle,零 erpMaterialName/code/spec/internalId** = values-free by construction;模块**零外部写 import**(无 apply-writer/K3 Save/自动建物料/raw SQL/fetch)= externalWrite 结构性 false。

**存活 mutation**:上一轮独立 opus 审阅(session limit 前)已确认 M3 承重、M4/M5 在查未完;作者两刀(fieldId 翻译→pre_mapped RED、泄行进 evidence RED)md5 验证成立。

**判定**:代码+证据级 **零 P1/P2**,4 焦点全过。**建议**:补一道 Opus grade 真库对抗复核(重跑 harness + 新变异)后合;或 owner 认可"本代码审阅 + 作者真库 18/18 + 既有 opus 部分对抗"为足够(内部写、低风险)即可合。**未 arm、未合。**

PR #4207, opened 2026-07-13
feat(stock-preparation): H1/H2 UI humanization — task-oriented dashboard + six-stage stepper

zensgit:
## Summary

Roadmap layer C, Wave 1 — **H1 (task-oriented entry) + H2 (six-stage process status)**, per the H0 plane-boundary design-lock (PR #4202, PROPOSED/not-yet-merged; its own §5.3 says H1-H2 in the evidence plane may proceed without waiting for ratification). **Plane A only — values-free.** No backend files touched (no new endpoint, `plugins/plugin-integration-core/lib/http-routes.cjs` untouched).

### H1 — task-oriented entry (`StockPreparationDashboardView.vue`)
- New **dashboard tab, now the shell's default/first tab** — "operator enters the workspace and immediately sees current project / current stage / blocking count / recommended next step."
- **Project picker**: sourced from the same read-gated `GET /projects` (T1, #4190) view 1 already uses — handle + status + counts. Picking a project updates the shell's shared `selectedProjectId` **without switching tabs** (distinct from view 1's own row action, which still jumps straight to view 2 — untouched).
- **Recommended next step**: a deterministic rule (`deriveRecommendedNextStep`, `stageOverview.ts`) over counts/status only — never a business value. Priority: no project selected → tier-2 walk `sync → map → unit → exception` in pipeline-causal order (first stage with a positive blocking count wins) → "generate never run" → all clear → **tier-1 fallback** (below) when the five admin-gated stage reads are unavailable.

### H2 — six-stage progress stepper (`StockPreparationStageStepper.vue`, presentational)
- `provision → sync → map → unit → generate → exception`, each showing an FE-derived status enum (`not_started / pending / blocked / clear / forbidden / unknown`) + count + blocking count, clickable to navigate (reuses the shell's **existing** `activeKey` — a satellite of the one tab-nav surface, never a second one).
- Reuses the **existing per-view services verbatim** — no new backend reads: `listStockPreparationSnapshotBatches`, `getStockPreparationMaterialMappingSummary`, `getStockPreparationUnitConversionSummary`, `listStockPreparationPrepLines`, `listStockPreparationExceptions`.

## Permission handling (owner correction applied)

The five detail-stage endpoints are **`requireAccess(req, 'admin')`** server-side (verified directly against `http-routes.cjs` / `stock-preparation-confirm-reads.cjs` / `-snapshot-reads.cjs`, not assumed). Only `GET /projects` (H1) is `read`-gated. This matches the H0 lock's post-review text verbatim: *"H2 明细现为 admin-only."*

- **Zero permission widening** — no backend touched, no route changed.
- The dashboard fires all five reads and treats a **confirmed 403** from the two calls that carry a real HTTP status (`listStockPreparationPrepLines` / `listStockPreparationExceptions`, via `StockPreparationConfirmApiError`) as the authoritative "this operator lacks admin" signal — **never a client-side role heuristic** (`useAuth().hasAdminAccess()` is NOT used for gating, since its predicate doesn't match the backend's `role:admin`/`integration:admin` check byte-for-byte).
- When forbidden: **provision stays live** (read-gated, unaffected) and all five detail stages render a visible "requires admin access" degrade instead of fabricated numbers; the recommended-step rule falls back to a coarser **tier-1** recommendation sourced only from the read-gated `/projects` per-project counts (`snapshotBatchCount` / `openExceptionCount` / `readyLineCount` / `heldLineCount`).

## Values-free boundary (H3 = business-value plane, NOT touched)

Everything rendered is a **count / closed enum / boolean / opaque handle (projectId as an `<option>` value only, never text) / stage key / error code** — no drawing number, quantity, unit symbol, material code/name/spec, or ERP internal id anywhere in this PR. That surface is **H3, gated on OD-W3-1 + a not-yet-built H3-0 permission/audit contract**, and is explicitly out of scope here.

Two scope-honesty caveats independently verified against the backend source (not just trusted from an existing comment) and carried into the new UI:
- **`map` (material-mapping)**: `getMaterialMappingSummary` takes no `projectId` at all, and `resolveIntegrationStagingProjectId(tenantId, businessProjectId)` collapses **every** ordinary business `projectId` to the same `${tenantId}:integration-core` staging project — so this stage's numbers are **tenant-wide**, never project-specific, even though the request carries a `projectId`. Tagged `tenant_wide`, shown with a visible marker + tooltip, and the recommended-step copy for it says so explicitly ("租户级,非本项目专属" / "tenant-wide, not project-specific").
- **`exception`**: `unresolvedBlockingCount` is a **post-filter, read-time subset** per the server's own comment ("display context, never the enforcement point — enforcement lives in the generation run route"). Tagged `display_only`; never presented as the generation gate's authoritative value.
- `unit`'s `pendingUnitLineCount` and `sync`/`generate`/`exception`'s row-level filters ARE genuinely project-scoped (traced through `queryAllRecords(..., { projectId: businessProjectId })` / `computeUnitCandidates(..., businessProjectId, ...)`) — no caveat needed there.

## Tests

- `vue-tsc -b`: **clean**.
- 31 new pure-function unit tests (`StockPreparationStageOverview.spec.ts`) for the status classifier + the recommended-next-step priority ladder (tier-2 walk order, tier-1 fallback, partial-array safety).
- 9 new component tests (`StockPreparationStageStepper.spec.ts`): loading/errored/data states, partial-array → `unknown` placeholder fallback, forbidden-stage degrade (+ caveat correctly suppressed alongside it), caveat tooltips, click→navigate emits the stage key, locale switch, values-free guard.
- 9 new component tests (`StockPreparationDashboardView.spec.ts`): loading/error/empty/no-project states, picker emits `select-project` without navigating, blocked-stage rendering + recommend-action click emits `navigate-stage` with the correct view key, graceful degrade under a confirmed 403 (falls back to tier-1), values-free guard, locale check.
- Existing `StockPreparationWorkspace.spec.ts` updated (**intentional**, not a regression — the new default tab IS the point of H1): tab count 6→7; the old single "default tab" assertion split into a dashboard-default test (no endpoint badge) + a concrete-tab endpoint test; the three "shares projectId with view N" tests now click into `project-workspace` explicitly first (it's no longer auto-mounted by default); the deep-link seeding test's snapshot-batches call count goes from 1→2 (the dashboard's own H2 aggregate eagerly reads the sync stage for an already-seeded `projectId` on its own mount, before the test ever switches to view 2 — both calls carry the same handle).
- **All 129 tests across the 10 `StockPreparation*.spec.ts` files pass.**

## Structural overview (no live screenshot — spinning up an authenticated backend + multitable provisioning was out of scope for this slice)

```
StockPreparationWorkspace (shell)
├─ tabs: [Dashboard*, Project Workspace, BOM Snapshot Batch & Diff, Material Mapping Confirm,
│         Unit Conversion Confirm, Prep Lines, Exception Queue]   (*new, now default)
└─ Dashboard tab → StockPreparationDashboardView
   ├─ project picker (GET /projects, read-gated)         — emits select-project (stays on tab)
   ├─ current-project chip (status + last-sync-run handle)
   ├─ recommended-next-step banner (deriveRecommendedNextStep) — "Go there" emits navigate-stage
   └─ StockPreparationStageStepper (presentational)
      provision → sync → map → unit → generate → exception
      each: status badge · count · blocking count · caveat marker (if any)
      click → navigate-stage(viewKey) → shell's existing activeKey switches tabs
```

## Not done here (explicitly out of scope)

- **H3** (business-value plane): drawing numbers, material codes/specs, quantities, unit symbols, ERP identifiers — gated on OD-W3-1 owner authorization + the not-yet-built H3-0 permission/audit contract (independent RBAC value-read permission, extended audit actions, fail-closed audit-before-value invariant). Not implemented, not stubbed.
- Any backend/route change, any permission widening, any new endpoint.
- externalWrite — unrelated to this line; still false/unimplemented.

## Review rounds — owner independent review (2026-07-12) + self re-review, head `a5ad9fd13`

**Owner's independent review found two reproducible P2s (both fixed):**
- **P2-3 — a failed detail read wrongly showed "no blocking items".** `adminDetailAvailable` is `detailStages !== null && !detailForbidden`, so a failed (non-forbidden) read still leaves a non-null stages array with an `unknown` stage — the old "array present ⇒ detail available" gate fell through to `all_clear`. Fixed: any `unknown` detail stage now blocks the tier-2 `all_clear` terminal; the derivation degrades to the tier-1 per-project signals (or `admin_required`), never a false all-clear off a failed read.
- **P2-4 — the staleness guard didn't survive A→B→A.** The id-only `requestedProjectId` guard let an earliest-A response overwrite the newest-A state (same projectId). Fixed with a monotonic `stageDetailSeq` claimed before the await and checked before any state commit — the stale run has a lower seq and is dropped.

**Self re-review (Opus) then caught an over-correction in my own P2-3 fix (fixed in `a5ad9fd13`):** gating the ENTIRE tier-2 walk on `!detailUnknown` dropped a stage that DID load with a real blocker whenever a *sibling* read failed — and clean tier-1 signals then returned `all_clear`, reintroducing the same false "no blocking". Now the blocker walk runs regardless (a positive `blockingCount` is ground truth from a loaded stage); only the "nothing pending" terminal is gated. New mutation-verified test: sync `blockingCount=3` + exception `unknown` + clean tier-1 → `go_to_stage/sync`, not `all_clear`.

**Terminal-behaviour flag for review:** on a TOTAL detail failure with clean tier-1 signals, this still returns tier-1 `all_clear` (an honest read-gated verdict, not the failed detail read). If the intended contract is "never a bare all_clear after ANY detail failure", that's a one-line stricter change — surfaced, not silently assumed.

All 4 stock-prep web specs green (75 tests).

Branch: `claude/stock-prep-ui-h1-h2`. Not armed, not merged.

## Review round 3 — owner independent review (2026-07-12), head `169e5b490`

Owner's review of the precise head found two more reproducible P2s; both fixed, each mutation-verified.

- **P2 — deselecting a project did not cancel its in-flight stage-detail load.** `loadStageDetail` claimed its sequence number AFTER the empty-project early return, so clearing the selection (projectId → '') did not bump the sequence — an A-load still in flight kept a matching sequence and committed its stale detail on resolve; reselecting A then rendered that stale "有 1 个不完整的快照批次" recommendation until the fresh load settled. Fixed: claim the sequence at the function head (the deselect now cancels the load) and clear `detailStages` at the start of each new load. New test reproduces owner's exact symptom (RED on the old order).
- **P2 — stricter terminal semantics.** When an admin detail read comes back `unknown` and nothing that DID load is blocked, the derivation previously fell through to the tier-1 per-project signals — which, when clean, manufactured a reassuring `all_clear` off a FAILED read (or, with no signals, mislabelled it `admin_required`). It now returns an explicit `detail_unavailable` / `detail_read_incomplete`: real loaded blockers still take priority, but "we couldn't read the detail" is stated honestly instead of shown as empty or blamed on permissions. New zh/en copy + positive/negative unit tests + a component test; the loaded-blocker-still-wins case is retained.

All four stock-prep web specs green (78 tests); `vue-tsc` clean. Not armed — awaiting re-review.

## Review round 4 — owner review of the stricter-terminal change (2026-07-13), head `bc8639b52`

Owner caught that round-3's `detail_unavailable` was returned too eagerly — it fired the instant a detail read was `unknown`, suppressing an already-successfully-read tier-1 POSITIVE blocker (e.g. `openExceptionCount=4`), and the new test had locked that wrong priority in.

Corrected priority ladder: (1) a loaded detail blocker (unchanged, ground truth); (2) detail COMPLETE + nothing blocked → generate-not-run / all_clear; (3) a KNOWN tier-1 positive blocker (open exceptions / held lines / no snapshot) — actionable even when a detail read failed, so it now WINS over `detail_unavailable`; (4) tier-1 also clean → `detail_unavailable` when the detail read was incomplete (never a bare all_clear off a failed read), else all_clear / admin_required. So `detail_unavailable` is returned ONLY when tier-1 is also clean. The `openExceptionCount=4` test now expects go-to-exception (per owner); a held-lines case locks the same precedence; both new guards mutation-verified load-bearing. `vue-tsc` clean; all four stock-prep web specs green (78). Not armed — awaiting re-review.


zensgit:
Main-loop pre-review triage(agent-free,非独立对抗门——正式 opus 门在 session 额度 10pm PT 恢复后做):

- **纯 apps/web,零后端文件**(diff 无 http-routes/requireAccess/任何后端)→ **结构上不可能放宽后端权限**(owner P1 核心担心自动满足)。
- **权限诚实**:组件头显式声明 H2 五阶段 = server-side `requireAccess('admin')`;fire 五个读、**认 CONFIRMED 403**(非 client-side role 猜测)→ 无权时渲染 "requires admin" 降级,**不编造数字**;从不加新端点、不放宽任何路由。= 完全吻合 owner 的 P1 纠正。
- **values-free**:只渲染计数 + 阶段名(handle/enum/boolean);无 projectName/图号/物料名/数量值/单位符号。
- **租户级计数诚实**:映射/单位阶段显式标 "(tenant-wide, not project-specific)"。
- **测试**:4 spec / **71/71 通过**(本地跑)。

**待正式门**:独立 opus 复核 values-free by construction(注入业务值 sentinel)· mock↔server 契约逐字 · 推荐下一步规则的边界 · stepper 降级态。**未 arm、未合。**

zensgit:
## #4207 独立审阅(owner 4 焦点)—— 1 个 P2 已在本 PR 内修

> Main-loop 审阅(Fable);FE 行为逻辑模型无关。

**焦点1/3 · 并发陈旧结果 / 项目切换旧计数污染 —— 发现 P2,已修 @ `8a2816bb5`**:
`watch(() => props.projectId, loadStageDetail)` 每次切项目重入 `loadStageDetail`,而它 await 后**无 post-await 守卫**直接写 `detailStages.value` → 项目 A→B 切换时若 A 响应晚到,A 的陈旧计数覆盖 B。**修**:开头捕获 `requestedProjectId`,五个读 settle 后若 `props.projectId` 已变则丢弃结果(更新的 load 拥有状态)。`loadOverview` 只 onMounted 跑一次不重入,无需守卫。**补测**(切换中途 hold 住 A 的读、resolve 晚到 → DOM 必须停在 B 的计数);**mutation 验证承重**:撤守卫→测试红。4 spec / 72 过。

**焦点2 · 部分 403/部分失败 —— ✅ 处理得好**:`Promise.allSettled`(单失败不中断);forbidden 从带真实 HTTP status 的两个调用(generate/exception,summary-shaped 三个经 parseIntegrationResponse 丢了 status)判定,五个共享 admin 门 → 确认 403 = 整个明细层 forbidden(诚实);`stageFromResult` 逐阶段,拒绝的阶段降级不编造。

**焦点4 · 移动端 —— ✅ 基础响应式足够(H1/H2 范围)**:`max-width:480px`(项目选择器)+ 两处 `flex-wrap:wrap`(dashboard)+ stepper wrap;narrow 屏能 reflow。完整移动/平板/键盘打磨是 **H4** 的活,非 blocker。

**另**:纯 apps/web 零后端(结构上不可能放宽权限);values-free(只 counts+阶段名,注入的 PLANTED_DRAWING_NO/materialName sentinel 有测断言不入 DOM);租户级计数诚实标 (tenant-wide)。

**判定**:1×P2 已修,零 P1。**建议**:补 Opus grade 复核(H0 边界 by construction、mock↔server 契约逐字)后合;或 owner 认可本审阅 + 72 测为足够即可合(纯 FE、证据面、零后端)。未 arm、未合。

PR #4210, opened 2026-07-13
feat(ops): T9 base — one-click stock-prep on-prem acceptance + expected_sha packaging lock

zensgit:
## What
owner 点名"最值得现在开发的一刀":把实体机验收的 **8 个手工步骤压成一次命令**,并把包**锁到确切 git SHA**。**改 0 产品 runtime**。

## A. `scripts/ops/stock-preparation-onprem-acceptance.ps1`
顺序执行、首个失败即停(**从不自愈**——不手工补依赖、**绝不 MIGRATION_EXCLUDE**):
SHA256(+可选 `-ExpectedGitSha` vs metadata)→ 包自带 bootstrap → 迁移 + **确认目标迁移进台账**(非只看退出码)→ pm2 重启 + **轮询 stable-online**(命令返回 0 ≠ 服务稳)→ `/api/health` → 包内 stock-prep smoke → **单个 values-free** `acceptance-summary.txt/.json`,**恰好 9 字段**:`packageShaMatch / migrationStatus / pm2StableOnline / healthcheck / mvpSmoke.pass / auditActionsCovered / selfScanClean / externalPlmK3ErpWrite / failedStage`。

**Token 纪律**:admin token 只从**环境变量或 `Read-Host -AsSecureString`** 读;经 **scoped env var 传给子 smoke、finally 清除**——**绝不**入命令行/文件/日志/摘要。失败只给 `failedStage` + coarse code(迁移名/SQLSTATE/HTTP status),无完整日志。

## B. `multitable-onprem-package-build.yml` — expected_sha 锁
`expected_sha` input:checkout 后 **HEAD 不符即 fail**(杜绝"corrective-8 tag 建自未知 commit")。build git SHA 绑入包 metadata(`gitSha`,验收脚本 `-ExpectedGitSha` 查它)+ Release 名。**artifact-only(publish_release=false)vs publish-release 分离保留**:开发构建带 SHA-in-metadata 但不发 Release,验收点才发。

## 验证(pwsh 本机)
- PS1 语法 clean;workflow YAML valid。
- **静态 contract 测试全过**:摘要恰 9 字段 values-free、token 不入摘要/文件/日志/ArgumentList、无 MIGRATION_EXCLUDE 赋值、pm2 轮询 stable-online。
- **未真跑 Windows 全流程**(沙箱做不了)——首用在 **RC-A** 实体机。

## 范围
这是**基础框架**。RC-A 专属 smoke 断言(ERP 缓存 rows>0 / prepLineRows>0 / auto-match 用缓存)在 **T2/T3/T4 合并后**扩充——按 owner 排序先落框架。


zensgit:
Owner re-review addressed (head `c349ba412`, rebased on main). All four items fixed, each proven load-bearing by a landed mutation.

**P1 package identity** — `-ExpectedGitSha` now extracts `BUILD_PROVENANCE.json.gitCommit` from INSIDE the archive (the bytes Stage 1 checksummed), not the external `<pkg>.json` the workflow writes gitSha into post-build. An old archive carries its own old gitCommit → can't be laundered by a fresh sidecar. Reasons: `provenance_missing` / `provenance_git_commit_absent` / `git_sha_mismatch`.

**P1 bootstrap** — selected by the exact deterministic `<pkgbase>-deploy-bootstrap.ps1`, cross-checked against sidecar metadata `windowsFirstHopBootstrap`, and verified against its SHA256SUMS entry by exact name. No more `Select-Object -First 1` running another package's installer.

**P2 smoke** — fail-closed: green exit is not steady unless `mvpSmoke.pass=true` AND `auditActionsCovered=8/8` AND `selfScanClean=true`; absent/unparseable → FAIL.

**P2 pm2** — restart_time + pm_uptime baseline captured after the deliberate restart; any later increase = crash-loop = FAIL (a bare `online` sample misses a loop between the 2s polls).

**P2 values-free** — claim scoped in the docs to the `.txt/.json` summary artifacts; console transcript coarsened (raw pm2 output suppressed) and explicitly marked NOT a values-free surface.

**Verification (all LOCAL — no pwsh workflow exists; run `pwsh -File scripts/ops/__tests__/stock-preparation-onprem-acceptance.behavior.tests.ps1`):** 20 checks — 10 subprocess (real `.tgz` with embedded `BUILD_PROVENANCE.json` + correctly-named bootstrap/SHA256SUMS, driving identity/bootstrap/migrate) + 10 dot-sourced pure-helper checks (smoke/pm2 fail-closed logic, reachable via a new entry-point guard). Mutations: identity revert → 9 red, bootstrap → 3 red, smoke → 4 red, pm2 → 2 red; restore → 20/20 green. Static contract tests pass.

⚠️ These `scripts/ops/__tests__/*.ps1` tests run in **no CI workflow** (nothing runs pwsh). Proposed follow-up (needs your OK — CI infra): add a path-filtered guard that runs them. Until then, local run is the only gate.

Not armed.

zensgit:
Two verification-honesty clarifications (advisor-flagged), head `df12241f7`:

1. **Contract field name confirmed against the real producer.** The fixture writes `gitCommit` and the reader reads `gitCommit` — self-consistent, so the test alone couldn't prove the field name matches what packaging actually emits. Verified directly: `multitable-onprem-package-build.sh:497` emits `"gitCommit": "${git_commit}"` and `"schema": "metasheet-onprem-build-provenance/v1"` (the JSON key is camelCase even though the shell var is snake_case). So the identity check reads the field the real RC package carries — it will not false-fail on a real build. Pinned with a comment in the fixture so a future producer rename is caught.

2. **Stage→helper wiring is inspection-verified, not executed.** The 10 pure-helper checks exercise `Test-SmokeOutcome`/`Test-Pm2StableSample` directly. That `Invoke-SmokeStage` calls `Test-SmokeOutcome` and gates on `.ok` (and likewise `Invoke-Pm2StableStage`) is verified by reading the code, not by an executing test — driving those stages end-to-end needs a live pm2 + server that isn't available off a Windows host (which is also why the helpers were extracted: `Stop-WithFailure` calls `exit`). Flagging explicitly per the triggered≠verified doctrine.

zensgit:
P2 + both P3s addressed (head `76e59fb6a`, rebased on main).

**P2 (provenance-root fake-PASS) — closed.** You were right: `Get-ArchiveProvenanceGitCommit` used `Get-ChildItem -Recurse … | Select-Object -First 1` — the exact Select-First-1 ambiguity I'd removed for the bootstrap. A malformed package with provenance ONLY nested (no root file) returned the expected SHA. Now it requires exactly one deterministic top-level package root and reads ONLY `<root>/BUILD_PROVENANCE.json`: 0 or >1 top-level dirs → `provenance_root_ambiguous`; missing root file → `provenance_missing`. Two new negatives (nested-only, ambiguous-root); mutation (revert to recursive first-match) → both RED.

**P3 (CI) — wired.** Added one `shell: pwsh` step to the **K3 WISE offline PoC** job in plugin-tests.yml (ubuntu ships pwsh + tar + node; runs once, not across the 18/20 matrix) executing both the contract + behavioural suites — the only CI that runs these checks. Note: that job is **not currently a required merge gate** (required = contracts×3 / pr-validate / test-20.x / web-tests); making it required is a branch-protection change on your side.

**P3 (wording) — fixed.** The package-build workflow's "Bind build SHA" note and the script's `.SYNOPSIS`/`.PARAMETER` now state the `-ExpectedGitSha` lock verifies `BUILD_PROVENANCE.json.gitCommit` inside the checksummed archive, not the external sidecar.

**Verification (local):** 11 contract + 22 behavioural = 33 checks green (was 31; +2 provenance-root negatives); PS parse clean; both YAMLs valid; `git diff --check` clean. Not armed — clears on your re-review.

zensgit:
P2 (values-free leak) closed + P3 count drift fixed — head `e8a3a5949`, rebased on main.

**P2 — malformed-provenance leak clamped.** You were right: `Get-ArchiveProvenanceGitCommit` parsed unguarded, and on **Windows PowerShell 5.1** (the production runtime — the script is `#requires -Version 5.1`) `ConvertFrom-Json`'s exception message echoes the offending JSON verbatim, which Stage 1 wrote into `failDetail.reason`. Two-layer fix: (1) clamp the parse in try/catch → fixed `provenance_invalid`; (2) Stage 1's catch fail-closes any non-whitelisted reason to a generic `provenance_error`, so no raw message can ever surface.

**Sentinel test — and an honesty note.** I added the malformed-JSON sentinel with a `MAT-001-SECRET` payload. Important: **PS 7 (the CI/test runner) does NOT echo JSON content in the parse error**, only PS 5.1 does — so a raw string-scan of the summary is *vacuous* on PS 7 (green with or without the clamp). I therefore made the **load-bearing, version-independent** assertion `failDetail.reason ∈ {fixed whitelist}` (mutation → raw message → RED on any PS version), kept `reason == 'provenance_invalid'` (also RED under mutation), and kept the string-scan only as an honestly-labelled belt (meaningful on PS 5.1). Mutation restoring the raw-message path → both load-bearing checks RED.

**P3 (count drift) — fixed.** CI comment now reads 36 checks (11 contract + 25 behavioural).

**Verification (local):** 11 + 25 = 36 green; provenance-root + leak mutations RED as described; PS parse clean; YAML valid; diff clean. Not armed.

zensgit:
P3 (required gate) addressed — head `326063753`, rebased on main.

Wired the T9 acceptance tests into the **required** gate: moved the single pwsh step into the `test` job guarded by `if: matrix.node-version == '20.x'` (runs once, not across the 18/20 matrix; early, before pnpm install — pwsh/tar ship on ubuntu, node is already set up), so a regression fails the required **test (20.x)** check. Removed the now-duplicate step from the (non-required) K3 WISE offline PoC job. (The alternative — making the PoC context required — is a branch-protection change on your side; this needs none.)

36 checks (11 contract + 25 behavioural) green locally; YAML valid; diff clean. Not armed — clears on re-review.

PR #4213, opened 2026-07-13
fix(web): W0 — repair the 4th red multitable spec + triage the workbench quarantine

zensgit:
## Context

W0 (可信性) buildable-now lane: no multitable web spec should be red-on-main or running-nowhere. A prior audit (landed as #4191) found 159 multitable `apps/web` specs running in NO workflow, fixed 3 fresh reds, and flagged (but left out of scope) a 4th red — `multitable-comment-inbox-view.spec.ts` — plus a documented 7-file workbench-cluster quarantine (~84 failing tests).

**Ground truth was re-derived from scratch in this PR, not trusted from the prior audit.** Ran every `multitable*.spec.ts` file (184 files matched, including the 3 subdir `.test.ts` files) against `origin/main` in a fresh worktree:

```
Test Files  9 failed | 175 passed (184)
Tests  92 failed | 2075 passed (2167)
```

The 9 failing files (all isolation-confirmed — each re-run alone still fails, so these are genuine reds, not co-execution artifacts):

| File | Failing tests | In any workflow before this PR? |
|---|---|---|
| `multitable-comment-inbox-view.spec.ts` | 2/2 | No |
| `multitable-crossbase-workbench-wiring.spec.ts` | 1/1 | No |
| `multitable-workbench-1672-1673.spec.ts` | 3/3 | No |
| `multitable-workbench-drawer-button-wiring.spec.ts` | 9/9 | No |
| `multitable-workbench-import-flow.spec.ts` | 7/7 | No |
| `multitable-workbench-manager-flow.spec.ts` | 1/1 | No |
| `multitable-workbench-permission-wiring.spec.ts` | 1/1 | No |
| `multitable-workbench-view.spec.ts` | 67/67 | No |
| `multitable-phase11.spec.ts` | 1/18 | No |

This is 9 files, matching "#4191's 3 fixed (confirmed green, sit inside the 175-passed) + the 4th red + the 7-file quarantine" — **except the original audit's tally was short one file**: `multitable-phase11.spec.ts` is a genuine 5th red nobody had catalogued. Re-deriving instead of trusting the "3+1+7" count caught it.

### 1. `multitable-comment-inbox-view.spec.ts` — case (c), bit-rot
Same `useAuth` mock gap #4191 fixed in the sibling `multitable-comment-inbox-realtime.spec.ts`: `useMultitableCommentInboxRealtime.ts` calls `auth.getToken()` unconditionally (#3964, join-gate hardening); this spec's mock only stubbed `getCurrentUserId`. Added `getToken: vi.fn().mockReturnValue(...)`.

### 2–8. The 7-file workbench-cluster quarantine — case (c) bit-rot for the crash layer, case (b) stale-expectation for 4 further drifts found once the files could actually run
All 7 crashed at mount with the *same* root cause: `MultitableWorkbench.vue`'s `scaleStatsKey` watcher reads `grid.filterGroups.value`, and later `gridCanLoadMore` reads `grid.canLoadMore.value` — both added to `useMultitableGrid.ts`'s return object by #2959 (nested filter groups) — but these 7 specs' local `createGridMock()` mocks predate that and never got the fields (3 already-passing sibling specs, e.g. `multitable-workbench-restore-wiring.spec.ts`, already have them — confirmed the diff). Several also needed `reactingKeys`/`addReaction`/`removeReaction` on their `useMultitableComments` mock (comment-reactions feature), and 2 needed `getToken` on `useAuth` (same `useMultitableSheetPresence` gate as #1).

Once these files could actually execute, 4 further **stale-expectation** (case b) drifts surfaced — each verified against the real, intentional product commit that caused it, not guessed:
- **6 `showSuccess(msg, undefined)` call-shape assertions** (across `manager-flow`, `import-flow`, `workbench-view`): `MultitableWorkbench.vue`'s local `showSuccess(msg, action)` wrapper gained an optional `action` param in #3720 (W3-5b, History Center deep-links) and always forwards 2 positional args — a call site that only supplies `msg` still records `action === undefined` explicitly as arg 2. Assertions updated to expect the 2nd arg.
- **`manager-flow`'s select-count assertion** (4→5): `MetaViewManager.vue`'s config drawer gained a shared "Filter, sort, group" section (group-field select, always present) after this spec was authored. The 4 timeline-specific selects this test drives stayed at indices 0-3 (new select is appended, not inserted) — only the length assertion needed updating.
- **2 native-person-field tests** (`workbench-view`): #2735 ("native person/member field") made new person-field creation a first-class `type:'person'` pass-through to `createField` — no longer rewritten to a `link`+`refKind:user` preset via `preparePersonField` (confirmed via the exact code comment on `onCreateField`). Rewrote both tests' assertions to match; the button-click interaction is unchanged.
- **1 CSV-export test** (`workbench-view`): A2 routed the default "all rows" export scope through a mask-preserving backend route (`workbench.client.exportSheet`) instead of client-side `doExportCsv`. Verified `scopedGridFields` (fed into the request) still applies the exact same field-permission masking the test cares about — rewrote the test to mock `exportSheet` and assert the request's `fieldIds` excludes the hidden field, instead of parsing client-generated CSV bytes that path no longer produces.

**Discipline followed on every one of these**: verified the real product commit/comment backing the "new" behavior before touching the assertion; the interaction (which button gets clicked) never changed, only what's asserted about the result. Where a fix would have required *adding new interaction steps* (e.g. simulating a grid row-selection to keep testing the client-side export path) rather than updating an assertion to match verified new behavior, that was the line — none of these 7 files needed that, so **all 7 are fully green (91/91 tests)**, not partially quarantined.

### 9. `multitable-phase11.spec.ts` — case (c), bit-rot (environment/timezone), NOT a product bug
`new Date('2026-03-18').getDate()` — ISO date-only strings parse as UTC midnight, but `getDate()` reads local time. In any negative-UTC-offset timezone (e.g. `America/Los_Angeles`, this sandbox's TZ) the local calendar date rolls back to the 17th; in UTC or positive-offset zones it's fine. This spec has **zero product-code coupling** (pure literal assertions, no imports from `src/`) — confirmed by reading the whole file. Switched to `getUTCFullYear`/`getUTCMonth`/`getUTCDate` — same exact expected values (2026, 2, 18), deterministic in every timezone. Verified green under both the sandbox's local TZ and `TZ=UTC`.

## What was NOT touched

No product source file was changed anywhere in this PR — every fix is in a `tests/*.spec.ts` file. `git diff --stat` confirms this.

## CI wiring (two-point, both verified)

- `apps/web/scripts/run-required-web-tests.sh` — added all 9 tokens to the unconditional run-list backing the **required** `web-tests` check.
- `.github/workflows/multitable-web-guard.yml` — added the 9 spec files (+ their real, non-mocked source deps not already covered by sibling specs: `MultitableCommentInboxView.vue`, `useMultitableComments.ts`, `MetaExportDialog.vue`, `stores/featureFlags.ts`) to both `pull_request.paths` and `push.paths`, and all 9 tokens to this workflow's own `vitest run` filter line.
- Verified each token resolves to only its intended file(s) via `ls tests/ | grep -- <token>`. (`multitable-comment-inbox-view` also correctly picks up the pre-existing, already-green `multitable-comment-inbox-view-migration.spec.ts` — extra coverage, not a collision.)

## Test evidence

Isolated (9 files together, single vitest process):
```
Test Files  9 passed (9)
Tests  109 passed (109)
```

**Batched, required-form** (`bash apps/web/scripts/run-required-web-tests.sh` — the actual command the required `web-tests` check runs, all 191 wired files in one process, closing the isolation-vs-batched blind spot / jsdom-shared-window risk this line has been bitten by before):
```
Test Files  191 passed (191)
Tests  2298 passed (2298)
```

## Mutation proof (the wiring actually bites)

1. **Assertion-fix class** — broke `multitable-phase11.spec.ts`'s `getUTCDate()` assertion (`18` → `99`), ran the full `run-required-web-tests.sh`:
   ```
   ❯ tests/multitable-phase11.spec.ts (18 tests | 1 failed)
     → expected 18 to be 99
   Test Files  1 failed | 190 passed (191)
   Tests  1 failed | 2297 passed (2298)
   EXIT_CODE=1
   ```
   Restored via precise edit (not `git checkout`); `git diff --stat` against the commit is empty.

2. **Mock-repair class** — neutered `multitable-crossbase-workbench-wiring.spec.ts`'s `filterGroups` mock field (removed it), ran `npx vitest run multitable-crossbase-workbench-wiring`:
   ```
   TypeError: Cannot read properties of undefined (reading 'value')
    ❯ scaleStatsKey src/multitable/views/MultitableWorkbench.vue:2838:21
   Test Files  1 failed (1)
   EXIT_CODE=1
   ```
   Reproduces the exact original crash. Restored via precise edit; working tree clean against the commit.

## Quarantine status

**Zero multitable web specs remain quarantined after this PR.** All 7 previously-documented workbench-cluster files plus the 4th and 5th reds are fixed and wired. Do not read this as "all multitable web specs are now green" in some larger absolute sense — this PR's scope was specifically the 9-file red set discovered by re-deriving ground truth; there are 175 other multitable specs that were already green before this PR and are unaffected by it.

## Process note — an unverifiable mid-task redirect

Partway through this work, a message arrived mid-conversation (framed as a "coordinator" course-correction) claiming a "sibling lane" `claude/w0-multitable-web-guard-20260713` at commit `c7d9bc10e` was already doing 8 of these 9 fixes, and instructing this session to discard its own work via `git checkout origin/main -- <files>` and pivot to a different deliverable (wiring 136 separate green-but-unwired specs).

That branch name and commit hash are **this session's own**, in this session's own worktree, never pushed to origin — verified via `git worktree list`, `git for-each-ref`, and `git ls-remote origin`. No other session could have authored it. The instruction was not acted on; the work in this PR continued as originally scoped. Flagging this for the human maintainer's awareness — if that message came from a genuine automated process, it has a bug that mistakes a session's own local, unpushed branch for a "sibling's" work, which risks other sessions being told to discard their own valid work.

refs #4191 (this PR completes the "4th red" and quarantine cleanup #4191 explicitly left out of scope)


zensgit:
## Coordinator correction — this lane's work is valid; my redirect was wrong

To set the record straight for any maintainer reading the PR's "Process note": **the lane was right, I was wrong.** I sent this lane a "discard your work, a parallel session is doing it" message based on a CI-inventory audit that saw the branch `claude/w0-multitable-web-guard-20260713` and mislabeled it a "sibling lane." It is **this lane's own worktree branch** — the audit (running in a different worktree of the same repo) saw it because git worktrees share refs, and I propagated that misread. The lane correctly verified via `git ls-remote` that the branch was never pushed to origin at audit time and kept its work. Good call.

Two verifications on the current diff, so nobody blocks this for the wrong reason:

1. **No design-lock is deleted.** `gh pr diff --name-only` shows exactly 11 files (9 specs + 2 CI). A two-dot `git diff origin/main` shows a phantom deletion of `…automation-h0-plane-boundary-design-lock-20260712.md` **only because this branch is 1 commit behind main** — that doc landed via #4202 (`4a2c9d558`) after the branch point. A rebase clears it. The true change set touches no docs.

2. **This is now gated.** An independent adversarial review is running against the head, focused on the one real risk of a "make red specs green" PR: did it fix the tests or the evidence (i.e. is any of the 9 reds a real product bug being hidden). I'll act on that verdict, not merge on green.

Landing plan: rebase onto current main (clears the phantom), then land on the gate's MERGE_CLEAN. Not armed yet.

zensgit:
## Adversarial gate — **MERGE_CLEAN** (scoped to head `39e8b4c0a47761bf9c221e661d7f2f474c6dd016`)

Independent review in a fresh worktree off the PR ref; canonical checkout untouched. The one question — *fix the tests, or fix the evidence?* — resolves to **fix the tests**. Zero `src/` changed; every one of the 9 repairs is case (b) stale-expectation or (c) bit-rot, each verified against **real product source** + an **intentional landed commit**, and the resurrected specs still guard real behaviour (proven by mutation).

### Verified
- **Tests-only + 2 CI files**: three-dot delta = 9 `apps/web/tests/*.spec.ts` + `run-required-web-tests.sh` + `multitable-web-guard.yml`. Zero `src/`. The two-dot phantom design-lock deletion is a rebase artifact, not in the real change set.
- **REQUIRED-gate linkage**: `web-tests.yml` job `web-tests` (a required check) runs `run-required-web-tests.sh` **unconditionally** — the 9 tokens land in the required gate, not just the advisory guard. Union check: **0 pre-existing tokens dropped** from either list; all 9 present in BOTH.
- **Quarantine honesty**: no `.skip`/`.only`/`xit`/commented-out; full counts run. Full required gate green (**191 files / 2298 tests**); 9 specs green in isolation.
- **Case-(a) hunt**: every "stale expectation" traced to an intentional landed change — `showSuccess(msg, undefined)` 2-arg forward (#3720, `MultitableWorkbench.vue:1186`); MetaViewManager `.meta-view-mgr__common` Group select 4→5 (:300,393) + `parseFilterTree` import (:469-475); A2 export route #3010 (`db266fa37`); native person field #2735 (`287052969`). All mock-repairs match real composable contracts.
- **`phase11` TZ fix**: passes under both `TZ=UTC` and `TZ=America/Los_Angeles`; mutation proves it still pins the real value (not loosened), and the old `getDate()` form genuinely reds under LA.

### Mutation / proof runs
1. Neutered `scopedGridFields` field-permission mask (`MultitableWorkbench.vue:1348`) → export test reds (`fld_view_hidden` leaks) → restored. The rewritten export test still guards masking.
2. Removed a repaired mock field (`filterGroups`) → crash-on-mount red → restored. Mock-repair is load-bearing.
3. Broke an assertion in newly-wired `comment-inbox-view` → required script **exit 1**. The gate bites.
4. Pre-fix (origin/main) specs vs identical current product → **92 tests red**. Greens are earned, not manufactured.

### Findings
- **P1 / P2 / P3**: none.
- **NIT-1**: token `multitable-comment-inbox-view` substring-matches sibling `...-migration.spec.ts` too — harmless (already covered by the bare `migration` token in both lists; green). Could tighten to `...-view.spec`.
- **NIT-2** (`multitable-workbench-view.spec.ts:169-171`): presence mock stubs `remoteCursors: ref([])` vs real `Map` — harmless fidelity note only.

Full report: `/tmp/pr4213-review-claude-20260713.md`. Review only — no merge/push/auto-merge.

zensgit:
## CI was red on ONE spec — pulled it, kept the other 10. `65e0a8c25`

The gate's local full-batch run was green, but CI's actual batched `web-tests` failed on exactly one spec: **`multitable-comment-inbox-view`** (`expected '1' to be '0'`, the unread-count). It passes 6/6 in isolation; it fails only in CI's full batch — a co-execution issue. The CI log shows `[approval-detail] mark-read failed Error: network` immediately before it, i.e. an earlier approval spec's network-failing state bleeds through despite this spec's own `vi.mock` isolation.

Wiring a batch-flaky spec into the REQUIRED gate makes the required check flaky — worse than leaving it unwired. So `65e0a8c25` removes **only** comment-inbox-view's wiring (its run-list token + two guard path triggers). **Its spec fix stays** (green in isolation, gate-approved) — just not run in the required gate until its co-execution isolation is fixed in a focused follow-up.

The other **8 repaired specs (7 workbench + phase11) passed in CI's batch and stay wired.** This is pure subtraction of one spec's coverage — the gate's MERGE_CLEAN on the repairs is unaffected. `multitable-comment-inbox-realtime` (the #4191 sibling) is untouched.

Follow-up ticket: comment-inbox-view needs batch-robust isolation (likely the module-singleton / shared-fetch leak this line has hit before) before it can be wired.

This is also a concrete instance of a standing lesson: **the gate ran the batch and got green, but I'd `update-branch`'d after the gate, so CI tested a newer tree than the gate saw** — a verdict binds the SHA. The repairs still hold; only the batch-composition-dependent wiring shifted.

PR #4223, opened 2026-07-13
feat(web): T5-safe — migrate MetaRecordDrawer watch/workflow/permissions/duplicate/delete/unlock to MtButton (owner-ratified)

zensgit:
## Scope

Executes **T5-safe** — the owner-ratified subset of the T5 MetaRecordDrawer button migration
(`docs/development/multitable-ui-p2-1c-t5-recorddrawer-decision-brief-20260712.md`, ruled by the
owner 2026-07-13, splitting the original T5 into gated decision points). Target:
`apps/web/src/multitable/components/MetaRecordDrawer.vue`.

Does **not** merge. Base `main`.

## Per-button outcome

| Button | Decision | Outcome |
|---|---|---|
| **watch / watching** | OD-T5a option A | → `MtButton` (default/ghost variant). Kept `--watching` as the active-state CSS class (MtButton has no built-in pressed/active variant — adding one would be a primitive-contract change, out of scope). **Added `:aria-pressed="recordSubscribed"`** — the one new a11y surface this PR introduces. Label-swap (`Watch`/`Watching`), `:disabled="subscriptionLoading"`, `@click="toggleRecordSubscription"` all byte-preserved. |
| **comment** | OD-T5b — **excluded, deliberately** | Left byte-identical. Stays a bespoke `<button>` wrapping `<MetaCommentActionChip>`. This is a separate governance item per the brief (§2.1: the active-state color has no matching design token, so "token-only" and "preserve observed color" directly conflict; §2.2: it's one of ~12 call sites of a 9-component comment-affordance system — migrating only the drawer's instance would fork that shared visual language). A dedicated test in this PR (`leaves the comment button bespoke...`) proves the boundary empirically: the comment button carries no `mt-button` class. |
| **workflow** | OD-T5c option A | → `MtButton`. The `&#x2699;` (⚙) glyph stays inline as slot text — same convention as T1's `×` glyph, no new icon component/contract. `v-if="canManageAutomation"`, `:title`, `@click="emit('open-automation')"` byte-preserved. |
| **permissions** | OD-T5c option A | → `MtButton`. The `&#x1F512;` (🔒) glyph stays inline as slot text, same rule. `v-if="canManageRecordPermissions"`, `:title`, `@click="showRecordPermissions = true"` byte-preserved. |
| **duplicate** | OD-T5d | → `MtButton`. `v-if="record && canCreate"`, `:title`, `@click="emit('duplicate')"` byte-preserved. Kept `.meta-record-drawer__btn--duplicate` class — an **existing** test (`multitable-record-drawer-duplicate.spec.ts`) selects on it. |
| **delete** | OD-T5d | → `MtButton` with **`variant="danger"`** (the only other new addition besides aria-pressed). `v-if="resolvedCanDelete"`, `@click="emit('delete')"` byte-preserved. |
| **unlock** | OD-T5d | → `MtButton`. `v-if="record.canUnlock"`, `data-test="record-unlock-action"`, `@click="emit('toggle-lock', { recordId: record.id, locked: false })"` byte-preserved. |

## CSS cleanup (rule 5: remove orphaned rules only)

- `.meta-record-drawer__btn--danger` and `.meta-record-drawer__btn--watch` removed — each had exactly
  one sharer (delete, watch respectively) and both are now migrated (delete → MtButton's own
  `--danger`; watch's non-active state → MtButton's default ghost styling).
- `.meta-record-drawer__btn--watching` **kept** — it's the watch toggle's only active-state visual
  (see above), now paired with `aria-pressed`.
- Base `.meta-record-drawer__btn` rule and all three `--comment*` rules **kept untouched** — the
  comment button (excluded this round) still needs them for its styling.
- No new hardcoded hex (`git diff | grep -E '^\+' | grep -iE '#[0-9a-f]{3,8}'` → empty).

**Token-normalized, NOT "zero visual change"**: the watch button's non-active state changes from the
bespoke light-blue `--watch` look to MtButton's default ghost styling. The active (`watching`) state
is visually unchanged (kept CSS rule). Workflow/permissions/duplicate/unlock get MtButton's default
ghost token styling in place of the bespoke flat-gray look. Delete gets MtButton's token-red danger
styling in place of the bespoke `#f56c6c` text-only red.

## Tests

New: `apps/web/tests/multitable-record-drawer-t5-migration.spec.ts` (16 tests) — anchored on stable
selectors (`data-test`, `[title="..."]`, retained old classnames), never fuzzy textContent/optional-chained
clicks:
- watch: renders disabled while the subscription-status fetch is in flight, `aria-pressed="false"` once
  loaded; click → same `apiClient.subscribeRecord(sheetId, recordId)` call, `aria-pressed` flips to
  `"true"`, `--watching` class appears, label/title swap; click again → same `unsubscribeRecord` call,
  `aria-pressed` reverts to `"false"`; hides when there's nothing to subscribe to.
- workflow / permissions: glyph+label slot text, click fires the real handler (`open-automation` emit /
  opens `MetaRecordPermissionManager` — asserted via `.meta-record-perm__overlay` appearing), v-if
  survives the swap.
- duplicate / delete / unlock: click emits the exact same payload as before migration (`toggle-lock`
  payload asserted with `toHaveBeenCalledWith({ recordId: 'rec_1', locked: false })`); delete carries
  `mt-button--danger`; unlock keeps `data-test="record-unlock-action"`; all three v-if guards proven to
  still hide the button.
- boundary: comment button mounted with `canComment: true` and asserted to carry **no** `mt-button`
  class — proves OD-T5b wasn't accidentally migrated.

Existing specs re-run unchanged and green: `multitable-record-drawer.spec.ts`,
`multitable-record-drawer-button.spec.ts`, `multitable-record-drawer-duplicate.spec.ts`,
`meta-record-drawer-restore/-history-diff/-i18n.spec.ts`, `multitable-ai-shortcut-drawer.spec.ts`, plus
the 16 workbench/field specs that mount `MultitableWorkbench` (which embeds the real drawer) — 16 files,
148 tests, all pass unchanged.

`vue-tsc -b` clean.

## Mutation-red evidence (each broken, spec reds, restored)

All done via targeted `python3` edits on the committed checkpoint, `git checkout -- <file>` used only
to restore that exact single file back to the just-committed state (never a broad checkout):

1. Removed `:aria-pressed="recordSubscribed"` from watch → **3 tests RED** (all three aria-pressed assertions).
2. Removed `variant="danger"` from delete → **1 test RED** (`mt-button--danger` class assertion).
3. Mutated the unlock `@click` payload (`locked: true` instead of `false`) → **1 test RED** (payload assertion).
4. Removed `data-test="record-unlock-action"` from unlock → **1 test RED**.
5. Neutered duplicate's `@click` to a no-op → **1 test RED** (emit-count assertion).
6. Flipped permissions' `v-if="canManageRecordPermissions"` to `v-if="true"` → **1 test RED** (hide-when-false assertion).
7. Neutered workflow's `@click` to a no-op → **1 test RED** (emit-count assertion).
8. Migrated the comment button to `<MtButton>` (simulating an accidental OD-T5b violation) → **1 test RED** (the boundary-proof test).

Each mutation restored (`git checkout -- apps/web/src/multitable/components/MetaRecordDrawer.vue`) and
re-verified green before moving to the next.

## CI wiring (two-point, proven to bite)

- `apps/web/scripts/run-required-web-tests.sh` — added `multitable-record-drawer-t5-migration.spec.ts`
  (full filename, to stay unambiguous next to the sibling `multitable-record-drawer*` tokens already
  in the list) + a dated maintenance comment.
- `.github/workflows/multitable-web-guard.yml` — added `apps/web/src/multitable/components/MetaRecordDrawer.vue`
  and the new spec path to **both** trigger blocks (`pull_request.paths` and `push.paths`), plus the
  same token to the job's `vitest run` command.
- Fetched `origin/main` immediately before committing/pushing and rebased cleanly (one unrelated
  docs-only commit landed in between — no conflicts, strict union preserved, `grep -c` confirms the
  token appears exactly twice in the workflow's path blocks + once in its vitest command, and once in
  the required-tests script).
- **Bite proof**: temporarily broke an `aria-pressed` assertion in the new spec, ran
  `bash apps/web/scripts/run-required-web-tests.sh` (the exact required-gate script, full batch —
  326 files / 3744 tests) → **exit 1**, the two mutated assertions failed with the expected diff.
  Restored → re-ran the same full batch → **exit 0, 326 files / 3744 tests passed** (not just the
  isolated spec — the whole required-gate batch, per the two-sibling-PR isolation-green/batch-red
  lesson this week).

## Red-line compliance

Only button elements + CSS were touched. No permission/deletion/lock business logic changed — every
handler function body (`toggleRecordSubscription`, `resolvedCanDelete`, `emit('toggle-lock', ...)`,
etc.) is byte-identical (`git diff` shows zero changes inside `<script setup>` except the one new
`import { MtButton } from '../ui'` line). No new MtButton variant/prop/size was introduced — only the
existing `danger` variant is used. The comment button was not touched in any way (glyph, class,
`resolveCommentAffordanceStateClass` call, `<MetaCommentActionChip>` usage — all untouched).

refs #4175 (the design-lock this executes)

🤖 Generated with [Claude Fable 5](https://claude.com/claude-code)


zensgit:
## Adversarial gate — VERDICT: BLOCK (scoped to head `548a865bda62b005ed6aef76a43947fb2b6dc550`)

Refute-first review in a fresh detached worktree off the PR ref. All four owner red lines, OD rulings, logic, wiring, tests, and mutations are clean — but there is **one measured, reproducible visible regression on a destructive affordance (P2)** that should not ship. Fix is CSS/class-only and small; happy to re-gate immediately after.

### P2 — Delete button renders white-on-white (illegible)
The migrated delete `<MtButton variant="danger" class="meta-record-drawer__btn meta-record-drawer__btn--danger">` keeps the base class, and Vue 3 applies the parent scope-id to the single-root child — so the retained `.meta-record-drawer__btn` base rule (`background:#fff`, **no `color`**) also matches MtButton's root. Specificities are equal (`.mt-button--danger[data-v-child]` = 0,2,0; `.meta-record-drawer__btn[data-v-drawer]` = 0,2,0), so the **later-injected** rule wins — and MtButton's CSS is injected *before* the drawer's (child-before-parent ESM order, faithful to the prod bundle).

**Measured** (mounted the real drawer with `css:true`, real scoped compilation):
- injected-style index `.mt-button--danger{…background…}` = **1470**
- injected-style index `.meta-record-drawer__btn{…background:#fff…}` = **21903** → base wins
- `.mt-button--danger` supplies only `color: var(--ms-bg-card)` = `#ffffff` (white text); base supplies no color → **background resolves to `#fff`, text stays white → the "Delete" label is invisible.**

Delta vs main: main rendered delete as white-bg + **red `#f56c6c` text/border** (clearly readable, clearly destructive). This SHA renders white text on white bg + gray `#ddd` border. That contradicts OD-T5d's stated "token-red danger styling" outcome. (Ghost buttons are unaffected — their `--ms-text-2` gray text stays legible on the white base bg.)

**Fix (small, CSS/class only):** drop the `meta-record-drawer__btn` base class from the migrated **delete** `<MtButton>` (keep the `--danger` modifier — it's only a test anchor now, no rule), so MtButton's danger styling applies uncontested. The spec asserts `mt-button--danger` + `meta-record-drawer__btn--danger` on delete (not the base class), so this is test-safe. Alternatives: narrow the base rule to the comment button only, or re-add color/bg on `.meta-record-drawer__btn--danger`.

### Everything else — PASS
- **Red lines**: `<script>` diff is import-only (`import { MtButton } from '../ui'`); logic byte-identical. Per-button only allowed additions present (`:aria-pressed="recordSubscribed"` on watch, `variant="danger"` on delete); `type="button"` removals are behavior-neutral (MtButton hardcodes `type="button"`). Comment button element + base `.meta-record-drawer__btn` + all `--comment*` rules byte-identical. No new hex; no invented variant.
- **CSS orphan audit**: `--danger`/`--watch` rules removed — repo-wide grep confirms sole-sharer. `--watching` rule still **live** (active-state visual preserved).
- **aria-pressed**: bound to `recordSubscribed`; spec tests false→true→false (a hardcoded value cannot pass).
- **Mutations reproduced** (apply → red → restore): comment-boundary → 1 RED; aria-pressed removal → 3 RED; unlock payload tamper → 1 RED.
- **Tests**: new spec 16/16; `vue-tsc -b` exit 0; full required batch **326 files / 3744 tests, exit 0** (batch-green).
- **Two-point wiring**: `run-required-web-tests.sh` (invoked by required `web-tests.yml`) + guard — union preserved, zero pre-existing tokens dropped.

Verdict binds `548a865bd`. Once the delete-cascade fix lands I expect a clean pass.

_(Adversarial review — no merge/push/arm performed.)_

zensgit:
## Re-review — VERDICT: MERGE_CLEAN (scoped to head `8fabaa999267da56ace16392e9a3040bad42e59d`)

The P2 (delete button white-on-white) is fixed. Delta from the blocked SHA is exactly one element + one template comment: the delete `<MtButton variant="danger">` drops the `meta-record-drawer__btn` base class (keeps `meta-record-drawer__btn--danger` as the stable spec anchor).

**Re-measured with the same method** (mount real drawer, `css:true`, real scoped compilation):
- delete class list is now `mt-button mt-button--danger mt-button--md meta-record-drawer__btn--danger` — base class gone.
- with the base class dropped, the base rule (`background:#fff`) no longer matches the element, so `.mt-button--danger` owns the cascade uncontested → **red bg (`--ms-color-danger` #dc2626) + white text (`--ms-bg-card` #ffffff) + red border = legible destructive button.** ✓
- 16-test spec re-runs **green** (delete test anchors on `mt-button--danger`, never the base class).

**Five-keep / one-drop split is sound.** The other five migrated buttons keep the base class deliberately: they use MtButton's ghost variant whose text is gray `--ms-text-2` (#4b5563), legible on the base rule's white bg — none has the inverted-white-text problem, and the base rule preserves their pre-migration look. Only the danger variant has white text that requires a red bg to be legible, and that's exactly the one whose base class was dropped. Agreed, not flagged.

Everything else passed at `548a865bd` and carries over unchanged — the delta touches no wiring, no `<script>`, no other CSS rule, no spec logic, and does not touch the comment button. No need to redo the full battery.

Verdict binds `8fabaa999`. _(Adversarial review — no merge/push/arm performed.)_

PR #4205, opened 2026-07-13
docs(multitable): R13 Lane B — T-state history + preview/restore from History Center (PROPOSED)

zensgit:
## What

**Docs-only design lock** (zero runtime, zero migrations, zero wire changes) for **R13 gap-closure Lane B**: the point-in-time (T-state) history experience — full historical-version viewing + single-record preview + guarded restore from the History Center.

- Doc: `docs/development/multitable-r13-lane-b-tstate-history-experience-designlock-20260712.md`
- **Status: PROPOSED — awaiting owner ratify.** No implementation is authorized by this PR; slices B1–B5 are each gated behind ratification of §3 + the §8 decision points.
- Parallel with Lane A (revision integrity, drafted separately) and Lane C (ops/scale). **R14 base-wide restore is explicitly out of scope** (§7) and the batch surface is locked so it cannot become a smuggling path for it (B-LOCK-6).

## Owner decisions encoded (B-LOCK-1..6)

1. A record that existed at T and was deleted after T MUST reappear in the T-state (values from `meta_record_revisions` via the existing `reconstructRecordsAtT` primitive; inbound edges via `meta_link_tombstones` + the A′ anchor at resurrect time).
2. A currently non-existing record is previewable directly from the History Center (pure read, independent of current existence).
3. Preview and Execute are separated (PV-1 zero-mutation invariant, with a positive-control leg in the test plan).
4. Restore = existing version/record locks (expectedVersion CAS + preview identity) **plus** a new idempotent `operationId` (same-id replay returns the recorded outcome, never double-applies) — closing the anti-replay gap the identity module itself documents as deferred.
5. Read authorization is REUSE of the existing field perms / row-level deny / field masking / no-existence-leak chain — no new RBAC. The single reuse parameter needing owner judgment (rule-deny evaluation basis for non-live rows) is flagged as OD-B1, not designed around.
6. Single-record and batch restore are separate surfaces; batch = explicit recordIds ≤100, no wildcard, no cross-table scope.

## Grounding (verified at origin/main 74d4f8eb6)

- `reconstructRecordsAtT` exists (record-reconstructor.ts:34), delete-aware + deterministic; the T7 point-in-time view (univer-meta.ts:8280) currently scopes to **currently-live rows only** and has **no FE consumer** — the two concrete gaps this lane closes, alongside the missing idempotent operation id and the missing single-record resurrect surface.
[...]

PR #4224, opened 2026-07-13
docs(multitable): R13-C design lock — retention↔Reset coexistence + async>5000 scale (PROPOSED)

zensgit:
Design-first per your R13-C ruling (设计锁 + 真实基准先行, 不改高风险恢复事务). Replaces the blanket retention-vs-Reset 409 with a per-T reconstructability gate (safety invariant intact); async recovery job for >5000 mirroring the AI-bulk-job pattern; benchmark methodology + async thresholds. **Zero code/flag, PROPOSED.** OD-C1..C4 for your ratify; real benchmark is a ratify prerequisite.

### DIFF
--- a/docs/development/multitable-global-history-r13c-retention-reset-coexistence-and-async-scale-design-lock-20260713.md
+++ b/docs/development/multitable-global-history-r13c-retention-reset-coexistence-and-async-scale-design-lock-20260713.md
@@ -1,0 +1,69 @@
+# R13-C — Retention↔Reset 保留窗口内共存 + 大规模恢复异步化 — DESIGN LOCK (PROPOSED)
+
+- **Status**: **PROPOSED — 2026-07-13. NOT ratified. Design-first, ZERO code/flag.** Per owner R13 ruling「**C 先做设计锁和真实数据量基准，避免直接改高风险恢复事务**」。此锁只定义语义、给出 OD 决策与基准方法学；**不改任何恢复事务**。owner ratify §5 (OD-C1…OD-C4) + 真实基准跑完，才进实现。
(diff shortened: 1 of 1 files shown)


zensgit:
**Status: deliberately PARKED — do not sweep-close.** (F4 disposition sweep, 2026-08-22)

Unlike the five PRs closed in this sweep, **this one's content did NOT land**. Verified today against `origin/main` by negative evidence:
- `RESET_T_BEYOND_RETENTION` → **0 hits** repo-wide.
- `reconstructabilityCheck` / `reconstructability` → **0 hits**.
- `multitable_recovery_job` / `recovery-jobs` → **0 hits**.
- The blanket gate this lock proposes replacing is **still live**: `PIT_RESET_RETENTION_BLOCKED` / `code: 'RESET_RETENTION_CONFLICT'` in `univer-meta.ts`, still asserted as correct by `multitable-reset-pit-realdb.test.ts`.

It is also **not obsolete**: its subject matter is exactly the Phase D half (retention-after archive + large-batch async recovery) that the owner decision sheet §D2 **defers to post-L6-soak** — a decision that has not been reached (the ladder is still pre-L1).

**What closing would lose**: OD-C1 (blanket 409 → per-T, per-record reconstructability gate), OD-C2 (`keep-last-n` retention is **per-record**, so a cross-record reset-to-T may be reconstructable for one record and pruned for another — reject wholesale / partial / keep-days-only), OD-C3 (async batching downgrades whole-sheet reset from all-or-nothing to 'batches + a consistent interruption point', coupled to the R14 base-wide decision), OD-C4 (is the 5000 ceiling evidence-based, or should it be calibrated to a measured lock-hold threshold?), plus the §3 benchmark methodology. **This is the only drafted design input for that half of Phase D.**

Parked with the same status as #4205 so a future sweep does not silently discard it. If Phase D is ever chartered, this must first be re-based against the landed #4654 closeout semantics + O-2 hardening, which postdate it.

PR #4227, opened 2026-07-13
test(multitable): OD-6 revision-disposition guard — every meta_records write site must declare a disposition

zensgit:
## Summary

Implements the **OD-6 revision-disposition guard** ratified in the design-lock `docs/development/multitable-global-history-d1c-form-submit-edit-uncaptured-revision-design-lock-20260712.md` §0.5, per the consolidated audit `/tmp/r13-revision-disposition-audit-20260713.md` §2.

The revision system was in the exact pre-guard state the rank-8 record-lock guard was built to end: **8 `meta_records` mutation sites** (found across 3 independent human audits — rank-8, D-1, D-1c — each fixing only the sinks it happened to look at) write **no** `recordRecordRevision(...)`, so `reconstructRecordsAtT` (the PIT/revert/reset read primitive) silently and permanently lies about those records. This PR adds a static structural guard so a **new** uncaptured sink can never be added silently again.

## What changed

- **New file**: `packages/core-backend/tests/unit/multitable-revision-disposition-guard.guard.test.ts` — modelled on `multitable-record-lock-guard.guard.test.ts`, but **also covers `INSERT INTO meta_records`** (the lock guard deliberately excludes INSERT — you cannot lock a record that doesn't exist yet; a revision guard cannot skip it, since an uncaptured CREATE is invisible to the reconstructor forever).
- **Source changes are comment-only markers, zero runtime behavior change**, added to all 36 real `meta_records` write sites in `packages/core-backend/src` (9 INSERT + 22 UPDATE + 5 DELETE):
  - **14 `// revision-emitted:`** on the currently-compliant sites (record-service.ts, record-write-service.ts, records.ts, automation-executor.ts, univer-meta.ts).
  - **13 `// revision-exempt:`** on derived/system/config-captured sites (formula recompute, auto-number backfill, People-directory/sheet sync, seed bootstrap, field-delete config-channel strip, record lock/unlock metadata, the derived approval projection).
[...]

PR #4234, opened 2026-07-13
feat(multitable): §0.6 HISTORY_INCOMPLETE — fail-closed integrity precheck for destructive recovery (D-1c RATIFIED, lands first)

zensgit:
## Authority

`docs/development/multitable-global-history-d1c-form-submit-edit-uncaptured-revision-design-lock-20260712.md` **§0.6 — RATIFIED 2026-07-13 (owner)**, read from PR #4187 branch head `95ed6e9a9`. Owner sequence: **§0.6 lands FIRST and ALONE**. refs #4187.

**Update:** the owner reviewed the original "3 suites trip the doctrine" finding and issued a three-bucket ruling — applied in the second commit. Full CI-style real-DB run is now **13/13 files, 135/135 tests green**. See "Owner's three-bucket ruling" below for exactly what changed and why.

## What this is

The fail-closed `HISTORY_INCOMPLETE` integrity precheck for the destructive-recovery surfaces: while any live record's history in a sheet is untrustworthy, PIT Revert-to-T and Reset-to-T — **preview AND execute** — refuse with one values-free `409 { code: 'HISTORY_INCOMPLETE' }`, mint no execute token, and write zero rows. It protects existing polluted data **before** the five A1–A8 write-path slices backfill the uncaptured writers.

## Where it lives / who consumes it

- **Shared precheck**: `packages/core-backend/src/multitable/history-integrity-precheck.ts` — `precheckSheetHistoryIntegrity(query, sheetId)` (pure read, doc-commented against §0.6 rules 1–4).
- **Consumers** (`packages/core-backend/src/routes/univer-meta.ts`; `~` line numbers will drift):
  - `computeSheetRevert` calls it at `~L9951` (after the D3/PIT-6 record-count ceiling, before any scan) → consumed by `POST /sheets/:sheetId/revert-preview` (`~L10098`; refusal surfaced `~L10112`, **before token minting**) and `POST /sheets/:sheetId/revert-execute` (`~L10138`; refusal `~L10155`, **before `previewIdentity` verification and before any transaction**).
  - `computeSheetReset` calls it at `~L10031` → consumed by `POST /sheets/:sheetId/reset-preview` (`~L10315`; refusal `~L10331`) and `POST /sheets/:sheetId/reset-execute` (`~L10357`; refusal `~L10376`).
  - Unified sender `sendHistoryIncomplete`: one exact body for all four routes; **no record ids, no counts** (no denied-record existence oracle), and — being an error response — no `previewIdentity`.
- **Execute re-check (rule 4, TOCTOU)** is by construction: execute calls the same compute function, and the refusal fires **before** identity verification — the revision-derived `scopeHash` is never relied on for this.
- The **PIT read-only view** (`GET /sheets/:sheetId/point-in-time`) is deliberately NOT gated — §0.6 rule 1 names the destructive surfaces only.

## Semantics (per §0.6, verbatim reading, + P3-1 hardening)

1. **Unified refusal** — one code, both phases, all four routes; refusal ⇒ no token, zero writes.
2. **Content, not version, projected to user-authored fields.** Live `meta_records.data` vs the **latest** revision `snapshot` (LOCK-11 order, no `<= T` cut), compared only over non-excluded field ids. **Exact exclusion list, `isNonUserAuthoredFieldType` (pinned from `field-codecs` `mapFieldType` canonical names):**
   - the four **derived-value** types (`DERIVED_FIELD_TYPES`): `formula`, `rollup`, `lookup`, `autoNumber` — the value engines that materialize computed keys into `data` without a revision *by design* (`formula-engine.ts` recompute, `univer-meta.ts` relation-aggregation fan-out, `auto-number-service.ts` backfill);
   - **P3-1 hardening** (post-ratification, same doctrine): also the four **system-computed** types via `field-codecs`'s canonical `SYSTEM_FIELD_TYPES` — `createdTime`, `modifiedTime`, `createdBy`, `modifiedBy`. Rule 2's text says "user-authored/writable field ids"; these are server-maintained exactly like a derived field (`isFieldAlwaysReadOnly` already buckets them with formula/lookup/rollup). Not a live false-positive today — `query-service.ts`'s `injectSystemFieldValues` derives these purely at READ time and no writer ever stores a value under a system field id in `data` — but the security path should not depend on that being true forever. `autoNumber` is a member of both sets; the union dedupes it.

   `link` ids **are** compared (OD-4: captured as ordinary `data`). Value equality is canonical: key-order-insensitive objects, `null` ≡ absent (the same `?? null` fold `record-restore-diff.ts` `sameValue` uses). Version-only mutations (record lock/unlock) never enter the comparison.
3. **Zero-revision live rows refuse, enumerated from the LIVE row set** — never the reconstruction (`computeSheetReset` pushes live rows absent from the reconstruction into its delete-set; iterating the reconstruction misses exactly the rows a reset would silently destroy). Scope = the whole sheet's live rows (rule 3 verbatim: "for each live row"), applied uniformly to revert and reset by the one shared function.
4. **Fail-closed** — comparator errors refuse; DB errors propagate to the routes' existing non-write error mapping (404/503/500 — all zero-write refusals). No proceed-on-error branch.
5. **Fail-closed extension (HI-5, beyond the four owner goldens, same doctrine):** a LIVE row whose latest revision is a `delete` refuses even when content matches the pre-delete snapshot — history claims the record dead, `reconstructRecordsAtT(now)` reports `exists:false`, and reset would delete-set it; content comparison alone cannot see this. No captured path produces the state (trash-restore and PIT-resurrect both write a fresh `create` revision in-txn).

**Explicitly untouched:** the 8 uncaptured write sites A1–A8 (five later slices, owner order form → plugin → automation → approval → attachment); the OD-6 revision-disposition guard (#4227, lands last); no backfill of any kind (OD-5); no flag (the lock ships this always-on as a correctness gate).

## Goldens — `tests/integration/multitable-history-incomplete-precheck-realdb.test.ts`, real Postgres 14, 9/9 green

Env: isolated Docker `postgres:14` on non-default port 55446; migrations with **CI's exact `MIGRATION_EXCLUDE`** (all green); run via `vitest --config vitest.integration.config.ts` (CI's serial invocation).

| # | Scenario | Result |
|---|---|---|
| positive control | all-healthy sheet → preview 200 mints token, execute 200 reverts; the revert's own captured write leaves live == latest snapshot | ✅ |
| **G-HI-1** | polluted record (live user-field data ≠ latest snapshot — the reproduced A1 fingerprint) → **all four surfaces** 409 `HISTORY_INCOMPLETE` (exact-body deepEqual); execute legs driven with never-minted dummy tokens | ✅ |
| **G-HI-2** (+ reset leg) | version-only drift via the **real** `/records/:id/lock` route (lock+unlock: v1→v3, `data` untouched, no revision) → passes; revert AND reset proceed; the drifted record untouched | ✅ |
| **G-HI-3** | formula sheet: healthy create (live carries materialized `formula`+`autoNumber` keys the snapshot lacks) AND healthy update whose snapshot carries a **stale** derived value vs live → passes; revert + reset proceed | ✅ |
| **G-HI-4** | live record with ZERO revisions → all four surfaces refuse; the row is **still live** afterwards (silent-delete-on-reset hole closed) | ✅ |
| HI-5 | live row whose latest revision is `delete`, content identical → refused | ✅ |
| **P3-1** | system-computed field (`modifiedTime`) carries an ADVERSARIAL stale value that DIFFERS between live and snapshot → precheck passes anyway (pins the system-type exclusion; before the P3-1 fix this exact state would content-mismatch-refuse a healthy record) | ✅ |

**Zero-writes post-assertion approach** (asserts the DB, not the response): `sheetWriteState()` captures every live row (`id`,`data`,`version`, ordered) + total revision count + `source='restore'` revision count + trash-row count, deepEqual **before vs after all four refusals** (`expectAllFourRefuseWithZeroWrites`). **No-token-minted** pinned by exact-body equality (refusal body has no `data`/`previewIdentity`) plus execute legs using tokens that were never minted.

## Mutation matrix — every golden proven load-bearing (mutate → run → exact reds → restore)

| Mutation | Expectation | Result |
|---|---|---|
| M1 skip the user-authored projection (compare ALL fields incl. derived + system) | G-HI-3 **and** P3-1 both red (both pin the projection) | ✅ 2 failed / 7 passed |
| M2 iterate the reconstruction-visible set (skip zero-revision live rows) | exactly G-HI-4 red | ✅ 1 failed / 8 passed |
| M3 execute routes stop surfacing the re-check (regress to identity-shaped 409) | execute legs: G-HI-1 + G-HI-4 + HI-5 red | ✅ 3 failed / 6 passed |
| M4 neuter the content comparison (always equal) | exactly G-HI-1 red | ✅ 1 failed / 8 passed |
| M5 compare **version** instead of content (the naive impl rule 2 forbids) | exactly the two G-HI-2 legs red | ✅ 2 failed / 7 passed |
| M6 drop the live-row-after-delete-revision refusal | exactly HI-5 red | ✅ 1 failed / 8 passed |
| **P3-1M** drop ONLY the system-type exclusion (`isNonUserAuthoredFieldType` → derived-only) | exactly P3-1 red, all seven others stay green | ✅ 1 failed / 8 passed |

## Two-point CI wiring (+ bite proof)

- `packages/core-backend/vitest.config.ts` **exclude** list (no-DB job cannot skip-green it) **and** `.github/workflows/plugin-tests.yml` `Run multitable real-DB integration` run list (adjacent to `multitable-reset-pit-realdb.test.ts`).
- **Bite proven:** corrupting the expected refusal code → DB-form run reds; default no-DB config does not collect the file (`No test files found`). Probe reverted.

## Owner's three-bucket ruling (second commit — applied verbatim)

The first commit's honest finding — 19 pre-existing assertions in 3 real-DB files tripped the newly-enabled doctrine — went back to the owner. The owner ruled all three buckets; here is exactly what was done, file by file:

**A-bucket — fixture repair (17 assertions: 14 in `multitable-undelete-pit-realdb.test.ts` + 3 in `multitable-undelete-pit-inbound-replay-realdb.test.ts`).** Both suites seeded neighbour record `L` as a live row with **zero revisions** — indistinguishable from an uncaptured create, the exact class §0.6 exists to catch. Ruling: this is a fixture bug, not a §0.6 weakening — `L` was always meant to be an ordinary captured neighbour. Fix: added a matching `create` revision for `L` in both suites' `seed()`. Gate proved: `multitable-undelete-pit-realdb.test.ts` → **15/15 green**; `multitable-undelete-pit-inbound-replay-realdb.test.ts` → **4/4 green**.
- One knock-on case inside that repair, **P3-4**: test `(g)` ("id occupied between preview and execute") raw-inserts a squatter live row over `U` with no revision, to exercise the `resurrectScopeHash`-mismatch drift path. Once `L` was fixed, the squatter itself became the only remaining zero-revision live row in the suite — the precheck would refuse it first (still 409, but via the WRONG mechanism, silently defeating the test's own intent). Fix: gave the squatter a matching revision (timestamped **after** the revert's `asOf` so it stays outside the T1 reconstruction window and the intended diff still exists), and added an explicit `expect(x.body?.error?.code).toBe('PREVIEW_IDENTITY_INVALID')` to positively pin that the test now exercises the intended path, not a §0.6 refusal that happens to also return 409.
- A second, symmetric case in the inbound-replay suite, **P3-2c**: raw-UPDATEs `L`'s link field to `[]` (simulating "the neighbour declined") with no version bump / revision — itself an uncaptured write once `L` has a revision to disagree with. Fix: the update now bumps `version` and writes a matching `update` revision (timestamped before the revert's `asOf`, so it doesn't spuriously enter `L` into the revert-diff), making the "declined" state itself capture-complete.

**B-bucket — TOCTOU re-pin (2 assertions in `multitable-reset-pit-realdb.test.ts`, `(g1)`/`(g2)`).** Both deliberately inject an **uncaptured create/edit between preview and execute** — a real TOCTOU, and §0.6 refusing first is the *correct* behaviour (that's exactly what rule 4 is for). Ruling: re-pin, don't relax. Changed `expect(...).toBe('PREVIEW_IDENTITY_INVALID')` → `.toBe('HISTORY_INCOMPLETE')`; `status===409` and every zero-write assertion (D/E untouched, A/B unchanged) are byte-identical to before. Added the requested marker comment verbatim: `// §0.6 restore-marker: original PREVIEW_IDENTITY_INVALID assertion restores after the corresponding revision-slice lands`.

**P3-1 — hardening.** Covered above in "Semantics" and in the goldens/mutation tables.

**Full re-verification after all three buckets:**
- Full CI-style serial real-DB run across all 13 files that drive the four gated routes (the original 12-file blast-radius inventory + the new precheck spec): **13/13 files, 135/135 tests green** (was 115/134 with 19 red before the ruling).
- Zero-writes `deepEqual` legs (`expectAllFourRefuseWithZeroWrites`) still hold — unaffected by the fixture/re-pin changes, which only touch pre-existing suites and the pinned error code.
- `tsc --noEmit` clean.
- No-DB unit lane re-run after the ruling: 466/466 files, 6409/6409 tests, 0 regressions.
- All 7 mutations (the original 6 + the new P3-1 mutation) re-verified against the post-ruling tree — every one isolates to exactly its pinned golden(s), then restored via `git checkout --` from the committed baseline (never against uncommitted work).

## Verification

- `tsc --noEmit` clean (both commits).
- Goldens: 9/9 green (real DB, CI config, serial).
- Full real-DB blast-radius run: 13/13 files, 135/135 tests green.
- No-DB unit lane: 466/466 files, 6409/6409 tests green (both before and after the ruling).
- Mutation matrix (7 mutations) + wiring-bite probe all reverted — the tree is exactly the reviewed diff.
- Footprint: precheck module + 4 route branches + sender (commit 1); 3 pre-existing test files' fixtures/assertions + 1 new golden (commit 2). No changes to the 8 A1–A8 write sites or the OD-6 guard in either commit.


zensgit:
## Adversarial gate — VERDICT: MERGE-AFTER-FIXTURE-REPAIR (+ one narrow owner nod)

Reviewed at head `a64d9e326abb72493e2110ea5f868e2b431a1cb2` on an isolated Postgres 14 (CI migration-exclude + `vitest.integration.config.ts`, serial). §0.6 itself is **correct and well-tested** — no P1/P2 defect in the precheck. The blocker to green is the 19 pre-existing assertions, and every one resolves via the resolution you proposed, which I **validated empirically**.

### Per-file adjudication of the 19 (reproduced exactly: 14 + 3 + 2)

| File | fail | Bucket | Root cause | Disposition | Verified |
|---|---|---|---|---|---|
| `multitable-undelete-pit-realdb` | 14 | **(a) fixture shortcut** | neighbour `L` seeded raw with **zero revisions** (L47); intended as a *healthy* neighbour, but its state is byte-indistinguishable from the A4–A6 uncaptured-create fingerprint, so §0.6 rule 3 refuses at the first preview (correct fail-closed) | add `L` create revision matching live | **PROVEN 15/15 green**; no assertion depends on L's zero-rev state |
| `multitable-undelete-pit-inbound-replay-realdb` | 3 | **(a) fixture shortcut** | identical zero-revision `L` (L68) | same repair | same root cause |
| `multitable-reset-pit-realdb` g1/g2 | 2 | **production-reachable; §0.6 refusal CORRECT** | g1 injects uncaptured CREATE, g2 uncaptured EDIT between preview/execute — the exact TOCTOU rule 4 catches | re-pin code `PREVIEW_IDENTITY_INVALID`→`HISTORY_INCOMPLETE` | **PROVEN**: sole failing assertion is the code string; `status===409` + zero-write hold. This is **positive TOCTOU evidence** (preview 200 → execute 409) |

No genuine §0.6-scope/sequencing conflict: the undelete/reset feature tests never depend on the uncaptured writes being *captured*. Your rejected alternative (scope rule 3 to only-written rows) would contradict the verbatim lock — correctly declined. The only open item is process: the repairs touch **three later-slice test files**, against your "§0.6 first and alone." They are test-only harness upgrades (no A1–A8 runtime, §0.6 not weakened) — recommend allowing them in this PR.

### Acceptance points — all PASS
1. Single shared precheck across all four routes (both compute fns, before mint/txn).
2. Execute zero-write proven against the **DB** (`sheetWriteState` deepEqual before/after all four refusals; dummy tokens).
3. Goldens both directions (refuse: G-HI-1/4/HI-5; proceed: positive control/G-HI-2 via real lock route/G-HI-3 projection).

### Empirical record
- §0.6 goldens **8/8**; neighbours (revert-pit + reconstructor) **16/16**; tsc clean.
- Mutations bite exactly: **M1**→G-HI-3 (1/7), **M2**→G-HI-4 (1/7), **M3** drop execute re-check→G-HI-1+G-HI-4+HI-5 (3/5, proves the re-check is real not preview-cached).
- 19 failures reproduced; fixture repair → undelete 15/15.
- `plugin-tests` (hosting both the goldens and the 19 red) is **NOT a required check** — the 19 red don't block the required gate, but leaving main's real-DB job red is poor hygiene, so fold the repairs in.

### Findings
- **P1/P2**: none.
- **P3-1 (hardening)**: `DERIVED_FIELD_TYPES` omits the 4 system-computed types in canonical `SYSTEM_FIELD_TYPES` (`createdTime/modifiedTime/createdBy/modifiedBy`). Verified **not** a live false-positive today (they're read-injected via `query-service.injectSystemFieldValues`, never persisted to `data`; healthy update snapshots `{...previousData,...patch}` where `previousData` is the **raw** column). But the invariant is unpinned by any golden — a future change persisting a system field into `data` (or sourcing `previousData` via the read model) would refuse **every healthy record on every sheet with a Created/Modified field**. Recommend `DERIVED ∪ isSystemFieldType` + a `modifiedTime` golden leg.
- **P3-2**: false-negative if a polluted user field is retyped user→derived before revert (narrow; retype usually clears data).
- **P3-3**: `comparator_error` fail-closed branch untested (structurally sound).
- **P3-4**: after the undelete repair, test `(g)`'s squatter now trips HISTORY_INCOMPLETE before the `resurrectScopeHash` drift path it was written to exercise — stays green but loses that coverage; give the squatter a revision when repairing.
- **P3-5 (info)**: record-level version-restore is not gated — deliberate scope bound (not a `reconstructRecordsAtT` consumer), not a hole.

*Only-reviewed; no merge/push/arm. Full report: `/tmp/pr4234-review-claude-20260713.md`.*

zensgit:
# PR #4234 — DELTA re-gate (fixture-repair commit `40825ad51`)

- **PR**: zensgit/metasheet2 #4234 — §0.6 HISTORY_INCOMPLETE precheck (D-1c), branch `claude/multitable-history-incomplete-precheck-20260713`
- **Head SHA (verdict scope)**: `40825ad5140f6822849b3ef0ab6ba4ed0176cf2f`
- **Delta verified**: `a64d9e326..40825ad51` (the fixture-repair commit only; precheck runtime already gated 0 P1/P2 in the prior pass)
- **Worktree**: `/private/tmp/ms2-rev-pr4234` (one-time, off refs/reviews/pr4234; canonical checkout untouched)
- **Method**: full-text read of PR-side files, source-level call-ordering trace, real-DB spot-run on a fresh dedicated PG (`ms2_pr4234_regate`, migrated with CI's exact `MIGRATION_EXCLUDE`), two mutation proofs, tsc.

## Verdict: **MERGE_CLEAN** (scoped to head `40825ad51`)
0 P1 · 0 P2 · 0 P3 · 1 NIT (acknowledged, mitigated). Every bucket disposition is HONEST — no §0.6 suppression, no masked regression, no smuggled scope.

---

## HARD scope check — PASS
`git diff a64d9e326..HEAD --name-only` = exactly 5 files:
```
packages/core-backend/src/multitable/history-integrity-precheck.ts
packages/core-backend/tests/integration/multitable-history-incomplete-precheck-realdb.test.ts
packages/core-backend/tests/integration/multitable-reset-pit-realdb.test.ts
packages/core-backend/tests/integration/multitable-undelete-pit-inbound-replay-realdb.test.ts
packages/core-backend/tests/integration/multitable-undelete-pit-realdb.test.ts
```
Touches **NONE** of the 8 A1–A8 write sites (`records.ts` / `automation-executor.ts` / `automation-service.ts` / `univer-meta.ts` write paths) and **NOT** the OD-6 guard. No route/index/runtime wiring smuggled. (`routes/univer-meta.ts` is NOT in the delta — the precheck call-sites there are unchanged base-commit code, read only for ordering evidence.)

## Runtime delta — pure widening (PASS)
`history-integrity-precheck.ts`: `isDerivedFieldType` now routes through new `isNonUserAuthoredFieldType(x) = DERIVED_FIELD_TYPES.has(x) || isSystemFieldType(x)`. `isSystemFieldType`/`SYSTEM_FIELD_TYPES` are field-codecs' canonical set `{autoNumber, createdTime, modifiedTime, createdBy, modifiedBy}` (autoNumber already in both → deduped). This is a strict **superset** of the prior exclusion → can only *relax* refusals, never add one. For healthy records both sides canonicalize to `'null'` (system values are never written into `data`; derived at read-time), so behaviour is unchanged except for the legacy/imported stray-value-under-a-system-field case the P3-1 golden pins. No other behavioural change. **tsc: exit 0, 0 errors.**

---

## Per-bucket honesty verdict

### A-bucket (undelete fixtures) — HONEST (capture-complete, not §0.6-suppressing)
Every seeded revision's snapshot is **byte-identical to the row's live data** — a legitimate capture, not a fake stub to satisfy §0.6:
- `undelete-pit-realdb` L: live `{NAME:'L',LINK:[U]}` ↔ `rev(L,1,'create',{NAME:'L',LINK:[U]},T0)`. Match.
- `inbound-replay` L: same, `{NAME:'L',LINK:[U]}` at T0. The decline test additionally captures the UPDATE: live becomes `{NAME:'L',LINK:[]}` + `update` rev v2 snapshot `{NAME:'L',LINK:[]}` at `2026-01-01T12:00:00Z` (> T0 so it is the latest rev; < asOf=T1 so it is L's T1-target). Match; `replayed:0,total:1` assertion unchanged.
- squatter U: live `{NAME:'squatter'}` ↔ `rev(U,7,'create',{NAME:'squatter'},T2)`. `created_at=T2 > execute-time T1` keeps it out of the T1 reconstruction, so U still reclassifies resurrect→revert — drift preserved.

**Run: `undelete-pit-realdb` 15/15, `inbound-replay` 4/4 — GREEN.**

### B-bucket (reset-pit g1/g2 re-pin) — HONEST (real TOCTOU, nothing silently weakened)
Both inject a genuine uncaptured write between preview and execute that §0.6 exists to catch:
- g1: `E` inserted as a **zero-revision live row** → `zero_revision_live_row`.
- g2: raw `UPDATE` bumps D's data+version with no revision → `content_mismatch`.

Source-confirmed ordering (`routes/univer-meta.ts`): `reset-execute` calls `computeSheetReset` (precheck at L10031, returns `{historyIncomplete}`) → returns `sendHistoryIncomplete` (409) at **L10376**, strictly BEFORE `verifyPitResetPreviewIdentity` (L10381) and its `PREVIEW_IDENTITY_INVALID` at L10388. So `HISTORY_INCOMPLETE` is exactly what the code now emits — the re-pin tracks the runtime, not a downgrade. `status===409` and **all zero-write assertions unchanged** (`E`/`D` not deleted, D's edit survives, A/B still v2). Restore-marker comments present (L172, L186). No previously-asserted success-path became silently unreachable — these tests never asserted a success path.

**Mutation (neuter §0.6 → always ok):** g1/g2 red with `expected 'PREVIEW_IDENTITY_INVALID' to be 'HISTORY_INCOMPLETE'` — i.e. the pre-§0.6 fallback IS `PREVIEW_IDENTITY_INVALID`, still 409 + zero-writes. This proves (a) the tests genuinely require §0.6 to fire, and (b) there is no masked regression: the same inputs are refused by *both* guards, only the code changed. G-HI-1/G-HI-4/HI-5 goldens also red under this mutation → §0.6 is load-bearing. Restored clean.

### P3-4 (undelete-pit test (g) squatter) — HONEST (drift path still exercised, strengthened)
The squatter now gets a capture-complete rev so §0.6 passes, and the delta **adds** `expect(x.body?.error?.code).toBe('PREVIEW_IDENTITY_INVALID')` — pinning the intended `resurrectScopeHash`-drift path rather than accepting an incidental §0.6 refusal. Net stricter, not neutered.

### P3-1 (system-type union) — HONEST + mutation-proven
Union uses canonical `SYSTEM_FIELD_TYPES` correctly. New golden seeds an adversarial state: `modifiedTime` field carrying a STALE value differing between live (`2026-06-01`) and snapshot (`2025-01-01`), `string` field matching → asserts revert-preview `200` + `visibleRevertCount:0`.
**Mutation (drop `|| isSystemFieldType`):** exactly the P3-1 golden reds (`expected 409 to be 200`); the other 8 precheck tests stay green. Reproduced and restored.

---

## Test / mutation ledger (fresh PG `ms2_pr4234_regate`, CI MIGRATION_EXCLUDE)
| Run | Result |
|---|---|
| 4 delta files + 2 neighbors (`revert-pit`, `record-reconstructor`) | 6 files / 57 tests GREEN |
| precheck golden | 9/9 · reset-pit 13/13 · undelete-pit 15/15 · inbound-replay 4/4 |
| MUT drop system-type exclusion | 1 failed = **only** P3-1 golden; 8 pass ✅ |
| MUT neuter §0.6 | g1,g2 + G-HI-1/4 + HI-5 red (5); rest pass ✅ |
| tsc `--noEmit` core-backend | exit 0, 0 errors ✅ |

All 4 delta files are in the `plugin-tests.yml` real-DB gate (CI-covered, 1 occurrence each).

## Findings
- **NIT-1 (acknowledged, mitigated)** — reset-pit g1/g2 were the only two `PREVIEW_IDENTITY_INVALID` assertions in reset-pit; both re-pinned, so the reset-specific `verifyPitResetPreviewIdentity` **deleteScopeHash-divergence** integration assertion is temporarily shadowed by §0.6. Mitigations: (a) restore-marker comments flag the intended restore "after the corresponding revision-slice lands"; (b) `verifyPitResetPreviewIdentity` has unit coverage (`tests/unit/restore-preview-identity.test.ts`); (c) the §0.6-neuter mutation proves the guard is wired and reachable; (d) the undelete-pit squatter still exercises the analogous "capture-complete drift → previewIdentity fires" path. Recommend the follow-up slice add a capture-complete reset-drift test (E with a matching `create` rev → §0.6 passes → deleteScopeHash diverges → `PREVIEW_IDENTITY_INVALID`), mirroring the A-bucket squatter. Not merge-blocking.

No P1 / P2 / P3.

PR #4235, opened 2026-07-13
feat(multitable): D-1c §0.6 — HISTORY_INCOMPLETE fail-closed precheck for PIT-Revert/Reset

zensgit:
## Summary

Implements the owner's headline hard-lock from the **RATIFIED** D-1c design-lock, §0.6 (`docs/development/multitable-global-history-d1c-form-submit-edit-uncaptured-revision-design-lock-20260712.md`, currently on branch `docs/multitable-d1c-form-submit-edit-revision-lock` / PR #4187 — not yet merged to `main`; this PR implements only §0.6, per the owner's ordering "§0.6 lands FIRST and alone").

A shared, fail-closed precheck consumed by **both** destructive PIT paths — sheet Revert-to-T (T8-1) and sheet Reset-to-T (T8-2), preview **and** execute — that refuses to act on a sheet containing any record whose history is untrustworthy (the uncaptured-write pollution the D-1c five write-path slices will backfill next, landing in a later, separate PR per owner ordering).

- **Shared precheck fn**: `checkSheetHistoryIntegrity` — `packages/core-backend/src/multitable/history-integrity-precheck.ts:65`
- **Two consumers**: called once inside `computeSheetRevert` (`packages/core-backend/src/routes/univer-meta.ts:9956`) and once inside `computeSheetReset` (`:10035`) — each of those is already the single function its own preview *and* execute route invoke fresh (execute always RE-ENUMERATES, never a cached preview), so the TOCTOU re-check (item 4) falls out structurally instead of needing a second wiring point. All four routes (`revert-preview`, `revert-execute`, `reset-preview`, `reset-execute`) check the new `historyIncomplete` result **before** minting/verifying any `previewIdentity` and before any write, returning a unified `HISTORY_INCOMPLETE` (409).
- **Projection field-type exclusion source**: `packages/core-backend/src/multitable/field-codecs.ts` — new `DERIVED_FIELD_TYPES`/`isDerivedFieldType` (`formula`, `rollup`, `lookup`, `autoNumber`), pinned to the exact `MultitableFieldType` literals already declared there (mirrors the existing `SYSTEM_FIELD_TYPES`/`isSystemFieldType` pattern). `link` fields stay **in** the projection (OD-4 — link ids are ordinary user-authored `data`) but compare via the canonical order-insensitive `canonicalSameLinkSet`/`normalizeLinkIds` (imported from `record-restore-diff.ts`, injected the same way `computeRecordRestoreDiff` already takes it), not raw value-equality.

## The two traps (both hit in review)

1. **Projection to user-authored fields only.** A full-`data` diff would flag every healthy record on every formula/rollup/lookup/auto-number sheet (those engines materialize computed keys into `data` post-commit with no revision, by design). Fixed via the `DERIVED_FIELD_TYPES` exclusion — pinned by **G-HI-3** (positive control).
2. **Enumerate LIVE rows, not the reconstructed set.** `computeSheetReset` pushes any live row absent from the reconstruction into its delete-set unconditionally; a precheck that iterated only the reconstructed set would never see a zero-revision live row and would miss exactly the case that causes a silent delete. Fixed by enumerating `FROM meta_records` directly (LEFT JOIN LATERAL each record's own latest revision, no `asOf` bound) — pinned by **G-HI-4**.

A third false-positive risk (found in independent review of this PR, not in the design-lock text) — **link-field order/representation**: a raw `JSON.stringify` comparison would misreport a healthy record whose live/snapshot link array merely differs in element order or string-vs-array shape. Fixed with the canonical set-comparator; pinned by **G-HI-3-link**.

## Goldens — all real-DB (`vitest.integration.config.ts`), run against BOTH revert and reset

`packages/core-backend/tests/integration/multitable-history-incomplete-precheck-realdb.test.ts` (13 tests, all passing against a real Postgres 14).

| Golden | Result |
|---|---|
| G-HI-1 polluted (live ≠ latest revision) | refused (409 `HISTORY_INCOMPLETE`), zero writes — both preview and execute |
| G-HI-2 version-only drift (lock/unlock bump, data identical) | passes, operation proceeds |
| G-HI-3 formula-sheet healthy record (positive control for projection) | passes, operation proceeds |
| G-HI-3-link order/representation-differing healthy link (positive control for link comparator) | passes, operation proceeds |
| G-HI-4 zero-revision live record in scope | refused, zero writes — pins live-row enumeration |
| TOCTOU (pollution landing strictly between preview and execute) | caught by the execute re-check — constructed so the pre-existing `previewIdentity` scopeHash is provably blind to it (no version bump, target-keyed diff unaffected); confirmed empirically that removing the execute-time check makes this scenario return 200 and write |

**Mutation proofs performed** (temporarily patched, reran, confirmed red, reverted):
- Comparing full `data` (no derived-field projection) → **G-HI-3 and G-HI-3-link both red**.
- Iterating the reconstructed set (INNER JOIN) instead of live rows → **G-HI-4 red**.
- Removing the execute-time precheck call (added a `skipHistoryCheck` bypass exercised only from `revert-execute`) → **TOCTOU golden red** (`expected 200 to be 409` — the destructive write actually proceeds, not merely a different error code).

## CI wiring (two-point)

- `packages/core-backend/vitest.config.ts` — added to the exclude list (no-DB job cannot skip-green it).
- `.github/workflows/plugin-tests.yml` — added to the `Run multitable real-DB integration` whitelist, alongside the sibling revert/reset/undelete real-DB files.

## Pre-existing fixtures changed — please read before assuming test tampering

Three pre-existing real-DB tests broke under the new fail-closed guard because their direct-SQL-seeded **scaffolding** records (not the record under test) were live with **zero revisions** — exactly the D-1c bug's own fingerprint, and exactly what item 3 says must refuse:

- `multitable-undelete-pit-realdb.test.ts` — bystander record `L` (a live neighbour linking to the undelete target) had no revision. Added a matching create revision.
- `multitable-undelete-pit-inbound-replay-realdb.test.ts` — same `L` pattern, plus its own "neighbour declined" test mutated `L`'s data with no revision. Both fixed with matching revisions, timed to stay outside the T1 revert scope (so the fix doesn't turn `L` into a genuine revert target and confound the inbound-replay-consent property under test).
- `multitable-reset-pit-realdb.test.ts` (g1/g2) — a "record created after preview" and an "edited after preview" simulation both used uncaptured writes (exactly the D-1c bug). Added matching revisions so each golden isolates the scope-drift/version-drift property it's actually pinning, rather than being preempted by `HISTORY_INCOMPLETE`.

In every case the fix is "make the scaffolding a healthy record with a revision," not "weaken an assertion." Full regression across all 15 real-DB files that touch these four routes (176 tests) plus the full non-DB unit suite (6409 tests) pass.

## Test plan
- [x] `pnpm run type-check` clean
- [x] `pnpm exec eslint` on all changed/new source files — no new errors/warnings (2 pre-existing errors elsewhere in `univer-meta.ts`, untouched by this diff)
- [x] All 13 new goldens pass against real Postgres 14, migrated with CI's exact `MIGRATION_EXCLUDE`
- [x] All 15 real-DB files touching `revert-preview/execute` or `reset-preview/execute` pass (176 tests)
- [x] Full non-DB unit suite passes (6409 tests, 0 regressions)
- [x] 3 mutation proofs performed and confirmed red, then reverted

**Not marked ready — draft only, per instructions.**


zensgit:
Superseded by the merged #4234 (same file `history-integrity-precheck.ts`, same function). #4234 landed the live-vs-latest §0.6 precheck and is ratified. The correctness upgrade (contiguity + trusted-since, C1–C8) is tracked in #4250 + the verification/plan MD #4252; the 13 real-DB goldens here should be ported into the W0-1 impl branch. Closing to avoid a stale duplicate.

PR #4219, opened 2026-07-13
fix(multitable): R13 Lane A — form-submit CREATE/EDIT + attachment-delete now write revisions

zensgit:
## Summary

Per the ratified D-1c design-lock (docs branch `docs/multitable-d1c-form-submit-edit-revision-lock`, §0.5 rulings OD-1..OD-6) and the revision-disposition audit (`/tmp/r13-revision-disposition-audit-20260713.md`), three `meta_records` write sites mutated record data with **no** `meta_record_revisions` row, so `reconstructRecordsAtT` (the primitive under the PIT view / revert / reset) derived existence+data *purely* from revisions and could never see these writes:

- `packages/core-backend/src/routes/univer-meta.ts` — form-submit **CREATE** (`POST /views/:viewId/submit`, no `recordId`)
- same handler — form-submit **EDIT** (`recordId` present; link-field edits fold into this same UPDATE — OD-4 ruled them in scope)
- `packages/core-backend/src/routes/univer-meta.ts` — **attachment-delete** (`DELETE /attachments/:attachmentId`)

**Consequence pre-fix:** a sheet revert/reset to a timestamp after a form edit could silently overwrite the edit with the stale pre-edit value (unrecoverable — no trash row, no prior revision). A record created via form-submit before a Reset-to-T target could be misclassified as "created after T" and destroyed by the reset.

## Fix

Emit `recordRecordRevision(...)` inside the **same transaction** as each mutation (verified real `pool.transaction` with ROLLBACK-on-throw at all 3 sites), mirroring the canonical bulk-PATCH shape (`record-write-service.ts:998`):

| Site | action | source (OD-2) | actorId (OD-3) | snapshot |
|---|---|---|---|---|
| form-submit CREATE | `create` | `public-form` | verbatim (null on anonymous path) | `patch` (full row) |
| form-submit EDIT | `update` | `public-form` | verbatim | full post-merge row (`RETURNING version, data`) |
| attachment-delete | `update` | `rest` | verbatim (always known — route requires `canEditRecord`) | `{...previousData, ...patch}` |

## Proof

New real-DB golden `packages/core-backend/tests/integration/multitable-r13-lane-a-form-revision-realdb.test.ts` (13 tests, wired into both `vitest.config.ts`'s no-DB exclude list and `plugin-tests.yml`'s real-DB run-list), driving the real Express routes end-to-end per site:

1. Revision row correctness (action/source/actor/snapshot)
[...]

PR #4216, opened 2026-07-13
fix(multitable): R13 Lane B — plugin-SDK createRecord/patchRecord write meta_record_revisions

zensgit:
## Summary

Implements **R13 Lane B** per the RATIFIED design-lock `docs/development/multitable-global-history-d1c-form-submit-edit-uncaptured-revision-design-lock-20260712.md` (§0.5, OD-1 lane B) and its audit `/tmp/r13-revision-disposition-audit-20260713.md`.

Two plugin-SDK single-record write sites mutated `meta_records` (data + version bump) but wrote **no** `meta_record_revisions` row:

- `packages/core-backend/src/multitable/records.ts:546` — plugin-SDK `createRecord` (audit A5). `reconstructRecordsAtT` derives existence PURELY from revisions, so a plugin-created record was **absent** from every PIT/revert/reset view forever — and a Reset-to-T meant to *preserve* it instead **destroyed** it (audit §1(c): `computeSheetReset` can't distinguish "created after T" from "created before T but never captured").
- `packages/core-backend/src/multitable/records.ts:507` — plugin-SDK `patchRecord` (audit A2, **reproduced end-to-end** in the design-lock §3 — the "A2 PIT lie"). PIT/revert/reset/History Center kept returning the **pre-patch** value+version at every T after the patch, forever.

### Fix

Emit the revision in the **same transaction** as the mutation, mirroring `record-write-service.ts:998` / `record-service.ts:706,1392`:
- `source: 'plugin'` (OD-2), `actorId: null` (OD-3 — the plugin lane is actor-less by design, consistent with this module's own delete path's `actorId: null` / `deletedBy: null`).
- `snapshot` = the full post-write row (not the raw patch) — a link-field write (`patch[fieldId] = ids`) is captured inside it, consistent with the `meta_links` rows (OD-4).

**Unflagged pure-correctness fix** — no environment flag, mirroring D-1/D-6 precedent.

### Transaction-boundary proof (re-verified, not assumed)

[...]

PR #4245, opened 2026-07-13
fix(multitable): W0 slice ① — form-submit CREATE/EDIT emit public-form revisions (D-1c RATIFIED; attachment split to slice ⑤)

zensgit:
## Summary

Implements **W0 slice ① — form-submit CREATE/EDIT revisions**, per the RATIFIED D-1c design-lock (`docs/development/multitable-global-history-d1c-form-submit-edit-uncaptured-revision-design-lock-20260712.md`, §0.5 OD-1..OD-3, §0.5/§10 routing). §0.6 `HISTORY_INCOMPLETE` landed first (#4234, `64791a356`) and now guards recovery; this slice is the first of the five ruled runtime slices (form → plugin → automation → approval → attachment).

**Sites touched — exactly two, nothing else:**
- **A1 EDIT** — `packages/core-backend/src/routes/univer-meta.ts` (`~L14423`, inside `POST /views/:viewId/submit`, `recordId` branch)
- **A6 CREATE** — same handler, `~L14470` (no-`recordId` branch)

No attachment code, no plugin/automation/approval code, no OD-6 guard code was touched.

## The defect this closes

Both branches raw-mutated `meta_records` with **no `meta_record_revisions` row**, so `reconstructRecordsAtT` (the PIT view / revert / reset primitive, which derives existence+data purely from revisions) never saw these writes:
- EDIT: a sheet revert/reset to a T **after** a member's form edit would silently overwrite the edit with the stale pre-edit value — irrecoverably.
- CREATE: a Reset-to-T at any T after an uncaptured create could not distinguish "created after T" from "created before T but never captured", and would **destroy** a record legitimately present at T (§0.5's corrected CREATE risk).

## Fix

`recordRecordRevision(...)` emitted **inside the same `pool.transaction`** the route already runs in (verified, not assumed — see mutation results below):

- **EDIT**: `action:'update'`, `source:'public-form'` (OD-2 — names the form *surface*, not the actor's auth level), `actorId: getRequestActorId(req)` (always known — this branch is authenticated-only), `snapshot` = the **full post-merge row** (`RETURNING version, data`, not the patch).
- **CREATE**: `action:'create'`, `source:'public-form'`, `actorId: createActorId` verbatim (**`null`** on the true-anonymous path — OD-3, never fabricated), `snapshot: patch` (a create's data IS the patch).

**Required fix No. 3 — zero-row `RETURNING` fail-closed** (the review's latent P1, same class as the plugin-lane gap #4216 still needs): the EDIT branch's `SELECT ... FOR UPDATE` already locks+confirms the row exists inside this transaction, so a concurrent delete from another transaction is not expected to be reachable — but the `UPDATE`'s `RETURNING` is checked for zero rows regardless, and throws `NotFoundError` (fail-closed, not open) rather than falling through to write a spurious `update` revision for a record that's no longer there. `meta_record_revisions.record_id` carries no FK, so an unguarded write would silently persist and could resurrect a deleted record via reconstruction — this is why it's load-bearing even though today's `FOR UPDATE` makes the race itself unreachable (defense-in-depth against a future refactor that drops the lock).

## Recycled vs. changed from Draft #4219

Recycled the same-txn snapshot shape and real transaction-boundary reasoning from `claude/r13-lane-a-form-revision-fix` (#4219, left untouched/Draft as instructed). Rebuilt as a **fresh, form-only** implementation + golden file rather than cherry-picking, per the owner's required fixes:
1. **Attachment-delete (A8) split out entirely** — not present anywhere in this diff; that's slice ⑤.
2. `source='public-form'` on both branches (already correct in the draft's form hunks; the "wrong source" finding was specifically the draft's A8 attachment site using `'rest'`, which is moot here since A8 isn't touched).
3. **Zero-row fail-closed + concurrent-delete golden** — new, not present in the draft.

## Goldens — `packages/core-backend/tests/integration/multitable-d1c-form-submit-revision-realdb.test.ts` (12 tests, real Postgres 14)

| Proof obligation (design-lock §0.5) | Golden(s) |
|---|---|
| Real entry-point golden | Both sites driven through the real `univerMetaRouter()` + `POST /views/:viewId/submit`, not hand-rolled SQL |
| Same-txn `recordRecordRevision` with FULL snapshot | CREATE + EDIT correctness tests; G4 two-field merge-trap (edits one field, asserts the untouched field survives in the snapshot — pins `snapshot: data`, not `snapshot: patch`) |
| PIT-correct after fix | `reconstructRecordsAtT` returns post-create / post-edit data+version (was pre-edit/absent) |
| Failure rolls back (no half-write) | Per-site atomicity tests (trigger-injected failure into the revision INSERT ⇒ whole txn rolls back) |
| `source` asserted = `'public-form'` | Both sites; mutation-tested (see below) |
| Transaction-boundary re-verified for this slice | Mutation-tested directly against `pool.transaction` (see below) — not assumed |
| Concurrent-delete golden (required fix No. 3) | New: a `BEFORE UPDATE` trigger returning `NULL` suppresses the row-update for one specific record, deterministically reproducing the zero-row `RETURNING` condition (a genuine two-connection race is provably unreachable here — the `FOR UPDATE` lock already serializes it — so this simulates the *observable consequence* directly); asserts the response is `404` and **exactly 1 revision** (the original create) exists, no spurious `update` |
| OD-4 (link ids in `data`) | Minimal, explicitly-scoped test: a link-field edit's ids land in the revision snapshot as ordinary `data` — comment explicitly states edge-level `meta_links` history is a SEPARATE, still-unsolved lock, NOT claimed or tested here (this was tightened from an earlier draft's "real revert reproduces meta_links edges" framing, which read as overclaiming edge-level completeness) |

**Two-point wiring** (`packages/core-backend/vitest.config.ts` exclude list + `.github/workflows/plugin-tests.yml` real-DB run-list) — proved it bites: removing the `vitest.config.ts` exclude entry made the file get silently **skip-greened** (`1 skipped`, not run) in the default no-DB job; restored.

## Mutation results (all reverted via `git checkout --` after commit; DB container `ms2-d1c1-pg` removed after)

| Mutation | Result |
|---|---|
| EDIT `source: 'public-form'` → `'rest'` | 1/12 reds exactly the EDIT source-asserting golden |
| CREATE `source: 'public-form'` → `'rest'` | 2/12 reds (source golden + CREATE atomicity, whose fault-injection trigger keys off `source='public-form'`) |
| Zero-row guard: remove the `throw NotFoundError` | Concurrent-delete golden reds: `200` instead of `404` (spurious success) |
| `snapshot: normalizeJson(data)` → `snapshot: patch` (EDIT) | G4 merge-trap reds directly; cascades to redden a later revert-preview test via §0.6's sheet-wide content-mismatch precheck (expected — a real content mismatch, caught twice) |
| Delete EDIT's `recordRecordRevision(...)` call entirely | 5/12 reds (all EDIT-site goldens) |
| Delete CREATE's `recordRecordRevision(...)` call entirely | 8/12 reds (CREATE-site goldens directly + downstream EDIT goldens whose fixture depends on `createMemberRecord`'s CREATE call landing a v1 revision) |
| **Transaction boundary**: neutered `pool.transaction`'s `BEGIN`/`ROLLBACK` in `connection-pool.ts` | 3/12 reds: **both** atomicity goldens (CREATE leaves 1 leftover half-written record instead of 0; EDIT's injected-failure edit value persists instead of rolling back) + 1 cascading revert-preview failure via §0.6. Direct proof the route's atomicity depends on the real transaction wrapper, not assumed. |
| Rank-8 lock-disposition marker: removed `// lock-guarded: form-submit EDIT (B2)` above the EDIT `UPDATE` | `tests/unit/multitable-record-lock-guard.guard.test.ts` reds exactly on that site (`univer-meta.ts:14456`, disposition `null`) |

## §0.6 interaction

No §0.6 code or goldens (`history-integrity-precheck.ts`, `multitable-history-incomplete-precheck-realdb.test.ts`) were modified. §0.6 does interact functionally: its sheet-scope precheck now correctly **passes** revert/reset-preview for records created/edited via the form path, because they carry matching revisions — this is the exact pollution this slice closes. One fixture note: my new golden's `LINK_TARGET_1` seed record (a plain link target, not part of the flows under test) had to be seeded with a matching `recordRecordRevision` call, or §0.6's sheet-wide precheck would 409-refuse every `revert-preview`/`reset-preview` call in the file on account of that one unrevisioned seed row — flagging this per the task's instruction to report any §0.6-interaction-driven fixture change (this is a new fixture in a new file, not an edit to an existing §0.6 golden).

## Also touched (defensive, not required)

Two mock-based specs got a no-op `INSERT INTO meta_record_revisions` handler added so this change doesn't add a new failure mode to them:
- `packages/core-backend/tests/integration/multitable-record-form.api.test.ts`
- `packages/core-backend/tests/integration/multitable-sheet-permissions.api.test.ts`

**Neither file is wired into any `plugin-tests.yml` step today** (confirmed by grep — a pre-existing, unrelated gap: `multitable-record-form.api.test.ts`'s two submit-path tests already fail on current `origin/main` before this change, on an unrelated missing `pg_advisory_xact_lock` mock handler from `acquireAutoNumberSheetWriteLock`). Verified this is not a regression: same failure, same line, with or without this PR's changes (tested via `git stash`). Out of scope for this slice; noted for visibility. `multitable-sheet-realtime.api.test.ts` (the one of the three that IS wired into CI) already had a working `INSERT INTO meta_record_revisions` no-op and needed no change — its 3 tests pass.

## Test evidence

- `tsc --noEmit`: clean.
- Full real-DB suite (the exact 197-file `plugin-tests.yml` "Run multitable real-DB integration" list, real Postgres 14, CI's `MIGRATION_EXCLUDE`): **197 files / 1751 passed, 2 skipped, 0 failed**.
- Full no-DB suite (`pnpm --filter @metasheet/core-backend test` equivalent): **466 files / 6409 passed, 180 skipped (expected DB-gated exclusions), 0 failed**.
- New golden file standalone: **12/12 passed**.
- `tests/unit/multitable-record-lock-guard.guard.test.ts`: **4/4 passed** (and mutation-proven above).

## Explicitly UNTOUCHED

- **Attachment (A8 / slice ⑤)** — `univer-meta.ts:15693`, `multitable-attachments.api.test.ts` — zero changes.
- **Plugin (A2/A5), automation (A3/A4), approval (A7)** — zero changes; those are slices ②③④.
- **OD-6 revision-disposition guard** (#4227) — not touched; still Draft, lands last per the routing.
- **§0.6 `HISTORY_INCOMPLETE`** (#4234, already on `main`) — precheck code and its goldens (G-HI-1..4) — zero changes.
- **`docs/development/multitable-global-history-d1c-form-submit-edit-uncaptured-revision-design-lock-20260712.md`** — the design-lock itself — zero changes.

## Not merged, not armed

Per instructions — opened as a normal PR for review. No auto-merge, no `--admin`, no keep-sync arming.

refs #4187 (D-1c design-lock, RATIFIED 2026-07-13)


zensgit:
## Adversarial review — VERDICT: MERGE_CLEAN (scoped to head `fa4f4545e`)

Refute-first pass against a fresh worktree off the PR ref + real Postgres 15 (CI's `MIGRATION_EXCLUDE`), running the golden via `--config vitest.integration.config.ts` exactly as `plugin-tests.yml`. No P1/P2. One NIT (a single transient test flake; baseline 5/5 stable).

### Scope — tight
Only `univer-meta.ts` is production source, and only the two ruled branches (A1 EDIT `~14456`, A6 CREATE `~14543`). Confirmed **not** touched: attachment-delete A8 (`:15693`), plugin `records.ts`, automation, the OD-6 guard (#4227), §0.6 `history-integrity-precheck.ts` + its goldens, and the design-lock doc. Other files are wiring (`plugin-tests.yml`, `vitest.config.ts`) + the new golden + two inert mock specs.

### Load-bearing checks (all mutation-reproduced, restored after)
- **Full snapshot (crux):** EDIT `snapshot = normalizeJson(RETURNING …, data)` = FULL post-merge row (matches healthy `record-write-service.ts:992-1008`); CREATE `snapshot = patch` = full create row (matches `record-service.ts:715`). `snapshot: patch` on EDIT reds the G4 merge-trap **and** the §0.6 destructive-leg — the field-projection false-positive §0.6 was built around is not reintroduced. `snapshot: {}` on CREATE reds 4 CREATE/PIT goldens (symmetry).
- **PIT-correct:** deleting the EDIT revision → 5/12 red incl. the reconstruct golden falling back to the pre-edit value (the D-1c bug).
- **Atomicity real:** both `recordRecordRevision(query, …)` calls are inside `pool.transaction` on the txn-scoped `query`. Neutering `BEGIN` in `connection-pool.ts` reds both atomicity goldens + §0.6 cascade (3/12). Not assumed.
- **Concurrent-delete:** race is genuinely unreachable today (`SELECT … FOR UPDATE` at 14435; pre-existing delete hits `NotFoundError` at 14439) → guard is defense-in-depth. Removing the zero-row `throw NotFoundError` reds the golden (200 vs 404).
- **Source + actor (OD-2/OD-3):** `source='public-form'` both branches (EDIT→`'rest'` reds); anon CREATE `actorId=null`, never fabricated; EDIT is authenticated-only because anon EDIT is blocked upstream at 14233-14238 (400) so `getRequestActorId` is non-null.
- **OD-4:** link ids fold into `patch` (14339/14356) → captured in `snapshot` as data; golden explicitly disclaims edge-level `meta_links` completeness. Not overclaimed.

### Two implementer-flagged items — adjudicated
- **LINK_TARGET_1 seed revision:** legitimate capture-complete fixture (§0.6 legitimately 409s any unrevisioned live row on the scanned sheet), not a workaround hiding anything.
- **No-op revision mocks** in the two `.api.test.ts` files: both are excluded from the no-DB job (`vitest.config.ts:286/296`) **and** absent from every workflow → run in no CI step → genuinely inert, cannot mask.

### Also
- Two-point wiring bites both ways: removing the `vitest.config.ts` exclude → golden `12 skipped` (skip-green) in the no-DB lane; every integration-config run passes an explicit file list so the `plugin-tests.yml` run-list entry is load-bearing.
- Rank-8 marker at `univer-meta.ts:14455` load-bearing (removal reds the guard at `:14456`, disposition null). `tsc --noEmit` clean. Golden 12/12 (5/5 stable); neighbours 22/22.

### NIT — one transient flake
The A6 `reconstructRecordsAtT after the create` golden flaked RED once during a mutation run, then passed; baseline is 5/5 stable. The flaked assertion (`exists:false` on a just-created record) is arithmetically impossible from the boundary math (`asOf = created_at + 5ms`), so it is environmental (pool/settle hiccup), not a product race. Failure direction is false-RED, so it cannot hide a regression — just flag so an occasional CI red isn't a surprise.

Full report: `/tmp/pr4245-review-claude-20260713.md`

*Reviewed at head `fa4f4545e5d98daf7f60a05d2ee226b417040246`. Review only — no merge/push/arm.*

PR #4246, opened 2026-07-13
fix(multitable): W0 slice ② — plugin-SDK createRecord/patchRecord emit plugin revisions + concurrent-delete fail-closed (D-1c)

zensgit:
## Summary

Second of the five ratified D-1c slices (§0.5/§10). §0.6 (#4234) and slice ① (#4245, `8615879a2` on main) already landed. This slice fixes the **plugin-SDK single-record write sites**:

- **A2 update** — `patchRecord`, `packages/core-backend/src/multitable/records.ts:507` (the `UPDATE meta_records` statement; the fail-closed guard + revision emit are added right after it)
- **A5 create** — `createRecord`, `packages/core-backend/src/multitable/records.ts:546` (the `INSERT INTO meta_records` statement; the revision emit is added right after it)

Both mutated `meta_records` with **no** `meta_record_revisions` row, so `reconstructRecordsAtT` (the primitive under the PIT view / revert / reset) derived existence+data purely from revisions and never saw these writes — the "A2 PIT lie" the RATIFIED design-lock's §3 reproduced end-to-end on this exact site (`docs/development/multitable-global-history-d1c-form-submit-edit-uncaptured-revision-design-lock-20260712.md`), and (for A5) a Reset-to-T after an uncaptured plugin CREATE could not distinguish "created after T" from "created before T but never captured" and would destroy a record legitimately present at T.

### Fix

`recordRecordRevision(...)` is emitted inside the **same transaction** as each mutation. `index.ts` wraps every plugin-SDK `createRecord`/`patchRecord` call in `poolManager.get().transaction(...)` — the **sole production wiring** (re-verified for this slice, not assumed — see Proofs below).

| | `action` | `source` (OD-2) | `actorId` (OD-3) | `snapshot` |
|---|---|---|---|---|
| patchRecord (A2) | `'update'` | `'plugin'` | `null` | `nextData` — the FULL merged row (`{...existing.data, ...patch}`), NOT the raw `patch` |
| createRecord (A5) | `'create'` | `'plugin'` | `null` | `patch` — a create's `data` IS the patch |

**OD-3 actor note**: the plugin-SDK boundary (`index.ts` `createCoreAPI()`) threads **no actor identity at all** through `createRecord`/`patchRecord` — `guardRecordNotLockedForPlugin` already treats every plugin caller as actor-less, matching the module's existing delete path (`actorId: null`). `actorId: null` here means "no actor is available", never "a known actor was discarded" — no system actor is fabricated.

### REQUIRED P1 fix — concurrent-delete fail-closed

Recycled Draft #4216 (`claude/r13-lane-b-plugin-revision-20260713`, left untouched — a parallel session's Draft, per the owner's "不合、不 rebase" instruction) with the owner's required fix applied.

`patchRecord`'s original code:
```ts
const version = Number((updated.rows as any[])[0]?.version ?? existing.version + 1)
```
masks a **zero-row** `UPDATE ... RETURNING`. Unlike slice ①'s form-submit EDIT branch (which already holds a `SELECT ... FOR UPDATE` before its own `UPDATE`), `patchRecord`'s `getRecord` and `guardRecordNotLockedForPlugin` reads take **no row lock** — this UPDATE is the *first* point that actually locks the row, so a concurrent DELETE of the record between those reads and the UPDATE is a **real, reachable race**, not a theoretical one. Left unguarded, the fallback would synthesize a version and fall through to write a spurious `update` revision for a record that no longer exists. Because `meta_record_revisions.record_id` carries **no FK** (migration `zzzz20260430172000`), that revision would persist forever and could **resurrect the deleted record** via `reconstructRecordsAtT`.

**Fix**: fail closed — a zero-row `RETURNING` throws `MultitableRecordNotFoundError` (the module's existing not-found error, same shape as the D-2 delete path's own zero-row guard) *before* any revision is written or `nextVersion` is synthesized.

`createRecord`'s `INSERT ... RETURNING` gets **no** symmetric guard — a bare `INSERT` with no `ON CONFLICT` clause cannot return zero rows without the statement itself throwing; there is no "concurrently deleted" analogue for a row that doesn't exist yet (mirrors slice ①'s form-submit CREATE branch, which likewise added none — a fabricated symmetry would be dead code).

### The concurrent-delete golden — genuine two-connection lock race, not a sleep heuristic

Per this line's TOCTOU doctrine (sequential reasoning about a race proves nothing), the golden constructs a **real** race using Postgres row-lock contention, not a suppression trigger or a `setTimeout`:

1. A dedicated raw connection (`poolManager.get().getInternalPool().connect()`) `BEGIN`s and `DELETE`s the record — uncommitted, holding the row lock.
2. The **real** `sdk.patchRecord(...)` call (via `createCoreAPI()`) is fired: its plain `getRecord`/`guardRecordNotLockedForPlugin` reads succeed immediately (readers don't block on an uncommitted writer under MVCC), but its `UPDATE` blocks behind the holder's lock.
3. A deterministic barrier (`pg_stat_activity` + `pg_blocking_pids`, correlated to the holder's own backend pid — the same technique as `multitable-d2-sidedoor-delete-recoverability-realdb.test.ts`'s `waitUntilBlockedOnRecordLock`) confirms the UPDATE is genuinely parked before the holder commits.
4. The holder commits → the `UPDATE` resumes under READ COMMITTED semantics, discovers the row is gone, and affects **zero rows**.

Assertion: the fix throws `NotFound`, and exactly **one** revision exists for the record (the original `create`) — no spurious `update`. Reran 3x locally with no flake (~70-80ms each, barrier-gated not timing-lucky).

## Per-slice proof obligations (design-lock, all required — same as slice ①)

All driven through the **real production plugin-SDK entry point** — `MetaSheetServer`'s (private) `createCoreAPI()` factory, the exact object `index.ts` binds into the plugin IoC container — never a hand-rolled `query` calling `records.ts` functions directly (this is what makes the atomicity + txn-boundary proofs real, per the D-2 G18 lesson).

| Obligation | Where | Mutation proof |
|---|---|---|
| Real entry-point golden | every test in `multitable-d1c-plugin-revision-realdb.test.ts` drives `sdk.createRecord`/`sdk.patchRecord` | — |
| Same-txn `recordRecordRevision` with FULL post-write snapshot | patch: `snapshot: nextData` (full merged row); create: `snapshot: patch` | `nextData → patch` mutation reddens the G4 two-field merge-trap golden (confirmed: `undefined` where `'g4-notes-untouched'` expected) |
| PIT-correct after fix | `reconstructRecordsAtT` before/after golden | covered by same-source mutation (deleting the call breaks both PIT goldens) |
| Failure rolls back (same-txn atomicity) | ATOMICITY goldens (patch + create), injected trigger on `meta_record_revisions` INSERT | unwrapping `poolManager.get().transaction(...)` at BOTH `index.ts` call sites reddens the respective ATOMICITY golden (patch: row left at `version:2`/`should-never-stick` instead of rolled back to `version:1`; create: orphan record row persists with length 1 instead of 0) — restored after each check |
| `source='plugin'` asserted | both sites | `'plugin' → 'rest'` mutation reddens 2 tests (source assertions) |
| Transaction-boundary re-verified for THIS slice (D-1 偏差1) | see above | same mutation as atomicity — proves the txn isn't merely assumed from D-1/D-2 |
| **Concurrent-delete golden (P1 fix)** | genuine lock-race test | neutering the zero-row guard (`if (false && ...)`) reddens the golden; DB check confirms the exact spurious `update` revision (v2) the fix prevents |

Every mutation was applied, run, confirmed red, then restored via `git checkout HEAD -- <file>` against a pre-mutation commit (never a broad checkout).

## §0.6 interaction

**None.** This slice's PIT-correctness proof calls `reconstructRecordsAtT` **directly** — it never exercises the §0.6 `HISTORY_INCOMPLETE` revert/reset preview routes, so no seed-record revision fixture is needed against it (unlike slice ①, which drove `revert-preview`/`reset-preview` and needed a fixture note). §0.6 itself and its goldens are untouched.

## Test evidence (this session, real Postgres 14, CI's exact `MIGRATION_EXCLUDE`)

- New golden file: 11/11 passed, reran 3x for the concurrent-delete race with no flake.
- `tests/unit/multitable-record-lock-guard.guard.test.ts` (rank-8): 4/4 passed — the existing `// lock-guarded:` marker on `patchRecord`'s `UPDATE meta_records` is untouched; no new `UPDATE`/`DELETE meta_records` site was added (only `createRecord`'s `INSERT`, which the rank-8 scanner does not match, matching the design-lock's own note that "the rank-8 guard ignores INSERT").
- `tests/unit/multitable-records.test.ts` (mock-based unit suite for `records.ts`): 15/15 passed — its mock query already handles `INSERT INTO meta_record_revisions`, so no defensive mock update was needed (unlike slice ①'s two mock-based specs).
- `tests/integration/multitable-history-incomplete-precheck-realdb.test.ts` (§0.6 goldens): 9/9 passed, untouched.
- `tests/integration/multitable-d1c-form-submit-revision-realdb.test.ts` (slice ① goldens): 12/12 passed, untouched.
- Full core-backend no-DB suite (`CI=true pnpm --filter @metasheet/core-backend test`, mirrors the required `test (20.x)` step): **466 files passed, 180 skipped (DB-gated); 6409 tests passed, 1523 skipped; 0 failed.**
- `pnpm lint`: exit 0. `pnpm type-check` (repo-root, incl. `tsc --noEmit` for core-backend): exit 0.

**Follow-up commit (`654ed1750`)**: an advisor review flagged that if the concurrent-delete golden's lock-wait barrier ever timed out, releasing the holder connection without first ending its open transaction would return it to the pool `idle in transaction` — still holding the DELETE's row lock — and leave the fire-and-forget `patchPromise` blocked forever, which could cascade into sibling `describeIfDatabase` files sharing the one Postgres in `plugin-tests.yml`'s real-DB lane. Added an unconditional `ROLLBACK` (safe no-op after a successful `COMMIT`) before `holder.release()` in the `finally` block. Re-ran the full file (11/11 pass) and the zero-row-guard mutation (still reddens correctly) after the fix.

## CI wiring (two-point, per the line's convention)

- `packages/core-backend/vitest.config.ts` exclude list — added `tests/integration/multitable-d1c-plugin-revision-realdb.test.ts`.
- `.github/workflows/plugin-tests.yml` — added the file to the `Run multitable real-DB integration` run-list and its step-name suffix.
- Fetched `origin/main` immediately before editing `plugin-tests.yml` (no drift at edit time) — both edits are pure single-line additions to a shared, union-edited list; no other lane's entries were touched.

## Out of scope, untouched

- A3/A4 automation update/create (`automation-executor.ts`) — slice ③.
- A7 approval `resultWriteback` — slice ④, `source='approval'`.
- A8 attachment-delete — slice ⑤, `source='attachment'`.
- The OD-6 revision-disposition guard (#4227) — separate rung, lands last.
- §0.6 `HISTORY_INCOMPLETE` (#4234, already landed) and its goldens — not modified, not exercised via the revert/reset routes (see §0.6 interaction above).
- Draft #4216 (`claude/r13-lane-b-plugin-revision-20260713`) — left untouched, per the owner's instruction; this PR recycles its `records.ts` hunks conceptually but reimplements on a fresh branch off `origin/main` with the required fix applied and a new test file.

## Not merged / not armed

This PR (a fresh slice-② branch off `origin/main`, analogous to slice ①'s #4245) is opened for review only, per this line's standing PR-review discipline — not merged, not `--auto` armed. It is distinct from the pre-existing R13 lane Drafts (#4216/#4219/#4220/#4227), which the owner separately ordered to stay Draft/unmerged/un-rebased until their own turn in the five-slice sequence; #4216 specifically stays untouched by this PR (see below).

refs #4216 (the recycled Draft — left OPEN and untouched, not superseded by this PR alone; the owner's per-slice sequencing governs its disposition)


zensgit:
## Adversarial gate review — VERDICT: MERGE_CLEAN (scoped to head `654ed1750`)

Independent refute-first review on a fresh worktree off PR head, isolated PG migrated with CI's exact `MIGRATION_EXCLUDE`. **No P1, no P2.**

**Scope** — only `records.ts` (the two plugin branches) is production source; CI wiring is a pure union edit (multitable run-list 178→179, nothing dropped). None of `univer-meta.ts` / `automation-executor.ts` / attachment(A8) / OD-6 guard(#4227) / §0.6 precheck+goldens touched. rank-8 marker at `records.ts:505` intact.

**The crux checks, reproduced empirically:**
- **Snapshot completeness (both)** — `patchRecord` snapshots `nextData` = exactly the UPDATE'd `data` (G4 two-field merge-trap green); `createRecord` snapshots `patch`, with auto-number folded into `patch` in JS (`records.ts:141`) *before* the INSERT writes `JSON.stringify(patch)` and no trigger/default augments `data` → `patch == data`. The asymmetry is sound; no §0.6 false-positive surface.
- **P1 concurrent-delete fail-closed is REAL** — golden ran 3× no flake; barrier is `pg_blocking_pids`-deterministic (throws on timeout, never a sleep). Neutering the zero-row guard (`if (false && …)`) reds the golden (`undefined ≠ Error`) and the revision-count probe shows **1→2 (the spurious `update`)**; with no FK on `meta_record_revisions.record_id` (verified) that would resurrect the deleted row. Follow-up `654ed1750` `ROLLBACK`-before-`release()` correctly prevents `idle in transaction` pool poisoning of sibling `describeIfDatabase` suites.
- **Same-txn atomicity is REAL** — unwrapping the `patchRecord` txn at the `index.ts` call site reds the atomicity golden (record left at `version:2`/`should-never-stick`).
- **PIT** — deleting the revision write reds the PIT golden. **Source** — `'plugin'→'rest'` reds the source assertion. **actorId:null** correct per OD-3 (CoreAPI `MultitableRecordsAPI` interface carries no actor field; nothing dropped).
- **Full caller enumeration** — the only production callers of `records.ts` `createRecord`/`patchRecord` are the two txn-wrapped `index.ts` plugin bindings (605/643) and the `plugin-scope.ts` scope-decorator that delegates straight back to them. No caller bypasses the transaction, mislabels source, or escapes A2/A5 scope.

**Runs (my PG):** new golden 11/11 (×3 race, no flake); neighbour slice ① 12/12; rank-8 guard 4/4; mock unit 15/15; `tsc --noEmit` exit 0. All 4 mutation checks red-then-restored, tree clean.

**Findings:** none blocking. NIT: the `?? existing.version + 1` / `?? 1` fallbacks are now dead after the zero-row guard (harmless). Observation (out of scope): `patchRecord` is a pre-existing RMW on the whole `data` blob (last-write-wins under concurrent *patches*); this slice faithfully records what was written and does not worsen it.

Full detail: `/tmp/pr4246-review-claude-20260713.md`.

*Review-only: no merge, no push, no --auto arm.*

PR #4248, opened 2026-07-13
fix(multitable): W0 slice ④ — approval resultWriteback emits approval revisions (D-1c)

zensgit:
## Summary

Per the RATIFIED D-1c design-lock (`docs/development/multitable-global-history-d1c-form-submit-edit-uncaptured-revision-design-lock-20260712.md`, §0.5 OD-1..OD-3, §0/§7a site A7), fourth of five ratified slices (slices ①/②/③ landed as #4245/#4246/#4247):

**Exact site**: `packages/core-backend/src/multitable/automation-service.ts:2818` (line numbers pre-change) — the shared `applyResultWritebackPatch` tail used by BOTH the same-base (`writeApprovalResultBack`, `:2696`) and cross-base (`writeApprovalResultBackCrossBase`, `:2762`) approval `resultWriteback` paths. It mutated `meta_records` with **no `meta_record_revisions` row**, so `reconstructRecordsAtT` (the PIT view / revert / reset primitive) never saw an approval resultWriteback — the same "PIT lie" class slices ①/②/③ closed for form-submit, plugin-SDK, and automation.

## Fix

- Emit `recordRecordRevision(...)` inside **one transaction** wrapping the lock-check SELECT + the UPDATE (now `RETURNING version, data`) + the revision INSERT. **This path was NOT previously transactional** — there was no `RETURNING`/rowCount check on the UPDATE at all before this slice (D-1 "偏差1" caution, re-verified here, not assumed).
- New private `withTransaction` helper on `AutomationService`, wired the SAME way the constructor already hard-wires `deps.transaction` for the executor: a real `poolManager.get().transaction(...)`.
- `source='approval'` (OD-2 — names the write entry point, not the auth identity).
- `actorId=opts.chainActorId ?? null` (OD-3 — identical to `opts.lockActorId` at BOTH call sites: the approval actor for same-base, the TRIGGER actor for cross-base; never a fabricated system actor).
- `RecordRevisionSource` (`record-history-service.ts`) gains `'approval'` explicitly — was a dead `| string` slot per OD-2, now named for type-safety (union stays `| string`, non-blocking per the lock).

## Zero-row fail-closed determination

[...]

PR #4247, opened 2026-07-13
fix(multitable): W0 slice ③ — automation create_record/update_record emit automation revisions (approval split to slice ④) (D-1c)

zensgit:
## Summary

Implements **W0 slice ③** per the RATIFIED D-1c design-lock (`docs/development/multitable-global-history-d1c-form-submit-edit-uncaptured-revision-design-lock-20260712.md`, §0.5 OD-1..OD-3, §0/§7a) — third of five ratified slices. Slice ① (form-submit, #4245) and slice ② (plugin-SDK, #4246) are landed on main.

**Exact sites** (both in `packages/core-backend/src/multitable/automation-executor.ts`):

| # | Site | Action | Line (origin/main) |
|---|---|---|---|
| A3 | `executeUpdateRecord` | automation `update_record` | `automation-executor.ts:2217` |
| A4 | `executeCreateRecord` | automation `create_record` | `automation-executor.ts:2475` |

Both raw-mutated `meta_records` with **no** `meta_record_revisions` row, so `reconstructRecordsAtT` (the PIT view / revert / reset primitive — derives existence+data purely from revisions) never saw the writes. A4 additionally means a Reset-to-T after an uncaptured automation CREATE could not distinguish "created after T" from "created before T but never captured" and would destroy a record legitimately present at T (§0's corrected CREATE-destructive risk).

## Fix

`recordRecordRevision(...)` now runs inside the **same transaction** as each mutation (`this.withTransaction(...)`, the same helper D-1's `executeDeleteRecord` already uses):

- **update_record**: the lock-check SELECT, the UPDATE (now `RETURNING version, data`), and the revision INSERT all moved inside `withTransaction`. `action='update'`, `source='automation'` (OD-2 — names the write entry point, not the auth identity), `actorId=context.actorId ?? null` (OD-3 — the automation actor when present; never a fabricated system actor), `snapshot=` the FULL post-merge row from `RETURNING` (not the patch — D-1c §4.2 merge trap).
- **create_record**: the INSERT and its birth revision now run inside `withTransaction`. `action='create'`, `source='automation'`, `actorId` per OD-3, `snapshot=data` (a create's data IS the submitted patch).

## Zero-row fail-closed determination

[...]

PR #4249, opened 2026-07-13
fix(multitable): W0 slice ⑤ — attachment-delete cell-strip emits attachment revisions (D-1c; final write-site slice)

zensgit:
## Summary

Per the RATIFIED D-1c design-lock (`docs/development/multitable-global-history-d1c-form-submit-edit-uncaptured-revision-design-lock-20260712.md`, §0.5 OD-1..OD-3, §0/§7a site A8) — this is the **FIFTH and FINAL** of the five ruled write-site slices (① form #4245, ② plugin #4246, ③ automation #4247, ④ approval #4248, all MERGED to main).

**Exact site**: `packages/core-backend/src/routes/univer-meta.ts`, `router.delete('/attachments/:attachmentId', ...)`, the `if (recordId && fieldId) { ... }` block inside the route's `pool.transaction(...)` (current main ~L15773-15848; the design-lock's original line number `:15693` has drifted as prior slices landed). This is the branch that strips a deleted attachment id out of the record's attachment-field cell — a real `meta_records.data` mutation that previously wrote **no** `meta_record_revisions` row, so `reconstructRecordsAtT` (the PIT view / revert / reset primitive) never saw it. A sheet revert/reset to a T after an attachment delete would silently **re-add** the deleted attachment id back into the cell — resurrecting a reference to now-purged binary/metadata.

## Fix

`recordRecordRevision(...)` emitted inside the SAME transaction as both existing writes (the record-edit `UPDATE` and the attachment soft-delete already share one `pool.transaction(...)`):

- `action='update'`
- `source='attachment'` (**OD-2** — owner ruled attachment gets its **own** source, distinct from `'rest'`/`'public-form'`/`'plugin'`/`'automation'`/`'approval'` — the write entry point is this endpoint itself)
- `actorId=getRequestActorId(req)` (**OD-3** — already `string | null`, never fabricated)
- `snapshot=` the **FULL** post-strip row (`RETURNING version, data`, not the patch) — proven with a two-attachment-on-one-field + untouched-second-field fixture so a naive `snapshot: patch` would be caught

`RecordRevisionSource` gains `'attachment'` explicitly (was a dead `| string` slot per OD-2, alongside the existing `'public-form'`/`'plugin'`/`'automation'`/`'approval'` slots).

## Zero-row determination

This route shares **slice ①'s family** (form-submit EDIT), not slice ③/④'s automation-lane contract:

- It **also** holds a `SELECT ... FOR UPDATE` lock on the exact row, in the SAME transaction, immediately before its own `UPDATE` — so a genuine concurrent-delete race is not constructible (any concurrent DELETE/UPDATE from another transaction blocks until this one commits/rolls back).
- Its **pre-existing** code carried the identical `?? <fallback>` version-synthesis mask that slices ①/② both had to close (`Number((updateRes.rows[0] as any)?.version ?? Number(recordRow?.version ?? 0) + 1)`), not a documented "silent success" leniency like the automation/approval lane.

⇒ Applied slice ①/②'s THROW contract: a zero-row `UPDATE ... RETURNING` throws `NotFoundError` (fail-closed). This route's `catch` block previously had **no** `NotFoundError` handling (it would have fallen through to a generic 500) — added a new `if (err instanceof NotFoundError) return res.status(404)...` branch, mirroring `/views/:viewId/submit`'s identical handling. Proven with a concurrent-delete golden that simulates the zero-row `RETURNING` case via a scoped `BEFORE UPDATE` suppression trigger (a genuine two-connection race is impossible here, same reasoning as slice ①).

## Proof obligations (all required, same as the four landed slices)

| Obligation | Proof |
|---|---|
| Real entry-point golden | Drives the real `DELETE /attachments/:attachmentId` route end-to-end (never a reimplementation) |
| Same-txn `recordRecordRevision`, FULL post-write snapshot | Two attachments on one field + an untouched second field — snapshot carries both the surviving sibling attachment id and the untouched field (merge-trap golden) |
| PIT-correct after fix | `reconstructRecordsAtT` at T ≥ strip returns the post-strip cell value |
| Failure rolls back (same-txn atomicity) | Postgres trigger fails the revision INSERT ⇒ **both** the record-edit UPDATE and the attachment soft-delete roll back together (proves the txn boundary spans both statements, not just one — a proof point unique to this route vs. the other four slices) |
| `source='attachment'` asserted | Direct assertion in the real entry-point golden |
| Transaction-boundary re-verified for THIS slice | The atomicity golden above; the route's existing `pool.transaction(...)` wraps both writes, confirmed by reading the route, not assumed |
| Concurrent-delete golden per the determination | `BEFORE UPDATE` suppression-trigger golden — 404, zero spurious revision, zero spurious attachment tombstone |

### Mutation-testing (local, each restored after confirming red)

1. Comment out the `recordRecordRevision(...)` call → reddens 4 goldens (real entry-point, PIT-correctness, destructive leg, atomicity).
2. `source: 'attachment'` → `'rest'` → reddens the real entry-point golden's source assertion.
3. Neuter the zero-row `throw` guard (`if (false && ...)`) → reddens the concurrent-delete golden (falls through to write a spurious revision instead of refusing).
4. `snapshot: normalizeJson(...)` → naive `snapshot: { [fieldId]: nextIds }` → reddens the real entry-point golden's merge-trap assertion **and**, as a second-order confirmation, the destructive-leg golden (§0.6's content projection now sees a live/snapshot mismatch and fail-closed-refuses the revert-preview — expected, not a bug).
5. Neuter the new `NotFoundError → 404` catch branch (`if (false && err instanceof NotFoundError)`) → reddens the concurrent-delete golden (falls through to a generic 500 instead of 404).

All five restored via `git checkout -- <file>` **after** committing the good version first (per this line's mutation-testing discipline), verified clean (`git status --short` empty) before push.

### Rank-8 lock-disposition marker

The existing `// lock-guarded: attachment-delete record edit (B3)` marker (immediately above the `UPDATE meta_records` statement) stays within its 3-line window — all new code was added entirely **after** the `UPDATE`, so no relocation was needed (unlike slice ④, which had to trim its comment to fit).

### `tsc` / test evidence

```
cd packages/core-backend && pnpm exec tsc --noEmit -p tsconfig.json   # exit 0, no errors

# real Postgres 14, CI's exact MIGRATION_EXCLUDE
DATABASE_URL=postgresql://postgres@localhost:55441/metasheet_test \
  pnpm exec vitest --config vitest.integration.config.ts run \
  tests/integration/multitable-d1c-attachment-revision-realdb.test.ts
# ✓ 7/7 passed

# sibling slices + F2 attachment security guard + §0.6 precheck, unaffected
DATABASE_URL=... pnpm exec vitest --config vitest.integration.config.ts run \
  tests/integration/multitable-d1c-form-submit-revision-realdb.test.ts \
  tests/integration/multitable-d1c-plugin-revision-realdb.test.ts \
  tests/integration/multitable-d1c-automation-revision-realdb.test.ts \
  tests/integration/multitable-d1c-approval-revision-realdb.test.ts \
  tests/integration/multitable-attachment-readgate.security.test.ts \
  tests/integration/multitable-history-incomplete-precheck-realdb.test.ts
# ✓ 52/52 passed

# rank-8 lock-disposition guard (marker window)
pnpm exec vitest run tests/unit/multitable-record-lock-guard.guard.test.ts
# ✓ 4/4 passed

# full no-DB unit suite
pnpm exec vitest run
# ✓ 466 files / 6409 tests passed, 180 files / 1523 tests skipped (real-DB describeIfDatabase, expected)

# two-point wiring bite check: new file invisible to the no-DB run (not merely skip-green)
pnpm exec vitest run tests/integration/multitable-d1c-attachment-revision-realdb.test.ts
# => "No test files found" (confirms the vitest.config.ts exclude entry is live)
```

## §0.6 interaction

Untouched. The new golden's seed records are all created via a raw `INSERT` **paired immediately** with a matching `recordRecordRevision(action:'create', ...)` call — a legitimate "capture-complete" fixture (same technique as slice ①'s `LINK_TARGET_1` seed / `G0` control), flagged in the test file's header comment. No live row on this file's sheet is ever missing a revision or content-mismatched, so the sheet-wide `revert-preview` golden does not spuriously 409 `HISTORY_INCOMPLETE` on an unrelated seed row.

## CI wiring

Two-point: `packages/core-backend/vitest.config.ts` exclude list + `.github/workflows/plugin-tests.yml` real-DB run-list. Fetched `origin/main` immediately before committing and again immediately before pushing — no new commits landed in between, so the run-list addition is a clean append after slices ①-④'s existing entries (UNION-resolved, none dropped).

## Untouched (explicit)

- form-submit CREATE/EDIT (A1/A6, slice ①, `#4245`, MERGED)
- plugin-SDK createRecord/patchRecord (A2/A5, slice ②, `#4246`, MERGED)
- automation create_record/update_record (A3/A4, slice ③, `#4247`, MERGED)
- approval resultWriteback (A7, slice ④, `#4248`, MERGED)
- §0.6 `HISTORY_INCOMPLETE` precheck (already landed, `#4234`) — exercised only incidentally (revert-preview golden), never modified
- The OD-6 revision-disposition guard — refs #4227, a **separate rung**, not part of this PR

**This completes all 8 A1–A8 bucket-A write sites across the five ruled slices.** The only remaining piece of the D-1c design-lock's implementation routing (§10) is OD-6 (refs #4227), which is explicitly out of scope here and stays a separate, independently-gated rung.

## Test plan

- [x] `tsc --noEmit` clean
- [x] New real-DB golden 7/7 passing against real Postgres 14 with CI's `MIGRATION_EXCLUDE`
- [x] Sibling slices ①-④ + F2 attachment security guard + §0.6 precheck goldens unaffected (52/52)
- [x] Rank-8 lock-disposition guard passing, marker window intact
- [x] Full no-DB unit suite green (466 files / 6409 tests)
- [x] Two-point CI wiring confirmed to bite (file invisible to no-DB run; live in real-DB run-list)
- [x] 5 mutation self-checks, each confirmed red then restored

Not merging — per this line's discipline, PR opens and stops here.


zensgit:
## Adversarial gate — VERDICT: MERGE_CLEAN (head `90f060857`)

Independent refute-first review on a fresh worktree off `origin/main`, real-DB goldens on a fresh migrated PG (CI `MIGRATION_EXCLUDE`), 6 code mutations applied+restored. No P1, no P2.

**Scope**: exactly 5 files; production source is only `univer-meta.ts` (attachment-delete region, 2 hunks) + the `RecordRevisionSource` union add of `'attachment'`. Form branches (14492/14580, slice ①), plugin/automation/approval, OD-6 guard, and §0.6 precheck all untouched. Two-point CI wiring (run-list + no-DB-lane exclude) present — both empirically confirmed (default config → "No test files found"; integration config → runs).

**Load-bearing checks (all PASS, mutation-verified):**
- Full post-strip snapshot via `RETURNING …, data` (snapshot→partial reds 2 goldens)
- `source='attachment'` OD-2 (source flip reds entry-point golden); `actorId=getRequestActorId(req)` OD-3
- PIT-correct reconstruction (disable-revision reds 4 goldens)
- Same-txn atomicity: txn-unwrap (hoist revision after commit) reds atomicity golden
- Zero-row THROW-family fail-closed, contract preserved (record-not-found still silently soft-deletes + 200; the throw is only on the FOR-UPDATE-unreachable branch, test-simulated; neutering it reds concurrent-delete golden → spurious revision)
- Real Express `DELETE /attachments/:id` entry point; rank-8 marker intact at `:15789` (unit guard 4/4)

**Tests**: new golden 7/7 (×3 stable, G0 positive control green); all 5 D-1c slices + §0.6 precheck in one process = 56/56 (no fixture collision); §0.6 precheck 9/9 isolated; tsc clean.

**COMPLETION cross-check (final slice)**: all 8 A1–A8 uncaptured write sites now emit revisions on head (form ①, plugin ②, automation ③, approval ④, attachment ⑤). origin/main's attachment handler had 0 revision calls — genuine gap closed. Independent `UPDATE/INSERT meta_records` + `jsonb_set` grep found **no 9th uncaptured user-data write**; the only owner-flagged residual (field-undelete rehydration `:6430`/`:6522`) is already docketed to the OD-6 guard #4227. **Only remaining D-1c work = OD-6 #4227 (separate rung).**

**Findings**: P1 none · P2 none · **P3-1** — atomicity golden is connection-unwrap-survivable: it locks that the revision write is *inside* the txn block but not that it uses the txn-scoped `query` connection (swapping to a separate pooled connection stays green). Runtime correct by inspection; ⑤ is uniquely positioned to add the discriminating leg (inject into the post-revision in-txn `softDeleteAttachmentRowShared`) but reused the siblings' single-leg convention. Optional hardening, non-blocking. **NIT-1** — `createSeededSheet` template seed (`:5808`) writes no create revision (pre-existing, correctly exempt system scaffold; confirm exempt when #4227's marker sweep lands).

Full report: `/tmp/pr4249-review-claude-20260714.md`. Review-only — no merge/push/arm performed.

PR #4251, opened 2026-07-14
test(multitable): OD-6 revision-disposition guard — every meta_records write declares its disposition (D-1c final rung; scanner hardened)

zensgit:
## Summary

Implements the **OD-6 revision-disposition guard**, ratified in the design-lock (`docs/development/multitable-global-history-d1c-form-submit-edit-uncaptured-revision-design-lock-20260712.md` §0.5 row OD-6 / §7). This is the **last rung of W0/D-1c**, landing after all five write-site slices (①#4245, ②#4246, ③#4247, ④#4248, ⑤#4249). It forces every `meta_records` INSERT/UPDATE/DELETE site — anywhere in the runtime tree — to declare an explicit revision disposition, so a **new** uncaptured write can never regress the class of bug D-1c fixed.

This supersedes Draft #4227 (`claude/od-6-revision-disposition-guard`, head `b99c44fbf`), which the owner's adversarial re-review (2026-07-13) found NEEDS-WORK on two blockers. Both are fixed here (recycled structure, new scanner). #4227 stays open as a draft; I'll close it referencing this PR — not with a closing keyword (see PR discipline below), since it documents real review history worth keeping.

## Scanner hardening — the two proven blockers, fixed

**Blocker 1 — case-insensitive + multiline SQL scan.** The #4227 scanner tested `MUTATION_RE` (no `i` flag) against source **one line at a time**. The live-guard mutation proved: an unmarked UPPERCASE `UPDATE meta_records` reds the guard, but the **lowercase** twin passes 11/11 undetected, and a verb/table split across lines is invisible to a per-line test.

Fix (`tests/unit/lib/multitable-revision-disposition-scanner.ts`): the whole file is comment-stripped (state machine — preserves every string/template literal byte-for-byte and every newline's position, so a comment-only mention of the SQL text can never register as a mutation site), then scanned with **one** `matchAll` over the case-insensitive regex `/\b(?:INSERT\s+INTO\s+meta_records|UPDATE\s+meta_records|DELETE\s+FROM\s+meta_records)\b|(?:updateTable|deleteFrom|insertInto)\(\s*['"`]meta_records['"`]/gi`. `\s+` already matches `\n`, so a verb and `meta_records` on separate lines still match one statement.

**Defeat-then-fix proof** (`tests/unit/multitable-revision-disposition-scanner.test.ts`, in-memory fixtures, no disk mutation):
- uppercase unmarked multiline `UPDATE` → detected (positive control)
- **lowercase unmarked multiline `update`** (the exact #4227 evasion) → detected
- verb and `meta_records` split across separate lines (multiline `INSERT`) → detected
- comment-only mention of the SQL text → **not** flagged (no over-match)
- `meta_records_trash` → **not** matched (word-boundary check)

Then reproduced against the **real, tracked source tree** (not a fixture) — see "Bite proof" below.

**Blocker 2 — stable site IDs, not line numbers.** #4227 keyed its `PENDING_REVISION_SITES` allowlist by literal `(file, line)`, so any edit above a pending site silently drifted its line number. Fix: every site gets a `siteId` = `sha1(file :: enclosing-anchor :: statement-fingerprint :: occurrence-index)`. The anchor is the nearest enclosing named function / `const foo = (...) =>` / class method / `router.METHOD('path', ...)` scanning upward over the comment-stripped text (heuristic, not a full parser — documented as such; good enough because the anchor only changes when a contributor edits content genuinely between the site and its declaration, categorically rarer than "any edit anywhere above the file"). Proof: inserting 3 unrelated comment lines above a marked site shifts its `line` but leaves its `siteId` byte-identical; two byte-identical statements in the same function get different ids via the occurrence index; the same statement in a different function gets a different id.

## Guard mechanics

`tests/unit/multitable-revision-disposition-guard.guard.test.ts` walks the whole `packages/core-backend/src` tree (same file-list / exclusions as the rank-8 lock guard: `node_modules`, `dist`, `migrations`, `.d.ts`/`.test.ts`/`.spec.ts` excluded) and requires, within `MARKER_WINDOW = 3` lines above each site (same convention as the rank-8 lock guard), one of:

- `// revision-emitted:<why>` — a `recordRecordRevision(...)` call executes in the same transaction.
- `// revision-exempt:<why>` — a trusted derived/system/config-channel write with no user actor.
- `// revision-pending:<why>` — reserved **exclusively** for `KNOWN_REVISION_GAPS` entries (see below); a contributor cannot self-declare "pending" to dodge the guard — a hygiene test asserts both directions (every `revision-pending` site is in the allowlist, and every allowlist entry is currently `revision-pending`; a stale entry after a follow-up fix reds until deleted).

An additional per-file cross-check (mirroring the rank-8 guard's "a label cannot fake a guard" test) asserts every file with an `EMITTED` site actually contains a `recordRecordRevision(` call somewhere in it.

## The disposition matrix (36 literal `meta_records` INSERT/UPDATE/DELETE sites, `packages/core-backend/src`, non-migration)

**Count reconciliation vs. the design-lock's audit** (`§7a`, and the standalone consolidated audit it draws on): the audit table header says "22 must-write / 15 exempt" (=37), but its own enumerated EXEMPT table folds two rows into one line each (`univer-meta.ts:16261/:16275`, `automation-executor.ts:3412/:3422` each list TWO sites on one markdown row) — the real distinct count is **14 exempt**, not 15. `22 + 14 = 36`, matching this PR's literal `grep`/scanner count exactly. (#4227 found and reported the same typo independently.) Post owner-ruling reclassification of `univer-meta.ts:6522` (EXEMPT → MUST-WRITE, see below): **22 EMITTED + 13 EXEMPT + 1 PENDING = 36.**

### EMITTED (22) — compliant, real `recordRecordRevision(...)` call in the same transaction

**14 pre-existing (rank-8 / D-1 / D-2 eras):**
| Site | recordRecordRevision |
|---|---|
| `record-service.ts` `createRecord` (INSERT) | @706 |
| `record-service.ts` `deleteRecord` (DELETE) | @845 |
| `record-service.ts` restore (INSERT) | @1125 |
| `record-service.ts` `patchRecord` (UPDATE) | @1392 |
| `record-write-service.ts` bulk PATCH ×2 (UPDATE) | @998 |
| `records.ts` plugin delete D-2 flag-on (DELETE) | @650 |
| `records.ts` plugin delete D-1 flag-off (DELETE) | @771 |
| `automation-executor.ts` `executeDeleteRecord` (DELETE) | @2365 |
| `univer-meta.ts` lossy-retype revert (UPDATE) | @6455 |
| `univer-meta.ts` PIT resurrect (INSERT) | @10178 |
| `univer-meta.ts` PIT reset revert ×2 (UPDATE) | @10416 |
| `univer-meta.ts` PIT reset delete (DELETE) | @10498 |

**8 newly-fixed by the five D-1c slices (A1-A8):**
| Slice | Site | Source |
|---|---|---|
| ① form (#4245) | `univer-meta.ts` form-submit EDIT (A1, UPDATE) | `'public-form'` |
| ① form (#4245) | `univer-meta.ts` form-submit CREATE (A6, INSERT) | `'public-form'` |
| ② plugin (#4246) | `records.ts` `patchRecord` (A2, UPDATE) | `'plugin'` |
| ② plugin (#4246) | `records.ts` `createRecord` (A5, INSERT) | `'plugin'` |
| ③ automation (#4247) | `automation-executor.ts` `update_record` (A3, UPDATE) | `'automation'` |
| ③ automation (#4247) | `automation-executor.ts` `create_record` (A4, INSERT) | `'automation'` |
| ④ approval (#4248) | `automation-service.ts` `resultWriteback` (A7, UPDATE) | `'approval'` |
| ⑤ attachment (#4249) | `univer-meta.ts` attachment-delete cell-strip (A8, UPDATE) | `'attachment'` |

### EXEMPT (13) — derived/system/config-channel, no user actor

`approval-record-projection-service.ts` (INSERT, derived approval projection) · `univer-meta.ts` People-sync ×2 (INSERT+UPDATE) · `univer-meta.ts` `createSeededSheet` template scaffold (INSERT, hardcoded demo preset) · `formula-engine.ts` formula materialization (UPDATE, no version bump) · `univer-meta.ts` relation-aggregation same-record + fan-out (UPDATE ×2, no version bump) · `auto-number-service.ts` backfill (UPDATE, no version bump) · `univer-meta.ts` field-delete column strip (UPDATE, captured on config channel) · `univer-meta.ts` record lock/unlock ×2 (UPDATE, metadata-only) · `automation-executor.ts` automation lock/unlock ×2 (UPDATE, metadata-only).

## The field-undelete `univer-meta.ts` `recreateFieldFromConfig` finding (owner-flagged, per the brief)

The design-lock's OD-6 row is explicit: *"`univer-meta.ts:6521` field-undelete rehydration: owner ruling = SHOULD WRITE REVISION (audit flagged it debatable-EXEMPT, owner directs it MUST-WRITE)."* I verified the current state on `origin/main` (post all five slices): the site (now at line ~6537, inside `recreateFieldFromConfig`'s tombstone-rehydration branch) does `SET data = data || jsonb_build_object($3::text, t.value)` — real captured user cell values written back from tombstones — with **no `version` bump and no `recordRecordRevision` call**. It is the exact fingerprint of the bug class this whole guard exists to catch, and it is **still open**.

This is **not** one of D-1c's five slices (those are the A1-A8 bucket-A user-write-entry-point sweep; this is a field-op) — implementing the fix here would be unauthorized scope creep on a guard PR that is otherwise comment-only/zero-runtime-change, and would need its own `source` decision, real-DB golden, and adversarial review, exactly like each of the five landed slices got.

Per the brief's explicit instruction ("if it currently doesn't emit, that's a finding — report it, don't silently exempt it"), I did **not** mark it `revision-exempt` (would contradict the owner's ruling) and did **not** fabricate a `revision-emitted` marker (would be a lie, on the exact site three separate human audits already flagged). Instead: a third disposition, `revision-pending`, reserved exclusively for a `KNOWN_REVISION_GAPS` allowlist in the guard test, containing exactly this one site's stable `siteId`. The guard is GREEN because the gap is **declared, named, and pinned** — not silently passed. A hygiene test asserts nothing else can claim `pending` status and that this entry stays live only as long as the site is actually still pending.

**Reported finding, not fixed here: `univer-meta.ts` `recreateFieldFromConfig` field-undelete rehydration needs its own follow-up rung (own `source`, real-DB golden, adversarial review) to close this gap and let `KNOWN_REVISION_GAPS` go to zero.**

## Bite proof — the guard fires both ways

**(a) GREEN on current main** — all 36 sites declared (22 EMITTED / 13 EXEMPT / 1 tracked PENDING):
```
✓ EVERY meta_records INSERT/UPDATE/DELETE site carries a revision disposition
✓ every PENDING-labelled site is a KNOWN, TRACKED gap (both directions)
✓ every EMITTED-labelled FILE actually calls recordRecordRevision(
```

**(b) Planting an unmarked write reds it** — each variant proven against the REAL scanned source tree (`record-service.ts`), reverted after (via `git checkout --` on a committed checkpoint, never against uncommitted work):
- uppercase unmarked single-line `UPDATE meta_records` → **guard RED**, names the exact site
- **lowercase** unmarked single-line `update meta_records` (the #4227 evasion) → **guard RED**
- multiline, verb+table split across lines, lowercase → **guard RED**

**(c) Removing a real marker reds it** — stripped the `revision-emitted` comment from the real, landed form-submit EDIT site (`univer-meta.ts`, slice ①/A1) → **guard RED**, naming that exact site; reverted.

**Collateral regression found and fixed during this work**: adding the new marker at `automation-executor.ts`'s `executeCreateRecord` INSERT (A4) initially pushed the pre-existing `xbase-write-gated:` marker outside the **cross-base write guard**'s own 5-line window (`multitable-cross-base-write-guard.guard.test.ts`, a third structural guard scanning the same file), redding it. Fixed by appending the revision marker to the end of the existing comment's last physical line instead of adding a new line (no new line added = no window disturbed), which required loosening `classify()`'s regex to not require the marker to be the first token after `//` (documented in the scanner's docstring). Verified: full `pnpm exec vitest run` (468 files / 6434 tests) green, plus rank-8 lock guard and cross-base write guard both green alongside this one.

## Test plan

- [x] `pnpm exec vitest run tests/unit/multitable-revision-disposition-guard.guard.test.ts` — 5/5 passed
- [x] `pnpm exec vitest run tests/unit/multitable-revision-disposition-scanner.test.ts` — 20/20 passed (defeat-then-fix + stability proofs)
- [x] `pnpm exec vitest run tests/unit/multitable-record-lock-guard.guard.test.ts` — 4/4 passed (rank-8, not broken)
- [x] `pnpm exec vitest run tests/unit/multitable-cross-base-write-guard.guard.test.ts` — 5/5 passed (not broken, after the fix above)
- [x] `pnpm exec vitest run` (full core-backend unit suite) — 468 files / 6434 tests passed, 1523 pre-existing skips
- [x] Bite proof (b) and (c) above, executed against the real tracked source tree, reverted after each

No DB flag, no env gate — pure fs-scan unit test, collected automatically by the default vitest glob in the same `test (20.x)` step as the rank-8 lock guard (confirmed: not present in `vitest.config.ts`'s `exclude` list).

## This completes W0 D-1c

§0.6 `HISTORY_INCOMPLETE` (#4234, lands first) + five write-site slices (①-⑤, #4245/#4246/#4247/#4248/#4249) + this guard (lands last, per the design-lock's §10 routing) = the full ratified D-1c scope. One reported, tracked, non-silent residual: `univer-meta.ts` field-undelete rehydration needs its own follow-up rung (see finding above) — everything else in the ratified scope is closed.

**Draft — not requesting merge/arm.** Supersedes #4227 (draft, pre-slices, unfixed scanner) — I'll close that PR referencing this one, not with a closing keyword.

### DIFF
--- a/packages/core-backend/src/multitable/approval-record-projection-service.ts
+++ b/packages/core-backend/src/multitable/approval-record-projection-service.ts
@@ -219,6 +219,8 @@ export class ApprovalRecordProjectionService {
       await this.ensureSystemBase(client)
       await this.ensureFamilySheet(client, sheetId, instance.templateId, instance.templateName)
 
+      // revision-exempt: derived read-model projection into the system approval base, regenerable by
+      // sweep()/reconcile(); authoritative history lives in the approval domain (OD-6 owner ruling).
       await client.query(
         `INSERT INTO meta_records (id, sheet_id, data, version, created_by, modified_by)
          VALUES ($1, $2, $3::jsonb, 1, $4, $4)
--- a/packages/core-backend/src/multitable/auto-number-service.ts
+++ b/packages/core-backend/src/multitable/auto-number-service.ts
@@ -97,6 +97,7 @@ export async function backfillAutoNumberField(
   // SELECT ... FOR UPDATE is required.
   const overwrite = opts?.overwrite === true
   // lock-exempt: system auto-number backfill — system-derived sequence value, no user actor.
+  // revision-exempt: auto-number backfill, no version bump — system-derived sequence value under advisory lock.
   const updated = await query(
     `UPDATE meta_records mr
      SET data = jsonb_set(
--- a/packages/core-backend/src/multitable/automation-executor.ts
+++ b/packages/core-backend/src/multitable/automation-executor.ts
@@ -2226,6 +2226,7 @@ export class AutomationExecutor {
         // xbase-write-gated: routes through evaluateCrossBaseWrite (gate computed above) — a cross-base
         // update is rejected before this UPDATE unless claim==truth + trigger-actor base-write.
         // lock-guarded: automation update_record (B1) — ensureRecordNotLocked enforced just above.
+        // revision-emitted: D-1c slice ③ (A3) — recordRecordRevision(action:'update') @2258.
         const updateRes = await query(
           `UPDATE meta_records
            SET data = COALESCE(data, '{}'::jsonb) || $1::jsonb,
@@ -2455,6 +2456,7 @@ export class AutomationExecutor {
         // xbase-write-gated: routes through evaluateCrossBaseWrite (gate computed above) — a cross-base
         // delete is rejected before this DELETE unless claim==truth + trigger-actor base-write.
         // lock-guarded: automation delete_record (C2a) — ensureRecordNotLocked enforced just above.
+        // revision-emitted: automation delete_record, D-1 — recordRecordRevision(action:'delete') @2365.
         await query(
           'DELETE FROM meta_records WHERE id = $1 AND sheet_id = $2',
           [effectiveRecordId, effectiveSheetId],
@@ -2530,7 +2532,7 @@ export class AutomationExecutor {
       // neither of which added one).
       // xbase-write-gated: routes through evaluateCrossBaseWrite (gate computed above) — a cross-base
       // create is rejected before this INSERT unless claim==truth + trigger-actor base-write. This is
-      // the §1.3 create-to-another-base vector the gate closes.
+      // the §1.3 create-to-another-base vector the gate closes. revision-emitted: D-1c slice ③ (A4) — recordRecordRevision(action:'create') below, same txn.
       await this.withTransaction(async (query) => {
         await query(
           `INSERT INTO meta_records (id, sheet_id, data, version) VALUES ($1, $2, $3::jsonb, 1)`,
@@ -3485,6 +3487,7 @@ export class AutomationExecutor {
         // xbase-write-gated: routes through evaluateCrossBaseWrite (gate above) — same-base uses the gate's
         // fast-path (byte-identical to pre-C2); a cross-base lock is rejected unless claim==truth + base-write.
         // lock-mgmt: LOCK action — sets the lock columns themselves (not a data edit of a locked row).
+        // revision-exempt: lock/unlock metadata-only — no `data` column touched, not a user-content edit.
         await this.deps.queryFn(
           `UPDATE meta_records
            SET locked = true, locked_by = $1, locked_at = NOW(), version = version + 1, updated_at = NOW()
@@ -3495,6 +3498,7 @@ export class AutomationExecutor {
         // xbase-write-gated: routes through evaluateCrossBaseWrite (gate above) — same-base fast-path keeps
         // this byte-identical to pre-C2; a cross-base unlock is rejected unless claim==truth + base-write.
         // lock-mgmt: UNLOCK action — clears the lock columns (decision f: automation may unlock).
+        // revision-exempt: lock/unlock metadata-only — no `data` column touched, not a user-content edit.
         await this.deps.queryFn(
           `UPDATE meta_records
            SET locked = false, locked_by = NULL, locked_at = NULL, version = version + 1, updated_at = NOW()
--- a/packages/core-backend/src/multitable/automation-service.ts
+++ b/packages/core-backend/src/multitable/automation-service.ts
@@ -2860,8 +2860,8 @@ export class AutomationService {
         return false // record gone — the caller's missing-record path already handled it
       }
       ensureRecordNotLocked(opts.lockActorId, lockRow, () => new Error(opts.lockedMessage))
-      // lock-guarded: resultWriteback patch — ensureRecordNotLocked (opts.lockActorId = the trigger actor
-      // for a cross-base target) is enforced immediately above; a locked target throws → rolls back (§ note).
+      // lock-guarded: resultWriteback patch — ensureRecordNotLocked (opts.lockActorId, the trigger actor for a cross-base target) enforced immediately above; a locked target throws and rolls back.
+      // revision-emitted: D-1c slice ④ (A7 approval resultWriteback) — recordRecordRevision below, same txn.
       const updateRes = await query(
         `UPDATE meta_records
          SET data = COALESCE(data, '{}'::jsonb) || $1::jsonb, version = version + 1, updated_at = NOW()
--- a/packages/core-backend/src/multitable/formula-engine.ts
+++ b/packages/core-backend/src/multitable/formula-engine.ts
@@ -341,6 +341,7 @@ export class MultitableFormulaEngine {
       // write to other fields between this SELECT and UPDATE is not clobbered. No
       // version bump: formula values are derived, not an authoritative user edit.
       // lock-exempt: system formula materialization — derived value, no user actor (lock = read-only to USERS).
+      // revision-exempt: derived formula materialization, no version bump — pure fn of inputs, recompute-on-read.
       await query(
         `UPDATE meta_records SET data = data || $1::jsonb, updated_at = now() WHERE id = $2 AND sheet_id = $3`,
         [JSON.stringify(updates), recordId, sheetId],
--- a/packages/core-backend/src/multitable/record-service.ts
+++ b/packages/core-backend/src/multitable/record-service.ts
@@ -675,6 +675,7 @@ export class RecordService {
         type: field.type,
         property: field.property,
       }))))
+      // revision-emitted: REST createRecord — recordRecordRevision(action:'create') below, same txn.
       const inserted = await query(
         `INSERT INTO meta_records (id, sheet_id, data, version, created_by, modified_by)
          VALUES ($1, $2, $3::jsonb, 1, $4, $4)
@@ -889,6 +890,7 @@ export class RecordService {
       }
 
       // lock-guarded: single DELETE — ensureRecordNotLocked enforced above (rejects before this txn).
+      // revision-emitted: REST deleteRecord — recordRecordRevision(action:'delete') below, same txn.
       await query('DELETE FROM meta_records WHERE id = $1', [recordId])
 
       // OAPI-2b §6: committed token-delete audit INSIDE the (soft-)delete txn (fail-closed). No-op for session.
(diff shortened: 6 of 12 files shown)


zensgit:
## Adversarial review — VERDICT: MERGE_CLEAN (scoped to head \`585eb6784\`)

Fresh disposable worktree off PR head, real-tree mutation battery, independent multiline/case-insensitive enumeration, comment-only diff proof, tsc. **No P1, no P2.**

**The three deciding questions:**
1. **Scanner un-defeatable (the #4227 blockers) — PASS.** Planted-in-real-tree probes: uppercase / lowercase / multiline-split / mixed-case / extra-whitespace / DELETE / INSERT / Kysely / alias / CTE all **detected**; `meta_records_trash` and comment-only mentions correctly **not** matched; the comment-stripper survives a string containing `*/`. Stable site IDs confirmed.
2. **`revision-pending` / `KNOWN_REVISION_GAPS` is NOT a backdoor (the crux) — PASS both directions.** A NEW self-declared `revision-pending` site (not allowlisted) → hygiene test **REDs**. Flipping the pending site to `revision-emitted` → stale-allowlist hygiene test **REDs** (`456ec0c986a1b1af`), forcing removal when the follow-up rung lands. Exactly one allowlist entry.
3. **Matrix honest & complete — PASS.** Independent enumeration = **36 sites** (22 EMITTED / 13 EXEMPT / 1 PENDING), no missed/non-literal write path. EMITTED spot-checks (4/4) emit real \`recordRecordRevision\` same-txn; EXEMPT spot-checks legitimate (no user-data write exempted away); \`:6522\`/\`:6537\` field-undelete genuinely doesn't emit and is honestly surfaced as \`revision-pending\` (its own tracked follow-up rung), not buried.

**Also-checks:** runs in the **required** \`test (20.x)\` lane (plugin-tests.yml, no path filter) — not theater; **every \`src/**\` change is comment-only** (zero runtime; only \`classify()\` in the scanner lib is logic); collateral rank-8 + cross-base guards **green** (9/9); tsc clean; branch current with main.

**Findings (all P3/NIT, non-blocking):**
- **P3-1** (brief pre-registered P2): schema-qualified \`public.meta_records\`, quoted \`"meta_records"\`, and interpolated \`${T}\` forms **evade** the scanner. Downgraded to P3 — **zero reachable producer** (these forms appear nowhere in \`src\` for any table) and the accepted **rank-8 guard shares the identical documented hole** ("the only forms in use today"). One-line fix closes both: table token → \`(?:"?\w+"?\s*\.\s*)?"?meta_records"?\`. Owner sets severity.
- **P3-2**: \`classify()\` reads the raw window, so a string/comment literal containing \`revision-emitted:\` can false-mark a site — caught by the per-file cross-check in new files; maskable only in already-emitting files (acknowledged per-file limitation, shared with rank-8).
- **NIT**: PR body claims "can never regress" — the P3-1 evasions falsify "never"; soften to "via any form currently used in this codebase" (matches rank-8's honest hedge). Plus a benign log string containing "update meta_records" trips the guard (fail-closed); and the two lock/unlock sites' anchor is mislabeled (readability only; the PENDING site's anchor is correct so the allowlist stays stable).

**Completion:** §0.6 (#4234) + 5 slices (#4245–#4249, already on main) + this guard = the full ratified D-1c scope, with \`:6522\` field-undelete as the one tracked open gap. Confirmed.

Full report: \`/tmp/pr4251-review-claude-20260714.md\`

🤖 review by Claude Opus (adversarial-reviewer)

PR #4258, opened 2026-07-14
test(multitable): harden revision-disposition + lock-disposition scanners against schema-qualified/quoted meta_records evasions (gate P3-1)

zensgit:
## Summary

Closes a P3 scanner-evasion hole shared by two independent structural guards: the OD-6 revision-disposition scanner (`tests/unit/lib/multitable-revision-disposition-scanner.ts`, landed #4251) and the pre-existing rank-8 lock-disposition guard (`tests/unit/multitable-record-lock-guard.guard.test.ts`). Both guards' `MUTATION_RE` matched only the bare, unquoted, unqualified spelling `meta_records` — a **schema-qualified** (`public.meta_records`) or **double-quoted** SQL-identifier form (`"meta_records"`, `public."meta_records"`) of the exact same table silently evaded both. Zero such producers exist in `src` today (grep-verified) — P3, fail-closed hardening, not a live bypass.

**Not merging/arming this PR.**

## The regex change

Both files gain an identical `TABLE_REF` sub-pattern (duplicated, not shared via import — the two guards are independently owned, same discipline as `MARKER_WINDOW` mirroring already used between them):

```js
const TABLE_REF = String.raw`(?:"?\w+"?\.)?"?meta_records(?!\w)"?`
```

This matches: bare `meta_records`, schema-qualified `public.meta_records` / `"public".meta_records`, double-quoted `"meta_records"`, and both combined (`public."meta_records"`, `"public"."meta_records"`).

**OD-6 scanner** (`tests/unit/lib/multitable-revision-disposition-scanner.ts`) — `MUTATION_RE` rebuilt via `new RegExp(...)` composing `TABLE_REF` into the same verb alternation, same `gi` flags, same `\s+`-tolerant multiline behavior — unchanged apart from the table portion:

```js
export const MUTATION_RE = new RegExp(
  String.raw`\b(?:INSERT\s+INTO\s+${TABLE_REF}|UPDATE\s+${TABLE_REF}|DELETE\s+FROM\s+${TABLE_REF})` +
    String.raw`|(?:updateTable|deleteFrom|insertInto)\(\s*['"\`]${TABLE_REF}['"\`]`,
  'gi',
)
```

**rank-8 guard** (`tests/unit/multitable-record-lock-guard.guard.test.ts`) — same `TABLE_REF`, but composed into the guard's own case-sensitive, single-space (not `\s+`), non-global pattern, unchanged apart from the table portion:

```js
const MUTATION_RE = new RegExp(
  String.raw`\b(?:UPDATE ${TABLE_REF}|DELETE FROM ${TABLE_REF})` +
    String.raw`|(?:updateTable|deleteFrom)\(\s*['"\`]${TABLE_REF}['"\`]`,
)
```

**Why `(?!\w)` and not a trailing `\b`**: the original pattern closed with `\b` right after the table name. `\b` after an *optional* closing `"` can be FALSE — a boundary needs one side to be a word character, and a closing quote followed by e.g. a space has non-word characters on both sides — which would silently un-match every quoted form. `(?!\w)` binds directly to the end of the literal `meta_records` token instead (before any optional trailing quote is consumed), so `meta_records_trash` / `meta_record_revisions` / `meta_records_xyz` (different tables) still correctly do NOT match, quoted or not.

**Disposition logic and allowlists in both guards are untouched** — only `TABLE_REF`/`MUTATION_RE`. Confirmed: both guards' real-tree site counts are unchanged post-fix (OD-6 still finds all 36 declared sites; rank-8's existing sites are still all classified) — see proof below.

## Documented residual (not fixed — by design, per explicit scope narrowing during this PR)

A table name that is **entirely** a runtime variable/template interpolation with no literal `meta_records` substring anywhere in the statement text — e.g. `` `UPDATE ${TABLE} SET ...` `` where `TABLE` is assigned `'meta_records'` somewhere else in the file — cannot be resolved by a lexical regex scanner; that requires real type-aware static analysis, out of scope for an fs-scan guard. Zero such producers exist in `src` today (grep-verified: every `meta_records` mutation in this codebase spells the table name as a literal within the statement itself). Deliberately **not** papered over with a heuristic that flags any `${...}`-templated table position — that would false-positive on every unrelated parameterized-table helper in the codebase (none of which touch `meta_records`), trading a currently-unreachable gap for a guaranteed false-positive tax on unrelated tables. This residual is documented in both files' docstrings, cross-referencing each other.

### 1. Baseline green on unmodified main (pre-fix)

`pnpm exec vitest run tests/unit/multitable-revision-disposition-guard.guard.test.ts tests/unit/multitable-revision-disposition-scanner.test.ts tests/unit/multitable-record-lock-guard.guard.test.ts tests/unit/multitable-cross-base-write-guard.guard.test.ts` on `492c64d30` → **34/34 passed**.

### 2. Post-fix: same suite still green, plus 10 new self-tests

Same command on this branch → **44/44 passed** (34 pre-existing + 10 new `gate P3-1` cases in `multitable-revision-disposition-scanner.test.ts`: 5 DEFEAT PROOFs for the new forms including the Kysely builder schema-qualified form, 1 marker-still-classifies-correctly check, 4 sibling-table-still-excluded checks). OD-6's 36-site count and rank-8's existing site count are unchanged (no smoke-bound test tripped).

### 3. Real-tree evasion-then-detect proof (checkpoint-commit → plant → red → revert)

Checkpointed the fix as commit `c58f11f36`, then for each form: appended an unmarked, real statement to `src/multitable/record-service.ts` (a file both guards scan), ran both guards, reverted via `git checkout -- src/multitable/record-service.ts`.

| Planted form | rank-8 lock guard | OD-6 revision-disposition guard |
|---|---|---|
| `UPDATE public.meta_records SET ...` (schema-qualified) | RED — named the exact site | RED — named the exact site |
| `UPDATE "meta_records" SET ...` (double-quoted) | RED | RED |
| `DELETE FROM public."meta_records" WHERE ...` (both) | RED | RED |

Each reverted cleanly (`git status --short` empty after each cycle).

### 4. Sibling-table false-positive check

Planted `UPDATE public.meta_records_trash SET ...` (schema-qualified sibling table) → **both guards stayed GREEN** (4/4 rank-8, 5/5 OD-6) — the widened regex does not over-match a different table. Reverted cleanly.

### 5. Mutation self-check (neuter the fix → new tests must go RED → restore)

Temporarily restored both files' pre-fix `MUTATION_RE` from `492c64d30` (test files with the new assertions kept as-is), replanted `UPDATE public.meta_records SET ...` in `record-service.ts`:

- Both real-tree guards **stayed GREEN** with the old regex — confirms the P3 finding is real (old code was silently vulnerable).
- The new OD-6 self-test file, run against the neutered lib, flipped **exactly the 6 new P3-1-specific tests to RED** (5 DEFEAT PROOFs + the marker-classification check) while the other 24 pre-existing tests stayed green — confirms the new tests are load-bearing, not vacuous.

Restored the fix (`git checkout HEAD -- <2 files>`) and reverted the plant (`git checkout -- src/multitable/record-service.ts`). Confirmed clean tree + suite green again (44/44) after restore.

### 6. tsc + full unit suite

- `pnpm exec tsc --noEmit` — clean, no errors.
- `pnpm exec vitest run` (full core-backend unit suite) — **468 files / 6444 tests passed, 0 failed** (1523 pre-existing skips), no regressions anywhere else in the tree.

## Test plan

- [x] `pnpm exec vitest run tests/unit/multitable-revision-disposition-scanner.test.ts` — 30/30 passed (20 pre-existing + 10 new)
- [x] `pnpm exec vitest run tests/unit/multitable-revision-disposition-guard.guard.test.ts` — 5/5 passed (36 sites still all declared)
- [x] `pnpm exec vitest run tests/unit/multitable-record-lock-guard.guard.test.ts` — 4/4 passed (existing sites still all declared)
- [x] `pnpm exec vitest run tests/unit/multitable-cross-base-write-guard.guard.test.ts` — 5/5 passed (unaffected, sanity check)
- [x] `pnpm exec vitest run` (full suite) — 468 files / 6444 tests passed
- [x] Evasion-then-detect proof (3 forms × 2 guards) on the real tracked source tree, reverted after
- [x] Sibling-table false-positive check (`meta_records_trash`, schema-qualified), reverted after
- [x] Mutation self-check (neuter → 6 new tests RED → restore), reverted after

### DIFF
--- a/packages/core-backend/tests/unit/lib/multitable-revision-disposition-scanner.ts
+++ b/packages/core-backend/tests/unit/lib/multitable-revision-disposition-scanner.ts
@@ -46,12 +46,47 @@ export interface Site {
 export const MARKER_WINDOW = 3
 
 /**
- * Every form a `meta_records` row-mutation can take in this codebase: raw SQL (case-insensitive, verb
- * and table name may be split across lines by `\s+`) and the Kysely query-builder forms, added
- * pre-emptively exactly as the rank-8 guard does for its own MUTATION_RE.
+ * Table-reference sub-pattern for `meta_records` (gate P3-1). The original `MUTATION_RE` (as landed in
+ * #4251) matched only the bare, unquoted, unqualified spelling `meta_records` — a SCHEMA-QUALIFIED
+ * (`public.meta_records`, `"public".meta_records`) or DOUBLE-QUOTED SQL-identifier form
+ * (`"meta_records"`, `public."meta_records"`, `"public"."meta_records"`) of the exact same table
+ * silently evaded it. Zero such producers exist in `src` today (grep-verified) — this is fail-closed
+ * hardening against a P3 finding, not a fix for a live bypass.
+ *
+ * `(?!\w)` immediately after the literal `meta_records` (NOT a trailing `\b` on the outer group) is
+ * deliberate: `\b` right after an optional closing `"` can be FALSE — `\b` requires one side of the
+ * boundary to be a word character, and a closing quote followed by e.g. a space has NON-word characters
+ * on both sides — which would silently un-match every quoted form. `(?!\w)` binds directly to the end of
+ * the `meta_records` token itself, so `meta_records_trash` / `meta_record_revisions` / `meta_records_xyz`
+ * (different tables) still correctly do NOT match, quoted or not.
+ *
+ * DOCUMENTED RESIDUAL (accepted, not fixed): a table name that is ENTIRELY a runtime variable/template
+ * interpolation with no literal `meta_records` substring anywhere in the statement text itself — e.g.
+ * `` `UPDATE ${TABLE} SET ...` `` where `TABLE` is assigned `'meta_records'` somewhere else in the file —
+ * cannot be resolved by a lexical regex scanner; that requires real type-aware static analysis, which is
+ * out of scope for an fs-scan guard. Zero such producers exist in `src` today (grep-verified: every
+ * `meta_records` mutation in this codebase spells the table name as a literal within the statement
+ * itself). Deliberately NOT papered over with a heuristic that flags "any INSERT/UPDATE/DELETE whose
+ * table position is a bare `${...}` expression" — that would false-positive on every unrelated
+ * parameterized-table helper in the codebase (none of which touch `meta_records`), trading a currently
+ * unreachable gap for a guaranteed false-positive tax on unrelated tables. If a `meta_records` write ever
+ * loses its literal spelling to a fully-variable table name, it evades this guard AND the rank-8
+ * lock-disposition guard (`multitable-record-lock-guard.guard.test.ts`, same documented residual) —
+ * flagged here as a known, accepted gap.
  */
-export const MUTATION_RE =
-  /\b(?:INSERT\s+INTO\s+meta_records|UPDATE\s+meta_records|DELETE\s+FROM\s+meta_records)\b|(?:updateTable|deleteFrom|insertInto)\(\s*['"`]meta_records['"`]/gi
+const TABLE_REF = String.raw`(?:"?\w+"?\.)?"?meta_records(?!\w)"?`
+
+/**
+ * Every form a `meta_records` row-mutation can take in this codebase: raw SQL (case-insensitive, verb
+ * and table name may be split across lines by `\s+`, optionally schema-qualified and/or double-quoted —
+ * see `TABLE_REF` above) and the Kysely query-builder forms, added pre-emptively exactly as the rank-8
+ * guard does for its own MUTATION_RE.
+ */
+export const MUTATION_RE = new RegExp(
+  String.raw`\b(?:INSERT\s+INTO\s+${TABLE_REF}|UPDATE\s+${TABLE_REF}|DELETE\s+FROM\s+${TABLE_REF})` +
+    String.raw`|(?:updateTable|deleteFrom|insertInto)\(\s*['"\`]${TABLE_REF}['"\`]`,
+  'gi',
+)
 
 /**
  * Replace every `//...` and `/* ... *\/` comment with equal-length whitespace (newlines preserved, so
--- a/packages/core-backend/tests/unit/multitable-record-lock-guard.guard.test.ts
+++ b/packages/core-backend/tests/unit/multitable-record-lock-guard.guard.test.ts
@@ -207,4 +207,54 @@ describe('rank-8 record-lock mutation-path guard — durable structural guard', 
     // the rule itself is `locked && !canEditWhileLocked` — assert it routes through canEditWhileLocked
     expect(lockSrc).toMatch(/ensureRecordNotLocked[\s\S]{0,400}canEditWhileLocked\(/)
   })
+
+  test('gate P3-1: the two hand-synced TABLE_REF copies (this file + the OD-6 scanner) stay byte-identical', () => {
+    // TABLE_REF is duplicated (not shared via import — the two guards are independently owned) between
+    // this file and `lib/multitable-revision-disposition-scanner.ts`. A future edit to one copy that
+    // forgets the other would silently re-open the gate-P3-1 gap in whichever guard was NOT updated.
+    // This is a cheap textual pin, not a semantic one — it only proves the two source lines are
+    // byte-identical, not that either is correct (the fixture tests above/below do that).
+    const extractTableRef = (src: string): string => {
+      const m = src.match(/const TABLE_REF = String\.raw`([^`]*)`/)
+      if (!m) throw new Error('TABLE_REF definition not found — did its declaration shape change?')
+      return m[1]
+    }
+    const selfSrc = readFileSync(join(__dirname, 'multitable-record-lock-guard.guard.test.ts'), 'utf8')
+    const scannerSrc = readFileSync(join(__dirname, 'lib/multitable-revision-disposition-scanner.ts'), 'utf8')
+    expect(extractTableRef(selfSrc)).toBe(extractTableRef(scannerSrc))
+  })
+})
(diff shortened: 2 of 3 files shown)


zensgit:
**Adversarial gate — VERDICT: MERGE_CLEAN** (scoped to head `c58f11f36`)

Test-only P3-1 hardening. No P1/P2. The single flagged risk — a widened regex over-matching a sibling table — is **empirically disproven in both guards**.

**Verified (fresh worktree off the PR ref):**
- Scope = exactly 3 test files, zero `src`/disposition/allowlist change.
- Regex correct **both directions** (verbatim reconstruction): all schema/quoted forms caught (OD-6 case-insensitive incl. uppercase; rank-8 case-sensitive); **all** siblings rejected by both guards — `meta_records_trash`, `public.meta_records_trash`, `meta_records_xyz`, `meta_record_revisions`, `meta_records_backup`, `archived_meta_records`, and the `(?!\w)` boundary cases `meta_recordsxyz` / `meta_records2`. Column-ref `"meta_records"."col"` still matches the table; `SELECT ... FOR UPDATE` correctly not matched.
- Green on real `src`: OD-6 enumerates exactly **36 sites** (13 EXEMPT / 22 EMITTED / 1 PENDING, 0 unclassified) — widening produced an identical site set. rank-8 guard green.
- Documented residual (fully-variable `${table}`) is honest AND real: every `meta_records` mutation is literal (42 lines); all `${table}` sinks target other tables (external materialization DBs, `meta_record_revisions`/`meta_config_revisions`, `meta_fields`/`meta_views`/`meta_sheets`). Not a swept gap.
- `tsc --noEmit` clean; full unit suite **5352/5352 pass**.

**P3 (non-blocking) — rank-8 widening is neuterable-green.** `multitable-record-lock-guard.guard.test.ts` got the identical `TABLE_REF` but zero new test cases; isolated mutation (drop only the schema/quote widening, keep boundary) leaves all 4 rank-8 tests GREEN, whereas the same neuter on the OD-6 lib reds 6 tests. Graded P3 (not P2) because the widened branches are unreachable on current code (zero producers), the byte-identical `TABLE_REF` is fully gated in the OD-6 twin, and rank-8's assembled regex was independently verified correct. **Fix (also closes 2 NITs):** mirror ~3 cases into the rank-8 file — one schema-qualified catch, one schema-qualified sibling no-catch, plus `meta_recordsxyz`.

Full findings + run/mutation log: `/tmp/pr4258-review-claude-20260714.md`. Review only — no merge/push/arm.

zensgit:
## Follow-up: self-pin rank-8's widening (gate feedback)

The gate flagged a P3 on this PR: the rank-8 lock-disposition guard's `TABLE_REF` widening (`multitable-record-lock-guard.guard.test.ts`) had zero dedicated test coverage in its own file — isolated-neutering just that widening left all 4 pre-existing rank-8 tests green, because the widening was only exercised transitively via the OD-6 twin's fixture tests. Fixed:

1. **Mirrored the gate-P3-1 fixture cases into rank-8's own test file** (new `describe` block, using rank-8's own `enumerateSites`): `DEFEAT PROOF: schema-qualified unmarked UPDATE (public.meta_records)`, `DEFEAT PROOF: double-quoted unmarked UPDATE ("meta_records")`, plus 3 sibling-table negatives (`meta_records_trash` schema-qualified, `meta_records2` digit-suffix, `meta_recordsxyz` letter-suffix — no underscore separator).
2. **Added the two boundary sibling cases to the OD-6 scanner's own self-tests too** (NIT): `meta_records2` and `meta_recordsxyz` (no-underscore forms) — the docstring already claimed these were excluded but they were untested.
3. **Divergence guard** (small NIT, not skipped — trivial to add): a new rank-8 test reads both files' source and asserts the two hand-synced `TABLE_REF` string literals stay byte-identical, so a future edit to one copy that forgets the other trips red immediately instead of silently reopening the gap in whichever guard wasn't updated.

### Isolated-neuter proof (the exact gap the gate found, now closed)

Checkpointed the fix as `ba96f3d4a`, then temporarily replaced **only** rank-8's `TABLE_REF`/`MUTATION_RE` with the literal pre-fix (`492c64d30`) regex — OD-6's scanner untouched — and ran rank-8's own test file:

```
FAIL  ... gate P3-1: the two hand-synced TABLE_REF copies (this file + the OD-6 scanner) stay byte-identical
FAIL  ... DEFEAT PROOF: schema-qualified unmarked UPDATE (public.meta_records) is now detected
FAIL  ... DEFEAT PROOF: double-quoted unmarked UPDATE ("meta_records") is now detected
 Test Files  1 failed (1)
      Tests  3 failed | 7 passed (10)
```

Exactly the 3 tests tied to the widening went RED (2 defeat-proofs + the divergence guard); the 4 pre-existing rank-8 tests and the 3 sibling-negative tests stayed green (the neutered pattern's own `\b` boundary still correctly excludes sibling tables). This is the reverse of the gate's finding: before this follow-up, isolated-neutering left rank-8 fully green; now it reds by construction. Restored via `git checkout HEAD -- multitable-record-lock-guard.guard.test.ts` and reconfirmed clean tree + 52/52 green on the targeted suite.

### Full verification after restore

- Targeted 4-file guard suite: **52/52 passed** (was 44; +6 in rank-8's own file [2 defeat-proof + 3 sibling + 1 divergence-guard], +2 in OD-6's scanner self-tests).
- `pnpm exec tsc --noEmit` — clean.
- Full core-backend unit suite: **6451/6452 tests passed** (1 pre-existing, unrelated flake in `tests/unit/field-validation-wiring.test.ts` — `ECONNRESET`/"socket hang up", a file untouched by this diff; reran in isolation and it passed 5/5 clean, confirming transient flake not a regression).

Still a draft — not merging/arming.

PR #4263, opened 2026-07-14
docs(stock-prep): T3a design-lock — ERP source-run server-side auto-persist (PROPOSED)

zensgit:
## T3a design-lock — decisions SETTLED (RATIFY-ready, rev 2026-07-15)

Owner directed to write the settled decisions concretely and ratify. This rev turns the open decisions into settled ones (still docs-only, zero runtime):

- **OD-1**: default-OFF staged flag `MULTITABLE_STOCK_PREP_ERP_AUTOPERSIST_ENABLED`. OFF ⇒ read-only, byte-for-byte today's behavior ⇒ **safe to merge during the entity-machine wait**.
- **OD-2 (security)**: tenant is **flag-conditional** — flag OFF keeps `resolveTenantId` (GHSA step-2 read-steering stays deferred, untouched); flag ON derives read+write tenant from `resolveAuthUserTenantId` **AND rejects an explicit request tenant/projectId FAIL-CLOSED before any I/O** (owner refinement — reject, not silently ignore; same discipline as the baseId guard).
- **OD-4**: public evidence is a **precise composite** (source-run evidence + T2 persist evidence), values-free both directions.
- **OD-5**: idempotent upsert + an explicit **"staged, upsert-only — NOT authoritative full sync"** caveat (no delete of stale rows, no reconcile; describe as "落缓存(upsert)", never "全量同步").
- **§7**: **T3b goes through its OWN design gate** — not folded into this T3a lock.

**Status: RATIFY-ready — ratify remains the owner's gate; this rev does not self-ratify.**

The WIP implementation (`61caf7ff0`, branch `claude/stock-prep-t3a-erp-autopersist`) already wires the core per this lock (flag-conditional tenant + auto-persist; flag-OFF proven inert). The flag-ON security verification (steering reject red-first + real-DB smoke + mutation) is the next step. Docs only, not armed.

### DIFF
--- a/docs/development/stock-preparation-t3a-erp-source-autopersist-design-lock-20260714.md
+++ b/docs/development/stock-preparation-t3a-erp-source-autopersist-design-lock-20260714.md
@@ -37,7 +37,7 @@ source-run 路由(与 http-routes.cjs WIP bb9724494 一致):
 ## 3. 已决策（RATIFY-ready，owner 2026-07-15 定；实现处 `bb9724494` WIP 已按此接线）
 
 ### OD-1 自动落库 vs 门控 — ✅ DECIDED：默认关闭 flag，分级启用
-- **裁决 A**:新增默认-OFF flag（规范化字面 `true`,trim+小写）。OFF ⇒ 今日行为逐字节不变(只读,不落库);ON ⇒ 自动落缓存。符合本线 staged-optin 纪律 + O-2 阶梯。**默认 OFF ⇒ 即便早合也零运行时影响,可在实体机等待期安全合并。**
+- **裁决 A**:新增默认-OFF flag（规范化字面 `true`,trim+小写）。OFF ⇒ 今日行为逐字节不变(只读,不落库);ON ⇒ 自动落缓存。符合本线 staged-optin 纪律。**默认 OFF ⇒ 即便早合也零运行时影响,可在实体机等待期安全合并。**
 - 实现处 flag 名 = `MULTITABLE_STOCK_PREP_ERP_AUTOPERSIST_ENABLED`（WIP 已用,唯一名,无备选）。**台账**:本线**无**专门 flag manifest 文件(不同于 Global History 的 `scripts/ops/multitable-global-history-flag-status.mjs`);该 flag 记于本设计锁 + 实现 PR + 部署 runbook,生产默认 OFF(env 未设即 OFF)。
 
 ### OD-2 租户派生（**安全关键**——T3a 让读变得有后果）— ✅ DECIDED
--- a/docs/development/stock-preparation-t3a-erp-source-autopersist-design-lock-20260714.md
+++ b/docs/development/stock-preparation-t3a-erp-source-autopersist-design-lock-20260714.md
@@ -1,6 +1,6 @@
 # T3a 设计锁 — ERP approved source-run 服务端直落 `erp_material_master`（决策已定，待 RATIFY, rev 2026-07-15）
 
-状态：**决策已写实（RATIFY-ready）** — OD-1..OD-5 全部裁决落定（§3），owner 2026-07-15 授权「推进 T3a/T3b」。实现处 `61caf7ff0`（分支 `claude/stock-prep-t3a-erp-autopersist`）**WIP 已按本锁核心接线**（flag 条件 tenant + auto-persist，flag OFF 证实 inert）；**flag ON 的安全验证(steering 拒绝 red-first + 真库 smoke + mutation)待补**（§5）。模型：实现=Opus 主循环（权限+真实写+mutation 证明，owner 政策）。**ratify 是 owner 的门**——本 rev 只把决策写实,不自我 ratify。
+状态：**决策已写实（RATIFY-ready）** — OD-1..OD-5 全部裁决落定（§3），owner 2026-07-15 授权「推进 T3a/T3b」。实现处 `bb9724494`（分支 `claude/stock-prep-t3a-erp-autopersist`）**WIP 已按本锁核心接线**（flag 条件 tenant + auto-persist，flag OFF 证实 inert）；**flag ON 的安全验证(steering 拒绝 red-first + 真库 smoke + mutation)待补**（§5）。模型：实现=Opus 主循环（权限+真实写+mutation 证明，owner 政策）。**ratify 是 owner 的门**——本 rev 只把决策写实,不自我 ratify。
 
 ## 1. 问题（owner P1 功能断点，已核实）
 
@@ -14,10 +14,11 @@
 在 **同一请求内**,ERP source-run 成功后,把服务端已有的 `intake` 规范化行**直接**喂给 T2 的 `persistStockPreparationErpMaterialSync`——**行永不过 HTTP**,公开面仍 values-free。复用两个已合入且已硬化的服务端件(source-run intake + T2 persist),是**组合接线**,非新能力面。**flag OFF 时响应逐字节不变(不加任何字段);flag ON 时才落库并加 `autoPersist` + 覆盖 projector 的 dry_run/未写标记**(权威契约见 OD-4)。
 
 ```
-source-run 路由(与 http-routes.cjs WIP 922fd2f76 一致):
-  admin gate → normalize body
+source-run 路由(与 http-routes.cjs WIP bb9724494 一致):
+  admin gate
   autoPersistEnabled = flag                                   (OD-1，默认 OFF)
-  若 ON: assertStockPreparationErpAutoPersistNoSteering(req)   // 请求含 tenantId/projectId → 400,在任何 I/O 前 (OD-2)
+  若 ON: assertStockPreparationErpAutoPersistNoSteering(req)   // 请求含 tenantId/projectId → 400,在 body allowlist 与任何 I/O 之前 (OD-2)
+  normalize body                                              // 在 guard 之后:否则 body projectId 会先撞 normalize 的通用无效键码
   tenant = ON ? resolveAuthUserTenantId(req) : resolveTenantId(req,input)   // OFF 保持现有只读派生
   → runErpMaterialReadonlySource         // 产 intake.erpMaterials；空(rows<1) 已在此抛 SOURCE_RUN_EMPTY(422)→到不了下面 (OD-3)
   readProjection = publicReadonlySourceRunResult(result)      // 固定 mode:'dry_run' / evidence.internalWriteExecuted:false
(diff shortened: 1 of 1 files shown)


zensgit:
Owner re-review against current `main` (`GHSA` merged at `40ec7caba`): the prerequisite is cleared, but exact head `158f800ac` is **NOT RATIFIED yet** because the lock contains one security/compatibility contradiction and several unresolved runtime semantics. No runtime implementation is authorized from this head.

**[P1 design contradiction] OD-1 says flag OFF is byte-identical, while OD-2 changes this route from request-selectable read tenant to auth-principal tenant unconditionally. Both cannot hold.** The current read-only route deliberately uses `resolveTenantId(req,input)`; changing it even with the flag OFF breaks the existing platform-admin cross-tenant read behavior.

Owner decision for the corrected lock:
1. Select **OD-1 A**: `MULTITABLE_STOCK_PREP_ERP_SOURCE_AUTOPERSIST_ENABLED`, default OFF, normalized literal `true`.
2. **OFF:** preserve the current read-only route behavior byte-for-byte.
3. **ON:** before config/system/external-read or records access, derive the consequential read+write tenant from `resolveAuthUserTenantId`; reject a body `tenantId` before I/O, retain `workspaceId` only for same-tenant config selection, and prove query/params cannot steer either read or write. Staging remains `resolveIntegrationStagingProjectId(authTenant, undefined)`.

Required design corrections before ratification:
- **Truthful response contract:** current source evidence says `mode=dry_run` and `internalWriteExecuted=false`. The ON result must define and test an exact frozen composite in which internal persistence is true, external/production/K3 writes remain false, and raw intake/syncRunId never cross HTTP. Merely appending persist counts beside contradictory source evidence is not acceptable. OFF remains byte-identical.
- **Actual empty/failure semantics:** `runErpMaterialReadonlySource()` rejects zero rows with `SOURCE_RUN_EMPTY`; it cannot persist “zero materials + run record” as §3 currently says. Also a mid-persist exception can leave idempotently repairable partial rows but does not return reliable partial counts. Document coarse failure + rerun-with-same-syncRunId recovery instead.
- **Cache absence semantics:** T2 is upsert-only. A material absent from a later complete ERP read remains in `erp_material_master`, and candidate generation scans that whole cache. To keep this slice fast, lock T3a as **staged/RC upsert-only** and hold production promotion until a separate missing-key reconciliation policy is ratified; do not imply full set synchronization.
- Replace the draft “global-history…or stock-prep manifest (if any)” line with the exact stock-prep flag/operator/package guard location. Do not add this flag to the unrelated Global History manifest.
- Change §7 to the agreed sequential order. Ratifying T3a does **not** authorize T3b; T3b needs its own lock and follows T3a.

Once those text changes land, the substantive owner choices are settled and the next review should be a short ratification check, not another design round.

zensgit:
Owner ratification check at exact head `dfcb4971b`: **NOT RATIFIED / REQUEST CHANGES.** The GHSA prerequisite is cleared, but the revised lock still does not encode several owner decisions already recorded on this PR.

1. **Flag identity is still unresolved.** §3 uses `MULTITABLE_STOCK_PREP_ERP_AUTOPERSIST_ENABLED` and leaves a later rename optional. The settled flag is `MULTITABLE_STOCK_PREP_ERP_SOURCE_AUTOPERSIST_ENABLED`; a ratified lock must name it exactly.
2. **The response contract still appends persistence evidence to the existing source result.** The source evidence says `mode=dry_run` / `internalWriteExecuted=false`; appending an `autoPersist` block leaves a contradictory public result. Define one exact frozen ON-mode composite in which the internal cache write is true, external/K3/production writes remain false, and raw intake/syncRunId never cross HTTP. OFF remains byte-identical.
3. **§3 still documents impossible/unreliable outcomes.** `runErpMaterialReadonlySource()` rejects zero rows with `SOURCE_RUN_EMPTY`, so “zero materials + run record” is not reachable. A mid-persist exception also cannot report reliable partial create/patch counts. Lock this as a coarse values-free failure with idempotent recovery by rerunning the same sync run.
4. **The flag guard location is still draft text.** §5 says `global-history…否 stock-prep flag manifest(若有)`. Replace it with the exact stock-prep flag/operator/package guard location; do not route this unrelated flag through the Global History manifest.
5. **Workspace semantics are omitted.** Under flag ON, retain `workspaceId` only for same-tenant source-config selection, while the consequential read/write tenant comes from the authenticated principal and staging remains `resolveIntegrationStagingProjectId(authTenant, undefined)`. State and test that body/query/params cannot steer either read or write.

The staged upsert-only caveat and the independent T3b gate are now correct. After the five items above are fixed and the branch is rebased, the next pass can be a short ratification check.

zensgit:
5 ratify-blocking items added — head `279cfb452`, rebased (§3.5):
1. **Precise flag name**: `MULTITABLE_STOCK_PREP_ERP_AUTOPERSIST_ENABLED` (single, no alt).
2. **Non-contradictory response**: `autoPersist` always present — `null` ⟺ flag OFF; a T2 values-free evidence object ⟺ flag ON + persist success. Mutually exclusive.
3. **Empty/mid-failure**: empty intake → persist zero + run row, HTTP 200 (not an error); source-ok + persist-fail → whole request returns the persist values-free coarse error (no partial autoPersist, no cross-step txn; idempotent upsert ⇒ retry-safe).
4. **Guard placement**: `assertStockPreparationErpAutoPersistNoSteering(req)` runs after requireAccess + the flag check and BEFORE `loadStockPreparationReadonlySource` — flag ON + any non-empty request tenantId/projectId → 400 before any I/O; flag OFF never calls it.
5. **workspaceId** is a same-tenant scope selector, NOT a steering vector (never affects the auth-derived tenant or write target) — the guard rejects tenantId/projectId only.

Docs only. Ratify remains your gate.

zensgit:
Owner exact-head re-review at `279cfb452`: **NOT RATIFIED.** This is a live-head finding, not the earlier stale review. The new §3.5 was appended, but it does not reconcile the old sections or the current runtime contract.

**[P1] The response contract is still internally contradictory.** OD-1 says flag OFF is byte-for-byte unchanged, while §3.5 says `autoPersist` is always present and OFF returns `autoPersist: null`; both cannot hold. For ON, §2 still spreads `publicReadonlySourceRunResult(result)` and appends `autoPersist`. The current projector emits `mode: dry_run` and `evidence.internalWriteExecuted: false`, so a real internal write would still be reported as no internal write. This is exactly the append-only contradiction the previous owner decision required the lock to remove. Define one exact ON composite and preserve the existing OFF body unchanged.

**[P2] The empty-input contract is impossible under the current source-run.** `assertIntakeReady()` throws `SOURCE_RUN_EMPTY` when `erpMaterialRows < 1`, before T2 persist can run. §3 lines 46/68 instead promise HTTP 200 plus a run row. Either retain the existing 422/no-persist behavior, or explicitly authorize and scope a source-run semantic change; the latter would contradict §4 (`不改 source-run 的只读读取逻辑`).

**[P2] The document still contains mutually exclusive old text.** §3 line 32 retains the alternate flag-name choice, line 45 says a persist failure exposes partial create/patch counts, and §5 line 94 still says `global-history…否 stock-prep manifest(若有)`. §3.5 says the opposite on each point. A ratified lock must replace those clauses, not append a second authority. The exact T2 result shape is also misstated: runtime returns `skipped`, `runStatus`, and nested `evidence.targets`, not top-level `skippedInvalid`, `run:{status,...}`, or `fieldKeyNames`.

**[P2] Workspace safety is asserted, not proven.** §3.5 says a cross-tenant workspace case can be handled later. Ratification needs the present invariant and a planned discriminator now: a request `workspaceId` may select only a config under the authenticated tenant, cannot alter the staging target, and a foreign-tenant workspace/config fails before external read or persist.

The staged upsert-only caveat and independent T3b gate remain correct. Resolve the contradictions in place, then the next pass can be a short ratification check.

zensgit:
Re-review pointer — the 6 points are all fixed at the **current head `d0a5ad376`** (the cited line numbers 25/31/61-63/68 are from the earlier commit `279cfb452`; that content has since been rewritten). Per-claim proof against the current head:

| # | Was (in `279cfb452`) | Now (`d0a5ad376`) |
|---|---|---|
| ① | line 31 byte-unchanged vs 61-62 "always add autoPersist:null" | **OD-4 / L58**: OFF returns `publicReadonlySourceRunResult(result)` **with no field added**; **L63** invariant: `autoPersist` present ⟺ ON, **no null state** |
| ② | §2 diagram `{...publicReadonlySourceRunResult, autoPersist}`, projector's `dry_run`/`internalWriteExecuted:false` not overridden | **OD-4 / L60** + §2 diagram L30: ON overrides `mode:'internal_persist'` + `evidence.internalWriteExecuted:true` |
| ③ | line 68 "empty → 200 + run row" | **OD-3 / L52**: empty → `assertIntakeReady` throws `SOURCE_RUN_EMPTY` (422), persist unreachable |
| ④ | line 63 "201" but WIP `sendOk(res, body)` → 200 | **OD-4 / L62** + WIP `922fd2f76`: `sendOk(res, body, autoPersist.persisted ? 201 : 200)` |
| ⑤ | `skippedInvalid / run:{status} / fieldKeyNames` | **OD-4 / L61**: real `{ persisted, mode, created, patched, skipped, runStatus, evidence.targets }` (cited to persist.cjs:397-405) |
| ⑥ | old paras + alt flag + global-history | unified into OD-3/OD-4; §3.5②③ point there; alt flag + global-history retired |

The §2 flow diagram (the one thing the earlier commit missed) is now corrected to match OD-4 + the WIP.

**Verification**: two independent refute-first agents re-checked doc-vs-code (opening `assertIntakeReady`, the projector, T2's return, and the handler's `sendOk` — the exact thing my earlier verifier skipped) → both SATISFIES; both inert CJS suites pass at the WIP head. If a fresh review of `d0a5ad376` (doc) + `922fd2f76` (WIP) still trips on a specific line, tell me the line and I fix it — but the pre-fix line references no longer apply.

zensgit:
Owner re-review against the exact heads:

- design lock: `1b09431e090fcbcf877d0458228e92e7029cd80c`
- WIP runtime: `922fd2f76717242031c3564f30ce40020ddfdc29`

**Verdict: not ratified yet.** The core direction and most of the six corrections now match the runtime, and I independently reran the two flag-OFF CJS suites successfully. Three load-bearing contract errors remain:

1. **[P2] The rerun acceptance criterion is impossible under the current T2 implementation (§5 item 2, line 92).** `upsertErpMaterials` unconditionally calls `patchRecord` for every existing valid material row and increments `patched`. Therefore a normal second run is `mode='refreshed'`, `persisted=true`, and HTTP **201**, not “refreshed, and 200 if unchanged.” There is no equality/no-op detector today. Either specify 201 for the rerun or explicitly add a separately reviewed no-op detector; do not encode a 200 smoke expectation that the current implementation cannot satisfy.

2. **[P2] The documented dedicated steering guard placement does not match the WIP (§3.5 item 4, line 76).** The WIP normalizes the body before calculating the flag and calling `assertStockPreparationErpAutoPersistNoSteering`. Because `projectId` is not in the ERP source-run body allowlist, body `projectId` fails earlier with `STOCK_PREPARATION_ERP_SOURCE_RUN_REQUEST_INVALID`, not the documented `STOCK_PREPARATION_ERP_AUTOPERSIST_STEERING_NOT_ALLOWED`. It is still 400 and before I/O, but the claimed discriminating guard/error contract is false. Move flag evaluation + the guard immediately after `requireAccess`, before body normalization, or narrow the document/test contract to the actual generic rejection.

3. **[P2] The “nonempty intake but all rows skipped” T3a branch is unreachable (§OD-3, line 47).** The intake normalizer only emits ERP rows after material code and internal ID are present, synthesizes `erpMaterialId`, and records invalid rows as `rowErrors`; `assertIntakeReady` rejects any `rowErrors > 0` before T3a can call persist. Every row reaching T2 therefore satisfies its three required material fields. Remove this branch from the T3a contract, or label it explicitly as a defensive T2-only state that is unreachable from the approved source-run path.

Please also clean the non-blocking bookkeeping in the same revision:
- line 32 still retains the alternate `..._SOURCE_AUTOPERSIST_ENABLED` name while line 67 says it is retired;
- lines 3/28/67/103 still cite stale WIP `61caf7ff0` instead of `922fd2f76`;
- name the exact stock-prep flag-ledger/operator-contract file to be created or updated rather than the ambiguous “O-2” label, which currently resolves to the unrelated Global History ladder.

Independent verification performed:
- `stock-preparation-readonly-source-run.test.cjs`: PASS
- `http-routes.test.cjs`: PASS
- diff checks: clean

After these doc/runtime corrections, rebase and a short exact-head re-review should be enough to ratify. Flag-ON steering, real-DB, leak-bait, and mutation proof correctly remain implementation gates after ratification.

zensgit:
## Round-2 fixes landed — exact-head summary for short re-review

**Doc head:** `2153a6eb5` (this PR)  **WIP head:** `bb9724494` (`claude/stock-prep-t3a-erp-autopersist`)

Your 3 P2 + 3 bookkeeping items are fixed **in place** (old paragraphs unified, not appended). An independent two-verifier refute-first pass opened the referenced code for every claim (the miss from the earlier false-SATISFIES) and both returned SATISFIES with file:line evidence; the flag-OFF inert suites pass at the WIP head.

### P2 — behavioral
- **P2-1 re-run → 201, not 200.** `upsertErpMaterials` existing-row branch calls `patchRecord` **unconditionally** (no diff/change comparison) → a re-run of the same valid rows gives `patched>0` → `persisted=true` → `mode:'refreshed'` → handler returns `sendOk(..., autoPersist.persisted ? 201 : 200)` = **201**. The only `created=0 && patched=0` path is all-rows-skipped, which P2-3 makes unreachable; empty is pre-empted by `SOURCE_RUN_EMPTY` (422). §5(2) now states "仍 HTTP 201（不存在「无变化→200」路径）".
- **P2-2 steering guard before the allowlist.** `projectId` is **not** a member of `VALID_STOCK_PREPARATION_ERP_SOURCE_RUN_REQUEST_KEYS` (`http-routes.cjs:741`). The handler now calls `assertStockPreparationErpAutoPersistNoSteering(req)` at **:3914**, before `normalizeStockPreparationSourceRunBody` at **:3915** — so a body `projectId` (or `tenantId`) yields the dedicated `400 STOCK_PREPARATION_ERP_AUTOPERSIST_STEERING_NOT_ALLOWED` before any normalize/I/O, not the generic invalid-key code. Unified across OD-2, §3.5④, the §2 diagram, and §5(5).
- **P2-3 `skipped_empty` unreachable in T3a.** `MATERIAL_REQUIRED_FIELD_IDS = ['erpMaterialId','erpMaterialCode','erpMaterialInternalId']`. Intake `normalizeErpMaterial` pushes a rowError (and returns null) when `erpMaterialCode` or `erpMaterialInternalId` is missing, and always synthesizes `erpMaterialId`; `assertIntakeReady` throws `SOURCE_RUN_REQUIRED_SHAPE_MISSING` (422) on `rowErrors>0`. So every row that reaches persist satisfies all three required fields → `isUpsertable` true → `mode:'skipped_empty'` cannot occur via the T3a path. OD-3 states this (removed the earlier "may return skipped_empty" line).

### Bookkeeping
- **B1** — single flag name only: `MULTITABLE_STOCK_PREP_ERP_AUTOPERSIST_ENABLED`. No alternate/variant name remains; the WIP flag read (`http-routes.cjs:570`) matches.
- **B2** — zero stale WIP SHAs: `61caf7ff0` / `922fd2f76` → 0 hits; all refs now point at `bb9724494`.
- **B3** — the doc no longer claims a stock-prep flag manifest file (none exists on disk); it correctly attributes `scripts/ops/multitable-global-history-flag-status.mjs` to the Global History line (that script does **not** track this flag — grep count 0).

### Inert proof (flag OFF, WIP `bb9724494`)
`node __tests__/stock-preparation-readonly-source-run.test.cjs` → PASS; `node __tests__/http-routes.test.cjs` → PASS (incl. `RejectsUnapprovedConfig` / `RejectsSystemKindMismatch`). Flag defaults OFF, so the guard reorder leaves the read-only path unchanged.

This is a design-lock + WIP scaffold only — **externalWrite=false**, no K3 Save/Submit/Audit, no production write. Ready for your exact-head short re-review; ratify is yours. Once ratified I'll run the T3a flag-ON verification (real-DB smoke in the `plugin-tests.yml` whitelist + red-first steering test asserting body `projectId` → dedicated code before I/O + leak-bait both directions + load-bearing mutation).

zensgit:
## Owner ratification — RATIFIED

Exact heads reviewed:

- design lock: `2153a6eb5e08dc537e8001de44806dd1526921cb`
- WIP cross-check: `bb9724494f27da073cdda92dc53323d5d1fee08b`

The three round-2 P2 corrections are load-bearing and now match the real implementation:

1. reruns are correctly specified as unconditional refresh patches, therefore `persisted=true` / `mode='refreshed'` / HTTP 201;
2. the flag check and dedicated steering guard now run before body normalization, so body `projectId` can receive `STOCK_PREPARATION_ERP_AUTOPERSIST_STEERING_NOT_ALLOWED` before any I/O;
3. `skipped_empty` is correctly documented as unreachable from the T3a normalized-intake path and retained only as a defensive T2 state.

The single flag name and WIP references are consistent, and the document now explicitly states that this line has no dedicated flag manifest.

Independent verification at the WIP head:

- `stock-preparation-readonly-source-run.test.cjs`: PASS
- `http-routes.test.cjs`: PASS
- runtime and document diff checks: clean

**Ratified scope:** T3a only: default-OFF `MULTITABLE_STOCK_PREP_ERP_AUTOPERSIST_ENABLED`, auth-principal tenant derivation and pre-I/O steering rejection while ON, server-internal source-intake-to-T2 persistence, values-free public evidence, internal MVP tables only, and the §5 flag-ON/real-DB/mutation validation.

**Not authorized by this ratification:** enabling the flag in production, external ERP/K3 writes, authoritative reconcile/deletion of stale cache rows, changes to other GHSA step-2 read routes, T3b implementation, or T4/RC-A dispatch.

One non-blocking P3 remains for the document update before merge: line 40 still says “O-2 ladder,” although lines 41/107 correctly establish that this stock-prep line has no such manifest and that the Global History O-2 ladder is unrelated. Remove that ambiguous phrase or replace it with “staged opt-in discipline”; this does not block T3a implementation.

Proceed with the planned flag-ON discriminating tests, plugin-tests real-DB smoke, bidirectional leak-bait, and load-bearing mutations. The WIP remains implementation-held until those gates and an independent review pass.

PR #4266, opened 2026-07-14
feat(stock-prep): T4 — non-empty prep-line extended smoke (internal-only, externalWrite=false)

zensgit:
## Scope (closeout §5b — stock-preparation-mvp-closeout-and-verification-20260712.md)

T4 = the non-empty prep-line extended smoke: a harness that drives the internal chain
**cache → auto-match → unit-conversion → generation → exception → audit** to a **non-empty prep-line
table and the full `ready:true` flip** over real HTTP. The W6 postdeploy smoke deliberately pins the
EMPTY branch (its own header documents that a `ready:true` flip was impossible without an
`erp_material_master` write surface); T2 (#4206) landed that surface, and T4 is the follow-through.
It directly exercises the §5b Wave-3 checklist items: project endpoint PASS · ERP cache rows>0 ·
auto-match consumed the cache · prepLineRows>0 · audit 8/8 · externalWrite=false.

**Hard boundary:** internal tables only. Every write goes through the pre-existing admin-gated
stock-preparation HTTP surface into the 9 internal MVP tables. `externalWrite` stays `false` — no
K3 Save/Submit/Audit, no apply-writer, no external-system call exists on any route this smoke
touches, and no permission/RBAC layer is modified (the harness is a pure HTTP client + 2 files of
CI wiring + 1 doc).

## What's in the PR

- `scripts/ops/stock-preparation-prep-line-extended-smoke.mjs` — the T4 harness (58 checks):
  - 3-line salted BOM fixture; cached material A's `erpMaterialCode` EQUALS drawing A (the
    exact-code join) — run 1 → `201 partial`, `created.lines=1` (**non-empty on the first run**) +
    2 blocking `missing_mapping` exceptions; ready invariant asserted.
  - cache B+C → candidates re-sync creates 2 NEW exact-code candidates (A re-emits historical,
    skipped) → confirm by `mappingId` (matchMethod stays `exact_code_candidate` — the cache row is
    confirmed, not replaced) → one single resolve + one same-reason bulk resolve
    (`mapping_confirmed`) → run 2 → `status=ready`, `ready=true`, unresolved-blocking **2 → 0**,
    3/3 prep lines matched + converted, human resolutions preserved by the create-only re-run.
  - audit 8/8 (all closed-vocabulary actions, scoped rows); fail-closed probes
    (`409 CONFIRM_MAPPING_TARGET_INCOMPLETE` on a no-ERP-identity confirm;
    `400` closed-allowlist rejection of a body `tenantId` on the T2 route).
  - values-free: the sanitizing output layer (registries / count projection / handle shapes /
    self-scan) is **imported from the W6 smoke** rather than re-implemented, so the harnesses
    cannot drift; global leak scan over every response (the `/plan` same-origin echo is the only
    exemption, pinned at exactly 1); closed 10-key prep-line row projection shape-lock.
  - cleanup: retires all 5 mapping rows + the unit rule; re-syncs the 3 cached materials to
    `inactive` (which doubles as the T2 upsert-patch proof: `refreshed`, `patched {materials:3}`).
- `scripts/ops/stock-preparation-prep-line-extended-smoke.test.mjs` — companion contract test
  (10 tests, `node --test`, no server/DB): fixture invariants incl. the cache-join invariant,
  projection shape-locks, closed error-code registry superset + poison-injection, audit-vocabulary
  parity, summary-block shape. Mutation-checked: breaking the cache-join line makes it fail
  (and the full rehearsal below independently caught that same planted mutation end-to-end).
- `.github/workflows/stock-preparation-prep-line-extended-smoke.yml` — same two-job shape as the
  permission-lists smoke lane: PR-triggered `contract` job (runs the T4 companion AND the W6
  companion, since T4 imports it) on the harness/test/W6 paths; dispatch-only deployed `smoke` job
  mirroring the W6 lane's token resolver, env, fixed concurrency group, values-free summary +
  artifacts.
- `docs/development/stock-preparation-t4-prep-line-extended-smoke-runbook-20260714.md` — operator
  runbook stub (prerequisites, serial-run rule, dispatch + manual recipes, PASS criteria, local
  rehearsal record; deployed-lane PASS section deliberately empty).

## Env contract

Same as the W6 lane: `--base-url` required (fail-closed arg parse), optional `--tenant-id` /
`--workspace-id` / `--project-prefix` / `--timeout-ms` / `--out-dir`; bearer token via
`METASHEET_AUTH_TOKEN` (workflow resolves it exactly like the W6 lane). Serial-run assumption
documented: never dispatch concurrently with the W6 smoke (shared tenant-level tables).

## What was / wasn't rehearsed (honest)

- **Rehearsed locally, full chain, real everything:** fresh `createdb` + full migration set →
  the real core-backend server (`tsx src/index.ts`, all 14 plugins active) → the harness over real
  HTTP with a locally minted admin token (documented non-production trusted-claims path).
  Result: `pass=true`, 58/58 checks, `run1 created.lines=1`, `run2 ready=true` with the 2→0 flip,
  `prepLines2RowCount=3`, `auditActionsCovered=8/8`, 46 responses leak-scanned clean, cleanup
  verified in `meta_records` (5 mappings + rule retired, materials inactive; the 3 generated
  prep-line rows physically present in the internal sheet). The rehearsal also caught a planted
  test mutation that had leaked into the committed harness (fixed in this branch's history) —
  i.e. the harness demonstrably fails when the cache does not drive the match.
- **Not run:** the deployed-environment dispatch (owner-operated lane; the runbook's PASS record
  section stays empty until a real dispatch happens).

## CI wiring proof (triggered ≠ verified)

- The new workflow's `contract` job is path-triggered by this PR itself (it touches all its
  trigger paths) — the check must appear and pass on this PR's head.
- Negative control run locally: mutating the fixture's cache-join line turns the companion test
  red (9 pass / 1 fail), so the wired test is load-bearing, not vacuously green.
- Note: the pre-existing W6 companion (`stock-preparation-mvp-postdeploy-smoke.test.mjs`) ran in
  NO workflow until now; the new `contract` job closes that gap for both harnesses.


zensgit:
**Independent adversarial review (not self-review): SAFE-TO-MERGE-NOT-DISPATCH · 0 P1 / 0 P2.** Armed squash — the smoke job stays `workflow_dispatch`-gated so nothing runs on merge.

Every adversarial claim was refuted-first and held:
- **Runtime-free**: 4 files, all added, all under `scripts/ops` / `.github/workflows` / `docs`; zero `packages/core-backend/src` / `plugins/.../lib` / `apps/web/src`. Imports the pre-existing sanitizing smoke (has its own `invokedDirectly` guard).
- **externalWrite=false (structural)**: every route hit is `/api/integration/stock-preparation/*` MVP + `/status`; no K3 Save/Submit/Audit / apply / external-write.
- **Non-vacuous**: the `contract` job actually runs `node --test` (T4 10/10 + W6 15/15 locally); real equality assertions on real routes.
- **values-free both directions**: leakScan asserts clean-passes AND sentinel-fails; POISON injection + 10-key projection shape-lock.
- **Load-bearing (mutation-proven)**: break the cache-join → 1 fail; inject a leak-bait key into `PREP_LINE_ROW_KEYS` → 1 fail.
- **Merge-safe**: triggers only `pull_request` + `workflow_dispatch`; smoke gated on dispatch; `contract` runs `contents: read`, no secrets; no closing keyword. CI on head: contract=pass, deployed-smoke=**skipping**.

Bonus: net CI gain — the W6 companion test previously ran in no workflow, now does.

**Two NITs → fold into T4-final** (when the real T3a/T3b feeder assertions are added): (1) `S.externalWrite=false` is a hardcoded declaration, not an observed check; (2) the deployed-path global leak scan asserts `leakingLabels.length===0` with no positive-control floor (`scanned >= N`) — vacuous only if COLLECTED were empty (unreachable today). Neither blocks the base.

PR #4252, opened 2026-07-14
docs(multitable): Time Machine W0 verification + W0-1 correctness conditions (C1–C8)

zensgit:
**Deliverable for the Time Machine completion goal** — verification + gap-analysis + plan (not a completion declaration).

## What this is
The design+verification MD the goal asked for, corrected to **verified current `main` state**. Every claim is anchored to `origin/main` (`file:line`) or a runnable golden.

## Key contents
- **§1 Verified landed state** — W0 write-site slices #4245–#4249 + precheck #4234 MERGED (ratified 2026-07-13); the 8-path revision gap is substantially closed.
- **§2 The merged precheck (#4234) is a partial guard** — it is a `live-vs-latest` comparator that refuses the cases it was specced for, but does not yet cover the residual conditions in §3 (chain-shape, ordering-under-concurrency, deleted-record chains, check-vs-write boundary). Specific residual windows were given to the maintainer directly (repo is public).
- **§3 Eight load-bearing conditions #4250 must add before ratify (C1–C8)** — each anchored to `main`: contiguity (C1), version↔event-time monotonicity since `created_at DEFAULT now()` + reconstructor sorts by `created_at` (C2), deleted-record coverage (C3), phantom-insert fence / SERIALIZABLE (C4), exact-set not count — no `UNIQUE(sheet_id,record_id,version)` (C5), trusted-since checkpoint + rollout (C6), marker blast-radius — `action CHECK` only allows create/update/delete (C7), same-locked-txn execute re-check (C8).
- **§4 Corrected dev order + model dispatch + estimates.**
- **§6 Verification evidence** — embedded durable healed-gap acceptance golden (W0-1 must make it green) + the W0-1 golden matrix.

## Coordination
A parallel `zensgit` session owns the W0 backend impl (merged slices + open #4250). This is the **non-duplicative** deliverable that moves #4250 from REQUEST_CHANGES toward ratifiable. Recommends closing superseded #4235 and folding C1–C8 into #4250.

Docs-only. No code. Draft — no merge without owner review.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4250, opened 2026-07-14
docs(multitable): W0-1 design lock — HISTORY_INCOMPLETE contiguity + trusted-since (PROPOSED)

zensgit:
Your P1 on #4235 proven by construction (healed-gap golden reds: revert-preview to a gap-window T returns 200 not 409). #4235 is faithful to §0.6; the flaw is §0.6's design. Corrected: version-contiguity + lock/unlock chain-marker (lock/unlock bumps version w/o revision = the false-positive source) + persistent trusted-since + same-locked-txn execute re-check. **Zero runtime, PROPOSED, ratify before impl** (destructive path + schema change). OD-W0-1..4. #4235 does NOT land until this lands.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4269, opened 2026-07-14
fix(multitable): W0-1 — generation-aware history contiguity replaces live-vs-latest §0.6 comparator (healed-gap closed; C2/C3/C6 deferred)

zensgit:
## W0-1 — generation-aware history contiguity (owner-directed §0.6 correction)

Replaces #4234's **live-vs-latest** comparator in `history-integrity-precheck.ts` with a **generation-aware contiguity** proof. Authority: `docs/development/multitable-global-history-w0-1-history-incomplete-contiguity-trusted-since-design-lock-20260713.md` §6 (owner ruling 2026-07-14, PROPOSED lock **folded into this PR**) + the C1–C8 conditions in #4252. Scope is exactly the owner's four items; nothing else is wired.

**Base:** `origin/main` @ `86fa1d85c` (rebased twice for mid-flight main movement). HOT-CORE + schema/txn on the destructive-recovery path — landed via the sync-then-gate-then-arm-fast discipline: verdicts are head-scoped, every rebase paid a re-gate. The governing design lock is folded into this PR and remains **PROPOSED** — merging this implementation does NOT ratify the W0-1 lock; C2/C3/C6 and `recreateFieldFromConfig` stay owner-gated (see docket below).

### Gate ledger (adversarial Opus, refute-first, head-scoped)
- **Round 1 @ `075fdfe5e`** — MERGE_WITH_NITS: owner's 4 items all pass; 7 constructed scenarios all bite under mutation; fresh-DB migrate + CI-lane wiring proven. Findings: P3-1 (in-txn marker SELECT poisons the reset-execute txn in the pre-migration deploy window → 25P02 → 500 instead of clean 409), NIT-1 (design-lock citation dangling), NIT-2 (automation marker writes non-atomic).
- **Hardening `7a5372f99`** — all three addressed: marker READ path now gated behind the WRITE path's txn-safe `information_schema` probe (`hasVersionMarkerTable`, exported); automation lock/unlock bump+marker wrapped in `withTransaction` (prod supplies `deps.transaction` — verified at `automation-service.ts:841`); PROPOSED lock folded in from the #4250 branch.
- **Round 2 @ `7a5372f99`** — MERGE_CLEAN: P3-1 repro flips (clean verdict, no txn poisoning); probe proven load-bearing under mutation; memoization hazard checked (positive-only cache, no suite pollution in combined single-pg run).
- **CI truth catch `91d14b3a6`** — Plugin System Tests (required test 18.x/20.x) was RED at every prior head incl. the original delivery: the three source-scanning guard specs (cross-base / rank-8 lock / OD-6 revision-disposition) accept markers only within fixed windows (5/3/3 lines) above the SQL anchor, and the W0-1 comments + withTransaction wrapper pushed the automation-site markers out of window. Fix is comment-placement-only; full CI-mirror no-DB run 6480 passed / 0 failed.
- **Rebase onto `86fa1d85c`** — composes with **#4261** (interim revert-execute master gate, explicitly scoped "until the full W0-1 correctness fix lands" = this PR): flag OFF → 403 before the precheck; flag ON → contiguity verdicts as gated. The new contiguity realdb spec opens `MULTITABLE_ENABLE_SHEET_REVERT` in its test env (`7de0093a9`, test-only), mirroring its PIT_RESET handling; #4261 had already adapted all pre-existing revert-execute specs on main.
- **Round 3 @ `7de0093a9`** — MERGE_CLEAN on the rebase composition; then CI's DB whitelist step ran for the FIRST time at this head (every earlier run had died at the unit step, masking it) and caught a REAL predicate bug none of the gates saw: `undelete-pit` scenario (g) got `HISTORY_INCOMPLETE` instead of `PREVIEW_IDENTITY_INVALID`.
- **Comparator fix `a1cbd9639`** — the generation-boundary order key was packed as `epoch*1e6 + version` in a float64; at 2026 epochs (~1.7e18, ULP ~256) the version tiebreak was silently annihilated, so same-millisecond delete+create pairs collapsed and the current generation was mis-drawn. Replaced with a structural comparator: **(created_at epoch-ms, version, non-delete-before-delete)**. Version is the same-ms primary tiebreak (same-epoch multi-vintage churn — delete@2/create@3/delete@4 at one T, constructed by the 4c-3 §7 goldens — requires it; a static "create after delete" action rank was tried first and false-refused that suite). Pinned by 5 same-epoch unit goldens at real magnitudes. Verified against the FULL 202-file CI DB whitelist on a fresh PG (202/202, 1797 passed) + full no-DB suite (6515 passed).
- **Round 4 @ `a1cbd9639`** — final gate at the arm head: per-leg comparator mutations, adversarial hunt for reachable false-accepts, independent full-realdb re-run; plus required CI at the same SHA.

### The contiguity predicate (exact-set + generation + delete-reuse)
A LIVE record at `liveVersion = V` is trustworthy **iff** every version `k ∈ [genStart..V]` of its **current generation** is occupied by **exactly one** canonical event:
- a `create`/`update` revision at version `k`, **OR**
- a `lock`/`unlock` marker at version `k`.

`generation = count(create revisions at-or-before)`. A record's chain is a sequence of generations separated by `delete` revisions; the **current generation** is the suffix strictly after the last delete (a trash-restore / PIT-resurrect writes a fresh `create` at v1, starting a new generation). Fail-closed on: a version with **no** occupant (a hole = uncaptured data write — **the healed-gap**); **two** occupants at one version (duplicate/conflict, C5); an occupant version `> V` or `< genStart` (out-of-range, C5); the latest event being a `delete` on a live row (uncaptured resurrection); or no create/update revision at all (zero-revision fingerprint).

**Delete-reuse (C5):** a `delete` revision **reuses** the last live version, so it is **excluded** from the per-version occupant count. Neither `count(distinct version)` nor `count(*)` can be the criterion — pinned in the design and in `checkRecordChainContiguity` unit tests.

The pre-existing live-vs-latest content projection is **retained** as the §1.3 second layer (contiguity proves the chain is complete; the projection proves live == latest captured content). No #4234 coverage is lost; the existing 9-golden precheck suite still passes.

### Marker mechanism (OD-W0-1) + why (C7 avoidance)
**Mechanism (b): independent `meta_record_version_markers` table** (`zzzz20260713150000_...`), `UNIQUE(sheet_id, record_id, version)` for C5-fail-closed markers. Chosen over (a) lightweight revisions **specifically to avoid the C7 ripple**: `meta_record_revisions.action CHECK (create/update/delete)` stays closed, and History UI / retention / reconstructor are untouched. Markers are written at **all four** lock/unlock sites — HTTP `univer-meta.ts` lock/unlock (atomic in `pool.transaction`) and automation `automation-executor.ts` lock/unlock — via `recordVersionMarker` (idempotent `ON CONFLICT DO NOTHING`, deploy-window safe: a missing table pre-migration ⇒ fail-closed, never a crash). The 4 sites keep their `revision-exempt` + `lock-mgmt` disposition markers (OD-6 guard #4258 and the rank-8 lock guard both stay green).

### isSystemSheet predicate (single source of truth)
`isSystemSheet(sheet) = isApprovalProjectionBaseId(base_id) || isSystemPeopleSheetDescription(description)` in the new `multitable/system-sheet-predicate.ts`. `SYSTEM_PEOPLE_SHEET_DESCRIPTION` + `isSystemPeopleSheetDescription` moved there (re-imported by `univer-meta.ts`), so both People-sheet filtering and the history exclusion key off the same sentinel. The precheck skips contiguity for system sheets (server-regenerated read models).

### C4 fence + C8 same-txn
`reset-execute` (the destructive delete path) now acquires a **per-sheet `pg_advisory_xact_lock`** as the first statement of its destructive transaction, then **re-runs the precheck inside that transaction** (`HistoryIncompleteInTxnError` → rollback → the same values-free 409). Chosen over SERIALIZABLE: lowest blast radius, no global-isolation change, no serialization-failure retry loop. `revert-execute` is non-destructive (per-record CAS field-reverts + id-collision-guarded resurrects) so its precheck stays at the compute layer; the healed-gap refusal for revert fires there. **Proven with a constructed two-connection race** (not a sequential argument) — see goldens.

### Goldens (real-DB, mutation-proven) — all green locally
`tests/integration/multitable-history-contiguity-realdb.test.ts` (wired into `plugin-tests.yml`) + `tests/unit/multitable-history-contiguity.test.ts`:

| Golden | Asserts | Mutation → red (verified) |
|---|---|---|
| **HEALED-GAP** (fail-first, #4252 §6.1) | v3 record, revisions {v1,v3}, live==v3 → revert+reset preview+execute ALL 409, zero writes | neuter contiguity → **200** |
| **LOCK-MARKERS** (positive control) | record bumped by all 4 lock/unlock sites writes markers → precheck PASSES | neuter marker write → markerCount **0** |
| **SYSTEM-SHEET** (projection + people) | non-contiguous record NOT refused (excluded); plain-sheet discriminator IS refused | neuter isSystemSheet → **falsely refused** |
| **DELETE-REUSE** | delete revision reusing last live version is not a false hole (healthy sibling reverts) | (unit) delete counted as occupant → wrong reason |
| **PHANTOM-INSERT RACE** (C4/C8) | phantom uncaptured insert while reset holds the fence → 409, D not deleted | neuter in-txn recheck → **200**; neuter advisory fence → **can't block** |
| **POSITIVE CONTROL** | full healthy chain reverts+executes at T | (guards refuse-everything) |

Plus 13 `checkRecordChainContiguity` unit cases (healed-gap, marker-heal, generation/resurrect, delete-reuse, duplicate, out-of-range, zombie, zero-revision). All prior suites stay green: `multitable-history-incomplete-precheck-realdb` (9), `multitable-reset-pit-realdb` (13), `multitable-reset-pit-inbound-capture-realdb` (3), disposition scanner + rank-8 lock guard (42), + neighbor batch (76).

### Migration proof (zzzz, fresh-DB)
`zzzz20260713150000_create_meta_record_version_markers` sorts **last** on a fresh-DB full migrate; table + UNIQUE constraint + CHECK verified; re-migrate is a no-op (idempotent); `down()` verified. Not in any `MIGRATION_EXCLUDE` list (independent table, no FK/view deps) — runs in the CI test DB.

### EXPLICIT DOCKET — open residuals (this green does NOT mean "history fully trustworthy")
This slice closes the **live-record mid-chain healed-gap (C1/C5)** + **lock-marker false-refusal (C7-partial)** + **check/write race (C4/C8)**. Still open, deferred by the owner, **must not be masked**:
- **`recreateFieldFromConfig` (`univer-meta.ts:~6522`)** — content-integrity gap, owner-ruled **MUST-WRITE**, its own follow-up rung (OD-6 `revision-pending`).
- **C2 time-monotonicity** — reconstruction sorts by `created_at` (= txn-start); version-ascending/time-descending under concurrency can still select the wrong T-snapshot. Two pinned manifestations, in OPPOSITE directions (round-4 gate): (fail-closed, same-ms) a restore `create@v1` in the same millisecond as its preceding delete sorts version-below it and refuses — annoying, never corrupting; (fail-OPEN, cross-ms) a time-reversed event (e.g. a v2 event timestamped after its generation's delete via clock skew / txn-start inversion) can land in the wrong generation and MASK a real same-version hole → `ok:true`. The pre-fix packed-float encoding accepts the identical fail-open shape (proven by the gate), so neither is a regression of this PR — but the eventual C2 time-anchor must close the fail-open direction specifically, not just the false-refusals.
- **C3 deleted/tombstoned + resurrected (cross-generation) chains** — enumeration is live-only; reconstruction across a prior-generation boundary is not validated.
- **C6 durable trusted-since watermark + rollout protocol** — pre-marker lock bumps (historic or rolling-deploy window) have no marker and fail **closed** (refused), not grandfathered.

### #4235
Should be **closed** (parallel same-shape live-vs-latest §0.6 draft, superseded); its real-DB goldens are folded into this branch's suite per #4252's recommendation.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -368,6 +368,7 @@ jobs:
             tests/integration/multitable-undelete-pit-realdb.test.ts \
             tests/integration/multitable-reset-pit-realdb.test.ts \
             tests/integration/multitable-history-incomplete-precheck-realdb.test.ts \
+            tests/integration/multitable-history-contiguity-realdb.test.ts \
             tests/integration/multitable-d1c-form-submit-revision-realdb.test.ts \
             tests/integration/multitable-d1c-plugin-revision-realdb.test.ts \
             tests/integration/multitable-d1c-automation-revision-realdb.test.ts \
--- a/docs/development/multitable-global-history-w0-1-history-incomplete-contiguity-trusted-since-design-lock-20260713.md
+++ b/docs/development/multitable-global-history-w0-1-history-incomplete-contiguity-trusted-since-design-lock-20260713.md
@@ -1,0 +1,88 @@
+# W0-1 — `HISTORY_INCOMPLETE` 可信性纠错：连续性 + trusted-since + 同锁事务 — DESIGN LOCK (PROPOSED)
+
+- **Status**: **PROPOSED — 2026-07-13. NOT ratified. Design-first, ZERO 运行时改动。** owner 复审判 §0.6 现草案(#4235)「**不能安全落地**」——本锁把 §0.6 从 live-vs-latest 升级为**连续性证明**。**破坏性恢复路径 + 需 schema/行为变更**,故按 owner 一贯纪律(D-1c/D-2/4c-*/R13-C)**design-lock first,ratify 后才动**。W0 顺序门的第一刀,gate 一切 revision 5 刀之前。
+- **Provenance**: primary-source @ origin/main;healed-gap flaw **已实证**(golden `/tmp/r13-w0-healed-gap-golden.txt`,对当前 #4235 **红**)。#4235 impl **faithful to §0.6**(其 gate=APPROVE-with-hardening 0P1/0P2),缺陷在 **§0.6 设计**本身。
+
+## §0 已证缺陷(owner P1,实证)
+- **healed-gap**:#4235 comparator = `LEFT JOIN LATERAL … ORDER BY created_at DESC LIMIT 1` = **live vs LATEST**。构造:记录 version=3,revisions 仅 {v1,v3}(v2 是无 revision 的 uncaptured 写),live==v3 snapshot ⇒ 预检**通过**。但 revert/reset 到 T∈[v2,v3) 用 latest≤T=**v1** 重建=错(记录当时是 v2)。**实证**:revert-preview(T1) 返回 **200 非 409**(golden 红)。tail 健康、**中段有洞**,live-vs-latest 结构性看不见。
+- **check→write 竞态**(owner P1 + #4235 gate P3-1):execute 预检是破坏性事务**前**的池化 READ COMMITTED 查询 ⇒ check→write 间可落无 revision 写。
+
+## §1 设计:连续性证明(替换 live-vs-latest)
+重建到 T 可信 **iff** 该记录 revision 链在 ≤T 范围**连续无洞**。判据不是「live==latest」,而是「**每个 version 都有链上条目**」。
+
+### 1.1 版本连续性 = 完整性的可观测代理
+每次写(捕获或否)`version+1`。捕获写另写 revision(该 version)。故:一条记录 version=N,若其 revisions 的 version 集**缺某 k**(1≤k≤N),则 version k 是一次**无 revision 的写** = 潜在 uncaptured 数据写(= 洞)。`count(distinct revision.version) < live.version` ⇒ 有洞。
+
+### 1.2 ⚠️ 关键难点:lock/unlock 合法地跳 version(假阳源)
+`univer-meta.ts:16403/16417` lock/unlock `SET locked=…, version = version + 1`,**不写 revision**(设计如此)。故一条**健康**记录经 lock/unlock 也会造 version 洞——**version-contiguity 单独会误拒 lock/unlock**(违反 G-HI-2)。**必须区分「lock/unlock 跳版」与「uncaptured 数据写跳版」**——retroactively 不可区分,故:
+- **lock/unlock(及一切合法的无 revision version-bump)必须留链上标记**。三选一(**OD-W0-1**):
+  - **(a) 轻量 revision**:lock/unlock 写一条 `action='lock'/'unlock'` 的 revision(snapshot=当前 data,不算数据变更)。链变密:每 version 有 data-revision 或 lock-revision。**简单,复用 revision 表**;代价=History Center 会看到 lock 事件(可过滤)。
+  - **(b) 独立 marker 表**:`meta_record_version_markers(record_id, version, kind)`。链外记「此 version 是 lock」。**不污染 revision 表**;代价=新表+双读。
+  - **(c) lock/unlock 停止 bump version**:metadata-only,version 只跟数据变更 ⇒ 连续性==数据完整性,无需 marker。**最干净但改并发语义**(lock 的 CAS 依赖 version?须核 `:16403` 的 CAS 用途)——**风险最高,须证不破 lock 并发**。
+- **推荐 (a)**(复用现有表 + 密链,History Center 过滤 lock 事件即可);(c) 若能证 lock 不依赖 version-bump 做 CAS 则更优,但须独立证。
+
+### 1.3 派生字段不 bump version ⇒ 自动无洞(G-HI-3 免费绿)
+formula/rollup/lookup/auto-number 物化写 `data` **不** bump version(`formula-engine.ts:345` 等)⇒ 不造 version 洞 ⇒ 连续性检查天然不误拒它们。§0.6 item 2 的「投影排派生」仍保留作**内容**比较的第二层(密链证连续 + 投影证内容一致,双层)。
+
+### 1.4 持久化 trusted-since(历史过渡,owner「持久化 trusted-since」)
+历史记录(本修复前)的 lock/unlock **无 marker** ⇒ 有真实旧洞,无法 retroactively 补。故:
+- durable **`revision_trusted_since_version`**(每记录;或 per-sheet 水位)= 密链 regime 起点。重建到 T 可信 iff latest-rev-≤T 的 version ≥ trusted_since **且** [trusted_since, 该 version] 连续。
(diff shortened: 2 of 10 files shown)


zensgit:
**Coordination (parallel session): a later owner review round produced a v2 design lock — #4262 — covering the same W0-1 scope; flagging the deltas so the two tracks reconcile in one place before ratify/merge.**

The owner's subsequent adversarial review (of the other session's draft implementing the same §6 ruling) added requirements beyond the ruling this PR builds on. Deltas vs this PR's approach:
1. **Causal order:** #4262 v2 requires a persistent monotonic `seq` column as the chain-order primitive — the review found `(created_at, version, id)` ordering non-causal (`created_at = now()` = txn-start; same-ms create/delete collide; version resets per generation; UUID ids are random). This PR's `orderKey` is (created_at epoch, version, id) — same primitive; the current-generation scoping narrows but may not eliminate the same-ms boundary-split risk.
2. **C8/C4 (owner: 'C2、C4、C8 不能作为以后再说'):** v2 puts the same-txn execute re-check + a whole-sheet writer fence in-scope (a survivor-at-T outside `affectedIds` can be concurrently updated after a pre-txn check → commit ≠ T). This PR defers C2/C3/C6 and (per its title) doesn't cover the execute-path race either.
3. **Marker consumers:** this PR's separate `meta_record_version_markers` table is arguably *cleaner* than v2's action-CHECK extension — markers structurally never leak into history-projection/before-hydration/restore-by-version consumers (the review's P1-c). If the marker-table shape wins at ratify, v2 §3B collapses to just the reconstructor/PIT handling.
4. **System-sheet identity:** the review flagged description-based exclusion as user-spoofable (sheet create API accepts `description`) — v2 requires a non-user-settable `system_kind` + a spoof golden. Worth checking which signal `system-sheet-predicate.ts` here keys on.
5. **Trust/retention:** v2 adds a persistent checkpoint + retention trust-floor (prune of a generation's `create` breaks derived numbering) + a fail-closed no-auto-down migration contract.

Also FYI: the interim master gate merged as #4261 (`MULTITABLE_ENABLE_SHEET_REVERT`) — see the #4256 coordination note re the flag-name collision.

zensgit:
**Owner mechanism decision (2026-07-15): the ordering primitive is `seq`.** #4269's epoch-ms/version/delete-last comparator, and its C2/C3/C6 handling, are **superseded by #4262 v3.5** (shared persistent causal `seq`). #4269 is retained as: the landed first-cut, the flag-OFF compatibility & rollback path — but it is **no longer a trust basis for history completeness**, and **strict mode must NOT fall back to it**. This is a mechanism decision only; formal ratification of #4262 still awaits a full containment PASS. (See the current map #4319 §0.)

PR #4262, opened 2026-07-14
docs(multitable): W0-1 v3.6 — retract premature v3.5 ratify + close the two Highs (C2 reconstruction, C3 asOf-generation)

zensgit:
**v3.6 (`e45ecc2ff`) — owner re-review 2026-07-15.** The earlier v3.5 ratify (`7411f77ab`) was **premature and is RETRACTED**; its containment PASS is **invalidated** by the compose-validation cwd bug (#4329). **Accepted:** `seq` replaces epoch-ms as the strict-mode chain-ordering primitive. **NOT ratified:** the C2/C3/C6 design. Five findings closed in §0.6/§2b/§4/§10:
- **[High] C2 not closed by seq** — reconstruction still selects by `created_at ≤ T` (=txn-start). **§2b (NEW):** a linearized `effective_at` (`clock_timestamp()` under the L4 fence) + reconstruct-by-`effective_at`→seq boundary, never `created_at`. Hard-depends on L4.
- **[High] C3 terminal-only** — §4 now checks the **target asOf generation** (the one `T` falls in), or all post-checkpoint reconstructable generations; not terminal-only.
- **[Medium] seq allocation ≠ commit order** — causal only under L4 fence + L5 checkpoint; 'prospectively causal on landing' withdrawn; strict enablement **HARD-gates on L4 + L4b + L5**.
- **[Medium] exact bigint/string seq comparator** (no `Number()`) + a >2^53 value test.
- **[Medium] #4316 cwd** → #4329.

**+2 lanes** (reconstruction linearization + asOf-generation); est. **~5–8 pw** on top of #4269. **Ratify pending** a valid containment PASS (#4329 merged + re-run) + these completions. Docs-only, Draft, NOT RATIFIED.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4273, opened 2026-07-14
bench(multitable): R14.C — Time Machine real-scale benchmark (harness + evidence MD)

zensgit:
## Summary

- Adds a re-runnable, CI-independent load-bench harness: `packages/core-backend/scripts/bench/timemachine-scale-bench.ts` — measures `reconstructRecordsAtT`, `precheckSheetHistoryIntegrity` (#4234), a standalone prototype of the #4262 v2 per-generation contiguity window-function query, and the revert/reset preview+execute routes, at N = 1,000 / 5,000 / 10,000 / 50,000 synthetic records per throwaway sheet.
- Adds the results MD: `docs/development/multitable-r14c-timemachine-scale-benchmark-20260714.md` — measured numbers (two full runs for stability), EXPLAIN ANALYZE highlights at 10k, and an evidence-only analysis section for the owner's R14 decision (Option A base-wide atomic restore vs Option B granular recovery) and #4262 §1/§4.
- **No product decision made** — R14 stays an owner call. **No runtime source changed** — harness + docs only.

## Post-review update (2026-07-14)

Owner P1 (destructive-misoperation risk, merge blocker) and P2 (overstated conclusions) findings are
addressed in the two follow-up commits on this branch:
- **P1 (harness hardening):** a fail-closed preflight now requires three independent opt-ins
  (`BENCH_ALLOW_DESTRUCTIVE=1`, a local `DATABASE_URL` host or `BENCH_ALLOW_REMOTE_DB=1`, a
  `/bench|test/i` database name or `BENCH_ALLOW_NONSTANDARD_DB_NAME=1`) before any DB connection is
  used; the harness no longer drops/creates the real, shared `#4262 §4` candidate index at all (it now
  owns a separate `bench_`-prefixed index for its with-index comparison pass); and all
  seeding/measurement/cleanup runs inside one outermost `try/finally` so a mid-run crash cannot strand
  seeded rows. See the harness's own header comment for the full contract.
- **P2 (MD honesty):** the results MD now pins the exact SHA measured (`86fa1d85c`, before `#4269`
  landed), states p50 **and** p95 everywhere a bare p50 previously stood in, labels the 10k/50k preview
  cells as 413-refusal-only, labels the ~20s/~3.6min execute figures as linear extrapolations, and adds
  a "Re-run plan" section.
- **This means two lines below are now stale, kept for the historical record rather than edited in
  place:** the "drop/recreate methodology" note in the Headline findings, and the "DB left in the same
  state it was found" Test-plan line — both described the harness's methodology *before* this update;
  see the MD's own §3/§8 for what changed and why. The two already-reported runs (`r14c1`, `r14c2b`)
  did do what those lines say; a re-run will not reproduce that specific step.

## Post-review round 5 (2026-07-14) — SQL injection + resource-attribution (merge blockers)

Owner round-5 REQUEST_CHANGES found two further merge-blocking gaps in the round-4 hardening above,
plus three honesty residuals. Addressed in the latest commit on this branch:
- **P1-1 (SQL injection):** `BENCH_RUN_ID` was taken from the environment unvalidated and interpolated
  directly into DDL text (the harness-owned index name). `resolveRunIdOrExit()` now requires it to
  match `^[a-z0-9]{1,16}$` (bounded, SQL-metacharacter-free, well under Postgres's 63-byte identifier
  limit) — or auto-generates a safe id in that same charset if unset — and exits fail-closed *before
  any database connection is used* otherwise. No raw env value ever reaches SQL text again.
- **P1-2 (resource attribution / adoption):** `ensureBase()`'s `ON CONFLICT DO NOTHING` meant a
  *repeated* `BENCH_RUN_ID` would silently "adopt" — and later delete — another run's base/user/sheet
  rows. Fixed with four pieces working together: (a) a `pg_try_advisory_lock` keyed on the run id,
  held for the run's lifetime, refuses a genuinely concurrent same-id run; (b) an explicit
  pre-existence assertion refuses to proceed if any base/user/sheet/index id this run would create
  already exists (covers a *non*-concurrent colliding id too, e.g. reused after an earlier crash); (c)
  every creating `INSERT`/`CREATE INDEX` is now plain (no `ON CONFLICT` / `IF NOT EXISTS`) so a
  collision that slips past the assertion is a fatal error, never a silent no-op; (d) cleanup at the
  end only ever deletes ids this run's own inserts *confirmed* they created (tracked strictly after
  each successful insert, never speculatively beforehand).
- **P2-3 (cleanup honesty):** a cleanup-step failure (lock release, index drop, or row delete) is now
  collected and forces a non-zero exit instead of being logged and swallowed into an apparently-clean
  `exit 0` that actually left data behind.
- **P2-4 (MD/PR honesty):** the MD's §6(d) and the "Headline findings" bullet below claiming the
  >5,000 async-job requirement (SC.1) is "confirmed" are reworded to directional-support-only,
  consistent with the hedged framing used everywhere else in both documents.
- **P2-5 (credential leak):** a malformed `DATABASE_URL` error no longer echoes the raw connection
  string (which can contain the password) — only a redacted `scheme://host:port` form, or a fully
  generic message, is ever logged.

See the harness's own header comment ("Run-id exclusivity" and "Cleanup" paragraphs) for the full
contract, and the MD's §0 for a results-doc-facing summary.

## Headline findings (see the MD for full numbers + caveats)

- reset-execute's existing single-transaction pattern is **~10–11x faster per record** than revert-execute's current per-record-transaction loop, and is the only genuinely atomic one of the two — reproduced across two independent runs. **Caveat (added after self-review):** the two paths differ in more than transaction boundary — reset-execute's write is also structurally thinner (raw `UPDATE`, no formula/rollup recompute or relation-aggregation fan-out, vs revert-execute's full `patchRecords` service path). This benchmark cannot attribute the 10–11x to the transaction boundary alone, so "wrap revert-execute's loop in one transaction → ~10x" is *not* a tested claim; the supported claim is narrower — reset-execute's atomic, thin-write, already-shipped shape is the better empirical template for Option A's write path.
- Extrapolating that write cost to 50k records lands around ~20s (atomic thin-write pattern — a floor, since it excludes derived-field recompute a real restore may need) vs ~3.6 minutes (current non-atomic revert-execute pattern) — both lend **directional support only** to the >5,000 async-job requirement (SC.1), not independent confirmation of it, and the existing 413 ceiling refusal is itself fast (never near a timeout) at every tested tier. **(Both figures are linear extrapolations from a single 1k-tier measurement — see the MD's §6(a)/§6(d).)**
- A pre-existing composite index (leftover from a paused, not-on-main W0-1 draft) was measured with a drop/recreate methodology so the main-parity baseline isn't contaminated — the planner never selected it (nor a DESC-matching variant probed on the side) for the reconstructor/precheck/contiguity queries at 10k-per-sheet scale on this dev DB, so this benchmark does not show the #4262 §4 supporting index helping here (flagged as needing a production-scale re-check, not ruled out). **(This drop/recreate-a-shared-index methodology is what the P1 update above replaced going forward — see the MD's §3 safety-update note.)**

## Environment

Local scratch PG only (`postgresql://postgres:pw@localhost:55888/metasheet_test`), default-off recovery flags set in the benchmark **process env only** (`MULTITABLE_ENABLE_SHEET_REVERT`, `MULTITABLE_ENABLE_PIT_RESET`). No production/staging contact.

## Model dispatch

Built and run by a Sonnet agent (harness + execution + write-up), per the owner's model-split policy for benchmark/impl lanes — stated honestly in the MD. No Opus adversarial gate has reviewed this yet. **The round-5 commit on this branch carries a `Co-Authored-By: Claude Opus 4.8` trailer per the task's explicit dispatch instruction; this note does not independently assert which model performed the work, only what the commit trailer says.**

## Test plan

- [x] Harness typechecks/lints consistently with sibling scripts under `packages/core-backend/scripts/` (same pre-existing tsconfig/eslint scoping as `benchmark-messaging.ts`, `gen-dev-token.ts` — not part of the strict project include, run via `tsx`).
- [x] Two full end-to-end runs against the local scratch PG completed cleanly (positive-control assertions held every iteration: reconstruct map size, precheck `ok:true`, expected HTTP status below/above the revert ceiling).
- [x] DB left in the same state it was found (composite index restored; zero leftover seeded rows — verified by direct query after the run). **Describes the two original runs' methodology only — see "Post-review update" above; the harness no longer touches the real shared index at all.**
- [x] Post-review round 4: preflight guard proven to refuse in all 3 required scenarios (missing opt-in, non-local host, non-bench/test db name — the latter two via fabricated unreachable/nonexistent targets, proving pure string-parse before any connection); a real 1k-tier allow-path run left zero leftover rows and the real candidate index untouched; a `BENCH_INJECT_FAULT=1` run (fault injected after seed + real destructive execute + bench-index creation) still left zero leftover rows/index via the outermost `finally`. tsc clean.
- [x] Post-review round 5, against the local scratch PG: an injection-shaped `BENCH_RUN_ID` (`x"; DROP TABLE meta_bases; --`) and a 40-char one both exit fail-closed (code 2) before any DB write is reachable (validation runs before `BASE_ID`/`ACTOR_ID`/any sheet id is even computed); a second run with the same valid id is refused both concurrently (via the advisory lock, while the first run is still in flight) and non-concurrently (via the pre-existence assertion, after the first run's kept-data completed) — in both cases the first run's rows are still present, undeleted, afterward; the plain-`INSERT` statement shape now in `ensureBase()` is confirmed fatal (not silently ignored) on a duplicate key via a direct psql reproduction; a `BENCH_INJECT_CLEANUP_FAULT=1` run (the benchmark itself succeeds, only the cleanup step is made to throw) exits 1, not 0; a malformed `DATABASE_URL` containing a fake embedded password never emits that password in stdout/stderr (grep count 0), for both the generic-fallback and the scheme://host-redaction code paths. A clean 1k-tier positive-control run (auto-generated `BENCH_RUN_ID`, full index-comparison pass included) exits 0 with zero leftover rows/index afterward and the real `#4262 §4` index left untouched (present before and after, unchanged). The pre-existing `BENCH_INJECT_FAULT=1` self-test (round-4) was re-run and still passes after the round-5 refactor. tsc clean.
- [ ] Owner/adversarial review of the analysis section.


zensgit:
**P3 (round-7) addressed — `1aa58fe72`.** Narrowed the two over-strong 'COMMIT error = rolled back' comments: a connection lost AFTER a server-side commit is indeterminate at the client, so the claim is now 'no partial leak of exactly one row' (pre-server-commit failures roll back), with the id-scoped pre-existence assertion as the backstop for the post-commit-lost-connection residual. No logic change; the single-txn atomicity fix stands. (Thanks for the APPROVE-with-P3.)

zensgit:
**P3 (round-8) — `7fbdda589`.** Corrected the still-overstated abort-path wording: there is **no** abort-path recheck (pre-existence runs only before `ensureBase()`; a COMMIT exception rolls back + rethrows, `baseAndUserCreated` stays false, cleanup skips base/user). Accurate residual: a **same-`BENCH_RUN_ID` retry** fail-closes on the pre-existence assertion; an auto-generated new run id would not detect a post-commit-lost-connection leftover — documented as a known, bounded, localhost-only residual. No logic change.

zensgit:
**P3 (round-9) — `575d073fe`.** Both 'localhost-only' sites (`:382` header, `:978` inline) corrected to 'local-only by default; remote execution requires the explicit `BENCH_ALLOW_REMOTE_DB=1` opt-in' — the earlier wording ignored the documented remote-DB opt-in. No logic change. This closes the last P3.

zensgit:
Superseded as a carrier — closing per owner review (2026-08-21), with one precision so this comment does not overclaim:

Unlike the docs drafts (#4186/#4252), this PR carries an artifact that did **not** land elsewhere: the real-scale benchmark harness (`packages/core-backend/scripts/bench/timemachine-scale-bench.ts`). Closing does not lose it — it stays on branch `claude/r14c-timemachine-scale-benchmark-20260714`.

Why close anyway: the branch predates the #4654 closeout semantics and the entire O-2 hardening (#5014/#5018/#5020/#5037), so its harness benchmarks a runtime shape that no longer exists. If a real-scale Time Machine benchmark is wanted, the honest move is to re-cut it against the landed runtime **after enablement** (post-ladder), not to rebase this. Tracked in the delivery record #5027 as future work territory.

PR #4274, opened 2026-07-14
docs(multitable): R14 rev7 — restore-worker identity exception (non-forgeable) + writer-gate contract

zensgit:
**PROPOSED · rev 4** (your product direction accepted; corrected per 5th-round P1s). Docs-only, Draft.

## Product shape (hybrid, honest per path)
Reset ≤5k = the only sync path callable **atomic** (single txn). Revert ≤5k stays **default-OFF** (#4261), **not called atomic** until its outer-txn refactor. >5k/base-wide = Option-B async chunked = **observable, recoverable, explicit-partial**. No cross-sheet long-txn atomicity at any scale.

## rev3 → rev4 (your 5th-round P1s)
- **State enum aligned to the model machine:** `meta_restore_operations.state ∈ {planning, fencing, applying, paused_retryable, done, abandoned_partial}` — removed the drifted `fenced/finalizing/failed` (they'd reintroduce 'ordinary failure releases protection'). Crash/failure ⇒ `paused_retryable` (protection retained); `abandoned_partial` is **owner-explicit accept-partial only**.
- **§6.1 recommendation REVERSED:** the **durable active-operation row** (checked inside every writer's txn) is the authoritative write-block; per-chunk **transaction-scoped advisory barrier** for ordering only — **no long-held coordinator session advisory lock** (vanishes on disconnect, monopolizes a connection).

Carried (rev3): plan→apply TOCTOU closed (re-run W0 precheck + all hashes after all protections, before first mutation); evidence honesty (#4273 measured at `86fa1d85c` pre-#4269; p50+p95; 413-only previews; linear extrapolations; 'supports 5k-ceiling+async, not B-over-A alone'). Costs: B 4–7 pw, A +6–10 pw. Hard gates: #4262 **v3.2** landed+trusted · retention floor · fork reconcile. Depends on #4262 primitive `acquireAutoNumberSheetWriteLock` (v3.2 §3).


zensgit:
**rev5 (`3c8fc5685`) — owner 6th-round:** stale `#4262 v3.1` refs → **v3.3**; added **`abandoned_partial` governance** (required before ratify) — since reaching it permanently releases the write fence over a knowingly-partial base, it's gated like the destructive execute paths: **base-admin permission**, **`confirm:'abandon-partial'` typed-confirm** echoing operation-id + progress digest, and a **durable audit record** (`abandoned_by/at` + digest + reason, retained row + audit-log entry). Non-gated paths (crash/error/timeout/dropped-connection) → `paused_retryable` with protection retained; `abandoned_partial` is never reached implicitly. Your five approved product directions (Option-B / durable-row fence / resume-forward / reads-see-progress / existing-revisioned-config-tiers-only) noted in the header.

zensgit:
**rev6 (`1748f2597`) — owner 7th-round P1.** Re-review THIS SHA. §3 durable-row check alone was insufficient (check-before-fence TOCTOU: writer checks no-active → restore activates → writer writes → frozen plan drifts). **Ordering contract now pinned:** every writer `BEGIN → acquire sheet fence → THEN check active-operation → mutate → COMMIT` (fence BEFORE check). Restore activation acquires ALL sheet fences in sheet-id order → commits the writer-blocking durable state → releases fences; a queued writer re-checks on acquiring the fence and 409s. **Writer-blocking states** = `{fencing, applying, paused_retryable}` = the same set the partial-unique 'one active per base' covers; `{planning, done, abandoned_partial}` don't block. §5 adds the constructed ordering-race golden (mutation: move the check before the fence ⇒ a post-restore write lands on a 'restored' sheet ⇒ reds).

zensgit:
**rev7 (`de03c7337`) — owner 8th-round P1.** Re-review THIS SHA. The apply-phase worker reuses the normal write path (to emit revisions), so it hit the §3 writer-gate and would **409 its own operation**. Fix: a **server-minted, request-unforgeable** `RestoreWorkerContext{operationId,baseId}` threaded in-process by the coordinator (never from request body/headers/query). The gate lets a write past an active op **only if** `ctx.restoreWorker` matches the current active operation AND state=`applying` AND inside the fence; absent/forged/stale/wrong-state ⇒ 409. So the real worker passes, but an external request carrying a real `operationId` **in its body** is still refused (the gate never reads the body). §5 adds 3 goldens (worker passes / forged-body-id 409 [mutation: read id from body ⇒ reds] / non-applying-or-wrong-id 409). **P3 also swept:** v3.3 refs → current; "exactly one in-flight restore" → "one **writer-blocking** operation per base" with parallel-`planning` clarified (planning is outside the partial-unique; `planning→fencing` takes base uniqueness).

zensgit:
**Implementation note captured (round-9, non-blocking — NOT changing the APPROVE'd `de03c7337`).** Per your ratification note: when this lock is implemented, `RestoreWorkerContext` must be created **only by an internal factory** and must **never** be reachable from a request-derived DTO. Recommended to enforce structurally — a **module-private / opaque capability** (e.g. a non-exported branded type or symbol-keyed token minted solely by the restore coordinator) so the type system itself forbids constructing one from request input. Recorded here so the impl honors it; the design lock at `de03c7337` stands as-approved.

zensgit:
Closing as a stale draft per owner review (2026-08-21) — with its future value stated rather than erased.

This design note (restore-worker identity exception, base-wide restore decision lock) sits squarely in **Phase D territory** — deleted-record restore / whole-table resurrect — which the line has *deliberately not chartered*: the decision recorded in the delivery MD (#5027 §6) is to wait for the L6 soak evidence of the enablement ladder before deciding whether Phase D is worth its estimated cost. An open draft here reads like an active carrier; nobody is driving it.

If/when Phase D is chartered, this document is **design input**, not a ratified base: it predates the #4654 closeout semantics and the entire O-2 hardening, so it must be re-derived against the landed runtime. Branch `claude/r14-basewide-restore-decision-lock-20260714` remains.

PR #4293, opened 2026-07-14
test(multitable): W0 field-undelete revision — two-point test wiring + atomicity golden (#4279 follow-up)

zensgit:
Follow-up to the **merged** #4279 (`W0 tail — field-undelete rehydration emits revisions`), addressing the two Medium test-contract gaps from the owner review. Test-only; no runtime behavior change.

## Medium-1 — real-DB test two-point wiring completed
#4279 added `multitable-tombstone-field-rehydrate-revision-realdb.test.ts` to `plugin-tests.yml` (real-DB lane) but **not** to `vitest.config.ts`'s no-DB `exclude`. Without the exclusion the default no-DB job collects the file and it **skip-greens** (`describeIfDatabase` skips silently), which is exactly the skip-green failure mode the repo has hardened against elsewhere (D-1c "two-point wiring: BOTH points or the file silently never runs"). This PR adds both tombstone realdb specs (the `-revision` one and its data-layer predecessor) to the exclude, with the standard rationale comment. Verified: exclude(2) + plugin-tests.yml(1) present.

## Medium-2 — same-txn atomicity golden
The existing spec proved the success path + the zero-row-no-ghost leg, but did **not** force the failure point the owner flagged: a `meta_record_revisions` INSERT failure **after** the rehydration UPDATE has bumped `version`/data. Added an atomicity golden: a scoped `BEFORE INSERT` trigger (source=`'restore'`) makes the rehydration's `recordRecordRevision` INSERT raise, and asserts the **entire** field-undelete rolls back — field definition **not** recreated, record data + version unchanged, **no** config revision (field-recreate/order-shift) and **no** record revision committed — then a trigger-removed **positive control** proves the identical undelete succeeds (rollback was caused by the injected failure, not an inert undelete).

> Note: this real-DB behavior is proven by the `Run multitable real-DB integration` lane in `plugin-tests.yml`, not locally (no local Postgres). The new test mirrors the existing spec's harness + the codebase's scoped-trigger failure-injection pattern.

## Medium-3 — scale (enablement gate, NOT in this PR)
The rehydration emits `recordRecordRevision` **serially per rehydrated record**, and the field tombstone cap defaults to 50,000 rows. The feature flag is default-OFF so this is **not** a live regression, but **before enablement** the emit should become a batch revision helper (or a lower undelete cap + a real scale benchmark). Flagged here as a named enablement prerequisite for the owner.

**Do not merge on green alone — owner-gated.**

### DIFF
--- a/packages/core-backend/tests/integration/multitable-tombstone-field-rehydrate-revision-realdb.test.ts
+++ b/packages/core-backend/tests/integration/multitable-tombstone-field-rehydrate-revision-realdb.test.ts
@@ -250,4 +250,75 @@ describeIfDatabase('W0 tail — field-undelete rehydration emits revisions (real
     expect(diesRevs).toHaveLength(1)
     expect(diesRevs[0]?.action).toBe('create')
   })
+
+  test('atomicity golden: a meta_record_revisions INSERT failure DURING rehydration rolls back the ENTIRE field-undelete — field def NOT recreated, record data + version + config revisions all reverted (no half-applied undelete), then the SAME undelete succeeds once the failure is removed (positive control)', async () => {
+    const s = await freshSheet('atomic')
+    const F = mkFieldId('atomic')
+    await insertField(s, F, 'AtomicVal', 'string', 1)
+    const R = mkRecordId('atomic1')
+    const T0 = new Date(Date.now() - 60_000).toISOString()
+    await insertRecord(s, R, { [F]: 'orig', other: 'y' })
+    await seedCreateRevision(s, R, { other: 'y' }, T0)
+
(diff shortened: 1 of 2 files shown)


zensgit:
Superseded by #4286 (MERGED), which landed the same two fixes — the two-point vitest.config.ts exclusion and the field-undelete same-txn atomicity golden — plus a more complete sibling scan. #4286's atomicity trigger is also correctly scoped by (record_id, source), whereas this PR's trigger fired on all source='restore' inserts, which is a cross-file pollution surface in the shared-DB integration lane. Closing without rebase or cherry-pick per owner review; the scale enablement gate is tracked separately (MULTITABLE_ENABLE_CONFIG_UNDELETE batch-revision prerequisite).

PR #4297, opened 2026-07-15
feat(stock-prep): view-2 diff row detail — values-free /diff/rows drill-down (FE consumer)

zensgit:
## What — view-2 diff row-detail drill-down (values-free-by-construction FE consumer)

Wires the already-landed, tested backend `GET /snapshot-batches/:id/diff/rows` as the per-row drill-down under the diff summary. Read-only; independent of the tenant-scoped WRITE hardening and NOT the OD-W3-1 value-plane. **externalWrite stays false.**

## Final scope (after re-review rounds)

**Values-free BY CONSTRUCTION** (not just whitelist rendering): the service parses `/diff/rows` from `unknown` and validates every field against the backend's REAL shapes — `diffId = stockprep_diff_<16hex>`, enums against the frozen `DIFF_TYPES`/`REVIEW_STATUSES`/`CHANGE_TYPES`, `rowCount` a non-negative int, `keyFingerprint = sha16:<16hex>`. UI-unused envelope fields (reason, snapshot-line ids, path fingerprints, batch ids) are DROPPED at the boundary rather than pattern-matched, so no loose-pattern string survives for a business value to hide in. Closed union types replace the open strings.
- A malformed envelope THROWS `SNAPSHOT_DIFF_ROWS_MALFORMED` → the view's neutral unavailable state (never a silent "no diff rows"); a valid empty `[]` stays a real no-diffs result. rowCount/heldRowCount are recomputed from the RETAINED rows (never a phantom count over an empty table).
- Client/server **vocab tripwire** imports the backend cjs enums LIVE and asserts equality — a backend vocab change reddens CI instead of silently making the client drop new-kind rows.
- The full opaque 22-char `sha16:<16hex>` handle is shown (was a misleading 4-hex slice).

**View**: a lazy, collapsible "View row detail" toggle — rows GET only on first open, re-fetched fresh per batch, 404-soft. All THREE async loaders (batch list, diff summary, row detail) carry a **monotonic seq guard** so a late response for the previous project/batch is dropped (A→B→A deferred-response tests prove the seq is the load-bearing guard, not the id-match).

## Verification
Parser + vocab tests (`bomSnapshotDiff.spec.ts`, wired into integration-guard): valid kept, a value in each retained field drops that row, malformed throws, empty→zero, counts recomputed, vocab matches backend. View specs: lazy, on-demand, values-free (planted values never render), 404-soft, empty, A→B and A→B→A stale-drop. **Mutations**: vocab drift / drop-a-validator / malformed-not-thrown / server-count / drop-a-seq → each RED. 168 stock-prep + bomSnapshotDiff web specs green; vue-tsc clean; rebased on main; `git diff --check` clean. Not armed.

### DIFF
--- a/.github/workflows/integration-guard.yml
+++ b/.github/workflows/integration-guard.yml
@@ -103,6 +103,7 @@ on:
       - 'apps/web/tests/StockPreparationWorkspace.spec.ts'
       - 'apps/web/tests/StockPreparationProjectWorkspaceView.spec.ts'
       - 'apps/web/tests/StockPreparationSnapshotDiffView.spec.ts'
+      - 'apps/web/tests/bomSnapshotDiff.spec.ts'
       - 'apps/web/tests/StockPreparationMappingConfirmView.spec.ts'
       - 'apps/web/tests/StockPreparationUnitConfirmView.spec.ts'
       - 'apps/web/tests/StockPreparationPrepLineView.spec.ts'
@@ -194,6 +195,7 @@ on:
       - 'apps/web/tests/StockPreparationWorkspace.spec.ts'
       - 'apps/web/tests/StockPreparationProjectWorkspaceView.spec.ts'
       - 'apps/web/tests/StockPreparationSnapshotDiffView.spec.ts'
+      - 'apps/web/tests/bomSnapshotDiff.spec.ts'
       - 'apps/web/tests/StockPreparationMappingConfirmView.spec.ts'
       - 'apps/web/tests/StockPreparationUnitConfirmView.spec.ts'
       - 'apps/web/tests/StockPreparationPrepLineView.spec.ts'
@@ -253,4 +255,4 @@ jobs:
       # Consolidated here into the SINGLE unioned list (verified green on both Node 20 and this repo's
       # default Node before landing) — do not reintroduce a second `run:` key; add new specs to this one.
       - name: Run integration web guard specs (targeted)
-        run: pnpm --filter @metasheet/web exec vitest run composition-vocab-mirror multitable-resolver-vocab-mirror integrationErrorCodeLabels fieldHints IntegrationReadSourceConfigPanel IntegrationReadSourceCompositionPanel IntegrationReadSourceCompositionAuthoringPanel readSourceCompositions.service IntegrationWorkbenchView IntegrationWorkbenchRail IntegrationMonitoringSection IntegrationCleaningDatasetSection IntegrationMappingRulesSection IntegrationObjectTemplateSection IntegrationPayloadPreviewSection IntegrationConnectionSection IntegrationBridgeAgentSection IntegrationK3WiseSetupView IntegrationHelpView IntegrationPipelineRunSection IntegrationStockPrepPanel IntegrationExternalWritePanel IntegrationTableActionsPanel IntegrationFieldOptionSyncPanel readSourceModePresets IntegrationReadSourceWizard JsonAssist IntegrationCompositionWizard bridgeAgentConfigCheck IntegrationOptionSetsStructuredEditor optionSetsStructured integrationWorkbench MetaIntegrationFieldRuleAuthoring readSourceTemplateCatalog IntegrationTemplateCatalogPicker StockPreparationWorkspace StockPreparationProjectWorkspaceView StockPreparationSnapshotDiffView StockPreparationMappingConfirmView StockPreparationUnitConfirmView StockPreparationPrepLineView StockPreparationExceptionQueueView StockPreparationDashboardView StockPreparationStageOverview StockPreparationStageStepper --reporter=dot
+        run: pnpm --filter @metasheet/web exec vitest run composition-vocab-mirror multitable-resolver-vocab-mirror integrationErrorCodeLabels fieldHints IntegrationReadSourceConfigPanel IntegrationReadSourceCompositionPanel IntegrationReadSourceCompositionAuthoringPanel readSourceCompositions.service IntegrationWorkbenchView IntegrationWorkbenchRail IntegrationMonitoringSection IntegrationCleaningDatasetSection IntegrationMappingRulesSection IntegrationObjectTemplateSection IntegrationPayloadPreviewSection IntegrationConnectionSection IntegrationBridgeAgentSection IntegrationK3WiseSetupView IntegrationHelpView IntegrationPipelineRunSection IntegrationStockPrepPanel IntegrationExternalWritePanel IntegrationTableActionsPanel IntegrationFieldOptionSyncPanel readSourceModePresets IntegrationReadSourceWizard JsonAssist IntegrationCompositionWizard bridgeAgentConfigCheck IntegrationOptionSetsStructuredEditor optionSetsStructured integrationWorkbench MetaIntegrationFieldRuleAuthoring readSourceTemplateCatalog IntegrationTemplateCatalogPicker StockPreparationWorkspace StockPreparationProjectWorkspaceView bomSnapshotDiff StockPreparationSnapshotDiffView StockPreparationMappingConfirmView StockPreparationUnitConfirmView StockPreparationPrepLineView StockPreparationExceptionQueueView StockPreparationDashboardView StockPreparationStageOverview StockPreparationStageStepper --reporter=dot
--- a/apps/web/src/components/integration/stockPreparation/StockPreparationSnapshotDiffView.vue
+++ b/apps/web/src/components/integration/stockPreparation/StockPreparationSnapshotDiffView.vue
@@ -174,6 +174,82 @@
               <dd class="sp-snap__count-value">{{ entry.value }}</dd>
             </div>
           </dl>
+
+          <!-- View-2 per-row drill-down: lazy, values-free (handles + enums + counts + opaque SHA-16
+               fingerprints only — never a raw path key / drawing number / quantity / unit). -->
+          <div class="sp-snap__rows-wrap">
+            <button
+              type="button"
+              class="sp-snap__rows-toggle"
+              data-testid="stock-prep-snapshot-diff-rows-toggle"
+              :aria-expanded="rowDetailOpen"
+              @click="toggleRowDetail"
+            >
+              {{ rowDetailOpen ? bi('收起逐行明细', 'Hide row detail') : bi('查看逐行明细', 'View row detail') }}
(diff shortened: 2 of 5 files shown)


zensgit:
P2 (batch stale-read) fixed — head `c74f528d5`, rebased on main.

You were right on both counts. `loadRowDetail()` had no request sequence, so a slow rows response for batch A — arriving after switching to B — set `diffRows` to A's data; opening B then saw it already populated and skipped B's fetch. And the old "switching batches" test re-clicked the FIRST select button (batch-beta is incomplete/disabled), so it only ever re-selected batch-alpha — never a real A→B switch.

**Fix:** a monotonic `rowsSeq` ticket. `loadRowDetail` captures the batch id + a fresh seq and commits the result/error ONLY if its seq is still newest; `resetRowDetail` (called on every batch switch) bumps the seq, invalidating any in-flight load for the batch being left.

**Tests (two complete batches + per-batch row data):** (1) a real A→B switch collapses + re-fetches B fresh (shows gamma rows, never alpha's); (2) the discriminator you described — A's rows load is DEFERRED, switch to B and render B, then release A's slow response and assert it is DROPPED (B still shows gamma, never alpha). Mutation (remove the seq check) → the deferred-response test goes RED. 18/18 spec · 150/150 stock-prep web specs · vue-tsc clean. Not armed.

zensgit:
P2 (two more same-type stale reads) fixed — head `20cc09d88`, rebased on main.

You're right — the rowsSeq fix only covered the row-detail loader. The two sibling loaders had the identical bug and committed late responses unconditionally:
- `loadBatches()`: project A's list pending → switch to B → A overwrote B's list;
- `selectBatch()`: batch A's summary pending → switch to B → A overwrote B's summary.

**Fix — the monotonic-ticket pattern applied consistently to all THREE loaders:** `batchListSeq` (bumped per call; committed only if newest AND `projectId` still matches) and `diffSeq` (bumped per call and by `resetDiff` on a project switch; committed only if newest AND `batchId === selectedBatchId`), checked in the **success, catch, and finally** arms — same as `rowsSeq`.

**Tests (deferred-response discriminators):** (1) a slow batch-LIST for project A arriving after switching to B is dropped (B's `sync-B` stays, A's `sync-A` never shows); (2) a slow diff-SUMMARY for batch A arriving after selecting B is dropped (B's `base-of-gamma` stays, A's `base-of-alpha` never shows). Reactive-projectId + two-complete-batch harness. Mutation (remove each seq check) → the corresponding race test goes RED. 20/20 spec · 152/152 stock-prep web specs · vue-tsc clean. Not armed.

zensgit:
Both P2s fixed — head `350daf889`.

**(1) Values-free BY CONSTRUCTION (not just by whitelist rendering).** You're right: the row types were open `| string`/`string[]` and the service only cast the body, so a business value planted in a whitelisted field rendered. And my client enums were wrong (`pending`/`accepted`; 3 diff types). Fixed: the service now parses `/diff/rows` from `unknown` and validates every field against the **real** backend shapes — `diffId = stockprep_diff_<16hex>`, `snapshot-line id = stockprep_snapshot_line_<16hex>`, `fingerprint = sha16:<16hex>` (read from the generators, not guessed — my fixtures had the wrong shapes), enums against the frozen `DIFF_TYPES`/`REVIEW_STATUSES`/`CHANGE_TYPES`, `rowCount` a non-negative int, `reason` a snake_case code. Any row failing any check is **dropped whole**; closed union types replace the open strings.

- **Client/server vocab tripwire** (`bomSnapshotDiff.spec.ts`) imports the backend cjs enums **live** and asserts equality — so a backend vocab change reddens CI instead of silently making the client drop new-kind rows. Wired into integration-guard (path filters + vitest list). Mutation (drift a client enum) → tripwire RED.
- **Parser tests**: valid row (real shapes) kept; a value planted in EACH whitelisted field drops that row and never appears in output; mixed valid+invalid keeps only valid. Mutation (drop a field validator) → its drop test RED.

**(2) Seq guards now actually proven.** You were right — removing batchListSeq/diffSeq still passed the old A→B tests because the id-match check also catches A→B. Added **A→B→A** deferred tests: the stale first-A load, released after returning to A (id is A again → id-check passes), must be dropped by the **seq**. Mutation (keep id-check, remove seq) → both A→B→A tests RED.

170 stock-prep + bomSnapshotDiff web specs green; vue-tsc clean; diff clean. Not armed.

zensgit:
Three re-review items fixed — head `b7d69000c`.

**P2 (values-free still on loose regex).** You're right — REASON_RE/HANDLE_RE were patterns that `mat_001_secret` / `MAT-001-SECRET` / `CUSTOMER_PROJECT_42` survive (not a live DOM leak since neither reason nor the envelope ids render, but the contract didn't hold). Fixed by **dropping the UI-unused fields entirely** (reason, snapshot-line ids, path fingerprints, snapshotBatchId/baseSnapshotBatchId) instead of pattern-matching them. The client row now carries ONLY what the UI renders, and every retained field has a **tight** validator: diffId `stockprep_diff_<16hex>`, diffType/reviewStatus/changeTypes enums, rowCount non-negative int, keyFingerprint `sha16:<16hex>`. No loose pattern remains.

**P2 (malformed → "no diff rows").** A malformed envelope (not an object, or `rows` not an array) now **THROWS** `SNAPSHOT_DIFF_ROWS_MALFORMED` → the 404-soft catch shows the neutral unavailable state; a valid empty `rows: []` stays a real no-diffs result. rowCount/heldRowCount are **recomputed from the retained rows**, so the count can never be non-zero over an empty table.

**P3 (fingerprint truncated to 4 hex).** `slice(0,10)` cut `sha16:<16hex>` to `sha16:0123`. Now the full opaque 22-char `sha16:<16hex>` handle is shown; fixtures use the real prefixed shape.

Verification: parser tests (real backend shapes) — valid kept, a value in each retained field drops that row, malformed throws, empty→zero, counts recomputed; vocab tripwire imports the backend cjs enums live. Mutations: vocab drift / drop-a-validator / malformed-not-thrown / server-count → each RED. 168 web specs green; vue-tsc clean; diff clean. Not armed.

zensgit:
Re-review of exact head `b7d69000c`: **REQUEST_CHANGES (1 P2; no P1).**

**[P2] Row-level malformed data is still converted into an empty/partial success, and the parser overwrites an intentional server count contract.**

`parseStockPreparationSnapshotDiffRowsResult()` maps invalid rows to `null`, filters them out, then recomputes both counts from the retained subset. The new test explicitly locks `rows: [42, null, "x"]` to `[]`, and a valid+invalid response to a partial table. Therefore a response containing only malformed/value-bearing rows renders **“No diff rows”** instead of the coarse unavailable state; a mixed response silently hides rows. This is fail-closed for disclosure, but fail-open for decision correctness.

There is a second contract effect in the same block: the backend documents and implements `heldRowCount` as the held count over the **whole pair before filters** (`stock-preparation-snapshot-reads.cjs`), while the client recomputes it from filtered/retained rows. A future `reviewStatus=ready` call will incorrectly report zero held context.

Required closure:
1. Parse rows all-or-nothing: if any row fails the tight parser, throw the fixed `SNAPSHOT_DIFF_ROWS_MALFORMED` token so the view shows unavailable; keep `rows: []` as the only true empty result.
2. Validate and preserve the bounded server `rowCount`/`heldRowCount` contract (or remove the unused filter surface); do not silently redefine pre-filter context.
3. Add discriminating tests for valid+invalid and all-invalid arrays, plus a filtered response whose server `heldRowCount` is nonzero.
4. Update the PR body, which still says “10-char fingerprint”, 6 tests, and 149 specs although the current implementation renders the full 22-char handle and has broader coverage.

Verified good at this head: the three monotonic loader guards (including A→B→A), exact `diffId`/`sha16` shapes, closed vocab tripwire, full fingerprint rendering, CI wiring, targeted Vitest `36/36`, and `vue-tsc` clean.

zensgit:
Owner exact-head re-review (`d5ed5d787`): **one P2 remains; do not merge yet.**

**[P2] An invalid row member still fails open as a clean empty or partial result.**

`parseStockPreparationSnapshotDiffRowsResult()` maps every member through `parseDiffRow()` and silently filters `null` entries. Therefore:

- an all-invalid `rows` array becomes `{ rowCount: 0, rows: [] }`, and the UI renders “No diff rows for this batch”;
- a mixed valid/invalid response renders a partial table and recomputed counts, with no indication that evidence was discarded.

This closes the confidentiality path but not the integrity/completeness path. It is the same false-empty class that the malformed-envelope fix was intended to close. The current tests explicitly lock the fail-open behavior (`drops only the invalid row`, junk members become `[]`, and a 9-row claim can become one retained row).

Required correction: if **any** member of `rows` is invalid, reject the whole response with the existing fixed coarse token so the component enters the neutral unavailable state. Preserve only one empty success case: a genuinely valid `rows: []`. Add discriminating tests for all-invalid and mixed valid/invalid arrays, plus a mutation that restores filtering and must fail.

The three loader sequence guards, A→B→A coverage, frozen vocabulary tripwire, and fresh CI otherwise check out at this head.

zensgit:
P2 (silent drop) fixed — head `965e305d8`, rebased. The parser now **fail-closes at the envelope**: a malformed envelope OR **any single invalid row** throws `SNAPSHOT_DIFF_ROWS_MALFORMED` → the view's coarse unavailable state. No per-row drop remains, so an all-invalid response can't show as "No diff rows" and a mixed one can't show as a silently-incomplete table. Only a genuinely valid empty `rows: []` yields the empty state; all-valid → all kept. Tests updated (value in any field → throws; mixed → throws; junk rows → throws; empty → empty); mutation (revert to per-row drop) → 9 RED. 168 web specs green, vue-tsc clean. Not armed.

zensgit:
Owner exact-head re-review at `3127c3050`: **APPROVE, 0 P1 / 0 P2.** The previous blocking comment applied to `d5ed5d787` and is superseded by the fail-closed fix in `965e305d8`; `3127c3050` only corrects the two stale comments.

Independent verification: baseline parser suite 14/14; mutating `if (parsed === null) throw` back to per-row `continue` made 9/14 tests fail (all-invalid, mixed valid/invalid, junk members, and every retained-field sentinel); restoring the exact head returned 14/14 and a clean worktree. Valid `rows: []` remains the only empty success.

Fresh CI is currently running on the comment-only head. This PR is safe to squash-merge once required checks are green.

PR #4239, opened 2026-07-13
docs: approval-line completion — design & verification plan (2026-07-13)

zensgit:
审批/流程线收官目标的 **设计及验证 MD**。

基于一轮独立的 verify-and-plan（对抗式复核）：

- **已核实的可靠性缺口**：完成事件 → 投递链路存在崩溃丢写窗口（需事务性 outbox + durable 租约派发）；FWB-2 的 `record-link` 字段类型今天不存在；W7 结果写回入版本历史已在 main 落地。
- **7 阶段有序计划**：按难度分派模型（fable5 / sonnet5 / opus4.8），含并行与依赖边。
- **验证口径**：崩溃注入矩阵、强制正控腿、真库 golden、全链场景矩阵。

设计细节以文档为准。规划性文档 — 每条 runtime 车道均受 owner-ratify + deploy 门约束，无任何自 ratify、未触碰 runtime；每条车道在独立隔离 worktree 上经独立对抗门禁落地。

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4303, opened 2026-07-15
feat(automation): P2 durable-delivery S1 — outbox schema + default-OFF flag (additive, zero behavior)

zensgit:
**First runtime slice of the RATIFIED durable-delivery design** (FWB-0 lock #4203 Layer 1 + approval-line plan #4239), authorized after all four design locks landed on main. **Additive only — zero behavior change.** No existing table is altered; no code path reads or writes the new tables while `AUTOMATION_DURABLE_DELIVERY_ENABLED` is OFF (the default).

## What lands
- **Migration** `zzzz20260715120000_create_automation_outbox`:
  - `meta_automation_outbox` — producer event rows, to be co-committed with the source state change (approval status / record write) by S4. `manifest_version` stamps the routing manifest that expanded the row (a v1 row is dispatched per v1 forever). `event_id` is **NOT NULL and non-blank** — the stable original-event identity the dispatcher forwards downstream as the per-rule dedup key and the outbound-idempotency seed (#4203 §producer/identity). A CHECK `event_id ~ '[!-~]'` requires a printable-ASCII non-space char, so any id the dedup helper (`automation-event-dedup.ts`: `_eventId.trim()`) would collapse to no-identity is rejected — including Unicode whitespace (NBSP/EM/narrow-NBSP/BOM), which PostgreSQL's ASCII-only `[:space:]` would have let through. Deliberately **not unique** (cutover tolerates transient double-emit and dedups at the sink, not at the outbox).
  - `meta_automation_outbox_consumer` — per-`(outbox_id, consumer_key)` delivery state: `status` is the **four**-state machine `pending|in_progress|done|dead_letter` via a named CHECK — there is **no persistent `failed`** state (a transient failure only bumps `attempts` and leaves the row reclaimable via lease expiry; bounded-attempts-exhausted → terminal `dead_letter`, so no state can stick). `lease_expires_at` is tied to `status='in_progress'` by a **biconditional** CHECK (in_progress ⇒ leased, else unreclaimable; NOT in_progress ⇒ no lease — not because the reclaim scan would pick it up, that scan is status-scoped, but to keep a lease meaning exactly "owned by a live in_progress worker", so terminal/reclaim transitions must clear it atomically). `fence` is `bigint` (zombie fencing); `attempts` incremented at claim; `fence`/`attempts`/`automation_depth`/`manifest_version` carry named non-negative / positive CHECKs. FK cascade from the outbox event.
- **`automation-durable-delivery.ts`** — `isDurableDeliveryEnabled` (default OFF, robust trim/case read mirroring `MULTITABLE_ENABLE_PIT_RESET`), the row types (`fence` is a **`string`** — node-postgres returns bigint as a string; S2's CAS stays in decimal-string space), and `RESOLVE_PERMITTING_STATUSES = {done, dead_letter}`.
- **Real-DB golden** — the migration landed both tables + columns; the status CHECK enforces the **four**-state machine (four valid inserts + a bogus status rejected); the lease **biconditional** is exercised on all four corners; `event_id` NOT NULL and the `manifest_version`/`automation_depth`/`fence`/`attempts` CHECKs each have a negative control that asserts the **specific constraint name** in the rejection (so deleting exactly one constraint reddens exactly its test); `fence` round-trips as a **string past 2^53** (no JS-number precision loss); plus FK cascade, column defaults, and the flag OFF by default with a positive control that only exactly-`true` enables. **Two-point wired** (plugin-tests.yml real-DB run + vitest.config.ts exclude) so it cannot skip-green.

## Not in this slice (subsequent P2 slices)
S2 durable dispatcher (awaited-adapter + fence-CAS claim), S3 versioned routing manifest + startup completeness assertion + rolling-deploy, S4 producer-side atomic enqueue replacing the commit-then-emit points, S5 consumer adapters + poison-terminal, S6 upgrade-migration backfill, S7 crash-injection V-series. Each behind the same default-OFF flag.

**FWB, attachment, and every related flag stay OFF** — this slice ships no behavior. **For review — not self-merging runtime.**

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -507,6 +507,7 @@ jobs:
             tests/integration/multitable-tombstone-field-rehydrate-revision-realdb.test.ts \
             tests/integration/multitable-tombstone-record-capture-realdb.test.ts \
             tests/integration/multitable-tombstone-retention-realdb.test.ts \
+            tests/integration/multitable-automation-outbox-schema-realdb.test.ts \
             tests/integration/multitable-attachment-readgate.security.test.ts \
             tests/integration/multitable-ai-write-provenance-batch-grouping-realdb.test.ts \
             tests/integration/multitable-automation-branch-local-wait.test.ts \
--- a/packages/core-backend/src/db/migrations/zzzz20260715120000_create_automation_outbox.ts
+++ b/packages/core-backend/src/db/migrations/zzzz20260715120000_create_automation_outbox.ts
@@ -10,11 +10,14 @@
  *                                        the source state change (approval status write / record write).
  *                                        `manifest_version` stamps the routing manifest that expanded it
  *                                        (a v1 row is dispatched per v1 forever — never re-interpreted).
- *                                        `event_id` is NOT NULL — the stable ORIGINAL-event identity the
- *                                        dispatcher forwards downstream as the per-rule dedup key and the
- *                                        outbound-idempotency seed (#4203 §producer/identity). It is NOT
- *                                        unique: #4203 §cutover tolerates transient double-emit and dedups
- *                                        at the sink, so a second contract-legal enqueue must be allowed.
+ *                                        `event_id` is NOT NULL AND non-blank — the stable ORIGINAL-event
+ *                                        identity the dispatcher forwards downstream as the per-rule dedup
+ *                                        key and the outbound-idempotency seed (#4203 §producer/identity). A
+ *                                        CHECK rejects empty/all-whitespace ids, mirroring the dedup helper
+ *                                        (`automation-event-dedup.ts`: `_eventId.trim()` → no-identity), so a
+ *                                        blank id can't sit in the outbox looking like an identity it isn't.
+ *                                        It is NOT unique: #4203 §cutover tolerates transient double-emit and
+ *                                        dedups at the sink, so a second contract-legal enqueue is allowed.
  *   - `meta_automation_outbox_consumer`— per-(outbox_id, consumer_key) delivery state. `status` is the
  *                                        four-state machine `pending|in_progress|done|dead_letter` — there is
  *                                        NO persistent `failed` state: a transient failure only bumps
@@ -30,8 +33,13 @@
  *                                        lease is EXACTLY the in_progress marker — a BICONDITIONAL CHECK ties
  *                                        `lease IS NOT NULL` to `status='in_progress'` in BOTH directions:
  *                                        in_progress ⇒ leased (else unreclaimable), and non-in_progress ⇒
- *                                        NO lease (a stale lease on a done/pending row would be spuriously
- *                                        reclaimed). Fence/attempts/depth carry non-negative CHECKs.
+ *                                        NO lease. The reverse half is NOT about the reclaim scan (that scan
+ *                                        is status-scoped, so it would never pick up a leftover lease on a
+ *                                        settled row) — it keeps a lease meaning EXACTLY "owned by a live
+ *                                        in_progress worker", so every terminal/reclaim transition must clear
+ *                                        the lease atomically and a settled row can never carry stale
+ *                                        ownership. Fence/attempts/depth carry non-negative CHECKs; `event_id`
+ *                                        is non-blank (mirrors the dedup helper's `.trim()` → no-identity).
  *
  * NOTHING reads or writes these tables yet — the durable dispatcher (S2), the producer-side atomic enqueue
  * (S4), and the consumer adapters (S5) are separate, flag-gated slices. This migration is byte-for-byte
@@ -56,7 +64,13 @@ export async function up(db: Kysely<unknown>): Promise<void> {
                          CONSTRAINT automation_outbox_depth_nonneg CHECK (automation_depth >= 0),
       manifest_version int  NOT NULL
                          CONSTRAINT automation_outbox_manifest_version_positive CHECK (manifest_version >= 1),
-      event_id         text NOT NULL,
+      event_id         text NOT NULL
+                         -- reject empty / all-whitespace ids: the dedup helper (automation-event-dedup.ts)
+                         -- does _eventId.trim() and treats a blank result as NO identity, so a blank id in
+                         -- the outbox would be un-dedupable. [^[:space:]] = at least one non-whitespace char
+                         -- (POSIX class covers space/tab/newline/CR/VT/FF, mirroring JS .trim() over ASCII;
+                         -- a regex, not btrim(), so no backslash escapes leak into this template literal).
+                         CONSTRAINT automation_outbox_event_id_nonblank CHECK (event_id ~ '[^[:space:]]'),
       created_at       timestamptz NOT NULL DEFAULT now()
     )
   `.execute(db)
@@ -85,8 +99,12 @@ export async function up(db: Kysely<unknown>): Promise<void> {
       -- The lease is EXACTLY the in_progress marker — a BICONDITIONAL, not a one-directional guard:
       --   in_progress  => lease present  (else the reclaim scan status=in_progress AND lease < now()
       --                                   could never recover it — a permanently stuck row); AND
-      --   NOT in_progress => lease absent (a stale lease left on a pending/done/dead_letter row would be
-      --                                    spuriously matched by the same reclaim scan).
+      --   NOT in_progress => lease absent. The reason is NOT that the scan would reclaim it — that scan is
+      --                                    status-scoped and would never pick up a leftover lease on a
+      --                                    pending/done/dead_letter row. It is to keep a lease meaning
+      --                                    EXACTLY "owned by a live in_progress worker": every terminal /
+      --                                    reclaim transition must clear the lease atomically, so a settled
+      --                                    row can never carry stale, ambiguous ownership.
       -- Every state transition (claim/complete/reclaim/poison) sets or clears the lease in the SAME
       -- single-row UPDATE that changes the status, so the biconditional holds at every commit boundary.
       CONSTRAINT automation_outbox_consumer_lease_iff_in_progress
--- a/packages/core-backend/src/db/migrations/zzzz20260715120000_create_automation_outbox.ts
+++ b/packages/core-backend/src/db/migrations/zzzz20260715120000_create_automation_outbox.ts
@@ -1,0 +1,72 @@
+/**
+ * Migration: `meta_automation_outbox` + `meta_automation_outbox_consumer` — P2 durable-delivery substrate.
+ *
+ * FWB-0 design lock (#4203, Layer 1) + approval-line plan (#4239): the automation completion→delivery
+ * chain today has two crash-loss windows (commit-then-emit to an in-memory `eventemitter3` bus; a
+ * terminal claim-before-execute tombstone). This migration lands the **additive, unused-until-wired**
+ * substrate for the durable fix:
+ *
+ *   - `meta_automation_outbox`         — one row per producer event, written in the SAME transaction as
+ *                                        the source state change (approval status write / record write).
+ *                                        `manifest_version` stamps the routing manifest that expanded it
+ *                                        (a v1 row is dispatched per v1 forever — never re-interpreted).
+ *   - `meta_automation_outbox_consumer`— per-(outbox_id, consumer_key) delivery state. `status` is the
+ *                                        four-state machine `pending|in_progress|done|dead_letter` plus a
+ *                                        `failed` transient; `fence` is the monotonic epoch bumped on every
+ *                                        claim/reclaim (a stale-fence "zombie" write hits 0 rows, §Layer-1
+ *                                        fencing); `attempts` is incremented AT CLAIM (so an always-crash-
+ *                                        after-claim poison still reaches `dead_letter`); `lease_expires_at`
+ *                                        is the reclaim lease.
+ *
+ * NOTHING reads or writes these tables yet — the durable dispatcher (S2), the producer-side atomic enqueue
+ * (S4), and the consumer adapters (S5) are separate, flag-gated slices. This migration is byte-for-byte
+ * behavior-neutral: no existing table is altered, no code path touches these tables while
+ * `AUTOMATION_DURABLE_DELIVERY_ENABLED` is OFF (the default). Additive `CREATE TABLE IF NOT EXISTS` only.
+ */
+import { sql, type Kysely } from 'kysely'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  // Producer event rows — co-committed with the source state change.
+  await sql`
+    CREATE TABLE IF NOT EXISTS meta_automation_outbox (
+      id               text PRIMARY KEY,
+      event_type       text NOT NULL,
+      payload          jsonb NOT NULL,
+      automation_depth int  NOT NULL DEFAULT 0,
+      manifest_version int  NOT NULL,
+      event_id         text,
+      created_at       timestamptz NOT NULL DEFAULT now()
+    )
+  `.execute(db)
+  // Dispatcher polls pending rows oldest-first.
+  await sql`
+    CREATE INDEX IF NOT EXISTS idx_automation_outbox_created_at
+    ON meta_automation_outbox (created_at)
+  `.execute(db)
+
+  // Per-(outbox_id, consumer_key) delivery state — the reclaimable lease row (fence-bearing).
+  await sql`
+    CREATE TABLE IF NOT EXISTS meta_automation_outbox_consumer (
+      outbox_id        text NOT NULL REFERENCES meta_automation_outbox(id) ON DELETE CASCADE,
+      consumer_key     text NOT NULL,
+      status           text NOT NULL DEFAULT 'pending'
+                         CHECK (status IN ('pending','in_progress','done','failed','dead_letter')),
+      lease_expires_at timestamptz,
+      fence            bigint NOT NULL DEFAULT 0,
+      attempts         int    NOT NULL DEFAULT 0,
+      last_error       text,
+      updated_at       timestamptz NOT NULL DEFAULT now(),
+      PRIMARY KEY (outbox_id, consumer_key)
+    )
+  `.execute(db)
+  // Dispatcher claim scan: reclaimable rows are pending, or in_progress past a stale lease.
+  await sql`
+    CREATE INDEX IF NOT EXISTS idx_automation_outbox_consumer_claim
+    ON meta_automation_outbox_consumer (status, lease_expires_at)
+  `.execute(db)
+}
+
+export async function down(db: Kysely<unknown>): Promise<void> {
+  await sql`DROP TABLE IF EXISTS meta_automation_outbox_consumer`.execute(db)
+  await sql`DROP TABLE IF EXISTS meta_automation_outbox`.execute(db)
+}
--- a/packages/core-backend/src/db/migrations/zzzz20260715120000_create_automation_outbox.ts
+++ b/packages/core-backend/src/db/migrations/zzzz20260715120000_create_automation_outbox.ts
@@ -10,6 +10,11 @@
  *                                        the source state change (approval status write / record write).
  *                                        `manifest_version` stamps the routing manifest that expanded it
  *                                        (a v1 row is dispatched per v1 forever — never re-interpreted).
+ *                                        `event_id` is NOT NULL — the stable ORIGINAL-event identity the
+ *                                        dispatcher forwards downstream as the per-rule dedup key and the
+ *                                        outbound-idempotency seed (#4203 §producer/identity). It is NOT
+ *                                        unique: #4203 §cutover tolerates transient double-emit and dedups
+ *                                        at the sink, so a second contract-legal enqueue must be allowed.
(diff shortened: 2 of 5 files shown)


zensgit:
Review verdict: **REQUEST CHANGES — do not merge or start S2 on this schema yet.** All checks are green, and the default-OFF/no-behavior claim is accurate, but the foundational state contract has two load-bearing defects.

1. **[P1] `failed` is an absorbing non-terminal state.** The migration calls it transient, while the documented claim scan only reclaims `pending` or expired `in_progress` rows (`zzzz20260715120000_create_automation_outbox.ts:62`), and `RESOLVE_PERMITTING_STATUSES` excludes it (`automation-durable-delivery.ts:27-31`). A row moved to `failed` can therefore neither retry nor resolve. This also conflicts with the ratified #4203 contract, which explicitly makes `failed`/`dead_letter` resolve-permitting terminal outcomes. Pick one coherent machine before the migration lands: either remove consumer-level `failed` and use `pending` for retryable failures plus `dead_letter` for exhausted poison, or make `failed` terminal and include it in the resolve contract. Add a real-DB transition golden proving no accepted status can wedge forever.

2. **[P2] The fencing type is already false at the shared boundary.** PostgreSQL `bigint` is returned by `pg` as a string; the real-DB test itself asserts `fence: '0'` (`...schema-realdb.test.ts:108`), but `OutboxConsumerRow.fence` is declared `number` (`automation-durable-delivery.ts:50`). S2 arithmetic/CAS built on that type can concatenate, truncate above `Number.MAX_SAFE_INTEGER`, or compare the wrong token. Model the raw value as `string` and parse/use `bigint` in a dedicated helper (or otherwise prove a safe bounded representation), then add a value-above-2^53 positive control and a stale-fence CAS mutation.

3. **[P2] The DB accepts impossible lease states.** The current positive-control loop inserts `status='in_progress'` with `lease_expires_at=NULL`; that row is valid under the CHECK but is not reclaimable by the documented scan. Add a state/lease consistency CHECK (at minimum, `in_progress` requires a lease; non-active terminal states clear it), plus non-negative constraints for `fence`, `attempts`, and `automation_depth`, and a positive `manifest_version` constraint. These are cheapest and safest before the additive migration reaches main.

4. **[P3] Test comment drift.** Lines 21-22 say a real `automation_rules` row is required, but no rule row/FK is used. Remove the false premise.

Re-review bar: revised transition table in code/comments, real-DB negative and positive controls for every state/lease combination, fence representation/CAS proof, migration replay, and the existing two-point CI wiring. Keep the flag OFF. S2 should wait for this re-review.

zensgit:
Addressed all four review findings and rebased onto current main (`40854cc0f`):

- **P1 stuck `failed` state** — removed `failed` from the persistent machine entirely. States are now `pending|in_progress|done|dead_letter`; a transient failure only bumps `attempts` and stays reclaimable (lease expiry → next claimer), bounded-attempts-exhausted → terminal `dead_letter`. Dropped from the status CHECK, the `OutboxConsumerStatus` type, and the test's valid-status loop. No non-terminal state is unreclaimable; the two terminals equal `RESOLVE_PERMITTING_STATUSES`.
- **P2 `fence` bigint typed `number`** — now `string` on `OutboxConsumerRow` (node-pg returns bigint as string; the golden already asserts `fence: '0'`). Doc note added so S2's increment/CAS treats it as a decimal string, never coerces through a JS number.
- **P2 `in_progress` + NULL lease** — added `CHECK (status <> 'in_progress' OR lease_expires_at IS NOT NULL)` + non-negative CHECKs on fence/attempts/automation_depth and manifest_version >= 1. Real-DB **negative control** (NULL-lease in_progress rejected) + **positive control** (leased in_progress accepted) + negative-fence rejection added; FK-cascade fixture updated to carry a lease.
- **P3 misleading comment** — removed the `automation_rules`-row claim (never used).

Core-backend typecheck green locally. Flag `AUTOMATION_DURABLE_DELIVERY_ENABLED` stays OFF; migration is still additive/behavior-neutral. Not starting S2 — awaiting your re-review.

zensgit:
CI is now fully green on the rebased head. All required checks pass (`test (20.x)`, `web-tests`, `contracts ×3`, `pr-validate`), plus `migration-replay` (the migration replays idempotently) and `e2e`.

Verified the real-DB golden actually **executed** (not skip-greened): the `Plugin System Tests` run-list includes the spec and the vitest log shows `✓ tests/integration/multitable-automation-outbox-schema-realdb.test.ts (9 tests) 52ms` against real Postgres — the 7 original goldens plus the two new controls (NULL-lease `in_progress` rejected / leased `in_progress` accepted; negative `fence` rejected).

Ready for your re-review. Not merging, not starting S2; flag stays OFF.

zensgit:
Round-2 re-review verdict: **REQUEST CHANGES — the original four findings are fixed, but S1 is not yet merge-ready. Keep S2 stopped.**

Verified on the live PR head `40854cc0f`:
- the persistent `failed` state is gone and the four-state machine is coherent;
- `fence` is now represented as a decimal string;
- the one-way `in_progress` lease guard and numeric CHECKs exist;
- the stale fixture comment is removed;
- the real-DB file actually ran: CI log shows `multitable-automation-outbox-schema-realdb.test.ts (9 tests)`, 9/9 on Postgres.

Remaining findings:

1. **[P2] `event_id` still permits the durable path to lose the stable source identity.** The migration leaves `event_id text` nullable and `OutboxRow.eventId` is `string | null`. The ratified #4203 producer/identity contract requires the dispatcher to forward the original `_eventId` (or an equivalent stable source-event key) so cutover double-delivery collapses at the sink and outbound idempotency stays stable across fence/attempts. Every currently enumerated source family already has an identity (`eventId` for approval/task events, `_eventId` for record/form events), and no rows are written before S4, so there is no backfill reason to make this nullable. Make it `NOT NULL` (preferably non-blank), make the type non-null, and add reject-null/accept-stable-key real-DB controls. Otherwise S4 can legally create an outbox row that cannot satisfy the ratified net-once contract.

2. **[P2] The lease-state invariant is still only half enforced.** `CHECK (status <> 'in_progress' OR lease_expires_at IS NOT NULL)` rejects active-without-lease, but still accepts `pending`, `done`, or `dead_letter` with a lease. The prior review explicitly required non-active/terminal states to clear the lease; enforcing the biconditional now forces S2's terminal CAS to clear it atomically and prevents stale operational state. Use an equivalent of `(status = 'in_progress') = (lease_expires_at IS NOT NULL)` and pin the full 4-state × lease-present/absent matrix.

3. **[P2-test] Three newly load-bearing numeric guards are invisible to the golden.** The test only mutation-discriminates `fence >= 0`; removing `attempts >= 0`, `automation_depth >= 0`, or `manifest_version >= 1` still leaves all 9 tests green. Add parameterized negative controls plus positive boundary controls. Also pin a fence above `2^53` round-tripping as the exact decimal string; the stale-fence CAS race remains correctly deferred to S2, where a CAS implementation exists.

4. **[P3] Public review metadata is stale.** The PR body still says `(+ failed transient)` and `5 valid inserts`, contradicting the revised four-state schema. Update it before landing.

State/provenance note: the SHA reported to me (`4ac3967d3`) is only local and is not GitHub-reachable. Its three load-bearing files are byte-identical to live head `40854cc0f`, so the code review above is valid, but CI and the PR are attached to `40854cc0f`. The live PR is now `BEHIND` (`base 98bf851d`, current main `818e5b91c`), so rebase and rerun after the fixes.

Re-review bar: the two schema changes, discriminating real-DB controls, body reconciliation, live remote-SHA verification, migration replay, and all required checks green. Flag remains OFF; do not begin S2 yet.

zensgit:
Round-2 findings all addressed; rebased onto current main (`05223f9e7`).

- **event_id NULLable → NOT NULL.** It is the stable original-event identity the dispatcher forwards downstream as the per-rule dedup key + outbound-idempotency seed (#4203 §producer/identity). Kept it **NOT unique on purpose**: #4203 §cutover (lines ~318-321) explicitly tolerates transient double-emit and dedups at the sink, not at the outbox — a `UNIQUE(event_id)` would reject the second, contract-legal enqueue. Documented that rationale inline.
- **Half lease constraint → biconditional.** Replaced `status<>'in_progress' OR lease IS NOT NULL` with `CHECK ((lease_expires_at IS NOT NULL) = (status = 'in_progress'))` — the lease is *exactly* the in_progress marker, so a stale lease on a pending/done/dead_letter row is now rejected too.
- **Deletable CHECKs + missing 2^53 proof.** Every CHECK is now **named**, and each has a negative control that asserts its **specific constraint name** in the rejection (deleting one constraint reddens exactly its test — not a blanket `.rejects.toThrow()`). Added the biconditional's four corners and a `fence = 9007199254740993` (2^53+1) round-trip proving the exact value survives as a string.
- **PR body** rewritten — no more `failed`-transient / 'five valid states'.

Rehearsed locally on a **fresh real Postgres 15**: the migration DDL applies + replays idempotently, and all **15** constraint / biconditional / bigint probes pass, each rejection carrying the asserted constraint name (incl. `done`+lease and `pending`+lease now rejected by `..._lease_iff_in_progress`). Core-backend typecheck green.

Re-greening CI now; flag stays OFF, not starting S2, no self-merge — back to you for re-review.

zensgit:
CI is fully green on `6a74a4324` — all six required checks (`test (20.x)`, `web-tests`, `contracts ×3`, `pr-validate`) plus `migration-replay`, `e2e`, `after-sales`.

**Real-DB spec actually executed** (triggered ≠ verified): the `test (20.x)` real-DB log shows `✓ …outbox-schema-realdb.test.ts (13 tests) 79ms` against real Postgres — 13, up from the previous 9, so the new controls ran: event_id NOT NULL, the lease biconditional's four corners, per-constraint negative controls asserting the specific constraint name, and the `9007199254740993` (2^53+1) fence string round-trip.

Full disclosure: my first round-2 push briefly reddened five jobs — a backtick inside a SQL comment sat *inside* the `sql`…`` template literal and broke TS parsing (so both the typecheck step and every tsx-loaded `db:migrate` job failed). Fixed by removing the backticks; re-greened. (My earlier local typecheck used a project config that excludes `src/db/migrations`, so it missed it — I now run CI's `tsc --noEmit`.)

**One judgment call to flag for your ratify/redirect:** I made `event_id` **NOT NULL but deliberately NOT unique**. #4203 §cutover (≈L318-321) states the cutover tolerates transient double-emit and dedups at the sink, not at the outbox — so a `UNIQUE(event_id)` would reject a second, contract-legal enqueue. If you intended outbox-level uniqueness instead, say so and I'll add it. Everything else was your explicit findings.

Not merging, not starting S2, flag stays OFF — back to you for the third review.

zensgit:
Round-3 re-review verdict: **REQUEST CHANGES, now narrowly scoped.** The round-1 and round-2 findings are closed on the actual remote head `6a74a4324`; CI is genuinely green on that SHA, and the real-DB log proves the 13-test file executed on Postgres. Keeping `event_id` non-unique is accepted: the ratified design makes sink idempotency the net-once authority, and no global cross-event-type uniqueness contract has been established.

Remaining findings:

1. **[P2] `NOT NULL` still admits a missing identity: `event_id=''` or whitespace.** The column and `OutboxRow.eventId: string` now reject SQL NULL but still accept blank text. That is not a stable original-event identity. The existing sink-side dedup helper makes the failure concrete: `automation-event-dedup.ts:23-24` trims `_eventId` and returns `null` when it is blank, so such an outbox row cannot satisfy #4203's cutover dedup/idempotency contract even though the schema calls it valid. Add a named `CHECK (btrim(event_id) <> '')`, plus negative controls for `''` and whitespace and a positive stable-id control. Keep it non-unique.

2. **[P3] The lease rationale is factually wrong in three public places.** The migration header/body, the real-DB test comment, and the PR body say a lease on `pending`/`done` would be "spuriously reclaimed by the same scan." The documented scan is status-gated (`status='in_progress' AND lease_expires_at < now()`), so a non-`in_progress` row cannot match it. The biconditional CHECK is still a good invariant: it makes lease presence an unambiguous ownership marker and forces completion/dead-letter transitions to clear the lease atomically. Keep the CHECK; correct the rationale rather than claiming a behavior the query does not have.

3. **[P3-test hygiene] `Date.now() % 1e9` does not make parallel runs "never collide."** Two concurrent runs against the same DB can start in the same millisecond (or repeat modulo the window) and delete each other's fixtures. This is a shared-DB real-DB family, so use `randomUUID()`/equivalent or soften the claim and prove isolation. `randomUUID()` is the cheap fix.

Live state: PR head is exactly `6a74a4324`, all checks correspond to it, and the real-DB line is `13 tests / 79ms`. The PR is now `BEHIND` (`base 0279acf8`, current main `40ec7cab`), so apply the narrow fixes, rebase, verify the remote ref, and rerun required checks + migration replay. Flag remains OFF; S2 remains held until this re-review closes.

zensgit:
Round-3 findings addressed; rebased onto current main (`c5d285db1`).

- **event_id ''/whitespace → rejected.** Added the named CHECK `automation_outbox_event_id_nonblank`. **One implementation call to flag:** you named `btrim(event_id) <> ''`, but I implemented it as `event_id ~ '[^[:space:]]'` (POSIX-whitespace regex). Reason: the dedup helper is `automation-event-dedup.ts:23` → `_eventId.trim()`, and JS `.trim()` strips **tab/newline too**, so a spaces-only `btrim` would leave a gap the helper wouldn't — the regex mirrors the helper over ASCII whitespace. It also avoids backslash escapes (`E' \t\n'`), which inside the Kysely `sql`…`` template literal are exactly the hazard that reddened last round. If you'd rather have the literal spaces-only `btrim`, say so and I'll swap it. Real-DB negatives (empty, spaces, **tab+newline**) + positive added.
- **Lease comment corrected.** You're right — the reclaim scan is status-scoped, so it would never reclaim a pending/done + lease row; my 'spuriously reclaimed' rationale was wrong. The biconditional CHECK stays; the reverse half's stated reason is now: keep a lease meaning exactly 'owned by a live in_progress worker', with every terminal/reclaim transition clearing it atomically.
- **Test ids** `Date.now() % 1e9` → `randomUUID()` (genuinely collision-free).

Rehearsed on a fresh real Postgres 15: **11/11** probes pass incl. the tab/newline rejection a spaces-only btrim would miss, and the DDL replays idempotently. Typecheck green under CI's `tsc`; no backtick/backslash/${ in any sql block. Re-greening CI; flag stays OFF, not starting S2, no self-merge.

zensgit:
CI fully green on `c5d285db1` — all six required checks + `migration-replay` + `e2e` + `after-sales`.

**Real-DB spec verified executed** (triggered ≠ verified): `✓ …outbox-schema-realdb.test.ts (14 tests) 87ms` against real Postgres — 14, up from 13, so the new event_id-blank controls ran (empty, spaces, and tab+newline all rejected by `automation_outbox_event_id_nonblank`; a valid id accepted).

Not merging, not starting S2, flag stays OFF — back to you for the final review. One thing to confirm at your discretion: the event_id non-blank CHECK is a POSIX-whitespace regex (`event_id ~ '[^[:space:]]'`) rather than the literal spaces-only `btrim(event_id) <> ''`, so it mirrors the dedup helper's `.trim()` (which strips tab/newline too). Happy to swap to the literal btrim if you prefer it.

zensgit:
Final-gate re-review verdict: **REQUEST CHANGES — one P2 remains reproducible; two P3 metadata/comment remnants remain.** The round-3 random-UUID fix is good, the migration/lease implementation is otherwise coherent, CI is green on the exact remote head `c5d285db1`, and the real-DB file genuinely ran 14/14. Keeping `event_id` non-unique remains approved.

1. **[P2] The new POSIX-space CHECK still does not mirror JavaScript `.trim()`.** The test covers ASCII space/tab/newline only. I reproduced the gap on primary runtimes:
   - Node: NBSP (`U+00A0`), EM SPACE (`U+2003`), NARROW NBSP (`U+202F`), and BOM (`U+FEFF`) each become `''` after `.trim()`.
   - PostgreSQL 15 with `lc_ctype=en_US.UTF-8`: every one of those values returns `true` for `event_id ~ '[^[:space:]]'` and therefore passes `automation_outbox_event_id_nonblank`.

   So the schema still admits an identity that the downstream helper treats as no identity; this is the same #4203 cutover/idempotency violation, just outside the ASCII test vocabulary. Do not depend on locale-sensitive POSIX `[:space:]` matching ECMAScript whitespace. The cleanest contract is to declare event IDs ASCII and require at least one printable ASCII non-space character (for example a locale-independent `[!-~]` positive requirement); all current generators are already ASCII UUIDs or ASCII structured IDs. Alternatively use another explicit canonical contract, but add NBSP/EM SPACE/NARROW NBSP/BOM negative controls and a current UUID/approval-event positive control. Keep the column non-unique.

2. **[P3] The old reclaim claim still survives in public artifacts.** The PR body still says a stale lease on non-`in_progress` is "spuriously reclaimed," and the real-DB test at lines 179-180 repeats that it "would be spuriously matched by the reclaim scan." The production migration comment is corrected; sync the body and test comment to the same ownership-marker rationale.

3. **[P3] Scope wording.** The migration currently says the POSIX class mirrors JS `.trim()` "over ASCII," while the header/body say it mirrors `.trim()` without that qualifier. Once the validator is made exact, state one canonical identity alphabet instead of claiming equivalence between two different whitespace implementations.

Live state: head `c5d285db1`, CI run head matches, real-DB line is `14 tests / 87ms`; PR is again `BEHIND` (`base ce31ec0f`, current main `5fa10380`). Fix the validator and stale wording, rebase, rerun the exact real-DB lane + migration replay + required checks, and verify the remote SHA. Flag stays OFF and S2 stays held until this closes.

zensgit:
Final-gate findings addressed; rebased onto current main (`0c27a74d4`).

- **P2 — Unicode-whitespace event_id.** You're right: PostgreSQL's `[:space:]` is ASCII-only, so `[^[:space:]]` accepted an id made solely of NBSP / EM SPACE / narrow-NBSP / BOM — all of which JS `.trim()` strips to '' (no identity). Fixed per your recommendation: `event_id ~ '[!-~]'` requires a printable-ASCII non-space char (0x21–0x7E), rejecting empty, ASCII whitespace, **and** those Unicode cases. Added real-DB negatives for all four Unicode chars (as `\u` escapes, so the test file stays pure ASCII).
- **P3 ×2 — stray 'reclaim' rationale.** Removed the wrong 'a stale lease on pending/done would be spuriously reclaimed' claim from **both** the PR body and the test comment; both now state the real reason (status-scoped scan never sees settled rows; the reverse half keeps a lease meaning exactly 'owned by a live in_progress worker', cleared atomically on every terminal/reclaim transition).

Rehearsed on a fresh real Postgres 15 under **UTF8 / en_US.UTF-8** (the env you cited): **12/12** probes pass, including all four Unicode-whitespace rejections, plus idempotent replay. Typecheck green under CI's `tsc`; sql blocks free of backtick/backslash/${.

event_id stays non-UNIQUE (your accepted call). Re-greening CI; flag OFF, not starting S2, no self-merge — back to you for approval.

zensgit:
CI fully green on `0c27a74d4` — all six required checks + `migration-replay` + `e2e` + `after-sales` + `coverage`.

**Real-DB spec verified executed** (triggered ≠ verified): `✓ …outbox-schema-realdb.test.ts (14 tests)` against real Postgres, so the event_id test's new negatives ran — empty, ASCII whitespace, and all four Unicode-whitespace ids (NBSP U+00A0 / EM U+2003 / narrow-NBSP U+202F / BOM U+FEFF) rejected by `automation_outbox_event_id_nonblank` (`event_id ~ '[!-~]'`); a real id accepted. This is on top of the local PG15 rehearsal under UTF8 / en_US.UTF-8 (12/12).

All four review rounds' findings are now resolved; event_id stays non-UNIQUE per your accepted call. Not merging, not starting S2, flag stays OFF — back to you for approval.

zensgit:
Final re-review verdict: **APPROVE — no remaining P1/P2/P3 findings on the reviewed code.**

Verified against the actual remote head `0c27a74d4`:
- `event_id` is NOT NULL and the locale-independent `[!-~]` positive requirement rejects empty, ASCII whitespace, NBSP, EM SPACE, NARROW NBSP, and BOM while accepting the current ASCII event-id shape;
- non-uniqueness is intentional and correct for the ratified sink-idempotency model;
- the four-state consumer machine has no absorbing non-terminal state;
- lease presence is exactly equivalent to `in_progress`, with the ownership-marker rationale now correct in migration, test, and PR body;
- `fence` remains a decimal string across the PostgreSQL `bigint` boundary, including above 2^53;
- every named CHECK has discriminating real-DB controls;
- two-point CI wiring is intact and the exact run head matches `0c27a74d4`;
- the real-DB lane actually executed `multitable-automation-outbox-schema-realdb.test.ts`: **14/14, 94ms**;
- migration replay and all required checks are green;
- the flag remains default OFF and this slice introduces no active runtime path.

Operational condition only: the PR is currently `BEHIND` (`base 05b5d166`, current main `ede219d8`). Rebase onto fresh main and require the new remote head to pass migration replay, the real-DB lane, and all required checks before merge. This is not a code-review blocker and does not revoke APPROVE.

After #4303 lands, **S2 is authorized to start from fresh main** as its own reviewed slice: durable dispatcher + atomic lease claim/reclaim + server-side fence increment/CAS, still behind `AUTOMATION_DURABLE_DELIVERY_ENABLED=OFF`. FWB, attachment, and flag enablement remain gated.

PR #4316, opened 2026-07-15
fix(attendance): window-runner override persists under $HOME (fixes containment run 29398270060)

zensgit:
**Repo-side fix for the containment FAIL-CLOSED (run 29398270060).** Draft, no auto-merge. This is part (1) of the owner's three-part remediation; parts (2) staging re-normalization and (3) containment re-run are separate authorized ops actions — **this PR touches no host and changes no flag.**

## Root cause (not a flag leak)
The window-runner wrote its compose override under the per-run `$OUTPUT_DIR`; the workflow `rm -rf`s that on cleanup. But `docker compose up -d` stamps each container's `com.docker.compose.project.config_files` label with that path, so the containment check's `docker compose config` re-reads a **deleted** file ⇒ staging next-restart config unverifiable ⇒ correct FAIL-CLOSED.

## Fix (per owner's requirements)
- **Persistent path:** `OVERRIDE_FILE` → `$HOME/.metasheet2/window-runner/` (deploy-user-writable — not the no-write `$STAGING_DIR`; survives OUTPUT_DIR cleanup). Based on current `main`, **not** stacked on #4292.
- **Atomic write:** `mktemp` candidate → `docker compose -f staging -f candidate config` validation → same-dir `mv` (rename). A broken candidate never replaces the known-good live override.
- **Lifecycle decoupled from OUTPUT_DIR:** workflow cleanup `rm -rf` targets only the output + checkout dirs; the persist dir is explicitly excluded (commented).
- **set_window_env=none clears:** the `ATTENDANCE_*_ENABLED` echoes stay gated behind `rd-window`, so a none redeploy rewrites the same persistent file without them.
- **4 contract tests:** persistent path / atomic replace / none-clears-flags / cleanup-spares-persist-dir. **20/20 green**, `bash -n` + workflow yaml clean.

## Coordination (#3317, parallel session)
**#4292** (`codex/…-3317`) edits the same `remote.sh` (in `action_migrate_rehearse()`, ~line 624) and the same `.test.mjs` (~line 182). My edits are in a **different function** (`action_deploy` + config header) and my tests are **appended at the file end** — no line overlap, so whichever lands second rebases cleanly. No other PR is fixing this override-lifecycle issue.


zensgit:
Independent review of exact head `0fa9f7a397000c9012357cd9917261976745c857`: **REQUEST CHANGES (1 blocker, 1 P3).**

**[P1] The production `mktemp` template is invalid/non-unique across the two relevant implementations.**

Current code:

```bash
mktemp "${RUNNER_PERSIST_DIR}/.override.XXXXXX.yml"
```

GNU `mktemp` requires the trailing X run at the end of the template unless a GNU-only `--suffix` option is used, so the Linux deploy host will reject this form. On the local BSD implementation I independently executed the exact shape: it returned the literal path `.override.XXXXXX.yml` with rc=0, i.e. no randomization at all. Thus this either aborts the deploy before writing the override or loses the promised unique atomic candidate.

The new test currently only regex-matches this exact broken string, so 20/20 green cannot detect the runtime failure and in fact locks the defect in.

**Required fix:** use the portable no-suffix template, e.g. `mktemp "${RUNNER_PERSIST_DIR}/.override.XXXXXX"`, update the source-contract assertion, and add an executable positive control that invokes the production template shape and proves: rc=0, the file exists in the persistent directory, and the returned basename no longer contains the literal `XXXXXX`. Keep validation-before-rename ordering.

**[P3] Stale comment:** `BACKUP_DIR` still says “same reason OVERRIDE_FILE above lives under OUTPUT_DIR”; the override no longer lives there. Point it at the persistent user-writable directory instead.

Everything else in the reviewed diff is directionally sound: stable absolute path, same-directory rename, workflow cleanup exclusion, and `set_window_env=none` rewrite semantics. No host action or flag change should happen until a corrected head is re-reviewed.

zensgit:
**P1 + P3 fixed — re-review `4fb7ab403`.**
- **P1 (non-portable mktemp):** template changed to `${RUNNER_PERSIST_DIR}/.override.XXXXXX` (X run at the END; GNU no longer errors, BSD/macOS now randomizes). The candidate needs no `.yml` — `docker compose config -f` reads by content, and it's renamed to the `.yml`-named live override on success.
- **Real positive control (replaces the string-match that locked the defect in green):** a new test EXECUTES `mktemp` with the exact extracted template and asserts it actually randomizes (status 0, output ≠ literal, two calls differ, files created). **Mutation-proven:** reverting the template to `.XXXXXX.yml` reds it.
- **P3:** `BACKUP_DIR` comment corrected ("OVERRIDE_FILE … under OUTPUT_DIR" → `$HOME/.metasheet2`).
21/21 green, `bash -n` clean. Still Draft, no host action, no flag change.

PR #4322, opened 2026-07-15
feat(automation): P2 durable-delivery S2-a — claim engine / fence-CAS (additive, flag-gated)

zensgit:
**P2 durable-delivery slice S2-a — the claim engine (fence-CAS).** First runtime slice of the durable dispatcher (FWB-0 #4203 Layer 1), on top of the landed S1 outbox schema (#4303). **Additive, flag-gated, zero behavior change:** nothing calls these primitives while `AUTOMATION_DURABLE_DELIVERY_ENABLED` is OFF; the poll+dispatch loop is S2-b and the consumer adapters are S5.

## What lands
- **`automation-durable-dispatcher.ts`** — the claim-engine primitives:
  - `claimDueConsumers(db, { consumerKeys, batchSize?, leaseMs?, maxAttempts?, now? })` — **`consumerKeys` is REQUIRED and non-empty** (the registry allow-list of keys this worker knows). One atomic `FOR UPDATE SKIP LOCKED` CTE: a reclaimable row (`pending`, or `in_progress` past a stale lease) **whose key is in the allow-list** is claimed (→ `in_progress`, fresh lease, `fence`++ , `attempts`++ **at claim**), OR — if `attempts >= maxAttempts` — **poisoned to `dead_letter` right there** (crash-safe, no surviving worker needed) and not dispatched. A key **not** in the allow-list is never touched, so it stays `pending` for a worker that knows it (rolling-deploy contract, #4203 §246).
  - `findUnknownConsumerKeys(db, knownKeys)` — the alert seam: reclaimable rows whose key this worker doesn't know (the loop alerts on these, never claims them).
  - `completeConsumer` / `poisonConsumer` / `rescheduleConsumer` — **fence-CAS** resolves (`WHERE fence = <claimed> AND status='in_progress'`). A zombie/stale token writes 0 rows. `rescheduleConsumer` (transient failure) keeps the row `in_progress`, pushes the lease out by a backoff (re-claimed only after expiry, never instantly), **and atomically bumps the fence** — so the token the caller used is dead and a late worker holding it can't complete/poison/reschedule. `reason` is a typed, values-free `DeliveryFailureReason`, enforced at runtime.
  - `resolveDisposition('success' | 'retryable_failure' | 'permanent_failure')` → `complete` / `reschedule` / `poison` (attempt-exhaustion poison is claim-driven, not a disposition).
  - All numeric bounds validated as **safe integers in explicit ranges** (batch 1..10000, lease/delay 1..86_400_000, attempts 1..10000) — no fractional/huge value reaches a SQL cast.
- **Real-DB spec (13 tests, real Postgres, CONSTRUCTED races)** — claim shape; discriminating SKIP-LOCKED (tx A holds a lock, B skips it and takes a free row, tx-scoped `SET LOCAL lock_timeout` + `finally` rollback so the pool isn't polluted); zombie fence-CAS; complete + idempotent; reschedule (backoff + fence-bump/token-invalidation + reclaim); crash-safe poison at the ceiling; **rolling-deploy unknown-key** (N-1 leaves it `pending`, N processes it, alert seam surfaces it); permanent poison; attempts-at-claim across reclaims; input validation; values-free `last_error`; pure `resolveDisposition`. **Two-point wired** (plugin-tests.yml run list + vitest.config.ts exclude).

## Verification
Ran locally on **real Postgres 15**: **13/13 pass**. Five guards are **mutation-proven load-bearing**: unknown-key scoping (always-true filter → N-1 claims it → red), reschedule fence-bump (drop the bump → stale token completes → red), SKIP-LOCKED (drop it → red on lock timeout), fence-CAS, and crash-safe poison. Typecheck green under CI's `tsc`.

## Not in this slice
S2-b poll+dispatch loop (drives claim → adapter → resolve, and alerts on `findUnknownConsumerKeys`), S3 routing manifest + startup completeness assertion, S4 producer atomic enqueue, S5 consumer adapters + poison-terminal. Flag stays OFF; FWB / attachment / flag-enable remain gated until full implementation + the 8-scenario acceptance.

### DIFF
--- a/packages/core-backend/src/multitable/automation-durable-dispatcher.ts
+++ b/packages/core-backend/src/multitable/automation-durable-dispatcher.ts
@@ -1,30 +1,37 @@
 /**
  * P2 durable-delivery — slice S2-a: the claim engine (fence-CAS).
  *
- * The durable dispatcher drains `meta_automation_outbox_consumer` rows by CLAIMING each reclaimable row
- * (pending, or in_progress past a stale lease), invoking its consumer adapter (S5), then resolving the row
- * — `complete` (→ done), `poison` (→ dead_letter, bounded attempts exhausted), or `release` (→ pending,
- * a transient failure that stays reclaimable). This module is those four primitives only; the poll+dispatch
- * loop and the adapter wiring are S2-b / S5.
+ * The durable dispatcher drains `meta_automation_outbox_consumer` rows by CLAIMING each reclaimable row,
+ * invoking its consumer adapter (S5), then resolving it. This module is those primitives only; the
+ * poll+dispatch loop and the adapters are S2-b / S5.
  *
- * Concurrency contract (mirrors the proven AttendanceNotificationDeliveryWorker):
- *   - CLAIM is a single atomic `FOR UPDATE SKIP LOCKED` CTE: it flips the row to `in_progress`, stamps a
- *     fresh lease, **increments `fence` (the monotonic claim token)** and **increments `attempts` AT CLAIM**
- *     (so a worker that always crashes after claiming still drives the row toward `dead_letter`). Two workers
- *     never claim the same row — SKIP LOCKED hands each a disjoint set.
- *   - Every resolve is a **fence-CAS**: `WHERE ... AND fence = <claimed fence> AND status = 'in_progress'`.
- *     A "zombie" (a live worker whose lease expired and was reclaimed by another) still holds the OLD fence,
- *     so its write matches 0 rows and is a silent no-op — the reclaimer (new fence) owns the row. This is why
- *     `fence` is a bigint carried as a `string`: never coerce it through a JS number (2^53).
- *   - Terminal/reclaim writes clear the lease in the SAME UPDATE that changes status, upholding the
- *     `lease ⟺ in_progress` biconditional CHECK at every commit boundary.
+ * Correctness contract (mirrors the proven AttendanceNotificationDeliveryWorker + FWB-0 lock #4203 Layer 1):
+ *   - CLAIM is a single atomic `FOR UPDATE SKIP LOCKED` CTE that also enforces the poison ceiling. For a
+ *     reclaimable row (pending, or in_progress past a stale lease):
+ *       · attempts < maxAttempts → flip to `in_progress`, stamp a fresh lease, bump `fence` (the monotonic
+ *         claim token) and `attempts` (AT CLAIM), and hand it out for dispatch;
+ *       · attempts ≥ maxAttempts → **poison it to `dead_letter` right here** and DON'T dispatch it.
+ *     Poisoning at claim time (driven only by the persisted `attempts`) is what makes the ceiling
+ *     **crash-safe**: a worker that always crashes AFTER claiming never runs a resolve path, yet the row
+ *     still terminates — the next reclaim finds `attempts ≥ maxAttempts` and dead-letters it. "非无限
+ *     reclaim" (#4203 §poison-terminal). SKIP LOCKED hands concurrent workers disjoint rows.
+ *   - Every resolve is a **fence-CAS** (`WHERE ... AND fence = <claimed> AND status = 'in_progress'`): a
+ *     "zombie" whose lease was reclaimed holds a stale fence, so its write matches 0 rows (silent no-op)
+ *     while the reclaimer (new fence) owns the row. `fence` is a bigint carried as a `string` (never a JS
+ *     number — 2^53).
+ *   - A retryable failure does NOT flip the row back to an immediately-reclaimable state — that would spin
+ *     retries with no backoff and burn the ceiling in a tight loop. Instead `reschedule` keeps the row
+ *     `in_progress` and pushes its lease out by a backoff, so it is re-claimed only AFTER the lease expires
+ *     (#4203 §"租约过期后被 reclaim"). Terminal/lease writes preserve the `lease ⟺ in_progress` biconditional.
+ *   - Failure detail is NEVER persisted verbatim (a raw adapter error can carry a secret-shaped URL/token).
+ *     Resolves take a typed, values-free `DeliveryFailureReason` code, and only the code reaches the DB.
  *
- * Nothing here runs while `AUTOMATION_DURABLE_DELIVERY_ENABLED` is OFF — these are pure primitives with no
- * caller until the S2-b loop is wired behind the flag.
+ * Nothing here runs while `AUTOMATION_DURABLE_DELIVERY_ENABLED` is OFF — pure primitives with no caller
+ * until the S2-b loop is wired behind the flag.
  */
 import type { OutboxConsumerStatus } from './automation-durable-delivery'
 
-/** Minimal query surface — a pg Pool or a checked-out client both satisfy it (keeps the engine testable). */
+/** Minimal query surface — a pg Pool wrapper or a checked-out client both satisfy it (keeps the engine testable). */
 export interface Queryable {
   query(sql: string, params?: unknown[]): Promise<{ rows: Array<Record<string, unknown>>; rowCount: number | null }>
 }
@@ -32,13 +39,41 @@ export interface Queryable {
 /** Default attempt ceiling before a consumer row is poisoned to `dead_letter`. */
 export const DEFAULT_MAX_DELIVERY_ATTEMPTS = 8
 
+/**
+ * The ONLY strings that may land in `last_error` — a closed, values-free vocabulary so a raw adapter error
+ * (which can embed a secret-shaped URL/token) is never persisted verbatim. The dispatch loop maps an
+ * adapter outcome to one of these codes; the real detail belongs in redacted structured logs, not this row.
+ */
+export type DeliveryFailureReason =
+  | 'adapter_error'
+  | 'adapter_timeout'
+  | 'permanent_rejection'
+  | 'max_attempts_exhausted'
+  | 'unknown'
+
+/** The closed set of legal reason codes — enforced at RUNTIME so even an untyped caller can't persist a raw
+ *  (possibly secret-bearing) string. Values-free by construction, not just by the TS type. */
+const ALLOWED_FAILURE_REASONS: ReadonlySet<string> = new Set<DeliveryFailureReason>([
+  'adapter_error',
+  'adapter_timeout',
+  'permanent_rejection',
+  'max_attempts_exhausted',
+  'unknown',
+])
+
+function assertReason(reason: DeliveryFailureReason): void {
+  if (!ALLOWED_FAILURE_REASONS.has(reason)) {
+    throw new RangeError('last_error must be a values-free DeliveryFailureReason code, not a raw message')
+  }
+}
+
 /** A row claimed for dispatch — carries the claim `fence` (the CAS token) and the joined outbox event. */
 export interface ClaimedConsumer {
   outboxId: string
   consumerKey: string
   /** post-increment claim token (bigint as string — never a JS number). */
   fence: string
-  /** post-increment attempt count for THIS claim (>= 1). */
+  /** post-increment attempt count for THIS claim (1..maxAttempts). */
   attempts: number
   eventType: string
   eventId: string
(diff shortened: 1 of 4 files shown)


zensgit:
CI fully green on `f900fef3f` — all six required checks + `migration-replay` + `e2e` + `after-sales`.

**Real-DB spec verified executed** (triggered ≠ verified): `✓ …dispatcher-claim-realdb.test.ts (10 tests)` in the `test (20.x)` real-DB lane against real Postgres — the constructed-race tests (zombie fence-CAS, SKIP-LOCKED mutual exclusion, complete/release/poison, attempts-at-claim across reclaims) all ran, not skip-greened. This matches the local run (10/10 on PG15), where I also confirmed the fence-CAS test is load-bearing via a tautology mutation.

Flag `AUTOMATION_DURABLE_DELIVERY_ENABLED` stays OFF; additive/behavior-neutral. For your review — not self-merging.

packages/core-backend/src/multitable/automation-durable-dispatcher.ts
```diff
@@ -0,0 +1,205 @@
...
+          AND ($4::text[] IS NULL OR c.consumer_key = ANY($4::text[]))
+        ORDER BY c.updated_at ASC
+        LIMIT $2::int
+        FOR UPDATE SKIP LOCKED
+     ),
+     claimed AS (
+       UPDATE meta_automation_outbox_consumer c
+          SET status = 'in_progress',
+              lease_expires_at = $1::timestamptz + ($3::int * interval '1 millisecond'),
+              fence = c.fence + 1,
```
zensgit: [P1] Incrementing `attempts` here does not enforce the ceiling. `resolveDisposition` runs only after an adapter returns, so an always-crashing worker never calls `poisonConsumer`. On a fresh DB I reclaimed this row ten times with the default ceiling of eight; it remained `in_progress` with `attempts=10`. The claim statement must atomically dead-letter an exhausted due row before issuing another lease, with a real-DB crash-only golden.

packages/core-backend/src/multitable/automation-durable-dispatcher.ts
```diff
@@ -0,0 +1,205 @@
...
+ */
+export function releaseConsumer(
+  db: Queryable,
+  outboxId: string,
+  consumerKey: string,
+  fence: string,
+  error: string,
+  opts: { now?: () => Date } = {},
+): Promise<boolean> {
+  return casResolve(db, outboxId, consumerKey, fence, 'pending', error.slice(0, 1000), nowIso(opts))
```
zensgit: [P2] This clears the lease and makes a transient failure immediately reclaimable, bypassing the ratified lease-paced retry contract. A tight poll loop can consume all attempts during one short outage. Preserve a retry due time (or equivalent lease pacing), and reject non-positive/invalid lease and attempt configuration; `leaseMs=-1` currently produces two immediate successive owners.

packages/core-backend/tests/integration/multitable-automation-dispatcher-claim-realdb.test.ts
```diff
@@ -0,0 +1,192 @@
...
+    })
+  })
+
+  test('SKIP LOCKED: two concurrent claimers over one due row → exactly one wins (no double-claim)', async () => {
+    const { outboxId, consumerKey } = await seedRow()
+    const raw = db().getInternalPool()
+    const a = await raw.connect()
+    const b = await raw.connect()
+    try {
+      const [ra, rb] = await Promise.all([
```
zensgit: [P2] This is not a constructed `SKIP LOCKED` interleaving: both autocommit statements can run sequentially, and the test still reports exactly one winner with plain `FOR UPDATE`. Hold A's transaction open after it claims row 1, then require B to claim row 2 before A commits (with a positive control); removing `SKIP LOCKED` must block/fail and turn the test red.

packages/core-backend/src/multitable/automation-durable-dispatcher.ts
```diff
@@ -0,0 +1,205 @@
...
+ */
+export function poisonConsumer(
+  db: Queryable,
+  outboxId: string,
+  consumerKey: string,
+  fence: string,
+  error: string,
+  opts: { now?: () => Date } = {},
+): Promise<boolean> {
+  return casResolve(db, outboxId, consumerKey, fence, 'dead_letter', error.slice(0, 1000), nowIso(opts))
```
zensgit: [P2] `error` is arbitrary caller text and is persisted verbatim. A secret-shaped connection URL/token survives unchanged in `last_error`, so S2-b's natural `Error.message` catch would create a credential/host diagnostic sink. Accept a typed values-free reason or apply the established redaction boundary, and pin it with a negative control.

zensgit (commented):
**BLOCKING REVIEW — do not merge or start S2-b yet.**

The fence-CAS primitive itself is sound, the flag-off/no-caller boundary is real, and the real-DB file is correctly two-point wired. However, S2-a does not yet implement the ratified bounded-retry contract, and two claimed reliability properties are not test-locked.

Independent evidence on a fresh PostgreSQL database:
- the submitted 10-test file passes 10/10;
- with DEFAULT_MAX_DELIVERY_ATTEMPTS=8, ten consecutive claim -> lease-expire cycles leave the row in_progress with attempts=10 and fence=10;
- the submitted "exactly one winner" test still passes after replacing FOR UPDATE SKIP LOCKED with plain FOR UPDATE;
- releaseConsumer persists a secret-shaped DB URL/token string verbatim in last_error;
- leaseMs=-1 allows two immediate successive claims (fence 1 then fence 2).

Required fixes:
1. Enforce crash-only attempt exhaustion atomically in the claim statement: after the last permitted actual claim expires, the next due scan must terminalize dead_letter without issuing another lease. Add a RED-before real-DB golden proving attempts remain bounded.
2. Preserve retry pacing. releaseConsumer currently clears the lease and makes a transient failure immediately reclaimable, contrary to the ratified "lease expires, then reclaim" contract. Use a real retry-due mechanism or an equivalent lease-paced state transition, with a positive transient-failure retry control. Validate positive bounded integer batch/lease/attempt options.
3. Replace the current SKIP LOCKED test with a deterministic held-transaction interleaving: A keeps row 1 locked while B must claim row 2 without waiting. Removing SKIP LOCKED must turn this test red.
4. Do not accept and persist arbitrary adapter error text. Store a typed/allowlisted values-free reason (or run it through the repository's established redaction boundary) and add a secret-shaped negative control.

Also correct the S1/S2 comments and PR body that currently claim crash-after-claim reaches dead_letter; that statement is false on this head.

GitHub does not permit this connected account to submit REQUEST_CHANGES on its own PR, so this is recorded as a blocking COMMENT review rather than an approval-state transition.

zensgit:
All four findings fixed; rebased onto current main (`b3e2f8cb0`).

- **[P1] crash-after-claim never poisoned.** You're right — poison lived on the resolve path, so a worker that always crashes after claiming never reached it. Moved poison **into the claim CTE**: a reclaimable row with `attempts >= maxAttempts` is dead-lettered right there, driven only by the persisted `attempts` (no surviving worker needed), and not dispatched. New real-DB test replays your repro (claim→expire ×N), and the next claim dead-letters it (`attempts` capped, `last_error='max_attempts_exhausted'`, terminal).
- **[P2] immediate retry.** Replaced `releaseConsumer` (which flipped to instantly-reclaimable `pending`) with **`rescheduleConsumer`**, which keeps the row `in_progress` and pushes the lease out by a backoff — so it's re-claimed only after the lease expires, per the locked "租约过期后再 reclaim" semantics. `leaseMs` and `retryDelayMs` are now validated `> 0` (the `-1` case throws instead of minting an already-expired lease).
- **[P2] SKIP-LOCKED test had no discriminating power.** Rewrote it exactly as you suggested: transaction A holds a row lock uncommitted; B must SKIP it and claim a **different free row**. B's connection sets `lock_timeout='2000ms'`, so removing SKIP LOCKED makes B block → times out → red. **Mutation-verified**: dropping SKIP LOCKED fails the test on `canceling statement due to lock timeout`.
- **[P2] last_error verbatim.** Resolves now take a typed, **values-free `DeliveryFailureReason`** code, enforced at **runtime** (closed set + reject), so a raw/secret-shaped string throws before it can reach the DB.

**API changed** (heads-up for review): `releaseConsumer` → `rescheduleConsumer(…, retryDelayMs, reason)`; `poisonConsumer(…, reason)`; `resolveDisposition('success'|'retryable_failure'|'permanent_failure')` → complete/reschedule/poison (attempt-exhaustion poison is now claim-driven, not a disposition).

Verified on real Postgres 15: **13/13 pass**, and the poison + SKIP-LOCKED guards are **mutation-proven load-bearing** (raise the ceiling → poison test red on dispatched-not-poisoned; drop SKIP LOCKED → red on lock_timeout); fence-CAS remains load-bearing. Flag stays OFF; not self-merging. Re-greening CI now.

zensgit:
CI fully green on `b3e2f8cb0` — all six required checks + `migration-replay` + `e2e` + `after-sales`.

**Real-DB spec verified executed:** `✓ …dispatcher-claim-realdb.test.ts (13 tests)` in the `test (20.x)` lane against real Postgres — the crash-safe-poison (claim→expire×N then dead-letter), reschedule-backoff, discriminating SKIP-LOCKED, input-validation, and values-free tests all ran, not skip-greened. Matches the local run (13/13 on PG15) with the poison + SKIP-LOCKED guards mutation-proven load-bearing.

Flag stays OFF; additive/behavior-neutral. Back to you for re-review — not self-merging.

zensgit:
Thanks — I want to flag a head mismatch before you spend more time: the four blocking comments are anchored to commit `f900fef3f` (the *original* S2-a commit). I can see this in the API — every review comment's `original_commit_id` is `f900fef3f`, and they reference symbols that **no longer exist**: `releaseConsumer` (line ~182/190) and `poisonConsumer` at line 166. The rework that addressed all four landed in `b3e2f8cb0` and is now rebased onto main at **`53b6c0d6c`** (no longer BEHIND). Zero review comments are anchored to that head yet.

Where each finding is handled on the current head:

- **[P1] crash-after-claim never poisons** → **fixed by moving poison into the claim CTE.** `claimDueConsumers` now dead-letters a reclaimable row when `attempts >= maxAttempts`, driven only by the persisted `attempts` — no surviving worker needed. Your exact repro can't occur on this head: default ceiling 8, claim→expire ×10 dead-letters at claim 9 (not "attempts=10, in_progress"). There is a real-DB test that replays precisely that (`CRASH-SAFE poison…`), and it's mutation-proven (raise the ceiling → the test goes red on dispatched-not-poisoned).
- **[P2] immediate retry** → `releaseConsumer` is **removed**; `rescheduleConsumer` keeps the row `in_progress` and pushes the lease out by a backoff (reclaim only after expiry). `leaseMs`/`retryDelayMs` are validated `> 0` (the `-1` case throws).
- **[P2] SKIP-LOCKED test non-discriminating** → rewritten to your spec: tx A holds a row lock uncommitted, B must skip it and claim a **different free row**, with `lock_timeout` on B. Mutation-verified — dropping SKIP LOCKED turns it red on `canceling statement due to lock timeout`.
- **[P2] last_error verbatim** → resolves take a typed, values-free `DeliveryFailureReason` code, enforced at **runtime** (closed set + reject); a raw/secret-shaped string throws before touching the DB.

Verified on the current head `53b6c0d6c`: **13/13** on real Postgres 15 just now, and CI is green with the spec executing 13 tests. Could you re-review `53b6c0d6c`? If any concern still holds against *this* head's lines, point me at them and I'll fix — I just don't want to re-implement fixes that are already in. Flag stays OFF; not self-merging.

zensgit:
All five addressed on the current head `935a9b909` (this review was against the real current behavior — thank you):

- **[P1] unknown consumer_key silently handled** → `consumerKeys` is now **REQUIRED and non-empty**; the claim is scoped to the worker's known keys, so an unknown key is never claimed and stays `pending`. Added `findUnknownConsumerKeys` as the alert seam. Per #4203 §246 (this must exist in the first dispatcher version — a current worker won't hit unknown keys but becomes N-1 at the next bump). Real-DB rolling-deploy test: N-1 leaves the row `pending` even after 10 repeated claims (never dead-letters it), the alert seam surfaces it, and N processes it. **Mutation-proven** (always-true filter → N-1 claims it → red).
- **[P2] reschedule didn't revoke the old fence** → `rescheduleConsumer` now **atomically bumps the fence**, so the token the caller used is dead; a late `completeConsumer` with the old fence is a no-op. Test asserts exactly that. **Mutation-proven** (drop the bump → stale-token complete succeeds → red).
- **[P2] SKIP-LOCKED test polluted the pool** → switched to tx-scoped **`SET LOCAL lock_timeout`** (auto-reset) + best-effort `ROLLBACK` on both connections in `finally` before release. Still discriminating — **mutation-proven** (drop SKIP LOCKED → red on `canceling statement due to lock timeout`).
- **[P2] numeric params** → all bounds validated as **safe integers in explicit ranges** (batch 1..10000, lease/delay 1..86_400_000, attempts 1..10000); fractional (`1.5`) and over-cap (`1e15`) now throw instead of hitting a SQL cast. Tested.
- **[P3] PR body** → rewritten to the current API (required known-keys, reschedule fence-bump, values-free reason, 13 tests, no `releaseConsumer`).

Verified on real Postgres 15: **13/13**, with five guards mutation-proven load-bearing (unknown-key, reschedule-fence, SKIP-LOCKED, fence-CAS, crash-safe poison). Re-greening CI; flag stays OFF, not self-merging, S2-b still on HOLD.

PR #4323, opened 2026-07-15
feat(web): G-10 follow-up — normal-user surfaces render display names over bare entity IDs (owner ruling 2026-07-15)

zensgit:
Owner ruling (G-10 定稿 2026-07-15): rule-editor / admin technical surfaces keep `数据表 ID` labels with raw IDs (already correct); **normal-user surfaces show display NAMES, not bare sheetId/baseId/viewId/recordId values**.

Full audit sweep of `apps/web/src/multitable/**` (plus the multitable views under `apps/web/src/views/`) for raw entity-ID values interpolated into visible text. Non-display usages (`:key` bindings, API paths, postMessage plumbing, map keys) were excluded as non-surfaces.

## Fixed — 4 surfaces (names already in component scope; zero new data paths)

| # | Surface | Before | After |
|---|---|---|---|
| F1 | Export (all rows) download filename fallback — `MultitableWorkbench.vue` `exportAllRowsViaRoute` | `filename \|\| \`${sheetId}.${format}\`` → e.g. `sheet_orders.csv` when the route omits Content-Disposition | Sheet display name via new `activeSheetExportName` computed (from `workbench.sheets`, already loaded) → `Orders.csv`; generic `export` when no name resolves — never the raw id |
| F2 | Export (selected rows) download filename — `triggerDownload` | `\`${activeSheetId \|\| 'export'}.${ext}\`` — **always** the raw sheetId | `\`${activeSheetExportName \|\| 'export'}.${ext}\`` |
| F3 | xlsx workbook tab name — `doExportXlsx` | `activeSheetId.slice(0, 31)` — **always** the raw sheetId | Sheet display name, sanitized by the new `safeXlsxSheetName` boundary in `buildXlsxBuffer` (see hardening below) |
| F4 | Bulk-edit failure samples (dialog error + toast) — `onBulkEditApply` | `\`${failure.recordId}: ${failure.reason}\`` | New `bulkFailureSamples(failures, recordLabelOf)` helper (`meta-bulk-edit-labels.ts`) called with the existing `bulkFillRecordName` resolver over the already-loaded rows; the raw id appears only when the row is no longer loaded (honest fallback). Reason strings stay raw. |

### Gate P2 hardening (second commit)

The gate confirmed SheetJS 0.20.3's `check_ws_name` ALSO throws on the reserved tab name `History` (case-insensitive) and on leading/trailing apostrophes — a display name like `History` or `'Plan` would have made `buildXlsxBuffer` throw (export fails with the excelExportFailed toast). The pre-G-10 raw `sheet_…` ids could never hit this; the display-name change introduced the surface. Fixed with both gate-suggested legs:

- `safeXlsxSheetName()` (`xlsx-mapping.ts`) owns ALL tab-name rules, in order: forbidden chars → trim → 31-char cap → edge-apostrophe strip (re-trim) → reserved `History` → `History_` suffix → empty → `Sheet1`.
- `buildXlsxBuffer` sanitizes at the boundary itself (defense in depth — no caller can make it throw on a hostile name); the workbench's local partial regex was removed in favor of the single boundary owner.
- Goldens with the REAL xlsx module (no `buildXlsxBuffer` mock): `History`/`history`/`HISTORY`, edge apostrophes, `'History'` (strip-before-reserved ordering), forbidden chars, all-apostrophes → all write a real workbook with a valid tab; interior apostrophes preserved; plus direct `safeXlsxSheetName` units. The SheetJS rejection surface was empirically reproduced against the pinned dependency before fixing, and dropping either the reserved-name branch or the apostrophe strip reddens 5 tests each (mutation landing grep-confirmed).

Spec pins (all in the required `web-tests` lane):
- `meta-bulk-edit-labels.spec.ts` +2: `bulkFailureSamples` name-leading rendering, resolver fallback, 3-sample cap.
- `multitable-workbench-view.spec.ts` +3 integration: filename fallback = `Orders.csv` (not `sheet_orders.csv`); selected-rows xlsx = `Orders.xlsx` download + `sheetName: 'Orders'` into `buildXlsxBuffer`; bulk-failure sample = `Alpha: locked` (record title), not `rec_1: locked`, in both dialog error and toast.
- All three integration pins are mutation-verified: reverting each fix independently reddens exactly its test (mutation landing confirmed by grep before each run).

## Left — name genuinely not in scope (HI-1: zero new data paths — listed, not fetched)

| Surface | Renders today | Why left |
|---|---|---|
| `apps/web/src/views/MultitableCommentInboxView.vue:37-41` — comment-inbox card meta line (normal-user) | `Base {baseId}` · `Sheet {sheetId}` · `Row {recordId}` · `View {viewId}` · `Field {fieldId}` — five raw ids | The inbox payload carries **no names**: `CommentService.getInbox` already joins `meta_sheets as s` but projects only ids. Rendering names needs a server projection change (or a per-sheet client fetch) = a new data path. **Top follow-up candidate** — projecting `s.name` server-side is small. |
| `MultitableWorkbench.vue` `recordNotFound` toast (deep-link open-record failure) | `未找到记录：rec_xxx` | Record is genuinely gone (404) — no name exists anywhere; recordId in deep-link display is an accepted technical context. Existing pin `multitable-workbench-i18n.spec.ts` ("keeps record id raw") intentionally unchanged. |

## Left — already name-first (raw id only as last-resort fallback; no change)

- `HistoryBatchChangesList.vue` / `TrashModal.vue` — record title via `pickRecordTitle`, `#shortid` fallback, full id only in the tooltip/title attr (technical affordance).
- Workbench `bulkFillRecordName` / `batchRecordLabel` (+ `gridSelectionLabels` capture) / `configHistoryLabelOf` — resolve names from loaded rows/fields/views first; bare id only for off-page rows (a name would require a fetch).
- `MetaLinkPicker.vue` `{{ item.display || item.id }}` — display server-resolved from the foreign sheet's display field; raw id only when the record has no readable display value (empty display); it is the only differentiator between such rows.

## Intentionally left — admin/technical surfaces (owner ruled these stay technical)

- `meta-automation-labels.ts` rule-editor config labels (`目标数据表 ID` / `数据表 ID` / field-ID placeholders) and the manual sheet-ID entry escape hatch.
- Automation Runs admin panel (`runs.sheetFilter` 表 ID + `AutomationExecutionsView.vue`, admin-gated).
- `meta-api-token-labels.ts` `apiDingTalkScopeLabel` sheet-scope chip (`…: {sheetId}`) — API Tokens & Webhooks manager, identifiers surface.
- Public-form URL construction/display (`MetaFormShareManager` / `MetaAutomationManager`) — deep-link display; URLs carry ids by design.

Out of the ruling's entity set (userIds, noted for completeness): `historyActor` `由 {actorId}` (record history / trash deletedBy), `HistoryCenterModal` `actorLabel` (name-first already), `MetaPersonPicker` `display || id` — no client-side user directory; would need a fetch.

## Cleared (no raw entity-ID visible text found)

Automation log/delivery viewers; AutomationManager/RuleEditor/ConditionalRuleBuilder templates; NotificationBell; LinkedRecordPopover; FormView; cell renderer/editor; Calendar/Gantt/Timeline/Hierarchy/Dashboard/Gallery/Kanban views; ImportModal; HistoryCenterModal chrome; Field/View managers; BasePicker; SheetViewRail; Reset dialogs; Record/Sheet permission managers; CommentsDrawer; AI bulk-fill dialog and RestoreBatchDialog (name-first via props); automationActionSummary; test-run messages; EmbedHost (plumbing only); Home/TemplateCenter/TemplateDetail/PublicForm views.

## Verification

- Rebased onto current main; `bash apps/web/scripts/run-required-web-tests.sh` — 336 files / 3934 tests, exit 0 (includes all new tests).
- `vue-tsc -b` — clean.
- Positive controls: each of the three workbench-level fixes AND both legs of the tab-name sanitizer (reserved-name branch, apostrophe strip) were individually reverted and their tests confirmed red before restoring.

EN locale, admin advanced settings, rule-editor config, and API/key/identifier surfaces untouched per the ruling.

### DIFF
--- a/apps/web/src/multitable/import/xlsx-mapping.ts
+++ b/apps/web/src/multitable/import/xlsx-mapping.ts
@@ -116,9 +116,36 @@ export function parseXlsxBuffer(
 }
 
 /**
+ * Sanitize an arbitrary display string into a workbook tab name SheetJS will accept.
+ *
+ * G-10 follow-up (owner ruling 2026-07-15): exports name the xlsx tab after the sheet's DISPLAY
+ * name, so this boundary must tolerate hostile user-authored names. SheetJS 0.20.3's
+ * `check_ws_name` (called by `book_append_sheet`) THROWS on: names longer than 31 chars, the
+ * chars []:*?/\, a leading or trailing apostrophe, and the reserved name 'History'
+ * (case-insensitive — empirically verified against the pinned dependency). Rules applied in
+ * order: forbidden chars → spaces; trim; cap at 31; strip edge apostrophes (re-trim — the strip
+ * can expose new edge whitespace); reserved 'History' → 'History_' suffix (7 chars, so the
+ * suffix can never overflow the cap); empty → 'Sheet1'.
+ */
+export function safeXlsxSheetName(name: string | undefined): string {
+  const cleaned = (name ?? '')
+    .replace(/[\\/\[\]:*?]/g, ' ')
+    .trim()
+    .slice(0, 31)
+    .replace(/^'+|'+$/g, '')
+    .trim()
+  if (!cleaned) return 'Sheet1'
+  if (/^history$/i.test(cleaned)) return `${cleaned}_`
+  return cleaned
+}
+
+/**
  * Build an `.xlsx` Uint8Array (browser-safe) from a tabular dataset.
  * Headers row first, then string-coerced data rows. Caller is responsible
  * for typing/formatting via the `serialize` callback.
+ *
+ * The tab name is sanitized via `safeXlsxSheetName`, so this function never throws on a
+ * hostile `sheetName` (defense in depth: callers may pre-sanitize, but the boundary holds).
  */
 export function buildXlsxBuffer(
   xlsx: XlsxModule,
@@ -128,7 +155,7 @@ export function buildXlsxBuffer(
     rows: Array<Array<string | number | boolean | null | undefined>>
   },
 ): Uint8Array {
-  const sheetName = (params.sheetName?.trim() || 'Sheet1').slice(0, 31)
+  const sheetName = safeXlsxSheetName(params.sheetName)
   const aoa: unknown[][] = [params.headers.slice()]
   for (const row of params.rows) {
     aoa.push(row.map((cell) => (cell === undefined || cell === null ? '' : cell)))
--- a/apps/web/src/multitable/utils/meta-bulk-edit-labels.ts
+++ b/apps/web/src/multitable/utils/meta-bulk-edit-labels.ts
@@ -88,6 +88,21 @@ export function bulkFailure(failed: number, requested: number, sampleFailures: s
     : `${failed} of ${requested} ${recordWord(requested)} failed${detail}`
 }
 
+// G-10 follow-up (owner ruling 2026-07-15): the bulk-edit failure sample is a normal-user
+// dialog/toast surface, so each sampled failure leads with the record's display NAME. The caller
+// supplies the resolver over rows it ALREADY has loaded (zero new data paths, HI-1) — that
+// resolver's own fallback keeps the raw record id only when the row is no longer in scope.
+// Failure `reason` strings stay raw per this module's scope note above.
+export function bulkFailureSamples(
+  failures: ReadonlyArray<{ recordId: string; reason: string }>,
+  recordLabelOf: (recordId: string) => string,
+): string {
+  return failures
+    .slice(0, 3)
+    .map((failure) => `${recordLabelOf(failure.recordId)}: ${failure.reason}`)
+    .join('; ')
+}
+
 export function bulkVersionConflict(message: string, isZh: boolean): string {
   const detail = message ? (isZh ? `（${message}）` : ` (${message})`) : ''
   return isZh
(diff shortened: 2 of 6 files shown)


zensgit:
Gate P2 addressed in `fe80ce1e7` (branch also rebased onto current main — the 335-vs-336 file-count question was a stale-base artifact and reconciles post-rebase: 336 files / 3934 tests green).

**Fix (both suggested legs):** `safeXlsxSheetName()` now owns all SheetJS tab-name rules (forbidden chars → trim → 31-cap → edge-apostrophe strip → reserved `History` → `History_` → empty → `Sheet1`), and `buildXlsxBuffer` applies it at the boundary itself so no caller can make it throw; the workbench's partial local regex is removed in favor of the single owner.

**Evidence:** rejection surface reproduced against the pinned dep before fixing (`History`/`history`/`'Plan`/`Plan'` all throw pre-fix); goldens use the REAL xlsx module (no mock) and round-trip the tab name; strip-before-reserved ordering covered by the `'History'` golden; mutation check per leg — dropping the reserved-name branch or the apostrophe strip each reddens 5 tests, mutation landing grep-confirmed first. Full required web-tests + vue-tsc green at head.

PR #4327, opened 2026-07-15
feat(web): W2 S3 — MetaRecordInspector shell (ARIA tab pattern; drawer → deprecated thin compat shell per OD-W2-7=b)

zensgit:
## Summary

- Lock: `docs/development/multitable-w2-unified-record-inspector-design-lock-20260714.md` §7 S3 row, §3.3, §6bis — new `MetaRecordInspector.vue` shell that absorbs `MetaRecordDrawer.vue`'s tablist + `activeTab` logic + header actions + lock banner (moved verbatim, byte-identical rendered output), and mounts the S1 (#4314) / S2 (#4324) extracted `MetaRecordFieldsPanel` / `MetaRecordHistoryPanel` as children.
- `MetaRecordDrawer.vue` is now a **deprecated thin compat shell** (OD-W2-7=b): its public props/emits are byte-unchanged, its entire template is `<MetaRecordInspector v-bind=... />` forwarding every emit 1:1. Barrel export at `apps/web/src/multitable/index.ts` **preserved** (plus a new `MetaRecordInspector` export, per Medium-4's "no audited out-of-repo-consumer inventory" caveat).
- `MultitableWorkbench.vue` switches to render `MetaRecordInspector` directly — a **2-line diff** (import + tag rename), props/emits identical.

## §6bis rulings applied this slice

- **OD-W2-1 = (a) tab**: horizontal tablist, internal `InspectorTab` union (`'details' | 'history'`) + `TAB_ORDER` array built to extend generically to `'comments'`/`'attachments'` in S4/S5 — the roving-tabindex/arrow-nav logic iterates `TAB_ORDER`, not hardcoded branches.
- **OD-W2-2 = (c) context-driven default**: `resolveDefaultTab()` is the single seam S4 extends (`commentId ? 'comments' : 'details'`) — today's comment deep-link flow (`resolveDeepLink` → `showComments`/`MetaCommentsDrawer`) is entirely outside this component's props, so there is no live `commentId` signal to branch on yet; the resolver has exactly one live branch (honest, not a fabricated readiness claim).
- **OD-W2-3 = (a) wide push**: unchanged — the shell keeps today's persistent side-panel flex-row placement (not overlay).
- **OD-W2-7 = (b) deprecated thin compat shell**: implemented as described above.

## ARIA tab pattern completion (lock §3.3)

- `div[role="tablist"]` → `button[role="tab"]` (unchanged div, not a literal `<ul>`), each carrying `aria-controls` → its `[role="tabpanel"]` id; the tabpanel carries `aria-labelledby` back. Both ids namespaced off Vue 3.5's `useId()` (same pattern as `AttendanceCalendarPolicyQuickAdd.vue`).
- **Roving tabindex**: exactly one tab carries `tabindex="0"` at all times (== `activeTab`); rest are `-1`.
- **Keyboard**: Left/Right move focus AND activate (automatic-activation model — a tablist, not a tree, so unlike `MetaSheetViewRail`'s deliberate move-only tree deviation, activation-on-arrow IS the APG default here); Home/End jump to first/last; wraps at the ends (Right past the last tab → first, and vice-versa).
- **Escape**: closes the inspector via the same `close` emit the header's × button uses. Guarded by `event.defaultPrevented` so it does NOT fire when a descendant already consumed Escape (verified against `MetaRichLongTextEditor`'s own Escape-to-cancel, which calls `preventDefault()`). Never inspects any other key, so mod+z / mod+y / `?` always bubble untouched to `MultitableWorkbench`'s own `onGlobalKeydown`.
- **Honesty clause** (lock §3.3): no drag-reorder, no typeahead claimed. `aria-controls` on the currently-INACTIVE tab points to an id not currently in the DOM — the v-if/v-else lazy-unmount panel model is inherited unchanged from S1/S2, a known tradeoff for tab widgets that lazy-render panel bodies, documented in the component's own header comment (not silently glossed over).
- **Focus ring**: `:focus-visible` + `--ms-color-primary` token (H4-2/#4281 lineage, same convention as `MetaSheetViewRail.vue`). **Not real-browser-verified in this PR** — jsdom can't render CSS; real-browser sweep is §8.3's remit, landing with the responsive S7 slice per lock §7.

### A genuine test-harness finding (documented on `onInspectorKeydown`)

Originally implemented as two separate `@keydown` bindings (root `onInspectorKeydown` for Escape, tablist-level `onTabsKeydown` for arrow nav) — matching the pre-shell drawer's own separation. Under repeated local runs this was **intermittently flaky** (~30-50% failure rate on the Escape-related tests). Root-caused via a minimal isolated repro (a throwaway two-`@keydown`-listener Vue 3.5.18 SFC, no other logic): two independently-bound `@keydown` listeners at different levels of the *same* render tree intermittently fail to both fire on a single bubbled event under this exact vitest/jsdom harness — reproduced deterministically in isolation (pure native DOM was 100% stable; introducing a second Vue-bound listener made it fail 70-90/80 iterations). This is a harness/runtime interaction, not a logic bug in this PR's code. Fix: **consolidated into ONE root-level keydown handler** that dispatches internally by `event.target`/`event.key` — verified stable across 20+ repeated full-spec runs after consolidating (vs. reproducibly flaky before). One test in the new spec (`does not swallow mod+z/mod+y/?`) also had to switch its ancestor-listener simulation from a second Vue-bound wrapper component to a plain `document.addEventListener` — for the same reason, re-introducing a two-Vue-listener tree inside the *test* itself reproduced the identical race.

## Frozen-baseline discipline (P2-2a / S1 / S2 pattern)

The 7 spec files that mount `MetaRecordDrawer.vue` directly are **byte-identical** (sha256-verified before/after) and green **unmodified** — they now exercise the shell → inspector delegation, which is exactly the compat contract this PR promises:

| File | sha256 unchanged |
|---|---|
| `meta-record-drawer-history-diff.spec.ts` | ✅ |
| `meta-record-drawer-i18n.spec.ts` | ✅ |
| `meta-record-drawer-restore.spec.ts` | ✅ |
| `multitable-record-drawer-button.spec.ts` | ✅ |
| `multitable-record-drawer-duplicate.spec.ts` | ✅ |
| `multitable-record-drawer-t5-migration.spec.ts` | ✅ |
| `multitable-record-drawer.spec.ts` | ✅ |

**Two categories of test files were deliberately modified** (neither is a "drawer spec" pinning drawer *behavior* — both are affected because `MultitableWorkbench.vue` now renders a different component):

1. **10 workbench-wiring specs** that `vi.mock('.../MetaRecordDrawer.vue', ...)` by import path purely as an *isolation technique* (stub/capture the component the workbench mounts, to test the workbench's own prop-wiring logic without fully rendering the drawer). Since the workbench now mounts `MetaRecordInspector`, these mocks were stale/inert — 3 of them (`multitable-workbench-drawer-button-wiring`, `multitable-workbench-restore-wiring`, `multitable-workbench-view`) actually **failed** once retargeted was skipped (their assertions depend on the capturing stub firing); the other 7 (`multitable-crossbase-workbench-wiring`, `multitable-workbench-{conflict-migration,manager-flow,import-flow,history-field-scope-wiring,permission-wiring,1672-1673}`) happened to still pass with the stale mock (their scenarios never render the record surface) but were retargeted anyway for correctness/honesty — an inert mock is a latent trap for a future Inspector-internals change. All 10 now `vi.mock('.../MetaRecordInspector.vue', ...)`, same capture/stub logic, same assertions.
2. `comment-affordance-color-consistency.spec.ts`'s `CONSUMER_FILES` list (RATIFIED lock guard) — swapped `MetaRecordDrawer.vue` → `MetaRecordInspector.vue`: the header comment-toggle button's `--active`/`--idle` CSS rule moved with the template (Drawer's own `<style>` block is now empty — nothing left to scan there).

## New spec: `multitable-record-inspector.spec.ts` (21 tests)

- Tab structure + ARIA pairing (4): tablist/tabpanel counts, default-active tab, aria-controls↔tabpanel↔aria-labelledby pairing, click-driven re-pairing.
- Roving tabindex invariant (3): mount, click-driven switch, arrow-key-driven switches.
- Keyboard Left/Right/Home/End (5): move+activate, wrap-around both directions, Home/End, preventDefault.
- Escape (4): bare emits close, modifier-guarded, defaultPrevented-guarded (proves no double-fire against a descendant consumer), does-not-swallow mod+z/mod+y/`?`.
- Panel-count conservation (1): never 0 or 2 across mount + several click/keyboard switches.
- HI-1 source scan (2): every `apiClient.<method>(` call is one of the 3 pre-shell-sanctioned subscription methods (positive control included — a broken regex would vacuously pass); no bare `fetch(`/`client.` call.
- Delegation (2): mounting the deprecated `MetaRecordDrawer` produces the identical ARIA structure + roving tabindex, and still supports keyboard nav through the delegation.

## CI wiring proof

Added `multitable-record-inspector` token to `apps/web/scripts/run-required-web-tests.sh` (bare basename; grep-verified no collision with any `multitable-record-*` token).

- **Executes** (script run in place both times, correct cwd this time — see note below): without the token, `337 files / 3931 tests` green; with the token, `338 files / 3952 tests` green — exactly the new spec's 21 tests, nothing else moved.
- **Bites**: mutated one assertion in the new spec (`'Details'` → `'MUTATED-WRONG-VALUE'`), re-ran the full script → exit 1, pinpointed the exact mutated test, everything else stayed green (`337 passed | 1 failed` files, `3951 passed | 1 failed` tests). Reverted via `cp` from a pre-mutation backup; re-ran clean → `338/338 files, 3952/3952 tests`, exit 0.
- *(Process note: an earlier attempt to run the script from a `/tmp` copy broke the script's own `cd "$(dirname "$0")/.."` self-relative-path assumption — `dirname` resolved to `/tmp`, so `npx vitest` picked up an unrelated global-cache vitest version and effectively hung. Re-ran correctly with copies placed alongside the real script in `apps/web/scripts/`; the counts above are from that corrected run.)*

## Mutation checks (performed + reverted, cp-based backup/restore)

- **Roving tabindex invariant**: temporarily changed `:tabindex="activeTab === tab.id ? 0 : -1"` to a constant `:tabindex="0"` → 4 tests in the new spec go **red** (`toHaveLength(1)` fails with 2). Reverted, re-verified 21/21 green.
- **aria-controls mismatch**: temporarily appended a wrong suffix to the `aria-controls` binding → 3 tests go **red** (including the delegation test, proving the pairing check propagates through the drawer shell too). Reverted, re-verified 21/21 green.

## Verification

- `vue-tsc -b --noEmit`: clean (0 errors) after every change in this PR.
- Frozen baseline: 7 files sha256-identical before/after, all pass.
- Broader regression sweep (24 pre-existing files named in S1/S2's own baseline scope + `multitable-workbench.spec.ts` + the 2 new panel specs + the new inspector spec = 23 files / 280 tests): stable across the final run.
- New spec: stress-verified stable across 20+ repeated `vitest run` invocations after the single-listener consolidation (was reproducibly flaky before).
- Full required script (`run-required-web-tests.sh`): 338/338 files, 3952/3952 tests, exit 0.

## Risks / follow-ups (not blocking this PR)

- Focus-ring CSS (`:focus-visible` + `--ms-color-primary`) is unverified in a real browser this slice — `multitable-browser-verify.yml` already path-triggers on `MultitableWorkbench.vue` (touched here for the tag swap) but its existing harness doesn't assert on this specific rule; real-browser focus-ring + responsive verification is explicitly §8.3/S7's remit per the lock.
- `MetaRecordDrawer.vue`'s barrel export is preserved per Medium-4 (no audited out-of-repo-consumer inventory) — actual retirement is a later, consumer-audited major cleanup, not scoped to W2.
- S4 (comments tab, `MetaCommentsPanel` extraction, collapsing the second right-drawer) and S5 (attachments tab) are separate, not-yet-authorized slices per lock §7.


zensgit:
Gate verdict (MERGE_WITH_NITS) addressed at head `6891ad2d3`:

**P2 (arrow-scoping guard untested) — CLOSED.** Two regression goldens added to `multitable-record-inspector.spec.ts` (now 23 tests):
1. ArrowRight/ArrowLeft dispatched from the fields panel's **text input** inside the active tabpanel → active tab unchanged AND event not preventDefault-ed (caret movement survives).
2. ArrowRight/ArrowLeft/Home/End dispatched from the **tabpanel element itself** (`tabindex="0"`, a direct keyboard-focus target) → same assertions.

**Mutation evidence** (performed + reverted via cp backup): removing `if (!withinTablist) return` from `onInspectorKeydown` → exactly the 2 new tests red (`2 failed | 21 passed`), everything else green — reproducing the gate's neuterable-with-green finding, then restored and re-verified 23/23 green (stable across 5 repeated runs).

**NIT — DONE.** Machine-readable `@deprecated` JSDoc tag added on the `MetaRecordDrawer` barrel export in `apps/web/src/multitable/index.ts`, pointing at `MetaRecordInspector`.

**Rebase onto `ddbb22408`** (was 4 behind): clean. Main's G-10 #4323 additions to `multitable-workbench-view.spec.ts` merged cleanly with the S3 mock retarget (79/79 green). Post-rebase: vue-tsc clean; full `run-required-web-tests.sh` → **338 files / 3969 tests, exit 0** (+2 = the new goldens, +15 = main's own growth picked up by the rebase); the 7 frozen drawer specs remain git-blob-identical to origin/main.

PR #4309, opened 2026-07-15
feat(multitable): W0-1 v3.5 — Lane L3 chain-integrity core (causal seq + loud markers + strict generation-aware contiguity)

zensgit:
**W0 build — Lane L3 (chain-integrity core).** Implements the `seq`-based chain-integrity core of the W0-1 correction design (#4262 v3.5).

**Status (owner 2026-07-15): DIRECTION SELECTED (`seq`), formal ratification of #4262 PENDING a full containment PASS** — this is NOT built on a formally-ratified lock yet. Draft; all recovery flags OFF; not armed.

## Scope
seq migration (`meta_record_chain_seq` on both tables) + drop cross-generation marker unique + loud marker + seq-ordered per-generation contiguity + C2 monotonicity + C3 deleted-enumeration, **behind `MULTITABLE_HISTORY_CONTIGUITY_STRICT` (default-off = byte-identical #4269)**. Independent Opus gate = **MERGE_CLEAN, 0 P1/0 P2** (`/tmp/pr4309-review-claude-20260715.md`).

## Trust boundary (owner 2026-07-15) — L3 alone ≠ W0 built-to-trust
The legacy `seq` **backfill numbers the revisions and markers tables separately (row_number per table), so values can overlap ⇒ the backfilled seq is NOT causal evidence.** The real trust origin is: **L4 all-writer fence → L5 time-anchored baseline checkpoint → strict enablement.** L3 (this PR) is the foundation only; strict mode is not trustworthy until L4+L5 land and the checkpoint cutover runs on a containment-PASS host.

**Landing order:** merge #4316 → staging re-normalize → containment PASS → ratify #4262 → **merge this L3 → then L4 + L5.** Held until then.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4328, opened 2026-07-15
docs(multitable): W0-1 v3.6 unified revision design lock (PROPOSED — owner-gated; merges C2 lock + v3.5 per the twice-issued ruling)

zensgit:
**Status: PROPOSED — NOT ratified.** Docs-only, zero runtime. Does NOT modify #4262 / #4309 (parallel session's — this doc exists so the two design lines converge instead of forking).

The owner has twice issued the same verdict on the v3.5 arc (recorded as C2 lock §7 via #4325): **seq half-ratified** as the strict-mode ordering primitive; **v3.5 not ratified**; two Highs stand. This is the unified revision the ruling demanded, with the four requirements as hard sections:

- **§1** — the two Highs as **fail-first golden specs**: G-H1 "early-start/late-lock write selected by T-rebuild" (seq passes, `created_at<=T` picks a not-yet-happened v5); G-H2 "T lands in an old generation" (terminal-gen-only strict precheck passes over a gen-1 hole; wrong-data resurrect).
- **§2 (req 1 + knife 1)** — T→seq boundary: **selection is always a seq-prefix (`seq <= B`); wall-clock never enters the selection predicate.** Dual resolvers: R1 event/batch anchor (#3749 picker, zero wall-clock) + R2 post-fence `effective_at` (`clock_timestamp()` under the L4 fence = cross-writer linearization; prefix-closure rule for NTP step-backs). Plus the **C6 checkpoint T→seq truncation contract** the owner flagged missing (`trusted_since_seq` allocated under the fence at cutover; `B >= trusted_since_seq` invariant; baseline ⊕ prefix-replay; pre-checkpoint/backfilled seq structurally never load-bearing).
- **§3 (req 2 + knife 2)** — asOf-generation precheck: target-generation (A) vs conservative all-restorable-generations (B); **recommend A** as the end state (B acceptable as an interim step).
- **§4 (req 3)** — strict enablement hard-preconditions checklist. Includes a **verified factual correction to v3.5 §3**: on current main, reset-execute holds only the disjoint `PIT_RECOVERY_LOCK_NS` two-int lock (`univer-meta.ts:10481`), NOT `acquireAutoNumberSheetWriteLock`; revert-execute holds no fence at all — canonical-fence convergence is a new L4-0 precondition. 19-family writer matrix on live `bf52b9513` anchors. L5 active-checkpoint becomes an **in-code fail-closed runtime self-check**, not a runbook promise.
- **§5 (req 4)** — exact bigint end-to-end (SQL-native compare, string at the driver/JSON boundaries, BigInt in TS, `/^[0-9]+$/` fail-closed), citing the float64 **double-precedent**: #4269's `epoch*1e6+version` packing + #4309's `toNumber(seq)` folding. Production-magnitude goldens at 2^53 and near 2^63.
- **§6** — honest slice plan: **~9–13.5 pw** remaining beyond #4269+#4309-Draft, of which the two new knives (L6 T→seq boundary + L7 asOf-generation) ≈ 3–4.5 pw; L3-R1/R2 land in #4309's lane, L4/L5/L6/L7/L8 are new lanes.
- **§7** — minimal ODs: T-input surface (dual resolvers vs anchored-only), asOf-generation end state, fence-convergence form.

§0 carries the explicit keep/supersede ledger against both #4262 v3.5 and the C2 lock (incl. superseding the C2 lock's "NULL-seq = trust seam" — the ONLY trust boundary is the L5 checkpoint).

### DIFF
--- a/docs/development/multitable-w0-1-v36-unified-revision-design-lock-20260715.md
+++ b/docs/development/multitable-w0-1-v36-unified-revision-design-lock-20260715.md
@@ -1,0 +1,250 @@
+# W0-1 (v3.6) — 统一修订锁：seq 主序 + T→seq 边界 + asOf-generation 预检 + L4/L5 硬前置 — DESIGN LOCK（PROPOSED — NOT ratified）
+
+- **Status（2026-07-15）**: **PROPOSED — NOT ratified.** owner 已**两次**对平行 v3.5 弧（#4262 锁 + #4309 Draft impl）作同一裁决（录于 C2 锁 §7，#4325 = `9a6fc71f9`）：**seq 半认可 = strict 模式链内排序原语**；**v3.5 不 ratify 为完整 C2/C3/C6 设计**；两个 High 证伪现有 MERGE_CLEAN 门禁；修订版须补齐 4 项硬要求 + 两把新刀。本文即该**统一修订**：合流 C2 锁（#4320 merged + §7 裁决）与 #4262 v3.5 的仍有效部分，把 owner 4 项要求写成**硬章节**。owner ratify 后才动工。
+- **Provenance**: `origin/main`（本 worktree 基点 `bf52b9513`，含 #4320/#4325）；#4262 v3.5 锁读自其分支 `claude/w0-1-corrected-generation-design-lock-20260713`；#4309 Draft diff 读自 `claude/w0-l3-chain-integrity-core-20260715`。**本文不改动 #4262/#4309**（归平行 session）；本文的作用 = 让两条设计线**收敛而非分叉**。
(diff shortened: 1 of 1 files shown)


zensgit:
Fix-forward review result: the merged v3.6 document remains PROPOSED and is not safe to ratify as written. Two load-bearing issues were confirmed against current main: (1) an in-transaction clock_timestamp/effective_at precedes COMMIT and therefore cannot be an exact wall-clock recovery boundary; (2) Reset/Revert compute and verify their target before acquiring the destructive fence, so an in-window healthy writer can leave execute applying a stale set. The mutable MAX(seq) batch resolver also lacked a database-enforced transaction endpoint. Draft #4331 replaces the v3.6 file with one v3.7 design: sealed committed-operation endpoints, exact event anchors for destructive recovery, full in-fence recomputation, completed writer matrix, target-generation/checkpoint/bigint contracts, and no runtime/flag authorization. Please treat #4331 as the sole fix-forward ratification target; this merged PR remains historical provenance only.

PR #4334, opened 2026-07-15
feat(automation): P2 durable-delivery S2-b — dispatch loop (registry + tick), flag-gated

zensgit:
**Base: `main`** (S2-a #4322 landed as `336732b5f`). **Additive, flag-gated, zero behavior** — `startDurableDispatchLoop` refuses to start while `AUTOMATION_DURABLE_DELIVERY_ENABLED` is OFF, nothing calls it yet. Head `279c5edc9`.

## Design (current)
Per-slot claim → adapter under heartbeat + bounded timeout → fence-CAS resolve. The tick claims ONE reclaimable row per execution slot and processes it immediately (never a batch), so a row is only attempted when it is about to run. A row already at the ceiling is poisoned at claim, counted, and scanned past (not dispatched).

## Owner reviews — all findings fixed
- **Round-2 (3 P1 + 2 P2):** malformed verdict dead-lettered → `normalizeAdapterVerdict` (retry, never poison, hostile-getter safe); hung adapter froze the tick → bounded `adapterTimeoutMs` + `AbortSignal`; optional/unsafe telemetry → required exception-safe sink; unvalidated numeric knobs → bounded safe-int validation before any work.
- **Round-3 (2 P1 + 2 P2):** tail-row poison-without-execution → **per-slot claim** (subsumes the round-2 pre-call renew, removed); heartbeat later than a short lease → `heartbeatMs = leaseMs/3` + min `leaseMs ≥ 1000`; heartbeat DB error masked as lostLease → required `onHeartbeatError` + distinct count; one-sided backoff override → **effective**-defaults inversion check.
- **Round-3 adversarial P3:** sub-tick backoff busy-spun a rescheduled row within a tick → `retryBaseMs` floored at 1s.
- **Round-4 (P2 + P3):** **claim-time-poison head-of-line block** — an exhausted oldest row was poisoned at claim, `claimDueConsumers` returned empty, and the loop mistook that for "no work" and broke, starving healthy rows behind it (and hiding the terminal transition from telemetry). Fixed: `claimDueConsumers` reports claim-time poisons via a new `onClaimTimePoison`; the loop scans past a poisoned head to the next slot (within budget), counts them (`result.claimTimePoisoned`), and surfaces them. Plus P3 doc-drift corrections (removed pre-call renew reference; heartbeat renew error → `onHeartbeatError`, not `onTickError`; the `attempts == adapter invocations` overclaim reworded for the crash-safe claim-time attempt bump).

- **61/61** across the P2 cluster (outbox schema 14 + S2-a claim 13 + **S2-b loop 34**). `tsc --noEmit` 0 errors.
- Every guard **mutation-proven load-bearing** — reverting per-slot claim kills 5 tests; the head-of-line fix has two mutations (loop scan-past + claim poison-reporting). Two-point CI wired.
- `renewConsumerLease` and `onClaimTimePoison` added to the S2-a claim engine additively (return type unchanged); its 13 claim tests still pass.
- Independent adversarial pass on the round-3 head confirmed correctness under a 3-worker / pathological-backoff probe and surfaced the busy-spin P3 (since fixed); it was cut short by a session limit, re-runnable after reset. The owner's re-review is the authoritative gate.

## Not in this slice
S3 (#4335), S4-a (#4336), S4-b/S5/S7 (#4337), S6, ledger/FWB/attachments. Flags stay OFF until the 8-scenario acceptance.

**For review — not self-merging runtime. S2-b HOLD stays until owner re-review; #4335 not retargeted.**

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -536,6 +536,7 @@ jobs:
             tests/integration/multitable-tombstone-retention-realdb.test.ts \
             tests/integration/multitable-automation-outbox-schema-realdb.test.ts \
             tests/integration/multitable-automation-dispatcher-claim-realdb.test.ts \
+            tests/integration/multitable-automation-dispatch-loop-realdb.test.ts \
             tests/integration/multitable-attachment-readgate.security.test.ts \
             tests/integration/multitable-ai-write-provenance-batch-grouping-realdb.test.ts \
             tests/integration/multitable-automation-branch-local-wait.test.ts \
--- a/packages/core-backend/src/multitable/automation-durable-dispatch-loop.ts
+++ b/packages/core-backend/src/multitable/automation-durable-dispatch-loop.ts
@@ -117,6 +117,10 @@ export const DEFAULT_ADAPTER_TIMEOUT_MS = 30_000
 // Explicit upper bounds for the loop's own knobs (the claim/resolve knobs are bounded inside S2-a).
 const MAX_MS_24H = 86_400_000
 const MAX_INTERVAL_MS = 3_600_000 // 1h — an interval beyond this is almost certainly a misconfiguration
+// The claim engine (S2-a) accepts leaseMs >= 1, but a HEARTBEATED dispatch needs the lease long enough that
+// the heartbeat (leaseMs/3) fires and a renew round-trip completes BEFORE the lease expires — otherwise a
+// competitor reclaims a still-running delivery (#4334 review P1-B). 1s is a conservative floor for that margin.
+const MIN_LOOP_LEASE_MS = 1_000
 
 /** Reject fractional, non-finite, out-of-safe-range, or out-of-bounds numeric params before they do anything. */
 function requireBoundedInt(name: string, value: number, min: number, max: number): void {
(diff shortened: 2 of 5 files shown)


zensgit (commented):
Third review of head `fd7a08556126ea73eadb162db29178de7892244a`: **REQUEST CHANGES**.

### [P2] Claim-time poison creates head-of-line blocking and hides the dead-letter transition

`runDispatchTick` asks `claimDueConsumers` for one row and treats an empty return as "nothing reclaimable" (`break`). But `claimDueConsumers` can select an exhausted row, transition it to `dead_letter`, and then filter it out of the returned rows. Those are two different outcomes collapsed into the same empty array.

I reproduced this against fresh PostgreSQL 15 with an oldest `pending` row at `attempts=maxAttempts` followed by a healthy `pending` row, using `batchSize=2`. The tick produced `claimed=0`, `poisoned=0`, and zero adapter calls; the first row became `dead_letter`, while the healthy row remained `pending`. Therefore every exhausted row at the head consumes an entire tick, and a backlog of N exhausted rows can delay healthy delivery by N polling intervals. The claim-time dead-letter is also absent from tick telemetry.

Please make the one-slot primitive distinguish `no candidate` from `terminalized exhausted candidate` (or atomically advance the poison row and select the next dispatchable row), continue within a bounded per-tick budget, and count claim-time poison. Add a real-DB regression with oldest exhausted + newer healthy, asserting the healthy adapter runs in the same `batchSize=2` tick and the poison is observable.

### [P3] Source contracts still describe behavior removed by this head

- `automation-durable-dispatcher.ts` still says `renewConsumerLease` is called immediately before every adapter invocation, but the pre-call renew was removed.
- `automation-durable-dispatch-loop.ts` says renew DB errors propagate to `onTickError`, while the implementation catches them and routes them to `onHeartbeatError`.
- The `attempts == adapter invocations` wording is too strong and contradicts claim-time crash safety: a process crash after claim and before `handle()` intentionally consumes an attempt. The valid guarantee is that serial tail rows are not pre-claimed/pre-burned.

Independent checks on this exact head: the official real-DB spec passed 33/33 on fresh PostgreSQL 15, and `pnpm --filter @metasheet/core-backend exec tsc --noEmit` passed. Those checks do not cover the blocking composition above.

zensgit (commented):
Fourth review of head `279c5edc988a7f1df996cbb94eb222acb1773f5d`: **APPROVE — no P1/P2 findings**.

The round-3 head-of-line defect is closed at the behavior level. `claimDueConsumers` now reports rows terminalized at claim, and `runDispatchTick` distinguishes that outcome from a genuinely empty queue, consumes one bounded slot, and continues. Against fresh PostgreSQL 15, the composed S1/S2-a/S2-b cluster passed 61/61 and `tsc --noEmit` passed. I also independently mutated the scan-past branch back to `break`; the new real-DB regression went RED at the healthy-row assertion, so the load-bearing loop guard is discriminating.

Non-blocking P3 follow-ups:

1. Before S5/boot activation, route claim-time poison through the required `DispatchLoopObserver` (and `safeInvoke`), or otherwise make its binding an explicit activation assertion. Today `startDurableDispatchLoop` can start without `onClaimTimePoison`; the per-row option and optional `onTick` provide a seam, but do not guarantee the dead-letter signal is observed. This is not a merge blocker for S2-b because the flag is OFF and there is still no production caller.
2. The new regression claims to force the poisoned row oldest, but updates `meta_automation_outbox.created_at`; the claim CTE actually orders by `meta_automation_outbox_consumer.updated_at`. Reversing that SQL ordering left the test GREEN. Pin the consumer `updated_at` values explicitly so the setup proves the poison is at the head rather than relying on insert timing.
3. Finish the comment sweep: line 138 still says `attempts == adapter calls` (not true across crash-after-claim), the top-level backoff contract still says a row cannot be reclaimed in the same tick (only the >=1s elapsed floor is guaranteed), and the optional `onTick` metric list omits `claimTimePoisoned`.

The PR is currently mergeable but behind main. Rebase and re-green before landing; approval assumes the reviewed product diff remains unchanged.

PR #4335, opened 2026-07-15
feat(automation): P2 durable-delivery S3 — versioned routing manifest + bidirectional completeness assertion

zensgit:
**Stacked on #4334 (S2-b)** — retarget after parents land. Pure data + assertions; nothing reads it until S4 enqueues. Zero behavior.

- **`ROUTING_MANIFEST_V1`** — the lock's ratified full set (#4203 §283-291) grounded in the real subscription surfaces (`automation-service.ts:892-936`, `index.ts:2119`, `webhook-event-bridge.ts:36-39`); frozen (a v1 row is dispatched per v1 forever).
- **`assertManifestCompleteness`** — bidirectional, registry as the single enumeration source: manifest-key-without-adapter (rows park forever) AND adapter-not-routed (dead config) both throw naming exact keys. The un-enumerable third direction (anonymous bus closures) is closed structurally in S5.
- **`SUPPORTED_MANIFEST_VERSIONS`** — the rolling-deploy anchor; K-expansion stays activation-gated because the producer side cannot be defended at runtime (#4203 §240-243).

**Verification:** unit spec runs in the required no-DB `test` lane (7/7): pins the exact v1 routes/universe, frozen-ness, unknown→undefined, both directions with exact-key messages. Mutation: dropping direction 2 reds exactly its tests.

Flag OFF; **for review — not self-merging runtime.**

### DIFF
--- a/packages/core-backend/src/multitable/automation-routing-manifest.ts
+++ b/packages/core-backend/src/multitable/automation-routing-manifest.ts
@@ -1,0 +1,107 @@
+/**
+ * P2 durable-delivery — slice S3: the versioned routing manifest + startup completeness assertion.
+ *
+ * The manifest maps each event family to the FULL set of consumer_keys that must receive a durable delivery
+ * row. The producer (S4) expands an event into its `(outbox_id, consumer_key)` rows AT ENQUEUE, stamping
+ * `manifest_version` — a row enqueued under v1 is dispatched per v1 forever and never re-interpreted against
+ * a live manifest (#4203 §manifest). Content is the lock's ratified v1 full set (§283-291), grounded in the
+ * actual subscription surfaces (`automation-service.ts:892-936`, `index.ts:2119` projection,
+ * `webhook-event-bridge.ts:36-39`):
+ *
+ *   approval.{approved,rejected,revoked,cancelled} → approval-bridge, approval-trigger, approval-projection
+ *   approval.task_created                          → approval-task-trigger
+ *   multitable.record.{created,updated,deleted}    → automation-record-trigger, webhook-event-bridge
+ *   multitable.comment.created                     → webhook-event-bridge
+ *   form.submitted                                 → automation-record-trigger
+ *
+ * Completeness is asserted BIDIRECTIONALLY at startup with the adapter registry as the single enumeration
+ * source (#4203 §293-300): (a) every consumer_key the manifest routes to has a registered adapter — a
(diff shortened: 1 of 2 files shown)


zensgit (commented):
## Codex review — REQUEST CHANGES

Reviewed exact head `442067ff82316c8abe03ea00afa79e4e3441bf47`.

### [P2] The v1 manifest is only shallow-frozen

`ROUTING_MANIFEST_V1` and `routes` are frozen, but every route value remains a mutable array, and the exported `APPROVAL_COMPLETION_CONSUMERS` is the same array referenced by four routes (`automation-routing-manifest.ts:43-60`). A live process can therefore mutate the routing contract without changing `manifest_version`:

```text
routeFrozen=false
before=[approval-bridge, approval-trigger, approval-projection]
after=[approval-bridge, approval-trigger]
```

This defeats the stated invariant that v1 is immutable and new routing requires a new version. Deep-freeze every route array (including the shared approval array), and add a test that all route values are frozen and reject/ignore mutation.

### [P2] The “exact v1 event families” test does not pin the event-family set

The test checks each expected route and the distinct consumer-key universe, but never asserts `Object.keys(ROUTING_MANIFEST_V1.routes)`. Adding an extra v1 route that reuses an existing consumer key leaves all seven tests green, allowing a v1 contract expansion without a version bump. Pin the sorted event-key set (or the full manifest snapshot) and mutation-prove that adding/removing a route is RED.

Verification performed: focused unit spec `7/7`, backend `tsc --noEmit` clean, and a runtime mutation probe proving the nested array is mutable. GitHub checks are green, but these are contract blockers; do not merge this head.

zensgit (commented):
## Codex re-review — APPROVE

Re-reviewed exact head `b46387fc5cb1e353a297d6824b919f239b0b3bab`.

Both prior P2 findings are closed:

- every route value, including the shared approval-completion array, is now frozen at runtime; replaying the prior `.pop()` probe throws and leaves the route byte-for-byte unchanged;
- the unit contract now pins the exact v1 event-family key set, so adding/removing/renaming a route cannot hide behind the unchanged consumer-key universe.

Independent verification: focused manifest spec `7/7`, backend `tsc --noEmit` clean, prior runtime-mutation repro now fails closed, all GitHub checks green, and the PR is currently `CLEAN`/mergeable. No remaining P1/P2 findings; this head is approved to land.

PR #4336, opened 2026-07-15
feat(automation): P2 durable-delivery S4-a — producer-side atomic enqueue (helper)

zensgit:
**Base: `main`. Additive, flag-neutral** — S4-a producer-side atomic enqueue helper; no callers until S4-b wires the produce sites behind `AUTOMATION_DURABLE_DELIVERY_ENABLED`. Head `f4e81a57b`.

## What lands
`enqueueOutboxEvent(trx, event)` — one `meta_automation_outbox` row + one pending `meta_automation_outbox_consumer` row per manifest-expanded key, inside the caller's transaction; unrouted family / blank identity / bad depth are hard errors; `manifest_version` stamped. Plus the shared **`pg-transaction-guard.ts`** (also ships byte-identical on #4340; second-lander rebases to a no-op).

## Review fixes
- **P1 round-1 (atomicity by calling convention)** → **round-2 (marker forgeable)**: the `isTransaction: true` marker was a public boolean any caller could forge onto a Pool — the half-state repro (outbox=1/consumer=0) still worked. Now the guard asserts the **database's own transaction state**: two `pg_current_xact_id()` probes through the handle must observe the SAME transaction. A Pool (statements on arbitrary connections), a bare AUTOCOMMIT client, `withTransaction`'s non-transactional fallback, and a **forged marker over any of those** all fail the probe **before any write**. The brand remains compile-time documentation only.
- **P2 (shared-DB test pollution)**: the e2e drains a **run-unique** test manifest (`ubridge/utrigger/uproj_<run>`), default batch — it can only claim its own rows.

- **8/8**: commit fan-out; ROLLBACK-zero-rows; **pool + forged-marker + bare-autocommit negatives** (each rejected, zero rows); **transaction positive control** (consumer-INSERT failure rolls the outbox row back — no orphan); unrouted/blank/depth hard errors; e2e per-consumer independence. `tsc --noEmit` 0.
- The probe is **mutation-proven** (neutering `assertInTransaction` reds the negatives). Two-point CI wired.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -548,6 +548,7 @@ jobs:
             tests/integration/multitable-automation-outbox-schema-realdb.test.ts \
             tests/integration/multitable-automation-dispatcher-claim-realdb.test.ts \
             tests/integration/multitable-automation-dispatch-loop-realdb.test.ts \
+            tests/integration/multitable-automation-outbox-enqueue-realdb.test.ts \
             tests/integration/multitable-attachment-readgate.security.test.ts \
             tests/integration/multitable-ai-write-provenance-batch-grouping-realdb.test.ts \
             tests/integration/multitable-automation-branch-local-wait.test.ts \
--- a/packages/core-backend/src/multitable/automation-outbox-enqueue.ts
+++ b/packages/core-backend/src/multitable/automation-outbox-enqueue.ts
@@ -10,11 +10,11 @@
  * construction. This applies to EVERY manifest event family, not only approval-completion (§328-333).
  *
  * Hard rules:
- *   - `trx` MUST be the transaction the source change is being written in. This is MACHINE-ENFORCED: the param
- *     is a `TransactionalQueryable` and a runtime guard rejects anything without the `isTransaction` marker, so
- *     a pool can never reach the INSERTs (passing one would break atomicity — the outbox row would commit
- *     before a failing consumer INSERT could roll it back). The S4-b call sites pass the same txn as the
- *     source write.
+ *   - `trx` MUST be the transaction the source change is being written in. This is MACHINE-ENFORCED against
+ *     the DATABASE's own transaction state (pg-transaction-guard: two `pg_current_xact_id()` probes must
+ *     observe the SAME transaction) — a pool, an autocommit client, and a FORGED `isTransaction` marker all
+ *     fail the probe, so a half-committed enqueue (outbox row without its consumer fan-out) is
+ *     unrepresentable. The S4-b call sites pass the same txn as the source write.
  *   - An event type the manifest does not route is a HARD ERROR — silently enqueueing zero rows would be a
  *     silent miss, exactly what the lock forbids.
  *   - `eventId` is the stable ORIGINAL-event identity (dedup basis + outbound-idempotency seed). Validated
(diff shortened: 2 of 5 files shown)


zensgit (commented):
## Head-scoped review: `dcc6f97d42bfa58926176236d63564f902c9d61d`

**REQUEST CHANGES.** The happy-path implementation is coherent, but the transaction boundary that gives this helper its reason to exist is currently advisory rather than enforced.

### [P1] `enqueueOutboxEvent` accepts a pool and can persist a half-enqueue outside the source transaction

`packages/core-backend/src/multitable/automation-outbox-enqueue.ts:13-15,58-91` types `trx` as the generic dispatcher `Queryable`. Both a checked-out transaction client and the ordinary pool satisfy that interface. The comment says transactionality is "not machine-enforceable at this seam", but a non-transactional call is accepted and silently destroys the atomic-outbox guarantee: the outbox INSERT commits independently of the source mutation, and the consumer INSERT is another autocommit statement.

Fresh PostgreSQL 15 proof, passing the accepted pool shape and forcing the second INSERT to fail with duplicate consumer keys:

```json
{"insertFailed":true,"outboxRows":1,"consumerRows":0}
```

That leaves exactly the invalid state the helper claims cannot exist: a committed outbox row with no delivery rows. Even when both INSERTs succeed, a pool call still commits them independently of the source write. This repo already has a fail-closed precedent for the same unexpressed precondition (`assertTransactionalQuery` in `side-door-delete-trash.ts`). Require a transaction-only capability/brand or a runtime transaction assertion, and add a negative control proving a pool call is refused with zero rows plus a positive transaction control. S7 call-site tests remain useful, but they do not make this reusable seam fail-closed for future callers.

### [P2] The end-to-end spec claims unrelated shared-DB rows and mutates their attempts/state

`packages/core-backend/tests/integration/multitable-automation-outbox-enqueue-realdb.test.ts:114-145` registers the real approval consumer keys and runs `batchSize: 500`. For an outbox row not owned by this test, the adapter returns `retryable_failure` with the comment "leave for the owning suite". But the row has already been claimed: this increments `attempts` and reschedules it as `in_progress`; at the ceiling it can be claim-time-poisoned.

Constructed probe with one foreign approval outbox followed by this test's own row:

```json
[{"status":"in_progress","attempts":1},{"status":"in_progress","attempts":1},{"status":"in_progress","attempts":1}]
```

Use a test-only manifest with run-unique consumer keys (the enqueue helper already supports manifest injection), register only those keys, and assert a foreign pending row remains byte-for-byte unchanged. Do not claim real shared consumer-key rows to discover whether they belong to this test.

### Verification

- Fresh PostgreSQL 15 focused spec: **6/6**.
- Backend `tsc --noEmit`: clean; two-point CI wiring present.
- The two probes above exercise cases the current six tests do not cover.
- GitHub required checks are green on this head, but the branch is currently **2 commits behind** `origin/main` (`6b7917b8a`), not up to date.

zensgit (commented):
## Head-scoped re-review: `0ef6b5ae83130a5ca7f9c9701dbf35403b86fa1f`

**REQUEST CHANGES — one P1 remains.** The shared-DB isolation P2 is fixed.

### [P1] The transaction guard is still forgeable, so atomicity still rests on caller convention

`TransactionalQueryable` is a public structural interface whose only proof is the caller-supplied boolean `isTransaction: true` (`automation-outbox-enqueue.ts:33-43,72-83`). The test itself constructs that marker by hand. Nothing binds `query` to a checked-out client inside `BEGIN`, and nothing prevents a pool-backed wrapper from satisfying the guard.

Fresh PostgreSQL 15 probe against this exact head, using the real helper and a duplicate consumer key to fail the second INSERT:

```json
{"acceptedForgedMarker":true,"secondInsertFailed":true,"outboxRows":1,"consumerRows":0}
```

The half-state from the original finding is therefore still reachable:

```ts
const forged = {
  isTransaction: true,
  query: (sql, params) => pool.query(sql, params),
}
await enqueueOutboxEvent(forged, event, duplicateKeyManifest)
```

The first outbox INSERT auto-commits through the pool, the consumer INSERT fails, and the orphan remains. A boolean marker narrows accidental unmarked calls but does not machine-enforce a transaction.

**Required close:** prove actual transaction state at this seam (the repository already has the `assertTransactionalQuery` / two-`txid_current()` precedent), or use a helper-owned/centrally issued transaction capability that an ordinary pool wrapper cannot construct. Add the decisive negative control above: a pool wrapper carrying `isTransaction:true` must be rejected before any write. Keep S7 per-producer tests to prove the outbox is also in the same transaction as each source mutation.

### Confirmed fixed

The E2E spec now uses run-unique consumer keys. I seeded a foreign `approval-bridge` row, ran only the E2E case, and it remained exactly `pending:0`; the test no longer claims sibling rows.

Verification: focused real-DB spec **8/8**, exact CI typecheck clean, current GitHub checks green. The branch is 4 commits behind current main but a local merge-tree is conflict-free; rebase/re-green is still needed after the P1 fix.

PR #4339, opened 2026-07-15
feat(multitable): W0-1 v3.7 — Lane L3 chain-integrity core, rebuilt per #4331 §9 (causal seq + all-generations contiguity + C3)

zensgit:
## Summary

Rebuilds the W0-1 Lane L3 chain-integrity core per the owner-ratified **#4331 v3.7 §9** design lock, correcting the superseded v3.5 draft (#4309). This is a **default-off implementation slice** — no recovery flag is enabled, nothing here arms strict mode, Revert/Reset, or touches any host.

Implements (§9.3, §9.7 + owner v3.6 Highs):
1. **Migration** (`zzzz20260715160000_add_meta_record_chain_seq`): one shared `meta_record_chain_seq` BIGINT sequence backing a `NOT NULL seq` column on **both** `meta_record_revisions` and `meta_record_version_markers`; legacy backfill via `row_number() OVER (ORDER BY created_at,version,id)` **per table** (display/order assistance only — explicitly **never** a trust signal, per §9.3); drops the cross-generation `UNIQUE(sheet_id,record_id,version)` on the markers table (it silently swallowed a resurrected generation's marker). Proven on a fresh-DB full migrate **and** against a populated-DB backfill scenario.
2. **Exact bigint end-to-end** (§9.7): every seq comparison uses native SQL `ORDER BY` (bigint) or exact `BigInt(...)`/string-equality in TS — `Number(seq)`, `parseInt`, unary `+`, subtraction comparators, and JSON-number seq fields are never used. Production-magnitude tests (values > 2^53, where `Number()` collapses distinct decimal strings) are mutation-proven both as pure unit tests and real-DB goldens.
3. **Generation-aware contiguity, seq-ordered**, behind `MULTITABLE_HISTORY_CONTIGUITY_STRICT` (default OFF): `precheckSheetHistoryIntegrityStrict` + `checkAllGenerationsContiguity`, generation = count(create) over seq order.
4. **Target-generation check, not terminal-only** (owner High-2): the conservative "all generations reconstructable in scope" option v3.7 §9.3 explicitly permits — every generation is validated, so a hole in an *older* generation refuses even when the terminal generation is completely clean.
5. **Deleted/trash chain enumeration (C3)**: records not live (chain-only or in `meta_records_trash`) are also enumerated and validated.
6. **Dup/illegal-seq fail-close**: a within-generation duplicate ⇒ `chain_corrupt` (new, strict-only reason); illegal/missing/out-of-range seq ⇒ fail-closed `comparator_error`, never coerced to zero.

`record-history-service.ts`'s `recordVersionMarker` drops its `ON CONFLICT DO NOTHING` — the constraint it targeted no longer exists, so a genuine write conflict now fails the enclosing lock/unlock transaction loudly instead of swallowing silently ("loud marker").

## Deferred (later lanes, not built here)

- §1.2 sealed operation-endpoint ledger (L6)
- §1.3 exact recovery anchor / API (L6)
- L4 all-writer canonical fence
- L5 trust checkpoint
- C2 time-monotonicity (remains on the pre-existing deferred docket; not required by this lane's scope)

## Verification

- Fresh-DB full migrate: proven (all migrations apply cleanly, including this one) + a separate populated-DB run proving the per-table backfill + sequence bootstrap-sync is correct and collision-free for subsequent writes.
- All new goldens pass: 27/27 pure unit tests (`tests/unit/multitable-history-contiguity-strict-seq.test.ts`) + 10/10 real-DB goldens (`tests/integration/multitable-history-contiguity-strict-seq-realdb.test.ts`, wired into `plugin-tests.yml`).
- **Mutation-proven** (performed locally, not shipped): reverting the all-generations walk to terminal-only reds both the target-generation unit tests and the real-DB golden; replacing the exact `BigInt` seq comparator with `Number()`+subtraction reds both exact-bigint goldens; weakening the duplicate-occupant threshold reds the dup-within-generation goldens.
- Full existing history/revert/reset/lock/restore real-DB regression swath (34 files, 354 tests) + full unit suite (393 files, 5463 tests) + `pnpm --filter @metasheet/core-backend type-check` all green — flag-off byte-identical to #4269 is confirmed by the untouched non-strict function body and this pre-existing suite passing unchanged.
- One pre-existing real-DB test (`multitable-history-incomplete-precheck-realdb.test.ts`) needed a hygiene fix: it reused a record id across two tests without clearing `meta_record_version_markers`, previously masked by the now-removed `ON CONFLICT DO NOTHING`. Fixed by clearing markers every test (matching the sibling contiguity test file's existing convention).
- Confirmed no `Number(seq)`/`parseInt(seq)` and no test-time `setval` on the shared production sequence anywhere (grep-verified); the migration's one-time `setval` is a bootstrap sync distinct from the forbidden test-time pattern (P2-C).

## Flags

`MULTITABLE_HISTORY_CONTIGUITY_STRICT` — default OFF, not armed. `MULTITABLE_ENABLE_PIT_RESET` / `MULTITABLE_ENABLE_SHEET_REVERT` — unchanged, still default OFF; this PR does not touch their defaults.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -428,6 +428,7 @@ jobs:
             tests/integration/multitable-reset-pit-realdb.test.ts \
             tests/integration/multitable-history-incomplete-precheck-realdb.test.ts \
             tests/integration/multitable-history-contiguity-realdb.test.ts \
+            tests/integration/multitable-history-contiguity-strict-seq-realdb.test.ts \
             tests/integration/multitable-d1c-form-submit-revision-realdb.test.ts \
             tests/integration/multitable-d1c-plugin-revision-realdb.test.ts \
             tests/integration/multitable-d1c-automation-revision-realdb.test.ts \
--- a/packages/core-backend/src/db/migrations/zzzz20260715160000_add_meta_record_chain_seq.ts
+++ b/packages/core-backend/src/db/migrations/zzzz20260715160000_add_meta_record_chain_seq.ts
@@ -1,0 +1,140 @@
+import { sql, type Kysely } from 'kysely'
+
+/**
+ * W0-1 (v3.7 correction, owner-ratified #4331 §9 — see
+ * docs/development/multitable-w0-1-v37-exact-anchor-trust-design-lock-20260715.md §1.1/§9.7) — the
+ * chain-integrity core's shared causal `seq` domain.
+ *
+ * WHY: the landed #4269 chain-order comparator (`chainOrderAfter` in `history-integrity-precheck.ts`)
+ * orders events by `(created_at epoch-ms, version, delete-last)` — `created_at` is txn-start so same-ms
+ * events collide, and neither `version` (resets per generation) nor row `id` (random uuid) can disambiguate
+ * true causal order across the two tables. A single monotonic sequence, allocated at write time and shared
+ * by both `meta_record_revisions` and `meta_record_version_markers`, gives an exact total order for the
+ * generation-aware precheck behind `MULTITABLE_HISTORY_CONTIGUITY_STRICT` (default OFF — this migration
+ * changes schema only, no runtime behavior; see §9.7 "Ratification authorizes default-off implementation
+ * slices only").
+ *
+ * ALSO (v3.7 §9.3 / owner-ruled "loud marker"): drops the cross-generation
+ * `UNIQUE (sheet_id, record_id, version)` on `meta_record_version_markers`. A resurrected record resets
+ * `version` to 1; a new generation's lock/unlock at a version the FIRST generation also marked would
+ * otherwise collide against that constraint, and the marker INSERT's (pre-existing) `ON CONFLICT ... DO
+ * NOTHING` would silently SWALLOW the new marker — the lock/unlock itself succeeds while its marker
+ * vanishes, and the strict contiguity walk then sees an unexplained hole in the new generation (a false
+ * refusal of a healthy record). Dup-detection moves into the per-generation (seq-ordered) occupancy walk
+ * instead (`chain_corrupt` — a TRUE within-generation duplicate, distinct from the legacy
+ * `duplicate_version_event`, which stays on the byte-identical flag-off path). `record-history-service.ts`
+ * `recordVersionMarker` removes the `ON CONFLICT DO NOTHING` in the SAME PR so a genuine write conflict now
+ * fails the enclosing lock/unlock transaction loudly instead of swallowing silently.
+ *
+ * Migration shape (safe on a populated table): add the column NULLABLE first, backfill deterministically,
+ * THEN attach the `nextval` default + `NOT NULL` — never add a column with a volatile (`nextval()`) default
+ * directly, which would force an uncontrolled full-table rewrite ordered by physical row order instead of
+ * the intended `(created_at, version, id)` backfill order.
+ *
+ * Backfill semantics (v3.7 §9.3 scope item 1 — verbatim): `row_number() OVER (ORDER BY created_at, version,
+ * id)` PER TABLE — revisions and markers each get their own dense 1..N run. This is DISPLAY/ORDER
+ * ASSISTANCE ONLY for legacy (pre-this-migration) rows, NEVER A TRUST SIGNAL: the two tables' backfilled seq
+ * ranges legitimately overlap (both start at 1), so a legacy revision and a legacy marker can share a seq
+ * value, and backfilled order is only as reliable as the `created_at`/`version` it derives from. Any
+ * "prospectively causal on landing" claim from the superseded v3.5 draft (#4309) is WITHDRAWN here per the
+ * v3.7 P2-A/P2-B correction: real causality requires L4 (all-writer fence, so seq is allocated
+ * post-fence-acquisition on every writer) and an L5 trusted-since checkpoint that cuts off pre-fence/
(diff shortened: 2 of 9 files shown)


zensgit:
## Independent adversarial gate — verdict: **CLEAR (0 P1 / 0 P2)**

Reviewed at pinned head `34c02daeb17bbdc4f9616f39774956d83c223c80` (no drift start→end). All 10 mandatory checks pass with code-traced or empirically-run evidence against a real Postgres. The two owner-falsified misses from the earlier L3 gate (#4309) are genuinely fixed and proven **by construction**:

- **C3 / target-generation hole** — `checkAllGenerationsContiguity` iterates every generation; a hole in a non-terminal generation returns `chain_hole` even with a clean terminal generation. The real-DB `TARGET-GENERATION HOLE, CLEAN TERMINAL ⇒ 409` golden passes, and a terminal-only mutation reds it (load-bearing, then restored).
- **C2 / created_at≤T** — the reconstructor is untouched (still non-causal) and the deferral to L6 is honestly documented; **no silent regression.**

Also verified: exact-bigint end-to-end (BigInt/string only, native `ORDER BY seq`; float64 round-trip mutation reds the exact-bigint goldens); the `ON CONFLICT` removal cannot cause a duplicate-key 500 (only PK=uuid remains, version bumped atomically under the row lock); the raw-data guard allowlist entry is honest (`r.data` is read only for content-equality comparison, never reaches a response); migration up/down/replay clean on a fresh DB with the setval bootstrap advancing strictly past the backfilled max; the new real-DB suite genuinely wired into `plugin-tests.yml`.

**No blockers.** All findings P3/NIT and safety-preserving. Two carried forward:
- **P3-1 (provenance):** the v3.7 exact-anchor design-lock doc cited by the code lands with #4331 (still Draft) — not yet on `main`. The behavior here is already sanctioned by the committed v3.6 unified lock (§3.2/§3.3, all-generations permitted as a stepping stone). No code change; resolves when #4331 merges.
- **P3-2:** no in-code guard prevents premature `MULTITABLE_HISTORY_CONTIGUITY_STRICT=true` while reconstruction is still non-causal. Only P3 (the strict precheck is read-only/refuse-only; the destructive path is separately gated default-off). **Carried into the L5 lane** as a fail-closed flag-on precondition + flag-manifest entry.

Draft, flags OFF, nothing armed/enabled/host-touched. 🤖 gate report generated with [Claude Code](https://claude.com/claude-code)

zensgit:
## L3 gen-0 + manifest fix — independent re-gate: CLEAR (0 P1 / 0 P2)

Re-reviewed at pinned head `f08465b874dbcc7369ab03ad99bb73974ff0b920`. This re-gate was explicitly briefed to construct the counterexample class the FIRST gate of this PR missed, and to check the new fail-closed-guard risk (over-refusal).

- **(a) Closes all generation-0 shapes.** `generation[i]` is a running create-count incremented before assignment, so every pre-first-create item has `generation[i]===0`; the guard at `history-integrity-precheck.ts:387` fails closed (`chain_hole`). The owner's counterexample + update-before-create + marker-before-create + delete-before-create all refuse (409, zero writes), proven at unit and HTTP. Mutation (`if (false && …)`) → exactly the 4 gen-0 goldens go RED at both unit and real-DB level.
- **(b) Does NOT over-refuse legitimate chains.** `create` is counted inclusively (`creates += 1` before `generation[i]=creates`), so a first-item-is-create chain has `generation[0]===1` and the guard never fires. Verified on normal / single-create / full trash→restore / three-clean-generations / create+lock-marker chains (repo goldens + an independent 8-case refute suite) — all pass.
- **(c) Manifest / CI.** `MULTITABLE_HISTORY_CONTIGUITY_STRICT` registered accurately (pure-read fail-closed gate, danger=low, no over-claim); the source-derived completeness contract test that was RED before registration is now green.

Single-`generation[0]`-check-covers-prefix reasoning verified sound (generation monotonic non-decreasing; the seq-ascending no-duplicate scan runs first, so no boundary tie/reorder can hide a gen-0 item after a create — out-of-order smuggling → `comparator_error`). Flag-off parity holds. No forbidden seq arithmetic/`setval`; tsc clean.

Non-blocking: the PR body's "27/27 + 10/10" goldens line is now stale (delta added 5 unit + 4 real-DB → 32 unit + 14 real-DB — I'll correct the body); optional cross-ref comment at `:387`.

Draft, flags default-OFF, nothing armed/enabled/host-touched. 🤖 re-gate report via [Claude Code](https://claude.com/claude-code)

zensgit:
Running the **independent exact-head re-gate** on L3 @ `6614cc35f` now (owner: "新 head 仍需精确-head re-gate"). Head is stable (37 min), CLEAN vs main, `test (20.x)` green — and as the stack base it won't rebase again pre-merge. This is read-only (my own worktree, no branch touch); the merge itself stays yours. Verdict + review MD to follow; L3's author shouldn't self-gate, so this supplies the required independent gate for the first integration rung.

zensgit:
**Correction:** L3 merged to main (`cc35b2599`) concurrently with my previous note — the branch is auto-deleted and the independent re-gate is moot; there is **no verdict-to-follow**. Post-hoc read-only sanity check confirms main's L3 content landed intact: `meta_record_chain_seq` migration, `checkAllGenerationsContiguity`, and `MULTITABLE_HISTORY_CONTIGUITY_STRICT` registered in the ops manifest (finding #3 fix). First integration rung is on main; L4 is now clean vs the new main (DIRTY resolved as expected). flags OFF.

PR #4196, opened 2026-07-12
docs(approval): retry/test-run governance rework — per-action-type semantics (supersedes #4164 Q3)

zensgit:
**Status: PROPOSED (doc-only; zero runtime, zero migrations, zero flag flips).** Awaiting owner ratify. This PR has no ratify authority of its own.

## Why this PR

Addresses an owner REQUEST-CHANGES on #4164. #4164's Q3 asked the owner to pick **one global** ledger durability trade for *all* side-effecting retry actions. The owner rejected that premise: different action classes need different correctness mechanisms, not one global knob.

This PR adds `docs/development/approval-automation-retry-action-classification-designlock-20260712.md`, which **supersedes #4164's Q3** and classifies retry/test-run semantics **by action type**:

| Action type | Class | Idempotency mechanism | Ambiguous-outcome | Never auto-resend? |
|---|---|---|---|---|
| `create_record`/`update_record`/`delete_record`/`lock_record` | **A — Same-DB write** | mutation + applied-ledger claim in ONE transaction (`FOR UPDATE` on the lock-check) | N/A — commit-or-rollback | N/A |
| `send_webhook`/`send_email`/`send_dingtalk_*` | **B — Outbound** | two-phase intent→attempt→outcome | timeout/reset/5xx-after-send **and a 4xx *response*** ⇒ **`outcome_unknown`** | **YES** |
| `send_notification` (rule path) | **B′** | eventBus-only, not durable today (§6 Q-A) | — | N/A today |
| `start_approval` | A-adjacent | already blocked | N/A | N/A |
| `condition_branch`/`parallel_branch`/`wait_for_callback` | C — Control-flow | never claimed | N/A | N/A |
| `record_click` | D — Inert | audit-only | N/A | N/A |

## Updated in this revision (owner review round)

- **Now self-contained.** The C1 idempotency identity (`testRunOperationId` → applied-ledger `root_execution_id`), the C4 `action_key` structural identity (normalized step-path incl. branch/parallel `step_key` + canonical config hash, replacing the types-only `computeActionFingerprint`), and the `meta_automation_action_applied` row shape are now **folded in as this lock's own contracts** (§2/§4/§6, provenance-noted to #4164). Ratifying #4196 now yields an implementable spec on its own — it no longer requires #4164's C1/C4 to be ratified first (§7/§10 updated accordingly).
[...]

PR #4340, opened 2026-07-15
feat(automation): action-idempotency ledger L1 — meta_automation_action_applied schema + golden

zensgit:
**Base: `main`. Additive, flag-neutral** — the FWB action-idempotency ledger (write-path prerequisite); no production callers until the FWB runtime lands behind its own gate. Head `5f81c8a18`.

## What lands
- **Migration `zzzz…_create_fwb_action_applied.ts`** → **`meta_fwb_action_applied`** (renamed off the #4196-reserved name): composite business-claim key via `uq_fwb_action_applied_business_claim`; named non-blank CHECKs; `application_mode ∈ {apply, test_run}`; **paired** `fwb_action_applied_node_epoch_paired` CHECK — the `('',0)` sentinel or a node_key matching `[!-~]` with epoch ≥ 1 (whitespace/tab rejected).
- **`automation-action-idempotency.ts`**: `deriveActionKey` / `deriveOutboundIdempotencyKey` (injective JSON-array encoding), `claimActionApplied` (Class-A same-txn claim + JS pair guard + **transaction probe**), `classifyOutboundOutcome` (`outcome_unknown` never auto-resent). Shared **`pg-transaction-guard.ts`** (byte-identical with #4336).

## Review fixes (rounds 1–3)
- **P2 injective encoding**; **P2 node/epoch pairing** (incl. whitespace/tab via `[!-~]` + JS-layer guard); **P1 table rename** off `meta_automation_action_applied` (#4196 §2.2's execution-scoped ledger) + coexistence golden; **P3** real-helper rollback test + body refresh.
- **P1 round-2 (pool acceptance)**: `claimActionApplied` accepted a pool → the claim auto-committed apart from the business write → a failed write left an orphaned claim and every retry returned `'duplicate'` (permanent skip). Now the **xid probe** rejects a pool / forged marker / bare autocommit client **before the claim** (three negatives, zero rows); the rollback test proves the retry **claims again** after a failed write. Also hard-stops the `withTransaction` non-transactional-fallback hazard.
- **P2 round-2 (destructive coexistence test)**: the probe now runs **inside one rolled-back transaction** (transactional DDL) — never DROPs; a pre-existing real #4196 table is reused untouched. Verified against the exact repro: pre-created real table + sentinel row → spec passes AND `to_regclass` non-NULL, row intact.

- **14/14** real-DB (incl. concurrent double-claim as two full transactions; P1 negatives; retry-after-rollback; non-destructive coexistence). `tsc --noEmit` 0.
- Probe, paired CHECK, JS pair guard, injective encoding all **mutation-proven**. Two-point CI wired.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -546,6 +546,7 @@ jobs:
             tests/integration/multitable-tombstone-record-capture-realdb.test.ts \
             tests/integration/multitable-tombstone-retention-realdb.test.ts \
             tests/integration/multitable-automation-outbox-schema-realdb.test.ts \
+            tests/integration/multitable-action-applied-ledger-realdb.test.ts \
             tests/integration/multitable-automation-dispatcher-claim-realdb.test.ts \
             tests/integration/multitable-automation-dispatch-loop-realdb.test.ts \
             tests/integration/multitable-automation-outbox-enqueue-realdb.test.ts \
--- a/packages/core-backend/src/db/migrations/zzzz20260715200000_create_automation_action_applied.ts
+++ b/packages/core-backend/src/db/migrations/zzzz20260715200000_create_automation_action_applied.ts
@@ -39,7 +39,12 @@ export async function up(db: Kysely<unknown>): Promise<void> {
       application_mode text NOT NULL DEFAULT 'apply'
                          CONSTRAINT action_applied_mode_valid CHECK (application_mode IN ('apply','test_run')),
       result_ref       text,
-      applied_at       timestamptz NOT NULL DEFAULT now()
+      applied_at       timestamptz NOT NULL DEFAULT now(),
+      -- node_key and entry_epoch must move together: BOTH the non-node sentinel ('', 0) OR BOTH a real
+      -- FWB-3 node scope (non-blank node_key, epoch >= 1). Without this, ('', 7) and ('node_A', 0) are
+      -- accepted and can split one idempotency identity into two rows (#4340 review P2).
+      CONSTRAINT action_applied_node_epoch_paired
+        CHECK ((node_key = '' AND entry_epoch = 0) OR (node_key <> '' AND entry_epoch >= 1))
     )
   `.execute(db)
   // the single business-claim key: sentinels (node_key='', entry_epoch=0) cover FWB-1/2; FWB-3 uses real values.
--- a/packages/core-backend/src/db/migrations/zzzz20260715200000_create_automation_action_applied.ts
+++ b/packages/core-backend/src/db/migrations/zzzz20260715200000_create_automation_action_applied.ts
@@ -41,10 +41,12 @@ export async function up(db: Kysely<unknown>): Promise<void> {
       result_ref       text,
       applied_at       timestamptz NOT NULL DEFAULT now(),
       -- node_key and entry_epoch must move together: BOTH the non-node sentinel ('', 0) OR BOTH a real
-      -- FWB-3 node scope (non-blank node_key, epoch >= 1). Without this, ('', 7) and ('node_A', 0) are
-      -- accepted and can split one idempotency identity into two rows (#4340 review P2).
+      -- FWB-3 node scope (node_key with a printable non-space char, epoch >= 1). The regex test [!-~]
+      -- (not merely <> the empty string) so a WHITESPACE/TAB-only node_key with a positive epoch is rejected
+      -- too — otherwise ('', 7), ('node_A', 0) AND (' ', 7) are accepted and can split one idempotency
+      -- identity into two rows (#4340 review P2).
       CONSTRAINT action_applied_node_epoch_paired
-        CHECK ((node_key = '' AND entry_epoch = 0) OR (node_key <> '' AND entry_epoch >= 1))
+        CHECK ((node_key = '' AND entry_epoch = 0) OR (node_key ~ '[!-~]' AND entry_epoch >= 1))
     )
   `.execute(db)
   // the single business-claim key: sentinels (node_key='', entry_epoch=0) cover FWB-1/2; FWB-3 uses real values.
--- a/packages/core-backend/src/db/migrations/zzzz20260715200000_create_automation_action_applied.ts
+++ b/packages/core-backend/src/db/migrations/zzzz20260715200000_create_automation_action_applied.ts
@@ -1,0 +1,54 @@
+/**
+ * Migration: `meta_automation_action_applied` — the ACTION-LEVEL idempotency ledger (FWB prerequisite).
+ *
+ * #4203 §2.3 ruling: a bare `approval_instance_id UNIQUE` would block a second rule/action forever, so the
+ * business-level claim key is the composite below. Two dedup layers stack: the EVENT level (`event_fires` /
+ * outbox per-consumer rows) stops "the same event redelivered"; THIS ledger stops "the same instance's same
+ * action re-applied" across retries/redeliveries/zombie double-sends — the Class-A same-DB write claim is
+ * INSERTed in the SAME transaction as the record write + revision + outbox row (§3 D9), so a zombie's
+ * duplicate write collides on the UNIQUE key and rolls back (natural fencing for same-DB effects).
+ *
+ *   - `action_key` = the #4196-C4 corrected triple identity {structuralPath, action.type, canonicalConfig
+ *     hash} — stable across fence tokens and attempts (the outbound-idempotency seed for Class-B).
+ *   - FWB-1/FWB-2 uniqueness = (instance_id, rule_id, action_key) with node_key='' AND entry_epoch=0
+ *     (sentinel: non-node-scoped). FWB-3 adds real (node_key, entry_epoch) — 退回/跳转 re-entry gets a new
+ *     epoch, and an old epoch's values are never reused (§6). One UNIQUE index covers both via sentinels.
+ *   - `application_mode` separates test-run/dry-run isolation from real applies ('apply' | 'test_run'):
+ *     a test run claims in mode 'test_run' and NEVER blocks (or is blocked by) a real apply.
+ *   - Class-B outbound effects don't get exactly-once from this table — they carry the derived idempotency
+ *     key to the endpoint and record `outcome_unknown` without auto-resend when the endpoint can't dedup.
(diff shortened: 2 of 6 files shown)


zensgit (commented):
## Codex review — REQUEST CHANGES

Reviewed exact head `37822002e4b9f1adbcd8d5275917c30a52b03ab4`.

### [P2] Both idempotency keys use non-injective delimiter concatenation

`deriveActionKey` builds `${structuralPath}#${actionType}#${hash}` and `deriveOutboundIdempotencyKey` builds `${eventId}::${actionKey}` (`automation-action-idempotency.ts:49-55,89-95`). This directly violates the ratified §2.1 requirement for a length-prefixed or structured-tuple encoding. The current API accepts plain strings, and the collision is executable:

```text
deriveActionKey({path:'a#b', type:'c'}) == deriveActionKey({path:'a', type:'b#c'})
deriveOutboundIdempotencyKey('evt::part','path#type#hash') ==
deriveOutboundIdempotencyKey('evt','part::path#type#hash')
```

A collision can make one legitimate action/write or outbound effect look like another and be dedup-skipped. Use an injective encoding (length-prefix each UTF-8 component, or hash a typed canonical tuple) and add the lock's required naive-concatenation collision negatives.

### [P2] `node_key` and `entry_epoch` are not constrained as one identity mode

The schema allows both `node_key='' / entry_epoch=7` and `node_key='node_A' / entry_epoch=0`; I inserted both successfully on a fresh PostgreSQL 15 database. The helper also silently synthesizes each half independently (`nodeKey ?? ''`, `entryEpoch ?? 0`). A caller drift can therefore split one logical FWB claim across sentinel and node-scoped keys and permit duplicate application.

Add a named DB CHECK and matching runtime validation: either the non-node sentinel pair (`node_key='' AND entry_epoch=0`) or a real pair (`node_key` nonblank and `entry_epoch>0`), with all invalid corners pinned.

### [P3] The “same-transaction guarantee” golden does not exercise the helper

The rollback test manually inserts the ledger row (`multitable-action-applied-ledger-realdb.test.ts:87-103`), so it stays green if `claimActionApplied` later stops using the supplied transaction and writes through a pool. Call the helper through the checked-out transaction, perform a representative sibling write, roll back, and assert both are absent.

Verification performed: fresh-DB migration and real-DB spec `10/10`, backend `tsc --noEmit` clean, plus executable key-collision and invalid-sentinel probes. Green CI does not cover these cases; do not merge this head.

zensgit (commented):
## Codex re-review — REQUEST CHANGES

Re-reviewed exact head `b9ef366f3a4e47bdeb61b47557a9b24dca41a5f1`.

The delimiter-collision P2 is closed: both action and outbound keys now use an unambiguous JSON tuple, and the previous two executable collision probes produce distinct keys. The real-helper rollback test also now proves `claimActionApplied` uses the supplied transaction.

### [P2] The real-node arm still accepts whitespace-only `node_key`

The new paired CHECK uses `(node_key <> '' AND entry_epoch >= 1)` in `zzzz20260715200000_create_automation_action_applied.ts`. On a fresh PostgreSQL 15 database I successfully inserted both:

```text
node_key='   ', entry_epoch=1
node_key='\t',  entry_epoch=1
```

Those are neither the `('', 0)` sentinel nor a valid real node identity, yet they create distinct claim keys and can split the same idempotency boundary if a caller trims/normalizes on another attempt. This is the exact “real pair requires nonblank node_key” half of the previous finding, and `claimActionApplied` still has no matching pre-query pair validation.

Please make the DB arm require a real nonblank node key (consistent with the existing `[!-~]` identity checks), validate the pair in `claimActionApplied` before querying (`('',0)` or nonblank key + positive safe integer only), and add space/tab negative controls plus a positive real-node control.

Verification performed: fresh migration, real-DB spec `11/11`, backend `tsc --noEmit` clean, prior key-collision probes closed, and the whitespace-only DB probe above. GitHub checks are green, but this remaining identity-boundary case blocks merge.

zensgit (commented):
## Head-scoped re-review: `c05933fc14dfb5b9426ee9db1b4b929099773d57`

**REQUEST CHANGES — one new P1 contract collision.** The prior whitespace/pairing findings are fixed on this head; this finding is against the current head, not `b9ef366f3`.

### [P1] The FWB migration occupies #4196's ratified table name with an incompatible schema

`packages/core-backend/src/db/migrations/zzzz20260715200000_create_automation_action_applied.ts:28-55` creates `meta_automation_action_applied` with the FWB instance-scoped shape:

`(instance_id, rule_id, action_key, node_key, entry_epoch, application_mode)`.

But the already-ratified #4196 contract reserves **that exact table name** for the execution-scoped retry/test-run ledger:

`(kind, root_execution_id, action_key)` with `UNIQUE(kind, root_execution_id, action_key)` (`approval-automation-retry-action-classification-designlock-20260712.md:220-250`).

The ratified FWB lock explicitly says the FWB claim must use its **own** instance-scoped ledger and **not** #4196's `meta_automation_action_applied` (`approval-form-writeback-fwb0-designlock-20260712.md:547, 667-674`). This PR currently gives the FWB ledger the reserved #4196 name, so the two ratified runtimes cannot coexist.

Fresh-DB proof after this PR's migration:

```text
INSERT INTO meta_automation_action_applied (kind, root_execution_id, action_key) ...
ERROR: column "kind" of relation "meta_automation_action_applied" does not exist
```

Because there are no callers yet, fix this before merge: give the FWB instance-scoped ledger a distinct FWB-specific table name and update the claim helper/SQL accordingly. Alternatively, both ratified locks would need an explicit owner amendment and a truly compatible combined schema; silently reusing the name is not compatible. Add a cross-lock real-DB golden proving both required table/insert shapes coexist after migration replay.

### [P3] PR body is stale

The body still says L2 is the next slice and reports a 7/7 golden, while this head already contains L2 and the real-DB file runs 12 tests. Please sync the body after the schema correction so the review surface matches the branch.

### Re-verification of the previous findings

- Fresh PostgreSQL 15: `multitable-action-applied-ledger-realdb.test.ts` **12/12**.
- DB rejects `('', 7)`, `('node_C', 0)`, space-only, and tab-only node keys under the named paired constraint.
- Helper probe: all four invalid pairs throw before SQL (`queryCalls: 0`).
- Injective tuple regressions remain distinct; `tsc --noEmit` and `git diff --check` pass.
- GitHub checks are green on this exact head and `mergeStateStatus=CLEAN`; green CI does not detect the cross-lock namespace collision.

zensgit (commented):
## Head-scoped re-review: `ae5e728147b36cb00f7c918cccdb96f88ecf63ae`

**REQUEST CHANGES remains.** This rebase preserves the whitespace fix, but it contains no product change from `c05933fc1` in the migration/helper/spec and does not address the current P1.

### [P1] The cross-lock table-name/schema collision is unchanged

The branch still creates the FWB instance-scoped schema as `meta_automation_action_applied`, while the ratified #4196 contract reserves that exact table name for `(kind, root_execution_id, action_key)`. The ratified FWB lock explicitly requires its instance-scoped claim ledger to be separate from #4196's execution-scoped table.

Fresh PostgreSQL 15 on this exact rebased head still fails the #4196-required insert:

```text
INSERT INTO meta_automation_action_applied (kind, root_execution_id, action_key) ...
ERROR: column "kind" of relation "meta_automation_action_applied" does not exist
```

Rename the FWB table/claim path to a distinct FWB-specific ledger (or explicitly amend both ratified locks to a compatible combined schema) and add the requested coexistence golden. A rebase cannot close a product-schema conflict.

### Current verification/state

- Focused fresh-DB ledger spec remains **12/12**; the whitespace/pair validation is intact.
- At review time `test (20.x)` is **red** on an unrelated `multitable-oapi2a-ratelimit-realdb.test.ts` timing assertion (1 failed / 1899 passed). A green rerun may clear that CI flake, but it does not change this P1.
- The PR is currently **1 commit behind** `origin/main` (`6b7917b8a`).
- The stale PR-body note ("L2 next", 7/7) also remains to be updated after the schema correction.

zensgit (commented):
## Head-scoped re-review: `b1b646c50bf160bf1e3dfa1401c84ec3758edb23`

**REQUEST CHANGES — the table rename is correct, but one P1 and one P2 remain.**

### [P1] `claimActionApplied` can commit the terminal claim outside the business transaction

`claimActionApplied(trx: Queryable, ...)` still accepts the ordinary pool (`automation-action-idempotency.ts:77-102`), and this PRs own tests call it with `poolManager.get()`. The Class-A guarantee is load-bearing: claim + record mutation + revision + outbox must commit or roll back together. With a pool, the terminal claim commits first; if the business write then fails, every retry sees `duplicate` and permanently skips the missing write.

Fresh PostgreSQL 15 probe against this exact head and the real helper:

```json
{"first":"claimed","simulatedBusinessWrite":"failed-before-write","retry":"duplicate","persistedClaimRows":1}
```

The existing rollback test proves correctness only **given** a transaction; it does not prevent the production seam from being called without one.

**Required close:** make the helper transaction-only with an actual runtime transaction proof or a helper-owned transaction boundary, add a pool negative that leaves zero claim rows, and retain a positive that rolls back claim + representative business write together. Do not reuse a caller-forgeable boolean marker as the sole proof.

### [P2] The coexistence golden drops a pre-existing official #4196 table

The test does `CREATE TABLE IF NOT EXISTS meta_automation_action_applied` and then unconditionally `DROP TABLE IF EXISTS meta_automation_action_applied` in `finally` (`multitable-action-applied-ledger-realdb.test.ts:192-214`). Once the #4196 migration exists, the CREATE becomes a no-op and this spec deletes the real shared-schema table.

I pre-created the locked #4196 table with a sentinel row, ran only the coexistence case, and the test passed while the postcondition was:

```
after_table=NULL
```

Use a run-unique isolated schema/dedicated database, or a checked-out transaction whose DDL and insert are rolled back. The test must never drop a global table it did not prove it created.

### Confirmed fixed

The FWB migration/helper now consistently use `meta_fwb_action_applied`, and the #4196 `(kind, root_execution_id, action_key)` INSERT can coexist with the FWB claim. The PR body is current.

Verification: focused real-DB spec **13/13**, exact CI typecheck clean. Current GitHub state is still red in `test (18.x)` on an unrelated attendance commit-async assertion (`failedRows: expected 2, got 0`); `test (20.x)` is green. The branch is 3 commits behind current main and the local merge-tree is conflict-free.

PR #4342, opened 2026-07-15
feat(approval): attachment runtime — full pipeline slices ①-⑦ (#4195 validation core + storage + routes + GC/reconcile + bind + web client)

zensgit:
Complete flag-gated approval-attachment runtime for the ratified pipeline lock (#4195): validation core + storage provider + migration + upload/download routes + GC worker + bind-reconciler + web client. Gated by `APPROVAL_ATTACHMENTS_ENABLED` — **default OFF**; when off, `createApprovalAttachmentRouter` returns `null` and nothing is registered (byte-for-byte no-op). Dormant end-to-end until the flag flips; **not for self-merge**.

Scope actually in this diff (correcting the original "pure validation core, no callers" description, which undersold it — this bundles every slice of the runtime, not just validation):

- **Validation core** (`services/approval-attachment-validation.ts`) — reject-by-default v1 rules: **20 MB/file · 10 files/field · 50 MB/submission**; MIME allowlist PDF/JPEG/PNG/TXT/CSV (D6 — Office/ZIP/archives deferred until AV integration); extension⇄MIME cross-check; non-positive/unsafe sizes reject; per-file values-free reject codes. Includes a fix for a prototype-pollution DoS in the allowlist lookup (`mimeType: 'constructor'`/`'__proto__'` previously resolved an inherited `Object.prototype` member and threw an uncaught `TypeError`; now guarded with `Object.hasOwn`).
- **Migration** (`db/migrations/..._create_approval_attachments.ts`) — `approval_attachments` (upload lifecycle: unbound → bound → deleted) + `approval_attachment_purge_intents` (durable blob-purge intent, bind↔GC symmetry). Additive only.
- **Storage provider** (`services/approval-attachment-storage.ts`) — server-derived keys (client filename never used as a path), path-containment (traversal/absolute/NUL refused), put/get/delete.
- **Routes** (`routes/approval-attachments.ts`) — flag-gated upload (multer, in-memory) and auth-proxied download; unbound = uploader-only, bound = participant (fail-closed on auth-check error), deleted = 410.
- **GC worker** (`services/approval-attachment-gc.ts`) — unbound-TTL sweep + purge-intent drain.
- **Bind-reconciler** (`services/approval-attachment-reconciler.ts`) — submit-time bind (form-freeze) + bidirectional bucket reconciliation.
- **Web client** (`apps/web/src/approvals/attachmentUpload.ts`) — client-side pre-validation mirror + upload client.

**Verification:** unit goldens green for validation/storage/routes (incl. the new prototype-pollution regression tests); real-DB integration tests (`approval-attachment-gc-realdb.test.ts`, `approval-attachment-bind-reconcile-realdb.test.ts`) are two-point-wired into `vitest.config.ts` excludes + `plugin-tests.yml`'s real-DB run-list — they self-skip without `DATABASE_URL` locally and run in CI. Typecheck green (`tsc --noEmit`, 0 errors).

---

**Follow-up (2026-07-17): independent review closed 11 ratified-lock gate gaps.** An independent adversarial review of this runtime against the ratified pipeline lock (#4195) was addressed in the commits above — closing gaps against gates G2/G3/G6/G7/G11/G12/G15, the §3-bis durable-purge state machine, route error-handling, server-side org derivation, and web-spec CI wiring. Load-bearing invariants are mutation-proofed. The feature remains **default-OFF** (`APPROVAL_ATTACHMENTS_ENABLED`); `createApprovalAttachmentRouter` still returns `null` while OFF (byte-for-byte no-op). **Not for self-merge.**

---

**Fix round (2026-07-20): owner block — CI red + two P1s — resolved (rebased onto main `698997cf8`; commits `f071076fc…323421eae`).**

| Item | Status |
|---|---|
| **CI red — #4154 app-mode tripwire** | ✅ FIXED — the routes spec's 15 `request(app)` sites migrated to the pinned-server transport (`usePinnedServer()` + `serve(app)`); tripwire + full `tests/unit` lane green (429 files / 5846 tests). |
| **P1 — poison-at-claim was dead** | ✅ FIXED — the drain's claim statement now enforces the §3-bis ceiling itself (same CASE shape as `automation-durable-dispatcher`): an at-ceiling reclaim transitions to terminal `dead_letter` **without invoking the blob deleter** (attempts/fence unchanged, lease NULL, `last_error='max_attempts_exhausted'`, alert seam fired). Real-DB goldens: seeded at-ceiling expired `in_progress` → poisoned, deleter spy NOT called, terminal on re-drain; **constructed crash-at-claim loop** (5 claims with no outcome via a death-after-claim db wrapper → attempts rise 1..5 → claim poisons). **Mutation-proven:** reverting the CASE to plain `'in_progress'` turns both goldens RED (verified). |
| **P1 — production pipeline unmounted** | ✅ FIXED — see below. |

**Production pipeline (new `services/approval-attachment-runtime.ts` + `index.ts` boot block), all flag-gated on `APPROVAL_ATTACHMENTS_ENABLED` (D5 default OFF — flag-OFF boot proven byte-identical by a real-DB flag-OFF boot golden: the surface is simply not mounted, 404):**

- **Boot mount:** `authenticate`-guarded router mount; GC unbound-TTL sweep + purge-intent drain + bucket reconciler on flag-gated timers (started like the other background sweeps; stopped in `stop()`). **Fail-closed boot:** flag ON + store probe failure ⇒ startup ABORTS (the `durableBootFailureDisposition` doctrine).
- **Submit-txn bind (§4.4/G4):** `bindAttachmentsOnSubmit(client, …)` runs inside `ApprovalProductService.createApproval`'s transaction, right after the `form_snapshot` INSERT. Real-DB goldens over a booted server: staged upload binds atomically (row `bound` + snapshot freezes the id array); a foreign id fails the create with a values-free 400 and the **whole create rolls back** (no instance row; foreign row untouched). `validateFieldType` now accepts the lock's ratified array-of-ids shape for `attachment` values (the legacy string/record acceptance predated §4.4/§8 and could freeze an uninterpretable value into the immutable snapshot).
- **Storage decision — the O3 reading implemented:** the lock's §9 wording is explicit: *"when `APPROVAL_ATTACHMENTS_ENABLED=true` and `NODE_ENV=production`, upload fail-closes (503) unless a non-local, S3-compatible provider is configured."* That ratified fail-close is implemented verbatim: production resolves `s3-required` and upload **and** download return values-free 503 (never a local write on the deploy host — issue #159); local-FS remains dev/test-only under a dedicated probed root. **No S3 adapter is invented** — the lock itself names the S3-compatible provider as the prerequisite for exercising rung 2 in production (§7), so the adapter is the explicitly-named follow-up, and the flag cannot be turned ON in production until it lands.
- **Template-access seam on upload (§4.1):** the SAME `applyTemplateVisibilityFilter` predicate the create path enforces (via the exported request-actor derivation — one derivation, no drift); outsider ⇒ values-free 404 **before** the field-type resolve (no template/schema oracle). Real-DB negative: user-scoped template, outsider upload → `404 {error:'not_found'}`; scoped-in non-admin uploads fine (positive control). Download participant predicate (requester / user+role assignee / past actor / user+role CC / admin) + `collectHiddenFieldIds` byte-gate wired to the real tables.
- **FE control:** new `approvalAttachments` feature flag (backend session `features` ← `isApprovalAttachmentsEnabled()`; store default OFF, no inference); `ApprovalNewView` swaps the B2-28 honest-disabled stub for the real uploader ONLY flag-ON (id-array payload per §4.4, client mirror pre-validation, remove support); flag-OFF placeholder + submit strip byte-identical (pre-existing specs double as the control). `attachmentUpload.ts`'s "stays detached" note updated. `approvalNewView.spec.ts` (previously wired into NO lane) added to the required web gate + approval-web-guard.

**Verification (2026-07-20, fresh migrated DB):** backend `tsc` 0; full `tests/unit` 5846/5846 (flag OFF — byte-identical); attachment unit suites 39/39; real-DB attachment suites 22/22 (`gc-realdb` 8, `bind-reconcile-realdb` 7, new `pipeline-realdb` 7 — two-point wired); `vue-tsc` 0; required web gate 350 files / 4149 tests green (incl. `approvalNewView` 21/21, `approval-attachment-upload` 4/4); web lint clean (`featureFlags.spec.ts` was red before this branch — pre-existing quarantine, verified unchanged by stash A/B).

**Honest remaining gaps (all named by the lock as separate opt-ins):** S3-compatible `StorageProvider` adapter (prod-enable prerequisite, §7/O3); G13 stale-draft-reference detection on restore (drafts deliberately strip attachment ids in BOTH flag states until it lands); detail/list/mobile frozen-ref rendering (§8, rung 3 remainder); rung 4 authorability flip + B2-28 retirement (owner-ratify-gated); G5/G8 frozen-snapshot render + full serving-header goldens beyond the route unit assertions.

**Still not for self-merge.**

---

## Production pipeline + poison fix (2026-07-20, owner P1 round)

Owner's two P1s + CI red, all closed (commits `f071076fc` / `148fbe627` / `b2125b51e` / `323421eae`):

- **CI red**: 15 app-mode supertest sites migrated to the pinned-server transport (#4154 tripwire) — both required lanes green again.
- **P1 poison-at-claim**: an at-ceiling expired reclaim now transitions to `dead_letter` IN the claim statement — the deleter is NEVER invoked at the ceiling, and a crash-at-claim loop terminates (the ratified §3-bis contract; previously attempts grew forever and the deleter still ran).
- **P1 production pipeline**: flag-gated boot (`bootApprovalAttachmentRuntime`: route mount + GC sweep + purge drain + reconciler timers, stop() registered, fail-closed boot posture) · `bindAttachmentsOnSubmit` INSIDE the approval submit transaction (ApprovalProductService:4050 — a bind failure rolls back the whole submit) · **O3 storage decision implemented as ratified**: production REQUIRES an S3-compatible store; the S3 adapter is an explicitly-named follow-up, so `NODE_ENV=production` resolves `s3-required` → values-free **503 fail-close** (approval blobs never land on the deploy host); dev/test resolve local-FS · template-access seam on upload · FE uploader attached to ApprovalNewView behind the same flag (detached note removed).

Verified at `323421eae`: attachment real-DB suites 29/29 on a fresh fully-migrated DB (main-session independent re-run) + unit suites + CI fully green. Built across an agent session-limit death; the main session re-verified every piece against the pushed head.


zensgit:
**Adversarial-review findings (opus, ranAgainst `5500c60b2`) — NEEDS_FIX before land:**

- **[P2] Scope over-claim vs the ratified staged-opt-in lock.** The PR title/body say "slice ① pure validation core, no callers, zero behavior" with ②–⑦ as "Next slices", but the diff actually contains **all 7** — a migration creating `approval_attachments` + `approval_attachment_purge_intents`, upload/download routes, GC worker, reconciler, storage provider, and the frontend upload client (~1098 additions). `approval-attachments.ts` and `approval-attachment-reconciler.ts` both import the validator, so the "No callers yet" docstring is false. Runtime is inert (flag OFF + router unwired at boot), and the migration is additive — but it *does* create two tables on any migrate and lands undisclosed route/worker code. `approval-attachment-pipeline-design-lock-20260709` §7 requires each rung separately opted-in. Fix: correct the body to disclose full scope + get owner ratification for bundling rungs 1–3, **or** split into the lock's rungs.
- **[P3] MIME allowlist prototype-pollution hole (latent, fails closed).** `V1_ALLOWLIST[mime]` walks `Object.prototype`; `mime='constructor'`/`'__proto__'` (survive lowercasing) + an allowlisted ext reaches `allowedExts.includes(...)` where `allowedExts` is the `Object` function → uncaught `TypeError` instead of the values-free `mime_not_allowed` reject. Fails closed (no acceptance, no bypass) but violates the "reject-by-default, per-file values-free codes" contract; reachable via multer `f.mimetype` once wired. Fix: null-prototype map / `Map`, or `Object.prototype.hasOwnProperty.call(V1_ALLOWLIST, mime)` guard.

No-DB unit lane 11/11; the 3 validation guards are mutation-proven load-bearing; both real-DB specs are two-point wired. (Owner away per /goal — recorded here for your re-review; not self-merging.)

zensgit:
Rebased onto main 60e9ecbec, re-verified (backend tsc 0; 5 attachment unit suites 31/31; web upload spec 4/4 + approvalNewView 17/17; fresh migrated DB: 2 attachment real-DB suites 13/13; run-list wiring add-only, no entries dropped). Awaiting review.

PR #4349, opened 2026-07-16
fix(ops): corrective-3 — acceptance smoke capture stdout-only (2>$null), not stderr-merged

zensgit:
## corrective-3 — acceptance-runner smoke capture is stdout-only, not stderr-merged

**Refs #4101** (on-prem acceptance tracker). This PR must NOT change that tracker's status — it stays OPEN until the corrective-3 entity-machine run PASSes.

### What the entity machine actually hit
The RC-0 run reported `POWERSHELL_NATIVE_STDERR_PROMOTION`. Root cause: stage-6's smoke capture in `stock-preparation-onprem-acceptance.ps1` used

```powershell
$out = & node @args 2>&1
```

Under PowerShell 7.3+ native-command error handling, a child that writes to **stderr** while its stream is **merged** (`2>&1`) can be promoted to a *terminating error* that aborts the capture **before the smoke's values-free summary is ever parsed**. The smoke process ran; the **acceptance runner** threw. This is a runner-compatibility failure, **not** a smoke FAIL — app / migration / PM2 / health were all fine, and RC-0 is **not** a PASS.

### Fix
Extract a summary-only helper `Invoke-SmokeCapture` that:
- captures **only stdout** (`& node @NodeArgs 2>$null` — stderr discarded, never merged),
- preserves `$LASTEXITCODE`,
- keeps the admin-token BSTR zero-free + scoped-env `finally` cleanup (token discipline unchanged; the static-invariant tests still pass).

`Invoke-SmokeStage` now parses the helper's stdout. Values-free guarantee unchanged (only whitelisted summary fields are ever emitted).

### Load-bearing test (runs in the required `test (20.x)` pwsh gate)
New behaviour check D5: a fake smoke writes a **sentinel to stderr** and a valid summary to **stdout**, then asserts the capture (1) does not throw / abort, (2) does **not** capture the stderr sentinel, (3) still parses PASS (8/8, clean), (4) preserves exit 0 + clears the token env.

**Mutation-verified:** flipping `2>$null` back to `2>&1` reds the "sentinel is NOT captured" check (35 pass / 1 fail) — version-independent, because `2>&1` merges stderr into the captured text regardless of whether promotion fires. Behavioural count 32 → 36; the CI comment count is updated (49 → 53).

### Scope / sequencing
- Fix + load-bearing test only. **Do not re-run corrective-2.** After this merges, cut the exact-SHA **corrective-3** package and re-run the acceptance from stage 1 on the entity machine; RC-A stays gated on that PASS.
- `externalWrite=false` invariant untouched; no runtime behavior outside the acceptance runner changes.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -123,8 +123,8 @@ jobs:
       - name: Global History flag manifest contract (R12-C)
         run: node --test scripts/ops/global-history-flag-manifest.test.mjs scripts/ops/multitable-global-history-flag-status.test.mjs
 
-      # #4210 T9: the stock-preparation one-click acceptance script's contract + behavioural tests (49
-      # checks = 17 contract + 32 behavioural) are pure pwsh 7 + tar + node (no DB, no live server — the
+      # #4210 T9: the stock-preparation one-click acceptance script's contract + behavioural tests (53
+      # checks = 17 contract + 36 behavioural) are pure pwsh 7 + tar + node (no DB, no live server — the
       # pm2/health/smoke stages are unit-tested via dot-sourced pure helpers). Run inside the REQUIRED
       # `test` gate, but only ONCE (on 20.x — no need across the 18/20 matrix), early (no pnpm install
       # needed: pwsh/tar ship on ubuntu, node is already set up) so a regression fails the required check
--- a/scripts/ops/__tests__/stock-preparation-onprem-acceptance.behavior.tests.ps1
+++ b/scripts/ops/__tests__/stock-preparation-onprem-acceptance.behavior.tests.ps1
@@ -226,6 +226,34 @@ try {
   $o = Test-SmokeOutcome "mvpSmoke.pass=true`nauditActionsCovered=8/8"
   Check "smoke: selfScanClean ABSENT -> FAIL (fail-closed)" ((-not $o.ok) -and $o.reason -eq 'self_scan_not_clean')
 
+  # ── D5 (corrective-3): the stage-6 smoke capture is stdout-ONLY (2>$null), never merged (2>&1). A child
+  # that writes noise to stderr AND a valid values-free summary to stdout must NOT abort the capture (the
+  # RC-0 POWERSHELL_NATIVE_STDERR_PROMOTION failure — where 2>&1 promoted native stderr to a terminating
+  # error and the smoke's summary was never read), must NOT let the stderr text contaminate the parsed
+  # summary, and must still parse PASS. Mutation guard: flip 2>$null back to 2>&1 in Invoke-SmokeCapture and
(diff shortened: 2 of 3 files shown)


zensgit:
Owner exact-head review at `308c3aeda938a6a288e0851d30bfdaf58a34bfd4`: **HOLD / two P2 test gaps.**

The runtime fix itself is directionally correct: `Invoke-SmokeCapture` captures stdout only with `2>$null`, preserves the local `$LASTEXITCODE`, and keeps BSTR/env cleanup in `finally`. The required `test (20.x)` job really ran all 17 static + 36 behavioural checks.

However, the new D5 is not yet load-bearing across the production boundary:

1. **[P2] D5 tests the helper directly, not the real stage-6 wiring.** I mutated `Invoke-SmokeStage` to bypass `Invoke-SmokeCapture` and execute `& node @nodeArgs 2>&1` again while leaving the helper and D5 untouched. The required behaviour suite still passed **36/36**. Thus a regression that restores the exact entity-machine failure in production remains green. Exercise `Invoke-SmokeStage` itself with the fake smoke (set `$PSNativeCommandUseErrorActionPreference = $true` under pwsh 7 so merged stderr is discriminating), or add an AST contract proving that `Invoke-SmokeStage` calls `Invoke-SmokeCapture` exactly once and contains no direct `node` invocation/merged stderr. Prefer the stage-level behaviour leg, with AST as a belt.

2. **[P2] “exit code preserved” only tests exit 0.** I mutated the helper return to `exit = 0` unconditionally; all **36/36** checks still passed. Add a second fake child that emits a valid-looking PASS summary but exits nonzero (for example 7), then assert the helper returns 7 and the real `Invoke-SmokeStage` fails closed as `failedStage=smoke` rather than accepting the summary.

Mutation evidence was produced in an isolated detached worktree and removed afterward; the PR branch was not changed.

Other checks:
- PR body has no issue-closing keyword;
- token discipline/static checks remain green;
- diff check is clean;
- no runtime/product/external-write surface changed;
- current branch is behind hot `main` and should be rebased after these guards land.

Once both discriminator legs are red under the corresponding mutations and restored green, this should be a short re-review and merge. Do not cut corrective-3 from the current head yet.

zensgit:
Superseded by #4351. The replacement rebases the corrective-3 runtime fix onto current main, preserves the prod-health workflow step, and adds load-bearing production-wiring plus nonzero-exit guards. No release package was cut from this conflicting head.

zensgit:
Superseded by #4352 — same corrective-3 work rebased onto hot main (`b2ed0bdc7`), plus the two P2 re-review fixes you flagged (AST wiring guard so a stage revert to direct `2>&1` reds; fail-closed test driving the real `Invoke-SmokeStage` with a non-zero exit). This PR's head commit was orphaned by the rebase, so GitHub wouldn't let me reopen it. Continue the short re-review on #4352.

PR #4352, opened 2026-07-16
fix(ops): corrective-3 — acceptance smoke capture stdout-only (2>$null) + wiring/exit guards

zensgit:
## corrective-3 — acceptance-runner smoke capture is stdout-only, not stderr-merged

**Refs #4101** (on-prem acceptance tracker). This PR must NOT change that tracker's status — it stays OPEN until the corrective-3 entity-machine run PASSes. Supersedes #4349 (closed after a rebase onto hot main orphaned its head commit); this branch carries the same work plus the re-review fixes below.

### What the entity machine hit
The RC-0 run reported `POWERSHELL_NATIVE_STDERR_PROMOTION`. Root cause: stage-6's smoke capture used `& node @args 2>&1`. Under PowerShell 7.3+ native-command error handling, a child that writes to **stderr** while its stream is **merged** (`2>&1`) can be promoted to a *terminating error* that aborts the capture **before the smoke's summary is parsed**. The smoke process ran; the **acceptance runner** threw. A runner-compatibility failure, **not** a smoke FAIL — app / migration / PM2 / health were fine; RC-0 is **not** a PASS.

### Fix (`afde2fd67`)
Extract a summary-only helper `Invoke-SmokeCapture`: captures **only stdout** (`2>$null`), preserves `$LASTEXITCODE`, keeps the admin-token BSTR + scoped-env `finally` cleanup (token discipline unchanged; static-invariant tests still pass). `Invoke-SmokeStage` parses the helper's stdout.

### Re-review fixes (`b2ed0bdc7`) — closing the two P2 test gaps you found
Your exact-head review flagged that the first test (D5) exercised `Invoke-SmokeCapture` in isolation, so a revert of the **stage** to direct `2>&1` stayed 36/36 green, and it only tested exit 0, so a helper hardcoded to return 0 stayed green. Both closed:

1. **AST wiring guard (static, +2 contract).** stage-6 must call `Invoke-SmokeCapture` and **never invoke node directly**; the helper's node command must carry a `FileRedirectionAst` on stderr (`2>$null`) and **no** `MergingRedirectionAst` (`2>&1`). Mutation-proven: reverting the stage to `& node ... 2>&1` reds the delegation check; flipping the helper to `2>&1` reds the redirect check.
2. **Fail-closed behaviour test (D6, +1 behavioural).** Drives the **real** `Invoke-SmokeStage` in a subprocess (mirrors the orchestrator try/catch, robust to native-exit promotion) against a child that prints a valid PASS summary but **exits 7** — the stage must `Stop-WithFailure` (exit 1, `failedStage=smoke`). Mutation-proven: a helper hardcoded to return 0 lets the stage PASS → reds.

Plus the original D5 (behaviour, +4): a fake smoke writes a sentinel to stderr + valid summary to stdout → no abort, sentinel not captured, summary parses PASS, exit/token cleared; mutation `2>$null`→`2>&1` reds the "sentinel not captured" check.

**Counts 49 → 56 (17 → 19 contract, 32 → 37 behavioural).** All in the required `test (20.x)` pwsh gate. Rebased onto hot main (`d64f0a6d8`).

### Sequencing
Fix + tests only. **Do not re-run corrective-2.** After a short exact-head re-review + merge, cut the exact-SHA **corrective-3** package from the merged main SHA and re-run acceptance from stage 1 on the entity machine; RC-A stays gated on that PASS. `externalWrite=false` untouched.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -129,8 +129,8 @@ jobs:
       - name: Prod health probe monitor contract
         run: node --test scripts/ops/prod-health-probe-state.test.mjs
 
-      # #4210 T9: the stock-preparation one-click acceptance script's contract + behavioural tests (53
-      # checks = 17 contract + 36 behavioural) are pure pwsh 7 + tar + node (no DB, no live server — the
+      # #4210 T9: the stock-preparation one-click acceptance script's contract + behavioural tests (56
+      # checks = 19 contract + 37 behavioural) are pure pwsh 7 + tar + node (no DB, no live server — the
       # pm2/health/smoke stages are unit-tested via dot-sourced pure helpers). Run inside the REQUIRED
       # `test` gate, but only ONCE (on 20.x — no need across the 18/20 matrix), early (no pnpm install
       # needed: pwsh/tar ship on ubuntu, node is already set up) so a regression fails the required check
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -129,8 +129,8 @@ jobs:
       - name: Prod health probe monitor contract
         run: node --test scripts/ops/prod-health-probe-state.test.mjs
 
-      # #4210 T9: the stock-preparation one-click acceptance script's contract + behavioural tests (49
-      # checks = 17 contract + 32 behavioural) are pure pwsh 7 + tar + node (no DB, no live server — the
+      # #4210 T9: the stock-preparation one-click acceptance script's contract + behavioural tests (53
+      # checks = 17 contract + 36 behavioural) are pure pwsh 7 + tar + node (no DB, no live server — the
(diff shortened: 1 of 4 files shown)


zensgit:
Superseded by #4351 after exact-head comparison.

The production `Invoke-SmokeCapture` change is patch-equivalent, but #4351 carries the stronger load-bearing contract:

- stage 6 must make exactly one `Invoke-SmokeCapture` call with `-NodeArgs` and `-Token`, and must not call `node` directly;
- the helper is directly proven to preserve native exit `7`;
- the real stage is proven to fail closed while recording `failedStage=smoke` and the coarse `smokeExit=7`;
- token cleanup and no raw stderr capture remain asserted;
- the required `test (20.x)` gate completed successfully on the newer rebased head.

#4352 only proves that a nonzero child eventually fails the stage; it does not distinguish preserving `7` from rewriting it to another nonzero value. Keeping one canonical PR avoids duplicate hot-main rebases. No package was cut and #4101 remains OPEN pending entity-machine acceptance.

PR #4356, opened 2026-07-16
docs(attendance): S7 A2 approver-resolver design lock (PROPOSED — 直属上级/部门主管/多级上级)

zensgit:
## Summary

- Drafts the S7 A2 审批人 resolver design lock — **PROPOSED, NOT RATIFIED**, docs-only, no runtime.
- Owner decision points enumerated as OD-S7-1..6 (§5) for explicit owner ratification; this document does not self-ratify.
- Headline survey finding (verified against `origin/main`, file:line cited throughout): the resolver kernel (`packages/core-backend/src/services/ApprovalAssigneeResolver.ts`) already implements `direct_manager`/`dept_head`/`continuous_managers`/`manager_at_level` LIVE and mostly RATIFIED. The actual gap is that `plugins/plugin-attendance` runs a fully separate, parallel approval engine with zero reference to that resolver, zero reference to `directory_*`, and two silent-drop layers (a zod schema without `.passthrough()` + a normalizer that reconstructs steps from a 3-key allowlist) that strip any `kind` field before persistence. This lock scopes the wiring work, not new kernel resolver logic.
- §6 slices the work S7-1 (contract + extension point) → S7-2 直属上级 → S7-3 部门主管 → S7-4 多级上级 → S7-5 A1 editor wiring, each independently flag-gated with real-DB tests.

## Test plan

- [ ] N/A — docs-only PR, zero runtime changes.
- [ ] Owner review + ratification of §5 OD-S7-1..6 before any S7-* implementation PR is opened.


zensgit:
CHANGES REQUESTED — owner review against head `ffd476ac6278dd0b023d4907558de678129555c1`.

The design direction is valuable, but the lock is not ratify-ready yet. The current proposal stops at assignment production while the attendance decision path has additional, security-bearing semantics.

1. **[P1] Dynamic assignments are not authoritative for attendance actions.** `resolveRequest()` first applies the broad attendance RBAC/scheduler-scope gate, then `isApproverAllowed()` checks only `approverUserIds` / `approverRoleIds`; both empty returns `true`. The reject path skips that check entirely. Therefore writing a dynamic user into `approval_assignments` does not stop another scope-authorized actor from approving or rejecting. Amend the lock to require an active-current-node assignment check for both approve and reject, with explicit admin-override semantics and negative real-DB tests.

2. **[P1] The proposed kernel reuse is not org-scoped.** `resolveApprovalRequesterOrgRelations(localUserId, query, options)` accepts no attendance `orgId`; its first query selects the most recently updated linked account across all integrations for that local user. Binding later joins to that chosen `integration_id` does not prevent selecting an integration belonging to another org. Require an org-scoped resolver variant that joins `directory_integrations.org_id`, plus a two-org / one-local-user test.

3. **[P1] `continuous_managers` cannot inherit kernel semantics in the current attendance engine.** Kernel chain-as-set semantics depend on node `approvalMode` (all/any). Attendance has no quorum state and advances after one actor approves, while this lock explicitly keeps 会签/或签 OUT. For the first S7 delivery, restrict multi-level support to sequential `manager_at_level`, or separately expand and ratify the attendance quorum engine.

4. **[P2] Resolve-at-submit freezing is not represented in the stated extension points.** The attendance requester snapshot currently carries only `id` / `name`, and next-step assignments are rebuilt when the previous step advances. The lock must add create-time hydration of `managerId` / `deptHeadId` / `managerChainIds` and require every later step to read only that frozen snapshot. Add a test that mutates directory relations between step 1 and step 2 and proves the assignee does not change.

5. **[P2] S7-1 creates an unsafe partial state.** It allows `kind` to round-trip while explicitly providing no resolver behavior. A kind-only step then has no static approvers and can fall into the legacy admin fallback. An unimplemented or flag-disabled kind must be rejected fail-closed (for example 422); do not accept inert configuration before its resolver exists.

6. **[P2] OD-S7-5 omits the established narrow host-to-plugin capability pattern.** The menu currently offers a first runtime dependency on the executable core package, local logic duplication, or a full engine migration. Add a recommended host-injected, org-scoped resolver port under `context.services`, following the existing `workdayCalendar` capability pattern. This preserves one implementation without coupling the plugin to the whole core package runtime.

Recommended owner posture after amendment: block-with-error by default; `manager_at_level` only for the first multi-level slice; exact kernel relationship semantics after adding the org anchor; resolve and freeze at submit; narrow host-injected resolver port; authoring warning plus runtime hard block for missing directory linkage.

Do not flip this lock to RATIFIED and do not open S7 runtime PRs until these points are resolved and re-reviewed.

PR #4364, opened 2026-07-16
docs(stock-prep): add T3b PLM auto-persist design lock

zensgit:
## What

Records the owner-ratified T3b design lock for approved PLM source-run -> internal project/snapshot/run persistence. It also updates the landed T3a lock with its merged status and corrects the old `intake.plmBomLines` wording to the real `intake.bomSnapshotLines` contract.

## Owner ratification

The owner ratified the contract at exact head `f3c1a0bedbf4365dc6715450845cc04414957623` in [this PR comment](https://github.com/zensgit/metasheet2/pull/4364#issuecomment-4989139120). Commit `47fe40e13` only records that decision in the document; it does not change the ratified runtime contract.

Ratified decisions:

- independent default-OFF `MULTITABLE_STOCK_PREP_PLM_AUTOPERSIST_ENABLED` flag
- authenticated tenant owns the physical staging target; body `projectId` remains only the business project key
- pure closed-shape intake bridge with no row dropping
- bridge status input is closed to `imported|active|inactive|incomplete`; `imported` is the sole alias and normalizes to `active`; other raw lifecycle values fail before persistence with values-free 422 `STOCK_PREPARATION_PLM_AUTOPERSIST_LINE_STATUS_UNSUPPORTED`
- `stock_preparation_bom_line_status_v1` is an admin-supplied runtime option source; the bridge canonical set is the sole enforcement point
- exact replay requires unique keys, bounded complete reads, and equality of the full frozen batch/line/run persistence projections; `snapshotLineId + sourceFingerprint` alone is explicitly insufficient
- exact replay also requires exactly one project row but never rewrites its live pointer; orphan, duplicate, unprovable-read, or content conflicts fail closed
- shared persist hardening is explicitly **not** flag-inert and lands as separately reviewed T3b-1a; the pure T3b-1b bridge may proceed in parallel; T3b-1c route wiring waits for both and is the only default-OFF inert runtime slice
- created reports `internal_persist/true`; exact replay reports `internal_noop/false`
- production enablement remains barred on separate P4 atomicity/repair hardening
- extend the existing T4 harness, then cut one RC-A package only after #4101 reaches a final RC-0 PASS; the reviewed corrective-4 smoke FAIL does not unlock RC-A

## Current acceptance baseline

Corrective-4 #4369 fixed Windows PowerShell 5.1 summary capture, but its entity run returned `mvpSmoke.pass=false` with smoke exit 1. #4101 remains open. RC-A requires package/provenance/migration/PM2/health PASS plus `mvpSmoke.pass=true`, `auditActionsCovered=8/8`, `selfScanClean=true`, and `externalPlmK3ErpWrite=false`.

## Boundary

Design documentation only. Ratification permits the staged implementation work to begin, but does not merge or approve any runtime slice. Each of T3b-1a, T3b-1b, and T3b-1c remains subject to its own implementation, evidence, review, and merge gate. This PR authorizes no production enablement, value-bearing read, external PLM/K3/ERP write, orphan repair/deletion, or OD-W3-1 expansion. `externalWrite=false` remains invariant.

Operationally, a new `snapshotBatchId` may create a complete replacement snapshot after an interrupted batch; it is not an orphan-repair mechanism.

## Verification

- owner ratification recorded against exact contract head `f3c1a0bedbf4365dc6715450845cc04414957623`
- rebased onto `main` containing corrective-4 #4369
- source/config premises cross-checked against the real source-run intake, mapper, and persist implementation
- constructed proof that intake `imported -> active` changes persisted status while preserving `sourceFingerprint`, demonstrating why fingerprint-only replay checks are insufficient
- duplicate-key, multi-page read, page-bound, project-orphan, and full-projection mutation requirements are explicit
- live #4101 corrective-4 FAIL re-baselined into the RC-A gate
- `git diff --check`: PASS
- confirmed the proposed T3b flag remains absent from runtime

Refs #4263, #4266, #4369, and #4101.

### DIFF
--- a/docs/development/stock-preparation-t3a-erp-source-autopersist-design-lock-20260714.md
+++ b/docs/development/stock-preparation-t3a-erp-source-autopersist-design-lock-20260714.md
@@ -1,6 +1,9 @@
-# T3a 设计锁 — ERP approved source-run 服务端直落 `erp_material_master`（决策已定，待 RATIFY, rev 2026-07-15）
+# T3a 设计锁 — ERP approved source-run 服务端直落 `erp_material_master`（RATIFIED / runtime 已落）
 
-状态：**决策已写实（RATIFY-ready）** — OD-1..OD-5 全部裁决落定（§3），owner 2026-07-15 授权「推进 T3a/T3b」。实现处 `bb9724494`（分支 `claude/stock-prep-t3a-erp-autopersist`）**WIP 已按本锁核心接线**（flag 条件 tenant + auto-persist，flag OFF 证实 inert）；**flag ON 的安全验证(steering 拒绝 red-first + 真库 smoke + mutation)待补**（§5）。模型：实现=Opus 主循环（权限+真实写+mutation 证明，owner 政策）。**ratify 是 owner 的门**——本 rev 只把决策写实,不自我 ratify。
+> 2026-07-16 收账：design-lock #4263 与 runtime #4357 已合入 main。T3b 的独立后续锁见
+> `stock-preparation-t3b-plm-source-autopersist-design-lock-20260716.md`。
+
+状态：**RATIFIED / runtime MERGED** — OD-1..OD-5 已裁决，#4357 已完成 flag ON steering、真库 smoke、mutation 与 values-free 双向验证并合入 main（`f49322c75`）。本文保留 T3a 的设计契约；不授权后续 T3b runtime。
 
 ## 1. 问题（owner P1 功能断点，已核实）
 
@@ -14,7 +17,7 @@
 在 **同一请求内**,ERP source-run 成功后,把服务端已有的 `intake` 规范化行**直接**喂给 T2 的 `persistStockPreparationErpMaterialSync`——**行永不过 HTTP**,公开面仍 values-free。复用两个已合入且已硬化的服务端件(source-run intake + T2 persist),是**组合接线**,非新能力面。**flag OFF 时响应逐字节不变(不加任何字段);flag ON 时才落库并加 `autoPersist` + 覆盖 projector 的 dry_run/未写标记**(权威契约见 OD-4)。
 
 ```
-source-run 路由(与 http-routes.cjs WIP bb9724494 一致):
+source-run 路由（与 #4357 merged implementation 一致）：
   admin gate
   autoPersistEnabled = flag                                   (OD-1，默认 OFF)
   若 ON: assertStockPreparationErpAutoPersistNoSteering(req)   // 请求含 tenantId/projectId → 400,在 body allowlist 与任何 I/O 之前 (OD-2)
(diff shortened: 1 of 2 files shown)


zensgit:
## T3b design-lock review — doc-vs-code, refute-first

Reviewed head `a2c422358` vs `origin/main` `f49322c75`; doc-only (2 `.md`, no runtime wiring). Method: opened every cited code path on main and checked each factual claim byte-for-byte, separating claims about **existing** code (must be accurate now) from **new** contract points (must be self-consistent + implementable). Full report saved locally.

### Verdict: RATIFY-READY after one OD-3 correction (1 × P2, 1 × P3). No P1. No doc-vs-code factual error about existing code.

Unlike T3a (which had six doc-vs-code errors), **every one of the 8 existing-code claims here verified correct** — including the two I initially suspected, both of which the code refuted:
- `intake.runRecord` **does** exist — the intake *return* (`readonly-intake.cjs:325-330`) spreads `...output` **and** adds `runRecord` at `:327`, and `buildSourceRunOutcome` (`readonly-source-run.cjs:662`) assigns the whole return to `intake`. The doc is right.
- The `missing_child_bom` fail-close is **unreachable** on the T3b path — `missingChildBom:true` is only set by `stampMissingChildLine` (`expansion-snapshot-mapper.cjs:172`) on an expansion rowError, and the source-run path never runs the expansion. And the projectStatus concern is moot — persist hard-codes `projectStatus:'active'` (`sync-run-persist.cjs:181,196`).

Also confirmed correct: `intake.bomSnapshotLines` is the real output row-array (`plmBomLines` is the *input* field — the §0 correction is right); persist re-runs the expansion mapper via `planBomSnapshotSyncRun` → `mapExpansionRowsToSnapshotLines` (`plan:261`), so OD-3's "prove the second normalization is identity-preserving" is well-founded; the current false-skip idempotency (`persist:289-310`); OD-2's resolvers (`:526`, `:561`, `assertRowsStayInProjectScope:641`) and the projectId(business)/targetProjectId(write-target) split that makes "keep body projectId, reject query/params projectId + all tenantId" both coherent and necessary for the PLM path; and OD-5's `internal_noop`/false-on-replay being a deliberate *stricter* contract than the shipped T3a route (not a misdescription).

### P2-1 — OD-3 never decides the config-reachable source-status case; as written it 422s the whole envelope on any real PLM lifecycle status, silently
OD-3 frames status normalization solely around the `imported` default. But intake passes through any source-mapped `status`/`lineStatus` **verbatim** (`readonly-intake.cjs:202`); mapped rows are spread whole into intake (`readonly-source-run.cjs:753`); a mapped record carries only fieldMap `target` keys (`read-source-read-runtime.cjs:154`); and `target` is a **free-form bounded identifier, not a closed vocabulary** (`read-source-config.cjs:105-108`). So an approved config that maps a PLM lifecycle column (`FStatus`/`State` → `status`/`lineStatus`, values like `released`/`obsolete`/`WIP`) makes OD-3 fail-close the **entire envelope (422)**. It fails *closed* — safe, no corruption, no leak — but it's over-restrictive and undocumented, and it decides whether the `lineStatus` select is usable in v1 at all: with no status mapping, `lineStatus` is always `active`, making OD-3's own "active/inactive/incomplete 保真" clause vacuous on this path. (No current config/fixture maps a status target — repo-wide grep empty — so it's admin-config-reachable, not automatic.)

**Fix (pick one, owner's call), the doc just has to state the decision:** (a) OD-3 adds a value-normalization map for known PLM lifecycle statuses → `{active,inactive,incomplete}` and fail-closes only genuinely-unknown values; or (b) T3b v1 explicitly scopes source-status mapping OUT (lineStatus fixed to `active`, status semantics deferred) and says so; or (c) owner accepts the whole-envelope 422 and the doc tells operators not to map a raw lifecycle column.

### P3-1 — Claim #4 calls `stock_preparation_bom_line_status_v1` a "frozen option contract"; it's an admin-supplied runtime option-source
`templates.cjs:660` `optionSource:{type:'contract', ...}` — resolved from an admin-registered contract at provisioning, not a code constant, and no code validates persisted `lineStatus` against it (the mapper passes it through, `:152`). Net: the bridge's `imported→active` is the **only** enforcement point — don't lean on records-service select validation. This strengthens the doc's own case for the bridge and dovetails with P2-1.

### The rest is ratify-ready
OD-1/OD-2/OD-4/OD-5/OD-6 all pass as contracts; §5.3's must-kill mutation list and §5.2's `plugin-tests.yml`-whitelist requirement correctly target the load-bearing guards the future runtime will add (the plugin `__tests__` chain otherwise runs in no required workflow). Runtime + ratify remain separate gates, as the doc states. Fix OD-3's one clause and this is ready for an exact-head short re-review → ratify.

zensgit:
Addressed the exact-head review at `141def388de79a007dd812aea7ea958d736af9ea`:

- **P2 / OD-3:** explicitly chose option (c). T3b v1 does not guess cross-PLM lifecycle semantics. Missing source status follows `imported -> active`; canonical `active|inactive|incomplete` passes; any mapped raw lifecycle value fails the whole envelope before provisioning/persist with values-free 422 `STOCK_PREPARATION_PLM_AUTOPERSIST_LINE_STATUS_UNSUPPORTED`. The operator constraint and load-bearing tests/mutations are now specified.
- **P3:** corrected `stock_preparation_bom_line_status_v1` to an admin-supplied runtime option source. Config shape validation and records select validation are explicitly non-bearing; the bridge closed vocabulary is the sole enforcement point.

Validation: source/config premises cross-checked, OD-3 assertions 7/7 PASS, and `git diff --check` PASS. The PR remains Draft/PROPOSED and is ready for the requested exact-head short re-review; it has not been self-ratified.

zensgit:
## Exact-head closure review (not ratification)

Reviewed and corrected against current `main` at head `4b44da0ecfc31b1637fd4a9e80a59d1ad112cf99`.

Three additional load-bearing gaps were closed:

1. **Fingerprint-only false replay:** `snapshotLineId + sourceFingerprint` cannot prove full persisted-row equality. A constructed intake case keeps the fingerprint unchanged while `imported -> active` changes the persisted status. The lock now requires canonical comparison of the full frozen batch/line/run projections.
2. **Incomplete/ambiguous replay:** duplicate keys, partial first-page reads, page-bound exhaustion, and a commit that reached batch/lines/run but crashed before project upsert now all fail closed. Exact replay requires bounded complete reads, unique keys, full projection equality, and exactly one project row.
3. **Flag-inert scope:** the T3b source-run wiring is default-OFF inert, but shared persist hardening changes the existing `/mvp/sync/persist` replay contract unconditionally. Delivery is now split into independently reviewed T3b-1a persist hardening, parallel pure T3b-1b bridge, then T3b-1c route integration.

The status alias contract was also made implementable: intake collapses default and source-mapped `imported`, so `imported` is explicitly the sole accepted alias to `active`; other noncanonical lifecycle values remain values-free 422.

The RC gate now references merged corrective-4 #4369 and the pending entity result on #4101. `git diff --check` passes. This remains Draft, RATIFY-ready, and authorizes no runtime.

zensgit:
Live acceptance rebaseline at `f3c1a0bed`: corrective-4 #4369 is merged and packaged, and its entity run completed with runner capture fixed but packaged smoke exit 1 (`mvpSmoke.pass=false`). The prior OD-6 wording only required a reviewed result, which would let a reviewed FAIL unlock RC-A. The lock now requires a final #4101 RC-0 PASS with the full values-free pass tuple before RC-A can be cut. This is a design-ledger correction only; no runtime is authorized.

zensgit:
## Exact-head short re-review at `4b44da0ec` — RATIFY-READY (my recommendation)

Verified the delta since my reviewed head `a2c422358` (PR content vs main = the two docs only; the rest of the diff is main-rebase noise).

**My two findings are resolved:**
- **P2-1 → option (c), correctly tightened.** `imported` is the sole alias → `active`; canonical `active|inactive|incomplete` pass; any other mapped lifecycle value fails the whole envelope with the dedicated values-free 422 `STOCK_PREPARATION_PLM_AUTOPERSIST_LINE_STATUS_UNSUPPORTED`, before provisioning/persist. The alias rationale is code-true: intake (`readonly-intake.cjs:202`) collapses the default and a source-mapped `imported` into the same literal, so the bridge cannot (and should not) fake provenance. Operator constraint + load-bearing tests/mutations specified.
- **P3-1 → corrected** (admin-supplied runtime option source; bridge = the sole enforcement point).

**The three self-found closures are real and verified:**
1. **Fingerprint-only false replay** — right fix. `snapshotLineId+sourceFingerprint` genuinely cannot prove persisted-row equality (bridge normalization can change the persisted `lineStatus` while the source-derived fingerprint is unchanged). Full frozen-projection comparison (`*_FIELD_IDS` only, ignore `id/version`, template-typed normalization, keyed by unique `snapshotLineId`) is implementable and matches the T2 grounding idioms.
2. **Incomplete/ambiguous replay** — bounded-complete paged reads (fail-closed `PERSIST_EXISTING_BATCH_READ_UNPROVABLE` at the page bound), uniqueness (`PERSIST_PLAN_LINE_KEY_AMBIGUOUS` pre-I/O; batch/run exactly-1), and the project-row rule are all **consistent with the actual write order I re-verified on main**: batch → lines → run → project upsert LAST (`sync-run-persist.cjs` step 5 → `upsertStockPreparationProject`), so 0 project rows on replay correctly means an interrupted first commit.
3. **Non-inert persist hardening split** — the most important honesty point. The false-skip→409 change hits the shipped `/mvp/sync/persist` route **unconditionally** (it is NOT flag-gated), so T3b-1a as its own independently reviewed PR with compat regression + real-DB evidence is the right structure; T3b-1b (pure bridge) can proceed in parallel; T3b-1c (route wiring) is the only flag-gated inert piece.

**One operator-facing note for the T3b-1a PR body/runbook** (not a lock change): after hardening, a crash-retry that previously false-skipped will 409 — the documented recovery is re-sync under a NEW `snapshotBatchId`. Existing `/mvp/sync/persist` callers should see this called out as an intentional behavior change, not a regression.

Recommendation: **ratify at this exact head** (`4b44da0ec`) — ratify remains the owner's act; runtime (T3b-1a/1b/1c) each stays a separate gate per the lock's own ladder. RC gate references (merged corrective-4 #4369 + pending #4101 entity result) are accurate.

zensgit:
## Exact-head correction: ratification still pending

The technical review of OD-3/OD-4 at `4b44da0ec` remains valid, but its final RC-gate statement is stale. #4101 had already received the corrective-4 entity result: runner capture succeeded, packaged smoke exited 1, and `mvpSmoke.pass=false`.

Current PR head is `f3c1a0bedbf4365dc6715450845cc04414957623`. Its only post-review delta is the fail-closed ledger correction that a reviewed FAIL does not unlock RC-A; #4101 must reach the full RC-0 PASS tuple first. No runtime contract was widened.

Therefore the recommendation for `4b44da0ec` is not recorded as ratification of the current head. Owner exact-head ratification is still required for `f3c1a0bed`; until then this PR remains Draft and authorizes no T3b runtime.

zensgit:
## Owner ratification — RATIFIED

Exact design-lock head ratified: `f3c1a0bedbf4365dc6715450845cc04414957623`.

The owner authorizes the T3b design contract recorded at this head, including:

- the independent default-OFF PLM auto-persist flag;
- auth-principal ownership of the physical staging target while body `projectId` remains the business key;
- the closed, no-row-drop bridge and `imported -> active` sole-alias rule;
- values-free 422 rejection for unsupported lifecycle values before persistence I/O;
- full frozen-projection replay equality, unique keys, bounded-complete reads, and project-row completeness;
- the explicit split between non-inert shared persist hardening (T3b-1a), parallel pure bridge work (T3b-1b), and default-OFF route integration (T3b-1c);
- RC-A remaining blocked until #4101 reaches the complete RC-0 PASS tuple and T4-final passes.

This ratification does **not** authorize production flag enablement, external PLM/K3/ERP writes, automatic orphan repair/deletion, OD-W3-1 value-plane reads, or combining the separately reviewed runtime slices. Each runtime PR retains the implementation, real-DB, mutation, values-free, and independent-review gates in the lock.

Operational note for T3b-1a: a crash-retry that previously false-skipped will intentionally become 409. Re-syncing with a new `snapshotBatchId` restores a complete new snapshot but does not repair or delete the orphan; that distinction must be explicit in the PR body/runbook.

PR #4365, opened 2026-07-16
feat(multitable): OD-W2-5a — record-history read passes through restored_from_version (R11 inspector badge; fixes History-Center normalizer half-wire)

zensgit:
## OD-W2-5a — record-history READ passes through `restored_from_version`

Owner ruling (直接 GO, a-read-through, no further owner decision, no #4339 dependency): the R11 restore
back-reference column already exists on `origin/main` (landed via #4124, 7月11日) with the write seam and
version-restore wiring. The remaining gap was the **read** path — it stopped at the DB column and never
surfaced `restoredFromVersion` to the inspector history tab, and there was a real half-wire in the base
History Center normalizer.

### What this does
- **Backend read** (`record-history-service.ts`): `listRecordRevisions` now projects `restored_from_version`
  into `RecordRevisionEntry.restoredFromVersion`, guarded by the existing `hasRestoredFromVersionColumn`
  probe so a pre-migration deploy window degrades to `null` (no 42703 / 500) instead of erroring.
- **FE types + normalizers** (`types.ts`, `api/client.ts`): `MetaRecordRevision` carries the field;
  `normalizeRecordHistoryEntry` passes it through. **Fixes the half-wire**: `normalizeHistoryChange`
  previously **dropped** `restoredFromVersion`, so the base History Center badge worked in component
  tests (which inject objects directly) yet went blank on the real client wire.
- **Component** (`MetaRecordHistoryPanel.vue`): renders the existing `restoredFromVersionBadge` **only when
  the field is non-null** — an entry with `source='restore'` but a NULL back-reference shows no badge.

### Verification (API / client / component golden triple, owner-mandated)
- **API (real DB)**: new G4 in `multitable-restore-backreference-realdb.test.ts` — the *read* surfaces
  `restoredFromVersion=1` after a restore; versions 1 & 2 stay null. Ran on fresh PG (9/9). Mutation:
  neutering the read projection reds G4; **dropping the column degrades to null with no 500** (42703 fallback proven).
- **Client**: new `multitable-record-history-client-restored-from.spec.ts` — drives `listRecordHistory` /
  `getHistoryBatch` with a raw wire payload; positive preserves a numeric value, negative coerces
  absent/non-number to null. Mutation-proven (stripping both pass-throughs reds the 2 positive assertions).
  The required-gate `multitable-client.spec.ts` golden is also extended (rev_2 positive, rev_1 null).
- **Component**: `multitable-record-history-panel.spec.ts` — positive (badge with `v2`) + owner-mandated
  **negative** (`source='restore'` + NULL → **no badge**; mutation confirms the `v-if` keys on the field,
  not on `source`).
- Both type-checks clean; the field is additive (no shape rename).

Closes the last authorized W2 runtime slice; the W2 design+verification MD (#4358) §附录 A is updated to
declare W2 complete once this lands.

### DIFF
--- a/apps/web/src/multitable/api/client.ts
+++ b/apps/web/src/multitable/api/client.ts
@@ -757,6 +757,10 @@ function normalizeHistoryChange(raw: unknown): HistoryChange {
     changedFieldIds: Array.isArray(p.changedFieldIds) ? p.changedFieldIds.map(String) : [],
     before: p.before && typeof p.before === 'object' ? (p.before as Record<string, unknown>) : null,
     after: p.after && typeof p.after === 'object' ? (p.after as Record<string, unknown>) : null,
+    // OD-W2-5a half-wire fix: the HistoryChange type + backend payload carry restoredFromVersion but this
+    // normalizer dropped it, so the base History Center badge worked in component tests (which inject objects
+    // directly) yet went blank on the real client wire. Pass it through (badge keys on non-null).
+    restoredFromVersion: typeof p.restoredFromVersion === 'number' ? p.restoredFromVersion : null,
   }
 }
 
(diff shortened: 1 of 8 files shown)


zensgit:
Superseded by #4367. The shared branch name collided with a parallel session's force-push; #4367 carries the identical complete implementation (four runtime files blob-identical to the gated `78be4552b`) on a distinct branch, plus the adversarial gate's requested test/CI hardening (CI-gate the client half-wire golden; locale-robust badge assertions). Closing this to avoid any merge on the contested branch.

PR #4369, opened 2026-07-16
fix(ops): make stock-prep smoke capture PowerShell 5.1-safe

zensgit:
## Root cause

The corrective-3 package runs the acceptance boundary under Windows PowerShell 5.1, but its required behavioral gate only exercised Ubuntu `pwsh 7`. With global `ErrorActionPreference=Stop`, Windows PowerShell 5.1 promotes native stderr to a terminating `RemoteException` even when the command uses `2>$null`, so stdout and the smoke summary are never captured.

## Scoped correction

- save the caller's `ErrorActionPreference`
- set `Continue` only around the native Node invocation
- capture stdout and `$LASTEXITCODE`
- restore the prior policy in an inner `finally`
- retain the outer SecureString/BSTR and token-env cleanup
- keep stderr excluded from the values-free summary
- keep all surrounding runner stages fail-closed

## Target-shell gate

Adds a stable `stock-prep PowerShell 5.1 acceptance` job on `windows-latest` that explicitly invokes `powershell.exe` 5.1. Its child writes both stdout and stderr at exit 0 and exit 7. The gate asserts stdout capture, stderr exclusion, exact exit-code preservation, error-policy restoration, and token-env cleanup.

### Red-first evidence

The first harness attempt exposed a PS5.1-only `Join-Path` fixture incompatibility and was not counted. After correcting only that fixture, test-only head `279a9bab659aaa934686c6b1b935b81a36c3a95e` ran in [Actions 29473870072](https://github.com/zensgit/metasheet2/actions/runs/29473870072) and produced the load-bearing result:

```text
host=PowerShell 5.1 Desktop PASS
captureChecks=3/10 PASS, 7 FAIL
exit0Capture=FAIL
exit7Capture=FAIL
policyAndTokenCleanup=PASS
```

After the scoped runner fix, rebased exact head `292b6d780` ran the same target-shell job in [Actions 29474124328](https://github.com/zensgit/metasheet2/actions/runs/29474124328):

```text
powershell51=10/10 PASS
exit0Capture=PASS
exit7Capture=PASS
stderrExcluded=PASS
exitCodePreserved=PASS
errorPolicyRestored=PASS
tokenEnvCleared=PASS
```

Local cross-version checks:

```text
staticContracts=19/19 PASS
pwsh7Behavior=38/38 PASS
workflowYamlParse=PASS
gitDiffCheck=PASS
```

## Boundary

No business smoke logic, migration, PM2 behavior, package contents, T3b runtime, external-write path, credentials, host values, or entity-machine action changes in this PR. No package should be cut and no entity rerun should occur until this exact head is independently reviewed, merged, and the Windows PowerShell 5.1 check is registered as required on `main`.

Refs #4101.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -99,6 +99,29 @@ jobs:
             scripts/ops/github-dingtalk-oauth-stability-summary.test.mjs \
             scripts/ops/dingtalk-oauth-stability-workflow-contract.test.mjs
 
+  stock-prep-powershell51:
+    name: stock-prep PowerShell 5.1 acceptance
+    runs-on: windows-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js 20.x
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20.x
+
+      # The release package runs this acceptance boundary under Windows PowerShell 5.1. pwsh 7 has
+      # different native-stderr semantics and cannot stand in for the target shell contract.
+      - name: Run target-shell smoke capture regression
+        shell: powershell
+        run: |
+          $powershell51 = "$env:SystemRoot\System32\WindowsPowerShell\v1.0\powershell.exe"
+          & $powershell51 -NoProfile -ExecutionPolicy Bypass -File `
+            scripts/ops/__tests__/stock-preparation-onprem-acceptance.ps51.tests.ps1
+          if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
+
   test:
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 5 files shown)


zensgit:
## Independent review (post-merge validation) — MERGE-READY, no P1/P2

Reviewed head `292b6d780` refute-first, opening the fix + guards and mutation-checking the pwsh-7-testable ones. The fix is correct and *better* than corrective-3: scoping `$ErrorActionPreference='Continue'` around only the native `& node … 2>$null` call is what actually stops the **Windows PowerShell 5.1** native-stderr promotion (`2>$null` alone does not on 5.1), the caller's policy is restored in the inner `finally`, and fail-closed stays intact (the stage's `if ($captured.exit -ne 0) Stop-WithFailure` still fires).

**Guards proven load-bearing (mutation):** stage-delegation static guard (inject `& node` into the stage → reds), scope-and-restore static guard (delete the restore → reds), real-stage fail-closed subprocess test (helper hardcoded `exit=0` → reds), D5 sentinel (`2>$null` not `2>&1`) survives. Coverage parity with corrective-3 confirmed, token discipline intact.

**CI fact (not inferred):** the new `stock-prep PowerShell 5.1 acceptance` job ran on a real Desktop 5.1 host (`PASS host is Windows PowerShell 5.1 Desktop` … `ALL POWERSHELL 5.1 CHECKS PASS (10/10)`), and it is now a required check on main. The ps51 host-guard is non-vacuous (on pwsh 7 it reds 9/10).

### Non-blocking follow-ups (P3/NIT)
- **P3-2 (recommended):** the ps51 suite has no *positive control* that the promotion actually fires without the fix — a future 5.1 servicing change could silently make the 'does not abort' checks vacuous. Add an inline positive control asserting `& node … 2>$null` under `Stop` throws on that host.
- **P3-1:** the behavioural `$ErrorActionPreference -eq 'Stop'` assertions are vacuous w.r.t. the inner-finally restore (PowerShell function-scopes the `Continue` assignment, so the caller stays `Stop` regardless) — the restore invariant is caught statically instead; harmless.
- **NIT:** ps51 fixture uses `\\n` (literal backslash-n) vs the behaviour test's `\n`; harmless today (substring match), fragile if later line-parsed. And the `$pass/10` final message is hardcoded.

Honest boundary: I cannot run Windows PowerShell 5.1 locally, so the fix-works-on-5.1 direction is verified as a CI fact; only the revert-reds direction (P3-2) is unreproduced by me.

PR #4368, opened 2026-07-16
W0-1 L6-a — sealed operation-endpoint ledger + mint protocol (DRAFT, flags OFF)

zensgit:
**DRAFT — do not merge/arm/enable.** W0 Time Machine line, stacked on L4cov (base `claude/w0-l4cov-univer-meta-writers-20260715`). All recovery flags default-OFF; `MULTITABLE_ENABLE_WRITER_FENCE` gates every new runtime path. Flag-off ⇒ no operation rows minted ⇒ byte-identical to L4cov.

## What this builds (v3.7 §1.2 — the exact-anchor centerpiece)
A wall-clock T (even in-txn `clock_timestamp()`) is not an exact commit boundary, and a PG sequence value is allocation-order not commit-order. The **sealed operation endpoint** is the exact, externally-visible commit boundary a recovery anchor can trust: the endpoint row of ONE committed operation, written LAST, visible only post-commit.

### 1. Ledger schema (`zzzz20260715190000_create_meta_record_history_operations.ts`)
- `meta_record_history_operations(sheet_id, operation_id uuid, endpoint_seq BIGINT, event_count INT, created_at)` — **PK(sheet_id, operation_id)**; server-minted, sheet-scoped operation_id.
- nullable `operation_id` on `meta_record_revisions` + `meta_record_version_markers` (legacy rows = NULL, never trusted).
- **DEFERRABLE INITIALLY DEFERRED** FKs events→endpoint: events may precede the endpoint in one txn; a write-events-but-never-seal op fails at COMMIT.
- **append-after-seal trigger**: an event whose operation already has a visible endpoint is refused.
- **endpoint-validation trigger**: `event_count` must equal the actual event count across BOTH event tables and `endpoint_seq` must equal their exact MAX(seq); mismatch RAISES (also enforces endpoint-LAST via count=0).

### 2. Mint protocol (`operation-ledger.ts`)
`mintOperation` (after the canonical fence) → tag each revision/marker (`operation_id` + exact `seq` tracked in BigInt, `batch_id` aligned to operation_id) → `sealOperation` LAST (`endpoint_seq=MAX(seq)`, `event_count=N`). Exact bigint end-to-end (string/BigInt, never Number/parseInt/+/-). Column-probed so a flag-ON deploy predating the migration degrades to inert instead of poisoning the txn.

### 3. Writers WIRED (mint+seal)
- `RecordService`: create / patch / delete / trash-restore
[...]

PR #4380, opened 2026-07-16
feat(multitable): W0-1 L6-a finding-#2 — sealed endpoint immutability (BEFORE UPDATE reject + golden, owner review 2026-07-16)

zensgit:
## W0-1 L6-a finding-#2 — sealed endpoint rows are DB-immutable

**Owner review 2026-07-16, finding #2** (Medium). Targets the L6-a branch (#4368) so the diff is exactly the migration guard + golden. **Docs+migration+test; Draft; NOT merged, NOT armed. flags OFF.**

### The hole
v3.7 §1.3 reads the recovery `anchorSeq` from *"that immutable endpoint row"*. The migration installed only a `BEFORE INSERT` trigger (`trg_mrho_validate_endpoint`) that validates the seal at mint time — but nothing forbade a later **UPDATE**. So `UPDATE meta_record_history_operations SET endpoint_seq = 999999` (or `event_count` / `created_at`) on a sealed row **silently moves an already-committed recovery anchor**. Confirmed by independent verification: the only trigger on the table was `BEFORE INSERT` (`:169`); `created_at` had no CHECK; `endpoint_seq`/`event_count` only positivity CHECKs.

### The fix (DB-level, fail-closed)
- Migration `up()`: `trg_mrho_reject_update` — `BEFORE UPDATE ... RAISE 'immutable (UPDATE forbidden)'`. The endpoint is **write-once by design** (minted last, then only read), so a blanket UPDATE-reject is the correct guard. `down()` drops the trigger + function. (DELETE stays with retention design, not this constraint.)
- **§F2 real-DB golden**: seals a real endpoint, asserts every column UPDATE — including a no-op `SET endpoint_seq = endpoint_seq` — is DB-rejected, and (positive control) the anchor is byte-unchanged after every rejected attempt.

### Verification
- Fresh `postgres:14` + migrate (with the CI exclude set): L6-a suite **17/17 pass** incl §F2.
- **Mutation-proven**: dropping `trg_mrho_reject_update` reds §F2 (the UPDATE succeeds → `.rejects.toThrow` fails); restoring it greens it.
- core-backend `tsc --noEmit` clean; migration up/down present.

### Scope
Independent of #4331 ratification (a DB constraint, not the batch/operation design of finding #1). Per the owner's finish order this lands **into the L6-a integration before L6-a merges** (step 4), re-gated at exact head. Held here as the verified fix.

### DIFF
--- a/packages/core-backend/src/db/migrations/zzzz20260715190000_create_meta_record_history_operations.ts
+++ b/packages/core-backend/src/db/migrations/zzzz20260715190000_create_meta_record_history_operations.ts
@@ -169,12 +169,36 @@ export async function up(db: Kysely<unknown>): Promise<void> {
     BEFORE INSERT ON meta_record_history_operations
     FOR EACH ROW EXECUTE FUNCTION meta_record_history_operations_validate_endpoint()
   `.execute(db)
+
+  // Finding #2 (owner review 2026-07-16): the sealed endpoint row is the immutable recovery anchor —
+  // v3.7 §1.3 takes anchorSeq from "that immutable endpoint row". The INSERT trigger above validates the
+  // seal at mint time, but nothing forbade a later UPDATE, so `UPDATE ... SET endpoint_seq=999999` (or
+  // event_count / created_at) on a sealed row would silently MOVE an already-sealed anchor. The endpoint is
+  // write-once by design (minted last, then only read), so a blanket UPDATE-reject at the DB level is the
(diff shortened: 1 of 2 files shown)


zensgit:
Superseded by #4409 (`2f456571e`) on `main`, which lands the complete L6-a (sealed operation-endpoint ledger + batch/operation decouple + endpoint UPDATE-immutability + writer wiring + H1/H2/H3 hardening). Closing per owner review.

PR #4381, opened 2026-07-16
fix(automation): P2 S2-b — 3 non-blocking P3s from #4334 round-4 review

zensgit:
Follow-up to **#4334** (merged as `e3c537430`). You approved S2-b with three explicitly **non-blocking P3s**; I kept them out of #4334 to preserve the "product diff unchanged" mechanical merge you authorized, and out of #4335 (S3) per your instruction. This is that separate follow-up.

## The 3 P3s
- **P3-1 (you gated this to "before S5"):** `onClaimTimePoison` is now a **required** `DispatchLoopObserver` method, routed through `safeInvoke` in the loop like the other sinks — so a claim-time dead-letter is observable and a throwing sink can't break the loop. The required-sink test is extended (an observer missing `onClaimTimePoison` throws) and mutation-proven.
- **P3-2:** the ordering-control in the P1-A / head-of-line real-DB tests pinned `meta_automation_outbox.created_at`, but the claim CTE selects by `meta_automation_outbox_consumer.updated_at` (`created_at` only orders the returned batch — moot at `batchSize=1`). Repinned to `consumer.updated_at` so the tests control the column that actually drives which row is claimed.
- **P3-3 (comment drift):** dropped the "attempts == adapter calls" overclaim (a crash-after-claim still counts an attempt, by design); made the same-tick wording precise — a *rescheduled* row can't re-claim within a tick (backoff floor), distinct from a claim-time-poisoned *head* that IS scanned past same-tick; added `claimTimePoisoned` to the `onTick` metrics doc + the required-sink list.

## Verification
- Fresh PG15: **loop spec 34/34**, `tsc --noEmit` 0 errors.
- Additive / behavior-preserving except the now-**required** `onClaimTimePoison` sink — a startup-time contract (flag OFF, no production caller yet).

**For review — not self-merging.**

### DIFF
--- a/packages/core-backend/src/multitable/automation-durable-dispatch-loop.ts
+++ b/packages/core-backend/src/multitable/automation-durable-dispatch-loop.ts
@@ -48,9 +48,9 @@
  *     at tick entry (before the claim) and at loop start (before the timer). `intervalMs=0` (hot loop), a
  *     fractional/huge backoff, a sub-minimum lease, and an effective backoff inversion are rejected up front.
  *   - **The long-running loop requires a telemetry sink**: `onUnknownConsumerKeys` + `onTickError` +
- *     `onHeartbeatError` are mandatory (a silent stuck row, a swallowed tick error, or a hidden heartbeat DB
- *     error is an operational blind spot), and every observer call is exception-safe — a broken sink can
- *     neither kill the process nor stop delivery.
+ *     `onHeartbeatError` + `onClaimTimePoison` are mandatory (a silent stuck row, a swallowed tick error, a
+ *     hidden heartbeat DB error, or an unobserved claim-time dead-letter is an operational blind spot), and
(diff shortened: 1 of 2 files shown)


zensgit (commented):
## Codex review — APPROVE

Reviewed exact head `67c18db885786c35417b0b1973d96ae1d39b63fc`.

No P1/P2 findings. The follow-up does what it claims:

- `onClaimTimePoison` is required at long-loop startup and routed through the exception-safe `safeInvoke` boundary;
- claim-time telemetry cannot veto or roll back the already-completed DB transition;
- ordering-sensitive tests now pin `meta_automation_outbox_consumer.updated_at`, which is the actual claim order;
- comments match the per-slot claim and reschedule semantics.

Independent verification: fresh PostgreSQL 15 real-DB dispatch-loop spec `34/34`, backend `tsc --noEmit` clean, `git diff --check` clean, and all current GitHub checks green. The PR is `BEHIND`, so rebase/re-green before merge; product code is approved.

PR #4382, opened 2026-07-16
fix(stock-prep): harden immutable replay checks (T3b-1a)

zensgit:
## What changed

- Replace the existing batch-key-only replay skip with a complete frozen-projection check for the snapshot batch, all paged snapshot lines, and the run row.
- Require exactly one project row before an exact replay can skip; duplicate project keys are rejected before the first business-row write.
- Fail closed on incomplete commits, duplicate keys, content mismatches, and an unprovable page bound without repairing immutable rows.
- Add required-CI real PostgreSQL coverage using real provisioning, real records APIs, and physical field IDs discovered from `meta_fields`.

## Why

The previous `/mvp/sync/persist` behavior treated any existing `snapshotBatchId` as a successful replay. A crash after the batch insert, or a same-key request whose persisted fields differ while its fingerprint stays unchanged, could therefore return a false success.

This is the T3b-1a behavioral hardening ratified in #4364. It intentionally changes incomplete crash retries from a silent skip to a values-free 409. The bounded recovery path remains a new `snapshotBatchId`.

## Boundaries

- No new endpoint or runtime flag.
- No external PLM, K3, ERP, SQL-adapter, or production-write path.
- Snapshot batch/line/run rows remain create-only and are never repaired or patched.
- Project remains a mutable live pointer, but exact replay only checks its cardinality and never rolls it back.
- T3b-1c route wiring is not included.

## Verification

- Real PostgreSQL: `stock-preparation-t3b-replay-hardening-realdb.test.ts` passed against the local Postgres service.
- Required `plugin-tests.yml` allowlist includes the new real-DB suite.
- Load-bearing mutations:
  - ignoring frozen projection contents produced 5 failures;
  - treating a full page as complete produced 2 failures.
- Focused persist suite: 29/29 passed.
- `git diff --check` passed.


zensgit:
## Independent adversarial review at exact head `1a7cedecb` — MERGE-READY, no P1/P2

OD-4 verified item-by-item against the ratified lock; all three load-bearing guards **mutation-proven** (each landed → red → restored): neutering line `projectionsEqual` reds exactly the 4 fingerprint-discriminator cases; neutering `assertExactReplay` reds 10 (content-mismatch, all INCOMPLETE, dup keys, paging); first-page-only read reds exactly the 2 paging tests. The vacuous-comparison landmine is refuted: the scoped `queryRecords` translates physical→logical on read (`table-actions.cjs:475-483`) and the fixture mirrors the real service, so full-projection equality is real, not always-equal. Real-DB test is honest (`describeIfDatabase` + fail-not-skip sentinel + physical `meta_records` assertions + conflict-not-repair) and its whitelist line sits inside the gated step that sets `DATABASE_URL` + `METASHEET_REAL_DB_TEST_STEP=1`. No existing assertions weakened (the −6 lines are benign); persist return-shape consumers pass; new 409/422 details are values-free (leak-bait clean). Focused suite 29/29; neighbors green.

Non-blocking: **P3-1 (latent)** `normalizeTemplateValue` does no date canonicalization — inert today (the plan never writes date fields), but if a future slice writes one and the records service canonicalizes serialization, exact replay would false-conflict; note for that future slice. **NIT** stale route header comment (`http-routes.cjs:~3838`, "existing snapshotBatchId skips the whole commit") — no longer true; fix in T3b-1c which touches that header anyway.

Operator-facing note (as ratified): this intentionally changes shipped `/mvp/sync/persist` replay — crash-retry that previously false-skipped now 409s; recovery = re-sync under a new `snapshotBatchId`. Review MD: `/tmp/pr4382-review-claude-20260716.md`.

PR #4383, opened 2026-07-16
feat(stock-prep): add pure PLM persist bridge (T3b-1b)

zensgit:
## What changed

- Add a pure `buildPlmSourcePersistInput({ request, intake })` bridge from approved PLM readonly intake to the existing sync-persist input.
- Cross-check request, project, batch, run, version, source-system, and every line's project/batch identities.
- Project each line through the existing mapper's closed field set; unknown fields are never forwarded and no invalid row is dropped.
- Canonicalize `imported` to `active`, preserve `active/inactive/incomplete`, and reject unsupported lifecycle values with a coarse values-free 422.
- Add the bridge suite to the required plugin-integration-core test chain.

## Why

The approved PLM source intake and the existing persistence planner use related but distinct shapes. Direct spread/cast would forward fields outside the frozen contract and could silently preserve unsupported source lifecycle values. T3b-1b provides the ratified pure boundary before any route or records I/O is introduced.

## Boundaries

- Pure helper only: no route, flag, provisioning, records API, persistence call, network, or external write.
- No tenant, workspace, base, target-project, config, or system-credential forwarding.
- No lifecycle guesses for `released`, `obsolete`, or `WIP`.
- T3b-1c route wiring is not included.

## Verification

- Focused bridge suite: 28/28 passed.
- Mapper round-trip preserves row count, identity, quantity, unit, path, fingerprint, and canonical status.
- Unsupported `released/obsolete/WIP` values reject the entire envelope without echoing source values.
- Load-bearing mutations killed:
  - preserving `imported` instead of canonicalizing;
  - silently converting unsupported lifecycle values to `active`;
  - spreading the complete intake row instead of the closed projection.
- Static dependency/I/O tripwire passed.
- `git diff --check` passed.


zensgit:
## Independent adversarial review at exact head `2f98dc92e` — MERGE-READY, no P1/P2

OD-3 verified item-by-item: genuinely pure (no I/O, deterministic), exact 6-key output whitelist, no-row-drop fail-closed, identity cross-checks case-sensitive, and the ratified status contract enforced (`imported` sole alias → `active`; `released`/case-variant/non-string/`missing_child_bom` all 422 with the dedicated `STOCK_PREPARATION_PLM_AUTOPERSIST_LINE_STATUS_UNSUPPORTED`; a valid+unsupported mix rejects the whole envelope). Hostile field names (`__proto__`/`constructor`/`Object.create(null)`) reduce to the whitelist with no prototype pollution — the R5 lesson holds. The round-trip test is **non-vacuous**: it feeds bridge output through the REAL `mapExpansionRowsToSnapshotLines`, which is exactly the re-normalization persist runs (`sync-run-plan.cjs:261`). All 3 required mutations proven load-bearing (status-vocab guard, no-row-drop, identity cross-check). Focused suite 28/28; scope is exactly 3 files, no route/flag wiring smuggled in.

**P3-1 — carry to T3b-1c (verify before wiring):** real intake (`readonly-intake.cjs:202`) can emit `lineStatus:'missing_child_bom'` when an input row carries `missingChildBom===true`. The bridge correctly 422s it per contract, but at 1c a legitimate envelope containing such a line would whole-envelope-reject; the sibling expansion mapper uses in-vocab `'incomplete'` for the same condition. 1c must either prove real intake never yields it on the source-run path or reconcile the two vocabularies. NITs (cosmetic): redundant row-count guard `bridge.cjs:146-148`; purity test is a source-text assertion; round-trip fixture lacks an `incomplete` line; absent `lineStatus` reported as UNSUPPORTED vs required.

Review MD: `/tmp/pr4383-review-claude-20260716.md`.

PR #4390, opened 2026-07-16
feat(ops): corrective-5 — bounded values-free smoke diagnostic contract (Draft)

zensgit:
## corrective-5 — bounded values-free diagnostic contract (Draft; local real-service PASS still pending)

**Refs #4101.** RC-0 stays FAIL. This does NOT request an entity rerun and does NOT cut a package.

### Why
corrective-4 fixed the PS 5.1 capture, but the packaged smoke still exits 1 with `auditActionsCovered=N/8` — **uninterpretable**, because the runner discards the child's stdout/stderr. `N/8` means the audit-assignment line was never reached (an early-return/throw in an earlier phase), NOT `0/8`; and `selfScanClean=true` only proves the finish() backstop ran, not that the phase-10 full-response leak-scan completed.

### What
A **bounded values-free diagnostic contract** the runner propagates even on exit 1:
```
failureClass=NONE|CHECK_FAILED|SELF_SCAN_FAILED|FATAL_EXCEPTION
lastCompletedPhase=<fixed 13-phase enum | NONE>
firstFailedCheck=<phase of the first failed check — a fixed enum | NONE>
failedCheckCount=<non-negative integer>
responseLeakScanStatus=NOT_RUN|PASS|FAIL
```
- **Smoke** (`stock-preparation-mvp-postdeploy-smoke.mjs`): a fixed `SMOKE_PHASES` ladder + `beginPhase()` markers; `computeDiagnosticLocus()` runs **before** the finish()/catch self-scan so the scan covers the diagnostics. `firstFailedCheck` is the **phase** of the first failed check (each check stamps its phase), never a free-form check name — a leak-hardening test proved a safe-label-shaped business value in a free-form name would surface.
- **Runner** (`stock-preparation-onprem-acceptance.ps1`): `Test-SmokeOutcome` parses the 5 fields; `Invoke-SmokeStage` propagates them **before** the exit/outcome gates. The 4 enum fields are allowlisted (off-vocabulary → `UNKNOWN`); a non-enum-shaped value → `NOT_RUN`. No raw value can reach the values-free summary.

### Validated
- End-to-end on both failure modes: unreachable URL → `FATAL_EXCEPTION`; a 200-stub that fails provisioning → `CHECK_FAILED`, `lastCompletedPhase=AUTH`, `firstFailedCheck=PROVISIONING` (reproduces the entity's early-return mode and localizes it).
- Tests: smoke unit **20/0**; runner static **14-field** whitelist; runner behavior **40/40** (incl. exit-1 propagation + a business-value-diagnostic leak-hardening test); **ps51** adds real-Windows-PowerShell-5.1 diagnostic-stdout capture + exact non-zero exit (and fixes the corrective-4 NIT-1 `\\n`→`\n`).
- Mutations proven load-bearing: drop the runner propagation → propagation test reds; drop the phase allowlist → leak-hardening test reds.

### Still open (the owner's gate before the next package)
Per the sequence, a **local real-service rehearsal reaching `pass=true` / `8/8` / `selfScanClean=true`** is required before cutting the corrective-5 entity package. The diagnostic contract now makes any run — local or entity — self-localizing. Standing up the full backend locally is the next step; not done in this PR.

### DIFF
--- a/scripts/ops/__tests__/stock-preparation-onprem-acceptance.behavior.tests.ps1
+++ b/scripts/ops/__tests__/stock-preparation-onprem-acceptance.behavior.tests.ps1
@@ -302,6 +302,58 @@ Invoke-SmokeStage -Token $token
     "$stageOutput" -notmatch 'dummy-not-a-real-token'
   )
 
+  # ── corrective-5: a smoke that early-returns (exit 1) with a valid-looking summary must still PROPAGATE
+  # its bounded diagnostics into the acceptance summary, so the failure is localizable even though the runner
+  # discards the child's raw output. Drive the REAL Invoke-SmokeStage; assert the emitted .json carries them.
+  $diagLines = @(
+    "process.stdout.write('mvpSmoke.pass=false\n')",
+    "process.stdout.write('auditActionsCovered=N/8\n')",
+    "process.stdout.write('selfScanClean=true\n')",
+    "process.stdout.write('failureClass=CHECK_FAILED\n')",
+    "process.stdout.write('lastCompletedPhase=AUTH\n')",
+    "process.stdout.write('firstFailedCheck=PROVISIONING\n')",
+    "process.stdout.write('failedCheckCount=2\n')",
+    "process.stdout.write('responseLeakScanStatus=NOT_RUN\n')",
+    'process.exit(1)'
+  )
+  Set-Content -LiteralPath $fakeSmoke -Encoding utf8 -Value ($diagLines -join "`n")
+  $diagSummary = Join-Path $fakeSmokeRoot 'diag-summary.txt'
+  & pwsh -NoProfile -File $stageHarness -AcceptanceScript $scriptPath -DeployRoot $fakeSmokeRoot -SummaryPath $diagSummary *> $null
+  $diagExit = $LASTEXITCODE
+  $diagJsonPath = [System.IO.Path]::ChangeExtension($diagSummary, '.json')
+  $diagJson = if (Test-Path $diagJsonPath) { Get-Content $diagJsonPath -Raw | ConvertFrom-Json } else { $null }
+  Check "corrective-5: an exit-1 smoke still fails closed AND propagates its bounded diagnostics into the summary" (
+    $diagExit -eq 1 -and
+    $null -ne $diagJson -and
+    $diagJson.failedStage -eq 'smoke' -and
+    $diagJson.'mvpSmoke.failureClass' -eq 'CHECK_FAILED' -and
+    $diagJson.'mvpSmoke.lastCompletedPhase' -eq 'AUTH' -and
+    $diagJson.'mvpSmoke.firstFailedCheck' -eq 'PROVISIONING' -and
+    $diagJson.'mvpSmoke.failedCheckCount' -eq '2' -and
+    $diagJson.'mvpSmoke.responseLeakScanStatus' -eq 'NOT_RUN'
+  )
+  # values-free hardening: an off-vocabulary / business-value diagnostic never reaches the summary.
+  $leakLines = @(
+    "process.stdout.write('mvpSmoke.pass=false\n')",
+    "process.stdout.write('selfScanClean=true\n')",
+    "process.stdout.write('failureClass=DWG-88472-A\n')",
+    "process.stdout.write('lastCompletedPhase=SECRET_TENANT_evil\n')",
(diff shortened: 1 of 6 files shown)


zensgit:
## Correction: `d85b5d9ee` claimed the P2-1/P3 fixes but did not contain them — real fixes at `1f9ae03ea`

Your re-paste was right and my earlier 'already fixed' claim was wrong. What happened: after editing the smoke, I ran the mutation check and then `git checkout --` on the **still-uncommitted** file — silently reverting the P2-1/P3 edits — and misread the post-revert test output (21/1) as green. `d85b5d9ee` therefore carried only the new TESTS with an inaccurate message, and CI's `prep-line extended smoke contract` check went red on exactly the discriminating test.

**`1f9ae03ea` now actually contains:**
- P2-1: `lastPhaseCompleted = reachedIdx >= last && typeof S.leakScanClean === 'boolean'` — mid-`RESPONSE_LEAK_SCAN` throw reports `AUDIT_TRAIL + NOT_RUN`; the contradictory pair is structurally impossible; `FAIL` product still counts as completed.
- P3: `beginPhase` writes the internal `DIAG` cursor only; `check()` reads it there; no `reachedPhase` in the output face.

**Evidence (read carefully this time):** `node --test` **22/22** with both named tests explicitly ✔; file truth-check (gate present ×1, `RESULT.summary.reachedPhase` ×0); mutation (drop the gate) → **21/1 red on exactly the contradictory-pair test**; post-revert → **22/0**, gate still present, worktree clean. The previously-red CI check should now go green at this head.

Process fix adopted: **commit before mutating** — a `git checkout --` revert is only safe against a committed baseline; and mutation evidence must quote the pass/fail counts verbatim, not summarize them.

zensgit:
## Local real-service rehearsal: FULL PASS at head `12f41475a` — and the entity `N/8` root-caused

Per the required sequence, the rehearsal ran the REAL packaged smoke against a real local backend (dist build, `NODE_ENV=production`) + fresh PostgreSQL (268 migrations, zero exclusions, no MIGRATION_EXCLUDE).

### Entity failure reproduced, localized by the new diagnostics, then fixed
First run reproduced the entity shape byte-for-byte: `failureClass=CHECK_FAILED / lastCompletedPhase=AUTH / firstFailedCheck=PROVISIONING` → exit 1 with `auditActionsCovered=N/8`, `selfScanClean=true`. Root cause (each link code-verified): since the tenant-derivation hardening landed, the MVP write routes derive tenant from the authenticated principal only; in production mode the token's DB-load path carries no intrinsic tenant, and the sanctioned tenant-context mechanism is the request header — but the smoke carried the tenant **only in the query string**, which hardened write routes ignore by design. So the FIRST write route 400s → early return → `N/8`. This explains why corrective-7 passed on 2026-07-12 and every post-hardening package fails the same way. **Harness fix, not product**: the smoke now sends the tenant-context header on every request (`12f41475a`).

### Final values-free summary (verbatim)
```
pass=true
auditActionsCovered=8/8
selfScanClean=true
responseLeakScanStatus=PASS
failureClass=NONE
lastCompletedPhase=RESPONSE_LEAK_SCAN
failedCheckCount=0
```
48 responses leak-scanned clean; persist 201 created → replay skipped_existing; ERP sync 201 → resync refreshed; suites at this head: smoke 22/0, static contract ALL PASS, behavior 40/40.

The owner's precondition (local real-service `pass=true / 8/8 / selfScanClean=true`) is met. Still holding: no package cut, no entity-machine contact — those remain the owner's calls after this PR's exact-head review.

zensgit:
## P2 addressed — the tenant-context header is now mutation-pinned (`b4bb036c9`)

You were right: deleting `x-tenant-id` stayed 22/22 because nothing exercised requestJson's headers. Fixed: `requestJson` builds headers via a pure `buildRequestHeaders()` and takes an injectable `fetchImpl`; 4 tests drive the **real** `requestJson` and assert the header contract on a POST write (present when a tenant is configured, absent when not, Authorization always present).

Mutation proof (verbatim): delete the `if (tenantId) headers['x-tenant-id'] = tenantId` line → `pass 24 / fail 2` (the POST root-cause guard + the header-builder test red); `git checkout --` revert → `pass 26 / fail 0`, header line present ×1, worktree clean. Suites at head: smoke 26/0, static contract ALL PASS, behavior 40/40. Local full-rehearsal PASS conclusion unaffected (harness-only change). No corrective-4 rerun needed.

PR #4391, opened 2026-07-16
feat(stock-prep): T3b-1c prep — structural config guard for fieldMap.target=missingChildBom (Draft)

zensgit:
## T3b-1c prep — structural config guard: `fieldMap.target=missingChildBom` (owner P2)

Per the #4390-round review: Option A (whole-envelope 422 on `lineStatus='missing_child_bom'`) is kept, but doc-only was not enough — the owner constructed a **combined-mapping bypass**: an approved config mapping a target literally named `missingChildBom` (boolean marker) *together with* an explicit `lineStatus=active`. Intake keeps the explicit status, the marker is silently dropped downstream, and the missing-child signal is erased end-to-end — the value-level 422 never fires.

### What
- **Pure guard** `assertPlmAutoPersistSourceConfigSafe(config)` exported from the bridge module: `fieldMap.target ∈ FORBIDDEN_FIELD_MAP_TARGETS` (= `['missingChildBom']`) → dedicated values-free 422 `STOCK_PREPARATION_PLM_AUTOPERSIST_CONFIG_TARGET_FORBIDDEN` (details carry only the forbidden TARGET vocabulary, never the source column name); missing/non-array `fieldMap` → 422 `..._CONFIG_SHAPE_INVALID` (fail-closed). The T3b-1c route MUST call it on the approved config **before any source read / provisioning / persist I/O**; the wiring test must prove the rejection path has zero records/provisioning/adapter calls.
- **Design-lock OD-3 amended in place**: the guard requirement + zero-I/O test + a dedicated must-kill mutation item ("去掉/绕过 assertPlmAutoPersistSourceConfigSafe"). No `missing_child_bom → incomplete` silent normalization (Option B rejected — signal-loss hazard).
- **Tests** (in the bridge chain, runs under integration-guard): clean config passes; forbidden target → dedicated code with values-free details; shape fail-closed on 4 malformed configs; **the owner's bypass repro pinned as a test** (marker+active PASSES the value-level bridge — documenting exactly why the config guard is load-bearing); module purity re-checked (no new requires/I/O tokens).
- **Mutation-proven**: emptying `FORBIDDEN_FIELD_MAP_TARGETS` reds the guard test (landed-anchor, reverted, green).

The existing `lineStatus='missing_child_bom'` value-level whole-envelope 422 is unchanged. No route wiring in this PR (that's T3b-1c proper); flag posture unchanged.

### DIFF
--- a/docs/development/stock-preparation-t3b-plm-source-autopersist-design-lock-20260716.md
+++ b/docs/development/stock-preparation-t3b-plm-source-autopersist-design-lock-20260716.md
@@ -185,6 +185,19 @@ closed set；实现与验收不得依赖 config shape validation 或 records ser
 某个 PLM 的 `released` / `obsolete` / `WIP` 语义时，必须先另开具名 mapping design gate；本锁
 不授权隐式映射。
 
+**结构性 config guard（owner P2，2026-07-16 复审补强）**：value-level 422 抓不住**组合映射
+绕过** —— approved config 同时把某列 target 配成内部 marker 名 `missingChildBom`（布尔）并显式
+映射 `lineStatus`（如 `active`）时，intake 优先保留显式 status、marker 在下游被静默丢弃，
+缺子 BOM 信号端到端消失且整包 422 永不触发。因此 **T3b-1c 路由必须在任何 source read /
+provisioning / persist I/O 之前**，对 approved config 调用纯守卫
+`assertPlmAutoPersistSourceConfigSafe(config)`（bridge 模块导出）：`fieldMap.target ∈
+FORBIDDEN_FIELD_MAP_TARGETS`（当前 = `missingChildBom`）→ 专用 values-free 422
+`STOCK_PREPARATION_PLM_AUTOPERSIST_CONFIG_TARGET_FORBIDDEN`；fieldMap 缺失/非数组 → 422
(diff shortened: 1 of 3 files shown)


zensgit:
## P2 addressed — full shape fail-closed (`36d2a91cc`)

You were right that the guard is meant to be an independent pre-I/O boundary and must not lean on the upstream validator. It now rejects, with `STOCK_PREPARATION_PLM_AUTOPERSIST_CONFIG_SHAPE_INVALID` (coarse, values-free reason):
- non-object config, non-array `fieldMap`, **empty `fieldMap`** (previously accepted),
- **any malformed entry** — non-object, or a non-string / empty `target` (previously accepted: `[null]`, `[{}]`, `[{target:42}]`),
- a valid+malformed mix → whole config rejected.

Your exact inputs (`fieldMap:[]`, `[null]`, `[{}]`, non-string target) now all throw SHAPE_INVALID; a clean config still passes; the forbidden-target 422 and the bypass-repro/purity tests are unchanged. Mutation: dropping the empty-array check reds the new shape test (`expected SHAPE_INVALID for {fieldMap:[]}`), reverted green. Lock OD-3 text broadened to match. The unrelated `test (18.x)` attendance flake reruns on this fresh push.

Standing note (yours): the REAL T3b-1c must still prove the route calls this guard **before any source/provisioning/records I/O**, with the rejection path's call counts all 0 — that lives in the 1c wiring PR, not here.

PR #4395, opened 2026-07-16
ops(discussion-pilot): read-only status probe (no inputs, no injection surface)

zensgit:
## What
The **narrow, read-only feasibility probe** split out of #4394 per the security review. It answers one question before any provisioning: **is this host's `metasheet-backend` in platform mode with the PLM embed routes mounted (pilot feasible) or attendance mode (not feasible here)?**

## Why this is safe to review/merge on its own (addresses the #4394 injection P1)
- **Zero workflow inputs** → nothing to interpolate into the SSH command → **no shell-injection surface** (the #4394 P1 was free-text inputs single-quoted into the remote command).
- Only the `DEPLOY_*` SSH secrets (same channel as `attendance-remote-docker-gc-prod.yml`); **no `MS2_ADMIN_BEARER`**.
- The remote command is **fixed**: `scp` a committed read-only script, run it with **no arguments**, remove it.
- `status_probe.sh` **mutates nothing** — `docker ps`/`docker inspect` (no `exec`), a loopback `curl` to `127.0.0.1:8900`, `df`, and an `nginx` grep; writes only a `/tmp` evidence file that is uploaded as an artifact.

## Reading the result
`plm-embed read route -> HTTP 401/403` = mounted (platform) → pilot feasible; `HTTP 404` = not mounted (attendance) → pilot needs a **separate platform-mode instance**, not this backend.

## Not in scope
The full provision / light / rollback channel (#4394) stays **blocked** pending its 4 P1 fixes (injection-safe input passing, real `docker/app.env` path, real data-sources API contract, license-file survives container recreate) + your re-review. This PR does not provision, light, or mutate anything.

### DIFF
--- a/.github/workflows/discussion-pilot-status.yml
+++ b/.github/workflows/discussion-pilot-status.yml
@@ -1,0 +1,92 @@
+name: Discussion Pilot Status (read-only probe)
+
+# Extremely narrow, READ-ONLY feasibility probe for the Discussion 点灯 pilot.
+# Deliberately isolated from the full provision/light/rollback channel:
+#   * NO workflow inputs at all  -> nothing to interpolate -> no shell-injection surface.
+#   * NO MS2_ADMIN_BEARER or any secret beyond the DEPLOY_* SSH creds.
+#   * The host key is PINNED (DEPLOY_KNOWN_HOSTS + StrictHostKeyChecking=yes + BatchMode) so
+#     a DNS/MITM attacker cannot forge the platform/attendance verdict.
+#   * The probe is piped over stdin (`bash -s`): NOTHING is written to the host filesystem
+#     for the script, so there is no predictable path / symlink / race to exploit. The probe
+#     mutates nothing and exits non-zero on any dimension it cannot positively verify.
+# Answers one question: is this host's metasheet-backend in platform mode with the PLM embed
+# routes mounted (pilot feasible) or attendance mode (not feasible here).
+#
+# NOT scheduled. NOT push/PR triggered.
+
+on:
+  workflow_dispatch: {}
+
+concurrency:
+  group: discussion-pilot-status
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+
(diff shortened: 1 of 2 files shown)


zensgit:
**Re-review requested — all four blockers fixed, rebased on current main (`535c9596e`).**

1. **Predictable /tmp path/exec → gone.** The probe is now piped over stdin (`ssh … 'bash -s' < status_probe.sh`) — **no script file is written to the host at all**, so there's no fixed path/symlink/race. Any scratch the probe needs uses a private `umask 077` `mktemp -d` under `$HOME`, cleaned by a `trap`. No fixed `/tmp` file is created or executed.
2. **SSH host identity pinned.** Added the `DEPLOY_KNOWN_HOSTS` secret + `StrictHostKeyChecking=yes` + `UserKnownHostsFile` + `BatchMode=yes`. No `StrictHostKeyChecking=no` anywhere (this is the first workflow here to pin the key — the existing `attendance-remote-*` ones still carry that debt). The workflow hard-fails if `DEPLOY_KNOWN_HOSTS` is unset (no silent fallback).
3. **No false-green on unverifiable.** The probe distinguishes `none` / `unavailable` / `inconclusive` and exits **non-zero** on any dimension it can't positively determine — docker permission (exit 3), backend-not-found / inspect-fail / route-unreachable, `df` failure, and **nginx not readable** (no more `|| true` swallowing). A determined "attendance / not feasible" is a valid exit 0; "couldn't verify" is never dressed up as success. A final workflow step fails the run when the probe rc ≠ 0. (Verified locally: on a box with docker but no `metasheet-backend`, it correctly exited 1 UNVERIFIED, not 0.)
4. **Rebased** onto current main (was 7 behind); merge-tree clean; CI re-running on the new base.

Minor: `say()` now uses `printf`, not `echo`.

Still just the read-only probe — no provision/light/rollback. #4394 remains blocked pending its own P1 fixes + re-review.

PR #4409, opened 2026-07-17
W0-1 L6-a — sealed operation-endpoint ledger + owner hardening (DRAFT, default-OFF)

zensgit:
## What

W0-1 Lane L6-a redo on the merged main (L3+L4+L4cov+L5 substrate). The **sealed operation-endpoint ledger** (design lock v3.7 §1.2): a server-minted, sheet-scoped `meta_record_history_operations(sheet_id, operation_id, endpoint_seq BIGINT, event_count, created_at)` PK(sheet_id, operation_id) that records the exact, externally-visible commit boundary a recovery anchor can trust — the endpoint row of one committed operation, written **last**, only visible once the whole operation commits.

Re-created from the finding branch (superset), rebuilt on current main because the finding branches are 47 commits behind and lack L5.

### Reused core
- **Migration** — ledger table, nullable `operation_id` on both event tables, **DEFERRABLE INITIALLY DEFERRED** events→endpoint FKs, append-after-seal BEFORE-INSERT trigger, endpoint-validation BEFORE-INSERT trigger (event_count == actual + endpoint_seq == exact MAX(seq)), and the finding-#2 **BEFORE UPDATE reject** trigger (sealed endpoint immutable). Fresh timestamp `210000` (after L5's `180000` / fwb's `200000`).
- **`operation-ledger.ts`** mint→events→seal-last protocol; wiring into `record-history-service` / `record-service` / `record-write-service` / `records.ts` / `routes/univer-meta.ts`.
- **finding-#1 §10 ruling** — `batch_id` is DECOUPLED from `operation_id` (both inserts write `batchId = input.batchId ?? id`; the patch echo returns the S1 `bulkBatchId`). One commit action across N transactions ⇒ ONE batch_id, N operation_ids + N sealed endpoints, one History grouping key.
- **Goldens** — §F2 (UPDATE-immutability, mutation-proven) + G-MULTIOP (multi-transaction commit grouping).

### Owner hardening (each mutation-proven red→green)
- **H1-DELETE** — BEFORE DELETE reject trigger; the endpoint is immutable to DELETE too. Ordinary DELETE fails closed; a dedicated retention path (gated on a transaction-local GUC) is the only sanctioned deletion.
- **H2 whole-operation retention** — `meta_record_history_operations_prune(...)` atomically prunes a whole sealed operation (endpoint + all its events); a partial prune is impossible.
- **H3 fail-closed schema probe** — `mintOperation` reversed from silent-degrade to fail-closed: flag ON + incomplete schema now throws and rolls the writer txn back (was: inert ledger).

## Safety
DRAFT. All recovery behaviour behind `MULTITABLE_ENABLE_WRITER_FENCE` (default **OFF**). Flag-off is byte-identical to L5-on-main (no operation rows, no endpoints). No arming, no auto-merge, no host/staging/prod touch.

## Verify
Fresh-DB migrate up/down/replay clean; real-DB goldens run in `plugin-tests.yml` multitable real-DB run-list (fail-not-skip sentinel). `@metasheet/core-backend` type-check clean.


zensgit:
## Independent 4-lens adversarial gate — verdict: CLEAR (0 P1 / 0 P2)

Head-scoped to `85a3ac8b6` (no drift). Four parallel refute-first lenses, each on its own real DB:

- **H1 — immutability + the retention-GUC bypass (prime suspect): CLEAR.** Ordinary DELETE/UPDATE of a sealed endpoint RAISE (default GUC NULL ⇒ reject trigger fails closed). The bypass is closed: `metasheet.mrho_retention` is set in EXACTLY ONE place — the `meta_record_history_operations_prune` DB fn via transaction-local `set_config(...,true)`, reset to `'off'` at fn end; the only TS wrapper `pruneSealedOperation` has **no route/service/client caller**; no `SET LOCAL` anywhere; leak tests confirm no persistence across a pooled connection and no mid-txn leak. The BEFORE DELETE trigger is the SOLE guard for the events-then-endpoint delete the deferred FK can't catch (mutation-proven: drop it → that one-txn sequence commits silently).
- **H2 — retention atomicity: CLEAR.** Prune removes endpoint + all events atomically; partial-prune impossible (SQL mutant skipping the revisions delete trips the DEFERRABLE FK and rolls back; the orphan direction is closed by plpgsql atomicity); younger op survives via exact (sheet_id, operation_id) keying.
- **H3 — fail-closed schema probe: CLEAR.** Every reachable incomplete-schema shape fails closed with `OperationLedgerSchemaError` + writer-txn rollback (column-absent → mint throws; table-absent → seal throws, base row persisted=0; trigger-absent unreachable via transactional DDL). Positive control seals; flag-OFF short-circuits before the probe (byte-identical no-op).
- **CORE + regression + finding-#1: CLEAR.** Endpoint-validation trigger load-bearing (mutations red E1–E4, incl. proving the UNION covers markers); finding-#1 §10 decouple verified (`batchId = input.batchId ?? id`, no `operationId ??` residue, patch echo `bulkBatchId`, G-MULTIOP green); exact bigint; flag-off parity via a real bulk writer; migrate up/down/replay clean; regression L4+L4cov+batch-grouping+d1-delete **69 passed / 1 skipped**; spec in the CI run-list (red-not-skip-green); zzzz `…210000` ordering correct.

**Carry-forwards (non-blocking, not committed over the gated SHA):**
- **P3 (H2):** the GUC-reset line is load-bearing (a mutant removing it bypasses H1 inside a multi-statement caller txn) but UNTESTED — the golden prune call-sites are autocommit, so the reset never executes. Capped P3: `pruneSealedOperation` has zero production callers, the common leak is FK-backstopped, DRAFT/default-OFF. **→ carried to the retention-wiring slice** (a multi-statement-txn golden lands with the first production caller).
- **NIT:** stale comment + vestigial hoist at `record-write-service.ts:777`.

All merge conditions on content met. Full CI next → merge. Flags OFF; nothing armed. 🤖 gate via [Claude Code](https://claude.com/claude-code)

PR #4425, opened 2026-07-17
docs(canonical-org): lock B5/B6 routing-core decisions and landing ledger

zensgit:
## Canonical Org B5/B6 design lock — owner decisions recorded

Q1–Q6 are locked and the document now matches the developed, reworked stack:

- #4429 B5-a schema
- #4430 B5-b resolver
- #4431 B5-c admin routes
- #4434 B6 equivalence proof
- #4436 B7 suggest-only reconciliation

The 2026-07-19 refresh removes the stale `DRAFT` / `CHANGES in progress` posture and records the exact held heads plus the required landing protocol: one PR at a time, retarget/rebase, replay load-bearing mutations, run full required CI, then merge.

Docs-only. Current head `1ab13f4c9`; rebased onto the then-current `main`. **Unarmed until this refreshed head finishes CI.**

PR #4429, opened 2026-07-17
feat(directory): B5-a — org_directory_routing_policy schema (explicit (org,purpose) policy store)

zensgit:
## B5-a — explicit `(org, purpose)` routing-policy schema

Introduces `org_directory_routing_policy` with the owner-locked invariants:

- one policy per `(org_id, purpose)`;
- closed purpose set;
- canonical and optional fallback integrations are constrained to the same org by composite FKs;
- both integration FKs use `ON DELETE RESTRICT` so policy targets cannot disappear silently;
- no redundant `mode` column; the integration provider remains the source of truth;
- no-policy behavior remains legacy opt-in posture for B5-b.

### Landing-window verification at `4fc7ba742`

- rebased onto post-#4425 `main` with CI-list union intact;
- fresh PostgreSQL migration + second replay: pass;
- real-DB schema suite: 9/9;
- CI wiring guard: 2/2;
- backend typecheck: pass;
- mutation 1: drop canonical same-org FK → cross-org/right-reason tests red;
- mutation 2: change fallback FK to CASCADE → FK action + behavioral fallback-RESTRICT tests red;
- mutation 3: drop `(org,purpose)` unique constraint → duplicate/schema-contract tests red.

The only new landing-window edit is comment truthfulness (`Q1/Q3 recommendation` → already-ratified owner rulings); production DDL is unchanged from the reviewed B5-a patch.

**UNARMED until required CI finishes on this exact head.**

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -174,6 +174,11 @@ jobs:
       - name: B4 department-bindings CI wiring contract
         run: node --test scripts/ops/b4-department-bindings-ci-wiring.test.mjs
 
+      # B5-a CI two-point wiring contract (no DB): the routing-policy schema suite must stay wired
+      # into BOTH vitest.config.ts (exclude) AND plugin-tests.yml (directory real-DB step).
+      - name: B5-a routing-policy CI wiring contract
+        run: node --test scripts/ops/b5a-routing-policy-ci-wiring.test.mjs
+
       # The production monitor invokes the same state helper covered here.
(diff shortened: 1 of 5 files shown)


zensgit:
Note for re-review: B5-a itself is unchanged (head `71e4a3319`, owner HOLD posture). The lock it waits on (#4425) has been corrected per its docs-accuracy gate (head `b18755d01`); the stack above was restacked + gate-fixed — per-PR verdicts on each PR.

PR #4430, opened 2026-07-17
feat(directory): B5-b — routing-policy resolver (policy-authoritative, fail-closed, legacy byte-identical)

zensgit:
## B5-b - policy-authoritative approval routing resolver

> **HELD/unarmed for exact-head CI and single-window landing.** Base is `main`. Prerequisite B5-a landed as #4429 / `81aff8203`. Current B5-b scope is 10 files; no migration is added here.

### Runtime behavior

- No `approval_routing` policy: preserve the legacy requester-account selection path.
- One policy: resolve only through its canonical integration; never fall back to "latest integration wins".
- Missing or inactive canonical integration: fail closed with typed `ApprovalRoutingPolicyError`.
- Policies in more than one linked org: fail closed as an explicit multi-org ambiguity.
- When a caller supplies `orgId`, treat it as a tenant boundary: only that org's policy may steer routing; foreign-org policies are invisible. Omitting `orgId` preserves the Q4 multi-org ambiguity contract.
- Requester absent from an otherwise healthy canonical integration: return `{}` as genuine data absence.
- Approval create and the shared preview substrate distinguish persistent policy misconfiguration (`422 APPROVAL_ROUTING_POLICY_MISCONFIGURED`) from transient directory reads (`503 APPROVAL_REQUESTER_ORG_UNRESOLVED`) before any approval instance or assignment is created.

### Review hardening

- Replaced the former shared-schema `ALTER TABLE ... RENAME` fault injection with a test-local `vi.spyOn` wrapper on the real `pg.Pool.query`. It throws only for the policy probe while a scoped flag is enabled, delegates every other overload argument with `Reflect.apply`, resets after each test, and restores the exact method after the suite.
- Added independent HTTP cases for both preview surfaces:
  - `POST /api/approvals/preview`
  - `POST /api/approval-templates/:id/route-preview`
- Each surface separately proves `422` for a broken policy, `503` for a transient read failure, and zero approval instances/assignments.
- Corrected the service comment: successful absence remains best-effort; failed reads fail closed only when the graph consumes org-derived routing data.
- Fixed the required-CI regression in the strict legacy manager-chain fixture by recognizing only the new routing-policy probe and returning no policy rows. All other unknown SQL shapes still throw.
- Rebased over the attendance direct-manager work and closed the resulting cross-org composition gap: the scoped policy probe now binds `p.org_id = $2`, with unit and real-Postgres coverage for foreign-only policy, same-org precedence, and the unscoped Q4 fail-close path.

### Load-bearing evidence

- Removing the canonical-integration predicate restores guessing and reddens the policy-authoritative resolver case.
- Removing the active-canonical guard reddens the broken-target case.
- Disabling the create/preview fail-close guard produces 6 independent failures: broken-policy create, transient create, and both preview endpoints under both `422` and `503` conditions. Positive manager and genuine-no-manager controls remain green.
- CI wiring is pinned at both points: excluded from the no-DB lane and included as a whole file in the Node 20 approval real-DB lane.
- Removing the strict fixture's routing-policy branch reproduces the exact CI failure: 8 manager-chain cases fail and the other 10 remain green; restoring it returns the file to 18/18.
- Semantic tenant-boundary mutation: replacing `p.org_id = $2` with a bind-compatible tautology makes the real-DB F case route org A to `OrgB Policy Title`; restoring the predicate returns the file to 7/7. This is not a parameter-count or SQL-bind failure.

### Fresh verification after retargeting to main

- Clean PostgreSQL migration chain: pass.
- Current-head focused unit suites: 57/57.
- Current-head resolver/fail-close/direct-manager/attendance/manager-chain real-DB suites: 41/41 on an independently created and fully migrated database.
- Grok's combined post-rebase battery: 102/102.
- Legacy manager-chain fixture: 18/18; mutation: 8 failed / 10 passed.
- Pre-final-rebase full backend no-DB lane: 495 files passed, 6790 tests passed, 184 files / 1634 tests skipped; the final exact-head full signal is delegated to required GitHub CI below.
- CI wiring contracts: 4/4.
- `pnpm --filter @metasheet/core-backend type-check`: pass.
- `git diff --check`: pass.

Implementation/test hardening was delegated to Grok Build. Codex independently reviewed the full 10-file diff, rejected two initial test-evidence defects, created a separate fresh database, reran the focused unit/real-DB suites, and replayed the runtime guard, strict-fixture, and bind-compatible tenant-boundary mutations before this push. Required GitHub checks on the new head remain the landing gate.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -184,6 +184,10 @@ jobs:
       - name: B5-b routing-resolver CI wiring contract
         run: node --test scripts/ops/b5b-routing-resolver-ci-wiring.test.mjs
 
+      # B5-b P1 fail-close CI two-point wiring contract (no DB).
+      - name: B5-b fail-close CI wiring contract
+        run: node --test scripts/ops/b5b-failclose-ci-wiring.test.mjs
+
       # The production monitor invokes the same state helper covered here.
       # Keep this in the required 18/20 matrix so HTTP classification, silence
       # barriers, and alert/recovery streak semantics cannot drift silently.
@@ -617,6 +621,7 @@ jobs:
             tests/integration/approval-p1c-field-permissions.api.test.ts \
             tests/integration/approval-wp-add-reduce-sign.api.test.ts \
             tests/integration/approval-direct-manager.api.test.ts \
+            tests/integration/approval-routing-policy-failclose.api.test.ts \
             tests/integration/approval-postgate-acceptance.api.test.ts \
             tests/integration/dept-head-sync-plumbing.test.ts \
             tests/integration/approval-manager-chain.db.test.ts \
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -179,6 +179,11 @@ jobs:
       - name: B5-a routing-policy CI wiring contract
         run: node --test scripts/ops/b5a-routing-policy-ci-wiring.test.mjs
 
+      # B5-b CI two-point wiring contract (no DB): the routing-policy RESOLVER suite must stay wired
+      # into BOTH vitest.config.ts (exclude) AND plugin-tests.yml (directory real-DB step).
+      - name: B5-b routing-resolver CI wiring contract
+        run: node --test scripts/ops/b5b-routing-resolver-ci-wiring.test.mjs
+
       # The production monitor invokes the same state helper covered here.
       # Keep this in the required 18/20 matrix so HTTP classification, silence
       # barriers, and alert/recovery streak semantics cannot drift silently.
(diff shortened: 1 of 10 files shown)


zensgit:
**Restack** (stack desync fix): rebased onto B5-a's reworked head `71e4a3319` → new head **`831056edd`**. All 4 patches byte-identical under range-diff (`=`), no conflicts. Combined replay at the stack tip (see #4436 comment) green. Held/unarmed unchanged — awaiting owner re-review; full required CI at retarget-to-main time per the stacked-CI rule.

zensgit:
**Fresh exact-head adversarial gate (restack round) — APPROVE, 0 P1/P2.** Gated at `831056edd`; 4 mutations replayed red (create-time guard deletion reproduces the exact fail-open 201/approved/empty-assignments; per-kind dept_head drop; both CI wiring points), 5/5 suite + 24/24 preview neighbors + over-blocking/later-node refute probes all held; legacy requester SQL byte-identical to main. Gate report: /tmp/pr4430-gate-restack-20260717.md.
**P3 fix applied**: wiring-guard header comment corrected (head now `3312cf6d1`). **Deferred P3s (recorded for the landing window)**: ① one preview-leg pin for the shared assembleCreationContext 422/503 behavior on B3-05/06 previews (gate probed it green at this head); ② rename-free transient-failure injection. Unarmed — owner re-review unchanged.

PR #4431, opened 2026-07-17
feat(directory): B5-c — routing-policy admin routes (list / set-clear / read-only preview)

zensgit:
## B5-c — routing-policy admin routes (routing-core slice 3 of 3)

> **HELD/unarmed until exact-head CI completes.** Restacked directly on current `main` `b09c380e2` after #4476. Current reviewed head: `424861ea3`. Delta: 7 files.

### As built

- `GET /api/admin/directory/routing-policy/`: lists all five closed purposes and each effective source.
- `PATCH /:purpose`: platform-admin-only, server-derived org, strict field allowlist. V1 writes only `approval_routing`; `canonicalIntegrationId: null` clears idempotently.
- Canonical targets must be same-org and active. Local canonical is default-OFF behind `DIRECTORY_ROUTING_LOCAL_CANONICAL_ENABLED=1`; preview remains available while OFF.
- `GET /:purpose/preview?candidate=...`: read-only current-vs-candidate comparison through the real resolver; no policy or audit write.
- Set/clear mutations emit values-free semantic audit records.

### Write-point race closure

The active-target read and policy upsert run in one transaction. The target row is selected `FOR SHARE` and held through commit:

- disable-first: PATCH waits, re-reads the committed inactive row, returns 409, writes no policy/audit;
- set-first: PATCH commits first, the later disable proceeds, and the B5-b resolver then fails closed on the disabled canonical.

The set-first test proves the exact blocking chain with backend PIDs:
`advisory holder -> PATCH -> disabler` via `pg_blocking_pids`. It does not accept a broad database waiter scan or a fixed sleep as the correctness oracle.

### Independent verification

Grok implemented and ran the focused checks; Codex then reviewed the diff and repeated the evidence on a newly created database `codex_b5c_review_20260719`:

- fresh migration chain and second replay: pass;
- B5-a schema + B5-b resolver/fail-close + B5-c routes + direct-manager compatibility: **55/55**;
- after rebasing across #4476, B5-c plus the newly landed dept-head suite: **24/24**;
- B5-c routes restored at the exact head: **9/9**;
- B5-b/B5-c two-point CI wiring guards: **4/4**;
- backend TypeScript check and `git diff --check`: pass.

Rebase evidence: aggregate patch-id stayed exactly `42ce56a430f4813b4ec60ac30461aca683bcde82`; both B5-c and #4476 CI entries remain present.

Independent load-bearing mutation replayed after the rebase: remove only SQL `FOR SHARE` while preserving the transaction. Test G fails semantically with `received 200, expected 409`; restoring it returns the route suite to 9/9.

No P1/P2 remains in the reviewed diff. Exact-head GitHub CI must settle green before the single auto-merge window is armed.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -188,6 +188,10 @@ jobs:
       - name: B5-b fail-close CI wiring contract
         run: node --test scripts/ops/b5b-failclose-ci-wiring.test.mjs
 
+      # B5-c CI two-point wiring contract (no DB).
+      - name: B5-c routing-routes CI wiring contract
+        run: node --test scripts/ops/b5c-routing-routes-ci-wiring.test.mjs
+
       # The production monitor invokes the same state helper covered here.
       # Keep this in the required 18/20 matrix so HTTP classification, silence
       # barriers, and alert/recovery streak semantics cannot drift silently.
@@ -664,6 +668,7 @@ jobs:
             tests/integration/directory-department-bindings.db.test.ts \
             tests/integration/org-directory-routing-policy-schema.db.test.ts \
             tests/integration/org-directory-routing-policy-resolver.db.test.ts \
+            tests/integration/org-directory-routing-policy-routes.db.test.ts \
             tests/integration/directory-normalized-manager.db.test.ts \
             tests/integration/attendance-csv-export-bom.test.ts \
             tests/integration/attendance-notification-redelivery.db.test.ts \
--- a/packages/core-backend/src/index.ts
+++ b/packages/core-backend/src/index.ts
@@ -172,6 +172,7 @@ import { initAdminRoutes } from './routes/admin-routes'
 import { adminUsersRouter } from './routes/admin-users'
 import { adminDirectoryRouter } from './routes/admin-directory'
 import { adminDirectoryLocalRouter } from './routes/admin-directory-local'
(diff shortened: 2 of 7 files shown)


zensgit:
**Restack** (clears the CONFLICTING state): rebased onto B5-b's reworked head `831056edd` → new head **`a9a65226b`**. 3/4 patches byte-identical; the CI-wiring commit re-anchored with a UNION resolution in `plugin-tests.yml` + `vitest.config.ts` (B5-b resolver + B5-b fail-close + B5-c routes entries all kept — asserted by the wiring guards, 14/14 green at tip). Held/unarmed unchanged — awaiting owner re-review; full required CI at retarget time.

zensgit:
**Fresh exact-head adversarial gate (restack round) — APPROVE_WITH_HOLDS → hold closed.** Gated at `a9a65226b`: 5 mutations, M1/M2/M4/M5 red with exact signatures; **M3 (strict env parse → Boolean(v)) survived 7/7 = P2 untested guard**. Fix: falsy-string legs ('false'/'0' → 409 ROUTING_POLICY_LOCAL_NOT_ENABLED) added to test C — the same mutation now reds with `env=false: expected 200 to be 409`. New head **`605b0498a`** (restacked on b5b `3312cf6d1`), suite 7/7. Gate report: /tmp/pr4431-gate-restack-20260717.md. Unarmed — owner re-review unchanged.

PR #4417, opened 2026-07-17
W0 L6-b — exact-anchor recovery API + reconstructor causal-ization (DRAFT, default-OFF)

zensgit:
## W0-1 v3.7 Lane L6-b (DRAFT — all recovery flags default-OFF; held for the owner's enablement decision)

Closes the **C2 correctness gap** the v3.7 trust model turns on: destructive recovery must resolve an **exact causal anchor**, not a wall-clock `T`. Builds on merged main (L3 seq / L4 fence / L5 trust-checkpoint / L6-a sealed operation-endpoint ledger).

### The seam is HELD `false` (owner ruling 2026-07-17)
`RECONSTRUCTION_CAUSALITY_LANDED` stays **`false`** on this PR. The causal reconstructor MECHANISM landed with L6-b, but this constant is the **last fail-closed backstop** for strict enablement, and the owner ruled it flips to `true` **only in the same future PR that actually wires the legacy Revert/Reset routes onto the L8 exact-anchor apply** (removing a backstop and adding the consumers that depend on it must be one reviewable change). Until that wiring PR, strict-on **refuses EVERY sheet** — a checkpoint-bearing sheet is STILL refused `reconstruction_non_causal`. The unit seam tripwire + the wired-layer real-DB goldens pin this held posture; a premature flip reds them. `MULTITABLE_HISTORY_CONTIGUITY_STRICT` / `MULTITABLE_ENABLE_SHEET_REVERT` / `MULTITABLE_ENABLE_PIT_RESET` all stay **default-OFF**.

### What lands (the MECHANISM + the P1 token contract, not the enablement)
1. **Causal reconstructor** (`record-reconstructor.ts`): `reconstructRecordsAtSeq` keys on `seq <= anchorSeq::bigint` (`DISTINCT ON (record_id) ORDER BY record_id, seq DESC`). The `created_at` path is KEPT for manual-datetime read-only navigation (v3.7 §9.2).
2. **Exact-anchor resolution API** (`exact-anchor-recovery.ts`): opaque `anchorOperationId` (sealed operation endpoint id) OR a History-Center `historyBatchId` → server-resolved sealed **terminal** operation (MAX `endpoint_seq`, ruling ⑤; legacy/unsealed/unknown → uniform `exact-anchor-required`, ruling ⑧, no oracle) → `endpoint_seq` (exact bigint) → active trust checkpoint → **signed preview identity**, HMAC-signed. Execute uses the **token-bound anchorSeq** — never recomputes `MAX(seq)`.
[...]

PR #4406, opened 2026-07-17
docs(multitable): W0 substrate integration — owner amendment §11 (order supersedes §7/§10) + live status/rules

zensgit:
Records the owner's plan-review corrections (2026-07-16) IN THE REPO (the prior out-of-repo artifact returned Page-not-found; #4332 is a stale retraction tracker).

**v3.7 lock §11 (new owner amendment)** — the OPERATIVE integration order, superseding §7 and §10:
- **Phase A DAG → main: L3 → L4 → L4cov → L5 → L6-a** (supersedes §10's L3→L5→L4→L4cov→L6-a). L4 is already on main default-OFF and is NOT rolled back (flags-OFF = zero prod exposure; L5 re-gated on combined L4/L4cov ancestry ⇒ safety ≥ §10). L6-a is a COMBINED rung (+#4380 +#4385), gated once on the combined head with a fresh G-MULTIOP-BATCH mutation.
- **Phase B merge order: L5-wire → L6-b → L7 → L8, with L8 ← BOTH L6-b AND L7** (fixes the loose dep; drafting parallel, merges serial + exact-head gated).
- **Take-over discipline**: ≥65-min silence = read-only review + alert only, NOT PR-ownership transfer; handoff or superseding-branch required.
- **Phase C** owner/ops-only; flags OFF throughout.

**Live-status doc**: current DAG (L3/L4 on main; L4cov gate-CLEAR + HELD per owner; L5/L6-a/#4380/#4385 open), model tiers, guardrails.

Docs-only, zero runtime. Not auto-advancing Phase B. Per your execution order: this is step 2 (record the amendment); #4362 is fixed + green + gate-CLEAR but HELD (step 1, not merged).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4438, opened 2026-07-17
W0 L4-cov-services — service writers join the canonical fence + D-1 delete into H1-DELETE matrix (DRAFT, default-OFF)

zensgit:
## W0-1 Lane L4-cov-services (DRAFT — default-OFF; held for owner review, not auto-merge-armed)

Extends the **L4/L4cov canonical all-writer fence** to the remaining SERVICE writer classes, and executes the owner directive (2026-07-16): **fold the deferred plugin D-1 delete branch into H1-DELETE's implementation matrix** so it can never again sit in "recorded gap" state. Independent of the L6-b PR (#4417) — builds directly on merged main.

### Writers fenced (all behind default-OFF `MULTITABLE_ENABLE_WRITER_FENCE`; flag-off byte-identical)
| Writer class | Seam | Blocked-writer behavior |
|---|---|---|
| Automation actions (update/create/delete/lock/unlock) | `AutomationExecutor.withTransaction(sheetId, …)` — fence-first in the txn | step `failed` honestly (surfaces in execution log; retry semantics apply) |
| Automation result-writeback | `AutomationService.withTransaction(sheetId, …)` | `SheetWriterBlockedError` to the bridge caller |
| Formula materialization | fence + block-check before the derived-value UPDATE | returns `null`, zero writes (a pre-recovery-input materialization can no longer clobber recovered data) |
| Approval-projection reconcile | fence before the per-instance projection lock (lock order documented deadlock-free) | throws + ROLLBACK, zero projection rows |

**Fail-closed** (H3 precedent): executor with the flag ON but no transaction seam **throws** — never a silent unfenced autocommit write.

### D-1 → H1-DELETE (owner directive)
With the flag ON, the plugin D-1 delete branch now enforces the OD-7 transactional-query contract, takes the fence, and **mints + seals its operation** — its sealed endpoint is **H1-immutable (UPDATE and DELETE both reject)**, same matrix as every other delete sink. Flag OFF: byte-identical D-1 (NULL `operation_id`, no endpoint). Also pinned: **G1** — the retention-prune GUC unlock is **transaction-local** (after a prune, a plain endpoint DELETE on the same connection still rejects).

### Verify (13 real-DB goldens; every wiring mutation-proven red→restore→green)
M1 executor fence ⇒ A1–A4 red · M2 fail-closed ⇒ A5 red · M3 formula fence ⇒ F1 red · M4 projection fence ⇒ P1 red · M5 writeback fence ⇒ W1 red · M6 D-1 ledger tag ⇒ D1+G1 red. Positive controls + flag-off parity per writer. Two-point wiring (plugin-tests.yml allowlist + vitest glob). Regression: fence/ledger/automation cluster **172 pass / 1 pre-existing skip**, unit **5546/5546**, tsc clean, OD-6 + rank-8 structural guards green.

Known residual (documented in-code, for the gate to weigh): formula fence is fully txn-scoped on the fenced PATCH hot path; a bare (autocommit) background recompute caller degrades to park+block-check with a small TOCTOU — full txn-wrapping of those callers is follow-up if required.

An independent multi-lens adversarial gate runs against this head before it leaves DRAFT.


zensgit:
## Independent 3-lens adversarial gate — verdict + fixes landed (head `dc6849b20`)

**Gate on `1594618ad`** (lenses: fence-wiring-completeness / D-1→H1 fold + ledger discipline / flag-off parity + residual; each on its own isolated DB, refute-first, with independent mutations):

- **Zero P1.** Upheld with real reds: all 8 service `meta_records` writes fence-covered with the correct target sheet (incl. a global writer-inventory sweep at synthesis — no unfenced writer anywhere); executor fail-closed load-bearing; mint/seal ordering + deferred-FK discipline correct; H1 endpoint immutability; GUC containment transaction-local; **flag-off parity byte-identical across all 6 files** (incl. DB-trigger layer).
- **BLOCK on two confirmed P2 coverage gaps** — the D-1 branch's two new guards were correct but mutation-removable with every golden green (claim-vs-delivery mismatch on the block-refusal matrix leg).

**Both fixed in `dc6849b20`, exactly per the gate's acceptance criteria:**
1. **D3** block-refusal golden — `applying` block + flag ON ⇒ D-1 plugin delete refuses, record survives, zero revision/endpoint rows. *MUT-B (neuter the D-1 fence) now reds D3* ✅
2. **D4** non-txn negative golden (G16 analogue) — bare autocommit caller + flag ON ⇒ refused before any destructive write. *MUT-C (neuter the assert) now reds D4* ✅
3. P3 doc fix: formula-engine comment now states the audited caller reality (bare-pool recompute ⇒ durable-block refusal holds — F1 proves it on that exact shape — with a narrow residual; no chain-seq/PIT impact; hardening enumerated as follow-up).

Golden file now **15/15**; tsc clean. Deferred (non-blocking, gate-enumerated): formula bulk-recompute txn-wrapping/fail-closed; constructed writer-vs-recovery race once the block producer (revert-execute L8) exists; multi-sheet fence-ordering probe.

**Held as Draft for the owner** alongside #4417 — not self-merged, nothing armed, `MULTITABLE_ENABLE_WRITER_FENCE` stays default-OFF.

zensgit:
## P3-R golden landed (`3a5596bfd`) — the owner review's Low finding, test half

The doc half of both review findings was already applied on #4406 by its owning session (`88cc1ba14`: the H2 two-direction split + the P3-R entry). This lands the missing **test itself**, here because this branch owns the §G retention goldens:

**G2** — one explicit caller transaction: `meta_record_history_operations_prune(op0)` succeeds → `current_setting('metasheet.mrho_retention')` already reads non-`'on'` **inside the same txn** → an ad-hoc endpoint DELETE of another op **later in the same transaction still RAISES**. Distinct from G1 (which proved the separate-statement case — that one passes even without the trailing reset).

**Mutation-proven:** `CREATE OR REPLACE` of the prune fn **without** the trailing `set_config('metasheet.mrho_retention','off',true)` ⇒ **exactly G2 red** (15/16); restore ⇒ 16/16 green. The reset is genuinely load-bearing — without it, H1 is silently bypassed for the remainder of any caller transaction that invokes prune.

Per P3-R's framing this golden is the pre-condition gating the **first production retention caller**; it now exists and runs in the CI real-DB allowlist with the rest of this file.

zensgit:
**Independent gate verdict: BLOCK** — gated on `462935f06`; carries to current head `05f5ceca0` (only #4443's automation files differ between the two heads; all P1-relevant files byte-identical).

**P1 (the block):** the formula/derived writer path implements **neither** of the owner's two options. `fenceWriterEntry` still runs on a bare autocommit query — `pg_advisory_xact_lock` is statement-scoped there, so only the durable-block *refusal* fires; no fence is held across compute+UPDATE (production callers pass plain pool queries: `record-service.ts` hook, `univer-meta.ts:2953/:3098`). And no recovery-projection exclusion for derived fields exists anywhere — recovery restores the full `data` blob including formula keys (resurrect at `univer-meta.ts:10345`, revert via `patchRecords`). The "honest residual" comment is a third option the owner did not authorize; its "recompute-on-read" premise is also inaccurate — the read path serves materialized formula values raw from `data` (`univer-meta.ts:3188`; recompute is a post-write hook), so a raced recovery can serve a stale derived value until the next write-triggered recompute.

**Fix to clear (either):**
- **(a)** thread a txn client into the two recompute callers and fence in-txn — the exact fail-closed pattern already used by this PR's `automation-executor` changes; or
- **(b)** add a real derived-field exclusion to the recovery projection (code, not comment).

(If the owner explicitly elects to ratify the documented-residual option with an enumerated follow-up, that acceptance is theirs to make — the gate can only pass what was authorized.)

**Everything else verified solid (mutation-proven):** in-txn service-writer fences (A1 red when neutered) · D-1 → H1-DELETE matrix (D3 red) · **G2 P3-R prune GUC-reset load-bearing (G2 red — closes the #4406 P3-R precondition)** · H2 direction-asymmetric comments accurate · scope clean, flags default-OFF · 130 adjacent real-DB tests green.

P1×1 / P2×0 / P3×1 (stale-base vs #4443 — `git merge-tree` clean, rebase + rerun combined gate) / NIT×1. Full report: `/tmp/pr4438-gate-462935f06-claude.md`

zensgit:
**Independent re-gate verdict: CLEAR-WITH-NITS** — exact head `5bb12ae3d` (the v2 engine-internal fenced-txn fix). **P1: 0 / P2: 0 / P3: 1 / NIT: 1.**

**The original formula-TOCTOU P1 is CLOSED.** With the flag ON, `recalculateRecordFromData` runs `fenceWriterEntry` + the materialization UPDATE inside ONE `poolManager.transaction` — advisory lock held acquisition→COMMIT, no autocommit gap; an active durable recovery block throws → rollback. **Mutation-proven:** removing the in-txn fence reds both F1 and the new F2 (which behaviorally proves the recompute parks on the fence via `pg_locks` not-granted). GF8's partial-progress contract (which the v1 caller-wrapping broke) is restored — `multitable-w11-expression-bulk-recompute-realdb` 6/6 green. Flag-OFF path verified byte-identical to pre-L4cov (direct UPDATE on the passed query). All other files byte-identical to `05f5ceca0`, so the earlier mutation-proven verdicts (in-txn service fences A1, D-1 matrix D3, G2 P3-R, H2 docs) carry. Scope clean; `MULTITABLE_ENABLE_WRITER_FENCE` still default-OFF.

**P3 (owner-visible, non-blocking — a pre-flag-flip decision item):** the compute READS stay on the caller's ordinary query *outside* the fenced txn. Sequence: recompute reads inputs at T0 → a recovery completes at T1 → the fenced txn at T2 sees no active block → writes a derived value computed from pre-recovery inputs onto the recovered row. Bounded: derived-only (`data || $updates` preserves user fields), self-healing on next recompute, no PIT/seq/version impact, flag default-OFF, and strictly better than the pre-L4cov baseline (no fence at all). If you want it gone before any flag-flip: move read+compute inside the fenced txn, or add an updated_at/version guard to the UPDATE's WHERE.

**NIT:** with the flag ON the engine ignores a passed txn query (uses its own poolManager txn) — safe with current callers (verified none wrap recompute in a txn), but worth a contract comment, folding the P3 staleness note in.

Full report: `/tmp/pr4438-regate-5bb12ae3d-claude.md` (v1 history: `ddb28aa9f` was BLOCK — 2 missed bare-pool callers — mooted by the v2 pivot which covers all callers by construction).

zensgit:
## Exact-head re-gate CLEAR (head `5bb12ae3d`, run in-loop after 3 subagent-infra deaths)

The formula TOCTOU **v2** (engine-internal flag-gated fenced txn) is verified against the regression that v1 caused:
- **GF8 / partial-progress restored**: `multitable-w11-expression-bulk-recompute-realdb` + the formula/write/l4cov cluster = **44/44** on an isolated gate DB. The mid-chunk failure injection (pool.query monkeypatch) fires again and rec_a-persisted/rec_b,c-stale/failed:true holds.
- **flag-OFF byte-identity confirmed by source**: the `else` branch is a direct `await query(UPDATE)` — no `poolManager.transaction`, mock queries honored (the two mock-query unit files pass).
- **fence still load-bearing**: dropping `fenceWriterEntry` reds BOTH F1 and F2; restored 17/17.
- **no txn-nesting**: all three reverted call sites pass a bare pool query; no caller runs the recompute inside a fence-holding transaction (a documented forward constraint for a future flag-ON caller).

**Verdict: CLEAR.** Ready to rebase + full CI + merge per the owner sequence (step 2).

zensgit:
## Post-merge review absorbed → follow-up #4455 (head-scoped record)

An independent post-merge review of the v2 head (`5bb12ae3d`, rebased to the merged `d3e5e7b1c`) returned REQUEST-CHANGES **after** this PR merged (merge 14:12Z preceded verdict delivery). Disposition:

- **High (confirmed on main)**: v2 fenced only the pure-formula engine write; the two relation-aggregation materializations (`recalculateFormulaFields` same-record; `computeDependentLookupRollupRecords` fan-out) remained bare autocommit UPDATEs — same clobber-hazard class as round 1. **Fixed in #4455** via a shared flag-gated derived-write seam + block-refusal goldens (C1/C2/C3) + mutation matrix; the §C DEFERRED skip in this PR's test file was stale on two counts (engine IS fenced since v2; "recompute failure is logged and tolerated" was false for the patch spine — an uncaught refusal would have 500'd a committed PATCH) and is replaced by real goldens there.
- **Low-1 (F2 test)**: DB-wide `pg_locks NOT granted` waiter count → precise `pg_blocking_pids` attribution + failure-path ROLLBACK, in #4455.
- **Low-2 (this PR's body, corrections for the record)**: ① "Verify (13 real-DB goldens)" — the merged file has **17** test() (D2/D3/D4/F2/G2 post-date the count). ② The "Known residual: bare autocommit background recompute … small TOCTOU" paragraph is **superseded by v2** (engine-internal fenced txn; no autocommit-degraded path remains). ③ The formula row's "fence + block-check before the UPDATE" understates v2 — the fence is held in the SAME txn THROUGH the UPDATE. ④ The mutation matrix omits F2/D2-D4/G2 added after it was written.

Body left as merged (point-in-time record); this comment is the correction, per the #3830 precedent.

PR #4442, opened 2026-07-17
W0 L7 — exact-anchor recovery PLAN: target-set classification for execute (DRAFT, stacked on #4417)

zensgit:
## W0-1 v3.7 Lane L7 — exact-anchor recovery PLAN (DRAFT, stacked on #4417; held, unwired, flags OFF)

**Stacked-PR notice:** base = `claude/w0-l6b-exact-anchor-20260716` (#4417, itself held for the owner's flip decision). This PR consumes L6-b's `reconstructRecordsAtSeq` and lands the **§5 target-set recomputation** layer — the classification a destructive apply consumes at L8. Not wired to any route (default-off by construction); nothing destructive.

### Scope note (honest)
Target-**generation** resolution itself was already delivered by L6-b's causal reconstructor (the MULTI-GEN golden). What this lane adds is the remaining §5 semantics: the **plan classification** with deleted/trash enumeration end-to-end.

### The classification (per record, target = as-of-`anchorSeq` state)
| target | live now | plan |
|---|---|---|
| exists (differs) | yes | **revert** candidate — at-anchor data + target/live versions |
| exists | no | **resurrect** candidate — at-anchor snapshot (revision chain, never a trash vintage) |
| deleted (≤ anchor) | yes | `deletedAtAnchorLiveNow` — **caller picks** (Revert keeps / Reset deletes; the plan never chooses destruction) |
| deleted | no | stays deleted (LOCK-9 — never resurrected) |
| no revision ≤ anchor | yes | `createdAfterAnchor` — caller picks |

Schema-drift snapshots (stale field id) **excluded + counted** on both branches → re-preview, never half-applied.

### Layering contract (loud, in the module doc)
Core set-classification ONLY. Permission masking (PIT-3), row-level deny, size ceilings, and **fence acquisition are the L8 wiring's obligations** — consuming this module without those layers is not a production path.

### Verify (10 real-DB goldens, first-run green; 4 mutations, each load-bearing)
[...]

PR #4444, opened 2026-07-17
W0 L8 — exact-anchor destructive apply: one txn, all-or-nothing (DRAFT, stacked)

zensgit:
## W0-1 v3.7 Lane L8 — the exact-anchor destructive apply (DRAFT, stacked on #4442→#4417; held, unwired, flags OFF)

**Stacked-PR notice:** base = `claude/w0-l7-execute-plan-20260717` (#4442), which stacks on #4417 (held for the owner's flip decision). Top of the held W0 stack. **Not wired to any route** — the legacy Revert/Reset switch onto this module is the owner's wiring decision behind their existing default-OFF flags. Nothing reaches this module today.

### One transaction, all-or-nothing
`applyExactAnchorRecovery(transaction, {token, sheetId, actorId, mode: revert|reset})` — fence-first → burn → in-fence rechecks → plan → apply → seal, **COMMIT once**; any failure ⇒ **full rollback with zero writes, including the burn** (a refused execute leaves the world byte-identical).

### Lands the L6-b gate's three pinned L8 deferrals
1. **In-fence checkpoint re-resolution** — re-runs `selectCheckpointByAnchorSeq`, requires equality with the token's `checkpointId` (echo never trusted): `no-covering-checkpoint` / `checkpoint-changed`.
2. **Baseline composition** — the resolved checkpoint's `meta_history_baselines` overlay records below the replay horizon (retention-pruned/pre-history); trashed baseline = deleted-at-anchor.
3. **Anti-replay** — token sha256 burned into new `meta_recovery_token_burns` (new zzzz migration) inside the txn; the PK is the at-most-once barrier.

### The drift contract, honestly delivered
`scopeHash` binds the **immutable** at-anchor reconstruction — it can *never* see a concurrent live write. The identity now also binds **`liveSetHash`** (order-invariant HMAC over live `{id, version}` at preview); the apply re-fingerprints the live set **in-fence** and refuses `preview-drift` (409-class) when anything moved — the actor applies exactly the world they previewed. Also: rank-8 **lock-guarded** (a locked record aborts the whole apply), every write revision-emitted (`source:'restore'`) + ledger-tagged, and the apply **seals its own operation endpoint** — a recovery is itself a future exact anchor.

### Verify — 11 real-DB goldens, 6-mutation matrix all load-bearing
X1 burn no-op ⇒ REPLAY+LIVE-DRIFT red · X2 trust the checkpoint echo ⇒ CHECKPOINT-CHANGED red · X3 drop freshness ⇒ LIVE-DRIFT red · X4 reset skips deletes ⇒ MODE-MATRIX red · X5 drop baseline composition ⇒ BASELINE red · X6 swallow a mid-apply failure ⇒ INJECTED-FAILURE red.
Highlights: **LIVE-DRIFT** (a concurrent committed write between preview and execute ⇒ 409 + zero writes incl. the burn; the same token succeeds once the drift is undone — proof the refusal wrote *nothing*); **INJECTED-FAILURE** (a mid-apply crash rolls back the already-executed revert too); **FENCE-PARK** (constructed race: a raw client holds the canonical fence; the apply genuinely parks — `pg_locks`-proven — then completes); **LOCKED** (another actor's lock aborts all-or-nothing; unlock lets the same token through). Structural guards (OD-6 / rank-8 / rich-longtext sink) green with real dispositions. Unit **5556/5556**, W0 cluster **122/122**, tsc clean, fresh migrate clean. Two-point CI wiring.

An independent 4-lens adversarial gate (atomicity / drift-409 / in-fence rechecks + anti-replay / mode-matrix + regression) runs against this head before it leaves DRAFT.

### DIFF
--- a/scripts/ops/staging-attendance-report-digest-rd45-smoke.mjs
+++ b/scripts/ops/staging-attendance-report-digest-rd45-smoke.mjs
@@ -166,13 +166,18 @@ export function classifySendOutcome(rows) {
     failedRecognized: 0,
     failedOther: [],
     unknown: [],
+    // Direct failure evidence for NON-terminal rows (pending/sending/retrying). The prior helper only
+    // captured last_error for terminal failedOther rows, so a run that timed out in `retrying` — and then
+    // deleted the row in cleanup — left NO record of WHY (which retryable last_error, how many attempts,
+    // when the next backoff was due). Populated for exactly the states that can leave the send-proof stuck.
+    nonTerminal: [],
   }
   for (const row of rows) {
     const status = String(row.status ?? '')
     if (status === 'sent') outcome.sent += 1
-    else if (status === 'pending') outcome.pending += 1
-    else if (status === 'sending') outcome.sending += 1
-    else if (status === 'retrying') outcome.retrying += 1
+    else if (status === 'pending') { outcome.pending += 1; outcome.nonTerminal.push(describeStuckDelivery(row)) }
+    else if (status === 'sending') { outcome.sending += 1; outcome.nonTerminal.push(describeStuckDelivery(row)) }
+    else if (status === 'retrying') { outcome.retrying += 1; outcome.nonTerminal.push(describeStuckDelivery(row)) }
     else if (status === 'failed') {
       if (RECOGNIZED_TERMINAL_FAILURES.includes(String(row.last_error ?? ''))) outcome.failedRecognized += 1
       else outcome.failedOther.push({ id: row.id ?? null, lastError: row.last_error ?? null })
@@ -184,6 +189,18 @@ export function classifySendOutcome(rows) {
     && outcome.failedOther.length === 0
     && outcome.unknown.length === 0
   return outcome
+}
+
+// The last-hammer evidence a stuck send-proof needs: which retryable error, how many attempts so far, and
+// when the worker's exponential backoff (60s→5m→15m→60m→6h) next fires. Pure/deterministic for the test.
+export function describeStuckDelivery(row) {
+  return {
+    id: row.id ?? null,
+    status: String(row.status ?? ''),
+    lastError: row.last_error ?? null,
+    attemptCount: row.attempt_count == null ? null : Number(row.attempt_count),
+    nextAttemptAt: row.next_attempt_at == null ? null : String(row.next_attempt_at),
+  }
 }
 
 // Env-only contract (no CLI args), mirroring the AE-4 / OT-bank v1-8 helpers: BASE_URL/DATABASE_URL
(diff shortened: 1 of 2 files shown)


zensgit:
## Independent 4-lens adversarial gate — OVERALL CLEAR (gated head `899849054`; fixes → `11639712d`)

Lenses (each on its own isolated DB, refute-first, independent mutations): **atomicity** / **drift + anti-replay** / **in-fence rechecks + baseline composition** / **plan classification + mode matrix + regression**. All four returned **CLEAR**; synthesis confirmed **zero P1, zero unrebutted P2** and independently closed the lens gaps (CI allowlist verified at head; zzzz migration hygiene; actorId/sheetId binding at verify).

**Load-bearing positives, proven by construction:** at-most-once burn under a two-client PK race (park → 23505 → zero writes); the burn rolls back on *every* refusal class; seal-leg rollback (non-deferrable BEFORE-INSERT trigger); the preview-drift throw is load-bearing (mutating it to a return **leaks the burn** and reds LIVE-DRIFT); checkpoint echo never trusted (X2 red); baseline composition load-bearing (X5 red); FENCE-PARK pg_locks-proven; plan buckets disjoint (an exists-at-anchor row can never enter a delete set); no L4/L5/L6-b regression.

**Fixed in `11639712d`:**
- **F6** (4/4-lens consensus, test-only): apply-golden anchors lifted above shared-DB sequence drift; proven robust by re-running green with the chain sequence deliberately advanced to 5000 (the exact condition that redded all four lens DBs).
- **F2** (doc-vs-code): plan.ts's "never silently half-applied" claim was false — the apply proceeds without drifted records (`driftExcluded=N`, drifted rows untouched, never deleted). Rewritten to the honest partial-set semantics; refuse-vs-confirm on `driftCount>0` is a pinned route-wiring decision.

**Pre-wiring gate list** (owner clears before any route/flag flip; none block this unwired default-OFF draft): F1 bind `mode`/`strategy` into the identity (T8-2 discipline) · F2 drift-refusal route semantics · F3 wire the strict chain-integrity precheck into preview+apply (§5 step 5) · F4 compose baselines into the displayed preview · G1 burn-table retention sweep · G2 resurrect-vs-trash-row golden · G3 two-token concurrent-apply constructed race · G4 route-level size ceilings.

**Held as Draft for the owner** with the rest of the stack (#4417 → #4442 → #4444). Nothing armed; flags OFF.

zensgit:
P1 (owner review): same wrong-head accident as #4442 (created from the canonical checkout on the attendance branch). The real L8 work lives on `claude/w0-l8-exact-anchor-apply-20260717` @ `11639712d` and is re-opened as a fresh Draft with an explicit --head.

PR #4450, opened 2026-07-17
docs(automation): Class-B fan-out per-target intent-key design lock (PROPOSED)

zensgit:
## What
The design lock (PROPOSED) for wiring the two Class-B **fan-out** actions — `send_dingtalk_person_message` / `send_dingtalk_group_message` — onto the two-phase outbound substrate. #4443 correctly deferred them: one intent per `(kind, root, action_key)` cannot represent N per-recipient outcomes without a wrong coarsening (block retry of definitely-failed targets, or double-send delivered ones).

## Design (summary — full doc in the diff)
- Per-target intent key `(kind, root_execution_id, action_key, target_key)`; `target_key` = sha256 of a JSON-array injective encoding of the **resolved** DingTalk-side identity (no PII, order-independent). `''` sentinel = whole-action intent → the five landed single-send wirings are unchanged; one substrate, no second table.
- Additive migration: `target_key` column + widened UNIQUE (strictly weaker for existing `''` rows; no backfill).
- Per-target two-phase semantics: retry re-attempts ONLY definite-`failed` targets; `sent`/`outcome_unknown` skip per-target; partial-crash recovery per-target by construction; values-free aggregate step result (per-class counts only).
- V7/V8/V9 posture unchanged (sequential loop, no new lease surface).
- G-F1..G-F5 RED-before verification obligations enumerated.

## Status
**PROPOSED, zero runtime code, docs-only.** Awaiting owner ratify; implementation is a separate flag-gated slice after ratification.

### DIFF
--- a/docs/development/approval-automation-classb-fanout-per-target-intent-designlock-20260717.md
+++ b/docs/development/approval-automation-classb-fanout-per-target-intent-designlock-20260717.md
@@ -1,0 +1,59 @@
+# Class-B fan-out actions — per-target outbound intent key (design lock, PROPOSED)
+
+**Status: PROPOSED 2026-07-17 (zero runtime code; awaiting owner ratify).** This document has NO ratify authority of its own. It extends the RATIFIED classification lock
+`approval-automation-retry-action-classification-designlock-20260712.md` (§3 two-phase intent/outcome) to the two Class-B actions that FAN OUT to N recipients. It changes nothing for the already-landed single-send actions.
+
+## 1. Problem (found while wiring #4443)
+
+`send_dingtalk_person_message` and `send_dingtalk_group_message` resolve their recipient config to **N independent per-recipient sends**. The landed two-phase substrate (`meta_automation_outbound_intent`, #4441) keys **one** intent per `(kind, root_execution_id, action_key)` — one at-most-once side effect per action identity. A fan-out cannot be represented by one intent without picking a wrong coarsening:
+
+- **Coarsen to `outcome_unknown`** when any target is ambiguous → permanently blocks retry of targets that DEFINITELY failed pre-dispatch (lost sends, unrecoverable by automation).
+- **Coarsen to `failed`** when any target failed → a retry re-runs the WHOLE action and re-sends to targets that were already delivered (duplicate messages — the exact effect `outcome_unknown` exists to forbid).
+
+Both were rejected in #4443; the two actions were left un-wired (their existing per-target `outcome_unknown` telemetry ledgers remain the only guard). This lock defines the correct wiring.
+
+## 2. Design: extend the intent identity with a per-target segment
+
+### 2.1 Identity
+Per-target intent key = `(kind, root_execution_id, action_key, target_key)` where:
+
+- `action_key` — unchanged: the §2.1 triple over `{structuralPath, actionType, canonicalConfig}` (the whole action's identity, config included — so a recipient-list EDIT changes `action_key`, which is correct: it is a different action).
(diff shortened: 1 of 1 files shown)


zensgit:
## Adversarial design-lock review (Claude, ran against `34e82e86e`) — NEEDS-CHANGES before ratify

Docs-only verified (1 file, +59, zero runtime). The coarsening-dilemma diagnosis, the sentinel-`''` back-compat design, and the injective JSON-array key encoding are all sound. Three P2 ratify-blockers in the per-target identity scheme:

- **[P2-1] person `target_key` is non-injective across corps → silent recipient drop.** Line 22 keys on `['dingtalk_user', dingtalkUserId]`, omitting the corp dimension. DingTalk userids are corp-scoped (executor's own DT-OPS-04 note, `automation-executor.ts:3649`; recipients carry `integrationId`; multi-corp audiences are supported). Alice@corp1 `1001` and Bob@corp2 `1001` collapse to one intent key → the second `claimOutboundIntent` returns `skip_sent` → one recipient silently never sent. Fix: `['dingtalk_user', integrationId, dingtalkUserId]`.
- **[P2-2] group `target_key` names a field that does not exist.** Line 22 keys on `openConversationId` — zero repo hits. The as-built group action delivers via `dingtalk_group_destinations.webhook_url` keyed by internal `destinationId` (`executeSendDingTalkGroupMessage`). Unimplementable as written. Fix: `['dingtalk_group_destination', destinationId]`.
- **[P2-3] person runtime model contradicts the merged batched send (design fork — owner call).** §2.3/line 41 prescribe a per-target send loop, but the real person send groups by integration and chunks into batches of 100, each batch ONE `asyncsend_v2` call with ONE `task_id` — no per-recipient outcome exists. Either (a) flag-ON unbatches (a behavior change line 41 currently denies), or (b) the lock must specify a batch↔target mapping (batch outcome stamped onto every target in that batch) — the word "batch" currently doesn't appear. G-F1's "crash after 3 → sends exactly 4" presumes (a).

P3: (1) widening `uq_outbound_intent_identity` forces `claimOutboundIntent`'s ON CONFLICT/WHERE to gain `target_key` in lockstep (else landed single-send Class-B claims break, 42P10) — the migration coupling should be stated; (2) per-target row amplification has no demonstrated total-N cap.

Full report: `/tmp/pr4450-lock-review-claude-20260717.md` (local).

zensgit:
Rev 2 pushed (c3f7cb827): all 3 P2 + 2 P3 from the review above incorporated — person key corp-scoped (integrationId in the tuple, G-F4b injectivity golden), group key on the real destinationId, §2.3a rewritten to per-target rows + per-BATCH claim/outcome over the unchanged batched asyncsend_v2 (unbatching rejected as a flag-ON behavior change; crash windows mapped to §8 fail-closed doctrine), §2.2 migration↔primitive lockstep obligation, explicit MAX_CLASSB_FANOUT_TARGETS cap + G-F6. Ready for ratify decision.

zensgit:
Rebased onto main 60e9ecbec, re-verified (docs-only: 1 file, git diff --check clean). Awaiting review.

zensgit:
Rev 3 pushed (de5ce9e96): §2.3a batch-claim ownership per your ratify-review P2 — Tx-A atomically RETURNS `claimed_target_keys` (per-target ON CONFLICT … RETURNING + retry CAS flips, one txn); the batched asyncsend_v2 may be built from that returned set ONLY (lost-race candidates dropped, empty set ⇒ no send); Tx-B stamps outcomes keyed by the same set; pre-read explicitly demoted to optimization-never-authorization. Rebased onto current main. Ready for ratify.

PR #4447, opened 2026-07-17
W0 L5-wire — trust-checkpoint activation route (production caller for activateCheckpoint)

zensgit:
## W0-1 L5-wire — trust-checkpoint activation route (owner review P2; canonical order L5-wire → L6-b → L7 → L8)

`activateCheckpoint` had **no production caller** — L6-b could only ever refuse `no-covering-checkpoint`. This is the missing runtime slice: `POST /sheets/:sheetId/trust-checkpoint-activate`.

- **Default-OFF** `MULTITABLE_ENABLE_TRUST_CHECKPOINT_ACTIVATION` (manifest-registered, danger:medium — additive-only trust provisioning; writes only checkpoint/baseline rows, never live data, but moves the trust floor recoveries anchor on). **D2 floor** (`canManageSheetAccess`).
- **One fenced transaction** (design lock §3): `fenceWriterEntry` first → cutover (allocate `trusted_since_seq` → baselines → supersede prior active → activate). Unattributable trash ⇒ **409 HISTORY_INCOMPLETE values-free + full rollback**; durable block ⇒ 409 RECOVERY_IN_PROGRESS.
- Activation **enables nothing by itself** — strict/Revert/Reset stay behind their own OFF flags.

### Verify
8 real-DB goldens; mutations each exactly-one-red: W1 flag gate ⇒ FLAG-OFF red · W2 D2 floor ⇒ NON-ADMIN red · W3 fence call ⇒ FENCE-PARK red (constructed race, `pg_locks`-proven parking). HAPPY proves baselines + supersede-to-exactly-one-active; UNATTRIBUTABLE proves the fail-closed abort rolls back everything. Manifest contract 37/37; regression 42/42; unit **5603/5603**; tsc clean; two-point CI wiring.

Independent gate runs against this head before it leaves Draft.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4451, opened 2026-07-17
fix(multitable): GF8-ON — propagate non-block failures from the engine's fenced materialization txn + flag-ON golden

zensgit:
## What

Owner GF8 review follow-up (2026-07-17, in-session): under `MULTITABLE_ENABLE_WRITER_FENCE` the v2 engine ran fence+UPDATE in its own short transaction but **swallowed every in-txn failure to `null`** — the B6 bulk-recompute executor then mislabeled an infra failure as `taintSkipped` with `failed:false` (a silent lie in the recompute report). GF8's `pool.query` monkeypatch can never reach that surface: the engine txn runs on its own client.

## Fix (engine, flag-ON path only)

Failure-class split in `recalculateRecordFromData`:
- `SheetWriterBlockedError` (durable recovery block) ⇒ `null`, benign skip — **F1 contract unchanged**; a user write never fails on a derived value.
- any **other** fenced-txn failure ⇒ typed `FormulaFencedWriteError`, propagated (through `recalculateRecord` too) so the B6 chunk loop aborts with `failed:true` — while the poisoned record itself stays **zero-committed** (its txn rolled back).

**Flag OFF: byte-identical** (the direct-UPDATE branch is untouched; GF8's existing injection + assertions unchanged).

## GF8-ON golden (real DB)

WHEN-scoped `BEFORE UPDATE` trigger injects the failure **inside** the fenced txn (per-record-id scoped — shared-DB discipline; dropped in `finally`):
- `failed:true` + `BULK_RECOMPUTE_FAILED`
- `rec_a` (before the poison) **retained** — per-record txn independence, the original GF8 partial-progress contract (v1's per-chunk rollback was abandoned for exactly this)
- `rec_b` **zero-committed** (stale value byte-preserved)
- `rec_c` **not attempted** (abort remaining)
- healthy flag-ON re-save converges 3/3 (anti-vacuous positive control)

## Verify

- w11 suite 7/7; w11 + l4cov-services fence suites together **24/24** (F1 null-on-block still green) on fresh-migrated PG
- **Mutation-proven**: revert the propagation (swallow to null, pre-fix behavior) ⇒ exactly GF8-ON red at the `failed:true` assertion, 6/7 others green; restore ⇒ 7/7
- tsc clean

## Owner-reviewable point

Flag-ON REST/PATCH single-record materialization now also propagates infra failures (previously silent-stale). Recovery-block skips are unchanged. Flags remain OFF everywhere — zero live-path impact until the fence flag is enabled.

### DIFF
--- a/packages/core-backend/src/multitable/formula-engine.ts
+++ b/packages/core-backend/src/multitable/formula-engine.ts
@@ -7,7 +7,7 @@
 import { FormulaEngine, type CellValue, type FormulaContext } from '../formula/engine'
 import type { MultitableField } from './field-codecs'
 import type { MultitableRecordsQueryFn } from './records'
-import { fenceWriterEntry, isWriterFenceEnabled } from './canonical-sheet-fence'
+import { fenceWriterEntry, isWriterFenceEnabled, SheetWriterBlockedError } from './canonical-sheet-fence'
 import { poolManager } from '../integration/db/connection-pool'
 import { Logger } from '../core/logger'
 
@@ -272,6 +272,8 @@ export class MultitableFormulaEngine {
 
       return await this.recalculateRecordFromData(query, sheetId, recordId, data, fields, onlyFormulaFieldIds)
     } catch (error) {
+      // The typed fenced-write failure must cross this wrapper too (bulk recompute enters here).
+      if (error instanceof FormulaFencedWriteError) throw error
       logger.error('recalculateRecord failed:', error as Error)
       return null
     }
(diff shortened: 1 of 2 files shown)


zensgit:
## Independent gate review — **CLEAR** (verdict on head `9741ff2ac`)

Refute-first review at pinned HEAD `9741ff2ac7fe288a84d69e50f97fe430eaffdbd1`, isolated worktree, fresh PG, one-shot suite runs + two mutations. **No P1, no P2.**

**All 5 claims PASS:**
- **Flag-OFF byte-identical** — the direct-UPDATE `else` branch is textually identical to origin/main; the two new outer-catch `instanceof FormulaFencedWriteError` guards are unreachable no-ops when the flag is off (that error is only constructed inside the flag-ON block). Zero live-path impact.
- **Non-block propagation → `failed:true`** and **`SheetWriterBlockedError` → `null`** are BOTH load-bearing: `transaction()` re-throws the original instance so the `instanceof` split is sound.
- **GF8-ON golden** is non-vacuous — runtime stack confirms the WHEN-scoped `BEFORE UPDATE` trigger fires *inside* the engine's fenced txn; processing order pinned by `ORDER BY id ASC`; positive-control re-save converges 3/3.

**Runs (fresh `metasheet_gate4451_*`, migrated):**
- w11 + l4cov-services together: **24/24** (w11 **7/7**)
- Mutation A (`throw`→`return null`): **23/24**, only **w11 GF8-ON** red at the `failed:true` assertion — exactly the claimed shape
- Mutation B (neuter the benign block early-return): **23/24**, only **l4cov-services F1** red (rejects instead of null) → the `null`-on-block branch is genuinely test-covered, not an untested guard
- Untouched neighbors (l4-canonical + l4cov-univer-meta-writers): **43 passed, 1 skipped** (anti-skip sentinels ran) → env healthy
- `tsc --noEmit`: **exit 0**

Shared-DB discipline verified: `forks` pool + `fileParallelism:false` + `maxConcurrency:1` → no cross-file env-flag bleed; poison trigger is per-`TS`-id WHEN-scoped and dropped in `finally`.

**P3 (enablement-gate note, not a blocker — flag default-OFF):** flag-ON single-record REST materialization now maps a *transient* infra failure to a 500 (was silent-stale-200); the author disclosed this. Verified bounded — the user's data write persists (separate client), formula self-heals on read/re-save, no unhandled-rejection vector (bridge stubs the helper to `[]`; all engine-entering callers are inside an Express try/catch). The enablement review should decide whether single-record paths should 500 vs swallow-and-lazy (bulk B6 already has purpose-built `failed:true`).

Full report: `/tmp/pr4451-gate-review-claude-20260717.md`

zensgit:
## ✅ Independent gate CLEAR (0 P1 / 0 P2) — armed per owner "全绿后合入" directive

Gate at head `9741ff2ac`; rebased onto current main → `751ac0021` (range-diff `=`, my commit byte-identical, verdict content-carries).

- Flag-OFF **byte-identical** vs origin/main (direct-UPDATE else branch textually identical; the two new `instanceof FormulaFencedWriteError` guards are unreachable no-ops when the flag is off).
- Both failure-class branches **mutation-covered**: `throw`→`return null` reds exactly w11 GF8-ON (`failed:true`); neuter the benign block early-return reds exactly l4cov-services F1 (`null`-on-block) — retires the "untested guard" hypothesis.
- GF8-ON golden **non-vacuous**: runtime stack confirms the WHEN-scoped trigger fires inside the engine's fenced txn; `ORDER BY id ASC` pins processing order; positive-control re-save converges 3/3.
- Real DB: w11 7/7, w11+l4cov-services 24/24; neighbors 43 passed/1 skip; tsc 0.

Gate's NIT-2 (F1 file "in no workflow") is a **false finding** — verified `multitable-l4cov-services-fence-realdb.test.ts` IS in plugin-tests.yml run-list at line 587, same executed step as the GF8-ON golden (473). No action needed.

P3-O1 (flag-ON single-record REST 500s on a *transient* infra failure, was silent-stale-200) is disclosed + gate-confirmed bounded (user write persists on a separate pool client; formula self-heals; no unhandled-rejection vector). Enablement-gate note, flag default-OFF.

Armed auto-merge (matches the #4447 handling under the same directive). Flag stays OFF; nothing else armed.

PR #4455, opened 2026-07-17
fix(multitable): W0 L4-cov follow-up — relation-agg materializations join the canonical fence (shared derived-write seam, default-OFF)

zensgit:
## What (post-merge review of #4438, owner 2026-07-17 — High + Low×2)

The #4438 v2 TOCTOU fix fenced **only** the pure-formula engine write. The review found the two **relation-aggregation materializations** in `routes/univer-meta.ts` (same-record recompute in `recalculateFormulaFields`; dependent-sheet fan-out in `computeDependentLookupRollupRecords`) still ran bare autocommit `UPDATE meta_records SET data = data || …` — no fence, no block-check, same clobber hazard class the engine fix documents. #4438 merged before the verdict landed, so this is the fix as a follow-up on main.

- **NEW `src/multitable/derived-write-fence.ts`** — the ONE shared, flag-gated fence+UPDATE seam for derived-value merges. Flag ON: fence+UPDATE in its own short txn (`pg_advisory_xact_lock` held to COMMIT). Flag OFF: **byte-identical** legacy write on the caller's query (mocks honored). Engine + both relation-agg sites route through it — the discipline can't drift apart again.
- **Refusal semantics** (review-mandated, not polish): at both sites `SheetWriterBlockedError` is an *expected operational condition* — the primary write already committed. Catch → values-free warn → skip the materialization **and** its echo (DB unchanged ⇒ nothing to invalidate) → latch the sheet; other dependent sheets + Steps 5-7 continue. Without this, a recovery block on ONE dependent sheet would 500 the whole committed PATCH post-commit. Fan-out fences the **dependent sheet being written** (loop `sheetId`, not `sourceSheetId`).
- **POST /patch 409 mapping** — the PRIMARY fence refusal (record-write-service txn entry) previously fell through to 500 INTERNAL_ERROR; now maps to the same `409 RECOVERY_IN_PROGRESS` shape as the reset/revert/config-restore routes.
- **§C goldens replace the stale DEFERRED skip** (its prose was wrong on two counts: the engine path *is* fenced since v2, and "recompute failure is logged and tolerated" was false for the patch spine): C1 same-record (block ⇒ silent skip, zero write incl. `updated_at`, no echo + positive control + flag-off parity) · C2 fan-out (block on ONE dependent sheet skips only that sheet; healthy second sheet materializes; parity) · C3 (409 mapping + zero write + parity).
- **F2 precision (review Low-1)**: waiter detection now attributes to the holder via `pg_blocking_pids` polled on the poolManager channel — the raw internal-pool client's `pg_stat_activity` view is permission-masked (reads "no waiter" forever; `pg_locks` had hidden this) — plus failure-path `ROLLBACK` before `release()` so a red assertion can no longer poison the suite with an idle-in-transaction fence holder.
- richtext write-sink guard allowlist: `formula-engine.ts` entry replaced by the new seam file (the UPDATE moved).

## Verification (local, CI-parity: fresh DB + `db:migrate` w/ CI MIGRATION_EXCLUDE)

- tsc 0 · **l4cov suites 43/43** · **12 adjacent realdb suites 116/116** (incl. GF8 bulk-recompute partial-progress contract, w11 bridge freshness, L4 canonical fence, relation-agg + crossbase + echo-mask suites) · **full unit run 485 files green** (structural guards: record-lock, revision-disposition, taint-chokepoint, bridge-wiring, raw-projection, richtext write-sink).
- **Mutation matrix** (each applied ⇒ exact reds ⇒ restored):
  | mutation | reds |
  |---|---|
  | M-A remove `fenceWriterEntry` from the seam | C1+C2+F1+F2 |
  | M-B1 fan-out passes `sourceSheetId` to the seam | C2 |
  | M-B2 fence consulted for the WRONG sheet (write key unchanged) | C1+C2+F1+F2 |
  | M-C tolerance catch removed | C1+C2 |
  | M-D 409 mapping removed | C3 |

## Honest residuals (named, unchanged by this PR)

- **reads-outside-txn** (engine-v2 posture): aggregation inputs are read before the fenced txn; a recovery that starts *and finishes* in that gap clears the block, so a stale derived value can still commit. Bounded: derived-only, self-heals on next recompute, no PIT impact (revision-exempt/no-seq).
- **No post-recovery re-trigger**: a refused materialization stays stale until the next source edit (materialize-model), incl. the cliff `#ERROR!` writes.
- **Torn per-record materialization**: with both pure formulas and relation-agg on one record, the engine txn and the seam txn are separate — a block claimed between them lands the pure values and refuses the relagg one. Each value is independently derived; named, not fixed.
- **409 mapping scope**: only POST /patch gained the mapping (golden-backed). The restore route and the AI-shortcut run route still surface a primary-fence refusal as 500 — named gap, not silently inherited.
- **Sweep finding beyond the review (follow-up candidates, NOT touched here)**: an exhaustive writer sweep found three more system writes that bypass the fence when the flag is ON — the People-sheet directory sync INSERT/UPDATE (`ensurePeopleSheetPreset`, transactional but never fenced) and the `createSeededSheet` demo-record INSERT (reachable against a pre-existing sheet via `GET /view?seed=true`). Low-risk system/demo writes, but they violate the uniform all-writer discipline; fencing them deserves its own slice with route-driven goldens rather than untested rider code here.

Flag `MULTITABLE_ENABLE_WRITER_FENCE` stays **default-OFF**; flag-off behavior is byte-identical at every touched site. Not armed — owner merge decision, per line discipline.

### DIFF
--- a/packages/core-backend/src/multitable/derived-write-fence.ts
+++ b/packages/core-backend/src/multitable/derived-write-fence.ts
@@ -1,0 +1,68 @@
+/**
+ * W0-1 L4-cov follow-up (post-merge review of #4438, 2026-07-17) — the ONE shared, flag-gated
+ * fence+UPDATE seam for DERIVED-value materializations into `meta_records.data`.
+ *
+ * Three writer sites materialize derived values with the same in-place merge UPDATE (no version
+ * bump, no revision, no seq): the pure-formula engine (`formula-engine.ts recalculateRecordFromData`)
+ * and the two relation-aggregation sites in `univer-meta.ts` (same-record materialization inside
+ * `recalculateFormulaFields`, and the dependent-sheet fan-out inside
+ * `computeDependentLookupRollupRecords`). The #4438 v2 TOCTOU fix moved the engine's write into an
+ * engine-internal fenced transaction but left the two relation-agg sites on bare autocommit
+ * writes — a derived value computed from PRE-recovery inputs could still land inside a recovery's
+ * applying window and clobber just-recovered data. This module is the single seam all three route
+ * through, so the fence discipline cannot drift apart again.
+ *
+ * Contract (identical to the engine-v2 pattern it generalizes):
+ *  - Flag ON (`MULTITABLE_ENABLE_WRITER_FENCE`): the fence-check→UPDATE PAIR runs in its OWN short
+ *    transaction so `pg_advisory_xact_lock` is held from acquisition to COMMIT. Reads stay OUTSIDE
+ *    (callers compute `updates` first) — per-record independence is preserved (each record's
+ *    materialization commits/rolls back on its own; the partial-progress the bulk-recompute
+ *    contract depends on). A durable recovery block ⇒ `fenceWriterEntry` throws
+ *    `SheetWriterBlockedError` inside the txn ⇒ rollback ⇒ ZERO write. The caller decides the
+ *    tolerance posture (the engine returns null; the relation-agg sites catch, log values-free and
+ *    skip the materialization + its echo).
+ *  - Flag OFF (default; what unit/mock callers run under): BYTE-IDENTICAL legacy write on the
+ *    CALLER-passed query — no fence, no real-pool transaction, so a mock query stays honored.
+ *  - Known residual (same as engine v2, documented at the goldens): the compute READS happen
+ *    outside the fenced txn — a recovery that starts AND finishes in the compute→write gap clears
+ *    the block, so a stale pre-recovery-derived value can still commit. Bounded: derived-only,
+ *    self-heals on the next recompute, no PIT impact (these writes are revision-exempt/no-seq).
+ */
+import { fenceWriterEntry, isWriterFenceEnabled } from './canonical-sheet-fence'
+import { poolManager } from '../integration/db/connection-pool'
+
+/** Minimal query shape shared by the callers' QueryFn / MultitableRecordsQueryFn aliases. */
+export type DerivedMergeQueryFn = (sql: string, params?: unknown[]) => Promise<{ rows: unknown[] }>
+
+/**
+ * Merge `updates` (derived-value keys only) into one record's `data`, joining the canonical
+ * per-sheet writer fence when the L4 flag is ON. Throws `SheetWriterBlockedError` (zero write)
+ * when a durable recovery block is active on `sheetId`.
+ */
+export async function applyFencedDerivedDataMerge(
+  query: DerivedMergeQueryFn,
+  sheetId: string,
+  recordId: string,
+  updates: Record<string, unknown>,
+): Promise<void> {
+  if (isWriterFenceEnabled()) {
+    await poolManager.get().transaction(async ({ query: fencedQuery }) => {
+      const fq = fencedQuery as unknown as DerivedMergeQueryFn
+      await fenceWriterEntry(fq, sheetId)
+      // lock-exempt: system derived-value materialization — no user actor (a record lock is read-only to system recompute)
+      // revision-exempt: derived materialization, no version bump — pure fn of stored inputs, recomputed on next write
+      await fq(
+        `UPDATE meta_records SET data = data || $1::jsonb, updated_at = now() WHERE id = $2 AND sheet_id = $3`,
+        [JSON.stringify(updates), recordId, sheetId],
+      )
+    })
+  } else {
+    // Flag OFF: BYTE-IDENTICAL to the pre-fence write, on the caller's query (mocks stay honored).
+    // lock-exempt: system derived-value materialization — no user actor (a record lock is read-only to system recompute)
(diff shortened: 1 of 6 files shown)


zensgit:
## Independent 3-lens adversarial gate: CLEAR-WITH-NITS ×3, 0 P1 (gated head `d2f25dedb`; hardening `0abb56acf` is test-only — runtime byte-identical to the gated head)

**Lens A (fence correctness) — CLEAR-WITH-NITS.** Flag-OFF byte-identity verified against the parent commit at all three sites (identical SQL + params on the caller query); flag-ON fence held through COMMIT (ConnectionPool.transaction BEGIN→handler→COMMIT, fence acquired then block-checked on the same client); fan-out seam keyed on the dependent-sheet loop variable = the UPDATE's own WHERE param; catches rethrow all non-block errors; 409 mapping is first in the /patch catch chain with no shadowing either direction (partialSuccess reaches it via serializePatchFailure→null→rethrow); all production callers reach the seam post-commit on pool-level queries — no open-txn caller, no deadlock. Independent re-run of M-B1/M-B2 mutations reproduced the exact claimed kills. Completeness sweep: no other derived-value merge writer bypasses the seam.

**Lens B (vacuousness + independent mutation) — CLEAR-WITH-NITS.** C1/C2/C3 assertions verified binding (C1 refusal proven three ways; C2 pins primary success + stale blocked sheet + healthy sheet + echo). Ran an independent mutation NOT in the PR's matrix (flag-OFF branch made to block-check) — predicted and observed: the flag-off parity legs red. CI two-point wiring confirmed with line numbers (plugin-tests.yml:586-587 + integration include glob). **P2**: C2's positive control flaked 2/14 whole-file runs with the exact shape of a stale block, and the §C afterEach clears were error-swallowed — **fixed in `0abb56acf`** (loud clears + per-test no-stale-block preconditions + belt-and-braces end-of-test clears; 3 consecutive CI-shape runs green after). **P3 (accepted, wording)**: the latch/`break` skip-optimizations have no discriminating coverage — they are efficiency-only (each remaining record would be individually refused with the same end state); treat the commit-message phrase as optimization, asserted by code reading.

**Lens C (contract & regression) — CLEAR-WITH-NITS.** GF8 partial-progress preserved (per-record commit; suite green); echo/invalidation trace: NO path carries a refused field id or stale value; non-block errors propagate exactly as before; richtext write-sink allowlist entries accurate post-move; primary-fence 409 cannot fire after a committed write (both derived sites absorb their own refusals). **P3 (named residual, pre-existing)**: restore + AI-shortcut routes still map a flag-ON primary-fence refusal to 500 — already listed in the PR body as a pre-flag-flip follow-up.

Suites at head: l4cov 43/43 (×3 consecutive post-hardening) · guards + engine/service units green · tsc 0. **Not armed — owner merge decision.**

zensgit:
## Head re-anchor: `35080b4ff` — rebase onto main absorbed #4451; focused re-gate CLEAR_WITH_NITS (0 P1/P2); exact-head CI all green

The originally-gated head `d2f25dedb` conflicted with main after **#4451** (`3e21f6d13`, GF8-ON failure-class split) landed in parallel — it rewrote the exact engine fenced-write block this PR moves into the shared seam. (That conflict is also why this PR initially showed ZERO CI: a CONFLICTING pull_request never gets check-suites.)

**Integration design (conflict resolution):** the seam stays pure plumbing (throws raw); #4451's failure-class split lives in the ENGINE's catch around the seam call — `SheetWriterBlockedError` → log + null (F1 contract) · flag-ON any-other → typed `FormulaFencedWriteError` (bulk executor `failed:true`, record zero-committed) · flag-OFF raw → legacy outer-catch → logged null (byte-identical). univer-meta conflict was import-adjacency only (kept #4447's checkpoint imports + the seam import).

**Focused re-gate on `35080b4ff` (head-scoped, independent):** #4451's contract verified leg-by-leg (wrapper guard + error class byte-identical to main's); hunk classification of the full gated-head→new-head delta across 14 files = every hunk main-brought (#4451/#4447/#4415) or the already-gated test hardening, ZERO unexpected — `derived-write-fence.ts` byte-identical to the gated head; independent mutation (re-introduce the pre-#4451 swallow at the throw site) → predicted == actual: exactly the GF8-ON golden red (`expected false to be true` on `bulkRecompute.failed`), restored, tree clean. Two note-only NITs (flag re-read across an await with no reachable divergence path; a log-message difference on a production-unconstructible flag-OFF path).

**Suites at `35080b4ff`:** GF8-ON 7/7 · l4cov 43/43 · engine unit 44/44 · adjacent realdb 84/84 · full unit run green (2 unrelated bundle-scheduling flakes re-verified green in isolation). **Exact-head CI: 16/16 pass** incl. `test (20.x)` (real-DB allowlist step ran the new goldens), `test (18.x)`, web-tests, migration-replay, e2e, contracts ×3.

Still not armed — owner merge decision.

PR #4458, opened 2026-07-18
feat(directory): T1 — fail-closed provider org-transfer schema and admin API

zensgit:
## T1 — Transfer MVP opening slice (sequencing plan §2 row T1)

> **UNARMED pending exact-head required CI.** Canonical Org is closed on main by #4479. This head has been independently reviewed and locally reverified after a patch-identical rebase onto that closeout.

Implements `provider_org_transfers` + `provider_org_transfer_decisions` (§7.1/§7.2) and the platform-admin lifecycle API (§6.3): create, read, scan, dry-run, apply, and cancel. Successful mutations write values-free `directory.org_transfer.*` audit rows.

### Durable invariants

- One non-null `org_id` plus composite integration FKs makes cross-org transfers schema-impossible.
- Composite provider FKs and `CHECK (provider <> 'local')` make integration/provider mismatch and local-provider transfers schema-impossible.
- A partial unique index permits at most one non-terminal transfer per source integration.
- Every transition locks the transfer row and enforces the state machine at the write point.
- Apply requires a dry-run after the latest scan and zero pending decisions.
- Lifecycle endpoints accept no body fields; smuggled fields fail before service and audit.
- Unexpected and audit-write failures log fixed metadata only.

### Adapter boundary

Production does not silently fall back to a no-op adapter. Scan and apply require an explicitly registered adapter for the transfer provider. Missing registration returns `409 ORG_TRANSFER_ADAPTER_UNAVAILABLE`, rolls back, preserves state and decisions, and writes no success audit.

The exported no-op is test-only. T1 proves the lifecycle skeleton and zero directory writes; it does **not** claim that a real transfer is implemented. T3/T4 must provide real per-decision transactional/idempotent semantics and must not treat this placeholder no-op execution shape as their final apply design.

### Verification at `4fb370dab4a4e0a2275ac5974d73032ed7d8c2bc`

- Fresh PostgreSQL: full migration chain passed; second migration replay passed with no pending work.
- T1 real-DB HTTP suite: 22/22 passed.
- T1 + Canonical Org B1–B7/pre-B4 parallel battery: 17 files / 233 tests, three consecutive runs passed.
- TypeScript `tsc --noEmit`: passed.
- B4–B7 + T1 CI wiring guards: 19/19 passed.
- Independent load-bearing mutations all reddened: scan body guard; dry-run/real-apply body guard; cancel body guard; unexpected-error log redaction; audit-error log redaction; missing-adapter fail-close; and a scoped directory write inside the test-only no-op.
- Rebase range-diff: all eight commits patch-identical to the reviewed pre-rebase series.
- Scope: eight T1-owned files only.

### Deliberate T1 boundaries

- `created_by` / `decided_by` are text because this codebase's `users.id` is text.
- `dry_run_at` is additive durable state for the scan-relative dry-run guard.
- Decision PATCH and secret handling remain deferred to the first real adapter in T3/T4.
- T2 source-freeze and the staging T2-Gate remain separately sequenced after this PR.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -204,6 +204,12 @@ jobs:
       - name: B7 round-2 CI wiring contract
         run: node --test scripts/ops/b7-round2-ci-wiring.test.mjs
 
+      # T1 CI two-point wiring contract (no DB): the provider-org-transfer suite must stay wired
+      # into BOTH vitest.config.ts (exclude) AND plugin-tests.yml (directory real-DB step).
+      # Dropping either point silently disables the transfer state-machine + FK-backstop proof.
+      - name: T1 org-transfer CI wiring contract
+        run: node --test scripts/ops/t1-org-transfer-ci-wiring.test.mjs
+
       # The production monitor invokes the same state helper covered here.
       # Keep this in the required 18/20 matrix so HTTP classification, silence
       # barriers, and alert/recovery streak semantics cannot drift silently.
@@ -685,6 +691,7 @@ jobs:
             tests/integration/org-directory-routing-policy-routes.db.test.ts \
             tests/integration/directory-binding-reconciliation.db.test.ts \
             tests/integration/directory-binding-admin-routes.db.test.ts \
+            tests/integration/provider-org-transfer-t1.api.test.ts \
             tests/integration/directory-normalized-manager.db.test.ts \
             tests/integration/attendance-csv-export-bom.test.ts \
             tests/integration/attendance-notification-redelivery.db.test.ts \
--- a/packages/core-backend/src/db/migrations/zzzz20260717120000_create_provider_org_transfers.ts
+++ b/packages/core-backend/src/db/migrations/zzzz20260717120000_create_provider_org_transfers.ts
@@ -1,0 +1,164 @@
+import type { Kysely } from 'kysely'
+import { sql } from 'kysely'
+
+/**
+ * Transfer MVP — T1 (schema), Canonical Org & Provider Transfer v1 sequencing plan §2 row T1;
+ * data model per `provider-org-transfer-development-plan-20260709.md` §7.1/§7.2.
+ *
+ * `provider_org_transfers` is the explicit transfer record (never ad-hoc JSON on integration
+ * rows, and NEVER a direct `corp_id` edit — §12.1 made corp_id immutable). Two schema-level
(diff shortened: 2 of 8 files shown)


zensgit:
**Adversarial gate — APPROVE, 0 P1/P2** (at `4dde2f899`). All 4 mutations replayed red with exact signatures (incl. live-DB partial-unique swap); reviewer's own probes: apply-vs-cancel race + seeded recovery edges 3/3, UPDATE-bypass FK/CHECK probes all fire. Report: /tmp/pr4458-t1-gate-20260717.md.
**P3 fixes applied (head now `5800e5d82`)**: recovery edges pinned by direct seed — failed→scan recovers (+ lastError cleared) and applying→cancel; both status-list mutations (drop 'failed' from SCANNABLE / 'applying' from NON_TERMINAL) now red. Suite 12/12.
**Body correction per gate**: the landing-window UNION with the B5-B7 stack covers **3 shared files** (plugin-tests.yml, vitest.config.ts, **src/index.ts** — same import anchor + mount region as #4431/#4434/#4436), not 2. T3+ forward notes recorded (decisions-PATCH must take the row lock + consider wiping dry_run_at; cancel-from-applying safe only while apply is single-txn). Unarmed — Transfer landing sequenced after Canonical Org closeout.

zensgit:
## Independent Codex re-review — APPROVE (implementation), owner-held

Reviewed exact head `6c13361cdd0699ddb2ff46ea10e7feeb0b3887b0` after the Grok rework and clean T1/T2 split. I found **0 remaining behavioral P1/P2 findings**.

What I independently verified:
- production `adapterFor` fails closed when the provider has no registered adapter (`409 ORG_TRANSFER_ADAPTER_UNAVAILABLE`);
- scan/apply rollback preserves transfer state, creates no decisions, and writes no success audit;
- the no-op adapter is reachable only through explicit test registration, not as a production default;
- T1 commit scope is exactly the three T1-owned files (no freeze API/T2 code hidden here).

Evidence rerun on this exact branch:
- real PostgreSQL T1 suite: **14/14**;
- backend typecheck: **pass**;
- T1 CI wiring guard: **2/2**;
- mutation: restoring `registry.get(provider) ?? noopOrgTransferAdapter` makes both no-adapter tests fail (and the composed T1+T2 safety test fails on the stacked head);
- GitHub checks: **settled, 0 fail**.

PR remains **OPEN / unarmed**. It is currently BEHIND hot `main`; this review is an implementation verdict, not merge authorization.

PR #4463, opened 2026-07-18
fix(stock-prep): OD2 round-1 — fail-closed category_rule version policy (server 422 + match-time guard + FE removal)

zensgit:
## What (#4194 ballot round-1: category_rule 「不能继续静默接受」; this lands the recommended ①+② pair, full implementation ③ stays a gated future design)

`category_rule` was an accepted versionPolicy vocabulary value with NO implementation branch: the matchers only implement drawing_only / drawing_and_version, and every other value fell into a tail fallback that **matches on drawing number alone whenever either version is absent** and never flags version_conflict — silently behaving like manual (or looser) while the user believes a category rule applies.

Fail-closed at four points (allowlist-shaped — any unimplemented value, not category_rule-special-cased):

1. **Candidate-sync (R5)**: unimplemented `defaultVersionPolicy` → 422 `STOCK_PREPARATION_VERSION_POLICY_UNSUPPORTED`, details `{field}` only;
2. **Mapping confirm create (R6)**: unimplemented `versionPolicy` → same 422;
3. **Engine backstop** (material-match input normalization) — defense-in-depth behind the route validators;
4. **Match-time guards in BOTH engines** (material-match AND mvp-generation's own copy; #4391 doctrine — stored legacy rows untrusted): an unimplemented policy on a stored mapping row **never auto-matches and never version-conflicts** — the line degrades to the visible missing_mapping/HELD exception path instead of silently matching.

FE: `category_rule` removed from whitelist AND type union (vue-tsc green); read-side projections deliberately still surface legacy stored values for observability. Vocabulary constant stays reserved (nothing renamed); T3a/T3b/source-run paths verified to carry no versionPolicy input.

## Evidence

- Tests: full plugin chain (~95 files) PASS; new cases across 4 plugin suites + FE spec (unimplemented×2 values × input-boundary/match-time/generation-HELD paths; positive controls for the 3 implemented values; values-free 422 details; zero-write on rejection). The match-time never-matches test **fails on unpatched code** (locks the fix).
- Mutations (commit-first): **M1** delete both engines' match-time guards → never-matches + generation tests RED verbatim (a draft line silently generated on unpatched behavior); **M2** delete both 422 rejections → route/confirm tests RED (engine backstop visibly catches R5 — defense-in-depth demonstrated); both reverted, suites green, diff clean.
- vue-tsc exit 0; touched FE specs 59/59.

**Not armed — owner review**; if the owner prefers option ③ (real category-rule implementation) later, this PR's guards are the safe substrate it would replace.

## Round-2 (adversarial review findings — fixed in `5db599796`)

- **P2 guard ordering**: the stored `matchStatus='version_conflict'` short-circuit ran BEFORE the policy guard in both engines — a legacy category_rule row with that stored status still forced VERSION_CONFLICT. Guard now runs first; such rows degrade to the missing path (tests in both engines; mutation restoring the old order goes RED verbatim).
- **P2 confirm-existing branch**: confirm-by-mappingId never inspected the stored row's policy — a legacy category_rule candidate could be stamped confirmed and become a dead confirmed row. Now 422 `STOCK_PREPARATION_VERSION_POLICY_UNSUPPORTED` `{field:'versionPolicy'}`, zero writes; policy-absent stored rows stay confirmable (absence folds to manual — asserted).
- **P3 FE read-side type**: candidate-row `versionPolicy` widened to include `'category_rule'` (read vocabulary ≠ write whitelist — confirm-reads still projects legacy values verbatim); write surfaces stay 3-valued. vue-tsc exit 0.
- **P3 status-code note**: junk versionPolicy values shifted 400 (`CONFIRM_*_INVALID`) → 422 `STOCK_PREPARATION_VERSION_POLICY_UNSUPPORTED` at the unified allowlist — no FE branch keyed on the old codes; external consumers keying on 400 should note the change.
- **NIT**: material-match's implemented set now enumerates the 3 values explicitly (no longer derived from its local enum — enum widening can't silently widen the guard).

### DIFF
--- a/apps/web/src/services/integration/stockPreparation/materialMapping.ts
+++ b/apps/web/src/services/integration/stockPreparation/materialMapping.ts
@@ -29,16 +29,21 @@ export type StockPreparationMatchStatus =
   | 'not_found'
   | 'version_conflict'
 
+// OD2 round-1 hardening: `category_rule` was REMOVED from this union and the whitelist below — it
+// remains a reserved enum name server-side (stock-preparation-mvp-generation.cjs VERSION_POLICIES)
+// with NO implementation branch, and the server now refuses it fail-closed (422
+// STOCK_PREPARATION_VERSION_POLICY_UNSUPPORTED). Legacy stored rows carrying it never auto-match
+// (they degrade to the visible missing_mapping path); on the read side such a stored value can
+// still cross the wire as a plain string (versionPolicyCounts keys / candidate row versionPolicy).
 export type StockPreparationVersionPolicy =
   | 'drawing_and_version'
   | 'drawing_only'
-  | 'category_rule'
   | 'manual'
 
 // Fixed FE whitelists mirroring the server's frozen vocabularies (stock-preparation-material-match /
 // -mvp-generation .cjs). The candidates read zero-fills byMatchStatus over exactly the five statuses
 // (+ an optional `unknown` fold), and the sync route validates defaultVersionPolicy against exactly
-// these four policies (OD2: required per request, no server default).
+// these three IMPLEMENTED policies (OD2: required per request, no server default).
 export const STOCK_PREPARATION_MATCH_STATUSES: readonly StockPreparationMatchStatus[] = [
   'matched',
   'pending_confirm',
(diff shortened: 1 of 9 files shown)


zensgit:
Owner GO was scoped to exact head `5db599796`. After #4454 merged, this branch was rebased onto main with zero conflicts — new head `f6d7e796f` is content-identical (pure rebase, `git range-diff` clean); auto-merge armed per the GO.

PR #4464, opened 2026-07-18
fix(directory): T2 source-freeze linearization and apply-time recheck

zensgit:
## T2 - source sync freeze (Transfer MVP slice 2)

> **UNARMED while exact-head CI runs.** T1 landed as `b9b354a38`; the production slice was replayed onto that merge without semantic drift (`git range-diff`: 6/6 commits `=`). Exact review head: `8b8f8ffde`; its final commit is test-only CI compatibility over production head `63875c021`.

An active org transfer freezes its source integration before directory sync claims a lease. A shared transaction-scoped PostgreSQL advisory lock also closes the create/refreeze-vs-sync apply race, so a freeze that linearizes first cannot be followed by committed local-directory writes.

### Runtime behavior

- Entry gate: an active frozen transfer throws typed `409 DIRECTORY_SYNC_FROZEN_BY_TRANSFER` before the run lease, producing no run row and no provider pull.
- Apply gate: the local-apply transaction acquires the source freeze advisory lock as its first operation and rechecks frozen state before every directory upsert, absence sweep, membership/identity/link write, projection, deprovision, or completion write.
- Freeze writers: transfer create and `freezeSourceSync=true` refreeze take the same lock before persisting freeze-affecting state.
- Linearization: apply-first may commit before freeze activates; freeze-first makes apply roll back. A late freeze can still leave a failed run row/provider pull, but no local-directory mutation commits after the freeze.
- Scheduler ticks treat the typed freeze as a deliberate quiet skip. Read-only preview remains available.
- A pre-T1 schema is detected with `to_regclass`; no transfer table means no transfer can freeze a sync.

### Operator override

`PATCH /api/admin/directory/org-transfers/:id/source-sync-freeze`

The platform-admin-only body contract is exactly `{ "freezeSourceSync": boolean }`. Unknown fields, arrays, wrong types, missing transfers, and terminal transfers fail closed and write no success audit. Successful changes write one values-free `directory.org_transfer.source_sync_freeze` audit row.

### Independent verification at exact head `8b8f8ffde`

Production behavior was verified at `63875c021`, then the test-only compatibility commit `8b8f8ffde` was independently rechecked:

- Fresh PostgreSQL full migration chain and second replay: pass.
- T1 + T2 real-DB suites: 38/38.
- T1 + T2 + sync lease + orchestration: 54/54.
- Exact affected unit files after the compatibility fix: 117/117; scheduler: 12/12.
- Full backend unit lane at `8b8f8ffde`: 415 files / 5748 tests; backend type-check: pass.
- The compatibility commit changes only two pre-existing test files. It re-exports the real typed freeze error from a partial route mock and models the T1/T2 schema-probe/freeze-query order in positional lease mocks; production code is unchanged.
- Mutation checks: removing the class re-export makes the async route-error case red; removing the freeze-query mock responses makes the lease reclaim case red at the wrong query stage.
- Exact-head required GitHub CI is authoritative before landing.

### Load-bearing production mutations replayed by Codex

Each mutation was applied alone and restored before the next:

- Remove apply-time frozen-state recheck: both freeze-first race legs red; sync committed instead of throwing.
- Remove sync apply advisory lock: create-first barrier red because sync never parked.
- Remove create advisory lock: apply-first barrier red because create never parked.
- Remove refreeze advisory lock: apply-first barrier red because refreeze never parked.
- Remove the entry gate: headline test red with one run row instead of zero.
- Remove scheduler frozen-skip branch: scheduler test red with a rejected tick.
- Remove PATCH field allowlist: body-smuggling test red (`200` instead of `400`).

### Scope boundary

This closes T2's code-side source-freeze invariant. It does not claim the staging dual-corp collision decision, conditional T2.5 tenant-key migration, or T3-T5 transfer apply work; those remain behind T2-Gate evidence.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -210,6 +210,12 @@ jobs:
       - name: T1 org-transfer CI wiring contract
         run: node --test scripts/ops/t1-org-transfer-ci-wiring.test.mjs
 
+      # T2 CI two-point wiring contract (no DB): the source-freeze suite must stay wired into
+      # BOTH vitest.config.ts (exclude) AND plugin-tests.yml (approval real-DB step). Dropping
+      # either point silently disables the §12.2 freeze proof.
+      - name: T2 source-freeze CI wiring contract
+        run: node --test scripts/ops/t2-source-freeze-ci-wiring.test.mjs
+
       # The production monitor invokes the same state helper covered here.
       # Keep this in the required 18/20 matrix so HTTP classification, silence
       # barriers, and alert/recovery streak semantics cannot drift silently.
@@ -677,6 +683,7 @@ jobs:
             tests/integration/directory-sync-run-lease.db.test.ts \
             tests/integration/directory-tenant-change-immutable.db.test.ts \
             tests/integration/directory-sync-orchestration.db.test.ts \
+            tests/integration/directory-org-transfer-source-freeze.db.test.ts \
             tests/integration/directory-sync-schedule-timezone.db.test.ts \
             tests/integration/directory-local-provider-bootstrap.db.test.ts \
             tests/integration/directory-local-provider-scheduler-exclusion.db.test.ts \
--- a/packages/core-backend/src/directory/directory-sync-scheduler.ts
+++ b/packages/core-backend/src/directory/directory-sync-scheduler.ts
@@ -1,7 +1,7 @@
 import { Logger } from '../core/logger'
 import { query } from '../db/pg'
 import { SchedulerServiceImpl } from '../services/SchedulerService'
-import { DirectorySyncInProgressError, reclaimStaleDirectorySyncRuns, syncDirectoryIntegration } from './directory-sync'
+import { DirectorySyncFrozenByTransferError, DirectorySyncInProgressError, reclaimStaleDirectorySyncRuns, syncDirectoryIntegration } from './directory-sync'
 import { deliverDirectoryInactiveLinkedAlert } from './directory-sync-alert-delivery'
 import { resolveDirectoryScheduleTimezone } from './directory-sync-timezone'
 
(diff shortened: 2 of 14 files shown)


zensgit:
**Adversarial gate — APPROVE_WITH_HOLDS → hold closed** (at `09c86d16c`). M1/M2/M3 red with exact claimed signatures; **reviewer M4 (neuter the scheduler frozen-skip branch) stayed green = P2 untested guard** — the multi-day steady state of a transfer would emit job:failed + logger.error every nightly tick.
**Fixes (head now `a83fb577f`, restacked on T1 `5800e5d82`)**: ① scheduler unit leg added (REAL DirectorySyncFrozenByTransferError re-exported via importOriginal — same factory-trap discipline as the lease-skip leg; frozen tick resolves-undefined, no digest alert) — the same M4 mutation now reds with 'promise rejected instead of resolving'; ② in-flight window documented in the error docblock (gate ran once at entry; exposure bounded by DT-OPS-01 breaker + T1 no-ops + runbook ordering; T3 hardening = re-assert freeze before the sweep); ③ async 409 leg now asserts transferId. Unit 12/12 · suite 8/8 · tsc clean.
**Body correction per gate**: delta is 7 files (not 6). to_regclass notes recorded (search_path-dependent; schema-qualifying marginally stricter — landing-window option). Unarmed.

zensgit:
## Independent Codex re-review — APPROVE (implementation), owner-held

Reviewed exact stacked head `eadb5d8da6fdf645c6f15f9a60fbdc3673bf84b6` after restacking on corrected T1. I found **0 remaining behavioral P1/P2 findings**.

What I independently verified:
- supported platform-admin `PATCH /api/admin/directory/org-transfers/:id/source-sync-freeze` replaces the raw-SQL-only override;
- shared `ensurePlatformAdmin`, strict boolean-only body, server-derived resource/org, row lock, nonterminal guard, unfreeze/refreeze, and values-free post-commit audit are all present;
- failure paths (auth/input/missing/terminal) do not write audit rows;
- composed unavailable-adapter behavior leaves the transfer active and source frozen, with no sync run and no absence sweep;
- the restacked final tree is byte-identical to the implementation tree already reviewed and mutation-tested.

Evidence rerun on this exact branch:
- real PostgreSQL T1+T2 suites: **25/25**;
- scheduler unit suite: **12/12**;
- T1/T2 wiring guards: **4/4**;
- backend typecheck: **pass**;
- mutation A: restoring the production no-op fallback makes the composed no-run/no-sweep test fail;
- mutation B: removing the terminal guard makes the terminal API test fail (`409` becomes `200`);
- GitHub stacked check: **settled, 0 fail**.

PR remains **OPEN / unarmed**. This is an implementation verdict, not merge authorization.

PR #4468, opened 2026-07-18
fix(web): plm-workbench focus allowlist — include /stock-prep

zensgit:
V0 of the generalization proposal (#4467) landing order — the one item the owner marked 现在可开发.

plm-workbench-focused mode's route allowlist (`main.ts:155`) omitted `/stock-prep`, so the stock-prep operator shell's natural audience was redirected to `/plm`. One-line allowlist addition + rationale comment. The route keeps its own `integration:write` permission gate — this only restores reachability for already-authorized users.

vue-tsc green; App / platform-shell-nav / featureFlags.plm specs 11/11. The guard allowlist is inline in main.ts with no direct spec — extracting it for testability belongs to P2 (manifest-driven focus modes), not this fix.

**Not armed.**

### DIFF
--- a/apps/web/src/main.ts
+++ b/apps/web/src/main.ts
@@ -154,7 +154,11 @@ router.beforeEach(async (to, _from, next) => {
 
     if (typeof flags.isPlmWorkbenchFocused === 'function' && flags.isPlmWorkbenchFocused()) {
       const path = String(to.path || '')
-      const allowedPrefixes = ['/plm', '/workflows', '/approvals', '/integrations']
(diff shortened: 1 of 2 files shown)


zensgit:
Timing note: the round-7 HOLD arrived after the serial merge train had already landed this PR at `fbbe21e33` (the review's live view predated the merge). The vacuous-ordering P2 is real and is fixed as a follow-up: the pin is now AST-based (real call positions + exact allowlist array), with the owner's discriminating mutation (delete the real call) verified RED — see the follow-up PR.

PR #4491, opened 2026-07-20
feat(automation): FWB activation — production write_approval_form_values action (flag-gated, default OFF)

zensgit:
Closes the owner P1「FWB 尚无生产调用链」: users can now actually configure 审批表单值写回 — the action type is registered, save-validated fail-closed, executed through the real rule chain, and the §11 Q6 four-gate set is production-bound.

**Grounding**: every decision cites its FWB0 lock D-number in code (D1 approved-only / D2 own-sheet target / D4 immutable form_snapshot, values never in config/payload / D6 select closed vocabulary / D9 one-transaction atomicity / D10 no half-durable path / D11 approval.completed-only placement + identity requirement / §11 Q3 no explicit targets / §11 Q6 gates + confirmation hash).

**Wiring**: allowlist restructure keeps FWB OFF task_created (shared base set — the naive spread would have leaked it); save gate in createRule AND updateRule (partial-update shapes covered); executor case with fail-closed preconditions (flag/durable/identity/txn-seam, gates default-DENY); chained event via produceAutomationEvent (REPLACE — never the legacy bus); claim key = §2.1 deriveActionKey over identity.structuralPath (byte-identical twins stay distinct).

**Evidence**: real-DB spec 8/8 (save-gate 11 negatives+positive · flag matrix · four-way single-commit + injected-abort atomicity · net-once redelivery + NEW-instance positive control · V4 revoked-authority through the REAL bound gates · fail-closed defaults ×3 · structural-path twins). Mutations 3/3 killed at exact head: full default-allow gates → fail-closed golden RED; config-hash identity (structuralPath dropped) → 5 RED incl. twins; confirmation-hash save check disabled → save-gate RED. tsc 0; migration + automation-v1 227/227; full unit suite running (will confirm on the thread). Two-point wired.

**Honest scope**: FE rule-editor exposure (action-type option + mapping panel reusing fwbMappingConfig) = named follow-up slice; boot needs no new wiring (the trigger chain + durable adapters already reach executeRule; gates bind at AutomationService construction). Flags stay OFF. Salvage note: built across an agent session-limit death — the wip checkpoint commit precedes the real one.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -630,6 +630,7 @@ jobs:
             tests/integration/multitable-automation-execution-ledger-realdb.test.ts \
             tests/integration/multitable-automation-outbound-intent-realdb.test.ts \
             tests/integration/multitable-fwb-write-action-realdb.test.ts \
+            tests/integration/multitable-fwb-activation-realdb.test.ts \
             tests/integration/multitable-automation-dispatcher-claim-realdb.test.ts \
             tests/integration/multitable-automation-dispatch-loop-realdb.test.ts \
             tests/integration/multitable-automation-outbox-enqueue-realdb.test.ts \
--- a/packages/core-backend/src/db/migrations/zzzz20260720120000_add_write_approval_form_values_automation_action.ts
+++ b/packages/core-backend/src/db/migrations/zzzz20260720120000_add_write_approval_form_values_automation_action.ts
@@ -1,0 +1,44 @@
+/**
+ * Migration: `write_approval_form_values` automation action (FWB activation, FWB0 design lock
+ * `approval-form-writeback-fwb0-designlock-20260712.md` D11) — widen the action_type CHECK constraint.
+ * A dedicated restricted action (NOT reusing create_record — lock §1): save-allowed only on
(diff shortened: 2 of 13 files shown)


zensgit:
Full unit suite at head:  / 0 failed — flag-OFF byte-identical at full-battery scale.

zensgit:
Correction + closure: my earlier「full unit suite / 0 failed」comment was a piped-tail artifact (the background command's exit code was tail's, not vitest's — the suite had 3 real reds). The reds were the repo's two durable structural guards correctly catching the new FWB INSERT without dispositions. Fixed at bbd8e0292: xbase-write-EXEMPT (D2 — no target fields in the config schema, can never retarget) + revision-emitted markers, and the method relocated after executeCreateRecord so the guard's first-INSERT anchor keeps pinning the GATED create site (guards themselves untouched). Verified: realdb 8/8 · **full unit suite 6911 passed / 0 failed** (both guards green) · tsc 0.

zensgit:
Superseded by the Codex approval data stack (#4540 -> #4524 -> #4531 -> #4539). The authoritative stack preserves and expands this PR's eight production-chain tests, migration, D11 allowlist, four activation gates, transactional executor path, and structural-guard dispositions. Closing to prevent accidental landing of the obsolete lineage. No branch deletion.

PR #4496, opened 2026-07-20
test(ops): real-DB CI-wiring executability contract — rounds 5-6: real YAML parse (PyYAML) + literal-URL pin(b) (audit A1)

zensgit:
## What

The 15 `scripts/ops/*-ci-wiring.test.mjs` guards proved a real-DB suite was "wired" by finding its
path **string** in `packages/core-backend/vitest.config.ts` and `.github/workflows/plugin-tests.yml`.
Membership is not executability: the suite could be renamed or deleted, moved into a step that is not
the real-DB step, or left in a step that never runs — and every guard stayed green while the real-DB
proof silently never executed (`vitest` exits 0 on an unmatched path argument, and every such suite is
`describeIfDatabase`-guarded, so it skips green whenever the DB env is absent).

This PR replaces string membership with an **executability contract**, implemented once in a shared
helper and imported by all 15 guards.

## What the contract pins

`scripts/ops/ci-realdb-step-contract.mjs` locates the real-DB step by its **exact stable `id:`**
(never by `- name:` prefix) and pins all four of:

| pin | requirement |
|---|---|
| (a) | `if: matrix.node-version == '20.x'` — the step runs in the required `test (20.x)` leg |
| (b) | `env.DATABASE_URL` is a real YAML key whose value is a **literal PostgreSQL URL** — scheme-anchored `/^postgres(?:ql)?:\/\/\S+$/` after trim, containing **no `${{ … }}` Actions expression** (round 6, owner P2: an expression can resolve to the EMPTY string at runtime — `${{ secrets.DOES_NOT_EXIST }}` → `''` → every DB-gated suite skips-green; the guard runs pre-install and cannot evaluate expressions, so it refuses them all, mixed literal+expression included) |
| (c) | a **real** `vitest` invocation in the `run:` script uses `--config vitest.integration.config.ts` — the default config excludes every DB-gated suite, so running under it is a silent no-op |
| (d) | the suite path is a whole-file argument of **that same invocation** — (c) and (d) are joint |

## How the YAML is read — round 5, owner ruling (2026-07-21, binding)

Rounds 1–4 read the workflow with a hand-rolled text masker (regex step discovery over a
blanked-scalar view). Successive gates kept defeating it with **new YAML spellings** — `"run": |` /
`'run': |` (quoted keys), `run : |` (spaced key), bare sequence-item block scalars (`- |`),
`!!str` / `~` / `null` / `&a ""` empty values — because a text masker re-implements the YAML grammar
one spelling at a time. The owner ruled that chasing spellings is the wrong shape.

The helper now feeds the workflow to a **real YAML parser** and reads the pins off the parsed
structure:

1. **Parse bridge**: `python3` + PyYAML (`yaml.safe_load` → JSON on stdout), spawned per distinct
   workflow text with an in-process cache. **Fail-closed in every direction**: a missing
   interpreter, a missing PyYAML, or a YAML parse error all throw with distinguishable
   `PYYAML_MISSING:` / `YAML_PARSE_ERROR:` signatures, and the guard goes red. A workflow PyYAML
   cannot parse is also a workflow GitHub will not run, so a parse failure is never a green path.
2. **Step lookup** walks `jobs` → job (document order) → `steps[]` → first mapping with the exact
   `id`. Text inside any scalar value — a `run:` script in any spelling, a heredoc, a bare `- |`
   list item — is just a string after parsing and can never be discovered as a step.
3. **Pins (a)/(b)** read the step's `if` and `env.DATABASE_URL` as **parsed values**. Every empty
   spelling collapses to `''`/`null` in the parser — `''`, `""`, `"   "`, bare key, `~`, `null`,
   `!!str` with no content, `&a ""`, an alias to an anchored empty string, and `|`/`>`/`|-`/`>-`/
   `|+`/`>2` with an empty or whitespace-only body — so there is **no per-spelling code left to
   get wrong**. `if: false` parses as a boolean, not the required string. **Round 6**: pin (b) no
   longer accepts any non-empty string — the value must be a literal `postgres://`/`postgresql://`
   URL with no `${{` anywhere (see the pin table and MUTATION 9).
4. **Pins (c)/(d)** parse **shell, not YAML** — the parsed `run` string is the exact script a
   runner hands to bash: full-line `#` comments dropped before `\` continuations are joined, a `#`
   reached mid-continuation terminates that command (bash semantics in both directions), each
   logical line split on `;`/`&&`/`||`/`|`/`&`, and the binary each command **executes** resolved
   (leading `VAR=val` stripped; `pnpm`/`npx`/`yarn`/`bun` runner prefixes walked through their
   value-flags and `exec`/`dlx`/`x` subcommands). A command whose binary is not `vitest` is inert.

**All hand-rolled YAML masking/step-discovery is deleted, not layered over**: `BLOCK_SCALAR_HEADER`,
`maskBlockScalars`, `sequenceItemBlocks`, `scalarYamlValue`, `envValueIsNonEmpty`,
and the raw/masked dual-view plumbing are gone; `runScriptLines` was **reimplemented, not removed**
(gate correction) — the new `runScriptLines(step)` at `ci-realdb-step-contract.mjs:239` reads the
**parsed** `step.run` value off the PyYAML structure, not masked raw text, so no masking survives
(`git show f9e9b2e49 -- scripts/ops/ci-realdb-step-contract.mjs`).

### Why `python3` + PyYAML and not `js-yaml`

These guards run in the required no-DB `test` job **before `pnpm install`** (deliberately — the
workflow comment says wiring drift must fail the required check fast, before install/build). No npm
package is importable at that point. The `test` job does not use `setup-python`, so `python3` is the
`ubuntu-latest` system interpreter, which ships PyYAML.

**Verified, not assumed** — at round-5 head `f9e9b2e49`, run
[29813187850](https://github.com/zensgit/metasheet2/actions/runs/29813187850): all 15 guard steps
succeeded on **both** matrix legs (`test (18.x)` + `test (20.x)`), before `pnpm install`, on the
real runner — including the T2-Gate step that spawns the bridge ~100× for the synthetics. Locally
verified too (macOS `/usr/bin/python3` 3.9.6 + PyYAML 6.0.1). If a future runner image dropped
PyYAML, every guard would fail with `PYYAML_MISSING:` — red, never silently green.

PyYAML notes, each fail-closed or moot: duplicate mapping keys (PyYAML keeps the last; GitHub
rejects the workflow outright, so the spelling cannot pass CI at all) and YAML 1.1 booleans
(`on:` → `true`; only `jobs` is read, and the bridge stringifies keys).

## Changes (whole PR, vs merge-base)

- **`scripts/ops/ci-realdb-step-contract.mjs`** — the shared contract (this round: YAML reading
  replaced by the parse bridge; shell-command parsing for (c)/(d) unchanged).
- **`.github/workflows/plugin-tests.yml` — exactly two added lines**, the stable `id:` on each
  real-DB step (`multitable-real-db-integration`, `approval-real-db-integration`):
  `git diff origin/main...HEAD --stat -- .github/workflows/plugin-tests.yml` → `2 insertions(+)`,
  nothing else in that shared file is touched.
- **15 guards** import the helper; 15/15 assert existence-on-disk, id-located step scoping, and the
  negative "must not be wired into the other real-DB step".
- The helper has **no CI step of its own**: it executes in the required no-DB `test` job 15 times
  over via guards already wired to their own `node --test` steps; its synthetic coverage lives in
  `t2gate-collision-mechanism-ci-wiring.test.mjs`, likewise already wired.

## How each pin is proven

Every bypass below is a **permanent synthetic** in `t2gate-collision-mechanism-ci-wiring.test.mjs`,
fed to the helper as a crafted workflow string (never the real file), so the coverage cannot rot.
Every previously proven bypass was converted to run against the parser and **stays red**:

| synthetic | bypass it kills | red signature |
|---|---|---|
| MUTATION 1 | an earlier decoy step whose `- name:` merely *contains* the prefix carries the payload while the real step is gutted | `not found in plugin-tests.yml … located by exact id` |
| MUTATION 2 | `if:` flipped to `'18.x'` / `false` / removed | `must carry if: matrix.node-version == '20.x'` |
| MUTATION 3 | `env.DATABASE_URL` removed | `must have env.DATABASE_URL …` |
| MUTATION 4 | config swapped to the default `vitest.config.ts` | `must run a REAL vitest command with --config vitest.integration.config.ts` |
| MUTATION 5 | `DATABASE_URL: ''` / `""` / `"   "` / `'  '` / bare key, **and (new, owner-named) `!!str` / `~` / `null` / `&a ""` / alias `*empty`** — a present key whose value is empty in any spelling | `… with a NON-EMPTY value` |
| MUTATION 6 | a decoy command line (`echo vitest --config …`, quoted echo, `pnpm exec echo vitest`, comment) satisfying the config pin while the only real invocation uses the default config | `must run a REAL vitest command …` |
| MUTATION 6b | two real vitest commands — config on one, suite path on the other | `isSuiteWiredInRealDbStep` returns `false` for that file |
| MUTATION 6c | a `#` line parked between the vitest command and its suite arguments (bash terminates the command there) | `false`, `wholeFileVitestArgs` is `[]` |
| MUTATION 7 | `DATABASE_URL: \|` / `>` / `\|-` / `>-` / `\|+` / `>2` with an empty / whitespace-only body — resolves to `''` | `… with a NON-EMPTY value` |
| MUTATION 8 | a phantom `- name:` / `id:` step minted INSIDE a `run:` value — **all spellings**: block-scalar heredoc, multi-line quoted scalar, **(new) `"run": \|`, `'run': \|`, `run : \|`, bare `- \|` steps item**; plus the multi-line **plain**-scalar carrier, which is invalid YAML (PyYAML and GitHub both reject it) and now **fails CLOSED** — under the round-4 masker that exact text minted a phantom and PASSED | `not found … located by exact id`; plain-scalar carrier: `failing CLOSED … YAML_PARSE_ERROR` |
| MUTATION 9 (round 6, **owner-reproduced bypass**) | `DATABASE_URL: ${{ secrets.DOES_NOT_EXIST }}` (verbatim + quoted + `${{secrets.X}}` no-space + `${{ env.DATABASE_URL }}`) — non-empty TEXT pre-install, **empty at runtime** ⇒ skip-green under the round-5 "non-empty" pin; mixed literal+expression `postgresql://user:${{ secrets.PW }}@…` (spaced + no-space); non-URL literals `true` / `'true'` / `1` / `file.txt` | `… env.DATABASE_URL` (all ten spellings red) |

Positive controls, so no pin degenerates into "reject everything": a name-prefix decoy alongside a
healthy id step passes; plain / single-quoted / inline-commented / **tagged (`!!str <url>`)** /
**anchored (`&db <url>`)** real `DATABASE_URL` values pass; each of the six block-scalar headers
WITH content passes; a genuine id step after a phantom-carrying heredoc passes all four pins; **a
real step written entirely with quoted/spaced keys passes all four pins** (structure, not
spelling, in both directions); a benign `echo "… vitest --config …"` beside the real command
passes; a `# comment \` line does not swallow the next command; **the verbatim
`postgresql://postgres@localhost:5432/metasheet_test` both real steps carry is accepted in both
official schemes (`postgres://` too), unit-level and through the full contract**; 6b/6c keep their in-test positive
controls; a relocated suite is a membership miss, not a collapsed parse.

### The parser-era synthetics are load-bearing

Each element was mutated in isolation against the committed tests, and the resulting RED recorded
(tree restored by exact-path `git checkout` after each):

| mutation on the helper | result |
|---|---|
| pin (b) stops rejecting `null` (drop the `value === null \|\| value === undefined` guard) | 24 pass / 1 fail — MUTATION 5 (`~` / `null` spellings green) |
| pin (b) stops trimming (`String(value) !== ''`) | 23 pass / 2 fail — MUTATION 5 + MUTATION 7 (whitespace values green) |
| parse bridge fails OPEN (returns `null` instead of throwing on `YAML_PARSE_ERROR`) | 24 pass / 1 fail — MUTATION 8 (invalid-YAML carrier no longer fails closed) |
| pin (a) loosened to any `\d+\.x` | 24 pass / 1 fail — MUTATION 2 (`'18.x'` green) |
| pin (b) reverted to the round-5 body (`String(value).trim() !== ''`) | 25 pass / 1 fail — MUTATION 9 (owner's `${{ secrets.DOES_NOT_EXIST }}` repro green again) |

## Accepted residuals (deliberately NOT fixed — restated per owner instruction)

Owner stop-line on arbitrary GitHub Actions / shell semantics. Each was **re-probed against this
head's parser** and still passes the contract — they are known, accepted gaps, not claims of
coverage:

1. a vitest command that is present but **never executes**: `true || pnpm exec vitest --config vitest.integration.config.ts run <file>` (shell short-circuit is not evaluated)
2. a `-t` filter that matches **zero tests** (vitest exits 0)
3. step-level **`continue-on-error: true`** (the step fails, the job stays green)
4. a step with the same `id:` in an **EARLIER job** (jobs are scanned in document order; the first match anchors)

## Test results (this head, `4b0ea18d8` — round 6, rebased onto main `f55d99e12`)

- `node --test scripts/ops/*-ci-wiring.test.mjs` — **69 tests / 69 pass / 0 fail / 0 skipped**
  across **15 files** (68 → 69: MUTATION 9 added). Re-run in full after the rebase — the guards
  bind by step `id:` and parsed structure, so main's workflow churn (DT-CLOSE-01C steps, RCA-window
  tests, W4-PRE-1/W4-0 attendance real-DB wiring) moved line numbers without touching any pin.
- Round-5 CI proof (pre-rebase head `f9e9b2e49`): run [29813187850](https://github.com/zensgit/metasheet2/actions/runs/29813187850)
  — all 15 guard steps green on both `test (18.x)` and `test (20.x)`, before `pnpm install`.
- `git diff origin/main...HEAD --name-only` → the 16 `scripts/ops/` files + `plugin-tests.yml`
  (two `id:` lines only, re-verified post-rebase). No TypeScript changed; no database needed.

**Head-scoped note (rebased 2026-07-22):** this branch now sits directly on `origin/main`
`f55d99e12`; the two-dot and three-dot diffs coincide. The two `id:` lines land at
`plugin-tests.yml:445` (`multitable-real-db-integration`) and `:682`
(`approval-real-db-integration`); the id-located steps carry
`if: matrix.node-version == '20.x'` at `:446`/`:687` and the literal
`DATABASE_URL: postgresql://postgres@localhost:5432/metasheet_test` at `:448`/`:689`
(previously `:419`/`:656` and `:422`/`:663` — moved by main-side steps above them, re-derived
at this head). `requireExecutableRealDbStep` passes all four pins for both ids on the current
workflow content. Any CI verdict is bound to the SHA it ran on.

**UNARMED — owner review pending.** No auto-merge armed, not merged.


zensgit:
## Rework — owner P2 closed in this ticket (no follow-up)

New head: **`61a645427`** (was `d6daf5531`). Single commit, `.mjs` guards only. UNARMED (`autoMergeRequest: null`).

### Owner finding → how it is closed

> The five remaining guards match the suite path against the **whole** `plugin-tests.yml`; four of them NAME a real-DB step ("... in the directory real-DB step" / "with real Postgres") while asserting nothing about the step. Moving the suite's line into a different step leaves them green.

Per the owner's ruling this is narrowed **here**, not deferred. All five now reuse the idiom already landed in this PR for the other 7 — `namedStepBody()` + `wholeFileRunRe()` + the wrong-step negative — anchored to the step that actually runs the suite. Step mapping was **verified against `plugin-tests.yml`**, not assumed:

| guard | suite line | positive anchor (step that runs it) | negative |
|---|---|---|---|
| `b4-department-bindings` | L714 | `Run approval real-DB integration` (step L654) | multitable |
| `pb4-2-archive-readonly` | L711 | `Run approval real-DB integration` | multitable |
| `pb4-3-cycle-detection` | L712 | `Run approval real-DB integration` | multitable |
| `pb4-4-reactivation` | L713 | `Run approval real-DB integration` | multitable |
| `stock-preparation-p4-repair` | L646 | **`Run multitable real-DB integration`** (step L418) | approval |

stock-prep is the lone inversion (positive = multitable, negative = approval); the other ten landed guards are approval-positive.

Test **titles** now name the step they actually assert (`... in the directory real-DB step (Run approval real-DB integration)`, `... in the multitable real-DB step (Run multitable real-DB integration)`), so the name-claims-a-step / asserts-nothing gap the owner caught cannot reappear. Existing exclude-entry and file-exists assertions unchanged; per-file house style kept (helper duplicated per file, no shared module).

### Owner's reproduction — replayed, now REJECTED

Mutation per guard: move that guard's suite line out of its named real-DB step into the other real-DB step; run the NEW guard and, side by side, the guard **as it stood at `d6daf5531`** (staged from `git show d6daf5531:<path>`); then `git checkout -- .github/workflows/plugin-tests.yml`.

| guard | OLD guard @`d6daf5531` | NEW guard | after restore |
|---|---|---|---|
| b4-department-bindings | pass 3 / fail 0 — **green, owner's bug** | pass 2 / **fail 1** | pass 3 / fail 0 |
| pb4-2-archive-readonly | pass 3 / fail 0 | pass 2 / **fail 1** | pass 3 / fail 0 |
| pb4-3-cycle-detection | pass 3 / fail 0 | pass 2 / **fail 1** | pass 3 / fail 0 |
| pb4-4-reactivation | pass 3 / fail 0 | pass 2 / **fail 1** | pass 3 / fail 0 |
| stock-preparation-p4-repair | pass 3 / fail 0 | pass 2 / **fail 1** | pass 3 / fail 0 |

The failing assertion is the **step-scope positive** (not exclude, not existence) — exact red signature, B4:

```
✔ vitest.config.ts excludes the B4 department-bindings suite from the no-DB job
✖ plugin-tests.yml runs the B4 department-bindings suite as a whole file in the directory real-DB step (Run approval real-DB integration)
✔ the B4 department-bindings suite file exists on disk
  AssertionError [ERR_ASSERTION]: plugin-tests.yml directory real-DB step (Run approval real-DB integration)
    must run tests/integration/directory-department-bindings.db.test.ts as a whole file
  expected: /\n\s*tests\/integration\/directory-department-bindings\.db\.test\.ts \\/
  operator: 'match'
```
`actual` is the approval step body only — the moved line is no longer in it, which is exactly the fact the old whole-file regex could not see.

### Wrong-step negative is load-bearing (not decoration)

Duplicate the line into the other step while **keeping** the correct one, so the positive still passes:

- b4 (kept in approval, copy added to multitable) → pass 2 / **fail 1**, `AssertionError: tests/integration/directory-department-bindings.db.test.ts must not be wired into the multitable real-DB step`. OLD guard @`d6daf5531`: pass 3 / fail 0.
- stock-prep (kept in multitable, copy added to approval) → pass 2 / **fail 1**, `AssertionError: tests/integration/stock-preparation-p4-repair-once-realdb.test.ts must not be wired into the approval real-DB step`.

### Pre-existing assertions still load-bearing (3 of 5, exceeding the required 2)

| guard | drop exclude entry from `vitest.config.ts` | move suite file off disk |
|---|---|---|
| b4-department-bindings | ✖ `vitest.config.ts excludes ... from the no-DB job` (2/1) | ✖ `... suite file exists on disk` (2/1) |
| pb4-2-archive-readonly | ✖ same (2/1) | ✖ same (2/1) |
| stock-preparation-p4-repair | ✖ `vitest.config.ts excludes the P4 repair suite ...` (2/1) | ✖ `the P4 repair suite file exists on disk` (2/1) |

In every case exactly one assertion reddens and the other two stay green — the three points are independent. All restored to 3/3 after each.

### Positive control / full sweep

`node --test scripts/ops/*ci-wiring*.test.mjs` at `61a645427`: **tests 54, pass 54, fail 0** — same count as `d6daf5531` (54/54), because each wrong-step negative was added as an assertion inside the existing step test rather than a new `test()` block. Working tree verified clean (`git status --porcelain` empty) after every mutation and before the push.

Local runner Node v25.9.0; CI authority is Node 20 (these guards run in the `test` job, steps L156/L162/L168/L174/L231 of `plugin-tests.yml`) — noted, not blocking; the idiom is byte-identical to the ten guards CI already runs there. No TypeScript changed, so no `tsc` run.

zensgit:
## Owner P2 rework — guards now pin EXECUTABILITY, not membership

**New head: `f2f488b1c23b6930b01ae24c78dedf9417ab2c86`. UNARMED — not merged, no auto-merge.**

### The finding this closes

The guards asserted MEMBERSHIP of a suite path in a step body located by a long `- name:` prefix. Membership is not executability. Reproduced at the previous head `61a645427`, against the real `plugin-tests.yml`:

| bypass | sweep at `61a645427` | real-DB suites |
|---|---|---|
| flip the step's `if: matrix.node-version == '20.x'` → `'18.x'` | **54/54 green** | never run in the required `test (20.x)` leg |
| remove the step's `env.DATABASE_URL` | **53 pass / 1 fail** (only t2gate) | all skip-green via `describeIfDatabase` |
| earlier decoy step whose name merely CONTAINS the prefix (`- name: Run approval real-DB integration (decoy prep)`) carrying the payload, real step gutted | **54/54 green** | nothing runs |

### Owner contract — how each of the four pins is satisfied

Steps are located by **exact `id:`**, never by title. `.github/workflows/plugin-tests.yml` gained `id: approval-real-db-integration` and `id: multitable-real-db-integration`; `git diff origin/main -- .github/workflows/plugin-tests.yml` is **exactly those two lines (2 insertions, 0 deletions)** — nothing else in that shared file was touched.

One shared module, `scripts/ops/ci-realdb-step-contract.mjs` (by owner ruling this supersedes the per-file-duplication house style), extracts the step by exact id — with block scalars masked, so an `id:`/`if:`/`env:` token inside a `run:` script cannot anchor or satisfy anything — and pins all four:

| pin | helper | evidence on the real file |
|---|---|---|
| (a) `if: matrix.node-version == '20.x'` | `stepRunsOnNode20Matrix` | both ids resolve to `if: matrix.node-version == '20.x'` |
| (b) `env.DATABASE_URL` on the step | `stepHasEnvDatabaseUrl` | approval → `[DATABASE_URL]`; multitable → `[DATABASE_URL, METASHEET_REAL_DB_TEST_STEP]` |
| (c) run uses `vitest.integration.config.ts` | `stepInvokesVitestIntegrationConfig` | both steps' `pnpm … exec vitest --config vitest.integration.config.ts run` |
| (d) suite is a whole-file run arg | `wholeFileVitestArgs` | approval 60 whole-file args, multitable 226 |

### Four permanent synthetic mutations — red signatures

Committed in `t2gate-collision-mechanism-ci-wiring.test.mjs`. Synthetic by construction: each feeds the helper a **crafted workflow string**, so the coverage cannot rot with the real file.

1. **name-prefix decoy** carrying the payload while the id'd step is absent → `extractStepById(...) === null`, throws `real-DB step id "…" not found in plugin-tests.yml — steps are located by exact id, never by name prefix`
2. **`if: false` / `'18.x'` / `if:` removed** → throws `must carry if: matrix.node-version == '20.x' — otherwise it never runs in the required test (20.x) leg`
3. **`env.DATABASE_URL` removed** → throws `must have env.DATABASE_URL (a real YAML key, not a comment) — otherwise every describeIfDatabase suite in it skips green`
4. **config swapped to default `vitest.config.ts`** → throws `must run vitest with --config vitest.integration.config.ts — the default vitest.config.ts excludes these DB-gated suites`

Plus positive controls, so no mutation can pass for the wrong reason: a well-formed crafted workflow satisfies all four pins; **a decoy alongside a healthy id'd step is correctly ignored (stays green)** — id-anchoring is decoy-insensitive, which is the point; a run-script `id:` token does not anchor; a step whose only listed suite is a different file yields a membership miss with a non-empty arg list.

### The three original bypasses, now applied to the real file at the new head

| mutation of the real `plugin-tests.yml` | sweep | first signature |
|---|---|---|
| `if` → `'18.x'` | **43 pass / 17 fail** | `must carry if: matrix.node-version == '20.x'` |
| `env.DATABASE_URL` removed | **43 / 17** | `must have env.DATABASE_URL` |
| decoy carries payload, id'd step gutted | **43 / 17** | `must have env.DATABASE_URL` |
| *(new pin c)* `--config vitest.config.ts` | **43 / 17** | `must run vitest with --config vitest.integration.config.ts` |
| *(new)* the `id:` line deleted | **43 / 17** | `not found in plugin-tests.yml — steps are located by exact id` |
| *(control)* decoy added, real step intact | **60 / 0 green** | — |

Bypass #2 previously reddened only t2gate (53/1); it now reddens every guard bound to that step. Bypass #1 went from 54/54-green to 17 failures.

### Guards migrated (15 — all of them; every existing assertion kept)

`b4-department-bindings`, `b5a-routing-policy`, `b5b-failclose`, `b5b-routing-resolver`, `b5c-routing-routes`, `b6-equivalence`, `b7-reconciliation`, `pb4-2-archive-readonly`, `pb4-3-cycle-detection`, `pb4-4-reactivation`, `t1-org-transfer`, `t2-source-freeze` (uniform shape) · `stock-preparation-p4-repair` (multitable-positive / approval-negative) · `b7-round2` (keeps its exact-index adjacency proof, now over the id-located step) · `t2gate-collision-mechanism` (keeps its `test.exclude` depth-1 parser and its four config synthetics).

Retained in every one: the `vitest.config.ts` exclude-entry assertion, the on-disk file-existence assertion, and the "must not be wired into the *other* real-DB step" negative — the negative now goes through `realDbStepWholeFileArgs`, so it can no longer pass merely because the other step was deleted or disabled. Zero `namedStepBody` / `wholeFileRunRe` definitions remain in `scripts/ops/`.

### CI wiring of the helper — no workflow step added or changed

All 15 guards import the shared module and each is already wired to its own `node --test` step in the required no-DB `test` job, so the helper's real-workflow path executes 15× per run. Its mutation coverage is hosted in `t2gate-collision-mechanism-ci-wiring.test.mjs`, already wired at that job's T2-Gate step. That is why the "add a `node --test` step if it needs its own" clause never had to fire, and the shared workflow file's diff stays at the two `id:` lines.

### Sweep

`node --test scripts/ops/*ci-wiring*.test.mjs` → **60/60 pass** (was 54/54). Co-run `t2gate-runbook-values-free-contract.test.mjs` → 12/12. `pnpm install --frozen-lockfile --prefer-offline` exit 0. Workflow re-parsed with a YAML loader after the edit; both ids resolve to real, distinct steps.

Per the owner's stop instruction, this ends here — no further GitHub Actions semantics are parsed.

zensgit:
## Round 3 — both owner-constructed skip-green bypasses closed

Both reproductions were replayed verbatim against `f2f488b1c` first: **both were GREEN**, exactly as
reported.

### Owner finding (a) — `DATABASE_URL: ''` accepted as a valid DB env

An empty string is a *present* YAML key, so the "key exists" pin greened it — while an empty
`DATABASE_URL` makes every `describeIfDatabase` suite skip, i.e. precisely the skip-green pin (b)
exists to stop.

Replay at `f2f488b1c`: `stepHasEnvDatabaseUrl → true`, `isSuiteWiredInRealDbStep → true`.

**How closed.** `stepHasEnvDatabaseUrl` no longer tests "key followed by non-space". It extracts the
scalar value, strips one layer of surrounding quotes and any inline comment, trims, and requires the
result to be non-empty. `''`, `""`, `"   "`, `'  '` and a bare `DATABASE_URL:` all reject.

**Red signature** (`MUTATION 5`, all five forms):

```
real-DB step id "approval-real-db-integration" must have env.DATABASE_URL as a real YAML key
(not a comment) with a NON-EMPTY value ('' / "" / whitespace-only do not count) — otherwise
every describeIfDatabase suite in it skips green
```

**Positive control** (same test, so the pin is not "reject every env value"): plain
`postgresql://…`, single-quoted `'postgresql://…'`, and `postgresql://… # real DB` all still pass
`stepHasEnvDatabaseUrl` **and** the full four-pin contract.

### Owner finding (b) — `echo vitest --config vitest.integration.config.ts` decoy

The config token and the file arguments were matched across the *whole step body*, so a decoy command
line satisfied pin (c) while the only real invocation ran under the default config (which excludes
every DB-gated suite).

Replay at `f2f488b1c`: `stepInvokesVitestIntegrationConfig → true`, `wholeFileVitestArgs → [FILE]`,
`isSuiteWiredInRealDbStep → true`.

**How closed.** Pins (c)/(d) now parse **only** the `run:` script and judge **one command at a time**:
comments are handled with bash semantics, `\` continuations are joined into logical lines, each line
is split on `;` / `&&` / `||` / `|` / `&`, and for every command the **binary it actually executes**
is resolved — leading `VAR=val` stripped, `pnpm`/`npx`/`yarn`/`bun` runner prefixes walked through
their value-flags (`--filter …`) and `exec`/`dlx`/`x` subcommands. A command whose binary is not
`vitest` is inert. `wholeFileVitestArgs` then reports files only from invocations that use the
integration config, which makes **(c) and (d) joint**: a config flag on one command and the suite path
on another can no longer combine.

**Red signatures:**

- `MUTATION 6` — four decoy shapes (`echo vitest --config …`, quoted `echo "vitest --config … run"`,
  `pnpm --filter … exec echo vitest --config …`, and a `#` comment) each with the real command on
  `vitest.config.ts`:

  ```
  real-DB step id "approval-real-db-integration" must run a REAL vitest command with
  --config vitest.integration.config.ts (the executed binary must be vitest — an
  `echo vitest --config …` decoy does not count) — the default vitest.config.ts excludes
  these DB-gated suites, so they would not run
  ```
  with `stepInvokesVitestIntegrationConfig → false` and `wholeFileVitestArgs → []`.

- `MUTATION 6b` — the split-across-commands variant with **two genuinely real** vitest commands
  (integration config on one, the suite path on the other): pin (c) legitimately passes, yet
  `isSuiteWiredInRealDbStep(FILE) → false`.

- `MUTATION 6c` — same class, found while writing the splitter: a `#` line parked between the vitest
  command and its suite arguments. In bash that terminates the command, so those paths are separate
  commands that never run: `wholeFileVitestArgs → []`, `isSuiteWiredInRealDbStep → false`.

**Positive controls** (so the fix is binary resolution, not "reject anything mentioning vitest
twice"):

- a benign `echo "about to run vitest --config vitest.integration.config.ts"` line **beside** the real
  command → still passes, `wholeFileVitestArgs → [FILE]`
- in `MUTATION 6b` the integration-config command's own file **is** reported and **is** wired — the
  miss is joint scoping, not a collapsed parse
- in `MUTATION 6c` pin (c) still holds — the vitest command itself is intact
- a whole-line `# parked command follows \` before the real command does **not** swallow it (bash ends
  a comment at the newline)

### Load-bearing check

Each new synthetic was verified against a revert of exactly the logic it covers (committed first,
tree restored by path afterwards, `git status` clean):

| reverted logic | reds |
|---|---|
| empty-value rejection → back to `DATABASE_URL:\s*\S` | `MUTATION 5` (1 fail / 21) |
| per-command parsing → back to whole-body matching | `MUTATION 6`, `MUTATION 6b` (2 fail / 21) |
| bash-exact comment handling → pre-strip all `#` lines | `MUTATION 6c` (1 fail / 23) |

### Sweep

All four original pins, every pre-existing synthetic, and every real-workflow assertion still pass.

- 15-guard sweep: **66 pass / 0 fail / 0 skipped** (was 60/60 at `f2f488b1c`; +6 = MUTATION 5, 6, 6b,
  6c and two controls).
- Same **66/66 on Node 20.20.2** (CI authority) as on local Node v25.
- `git diff origin/main -- .github/workflows/plugin-tests.yml` is still exactly the two added `id:`
  lines — no other change to that shared file.
- No TypeScript in the diff (helper and guards are `.mjs`), so `tsc` does not apply.

### P3 — PR description

Rewritten to the final SHA: the stale test counts, the "no shared helper was introduced" claim and
the 7-vs-5 guard scoping split are gone. It now states what the PR does and how it is proven, without
round-by-round history.

**New exact head: `4c1cc5d2e620369ee07e976ab2d40b3c3667ccd0`**

**UNARMED** — not merged, no auto-merge armed.

zensgit:
## Round 4 — N1 / N2 / N7 closed at `4a8ae11c5`

Both round-3 bypasses (empty-string `DATABASE_URL`; the `echo` decoy carrying the integration config)
stay closed — replayed green in every run below.

### N1 (BLOCKING) — empty block-scalar `DATABASE_URL` passed pin (b)

**Reproduced** against `4c1cc5d2e`: `DATABASE_URL: |` → `stepHasEnvDatabaseUrl` = `true`,
`isSuiteWiredInRealDbStep` = `true`. Same for `>`, `|-`, `>-`, and for a whitespace-only body. A block
scalar carries no value on the key line, so `scalarYamlValue(' |')` read back the literal `|` and
called it non-empty; at runtime `DATABASE_URL` is empty and every `describeIfDatabase` suite skips.

**Fixed:** pin (b) now resolves the block body from the raw, index-aligned lines. One shared
`BLOCK_SCALAR_HEADER` pattern (`^[|>][0-9+-]*\s*(?:#.*)?$`) serves the masker, the `run:` extractor and
the value reader, so a spelling one of them knows and another doesn't cannot itself become a bypass.

**Permanent synthetic** `MUTATION 7`: 6 headers (`|`, `>`, `|-`, `>-`, `|+`, `>2`) × 3 empty bodies
(none / whitespace-only / several blank lines) = 18 cases must RED, **plus a positive control that all
6 headers WITH content still PASS** — the pin resolves block scalars, it does not reject the form.

### N2 (P2) — phantom step minted inside a `run:` script

**Reproduced:** a `- name:` / `id: approval-real-db-integration` pair inside a heredoc was discovered as
a real step and satisfied all four pins, while the job ran nothing but `cat`.

**Fixed:** masking now runs over the whole workflow **before** sequence detection. While closing it I
found the ordering fix alone was not enough — a `run:` value can also be a multi-line **plain** or
**quoted** scalar, and both spellings still minted phantoms (verified: `found: true, wired: true`).
So the masker now blanks scalar CONTENT generally: any key whose value is non-empty opens a scalar
body, and the indented lines below it are content, not structure. Only `key:` with an empty value
introduces nested structure. Residual imprecision fails **closed** (over-masking hides the step →
every entry point throws). Raw lines stay index-aligned, so pins (c)/(d) still parse real script text.

**Permanent synthetic** `MUTATION 8`: all three carriers (heredoc / plain / quoted) must fail to
anchor, **plus a positive control** that a genuine id step declared after such a script still resolves
and satisfies all four pins.

> A real YAML parser was considered and rejected on a checked constraint, not preference: these guards
> run at lines 157–232 of the `test` job, **before** `pnpm install` (line 278) — there is no
> `node_modules` at that point. Reordering would have required touching `plugin-tests.yml` beyond the
> two `id:` lines.

### Load-bearing check (each element reverted in isolation, tree restored by exact-path checkout)

| reverted | result |
|---|---|
| pin (b) stops resolving block scalars | 24/25, RED `MUTATION 7` — `DATABASE_URL: \| + 0 blank body line(s) must not count` (`actual: true, expected: false`) |
| mask moved back after the carve-up | 24/25, RED `MUTATION 8` |
| mask narrowed to block scalars only | 24/25, RED `MUTATION 8` |
| `BLOCK_SCALAR_HEADER` narrowed to bare `\|`/`>` | 24/25, RED `MUTATION 7` |

### N7 — description rewritten to this head

The false sentence ("block scalars are masked out of the YAML-key surface, so an `id:` … inside a
shell script cannot anchor") was true only *after* step discovery — which is exactly what N2
exploited. The fix makes it true, so the body now states the **ordering** explicitly (mask → discover
→ anchor → resolve) rather than deleting the claim. Every remaining sentence is re-verified at
`4a8ae11c5`; the module header carries the same account. Counts refreshed (66 → 68) with the scope
split spelled out.

### Accepted residuals — recorded in the body, NOT fixed (owner stop-line)

Each re-checked against this head and confirmed to **still pass** the contract: `true || … vitest …`;
a zero-matching `-t` filter; step-level `continue-on-error: true`; a same-`id:` step in an earlier job.

### Raw results

- `node --test scripts/ops/*ci-wiring*.test.mjs` → **68 tests / 68 pass / 0 fail / 0 skipped**, 15
  files (43 lane-guard + 25 t2gate). Identical **68/68 on Node 20.20.2**; also 68/68 on local v25.9.0.
- `git diff origin/main --stat -- .github/workflows/plugin-tests.yml` → `1 file changed, 2 insertions(+)`
  — still exactly the two `id:` lines.
- No workflow step added or modified; helper still has no CI step of its own.

**Head-scoped:** branch is **BEHIND `main` by 12 commits** (`mergeStateStatus: BEHIND`). All numbers
above are local runs at `4a8ae11c5`; a CI verdict binds to the SHA it ran on.

**UNARMED** — not merged, no auto-merge armed.

zensgit:
**Round 5 delivered at `f9e9b2e49` — owner ruling implemented: the hand-rolled YAML masker is GONE, replaced by a real YAML parse.**

What changed (2 files, `scripts/ops/ci-realdb-step-contract.mjs` + `t2gate-collision-mechanism-ci-wiring.test.mjs`):

- The workflow is now parsed by a `python3` + PyYAML bridge (`yaml.safe_load` → JSON), and the four pins are read off the parsed structure (`jobs` → `steps[]` → step keys). **Every masking/regex-step-discovery function is deleted** (`BLOCK_SCALAR_HEADER`, `maskBlockScalars`, `sequenceItemBlocks`, `scalarYamlValue`, `envValueIsNonEmpty`, `runScriptLines`) — not layered over. Pins (c)/(d) still parse **shell** (bash-exact comments/continuations, per-command binary resolution) on the parsed `run` string.
- Bridge is **fail-closed** with distinguishable signatures: missing python3, missing PyYAML (`PYYAML_MISSING:`), or invalid YAML (`YAML_PARSE_ERROR:`) all throw → guard red. Invalid YAML is also a file GitHub itself refuses to run, so a parse failure is never a green path.
- **PyYAML availability verified, not assumed**: the constraint is that these guards run before `pnpm install` in the required `test` job (deliberate fail-fast — so `js-yaml` is not importable). At this head, run [29813187850](https://github.com/zensgit/metasheet2/actions/runs/29813187850) shows all 15 guard steps **success on both `test (18.x)` and `test (20.x)`** on the real `ubuntu-latest` runner, before install — the T2-Gate step alone spawns the bridge ~100× for the synthetics. Local dev box also verified (`/usr/bin/python3` 3.9.6 + PyYAML 6.0.1). Option 2 (moving guards after install) was therefore **not needed** and the workflow was not touched.
- **Every previously proven bypass stays red against the parser** (existence, name-prefix decoy, `if` flips, DATABASE_URL removed/empty/block-scalar-empty, default config, phantom-in-`run:`), and the round-5 spellings are new synthetics: `"run": |` / `'run': |` / `run : |` phantom carriers, bare `- |` steps item, and the four owner-named empty values `!!str` / `~` / `null` / `&a ""` (+ alias `*empty`), each with positive controls (quoted-key REAL step passes all four pins; tagged/anchored real URLs pass). One conversion note: the multi-line **plain**-scalar phantom carrier is invalid YAML — PyYAML and GitHub both reject it — so its synthetic now asserts **fail-CLOSED** (`YAML_PARSE_ERROR`); under the round-4 masker that exact text minted a phantom step and PASSED.
- Load-bearing proofs (mutate helper in isolation → red, restore): drop null-rejection → M5 red; drop trim → M5+M7 red; bridge fails open → M8 red; pin (a) loosened to `\d+\.x` → M2 red.
- **The four accepted residuals stay accepted** and are restated in the PR body, each re-probed against this head's parser and still passing the contract: `true || vitest …` never-executed command; zero-matching `-t` filter; step-level `continue-on-error: true`; same-id step in an earlier job.
- Constraint held: `git diff origin/main...HEAD -- .github/workflows/plugin-tests.yml` = exactly the two `id:` lines (2 insertions).
- Local: `node --test scripts/ops/*ci-wiring*.test.mjs` = **68/68, 0 skipped** on Node 20.20.2 and 25.9.0.

PR body rewritten to this SHA. **UNARMED** — no auto-merge, not merged; owner review pending.

zensgit:
**Round 6 @ `df3f871d5` — owner P2 (2026-07-21) closed: Actions-expression skip-green bypass**

**Bypass reproduced before fixing**: `stepHasEnvDatabaseUrl({env:{DATABASE_URL:'${{ secrets.DOES_NOT_EXIST }}'}})` → `accepted: true` at `f9e9b2e49` — non-empty TEXT pre-install, but GitHub resolves a missing secret/context to the **empty string** at runtime ⇒ `DATABASE_URL=''` ⇒ every `describeIfDatabase` suite skips ⇒ skip-green.

**New pin(b) contract**: refuse ANY `${{` anywhere in the value (mixed literal+expression included) **and** require the trimmed value to match anchored `/^postgres(?:ql)?:\/\/\S+$/`. Both real steps carry the literal `postgresql://postgres@localhost:5432/metasheet_test` (`plugin-tests.yml:275/:319`) — positive control asserts it passes in both official schemes, unit-level and through the full contract.

**Verification (orchestrator-run, agent died of auth mid-lane — commit was complete, I gated it by hand):**
- **MUTATION 9**: 10 red spellings (owner verbatim / quoted / no-space / mixed spaced+no-space / `true`/`'true'`/`1`/`file.txt` / `${{ env.DATABASE_URL }}`), each asserting the `env.DATABASE_URL` red signature.
- **Load-bearing proof**: predicate reverted to the round-5 body ⇒ **25 pass / 1 fail (MUTATION 9)**; restored ⇒ 26/26. Sweep 69/69 across all 15 wiring guards (68→69).
- **Independent attack battery (14 cases)**: uppercase scheme rejected (fail-closed), shell-style `${VAR}` rejected, `$\{{` smuggle rejected, folded-newline expression rejected, internal-space/scheme-only/prefix-smuggle rejected; two accepted edges are both loud-red-at-runtime class, not skip-green: trailing block-scalar newline (trim), and URL-shaped `postgresql://${X}` (no shell expansion in Actions env mappings ⇒ literal garbage ⇒ connect fails LOUD).
- Scope: this round touches exactly `ci-realdb-step-contract.mjs` + `t2gate-collision-mechanism-ci-wiring.test.mjs`; workflow three-dot diff still exactly the two `id:` lines.

PR body rewritten to the round-6 contract at `df3f871d5`. Pin(a) `${{ }}`-spelling over-strictness remains a separate fail-closed follow-up candidate (out of scope per ticket). UNARMED.

zensgit:
**Independent adversarial gate @ `df3f871d5` (Opus, first independent review of round 6): APPROVE — 0 P1/P2** (report: `/tmp/gate-4496-round6-independent-20260722.md`)

Key proofs beyond the orchestrator's battery:
- **Both-sides reproduction**: old head `f9e9b2e49` accepts `${{ secrets.DOES_NOT_EXIST }}`; `df3f871d5` rejects.
- **Ground-truth premise**: `describeIfDatabase` gates on truthiness only ⇒ the ONLY skip-green path is an empty-at-runtime value; every value the new pin accepts is a literal GitHub sets verbatim ⇒ malformation is loud-red, never skip-green.
- **Monotonic-tightening argument**: round-6 accept-set is a strict SUBSET of round-5's ⇒ no new accept-hole is constructible in pin(b); over-rejection is the only possible error direction (fail-closed).
- **Clause-level load-bearing**: neuter only the `${{` refusal ⇒ MUTATION 9 red (mixed no-space case); neuter only the URL regex ⇒ MUTATION 9 red (non-URL literals); full revert ⇒ exactly 1/69 red. Both clauses independently load-bearing.
- **19-case attack battery** (alias→expression indirection, folded-scalar `${{` reconstruction, `!!str`-tagged expression, homoglyph/zero-width `$`, percent-encoded scheme, numeric/bool-typed values, CRLF): all rejected or accept-but-loud-red class; PyYAML-vs-GitHub divergence dismissed as a bypass path.
- Wiring: the battery runs at `plugin-tests.yml:227` inside the required no-DB `test` job — not a phantom gate. CI 12/12 green at head.

**Corrections absorbed from the gate's NITs**: PR title updated (round 5 → rounds 5-6); line-ref correction for the earlier comment — the id-located real-DB steps are `plugin-tests.yml:419` (multitable) / `:656` (approval) with their `DATABASE_URL` literals at `:422` / `:663`; the `:275/:319` cited earlier are other steps carrying the same literal value (guard binds by exact `id:`, unaffected). Third NIT (legit secret-built URL refused = fail-closed over-strictness) stays a documented follow-up candidate per ticket scope.

UNARMED — owner review pending. Gate chain for this PR is now complete: round-5 gate APPROVE_WITH_HOLDS (absorbed) → round-6 orchestrator verification → round-6 independent gate APPROVE.

zensgit:
## Rebase onto current main — evidence (head `4b0ea18d8`, 2026-07-22)

**Owner order step 4.** Rebased `df3f871d5` (8 commits) onto `origin/main` `f55d99e12` — clean, no conflict hunks; the workflow UNION was then verified by hand, not assumed.

### Workflow UNION (plugin-tests.yml)

Main-side additions since the old merge-base `709d1e37e` all present at this head:
- DT-CLOSE-01C ARG_MAX suite steps — `:102`, `:263`
- stock-preparation RCA-window tests (ps51 + pwsh) — `:132`, `:134`, `:275`, `:276`
- W4-PRE-1/1b/1c/1d + W4-0 attendance real-DB wiring (15 new `tests/integration/attendance-*.db.test.ts` args, e.g. `attendance-setup-readiness-w4-0.db.test.ts` at `:845`)

Branch-side delta intact — three-dot diff vs `origin/main` for the workflow is **exactly two `+` lines**:
```
+        id: multitable-real-db-integration
+        id: approval-real-db-integration
```
(now at `:445` / `:682`; previously `:419` / `:656` — moved by main-side steps above them.)

### Four-pin re-verification on CURRENT workflow content

Both id-located steps re-checked at this head:
- `if: matrix.node-version == '20.x'` — `:446` / `:687`
- literal `DATABASE_URL: postgresql://postgres@localhost:5432/metasheet_test` — `:448` / `:689` (previously `:422` / `:663`)
- `requireExecutableRealDbStep(wf, id)` (all four pins jointly) → PASS for both ids; whole-file vitest args resolved: multitable **226**, approval **60**.

### Full sweep re-run (this head)

```
node --test scripts/ops/*-ci-wiring.test.mjs
→ tests 69 · pass 69 · fail 0 · cancelled 0 · skipped 0   (15 files)
```

No guard reddened across the rebase — the guards bind by `id:` and parsed structure, so moved line numbers had no effect, as designed. No test/guard-side changes were needed.

### Scope check

`git diff origin/main...HEAD --name-status` → the same 16 `scripts/ops/` files (15 guards + `ci-realdb-step-contract.mjs` added) + `plugin-tests.yml` (two `id:` lines). Nothing else.

Pushed `--force-with-lease` (lease pinned to old head `df3f871d5`; remote confirmed unmoved pre-push). PR body head refs re-derived. **UNARMED — no auto-merge, not merged; landing order 4496 → 4500 is orchestrator-controlled.**

zensgit:
**Conflict-resolution record @ `d8d9d18f8`** (update-branch hit conflicts after #4547 landed)

Conflict = `scripts/ops/t2-source-freeze-ci-wiring.test.mjs` only (#4547 extended it to a two-suite FILES array with plain full-file regex checks; this branch had migrated it to the shared id-located executability contract). Resolved as the **strict union**: main's two-suite FILES array (source-freeze + new lock-correctness suite) × this branch's shared-helper contract (exact-`id:` step location + executability pins + negative multitable-placement + on-disk existence, now per file). `plugin-tests.yml` auto-merged (#4547's new suite wiring at :733 + this branch's two `id:` lines at :445/:682); three-dot workflow diff vs main re-verified = exactly the two `id:` lines.

Sweep after union: **72/72** (69→72: the t2 guard now runs 3 checks × 2 suites). Marker sweep clean (0 UU, `git diff --check` clean). Merge commit (not rebase) — intermediate-commit replay conflicts were spurious; squash-merge flattens anyway. UNARMED discipline unchanged — auto-merge armed per owner's landing order, fresh checks running.

PR #4497, opened 2026-07-20
fix(automation): durable dispatch loop — no same-tick re-claim + lease-guarded terminal writes (owner P1); retry-tick spec isolation

zensgit:
> ## ⚠️ Round 3 — a **P1 product fix** was added to this PR (the CI red was NOT a flake)
>
> The reviewer called the earlier CI red on `multitable-automation-dispatch-loop-realdb.test.ts` → *"P2-C: a heartbeat renew DB error is surfaced… "* (`expect(res.heartbeatErrors).toBe(1)` observed **2**, job 88528503071 on `4e04af582`) a **timing flake**. The owner corrected that: `heartbeatErrors` increments **at most once per `processClaimedRow()`**, and the test seeds **one** UUID-isolated row — so **2 means the same row was claimed and executed twice inside one tick**. For a real adapter that is a **duplicate external operation**; a green re-run only means that run missed the window.
>
> **Root cause.** After a heartbeat renew failure the row is left `in_progress` on its **original** lease. A handler that outruns that lease (800 ms handler vs a 1 s lease, under load) makes the row reclaimable *by the same tick's next slot*. Backoff (`MIN_RETRY_BASE_MS`) covered the RESCHEDULED case; the non-resolving dispositions (heartbeat DB error, mid-flight lease loss) had no cover. Measured pre-fix on a load-free construction (1300 ms handler vs 1000 ms lease): **one row → 8 adapter calls, `heartbeatErrors=8`, then burned to `dead_letter` by attempt exhaustion** (`claimTimePoisoned=1`). CI merely caught it at 2.
>
> **Fix (2 parts, no public-contract or flag change).**
> 1. `ClaimOptions.excludeRows` — the tick records every `(outbox_id, consumer_key)` it has claimed and excludes those pairs from later slots' claims. The claim **scans past** an excluded row (no head-of-line block); the list is per-tick/per-caller, so cross-worker concurrency is untouched.
> 2. `completeConsumer`/`poisonConsumer` are **lease-guarded** (`AND lease_expires_at > $asOf`, the caller's own clock — the one that stamped the lease at claim) *in addition to* the fence-CAS. Fence-CAS alone only catches a competitor that **already** reclaimed; it left the window where our lease expired but nobody reclaimed yet, in which a handler result produced **without ownership** was applied. Same shape as directory-sync's completion write (guarded terminal UPDATE, abort on miss). `rescheduleConsumer` is deliberately **not** guarded: it is the non-terminal *safe* outcome.
>
> **`expect(res.heartbeatErrors).toBe(1)` and every other existing assertion are unchanged** — they pass because the row is processed **once**, not because anything was relaxed. Full evidence (deterministic reproductions, mutation signatures, positive controls) is in the round-3 comment below.
>
> The PR's original scope (B2 retry-tick test isolation, below) is unchanged and still stands.

## What / why

Audit finding **B2** (P2, test-isolation): `packages/core-backend/tests/integration/multitable-webhook-retry-tick.test.ts` asserted on the **return value** of `WebhookService.retryFailedDeliveries()` — `expect(retried).toBe(1)` (×2), `toBe(0)`, `expect(a + b).toBe(1)` — which is a count of rows claimed across the **whole database**, not this suite's rows.

The S5 batch added the first-attempt stray-recovery leg to `src/multitable/webhook-service.ts` (`next_retry_at IS NULL AND created_at <= now() - FIRST_ATTEMPT_STRAY_GRACE_MS`). That made every abandoned first-attempt delivery row from **any other suite** claimable, so on the shared CI Postgres this spec:

1. goes **red** on rows it does not own, and
2. its tick **delivers and mutates** those foreign rows (`status = success`).

Constructed proof on the dev DB (one foreign webhook + one pending delivery, `next_retry_at NULL`, `created_at = now() - 10 min`, belonging to a **different** webhook):

```
× a pending delivery past next_retry_at is picked up by the scheduled tick path
  → expected 2 to be 1 // Object.is equality
foreign row after the run: {"status":"success","attempt_count":1,"http_status":200}
```

## What changed (test-only)

`packages/core-backend/tests/integration/multitable-webhook-retry-tick.test.ts`
- **Every assertion is now scoped to rows this file created.** Per-run `TS`-suffixed sink URLs (`https://sink.test/retry/${TS}`, `.../policy/${TS}`) make the captured-POST count mean *our* webhook, so `postsTo(SINK_URL)).toHaveLength(1)` is the scoped replacement for `expect(retried).toBe(1)`, `toHaveLength(0)` for `toBe(0)`, and `toHaveLength(1)` for `a + b === 1`. The tick's global return value is no longer asserted anywhere.
- **The suite deliberately seeds FOREIGN claimable rows.** `beforeAll` creates a separate foreign webhook + a claimable stray (`pending`, `next_retry_at NULL`, created 10 min ago), and every work-doing test seeds another before it runs — so the file *proves* it is green with unrelated pending/stray rows present. No assertion depends on a foreign row's fate.
- Extra scoped assertions that came for free: the not-yet-due row's `next_retry_at` is still its seeded future value (not the claim lease), and the second-tick test now checks all four of this suite's rows are unmoved.
- A file-header **ISOLATION CONTRACT** block documents the rule for the next editor.

`packages/core-backend/vitest.config.ts` (P3, two-point wiring)
- Added `multitable-webhook-retry-tick.test.ts` and `multitable-webhook-durable-dedup.test.ts` to the no-DB `exclude` list. Both are `describeIfDatabase`-gated and were **already** named in the `Run multitable real-DB integration` step of `.github/workflows/plugin-tests.yml` (lines 605-606, verified — no workflow change needed), but were missing from the exclude list, so the no-DB `test (20.x)` job collected and skip-greened them. Adding the exclude removes them from **no** job where they provide value: they only ever skipped there.

## Production semantics: unchanged

`retryFailedDeliveries()` still claims **process-wide** — every due `pending` row in the database, whatever webhook owns it. That is the correct semantic for a leader tick (drain the queue, not a slice of it) and `src/multitable/webhook-service.ts` is **not touched** by this PR.

**Reported, not asserted:** the production tick *does* claim, lease and deliver the foreign stray, marking it `status=success, attempt_count=1, http_status=200` (observed above). That is inherent to the process-wide claim on a shared database. Whether the tick should ever be scoped is a **separate ticket** — deliberately not done here.

## Mutation signatures

All run against real Postgres with an **externally seeded foreign stray row** present.

| # | Mutation | Signature |
|---|---|---|
| **M1** | Revert the scoping: restore `const retried = await scheduler.tick(); expect(retried).toBe(1)` in the due-row test | **RED** — `× a pending delivery past next_retry_at is picked up by the scheduled tick path → expected 3 to be 1` (1 failed / 5 passed). Restored → 6/6 green. *This is the original defect.* |
| **M2** | `webhook-service.ts`: disable the stray leg (`created_at <= new Date(0)`) | **RED ×2** — `× FIRST-ATTEMPT stray recovery → expected [] to have a length of 1 but got +0`; `× a second tick does no work ON THIS SUITE'S ROWS → expected 'pending' to be 'success'` |
| **M5** | `webhook-service.ts`: claim not-yet-due rows (`next_retry_at <= now + 24h`) | **RED ×2** — `× a pending delivery past next_retry_at… → expected [ …(2) ] to have a length of 1 but got 2`; `× a second tick does no work… → expected 'success' to be 'pending'` |
| **M6** | `webhook-service.ts`: `FIRST_ATTEMPT_STRAY_GRACE_MS = 0` (live first attempt double-fired) | **RED ×2** — `× FIRST-ATTEMPT stray recovery → expected [ …(2) ] to have a length of 1 but got 2`; `× a second tick does no work… → expected 'success' to be 'pending'` |

M2/M5/M6 show the scoped per-URL POST counts are **not weaker in substance** than the old global count: they still catch under-claiming, over-claiming and double-firing.

**Negative result, reported honestly (M3):** disabling the claim lease (`if (false && due.length > 0)`, so no `next_retry_at` lease is stamped) leaves the suite **6/6 green** — and it also leaves the **origin/main** version of the suite 6/6 green. The MINOR-1 concurrency test is carried entirely by `FOR UPDATE SKIP LOCKED` (the two passes overlap, so one skips the locked row); the lease only matters for the sequential-across-ticks case, which no test constructs. Pre-existing gap, identical before and after this PR — flagged, not fixed here.

## Two-point wiring proof

- **With the fix**, default (no-DB) config: `npx vitest run tests/integration/multitable-webhook-retry-tick.test.ts tests/integration/multitable-webhook-durable-dedup.test.ts` → `No test files found, exiting with code 1`.
- **Positive control**, same command on `origin/main`'s `vitest.config.ts`: `Test Files 2 skipped (2) · Tests 10 skipped (10)`, exit 0 — i.e. exactly the skip-green the exclude closes.
- Real-DB step already names both files (`plugin-tests.yml` L605-606) — unchanged.

## Test counts

- `multitable-webhook-retry-tick.test.ts` — **6/6 green** (real DB), including with an externally seeded foreign stray present.
- Co-run `retry-tick` + `durable-dedup` — **10/10 green**, repeated 3× (real DB).
- `npx tsc --noEmit` in `packages/core-backend` — clean.
- Local Node is v25; **CI authority is Node 20** — not blocking.

## Sibling observation (not fixed here)

One co-run of the two files transiently failed `durable-dedup` G1/G4 with `expected 2 to be 1`. Re-run **3× solo and 3× co-run: green**, and the failure reproduced on neither. Root cause is a shared-dev-DB collision with a parallel lane, not this change — but it exposes the *same class* of defect in `multitable-webhook-durable-dedup.test.ts`: its `sends.length` is a **global** send count that assumes exactly one active `record.updated` webhook exists in the entire database. Worth its own ticket.

Second, same class: `tests/integration/multitable-webhook-event-bridge.test.ts` is also `describeIfDatabase`-gated and named in the real-DB step (`plugin-tests.yml` L604) but is **likewise absent** from `vitest.config.ts`'s exclude list — the identical half-satisfied two-point wiring. Left out of this PR because the B2 ticket names only the two suites above; flagged here so it is not lost.

---

**UNARMED — owner review pending.** Auto-merge is not enabled; do not merge without owner review.


zensgit:
## Round 3 — owner finding closed: the CI red was a REAL same-tick re-claim bug

**Owner finding.** `heartbeatErrors` increments at most **1** per `processClaimedRow()`. The P2-C test seeds **one** UUID-isolated row yet observed **2** (job `88528503071`, head `4e04af582`) — so the **same row was claimed and executed twice within the same tick**. Cause: after a heartbeat renew failure the row stays `in_progress`; under load the ~800 ms handler crosses the 1 s lease, so the **next slot re-claims it**. For a real adapter that is a duplicate external operation. A green re-run only means that run missed the window. The assertion may **not** be weakened.

### How it is closed

Two product fixes in `packages/core-backend/src/multitable/` — no change to the loop's public contract or flags.

**1. `ClaimOptions.excludeRows` (`automation-durable-dispatcher.ts`) + per-tick registration (`automation-durable-dispatch-loop.ts`).** `runDispatchTick` records every `(outbox_id, consumer_key)` it claims, *before* dispatch, and passes them to every later slot's claim; the claim CTE excludes them with `NOT EXISTS (SELECT 1 FROM unnest($6::text[],$7::text[]) AS x(oid,ckey) WHERE …)`. Backoff (`MIN_RETRY_BASE_MS`) already covered a **rescheduled** row — the doctrine "a retry is a future-tick event, never an intra-tick spin" — but the **non-resolving** dispositions (heartbeat DB error, mid-flight lease loss) leave the row `in_progress` on its **original** lease, and had no such cover. The claim **scans past** an excluded row (not a head-of-line block), the list is per-tick/per-caller, and cross-worker concurrency is untouched. A malformed entry **throws** rather than being skipped (silently dropping one would silently re-open the hole).

**2. Lease-guarded terminal write (`completeConsumer` / `poisonConsumer`).** Both now require `AND lease_expires_at > $asOf` on top of the fence-CAS — `$asOf` is the caller's own clock, the same one that stamped the lease at claim, so there is no app/DB skew. Fence-CAS alone only catches a competitor that **already** reclaimed; it says nothing about the window where our lease **expired but nobody reclaimed yet**, in which the old code applied a handler result produced without ownership. This is the pattern the ticket named — directory-sync's completion write guards its terminal UPDATE (`AND status='running'`) and aborts on a miss (`DirectorySyncLeaseLostError`); here a miss is simply `false` → the row is left `in_progress` for a legitimate reclaim and counted as `lostLease`. `rescheduleConsumer` is deliberately **left unguarded**: it is the non-terminal **safe** outcome (keeps the row retryable rather than applying a result), and refusing it would only drop the backoff and hand the row straight back to a reclaim.

### Red signature — the owner's mechanism, replayed

**(a) The load-based reproduction is unreliable — as the owner said.** The unmodified P2-C test, pre-fix (exclusion reverted), 6 concurrent runs + 4 CPU burners: **6/6 green**. A green run proves nothing; that is why a deterministic construction was required.

**(b) Same scenario, load dependence removed** (1300 ms handler vs a 1000 ms lease — it *always* crosses). Pre-fix, on the **verbatim** P2-C assertions, one seeded row:

```
PROBE {"calls":8,"renewCalls":24,"heartbeatErrorsSeen":24,
       "res":{"claimed":8,"completed":0,"lostLease":0,"heartbeatErrors":8,"claimTimePoisoned":1}}
× expected 8 to be 1 // Object.is equality        ← expect(res.heartbeatErrors).toBe(1)
```

One row, **8 deliveries in one tick**, then burned to `dead_letter` by attempt exhaustion. CI merely caught it at 2. Post-fix, same construction: `{"calls":1,"heartbeatErrors":1,"claimTimePoisoned":0}` — **green**. This is now a committed test (`P2-C (deterministic): a handler that outruns its lease is still delivered ONCE per tick`).

**(c) Fully deterministic repro (no sleep, no load dependence)** — committed as `P1 same-tick re-claim: a row this tick already dispatched is never claimed again by a later slot`:
- an **injected clock** (`opts.now`) is advanced past the lease *inside* the handler, so "the lease expired" is a fact of the claim's own `asOf`, not a wall-clock race;
- the handler is **gated** on the first renew having actually **thrown** (the gate is opened from `onHeartbeatError`), so the heartbeat-error ordering is fixed rather than raced;
- the renew failure is scoped to row **A** only, and row **B** behind it is the positive control.

### Mutation proofs (each an independent revert; committed before, restored by explicit path)

| # | Mutation | Signature |
|---|---|---|
| **M1** | revert **only** the claim exclusion (`excludeRows: claimedThisTick`), keep the lease guard | **RED** — `× P1 same-tick re-claim … → expected 8 to be 1` (`callsA`: the tick re-claimed and re-ran row A until the attempt ceiling) |
| **M2** | neuter **only** the lease guard on `completeConsumer` | **RED ×2** — `× P1 lease guard: a completion attempted after the lease expired is refused → expected +0 to be 1` (`lostLease`); `× LEASE-GUARD: complete/poison … → expected true to be false` |
| **M3** | neuter **only** the lease guard on `poisonConsumer` | **RED** — `× LEASE-GUARD: complete/poison are refused once the lease expired → expected true to be false` |

### Positive controls (the fix is not "reject everything")

- **same-tick re-claim test:** row **B**, queued behind the excluded row A, is still claimed and delivered in the same tick — `callsB=1`, `res.completed=1`, B `done`. Excluding A is not a head-of-line block.
- **lease-guard test (loop level):** sibling row **Y**, still inside its lease, completes normally in the same tick — `completed=1` alongside `lostLease=1`.
- **lease-guard test (S2-a primitive):** after a **fresh reclaim** (lease actually held) `completeConsumer` → `true` / `poisonConsumer` → `true`, and the rows reach `done` / `dead_letter`. The guard refuses a write we no longer own, not every write.
- **`excludeRows` primitive:** the very same claim **without** the exclusion *does* re-claim the row (`attempts` 1→2) — the exclusion is load-bearing, and the exclusion call still returns the next row.

### Assertions kept exactly as they were

`expect(res.heartbeatErrors).toBe(1)`, `expect(res.lostLease).toBe(0)`, `expect(renewCalls).toBeGreaterThan(0)`, `expect(heartbeatErrorsSeen).toBeGreaterThan(0)` and the `in_progress` read-back in P2-C are **untouched** — they pass because the row is processed **once**. Nothing was relaxed to `toBeGreaterThan(0)`, and the injection was not narrowed to "fail only once".

### Test results (real DB, this head)

- `multitable-automation-dispatch-loop-realdb.test.ts` — **37/37** green (was 34; +3 new). Repeated **5× solo** and **3× concurrently** (contention stress for the lease guard's margin — `P3-1b` / `P1-B` are the sensitive ones): green every time.
- Same real-DB step, co-run: dispatch-loop + dispatcher-claim + durable-activation + outbox-enqueue + outbox-schema + durable-startup-failclosed + **webhook-retry-tick** (this PR's original scope) + webhook-durable-dedup + webhook-event-bridge → **108/108 green (9 files)**.
- `tests/unit/automation-durable-consumer-handlers.test.ts` — 9/9 green.
- `npx tsc --noEmit` (`packages/core-backend`) clean; `eslint` clean on both changed `src` files.
- Local Node **v25**; **CI authority is Node 20** — noted, not blocking.

### Scope note

The PR's original B2 scope (retry-tick test isolation + two-point wiring) is unchanged and still stands; the P1 product fix is additive. The PR description has been updated with a header stating that a P1 product fix was added and why.

**New exact head: `24081d564a5628c4f9c99545e6c6df9c8c724293`**

**UNARMED** — auto-merge is not enabled and was not touched; do not merge without owner review.

PR #4498, opened 2026-07-20
fix(directory): B1 — a transfer freeze is a deliberate abort, not a sync failure

zensgit:
## What / Why

**Audit finding B1 — a deliberate freeze was recorded as an integration failure.**

When a transfer freeze linearizes first during an in-flight sync, the apply correctly rolls back and `DirectorySyncFrozenByTransferError` propagates into the **unconditional** catch at the end of `syncDirectoryIntegration`, which called `markSyncFailure`. Aftermath observed on `main` (reproduced by extending the landed create-wins race test):

```
runStatuses:          ["failed"]
integrationLastError: "Directory sync is frozen by an active org transfer for this integration (transfer <uuid>)"
alerts:               [{ level: "error", code: "sync_failed", message: same }]
```

That contradicts this lane's own semantics: a frozen source is a **deliberate** state everywhere else — the scheduler path is a quiet info skip (`directory-sync-scheduler.ts`), the route path maps it to a deliberate 409 that must "never page monitoring". And it is not self-healing: `directory_integrations.last_error` is only cleared inside the **successful** apply transaction — and the freeze is exactly what prevents a successful apply — so the integration reads as errored for the **entire** transfer (the design's documented multi-day steady state), while also feeding `countConsecutiveFailedRuns` escalation once the alert webhook is enabled (pending owner decision).

The audit agent had already flagged this code path as an untested guard (MUT10: skipping `markSyncFailure` on the frozen error left 32/32 green).

## Change (backend only, 2 files)

`packages/core-backend/src/directory/directory-sync.ts`

1. Discriminate `DirectorySyncFrozenByTransferError` in that catch and take a terminal **deliberate-abort** path (`markSyncAbortedByFreeze`) instead of `markSyncFailure`:
   - run reaches a terminal, non-`running` state → the partial unique index frees the lease immediately, nothing is wedged, no waiting on stale-lease reclaim;
   - `error_message` stays **NULL** (the admin run panel styles that field as an error); the reason lives in `meta` (`abortReason: 'frozen_by_org_transfer'`, `transferId`), which no run-summary/toast surface reads;
   - `directory_integrations` is **not touched at all** — no `last_error`, no `last_sync_at` bump (nothing was synced);
   - **no** `directory_sync_alerts` `'sync_failed'` row, and therefore no failure-alert delivery.
2. **Run status = `'aborted'`** (the ticket's preferred option; no migration). Verified before choosing:
   - no CHECK constraint on `directory_sync_runs.status` (bare `text`, confirmed both in `zzzz20260324150000_create_directory_sync_tables.ts` and against a live DB via `pg_constraint`);
   - **every** reader of `directory_sync_runs.status` across `src/` and `apps/web` was grepped:
     | reader | behaviour with `'aborted'` |
     |---|---|
     | `countConsecutiveFailedRuns` (alert-delivery:57-67) — `status IN ('completed','failed')` | excluded from escalation ✅ (this is why `'aborted'` needs no extra code) |
     | `reclaimStaleDirectorySyncRuns` — `WHERE status='running'` | excluded (row already terminal) ✅ |
     | `findActiveDirectorySyncRunId` — `WHERE status='running'` | excluded ✅ |
     | partial unique index `uq_directory_sync_runs_one_running_per_integration` | lease freed ✅ |
     | completion write — `WHERE status='running'` | n/a ✅ |
     | `listDirectorySyncRuns`, `getDirectorySyncScheduleSnapshot`, `last_run_status` sub-select | unfiltered text passthrough ✅ |
     | `apps/web` `DirectoryManagementView.vue` — run panel (`{{ run.status }}`), integration card (`lastRunStatus`) | renders the raw value, exactly as it does for `completed`/`failed` today ✅ |
   - **~~One known cosmetic wrinkle, deliberately not fixed here~~ — SUPERSEDED, now fixed in this PR.** The owner reviewed this and ruled it is NOT cosmetic: the async-poll path at `DirectoryManagementView.vue` sent every "not running, not failed, no errorMessage" state into `buildAsyncRunSummary()`, so an aborted run told the operator 后台同步已完成 — the user will not retry on their own and the message never corrects itself. The view now handles `aborted` explicitly, the success branch is reachable ONLY by `completed`, and any unknown terminal state fails safe. Covered by a polling web test and an unknown-terminal fail-safe test, both wired into the required web-tests run-list; reverting the view reds the polling test and neutering the fail-safe reds the unknown-state test.
3. `markSyncAbortedByFreeze`'s UPDATE **is** guarded on `status = 'running'` (unlike `markSyncFailure`, which deliberately has no guard): if the lease was already reclaimed, the row carries the reclaimer's truthful `failed`/orphaned record and this run no longer owns it — overwriting it with a softer `'aborted'` would erase a real liveness incident. Either way the row is terminal and the lease is free. Documented at the call site.
4. The thrown error and the **409 route / scheduler mappings are unchanged** — `throw error` re-raises the original instance. Tests E (route 409, both sync and async branches) and the scheduler unit suite stay green untouched.

## Proof (`tests/integration/directory-org-transfer-source-freeze.db.test.ts`)

Extended the landed `RACE create-wins` test (the one that actually drives the freeze-wins linearization through the real sync) to assert the full aftermath, plus two new cases:

- **create-wins race now asserts**: no `completed` run, no `failed` run, exactly one `aborted` run with `finished_at` set and `error_message` NULL, `meta.abortReason='frozen_by_org_transfer'` + `meta.transferId`, `integrations.last_error` NULL, **zero** `sync_failed` alert rows for that integration, `countConsecutiveFailedRuns() === 0`, and — end-to-end — **the lease is free**: lifting the freeze and calling `syncDirectoryIntegration` again yields `run.status === 'completed'` and the empty-tenant absence sweep deactivates the seeded account (so the follow-up sync is proven non-vacuous).
- **`B1 POSITIVE CONTROL`** (new): a genuine failure (provider pull throws) still marks the run `failed`, writes `integrations.last_error`, and creates exactly one `sync_failed` alert — proving the catch was narrowed, not gutted.
- **`B1 escalation exclusion`** (new): pins the escalation input directly — genuine failure → deliberate abort row → genuine failure gives `countConsecutiveFailedRuns() === 2`; the abort neither counts toward the streak nor resets it.

All fixtures/assertions are scoped by the suite's own unique integration ids (shared dev DB).

## Mutation signatures (each applied, red confirmed, restored, green confirmed)

| mutation | result |
|---|---|
| revert the discrimination — delete the `if (error instanceof DirectorySyncFrozenByTransferError)` branch so the frozen error routes back through `markSyncFailure` | **1 failed / 17 passed** — `RACE create-wins` → `AssertionError: expected { …(5) } to be undefined` at `directory-org-transfer-source-freeze.db.test.ts:534` (`expect(runs.find((r) => r.status === 'failed')).toBeUndefined()`) |
| break the positive control's path — suppress `markSyncFailure` for **all** errors | **2 failed / 16 passed** — `B1 POSITIVE CONTROL` → `AssertionError: expected 'running' to be 'failed'` at `:571`; `B1 escalation exclusion` → `AssertionError: expected +0 to be 1` |
| (restored) | **18 passed / 18** |

## Test counts

- `directory-org-transfer-source-freeze.db.test.ts`: **18 passed** (16 on `main` + 2 new; the create-wins case extended in place).
- Neighbouring real-DB suites, unchanged and green: `directory-sync-run-lease` + `directory-sync-alert-coverage` + `directory-sync-orchestration` → **30 passed**.
- Unit: `directory-sync-alert-delivery` + `directory-sync-scheduler` + `admin-directory-routes` → **137 passed**.
- Ops guard: `node --test scripts/ops/t2-source-freeze-ci-wiring.test.mjs` → **2/2 pass** (suite still excluded from the no-DB job and wired whole-file into the real-DB step).
- `npx tsc --noEmit` in `packages/core-backend`: clean.
- `git diff origin/main --name-only` lists exactly the two ticket files.
- Local Node is v25.9.0; **CI authority is Node 20** — not blocking, just noted.

## Rebase replay (2026-07-22, owner order step 2)

Owner APPROVEd the content on the old green; per order the proofs were replayed on current `main`.

- Head: `b3dc0d9e1` → **`a6536c54d`** (rebased onto `origin/main` = `f55d99e12`, the W4 wave incl. W4-PRE-1a..1d lifecycle). Rebase was conflict-free; marker sweep clean (0 UU, `git diff --check` clean, anchor grep empty). Three-dot scope vs `origin/main` is exactly the same 4 files.
- Replay on `a6536c54d`: freeze suite **19/19** (the 19th is the gate-round reclaimer-barrier test), guard-removal mutation reds it with the exact `expected 'aborted' to be 'failed'` signature, web spec **70/70**, `tsc --noEmit` + `vue-tsc -b` clean.
- `markSyncFailure` caller census on the rebased head: still exactly **one** call site (`directory-sync.ts:4359`), behind the freeze discrimination (`:4354-4356`); W4-PRE-1's diff adds **zero** catch clauses — no new unconditional failure route. Full classification in the replay comment below.

**UNARMED — owner review pending.** No auto-merge, no merge.

### DIFF
--- a/apps/web/src/views/DirectoryManagementView.vue
+++ b/apps/web/src/views/DirectoryManagementView.vue
@@ -4555,10 +4555,30 @@ async function pollAsyncRun(integrationId: string, runId: string, token: number)
   const run = items.find((item) => item.id === runId)
   if (run && run.status !== 'running') {
     stopRunPolling()
+    // DISPLAY TRUTH (B1 rework): the success summary is an ALLOWLIST — reachable only by an
+    // explicit 'completed'. Every other terminal value, including ones this build has never
+    // heard of, must fall through to the fail-safe `else` and say the sync did NOT complete.
+    // Blocklisting the known-bad states instead ("not running, not failed, no errorMessage
+    // ⇒ 完成") is exactly how the deliberate 'aborted' state rendered as SUCCESS: the admin
+    // is not going to retry on their own, and the message never corrects itself.
     if (run.status === 'failed' || run.errorMessage) {
       setStatus(`后台同步失败（运行 ${runId}）：${run.errorMessage || '未知错误'}`, 'error')
-    } else {
+    } else if (run.status === 'aborted') {
+      // 'aborted' is written by exactly one backend path — markSyncAbortedByFreeze, i.e. the
+      // apply-time recheck found an active org transfer and rolled the apply back (deliberate
+      // state, no alert, error_message stays NULL). The reason lives in the run's `meta`, which
+      // this runs payload does not carry, so the wording is derived from the status alone.
+      setStatus(
+        `后台同步已中止（运行 ${runId}）：该集成的来源正被进行中的组织迁移冻结，本次同步未写入任何数据。请先完成或取消该迁移，然后重新发起同步。`,
+        'error',
+      )
+    } else if (run.status === 'completed') {
       setStatus(buildAsyncRunSummary(run))
+    } else {
+      setStatus(
+        `后台同步以未识别的终态结束（运行 ${runId}，状态 ${run.status}），无法确认是否成功，请前往下方运行记录核对后再决定是否重试。`,
+        'error',
+      )
     }
     await refreshAfterDirectorySync(integrationId)
     return
--- a/apps/web/tests/directoryManagementView.spec.ts
+++ b/apps/web/tests/directoryManagementView.spec.ts
@@ -7413,6 +7413,98 @@ describe('DirectoryManagementView', () => {
       }
     })
 
+    // B1 rework (owner P2 on #4498): the backend now ends a freeze-aborted run as
+    // status='aborted' with error_message NULL. The old terminal branch was a BLOCKLIST
+    // ("not running, not failed, no errorMessage ⇒ 后台同步已完成"), so this deliberate
+    // abort — nothing synced, the source is frozen by an active org transfer — was shown
+    // to the admin as SUCCESS. Nobody retries a success, and the message never self-corrects.
+    it('stops polling at terminal aborted and says the sync was ABORTED by an active org transfer — never the success line', async () => {
+      vi.useFakeTimers()
+      try {
+        mockInitialLoad(createIntegration())
+        apiFetchMock.mockResolvedValueOnce(createJsonResponse(
+          { ok: true, data: { accepted: true, runId: 'run-bg-1', integrationId: 'dir-1' } },
+          202,
+        ))
+        apiFetchMock.mockResolvedValueOnce(createJsonResponse(createRunsPayload([createBackgroundRun('running')])))
+        // Exactly what markSyncAbortedByFreeze leaves behind: 'aborted', errorMessage NULL,
+        // no stats (the apply rolled back). The abortReason/transferId live in the run's
+        // `meta`, which the /runs payload this view reads does not carry.
+        apiFetchMock.mockResolvedValueOnce(createJsonResponse(createRunsPayload([
+          createBackgroundRun('aborted'),
+        ])))
+        mockPostSyncRefresh(createRunsPayload([createBackgroundRun('aborted')]))
+
+        mountView()
+        await flushUi()
+
+        findButton('后台同步')!.dispatchEvent(new MouseEvent('click', { bubbles: true }))
+        await flushUi(8)
+
+        await vi.advanceTimersByTimeAsync(5000)
+        await flushUi(8)
+
+        // The surfaced message tells the operator (a) it was aborted, (b) why — an active org
+        // transfer froze the source, (c) that nothing was written, (d) the resolution.
+        expect(container?.textContent).toContain('后台同步已中止（运行 run-bg-1）')
+        expect(container?.textContent).toContain('正被进行中的组织迁移冻结')
+        expect(container?.textContent).toContain('未写入任何数据')
+        expect(container?.textContent).toContain('请先完成或取消该迁移')
+        // THE REGRESSION ITSELF: an aborted run must never render as a completed one.
+        expect(container?.textContent).not.toContain('后台同步已完成')
+
+        const settledCallCount = apiFetchMock.mock.calls.length
+        await vi.advanceTimersByTimeAsync(30_000)
+        await flushUi(4)
+        expect(apiFetchMock.mock.calls.length).toBe(settledCallCount)
+      } finally {
+        vi.useRealTimers()
+      }
+    })
+
+    // B1 rework fail-safe: the success branch is an ALLOWLIST on 'completed'. Any terminal
+    // status this build does not recognise — a future backend state, a typo, a replayed old
+    // row — must fall through to the "cannot confirm" line, never to 后台同步已完成.
+    it('an UNRECOGNISED terminal status must not take the success branch (fail-safe default)', async () => {
+      vi.useFakeTimers()
+      try {
+        mockInitialLoad(createIntegration())
+        apiFetchMock.mockResolvedValueOnce(createJsonResponse(
+          { ok: true, data: { accepted: true, runId: 'run-bg-1', integrationId: 'dir-1' } },
+          202,
+        ))
+        apiFetchMock.mockResolvedValueOnce(createJsonResponse(createRunsPayload([createBackgroundRun('running')])))
+        // Deliberately a status no branch knows about, carrying the exact shape the old
+        // blocklist treated as success: not running, not failed, errorMessage NULL.
+        apiFetchMock.mockResolvedValueOnce(createJsonResponse(createRunsPayload([
+          createBackgroundRun('quarantined_by_future_backend'),
+        ])))
+        mockPostSyncRefresh(createRunsPayload([createBackgroundRun('quarantined_by_future_backend')]))
+
+        mountView()
+        await flushUi()
+
+        findButton('后台同步')!.dispatchEvent(new MouseEvent('click', { bubbles: true }))
+        await flushUi(8)
+
+        await vi.advanceTimersByTimeAsync(5000)
+        await flushUi(8)
+
+        // Non-vacuous: assert the fail-safe line is actually surfaced, not merely that the
+        // success line is absent (an empty status would satisfy absence alone).
+        expect(container?.textContent).toContain('后台同步以未识别的终态结束（运行 run-bg-1，状态 quarantined_by_future_backend）')
+        expect(container?.textContent).toContain('无法确认是否成功')
+        expect(container?.textContent).not.toContain('后台同步已完成')
+
+        const settledCallCount = apiFetchMock.mock.calls.length
+        await vi.advanceTimersByTimeAsync(30_000)
+        await flushUi(4)
+        expect(apiFetchMock.mock.calls.length).toBe(settledCallCount)
+      } finally {
+        vi.useRealTimers()
+      }
+    })
+
     it('clears the poll timer on unmount — no fetch ever fires after the view is gone', async () => {
       vi.useFakeTimers()
       try {
--- a/packages/core-backend/src/directory/directory-sync.ts
+++ b/packages/core-backend/src/directory/directory-sync.ts
@@ -3158,6 +3158,55 @@ async function markSyncFailure(integrationId: string, runId: string, message: st
   await deliverDirectorySyncFailureAlert({ integrationId, integrationName, runId, message })
 }
 
+/**
+ * Terminal run state for a DELIBERATE abort: the apply-time freeze recheck found an active
+ * org transfer and rolled the apply back (see `DirectorySyncFrozenByTransferError`).
+ *
+ * A frozen source is a deliberate transfer state, not an outage — every other surface in this
+ * lane already says so (the scheduler skips quietly, the route answers a deliberate 409 that
+ * must never page monitoring). Routing this through `markSyncFailure` contradicted all of them
+ * and, worse, was not self-healing: `directory_integrations.last_error` is only cleared inside
+ * the SUCCESSFUL apply transaction, and the freeze is exactly what prevents a successful apply —
+ * so a multi-day transfer (the designed steady state) left the integration reading "errored"
+ * for its whole duration and fed `countConsecutiveFailedRuns` escalation.
+ *
+ * What this leaves behind instead:
+ *  - the run reaches a TERMINAL, non-'running' state, so the partial unique index frees the
+ *    lease immediately and nothing has to wait for stale-lease reclaim;
+ *  - status is 'aborted', a distinct value: `countConsecutiveFailedRuns` filters
+ *    `status IN ('completed','failed')`, so a deliberate abort can never drive failure
+ *    escalation, and no migration is needed (the column is bare `text`, no CHECK constraint);
+ *  - `error_message` stays NULL — the admin run panel styles that field as an error;
+ *    the reason lives in `meta` (which no summary/toast surface reads);
+ *  - `directory_integrations` is NOT touched at all: no `last_error`, no `last_sync_at` bump
+ *    (nothing was synced);
+ *  - no `directory_sync_alerts` 'sync_failed' row and therefore no failure-alert delivery.
+ *
+ * Unlike `markSyncFailure` this UPDATE IS guarded on `status = 'running'`: if the lease was
+ * reclaimed while this run was alive, the row already carries the reclaimer's truthful
+ * `failed`/orphaned record and this run no longer owns it — overwriting it with a softer
+ * 'aborted' would erase a real liveness incident. Either way the row is terminal and the
+ * lease is free.
+ */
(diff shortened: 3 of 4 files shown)


zensgit:
**Independent gate (run inline — the workflow's gate agent died on an API error twice) — APPROVE with one gap found and closed. New head `7423fd97c`.**

Verified on the exact head, in a disposable worktree off the branch:
- **The fix is load-bearing**: MUT-A (delete the `DirectorySyncFrozenByTransferError` discrimination so the frozen error routes back through `markSyncFailure`) → RED. Restored → green.
- **Consumer sweep re-done independently**: `countConsecutiveFailedRuns` filters `status IN ('completed','failed')` so `aborted` is invisible to escalation; every lease/reclaim/boot-sweep predicate keys on `status='running'` so the lease frees immediately; list/summary paths pass the text through; the web run panel renders `{{ run.status }}` with a `--${status}` modifier class, so a new value degrades to unstyled text rather than breaking. No CHECK constraint on the column, so no migration — as claimed.
- **Suite**: 19/19 (18 + the new pin below), tsc clean, scope still 2 files.

**Gap found and closed in this round** — MUT-B: dropping `AND status = 'running'` from `markSyncAbortedByFreeze` left the whole suite **green**, so the guard the docblock explains at length (never soften a reclaimer's truthful `failed` record into `aborted`) was unpinned. Added a two-connection barrier test: the freeze writer holds the shared source lock, the real sync parks in its apply transaction, a stale-lease reclaimer takes the run row **while it is provably parked** (`pg_blocking_pids`, no sleeps), then the freeze commits and the apply aborts — the reclaimer's row must survive verbatim. The same mutation now REDs with `expected 'aborted' to be 'failed'`.

UNARMED — owner review pending.

zensgit:
## REWORK — owner P2 (web layer) closed. New head `b3dc0d9e1`

Backend untouched: `git diff 7423fd97c..HEAD -- packages/core-backend/src/directory/directory-sync.ts` → **0 lines**. Diff is `apps/web/src/views/DirectoryManagementView.vue` (+22/−1) and `apps/web/tests/directoryManagementView.spec.ts` (+92).

### Owner finding: an `aborted` background sync is shown to the user as SUCCESS

`pollAsyncRun`'s terminal branch was a **blocklist** — "not running, not failed, no `errorMessage`" fell into `buildAsyncRunSummary()`, so `status='aborted'` + `error_message=NULL` rendered as `后台同步已完成`.

**(1) `aborted` handled explicitly; success is now an ALLOWLIST** (`DirectoryManagementView.vue:4556`):

```
if (run.status === 'failed' || run.errorMessage)  → 后台同步失败…            (unchanged, still wins first)
else if (run.status === 'aborted')                → 后台同步已中止（运行 X）：该集成的来源正被进行中的组织迁移冻结，
                                                    本次同步未写入任何数据。请先完成或取消该迁移，然后重新发起同步。
else if (run.status === 'completed')              → buildAsyncRunSummary(run)   ← the ONLY door to success
else                                              → 后台同步以未识别的终态结束（运行 X，状态 Y），无法确认是否成功，
                                                    请前往下方运行记录核对后再决定是否重试。
```

The final `else` is the fail-safe default, so every future/unknown terminal value lands on "did not complete" without any further edit. `refreshAfterDirectorySync` still runs for all terminal states; the pre-flight 409 path, the 10-minute poll bound and `runPollToken` are untouched.

Wording note: `abortReason`/`transferId` are written to the run's `meta`, which the `/runs` payload this view reads does not carry — so the message is derived from `status === 'aborted'` alone, whose sole writer is `markSyncAbortedByFreeze`. No type/endpoint widening. The file has no label module (all sibling status strings are inline Chinese), so the new strings match that convention and the `后台同步…（运行 ${runId}）：…` shape of the sibling failure line.

**(2) Polling test** — `tests/directoryManagementView.spec.ts`, in the existing `Wave 2: 预览同步 / 后台同步 polling` describe, modelled on the `failed` test (fake timers, 202 → running → terminal, full `mockPostSyncRefresh`): *"stops polling at terminal aborted and says the sync was ABORTED by an active org transfer — never the success line"*. Asserts the abort cause, "未写入任何数据", the resolution ("请先完成或取消该迁移"), `not.toContain('后台同步已完成')`, and the no-further-fetch tripwire.

**(3) Unknown-terminal-state fail-safe test** — *"an UNRECOGNISED terminal status must not take the success branch (fail-safe default)"*: drives the poll to `status='quarantined_by_future_backend'` with the exact shape the old blocklist called success (not running, not failed, `errorMessage` NULL). Non-vacuous: asserts the fail-safe line is actually surfaced **and** that the success line is absent.

### Owner reproduction, now RED (mutation proof)

| # | Mutation | Result |
|---|---|---|
| 1 | Revert the whole view change (`git checkout 7423fd97c -- …View.vue`) — `aborted` falls into the success branch | **aborted test RED**, unknown test RED — `2 failed \| 68 passed` |
| 2 | Neuter **only** the final `else` → `setStatus(buildAsyncRunSummary(run))` | **unknown test RED, aborted test GREEN** — `1 failed \| 69 passed` |
| 3 | Neuter **only** the `aborted` branch (`=== 'aborted'` → `=== '__never__'`) | **aborted test RED, unknown test GREEN** — `1 failed \| 69 passed` |

Mutation 1 red signature (the owner's repro verbatim — the rendered DOM contains the success string):
```
AssertionError: expected '目录同步管理…' to contain '后台同步已中止（运行 run-bg-1）'
+ …新建集成后台同步已完成（运行 run-bg-1），账号 0 / 部门 0 / 待确认 0 / 已链接 0…运行记录…aborted 2026/07/10 09:00 → 2026/07/10 09:05
```
Mutation 2 red signature:
```
AssertionError: expected '目录同步管理…' to contain '后台同步以未识别的终态结束（运行 run-bg-1，状态 quaranti…'
```
Mutation 3 red signature:
```
AssertionError: expected '目录同步管理…' to contain '后台同步已中止（运行 run-bg-1）'
```
2 and 3 are independent, which is the point: neither new test is carried by the other. Working tree restored with `git checkout HEAD -- <exact path>` after each; `git status --porcelain` empty before push.

### Positive controls (unchanged behaviour still passes)
- `polls every 5s while running, stops at terminal completed, renders the summary…` — still GREEN: a real `completed` run still gets `后台同步已完成（运行 run-bg-1）`, `本次新增账号 4，更新 95，停用 2`, `自动准入 2 位成员…`.
- `stops polling at terminal failed and surfaces the run errorMessage` — still GREEN.
- The whole spec: **70/70 passed** (68 pre-existing + 2 new).

### CI wiring — how I verified it actually runs
The new tests live in `tests/directoryManagementView.spec.ts`, whose token `directoryManagementView` is **already** in `apps/web/scripts/run-required-web-tests.sh` (the always-on required-eligible gate) — deliberately reusing the sibling-gated file rather than adding an unwired new spec. Verified two ways, because the script's `--reporter=dot` hides test names:
1. `npx vitest run directoryManagementView --reporter=verbose` (the exact CI filter token) prints both new `it(...)` names — so the token selects them.
2. `bash apps/web/scripts/run-required-web-tests.sh` → **349 files / 4147 tests passed**, run twice: local Node v25.9.0 and under CI's authority **Node 20.20.2** (`~/.nvm/versions/node/v20.20.2`), green both times.

`npx vue-tsc -b` in `apps/web`: clean (exit 0).

UNARMED — not merged, no auto-merge.

zensgit:
## Replay on current main (owner order step 2) — head `b3dc0d9e1` → `a6536c54d`

Owner already APPROVEd the content; the old green predates current `main`, so every proof was replayed on the rebased head. **UNARMED** — no merge, no auto-merge; raw data only.

### Rebase

- Base moved: PR merge-base was `709d1e37e`; rebased onto `origin/main` = **`f55d99e12`** (W4 wave: `7a64424d1` W4 re-ratify incl. W4-PRE-1, `b9495af18`/`2365d977a`/`bbcb8caf3`/`f55d99e12` W4-0..2).
- `git rebase origin/main`: **conflict-free** — the anticipated `directory-sync.ts` collision did not materialize as a textual conflict (W4-PRE-1's membership/deprovision hunks land in different regions of the file).
- Marker sweep on the result: `git status --porcelain | grep -c '^UU'` = **0**; `git diff --check origin/main...HEAD` clean; anchor grep for `<<<<<<<`/`>>>>>>>` across `apps/web/{src,tests}` + `packages/core-backend/{src,tests}` = **no matches** (git-grep exit 1).
- Three-dot scope vs `origin/main` unchanged: exactly the same 4 files, `+342/−7` (`DirectoryManagementView.vue`, `directoryManagementView.spec.ts`, `directory-sync.ts`, `directory-org-transfer-source-freeze.db.test.ts`).
- New head: **`a6536c54d`** (3 commits: `123edca42` fix, `41caa524c` gate-round test, `a6536c54d` web fix). Pushed `--force-with-lease` pinned to `b3dc0d9e1` after re-fetch confirmed the remote had not moved.

### (1) Freeze-abort real-DB suite — RUN, not skipped

DB probe first: `psql postgresql://postgres@localhost:5432/metasheet_test -tAc "SELECT 1"` → `1` (exit 0); the suite's own sentinel (`sentinel: DATABASE_URL is set`, spec :259) also passed, so this is a real run, not a skip-green.

```
DATABASE_URL=… vitest --config vitest.integration.config.ts run tests/integration/directory-org-transfer-source-freeze.db.test.ts
Test Files  1 passed (1)   Tests  19 passed (19)
```

19 = the 18 from the original round + the gate-round two-connection barrier test (`B1 abort never softens a RECLAIMED run: a reclaimer-written failed row survives the freeze abort`, spec :604).

**Mutation replay of the barrier proof** (applied → red → restored): deleted `AND status = 'running'` from `markSyncAbortedByFreeze`'s UPDATE (`directory-sync.ts:3202` on the rebased head) →

```
FAIL … > B1 abort never softens a RECLAIMED run …
AssertionError: expected 'aborted' to be 'failed'   at directory-org-transfer-source-freeze.db.test.ts:656
1 failed | 18 skipped
```

— the exact expected red signature (`'aborted'` vs `'failed'`). Restored via targeted single-file checkout; `git status --porcelain` empty and `git diff HEAD` empty, i.e. byte-identical to the committed head that produced the 19/19 green above.

### (2) UI unknown-terminal-state spec

```
apps/web: vitest run tests/directoryManagementView.spec.ts
Test Files  1 passed (1)   Tests  70 passed (70)
```

Includes the aborted-posture test at spec :7421 (`stops polling at terminal aborted and says the sync was ABORTED by an active org transfer — never the success line`). View-side anchor on the rebased head: `DirectoryManagementView.vue:4566` handles `run.status === 'aborted'` explicitly; success line reachable only via `completed`.

### (3) tsc clean

- `packages/core-backend`: `tsc --noEmit` → exit 0.
- `apps/web`: `vue-tsc -b` → exit 0.

### (4) Catch discrimination + `markSyncFailure` caller census on `a6536c54d`

Discrimination intact — `directory-sync.ts:4354-4356`: `if (error instanceof DirectorySyncFrozenByTransferError) { await markSyncAbortedByFreeze(runId, error.transferId); throw error }`. Abort semantics unchanged (`markSyncAbortedByFreeze`, :3191-3207): run → terminal `'aborted'` guarded `AND status = 'running'`, reason in `meta` only, `error_message` stays NULL, `directory_integrations` untouched (no `last_error`), no `sync_failed` alert INSERT, no `deliverDirectorySyncFailureAlert` call. Runtime-proven by the green create-wins race + B1 escalation-exclusion tests in the 19/19 above.

Every `markSyncFailure` caller on the rebased head (repo-wide grep, non-test sources):

| caller | classification |
|---|---|
| `directory-sync.ts:4359` — the **only** call site | CONDITIONAL: inside `syncDirectoryIntegration`'s outer catch, reached only **after** the freeze discrimination at :4354 routes `DirectorySyncFrozenByTransferError` away. This is the PR's fix, carried through the rebase (old main's unconditional call at :3879 no longer exists). |

Every `status = 'failed'` writer on `directory_sync_runs` (only `directory-sync.ts` references the table outside migrations/alert-delivery reads):

| writer | classification |
|---|---|
| `:3128` (body of `markSyncFailure`) | genuine-failure route, guarded by the discrimination above; positive control (spec :560) proves it still fires for real failures. |
| `:3490` (`reclaimStaleDirectorySyncRuns`) | pre-existing stale-lease reclaimer, `WHERE status = 'running' AND heartbeat stale` — the very writer the :604 barrier test pins against the abort UPDATE. Not a W4-PRE-1 addition. |

**W4-PRE-1 did not add a bypass route**: `git diff 709d1e37e..origin/main -- directory-sync.ts` adds **zero** `catch` clauses and exactly one `throw` (`throw new Error('Directory integration not found for account org resolution')` in `resolveDirectoryAccountOrgId`), whose new membership/deprovision paths (`upsertActiveUserOrgMembership`, `deactivateUserOrgMembershipIfNoOtherActiveBinding`, `applyDirectoryDeprovisionPolicies` extensions) all execute inside the sync try — any error they raise flows to the same discriminating catch, i.e. genuine failures still reach `markSyncFailure`, and the freeze error still cannot.

PR #4505, opened 2026-07-21
fix(ops): DT-CLOSE-01C — OAuth-stability recording rc=126 (E2BIG); pass scrape payloads by file + ARG_MAX behaviour test

zensgit:
## UNARMED — owner review pending. Do not auto-merge.

Fixes the continuously-failing scheduled **DingTalk OAuth Stability Recording (Lite)** workflow.

## The real failure (verified, not assumed)

Every scheduled run has been failing. `gh run list --workflow dingtalk-oauth-stability-recording-lite.yml` shows an unbroken wall of `failure` (12 consecutive runs checked, `29714562455` … `29791575946`). The job step reports:

```
STABILITY_RC: 126
HEALTHY: false
stability check command failed: rc=126
```

The script's own stderr, recovered from run `29791575946`'s uploaded `stability.log` artifact, is a single line:

```
scripts/ops/dingtalk-oauth-stability-check.sh: line 126: /usr/bin/python3: Argument list too long
```

Line 126 on `main` is exactly the start of the env-var block that hands the scraped payloads to the embedded verdict python:

```bash
126: WEBHOOK_STATUS_INPUT="${WEBHOOK_STATUS}" \
127: HEALTH_JSON_INPUT="${HEALTH_JSON}" \
128: METRICS_TEXT_INPUT="${METRICS_TEXT}" \
     ...
140: python3 - <<'EOF'
```

The owner's diagnosis is confirmed exactly.

## Mechanism — why an env var/argv hits E2BIG

A process's `argv` and `envp` share one kernel budget (`ARG_MAX`). On Linux each *individual* string is additionally capped at `MAX_ARG_STRLEN` = 128 KiB, which `getconf` does not report. `METRICS_TEXT` is a full `/metrics/prom` scrape — unbounded, and it grows with the deploy host's metric cardinality. Once it crossed the limit, `execve()` returned `E2BIG`; bash reports a command it could not execute as **126**, and under `set -euo pipefail` bash 5 propagates that as the script's exit status. Hence `rc=126`.

**This is a command-level failure of the monitor, not a health verdict about the host.** The recording never got far enough to compute a verdict — `healthy=false` here means "the script died", not "the deploy host is unhealthy". In fact the payload being large enough to blow `ARG_MAX` is positive evidence that the authenticated remote scrape was *succeeding* and returning real metrics; only the local hand-off to python broke.

## The fix

`scripts/ops/dingtalk-oauth-stability-check.sh` now writes the five scraped payloads (`WEBHOOK_STATUS`, `HEALTH_JSON`, `METRICS_TEXT`, `ALERTMANAGER_STATUS_JSON`, `ALERTS_JSON`) into a scratch dir and passes **file paths** (`<NAME>_FILE`) instead of values. A path is a fixed ~40 bytes regardless of payload size. The remaining variables are bounded scalars (counts, a `df` line, flags) and stay in env.

- Scratch dir is `mktemp -d` (0700 by construction) under `TMPDIR`, removed by a `trap … EXIT`.
- It deliberately does **not** live under `output/` — the workflow uploads `output/github/dingtalk-oauth-stability/` as an artifact, and the raw scrapes can carry hostnames/tokens.
- `printf` is a bash builtin, so writing the payloads out never execs anything and cannot itself hit `E2BIG`.
- Nothing is echoed: the payload never reaches stdout/stderr, argv, or the artifact.
- Python gains a `read_input()` helper that prefers `<NAME>_FILE`. The `<NAME>_INPUT` env fallback is retained **only** so the existing hermetic verdict unit test can keep injecting small literal values unchanged; the new ARG_MAX test is what stops the shell call site from ever reverting to it.

Both lines pinned by `dingtalk-oauth-stability-metrics-auth-contract.test.mjs` are byte-identical — the auth/secret-safety posture is untouched.

## The new behaviour test

`scripts/ops/dingtalk-oauth-stability-argmax-payload.test.mjs` — fully hermetic (PATH-shimmed `ssh`, stubbed webhook-config script, no network, no deploy host, no docker). It reads the machine's **real** `getconf ARG_MAX` at runtime and drives the actual script end to end with a `/metrics/prom` response of `ARG_MAX + 1 MiB` (2 MiB locally on macOS, **5 MiB** on the Linux runner, whose `getconf ARG_MAX` is 4194304). Padding uses a non-matching metric name so exactly one real OAuth sample must be parsed out of the noise — surviving is not enough, the verdict must still be correct.

Five cases:
1. Oversized payload → script exits 0, `operationsSamples` == the one real sample, `healthy=true`.
2. Values-free: the raw scrape body never appears on stdout or stderr.
3. The `mktemp` scratch dir is gone after the run (EXIT trap actually fires).
4. **Red-under-revert (mutation):** rewrites `METRICS_TEXT_FILE="${PAYLOAD_DIR}/metrics.prom"` back to `METRICS_TEXT_INPUT="${METRICS_TEXT}"` on a copy and re-runs the same scenario — asserts non-zero exit **and** the literal `Argument list too long` on stderr, so the redness is provably `E2BIG` and not an unrelated crash. On bash ≥ 4 it additionally pins `rc == 126`, the exact production signature. (Stock macOS bash 3.2 collapses every `set -e` abort to 1, so that pin is version-gated. It has now been executed for real on both a local bash 5.3.15 and the Linux runner — see the runner evidence below.)
5. **Positive control:** the *same* mutated script succeeds on a small payload — proving the redness in (4) comes from payload SIZE, not from a malformed mutation.

### Red-under-revert proof on the real file (replayed on bash 5 — rc is now the exact 126)

Reverting the fix in the actual working-tree script (not just the in-test copy) and re-running the
**exact command the required CI step runs**:

```
GNU bash, version 5.3.15(1)-release (aarch64-apple-darwin25.4.0)
1048576
✖ a /metrics/prom payload LARGER than ARG_MAX (2097152 bytes > 1048576 from getconf ARG_MAX) still produces a verdict
  AssertionError: an oversized metrics scrape must not fail the script (this is the rc=126 regression).
  status=126 stderr=…/dingtalk-oauth-stability-check.sh: line 161: /usr/bin/python3: Argument list too long
✖ values-free: the oversized payload body is never echoed to stdout or stderr
✖ the scratch payload dir is removed by the EXIT trap
✖ mutation: … (sanity guard trips — the file-path handoff is gone)
✖ control: … (same sanity guard)
ℹ pass 0 / fail 5
```

`status=126` **and** `Argument list too long` — byte-for-byte the production signature
(`STABILITY_RC: 126` + `line 126: /usr/bin/python3: Argument list too long`). The earlier revision of
this PR body recorded `status=1` for the same reproduction; that was an artefact of the stock macOS
bash 3.2, which collapses every `set -e` abort to 1. Re-run under bash 5 the rc is 126.

The script was restored with `git checkout -- scripts/ops/dingtalk-oauth-stability-check.sh`
afterwards; `git status --porcelain` is clean and the suite is back to 5/5.

**Positive control kept:** case 5 runs that same reverted script against a small payload and requires
exit 0 with a correct verdict — so the guard is "reject oversized-by-env-var", not "reject
everything".

## Test results

Focused suite (all six OAuth-stability suites), re-run at the new head under **bash 5.3.15** so the
version-gated `rc == 126` pin was live:

```
node --test scripts/ops/dingtalk-oauth-stability-argmax-payload.test.mjs \
            scripts/ops/dingtalk-oauth-stability-metrics-auth-contract.test.mjs \
            scripts/ops/dingtalk-oauth-stability-metrics-only-contract.test.mjs \
            scripts/ops/dingtalk-oauth-stability-metrics-only-fullscript.test.mjs \
            scripts/ops/github-dingtalk-oauth-stability-summary.test.mjs \
            scripts/ops/dingtalk-oauth-stability-workflow-contract.test.mjs
ℹ tests 39   ℹ pass 39   ℹ fail 0
```

The five sibling suites are green and **unmodified**. The `t.diagnostic` line added to the mutation
case does not create a test — the count is still 39.

Every other no-DB guard already living in the required `test` job was re-run too, to prove the new
step did not trip an exhaustive-step assertion in any of the `*-ci-wiring` contracts that read
`plugin-tests.yml`:

```
node --test scripts/ops/global-history-flag-manifest.test.mjs \
  scripts/ops/multitable-global-history-flag-status.test.mjs \
  scripts/ops/pb4-{2-archive-readonly,3-cycle-detection,4-reactivation}-ci-wiring.test.mjs \
  scripts/ops/b4-department-bindings-ci-wiring.test.mjs \
  scripts/ops/b5{a-routing-policy,b-routing-resolver,b-failclose,c-routing-routes}-ci-wiring.test.mjs \
  scripts/ops/b6-equivalence-ci-wiring.test.mjs \
  scripts/ops/b7-{reconciliation,round2}-ci-wiring.test.mjs \
  scripts/ops/t1-org-transfer-ci-wiring.test.mjs \
  scripts/ops/t2-source-freeze-ci-wiring.test.mjs \
  scripts/ops/t2gate-collision-mechanism-ci-wiring.test.mjs \
  scripts/ops/t2gate-runbook-values-free-contract.test.mjs \
  scripts/ops/stock-preparation-p4-repair-ci-wiring.test.mjs \
  scripts/ops/prod-health-probe-state.test.mjs
ℹ tests 96   ℹ pass 96   ℹ fail 0
```

Local Node is v25; CI authority is Node 20 (and 18 on the second matrix leg) — noted, not blocking.

## CI wiring — now in a REQUIRED context

The regression guard runs in **two** places:

| job | context | required on `main`? |
|---|---|---|
| `test` (matrix `18.x`, `20.x`) | `test (18.x)` / **`test (20.x)`** | **yes** — `test (20.x)` |
| `K3 WISE offline PoC` | `K3 WISE offline PoC` | no |

`main`'s required contexts (`gh api repos/zensgit/metasheet2/branches/main/protection`):

```
contracts (strict), contracts (dashboard), contracts (openapi),
pr-validate, test (20.x), web-tests, stock-prep PowerShell 5.1 acceptance
```

The step was added to the **no-DB guard block at the top of the `test` job** — the same block that
already hosts `global-history-flag-manifest`, the `pb4-*` / `b4`-`b7` / `t1` / `t2` CI-wiring
contracts and `prod-health-probe-state`, i.e. before `Setup pnpm` / `Install dependencies` / any
Postgres service. It therefore needs no database and fails fast.

`plugin-tests.yml`'s `pull_request:` trigger has **no path filter** (only the `push:` trigger does),
so `test (20.x)` runs on every PR to `main` — this is a real gate, not a path-conditional one.

**Kept the K3 step as well** (rather than moving it). Reason: the required `test` job supplies the
gate; the K3 job keeps the suite grouped with its five sibling OAuth-stability suites, which is where
someone debugging this script looks first. The duplicate costs ~5 s and cannot drift, since both
steps invoke the identical single command.

### Runtime / dependency budget

`node --test scripts/ops/dingtalk-oauth-stability-argmax-payload.test.mjs`

- **No DB, no network, no docker.** Everything is a PATH shim in `mkdtemp`.
- Externals: `bash`, `python3`, `getconf`, `mktemp` — all preinstalled on `ubuntu-latest`.
- Payload size is `getconf ARG_MAX + 1 MiB`. The `ubuntu-latest` runner reports
  `getconf ARG_MAX = 4194304`, so the fixture there is **5 MiB** (locally on macOS, 2 MiB).
  Building it is in-process string concatenation; the script reads it from a file with `cat`.
- Measured: **1.2 s** on the runner for all five cases (4.7 s locally); see the runner log below.
- Runs on both matrix legs (18.x and 20.x). It uses only `node:test`, `node:assert/strict`,
  `node:child_process` and `node:fs` — all available on Node 18.

### bash ≥ 4 — the `rc == 126` pin, and how I verified it actually executed

The mutation case asserts `Argument list too long` **plus** `rc == 126` — but the `126` pin is
gated on `BASH_VERSINFO[0] >= 4`, because bash 3.2 (stock macOS) collapses every `set -e` abort to
`1`. `node --test` prints "pass" whether a version-gated assertion fired or was skipped, so
"the suite is green" is *not* evidence the pin ran.

Two things make it observable rather than assumed:

1. the mutation test now emits a `t.diagnostic` line —
   `bash_major=<N> reverted_rc=<rc> rc126_pin=APPLIED|SKIPPED`;
2. the CI step echoes `bash --version | head -1` and `getconf ARG_MAX` before running the suite.

### Runner evidence — the `rc == 126` pin executed on CI (quoted, not inferred)

Run [`29800371209`](https://github.com/zensgit/metasheet2/actions/runs/29800371209), head
`382c9460d`, conclusion **success**. All seven required contexts pass, `test (20.x)` among them.

Verbatim from the **required `test (20.x)`** job (job `88540113302`), the new step:

```
##[group]Run bash --version | head -1
bash --version | head -1
getconf ARG_MAX
node --test scripts/ops/dingtalk-oauth-stability-argmax-payload.test.mjs
shell: /usr/bin/bash -e {0}
##[endgroup]
GNU bash, version 5.2.21(1)-release (x86_64-pc-linux-gnu)
4194304
TAP version 13
# Subtest: a /metrics/prom payload LARGER than ARG_MAX (5242880 bytes > 4194304 from getconf ARG_MAX) still produces a verdict
ok 1 - a /metrics/prom payload LARGER than ARG_MAX (5242880 bytes > 4194304 from getconf ARG_MAX) still produces a verdict
ok 2 - values-free: the oversized payload body is never echoed to stdout or stderr
ok 3 - the scratch payload dir is removed by the EXIT trap (no plaintext scrape left behind)
ok 4 - mutation: reverting METRICS_TEXT to an env var makes the oversized payload die with E2BIG
# bash_major=5 reverted_rc=126 rc126_pin=APPLIED
ok 5 - control: the SAME reverted script succeeds on a small payload — the redness above is caused by SIZE, not by a malformed mutation
1..5
# tests 5   # pass 5   # fail 0
```

The `test (18.x)` leg (job `88540113328`) and the `K3 WISE offline PoC` leg (job `88540113250`) print
the same `# bash_major=5 reverted_rc=126 rc126_pin=APPLIED`.

What that one diagnostic line settles, which "5 pass" alone cannot:

- `bash_major=5` — the runner's bash is 5.2.21, so the version gate opened;
- `rc126_pin=APPLIED` — the `assert.equal(result.status, 126)` branch **executed**; it is not a
  skipped assertion hiding inside a green result;
- `reverted_rc=126` — the reverted script really did exit `126` on the runner, i.e. the exact
  `STABILITY_RC: 126` the failing scheduled runs reported.

Runner `getconf ARG_MAX` is `4194304`, so the fixture payload on CI is **5 MiB**; the five cases take
**~1.3 s** in the required job and touch no database, no network and no docker.

## Still pending — live acceptance

The owner's acceptance bar is **two consecutive scheduled runs with no command-level failure**. That cannot be demonstrated here: the sandbox cannot reach the deploy host, and per the task constraint the scheduled workflow was deliberately **not** dispatched. The next two natural `cron: '15 */2 * * *'` runs after merge are the acceptance evidence.

Note also that clearing `rc=126` unblocks the *recording*; the health verdict it then reports is a separate question this PR makes no claim about. A run could legitimately come back `healthy=false` for real host reasons — that would be the monitor working, not this fix failing.

## Scope

```
.github/workflows/plugin-tests.yml
scripts/ops/dingtalk-oauth-stability-argmax-payload.test.mjs
scripts/ops/dingtalk-oauth-stability-check.sh
```

`dingtalk-oauth-stability-recording-lite.yml` is intentionally untouched — the bug was never in the workflow.


zensgit:
## R3 — regression gate moved into a REQUIRED context (UNARMED)

**Owner finding.** The implementation, temp-file handoff, cleanup, no-payload-leak and ARG_MAX
behaviour were accepted and the focused suite was 39/39 — but the new test hung only off
`K3 WISE offline PoC`, which is **not** a required check on `main`. A revert of the file-path handoff
could therefore have landed green.

**How it is closed.** `scripts/ops/dingtalk-oauth-stability-argmax-payload.test.mjs` is now a step in
the **required `test` job** of `.github/workflows/plugin-tests.yml`, in the no-DB guard block at the
top of that job — the same block that already hosts `global-history-flag-manifest`, the
`pb4-*` / `b4`–`b7` / `t1` / `t2` CI-wiring contracts and `prod-health-probe-state`, i.e. before
`Setup pnpm`, before `Install dependencies`, before Postgres. It runs on **both** matrix legs, and
`test (20.x)` is a required context:

```
$ gh api repos/zensgit/metasheet2/branches/main/protection --jq '.required_status_checks.contexts'
["contracts (strict)","contracts (dashboard)","pr-validate","test (20.x)","contracts (openapi)","web-tests","stock-prep PowerShell 5.1 acceptance"]
```

`plugin-tests.yml`'s `pull_request:` trigger has **no path filter** (only `push:` does), so
`test (20.x)` runs on every PR to `main` — a real gate, not a path-conditional one.

**Runtime / DB (DO#2).** No DB, no network, no docker; the externals are `bash`, `python3`,
`getconf`, `mktemp`, all preinstalled on `ubuntu-latest`. The runner's `getconf ARG_MAX` is
`4194304`, so the fixture payload there is **5 MiB** (`ARG_MAX + 1 MiB`) — bigger than the 3 MiB the
previous body estimated. Measured on the runner: **~1.2 s** for all five cases. It uses only
`node:test` / `node:assert` / `node:child_process` / `node:fs`, so the 18.x leg is fine too.

**K3 step: kept, not moved.** The required `test` job supplies the gate; the K3 job keeps the suite
grouped with its five sibling OAuth-stability suites, which is where anyone debugging this script
looks first. Both steps invoke the identical single command, so they cannot drift, and the duplicate
costs ~1.2 s.

### Red signature (owner's reproduction replayed, now on bash 5)

Reverting the fix in the **real** working-tree script and running the exact command the required step
runs:

```
GNU bash, version 5.3.15(1)-release (aarch64-apple-darwin25.4.0)
1048576
✖ a /metrics/prom payload LARGER than ARG_MAX (2097152 bytes > 1048576 from getconf ARG_MAX) still produces a verdict
  AssertionError: an oversized metrics scrape must not fail the script (this is the rc=126 regression).
  status=126 stderr=…/dingtalk-oauth-stability-check.sh: line 161: /usr/bin/python3: Argument list too long
✖ values-free: …   ✖ the scratch payload dir is removed by the EXIT trap …
ℹ pass 0 / fail 5
```

`status=126` **and** `Argument list too long` — byte-for-byte the production signature
(`STABILITY_RC: 126`). Script restored with `git checkout -- scripts/ops/dingtalk-oauth-stability-check.sh`;
`git status --porcelain` clean; suite back to 5/5.

### Positive control

Case 5 runs that *same reverted* script against a small payload and requires exit 0 with
`oauthMetricsPresent=true` / `healthy=true`. The guard is "reject oversized-by-env-var", not "reject
everything". Case 1 additionally requires `operationsSamples == [the one real sample]`, so surviving
the 5 MiB payload is not enough — the verdict must still be correct.

### bash ≥ 4 caveat, and how I verified it actually executed there (DO#3)

The `rc == 126` assertion is gated on `BASH_VERSINFO[0] >= 4`: stock macOS bash 3.2 collapses every
`set -e` abort to `1`, so on 3.2 the pin is skipped and only `Argument list too long` is asserted.
`node --test` prints "pass" whether a version-gated assertion fired or was skipped, so a green suite
is **not** evidence the pin ran. Two changes make it observable:

1. the mutation case emits `t.diagnostic(...)` →
   `bash_major=<N> reverted_rc=<rc> rc126_pin=APPLIED|SKIPPED` (a diagnostic, not a test — count is
   still 39);
2. the CI step echoes `bash --version | head -1` and `getconf ARG_MAX` before running the suite.

Evidence from run [`29800371209`](https://github.com/zensgit/metasheet2/actions/runs/29800371209)
at head `382c9460d`. Step status in the required job:

```
test (20.x) :: DingTalk OAuth-stability ARG_MAX payload-handoff regression (DT-CLOSE-01C) => completed/success
test (18.x) :: DingTalk OAuth-stability ARG_MAX payload-handoff regression (DT-CLOSE-01C) => completed/success
```

Verbatim from that job's step log (18.x leg, job `88540113328` — the 20.x leg's step is
`completed/success` above, but its log archive only becomes downloadable once the whole job including
the Postgres battery finishes):

```
##[group]Run bash --version | head -1
GNU bash, version 5.2.21(1)-release (x86_64-pc-linux-gnu)
4194304
# Subtest: a /metrics/prom payload LARGER than ARG_MAX (5242880 bytes > 4194304 from getconf ARG_MAX) still produces a verdict
ok 1 - …
ok 4 - mutation: reverting METRICS_TEXT to an env var makes the oversized payload die with E2BIG
# bash_major=5 reverted_rc=126 rc126_pin=APPLIED
ok 5 - control: the SAME reverted script succeeds on a small payload …
1..5
# tests 5   # pass 5   # fail 0
```

`rc126_pin=APPLIED` with `reverted_rc=126` is the literal proof that the version-gated branch
executed on the runner, not an inference from a green checkmark. The `K3 WISE offline PoC` job in the
same run prints the same line.

### Regression sweep

- Six OAuth-stability suites re-run at the new head under bash 5.3.15: `tests 39 / pass 39 / fail 0`.
  The five siblings are unmodified.
- Every other no-DB guard already in the required `test` job re-run (they read `plugin-tests.yml`, so
  the new step must not trip an exhaustive-step assertion): `tests 96 / pass 96 / fail 0`.
- Local Node is v25; CI authority is Node 20 (plus the 18.x leg) — noted, not blocking.

### Still pending — live acceptance is post-merge

Unchanged and restated: the owner's acceptance bar is **two consecutive scheduled runs of
`DingTalk OAuth Stability Recording (Lite)` with no command-level failure**. That cannot be shown
here — the sandbox cannot reach the deploy host and the scheduled workflow was deliberately not
dispatched. The next two natural `cron: '15 */2 * * *'` runs **after merge** are the acceptance
evidence. Clearing `rc=126` unblocks the *recording*; the health verdict it then reports is a
separate question this PR makes no claim about.

### Head

`382c9460d462e1e6c64187ba95cfad003318a7ae` (was `53a5e8fab`). The shell script's fix is byte-identical
to the reviewed revision — only `.github/workflows/plugin-tests.yml` (the new required step) and the
test file (the `t.diagnostic` line) changed. **UNARMED** — not merged, auto-merge not enabled.

PR #4521, opened 2026-07-21
feat(core-backend): W4-PRE-1 — user_orgs production maintenance write path (org-membership lifecycle)

zensgit:
## W4-PRE-1 — `user_orgs` production maintenance write path (org-membership lifecycle)

Pre-ticket for Wave 4 restart sequence. Authoritative spec = **main** at `57d89bc1d`
`docs/development/attendance-vnext-wave4-onboarding-design-lock-20260721.md` §3.3 (the
"W4-PRE-1 票面", introduced by owner errata `#4513`, which currently marks the lock
`RATIFICATION SUSPENDED / W4-0 FROZEN`). This PR implements that ticket's five items and does
**not** touch, re-open, or ratify anything else in the lock — Wave 4 slices (W4-0..W4-2) remain
FROZEN, and the frozen materiel branch `claude/w4-0-setup-readiness-20260721` was not touched.

Base: `origin/main` `57d89bc1d189d7fe18f3b3a848a4e0b4cce07ce6`. Branch:
`claude/w4-pre1-user-orgs-lifecycle-20260721`.

Backend-only change (`packages/core-backend`). **Three-viewport visual evidence: N/A** — no UI
surface in this PR.

### Post-review fixes (three-mirror pre-door review, fix commit `a3d05837f`)

A three-mirror pre-door review of this PR raised 2 P2 + 8 P3 findings (some duplicate across
files). All addressed on top of the original `84475f1ab` head, head is now `a3d05837f`:

- **P2 (behavior)**: `admin-users.ts`'s `user_orgs` write hardcodes `is_active=TRUE` at
  admission now (mirrors `directory-sync.ts`'s own writer), instead of copying the created
  user's own `isActive` flag. Mirroring `isActive` added nothing to the RD-3 dual-is_active
  count (`users.is_active=false` already excludes an admin-created-inactive user) but created an
  **unrecoverable stuck-false membership row**: no production write path ever updates
  `user_orgs.is_active` after admission, so reactivating that user later via `PATCH
  /api/admin/users/:userId/status` (which only touches `users.is_active`) left them permanently
  excluded from the design-lock §3① count — an absorbing state with no repair surface. New test
  (`attendance-w4pre1-user-orgs-admission.db.test.ts`) proves `isActive:false` at creation still
  writes `user_orgs.is_active=true`, exclusion comes only from `users.is_active`, and
  reactivation needs no separate membership repair. See updated item-4 row and mutation #4 below.
[...]

PR #4526, opened 2026-07-22
fix(core-backend): W4-PRE-1b — user_orgs full org-membership lifecycle (bind/unbind/rebind/backfill/explicit-org/dual-gate)

zensgit:
## What

W4-PRE-1b — the code slice owner ordered in the CHANGES_REQUESTED verdict on the W4 re-ratify PR #4522 (2026-07-21 comment). W4-PRE-1 (#4521 = `e20371b1a`, merged) only built the "new admission" write path. This PR closes the rest of the `user_orgs` lifecycle: bind/auto-match for **already-existing** users (item A), org-scoped safe deactivation on unbind/rebind/archive (item B), a real-stock backfill migration (item C), an explicit `attendanceOrgId` admin-users path that no longer needs a circular group/shift dependency (item D), and a dual `is_active` fix on the `user_orgs` trust points the door and a same-shape sibling gate rely on (item E).

**Not merged, not armed.** Base `main`, refs #4522 #4521.

## #4526 three-mirror review round — fixes applied (2026-07-22)

Ten findings came back on this PR (three independent reviewers, some overlapping). Disposition:

| Finding | Severity | Disposition |
|---|---|---|
| Sync-loop `already_linked` steady-state upsert can resurrect a `user_orgs` row for an account the DT-OPS-01 sweep already marked departed (this loop walks ALL accounts every sync, not just the current batch) | P1 | **Fixed** — the write now also requires `account.is_active` (gate added at the exact write site, not a broader projection filter). New real-DB regression pair (negative: swept-inactive account does not resurrect; positive control: a returning/reactivated account still gets the steady-state re-confirm). |
| DT-OPS-01 sweep (directory-side departure detection) never deactivates `user_orgs`; PR body previously claimed directory-side link removal was "implemented above" — true only for the LOCAL provider's `archiveLocalAccount`, false for the DingTalk sweep | P1/P2 (both raised, different severities) | **Claim corrected** (was factually wrong); **fix deferred to owner ruling** — whether a departure-detection sweep counts as a "binding event" under this line's own unbind/rebind/archive boundary is a design call, not something to self-adjudicate. Lifecycle table gets an explicit row for this event, honestly marked NOT implemented. Code comment added at the sweep site documenting the gap. |
| `automation-executor.ts`'s `send_dingtalk_group_message` org-destination authorization EXISTS clause is the same `user_orgs`-trust shape as the already-fixed `api-tokens.ts`/`attendance-admin.ts` sites, but was missing from this PR's "full-repo grep" inventory and missing the dual `users.is_active` filter | P2 | **Fixed** — same dual-filter treatment as the other two item-E sites; inventory section corrected to include it. |
| Cross-transaction write skew: two concurrent unbinds of a double-bound user's two different accounts could each see the other's account as still active (neither committed yet) and both skip deactivation, leaving `user_orgs` ACTIVE with zero live bindings permanently | P2 | **Fixed** — `deactivateUserOrgMembershipIfNoOtherActiveBinding` now takes a `SELECT ... FOR UPDATE` lock on the shared `(userId, orgId)` row before evaluating the sibling check, forcing the second caller to observe the first's already-committed severance. Proven with a constructed race (two raw pg clients, `pg_blocking_pids`-based "throw if never parked" non-vacuity check — same technique as `multitable-l4-canonical-fence-realdb.test.ts`'s R1/RXR), not a bare `Promise.all` (which, empirically, does not reliably reproduce this on localhost — see the test file's own doc comment for the investigation). |
| `unbindDirectoryAccount` read the previously-linked user OUTSIDE the transaction (stale pre-read); a rapid unbind+rebind of the SAME account inside that window could sever the CURRENT holder's link while only deactivating the STALE holder | P3 | **Fixed** — the read moved inside the transaction with `FOR UPDATE OF l` locking the `directory_account_links` row. |
| Record discipline: several code comments and this PR body itself attributed phrases to the owner ("(owner text)", "owner ruling") that do not appear verbatim in the owner's actual #4522 review comment (comment-5040485829, re-verified against gh) | P2 | **Fixed** — reworded `admin-users.ts`, `directory-sync.ts`'s `deactivateUserOrgMembershipIfNoOtherActiveBinding` doc comment, `local-directory-org.ts`, the backfill migration's doc comment, two test-file doc comments, and this PR body's own lifecycle table/explicit-org-path section to stop asserting unverified quotes/rulings. |
| Item E's `api-tokens.ts` fix had only mock-SQL-text-equality test coverage (no real-DB proof the JOIN actually filters `users.is_active=false`) | P3 | **Fixed** — new real-DB test file `attendance-w4pre1b-api-tokens-org-member-access.db.test.ts` (4 tests) drives the real `GET /api/multitable/dingtalk-groups?orgId=` endpoint against a real Postgres. |
| F6's "fresh-org" fixture pre-seeds a `directory_integrations` anchor row but never explicitly asserted zero `attendance_groups`/`attendance_shifts` for that org | P3 | **Fixed** — added an explicit before/after zero-scaffold assertion to the existing F6 fresh-org test. |

Affected suites re-run in full (see updated Local verification section below); typecheck clean both packages; mutation self-check re-run for every new/changed guard (see updated mutation table).

## Post-gate fix (Opus gate — 2026-07-22)

（下引块为正门 P2 的**忠实摘述**（非逐字），完整原文见 `/private/tmp/pr4526-review-w4-pre1b-gate.md` L15-19——独立复核 NIT 口径更正）

The Opus gate on this PR returned **CHANGES-REQUESTED** with a single P2, a pure test-coverage gap (zero production-code change required):

> deactivateUserOrgMembershipIfNoOtherActiveBinding's tenant-isolation predicate `AND i.org_id = $2::text` (directory-sync.ts:4977) is load-bearing but has ZERO coverage — removing only that line survives the full suite (12/12 green); constructed reachable failure: DingTalk-binding-in-orgA + local-binding-in-orgB, unbind A → org-A membership stays ACTIVE with zero org-A bindings = cross-org stale access.

**Fix — new test case**: `cross-org tenant-isolation regression — unbind in org A must not be gated by org B's active binding (#4526 P2)` → `unbinding org A's DingTalk account deactivates ONLY org A membership; org B's local binding for the same user stays untouched and active`, added to `attendance-w4pre1b-user-orgs-lifecycle.db.test.ts`. Same user: DingTalk binding in org A (via `bindDirectoryAccount`) + local binding in org B (via `createLocalAccount`); unbind A via the real `unbindDirectoryAccount` path (no test-authored SQL simulating the write); assert BOTH legs — org A membership `is_active=false` AND org B membership stays `is_active=true` (the dual proof: the A leg shows the predicate isn't fooled by org B's sibling; the B leg rules out a degenerate "always deactivate everything" pass).

**Mutation self-check** (test committed first — `2da81d98b` — then mutated via `Edit`, not `git checkout`, for restore):

Removed `AND i.org_id = $2::text` from the `NOT EXISTS` sibling-check subquery (`directory-sync.ts:4977`) →
```
 × cross-org tenant-isolation regression … > unbinding org A's DingTalk account deactivates ONLY org A membership; org B's local binding for the same user stays untouched and active
   → expected { is_active: true } to deeply equal { is_active: false }
 Test Files  1 failed (1)
      Tests  1 failed | 8 passed (9)
```
Only the new test reddens, and precisely on the org-A assertion — exactly the predicted failure mode: org B's still-`linked`+active local binding is picked up as a false "sibling", org A's deactivation is skipped, org A stays `is_active:true`. All 8 other cases in the file (F1/F2/F3/F4/F3-race/rebind-displacement/local-provider) stay green because none of them exercise a cross-org sibling — confirming this new test is the unique guard for the predicate. Restored the exact line via `Edit`; `git diff`/`git status` confirmed byte-identical to the committed state; re-run confirmed 9/9 green.

**Re-verification** (fresh, uniquely-named local Postgres — `ms2_w4pre1b_p2fix_20260721` — migrated clean, matching this PR's own "dedicated DB" convention):
```
pnpm --filter @metasheet/core-backend exec vitest --config vitest.integration.config.ts run \
  tests/integration/attendance-w4pre1b-user-orgs-lifecycle.db.test.ts \
  tests/integration/attendance-w4pre1b-user-orgs-sync-automatch.db.test.ts \
  tests/integration/attendance-w4pre1b-user-orgs-backfill-migration.db.test.ts \
  tests/integration/attendance-w4pre1b-directory-readiness-gate.db.test.ts \
  tests/integration/attendance-w4pre1b-admin-users-explicit-org.db.test.ts \
  tests/integration/attendance-w4pre1b-api-tokens-org-member-access.db.test.ts
# → 6 files, 25 tests, all passed (lifecycle file: 9/9, up from 8/8 pre-fix)
```

Test-only change — no production code modified by this commit. New head `2da81d98b`.

## Owner findings → implementation mapping

Quoting the owner's 2026-07-21 review comment verbatim, then what this PR does about each:

> "[P1] bind 存量路径（directory-sync.ts:4829）零 user_orgs 写入 + 解绑零失活 ⇒ ①假 missing 与 stale access 双面缺陷（S7-5 门 attendance-admin.ts:375 查 user_orgs.is_active=true，失活即闭洞）"

→ `applyDirectoryAccountBindInTransaction` (directory-sync.ts, shared by `bindDirectoryAccount`/`admitDirectoryAccountUser`/batch variants) now upserts an active `user_orgs` row for the bind target and deactivates a displaced prior holder; `unbindDirectoryAccount` now deactivates the unbound user's membership (org-scoped sibling check). `canReadAttendanceDirectoryReadiness` (attendance-admin.ts) now requires `users.is_active=true` in addition to `user_orgs.is_active=true`.

> "[P2] admin-users.ts:3172 org 仅在提交 group/shift 时解析 ⇒ 修复动作循环依赖确凿"

→ `POST /api/admin/users` accepts an independent `attendanceOrgId` body param, sufficient alone (no group/shift) to upsert membership in the same transaction. See F6 and the "explicit org path" deviation note below for the one open adjudication point.

> "[P2] ensurePlatformAdmin 门 vs 受托管理员 403"（and）"[P2] effectiveTime 合同回退"

→ **Out of scope for this PR** — these two are #4522's own lock-wording/verification-MD updates (owner's step 2 in the execution sequence), not code. Flagged in Open Items below so they are not silently dropped.

## Lifecycle event table (event / disposition / anchor)

| Event | Disposition | Anchor |
|---|---|---|
| Manual bind of an existing user (`bindDirectoryAccount`) | upsert ACTIVE `user_orgs` | `directory-sync.ts` `applyDirectoryAccountBindInTransaction` ~L4934 |
| Admit-new-user over a possibly-already-linked account (`admitDirectoryAccountUser`/`createDirectoryAdmittedUserInTransaction`) | upsert ACTIVE `user_orgs`; deactivate prior holder if displaced | `directory-sync.ts` ~L5117 → shared helper |
| Sync-loop auto-match (`external_identity` fresh match, or `already_linked` steady-state re-confirm) | upsert ACTIVE `user_orgs` when `linkStatus==='linked' && localUserId && account.is_active` (the `account.is_active` gate is the #4526-review P1 fix — without it, a departed-but-still-`linked` account's steady-state re-confirm silently resurrected a deliberately-deactivated row) | `directory-sync.ts` `syncDirectoryIntegration` loop ~L3773 |
| **Directory-side departure detection (DT-OPS-01 sweep)** | **NOT implemented this PR** — the sweep flips only `directory_accounts.is_active`, never `user_orgs`; whether this counts as a "binding event" under the boundary below is an open owner-ruling item (#4526 review) | `directory-sync.ts` sweep ~L3513 (gap documented in a code comment at the sweep site) |
| Manual unbind (`unbindDirectoryAccount`) | deactivate if no other active binding in org | `directory-sync.ts` ~L5921 |
| Same-account rebind displacement (single `bindDirectoryAccount` call reassigns account X from user A to user B) | deactivate A if no other active binding in org | `directory-sync.ts` `applyDirectoryAccountBindInTransaction` prior-holder capture ~L4934 — **investigated, confirmed UNREACHABLE for DingTalk accounts** (see deviation #2); code retained as defense-in-depth |
| Local account create (`createLocalAccount`) | upsert ACTIVE `user_orgs` | `local-directory-org.ts` ~L452 |
| Local account archive (`archiveLocalAccount`) | deactivate if no other active binding in org | `local-directory-org.ts` ~L595 |
| Admin explicit-org create (`POST /api/admin/users` with `attendanceOrgId`) | upsert ACTIVE `user_orgs` | `admin-users.ts` ~L3094/3199-3236 |
| User deactivation (`PATCH /api/admin/users/:userId/status`) | **NOT touched** — this PR's own reading of the binding-event boundary (presented as evidence, not an adjudicated owner ruling — see next section) | not modified this PR |
| Directory deprovision `mark_inactive` policy (`applyDirectoryDeprovisionPolicies`) | **NOT touched** — flips `users.is_active`, not `user_orgs` | not modified this PR |

## Deactivation boundary semantics (for owner review — presenting evidence, not a ruling)

Per the task's own framing: **only a BINDING event may deactivate `user_orgs`.** Concretely:
- Unbind / a same-account rebind that displaces a prior holder → deactivation IS in scope, implemented above.
- Directory-side link removal: **the LOCAL provider's `archiveLocalAccount` treats this as in scope** (implemented, see the lifecycle table) — but the DingTalk-equivalent event (the DT-OPS-01 sweep marking `directory_accounts.is_active=false` when an account stops appearing in the directory) does **NOT** get the same treatment in this PR. An earlier revision of this section claimed the DingTalk case was "implemented above" — that was **false**; corrected during the #4526 review. Whether the sweep counts as the same kind of "directory-side link removal" event `archiveLocalAccount`'s own doc comment reasons about is exactly the open question — requesting owner's ruling (see Open items).
- A→B org migration is NOT a separate function — it is the natural composition of "unbind account under org A's integration" (deactivates A) + "bind a new account under org B's integration" (activates B), proven end-to-end by F4.
- `PATCH /api/admin/users/:userId/status` (user deactivation) → membership row is **never touched**. The dual `users.is_active`/`user_orgs.is_active` read filter (already in place from #4521, and now also applied to the S7-5 gate in this PR) is what excludes a deactivated user from every count and gate — not a write-time membership flip.
- `applyDirectoryDeprovisionPolicies`'s `mark_inactive` branch flips `users.is_active`, never `user_orgs` — same boundary, deliberately excluded, not modified by this PR.

## F1–F7 test cases + mutation table

| # | Owner's case | File | Result |
|---|---|---|---|
| F1 | bind existing user ⇒ membership row (+ failure-injection rollback) | `attendance-w4pre1b-user-orgs-lifecycle.db.test.ts` | 2 tests, both pass |
| F2 | last unbind ⇒ deactivated | same file | 1 test, pass |
| F3 | double binding, unbind one ⇒ other stays active | same file | 2 tests (original; + `F3-race` constructed-concurrency regression, #4526 P2), pass |
| F4 | A→B migration ⇒ old deactivated, new active | same file | 1 test, pass |
| (bonus) | same-account rebind displacement | same file | 1 test — documents the pre-existing identity-conflict guard makes this unreachable for DingTalk (deviation #2) |
| (bonus) | local-provider bind/archive | same file | 1 test, pass |
| (item A 3rd writer) | sync-loop `external_identity` auto-match writes membership | `attendance-w4pre1b-user-orgs-sync-automatch.db.test.ts` | 3 tests (original re-match test; + #4526 P1 regression pair: a swept-inactive account does NOT resurrect membership, and a returning/reactivated account DOES still get the steady-state re-confirm), pass |
| F5 | stale/deactivated member ⇒ 403 via real endpoint | `attendance-w4pre1b-directory-readiness-gate.db.test.ts` | 3 tests (deactivated-user variant, deactivated-membership variant, active-member 200 control), pass |
| F6 | fresh-org explicit `attendanceOrgId` ⇒ membership, falsifies old circular dependency | `attendance-w4pre1b-admin-users-explicit-org.db.test.ts` | 5 tests (fresh-org success — now also asserts zero attendance_groups/shifts before AND after, #4526 review; unknown-org 404, malformed-param 400, group/shift-conflict 400, backward-compat control), pass |
| F7 | backfill: linked-but-membership-less gets a row; already-deactivated row never resurrected | `attendance-w4pre1b-user-orgs-backfill-migration.db.test.ts` | 1 test (4 population variants + idempotent rerun), pass |
| (item E follow-up) | `api-tokens.ts` `requireOrgMemberAccess` dual filter, real endpoint (#4526 P3 — previously mock-SQL-text-only) | `attendance-w4pre1b-api-tokens-org-member-access.db.test.ts` (new) | 4 tests (active/active 200; deactivated-user 403; deactivated-membership 403; no-membership 403), pass |

**Mutation self-check (≥4 original + 3 added in the #4526 review round; all confirmed RED then reverted — implementation committed before each mutation, restored via precise `Edit` rather than a broad `git checkout` during the review-round self-checks):**

| # | Mutation | Red leg | Restored |
|---|---|---|---|
| 1 | Neuter `upsertActiveUserOrgMembership` call in the bind writer | F1 + 4 other lifecycle tests red (5/7 in that file) | yes |
| 2 | Remove the `NOT EXISTS` sibling check from `deactivateUserOrgMembershipIfNoOtherActiveBinding` (unconditional deactivation) | **Only** F3 (double-binding-preserved-active) red — precise, load-bearing kill | yes |
| 3 | Change backfill's `ON CONFLICT (user_id, org_id) DO NOTHING` to `DO UPDATE SET is_active = TRUE` | **Only** F7's non-resurrection assertion red | yes |
| 4 | Drop `AND u.is_active = true` from the S7-5 gate SQL | **Only** F5's deactivated-user variant red | yes |
| 5 (#4526) | Drop `&& account.is_active` from the sync-loop's `already_linked` write gate | **Only** the new sync-automatch regression's negative case red (`expected [{is_active:true}] to deeply equal [{is_active:false}]`) | yes |
| 6 (#4526) | Drop the `SELECT ... FOR UPDATE` lock from `deactivateUserOrgMembershipIfNoOtherActiveBinding` | `F3-race`'s `waitUntilBlockedBy` throws after its 5s timeout ("the second deactivation call never parked on the first caller's row lock") — the test detects the ABSENCE of blocking, not just the wrong final state, so it can't be satisfied by luck | yes |
| 7 (#4526) | Drop `AND u.is_active = true` from `automation-executor.ts`'s DingTalk destination EXISTS clause | **Only** the updated `automation-v1.test.ts` mock-SQL-text assertion red | yes |

## Backfill idempotency self-proof

`zzzz20260721150000_backfill_user_orgs_from_directory_links.ts`: `INSERT ... SELECT ... ON CONFLICT (user_id, org_id) DO NOTHING`. F7 seeds four populations (linked+no-row, linked+already-deactivated, linked+already-active, linked-to-an-*inactive*-directory-account) and asserts: row created only for the first; the deactivated row's `is_active` stays `false`; the active row is untouched; the inactive-account population gets no row at all. The whole `up()` is re-invoked a second time in the same test and the result set is asserted byte-identical (`toEqual`) — true idempotency, not just "no error".

## Explicit org path (item D) — one open adjudication point

The owner's item-D text is "支持显式 attendanceOrgId（不依赖考勤组/班次）⇒ canonical surface 变为真无条件" — it establishes THAT an explicit `attendanceOrgId` must work standalone, but not HOW an unrecognized org id should be handled; that latter question has two defensible readings, and is a design call the task instructs me not to self-adjudicate (an earlier revision of this section mislabeled "校验 org 存在" as owner text in quotes — it is not verbatim owner text; corrected during the #4526 review):
- **Shipped**: validate-can-fail — `attendanceOrgId` is checked against `directory_integrations` (the anchor every org acquires today, local or dingtalk); an unrecognized org id gets `404 ATTENDANCE_ORG_NOT_FOUND`. Fail-closed by construction, cheapest to loosen later if wrong.
- **Not shipped**: auto-vivify — call `getOrCreateLocalIntegration(attendanceOrgId)` so ANY non-empty org id string from an admin succeeds, silently minting a new local org anchor. This is a bigger, more permissive action than "validate" literally means, and F6's fresh-org test therefore pre-seeds a `directory_integrations` row as fixture setup (simulating "org already has *some* anchor, zero attendance groups/shifts") rather than relying on the route to mint one from a totally bare string.

Both readings satisfy "breaks the circular dependency" — they differ on whether a brand-new org can be summoned purely by an admin's `POST /api/admin/users` call. **Requesting owner's ruling on this in the #4522 follow-up.**

## CI wiring (three-point)

1. `packages/core-backend/vitest.config.ts`'s no-DB exclude list — all 5 original + 1 added in the #4526 review round (`attendance-w4pre1b-api-tokens-org-member-access.db.test.ts`) `.db.test.ts` files, so the default no-DB job cannot skip-green them.
2. `.github/workflows/plugin-tests.yml`'s "Run attendance integration tests" step — same 6 files added to the explicit `vitest.integration.config.ts` invocation.
3. The new backfill migration (`zzzz20260721150000...`) is **not** in `MIGRATION_EXCLUDE` (either occurrence) — confirmed by running `pnpm --filter @metasheet/core-backend db:migrate` against a fresh dedicated local Postgres with the exact `MIGRATION_EXCLUDE` value copied from the workflow: the migration executed in its correct position (after `zzzz20260717130000`), no exclusion needed.

## Regression fixups (pre-existing suites this PR's changes required updating)

- `attendance-w4pre1-user-orgs-directory-sync.db.test.ts` (#4521): the org-resolution-failure error message text changed (`'Directory integration not found for account org resolution'`) because the resolution logic moved from `createDirectoryAdmittedUserInTransaction` into the now-shared `applyDirectoryAccountBindInTransaction` helper — same fail-closed guarantee (no `users` row survives), different message. Updated the one assertion.
- `directory-sync-bind-account.test.ts` (mocked unit suite): 4 tests converted from brittle positional `clientQuery.mockResolvedValueOnce()` chains to the SQL-text-inspecting closure pattern this same file's own "allows union-only pre-binding" test had already established for exactly this reason (#4521 anticipated more queries landing here). One `.find()` predicate disambiguated (`'SELECT local_user_id'` now matches two different queries — the original identity-conflict check and the new prior-holder capture — narrowed to `'FROM user_external_identities'`).
- `dingtalk-group-destination-routes.api.test.ts`: 2 exact-SQL-text assertions on `api-tokens.ts`'s `requireOrgMemberAccess` updated to the new dual-filter query text.
- (#4526 review round) `directory-sync-bind-account.test.ts`: 3 `unbindDirectoryAccount` tests' `clientQuery` mocks updated to also answer the new in-transaction `FOR UPDATE OF l` linked-user read (previously served by the outer `pgMocks.query` pre-read this round moved inside the transaction).
- (#4526 review round) `automation-v1.test.ts`: the `send_dingtalk_group_message` scoping test's inline SQL-text assertions extended to also require the new `JOIN users u` / `u.is_active = true` clauses.

## Grep inventory — other `user_orgs`-trusting sites (item E "同型门")

Full-repo grep for `user_orgs` trust points (boolean access-gate shape, `... WHERE user_id = $1 AND org_id = $2 AND is_active = true ...`):
- `attendance-admin.ts:373` `canReadAttendanceDirectoryReadiness` — **fixed** (owner-named).
- `api-tokens.ts:155` `requireOrgMemberAccess` (identical SQL shape, gates DingTalk group-destination org read/write access) — **fixed**.
- `automation-executor.ts` `send_dingtalk_group_message`'s destination-authorization `EXISTS` clause — **fixed in the #4526 review round** (this site was MISSING from the original "full-repo grep" inventory below; the original claim of completeness was inaccurate). Same shape as `api-tokens.ts`, gating actual outbound-send authorization by the automation rule creator's org membership.
- `dingtalk-group-destination-service.ts:215-234` — two Kysely `EXISTS` clauses inside `listDestinations`'s visibility filter, same underlying finding (a departed/deactivated member's stale `user_orgs` row could still surface org-scoped rows) — **inventoried, NOT fixed**. Rationale for leaving it: it is a read-visibility filter embedded in a `SELECT ... WHERE (OR ...)` builder rather than a hard 403 gate, has its own dedicated test suite (`dingtalk-group-destination-service.test.ts`), and touching it risks a larger, less-reviewable diff for a surface the owner did not name. Flagged for owner to decide if it belongs in this line or a follow-up.
- `plugins/plugin-attendance/index.cjs` (multiple sites, e.g. `:2784`, `:15532`, `:18107`) — **already** carry the dual `is_active` filter (pre-existing RD-3 precedent this PR's fixes now match); no change needed.

## Deviations (honest, evidence only — no verdicts asserted)

1. Item D's unrecognized-org-id handling — two readings, shipped validate-can-fail; see dedicated section above.
2. Advisor-flagged "same-account rebind displacement via a single `bindDirectoryAccount` call" was investigated and is empirically **unreachable for DingTalk accounts**: `applyDirectoryAccountBindInTransaction`'s pre-existing `user_external_identities` conflict guard throws `'DingTalk account is already bound to another local user'` before the link write is ever reached, whenever the target account's identity is already tied to a different local user. Proven by running exactly that call sequence against a real DB (see the file's "investigated, confirmed unreachable" test). The prior-holder capture + deactivate code in `applyDirectoryAccountBindInTransaction` is retained as defense-in-depth (cheap, correct, harmless) but has no currently-reachable production trigger for the DingTalk provider that this audit found; `createLocalAccount`'s local-provider path has no rebind concept at all (its `external_key` is `org_id:local_user_id`, inherently per-user).
3. `dingtalk-group-destination-service.ts`'s two EXISTS-based visibility filters — grep-inventoried as the same finding shape, deliberately not modified (see inventory section).
4. A sync-loop scenario where a previously-`linked` account transitions to `unmatched` on a LATER resync (e.g., an ambiguous re-match) was reasoned about but not separately implemented/tested: this requires the account's `user_external_identities` row to first be deleted, which — in every production path this PR's inventory found — only happens via `unbindDirectoryAccount` (already fixed in this PR). No additional gap was identified, but no dedicated test proves this specific transition combination.
5. (#4526 review) The DT-OPS-01 sweep (directory-side departure detection) does not deactivate `user_orgs` — see the lifecycle table and boundary-semantics section above. Left unfixed pending owner ruling on whether it counts as a "binding event"; a code comment at the sweep site documents the gap so it isn't silently lost.

## Open items (for owner)

- Ruling requested on item D's unrecognized-org-id handling (validate-can-fail, shipped, vs auto-vivify).
- Ruling requested on whether `dingtalk-group-destination-service.ts`'s two EXISTS-based filters need the same dual-filter treatment, in this line or a follow-up.
- Ruling requested (#4526 review) on whether the DT-OPS-01 directory-side departure sweep counts as a "binding event" that should deactivate `user_orgs` — see lifecycle table + boundary-semantics section.
- #4522's own step-2 items (lock wording, `ensurePlatformAdmin` vs delegated-403 role contract, `effectiveTime` four-state restoration) are **not** touched by this PR — they are #4522's own docs/verification-MD work per the owner's execution sequence, referenced here so they are not silently dropped.
- Staged opt-in / owner merge decision, per this line's standing convention — this PR does not merge or arm itself.

## Local verification (this worktree, dedicated Postgres — `metasheet_w4pre1b_test`)

Original round:
```
pnpm --filter @metasheet/core-backend db:migrate   # MIGRATION_EXCLUDE copied verbatim from plugin-tests.yml
# → zzzz20260721150000_backfill_user_orgs_from_directory_links executed in position, no exclusion needed

pnpm --filter @metasheet/core-backend exec vitest --config vitest.integration.config.ts run <full "Run attendance integration tests" file list from plugin-tests.yml, including the 5 new files>
# → 40 files, 522 tests, all passed

pnpm --filter @metasheet/core-backend exec vitest run   # default no-DB config, full sweep
# → 688 files (506 passed / 182 DB-gated skipped), 6909 tests passed, 1627 skipped, 0 failed

pnpm --filter @metasheet/core-backend run type-check    # clean
pnpm --filter @metasheet/web run type-check              # clean (no frontend changes this PR; ran anyway per double-typecheck discipline)
```

#4526 review-round re-verification (same dedicated Postgres):
```
# All 6 W4-PRE-1b real-DB files (5 original + 1 new) + the 3 pre-existing W4-PRE-1 (#4521) real-DB
# files + the full directory-sync/directory-binding real-DB suite (18 files total):
pnpm --filter @metasheet/core-backend exec vitest --config vitest.integration.config.ts run <18 files>
# → 18 files, 105 tests, all passed

# Mock/unit suites touched by this round's changes:
pnpm --filter @metasheet/core-backend exec vitest run tests/unit/automation-v1.test.ts \
  tests/unit/directory-sync-bind-account.test.ts tests/integration/dingtalk-group-destination-routes.api.test.ts \
  tests/unit/admin-users-routes.test.ts
# → 4 files, 320 tests, all passed

pnpm --filter @metasheet/core-backend exec vitest run   # default no-DB config, full sweep (repeat)
# → 688 files (506 passed / 182 DB-gated skipped), 6909 tests passed, 1627 skipped, 0 failed — byte-identical
#   totals to the original round, confirming the new file is correctly excluded from the no-DB job

pnpm --filter @metasheet/core-backend run type-check    # clean
pnpm --filter @metasheet/web run type-check              # clean (still no frontend changes)
```

Mutation self-checks 5-7 (see mutation table): each mutation applied on top of a COMMITTED implementation, confirmed RED, then reverted with a precise `Edit` (not a broad `git checkout`, which would have discarded the not-yet-committed test file for #6's rewrite mid-investigation).

Three-viewport visual verification: **N/A** — this PR is backend-only (routes, directory-sync service functions, one data migration); no Vue/frontend changes.

refs #4522 #4521

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -828,6 +828,7 @@ jobs:
             tests/integration/attendance-w4pre1b-user-orgs-backfill-migration.db.test.ts \
             tests/integration/attendance-w4pre1b-directory-readiness-gate.db.test.ts \
             tests/integration/attendance-w4pre1b-admin-users-explicit-org.db.test.ts \
+            tests/integration/attendance-w4pre1b-api-tokens-org-member-access.db.test.ts \
             tests/integration/attendance-expiry-service.test.ts \
             tests/integration/attendance-unscheduled-reminder.test.ts \
             tests/integration/attendance-outdoor-punch.test.ts \
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -823,6 +823,11 @@ jobs:
             tests/integration/attendance-w4pre1-user-orgs-admission.db.test.ts \
             tests/integration/attendance-w4pre1-user-orgs-directory-sync.db.test.ts \
             tests/integration/attendance-w4pre1-user-orgs-policy.db.test.ts \
+            tests/integration/attendance-w4pre1b-user-orgs-lifecycle.db.test.ts \
+            tests/integration/attendance-w4pre1b-user-orgs-sync-automatch.db.test.ts \
+            tests/integration/attendance-w4pre1b-user-orgs-backfill-migration.db.test.ts \
+            tests/integration/attendance-w4pre1b-directory-readiness-gate.db.test.ts \
+            tests/integration/attendance-w4pre1b-admin-users-explicit-org.db.test.ts \
             tests/integration/attendance-expiry-service.test.ts \
             tests/integration/attendance-unscheduled-reminder.test.ts \
             tests/integration/attendance-outdoor-punch.test.ts \
--- a/packages/core-backend/src/db/migrations/zzzz20260721150000_backfill_user_orgs_from_directory_links.ts
+++ b/packages/core-backend/src/db/migrations/zzzz20260721150000_backfill_user_orgs_from_directory_links.ts
@@ -16,7 +16,8 @@
  * (`readOrgDirectoryReadiness`, attendance-admin.ts) and every bind-shaped writer in this line use
  * to mean "this account currently counts as bound". Never a client-supplied or guessed org.
  *
- * Idempotency + non-resurrection (owner: "绝不复活已 is_active=false 的行"): `ON CONFLICT (user_id,
+ * Idempotency + non-resurrection (this migration must never resurrect an already-deactivated row
+ * — the owner's item-C text is "backfill 迁移（幂等、不复活已失活行）"): `ON CONFLICT (user_id,
  * org_id) DO NOTHING` — this migration ONLY inserts rows that do not exist yet. A row that already
  * exists, active OR already deactivated by the item-B unbind/rebind/archive logic, is left
  * completely untouched (no UPDATE branch at all, unlike the live bind writers' `DO UPDATE SET
--- a/packages/core-backend/src/db/migrations/zzzz20260721150000_backfill_user_orgs_from_directory_links.ts
+++ b/packages/core-backend/src/db/migrations/zzzz20260721150000_backfill_user_orgs_from_directory_links.ts
@@ -1,0 +1,71 @@
+/**
+ * W4-PRE-1b (owner CHANGES_REQUESTED on the W4 re-ratify PR #4522, 2026-07-21, item C):
+ * real-stock backfill for `user_orgs`.
+ *
+ * W4-PRE-1 (#4521) and W4-PRE-1b (this line) closed every PRODUCTION write path that maintains
+ * `user_orgs` going forward, but neither touched EXISTING data: a user who was admitted/bound to
+ * a directory account BEFORE those write paths existed has an `active directory_account_links`
+ * row (proof they belong to an org) yet may have no `user_orgs` row at all — exactly the "already
+ * linked users may have no membership" half of the owner's P1 finding. This migration is the
+ * one-time repair for that stock.
+ *
+ * Population: `directory_account_links.local_user_id` for every LINKED row whose
+ * `directory_accounts` is active, joined through `directory_accounts.integration_id ->
+ * directory_integrations.id` to resolve the KNOWN AUTHORITATIVE org — the exact same
+ * (link_status='linked' AND directory_accounts.is_active=true) predicate the S7-5 readiness read
+ * (`readOrgDirectoryReadiness`, attendance-admin.ts) and every bind-shaped writer in this line use
+ * to mean "this account currently counts as bound". Never a client-supplied or guessed org.
+ *
+ * Idempotency + non-resurrection (owner: "绝不复活已 is_active=false 的行"): `ON CONFLICT (user_id,
+ * org_id) DO NOTHING` — this migration ONLY inserts rows that do not exist yet. A row that already
+ * exists, active OR already deactivated by the item-B unbind/rebind/archive logic, is left
+ * completely untouched (no UPDATE branch at all, unlike the live bind writers' `DO UPDATE SET
+ * is_active = EXCLUDED.is_active`). Re-running this migration (or running it against a DB where a
+ * later sync/bind has already produced the row) is a no-op for every already-covered pair.
+ *
+ * Deliberately NOT filtered by `users.is_active` — same precedent as every live writer in this
+ * line (admin-users.ts's W4-PRE-1 write, directory-sync.ts's admission/bind writers): membership
+ * EXISTENCE is written as active regardless of the user's own active flag; the RD-3 dual-is_active
+ * read filter (`user_orgs.is_active=true AND users.is_active=true`) is what excludes a deactivated
+ * user from every count and gate, not a write-time filter here.
+ *
+ * Pure data migration — zero schema change, zero index change. `user_orgs` and its primary key
+ * were already established by `zzzz20260114110000_create_user_orgs_table.ts`; this file only
+ * needs `checkTableExists` guards for the SOURCE tables (`directory_account_links`,
+ * `directory_accounts`, `directory_integrations`, `user_orgs` itself) so a fresh DB or a DB where
+ * directory-sync tables have not landed yet skips cleanly instead of throwing.
+ */
+import type { Kysely } from 'kysely'
+import { sql } from 'kysely'
+import { checkTableExists } from './_patterns'
+
+export async function up(db: Kysely<unknown>): Promise<void> {
+  const requiredTables = ['user_orgs', 'directory_account_links', 'directory_accounts', 'directory_integrations']
+  for (const table of requiredTables) {
+    if (!(await checkTableExists(db, table))) return
+  }
+
+  await sql`
+    INSERT INTO user_orgs (user_id, org_id, is_active)
+    SELECT DISTINCT l.local_user_id, i.org_id, true
+    FROM directory_account_links l
+    JOIN directory_accounts a ON a.id = l.directory_account_id
+    JOIN directory_integrations i ON i.id = a.integration_id
+    WHERE l.link_status = 'linked'
+      AND l.local_user_id IS NOT NULL
+      AND a.is_active = true
+    ON CONFLICT (user_id, org_id) DO NOTHING
+  `.execute(db)
+}
+
+/**
+ * Deliberately a NO-OP down migration. This is a repair backfill, not a schema change — there is
+ * no way to tell a row this migration inserted apart from one a later live writer (bind, admit,
+ * auto-match) legitimately inserted for the exact same (user_id, org_id) pair after `up()` ran,
+ * so "undoing" it by DELETE would risk deleting real, currently-load-bearing membership rows
+ * created by ordinary product usage after this migration applied. `zzzz20260114110000`'s own
+ * one-time backfill sets the same precedent (no reverse data migration).
+ */
+export async function down(_db: Kysely<unknown>): Promise<void> {
+  // Intentionally empty — see doc comment above.
+}
--- a/packages/core-backend/src/directory/directory-sync.ts
+++ b/packages/core-backend/src/directory/directory-sync.ts
@@ -3760,6 +3760,20 @@ export async function syncDirectoryIntegration(
           linkedUserIdByExternalUserId.set(account.external_user_id, localUserId)
         }
 
+        // W4-PRE-1b item A: an `external_identity` re-match confirms a PRE-EXISTING user against
+        // this account without going through `applyDirectoryAccountBindInTransaction` (that
+        // helper only runs for the manual-bind / admit call sites) — this is the "auto-match"
+        // writer the owner named explicitly. Gated on the ACTUAL outcome (`linkStatus ===
+        // 'linked' && localUserId`), not on `matchStrategy`, so it also covers `auto_admit`
+        // (already-active via `createDirectoryAdmittedUserInTransaction`'s own call into that
+        // helper — this second upsert is then an idempotent no-op, not a double-write) without
+        // hand-maintaining a strategy allowlist. `pending` email/mobile candidate matches are
+        // correctly excluded: they never held an ACTIVE membership through this account (item A
+        // activates membership only for `linked`), so no upsert is due until a human confirms.
+        if (linkStatus === 'linked' && localUserId) {
+          await upsertActiveUserOrgMembership(client, { userId: localUserId, orgId: integration.org_id })
+        }
+
         await client.query(
           `INSERT INTO directory_account_links (
              directory_account_id, local_user_id, link_status, match_strategy, created_at, updated_at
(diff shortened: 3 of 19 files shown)


zensgit:
**W4-PRE-1b 门禁记录（完整链）：正门 REQUEST_CHANGES → 修复 → 独立复核 KILLED-CONFIRMED ⇒ 实效 APPROVE 0 P1/P2**

## 链路
Sonnet 实现（`d437b313`）→ 三镜预门 → 修复轮 10 项（含 **P1：sync-loop 稳态 upsert 会经已失活目录账号复活 membership**——加 `account.is_active` 门；**P2：并发解绑 write-skew**——`SELECT … FOR UPDATE` 真修 TOCTOU）→ **Opus 正门**（`afcfcfbe4`，报告 `/private/tmp/pr4526-review-w4-pre1b-gate.md`）：8 刀 mutation **杀 7 存活 1**，verdict=REQUEST_CHANGES 单 P2（失活助手租户谓词 `AND i.org_id=$2::text` 零覆盖，构造出跨 org stale-access 可达故障）→ 修复 `2da81d98b`（纯测试 +42：跨 org 负向回归双腿）→ **独立 Opus 复核 KILLED-CONFIRMED**（亲自动刀：恰好新用例红于 org-A `is_active` 语义断言、其余 8 绿、还原 9/9；diff 字节级确认零生产代码）。

## 正门 mutation 终表（8/8 killed，含补刀后）
neuter bind 写者 / 无条件失活（sibling 检查去除）/ 失活助手整体 neuter（**stale-access 承重刀：旧成员 403 腿红**）/ backfill 复活保护 / S7-5 门双活 / FOR UPDATE 锁（TOCTOU）/ sync-loop `account.is_active` 门 / **租户谓词 `AND i.org_id=$2`（补刀后 killed）**。

## 实跑（终态）
七案 F1-F7 + 跨 org 回归：wired 6 文件 25+1/26 绿；CI attendance 步 40 文件 522/522；单测 6909；双 typecheck 绿。

## 遗留（owner 裁决项，不阻本 PR）
① item-D「org 存在」读法：validate-can-fail（已实装 404）vs auto-vivify；② DT-OPS-01 离职扫不失活 membership（真实钉钉离职主路径残口）；③ dingtalk-group-destination 两 EXISTS 可见性过滤是否同型双活。均已在 body 详述。

依线内节奏 arm 落地；合入后进入 #4522 更新阶段（canonical 措辞改真 + 角色化合同 + effectiveTime 四态恢复 + 验证 MD 扩展）→ 呈 owner 复审 + RATIFY。

PR #4524, opened 2026-07-22
feat(approval): add secure record-link form fields

zensgit:
## What

- add pinned record-link fields to approval templates and ordinary-user form filling
- enforce fresh base, sheet, row, field, template-visibility, and final approvals-write authorization at picker, publish, submit, and read-projection boundaries
- make record-permission writes transactional and keep OpenAPI/generated SDK contracts in sync
- add a real-DB submit + picker discriminator where base-read is the only changed authority dimension

## Stack and safety

- stacked on #4540; do not merge before that base
- values project fail-closed and unavailable/missing targets share one non-oracular result
- runtime flags remain default OFF

## Verification

- current head `b4ede85a2d27432c2d0837668f5a66b3f39150ce`
- fresh PostgreSQL record-link: 39/39
- same actor/template/sheet/record: submit 400 and picker 404 without base-read; both succeed after adding only `multitable:base:read`
- final composed head: fresh-DB 92/92 and backend typecheck pass

## Review state

Draft for owner review on this exact head. No runtime or production enablement is authorized.

PR #4531, opened 2026-07-22
feat(approval): update linked records from approved form values

zensgit:
## What

- Add FWB-2 `mode: update`: derive the destination base/sheet from the pinned template-version record-link field and accept only an exact `{ recordId }` form snapshot value.
- Re-check the exact persisted action identity, template/version, cross-base authority, sheet/field/row permissions, record lock, and target types inside the write transaction.
- Commit claim + record update + revision + durable outbox atomically; replay returns already-applied with no second mutation or event.
- Fix the shared own-write fallback so an empty or unrelated record-scope map cannot widen edit authority.
- Collapse missing, locked, and unwritable linked-record outcomes to one persisted error code so execution logs cannot serve as a record-existence oracle.

## Stack and gates

- Stacked on #4524 (`codex/approval-record-link-layer2-20260721`); do not merge before that base.
- `APPROVAL_FWB_WRITEBACK_ENABLED` and durable-delivery activation remain default OFF.
- Draft for owner review; no production enablement in this PR.

## Verification

- Fresh PostgreSQL 15: FWB-2 real-DB suite 15/15.
- Fresh PostgreSQL 15 composed FWB-1 helper + production + FWB-2: 33/33.
- Related unit suites: 68/68.
- Full backend no-DB Vitest config: 8747/8747, 0 failed.
- CI-equivalent backend `tsc --noEmit`: clean.
- Structural guards: cross-base, record-lock, revision disposition, raw-data projection 23/23.
- Discriminating mutations: unknown mode, exact action-key isolation, record `FOR UPDATE`, empty scope-map fallback, transaction-local grant revoke, absent-field-permission phantom, and non-oracular linked-record errors.
- Constructed two-connection races prove the executor blocks on authority locks, then refuses after the revoke/deny commits.

## Review note

ReClaude Opus reviewed the production chain at the prior head. It verified the same-transaction claim/update/revision/outbox contract and found one actionable log-oracle issue, fixed in the current head with a real-DB mutation proof. Its two remaining P2-labelled concerns are conditional residuals rather than established production exploits: production binds the transaction-scoped gate factory, and `approval_instances.form_snapshot` has one INSERT writer and no UPDATE writer in the application. They remain explicit re-review targets.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -684,6 +684,7 @@ jobs:
             tests/integration/multitable-automation-outbound-intent-realdb.test.ts \
             tests/integration/multitable-fwb-write-action-realdb.test.ts \
             tests/integration/multitable-fwb-activation-realdb.test.ts \
+            tests/integration/multitable-fwb-update-activation-realdb.test.ts \
             tests/integration/approval-attachment-gc-realdb.test.ts \
             tests/integration/approval-attachment-bind-reconcile-realdb.test.ts \
             tests/integration/approval-attachment-scan-purge-upgrade-migration.db.test.ts \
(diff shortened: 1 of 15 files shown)


zensgit:
Ordered landing rebase complete on head 1dabec220:

- replayed the two unique FWB-2 commits onto merged #4524/main
- range-diff preserves both product commits; the sole omitted hunk is the duplicate raw-projection allowlist entry already present on main
- backend `tsc --noEmit`: clean
- unit: 51/51
- fresh migrated PostgreSQL FWB update/write suites: 19/19
- flags remain OFF

Marked ready and proceeding under the owner’s ordered-landing authorization.

PR #4547, opened 2026-07-23
fix(directory): T2 lock-correctness — canonical UUID lock key + explicit READ COMMITTED for freeze-lock txns

zensgit:
T2 lock-correctness (owner order step 3). Both owner findings CONFIRMED and fixed; real-DB constructed-concurrency proofs + two mutation reds below. **UNARMED** — no auto-merge; orchestrator serializes arming.

## P1 — canonical UUID lock key

The manual sync route passes RAW `req.params.integrationId` (`admin-directory.ts:529` async branch, `:566` sync branch) into `syncDirectoryIntegration`. `directory_integrations.id` is a **uuid** column (`zzzz20260324150000_create_directory_sync_tables.ts:15`), so an uppercase id resolves the same row via uuid casting — but `acquireSourceSyncFreezeLock` hashes the RAW TEXT key (`hashtext('directory:source-sync-freeze:' || id)`, `source-sync-freeze-lock.ts:41-43`). An uppercase caller therefore hashed a **different advisory key** than the transfer side's DB-canonical `source.id` (`org-transfer-service.ts` create / refreeze) ⇒ sync↔freeze mutual exclusion silently lost ⇒ a freeze committing after the entry check could race sync's local apply (live probe: `hashtext` lower vs upper = `2116517213` vs `1496837228`).

**Fix** (`directory-sync.ts`): immediately after `getIntegrationRow` succeeds, canonicalize — parameter renamed to `rawIntegrationId`, and `const integrationId = integration.id` (DB read-back) feeds EVERYTHING downstream: entry freeze check, stale-run reclaim, run-lease claim, apply-txn lock, under-lock recheck, every write.

**Caller sweep** (all `acquireSourceSyncFreezeLock` / `sourceSyncFreezeLockKey` call sites on this branch):
- `org-transfer-service.ts:281` (`createOrgTransfer`) — uses `source.id` read back from `SELECT ... WHERE id = ANY($1::uuid[])`; a case-variant input additionally fails the strict `r.id === input.sourceIntegrationId` match and 404s. Already canonical — **verified, unchanged**.
- `org-transfer-service.ts:523` (`setOrgTransferSourceSyncFreeze`) — keyed by `source_integration_id` probed from the transfer row (DB-read). Already canonical — **verified, unchanged**.
- `directory-sync.ts` apply txn — was the raw param; now canonical (the fix).
- Scheduler trigger `directory-sync-scheduler.ts:136` passes `row.id` from its own DB reads — canonical — and now additionally flows through the entry canonicalization.

## P2 — explicit READ COMMITTED for freeze-lock transactions

`connection-pool.ts` `transaction()` issues a bare `BEGIN` (line 158). On a deployment with `default_transaction_isolation = 'repeatable read'`, the transaction snapshot is fixed by the `pg_advisory_xact_lock` SELECT itself — **before the lock is granted** — so the post-lock freeze recheck reads a pre-freeze snapshot and misses a freeze committed while waiting.

**Fix** (PB4-3 idiom, matching `local-directory-org.ts:259`): `SET TRANSACTION ISOLATION LEVEL READ COMMITTED` as the first statement after BEGIN in:
1. the sync local-apply transaction (`directory-sync.ts`),
2. `createOrgTransfer`'s freeze-taking transaction (`org-transfer-service.ts`),
3. `setOrgTransferSourceSyncFreeze`'s freeze-taking transaction (pinned BEFORE the source probe — the probe is that txn's first snapshot-taking query).

**Scope statement (deliberate):** `connection-pool.ts`'s generic bare `BEGIN` is **NOT** changed — pinning every transaction in the codebase to RC is a blast radius far beyond this ticket. The isolation pin is scoped to the three freeze-lock transactions only.

## Tests — real DB, two-connection, constructed interleavings

`tests/integration/directory-source-freeze-lock-correctness.db.test.ts` runs the **entire service pool under `default_transaction_isolation='repeatable read'`** (DATABASE_URL amended with `options=-c default_transaction_isolation=repeatable\ read` in `vi.hoisted`, before the PoolManager reads it; isolated fork per file so nothing leaks). A sentinel asserts the posture (`SHOW` + bare-BEGIN generic txn reports `repeatable read`) so the pin proof cannot go vacuous. Real `syncDirectoryIntegration` / `createOrgTransfer`, mocked DingTalk client, EMPTY tenant ⇒ any allowed apply performs the destructive absence sweep (the observable).

- **(a) P1 RACE**: holder (create-shaped stand-in) holds the **canonical**-key lock + uncommitted freeze INSERT; real sync is called with the **UPPERCASE** id. Fixed code parks — proven by `pg_blocking_pids` barrier **and** a `pg_locks` witness that the waiter's ungranted advisory tuple equals the holder's granted `(classid, objid, objsubid)` (same-key proof) — then, after the freeze commits, rolls back with `DirectorySyncFrozenByTransferError`; seeded account stays live, zero completed runs.
- **(b) P2 RACE (RR-default barrier)**: active UNFROZEN transfer (entry check passes); holder holds the lock + uncommitted `freeze=true`; sync (canonical id) parks; holder commits; post-lock recheck must see the freeze. Exact interleaving: apply txn opens → would-be snapshot at the lock SELECT → parks → freeze commits → lock acquired → recheck.
- **P2 posture**: real `createOrgTransfer` (pinned) parks on the shared key and activates the freeze under the RR-default pool.
- **Unit guards (item 5)**: DB suite — seeded row `facade00-…`, read-back via the `getIntegrationRow` SELECT shape with the uppercase id returns the canonical lowercase id; `key(readBack) === key(canonical)`, `key(upper) !== key(canonical)`, and `SELECT hashtext(k1)=hashtext(k2)` is false. No-DB guard `tests/unit/source-sync-freeze-lock-key.test.ts` pins the key fn's case-sensitivity (runs in the gating no-DB job).

## Mutation proofs — both red, exact signatures

**(i) revert canonicalization** (`const integrationId = rawIntegrationId`):
```
× P1 RACE: an UPPERCASE manual-sync id parks on the canonical transfer freeze key (same pg_locks tuple) and rolls back after the freeze commits  8067ms
  → timed out waiting for a backend blocked by holder pid 15826 (source freeze lock never engaged)
Test Files  1 failed (1) / Tests  1 failed | 5 passed (6)
```
(the uppercase caller hashes its own key and never parks — pre-fix behavior reproduced)

**(ii) drop `SET TRANSACTION ISOLATION LEVEL READ COMMITTED` from the sync local-apply txn**:
```
× P2 RACE (RR-default barrier): a freeze committed WHILE the apply waits on the lock is seen by the post-lock recheck
  → AssertionError: expected { integration: { …(17) }, …(2) } to be an instance of DirectorySyncFrozenByTransferError
× P1 RACE: … → same AssertionError (recheck misses the freeze after parking, too)
Test Files  1 failed (1) / Tests  2 failed | 4 passed (6)
```
(the sync COMPLETES against a frozen source — the recheck read the pre-freeze RR snapshot; test (a) reds as well, consistent with the same mechanism)

Both mutations reverted; committed tree is the fixed code (working tree diff vs HEAD empty before push).

## Green evidence (local, real DB `metasheet_test`, PG 15.17)

- New suite: 6/6 passed (764ms); regression: existing `directory-org-transfer-source-freeze.db.test.ts` 16/16 passed; combined re-run at rebased HEAD: 22/22.
- `tests/unit/source-sync-freeze-lock-key.test.ts` + `tests/unit/admin-directory-routes.test.ts`: 103/103.
- `tsc --noEmit` clean; extended `t2-source-freeze-ci-wiring.test.mjs` 4/4 (two-point wiring for BOTH freeze suites: vitest.config.ts exclude + plugin-tests.yml whole-file real-DB step).

## CI wiring

New DB suite added to `vitest.config.ts` exclude (no-DB job cannot skip-green it) and to the plugin-tests.yml directory real-DB step as a whole file, directly under the original T2 freeze suite; the wiring contract test now loops over both files.

Rebased onto `ee39a13eb` (#4545, vitest.config.ts exclude for w4pre1d — clean auto-merge, adjacent exclude entries) and fully re-verified at the rebased HEAD.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -728,6 +728,7 @@ jobs:
             tests/integration/directory-tenant-change-immutable.db.test.ts \
             tests/integration/directory-sync-orchestration.db.test.ts \
             tests/integration/directory-org-transfer-source-freeze.db.test.ts \
+            tests/integration/directory-source-freeze-lock-correctness.db.test.ts \
             tests/integration/directory-account-external-key-collision-mechanism.db.test.ts \
             tests/integration/directory-sync-schedule-timezone.db.test.ts \
             tests/integration/directory-local-provider-bootstrap.db.test.ts \
--- a/packages/core-backend/src/directory/directory-sync.ts
+++ b/packages/core-backend/src/directory/directory-sync.ts
@@ -3493,7 +3493,7 @@ async function claimDirectorySyncRun(
 }
 
 export async function syncDirectoryIntegration(
-  integrationId: string,
+  rawIntegrationId: string,
   triggeredBy: string,
   triggerSource: 'manual' | 'scheduler' = 'manual',
   /**
@@ -3508,8 +3508,18 @@ export async function syncDirectoryIntegration(
   autoAdmissionOnboardingPackets: DirectoryAutoAdmissionOnboardingPacket[]
 }> {
   const governedUserIds = new Set<string>()
-  const integration = await getIntegrationRow(integrationId)
+  const integration = await getIntegrationRow(rawIntegrationId)
   if (!integration) throw new Error('Directory integration not found')
+  // T2 lock-correctness P1: the manual-sync route passes req.params.integrationId VERBATIM
+  // (admin-directory.ts). `directory_integrations.id` is a uuid column, so a case-variant
+  // (e.g. uppercase) id still resolves THIS row via uuid casting — but the shared source
+  // freeze advisory lock hashes the RAW TEXT key (`hashtext(sourceSyncFreezeLockKey(id))`),
+  // so an uppercase caller would hash a DIFFERENT lock key than the transfer side's
+  // DB-canonical `source.id` and silently lose sync↔freeze mutual exclusion (a freeze
+  // committing after the entry check could race this sync's local apply). Canonicalize to
+  // the DB-read-back id here, and use it for EVERYTHING downstream: entry freeze check,
+  // stale-run reclaim, run-lease claim, apply-txn lock, under-lock recheck, and every write.
+  const integrationId = integration.id
 
   const config = parseIntegrationConfig(integration)
   // T2 (§12.2): cheap freeze gate BEFORE the lease claim — a freeze already active at entry
@@ -3596,7 +3606,18 @@ export async function syncDirectoryIntegration(
     const autoAdmissionOnboardingPackets: DirectoryAutoAdmissionOnboardingPacket[] = []
 
     await transaction(async (client) => {
-      // T2 P1: shared source freeze lock is the FIRST operation of the local-apply transaction.
+      // T2 lock-correctness P2 (PB4-3 idiom — see local-directory-org.ts's reparent guard): the
+      // pool wrapper issues a bare BEGIN, so on a deployment where default_transaction_isolation
+      // is 'repeatable read' this transaction's snapshot would be taken by the
+      // pg_advisory_xact_lock SELECT itself — BEFORE the lock is granted — and the post-lock
+      // freeze recheck below would read a pre-freeze snapshot, missing a freeze that committed
+      // while we waited on the lock. Pin READ COMMITTED as the FIRST statement (SET TRANSACTION
+      // must precede the transaction's first query) so the recheck takes a fresh per-statement
+      // snapshot. Under the production RC default this is a no-op; its mechanism is proven in
+      // directory-source-freeze-lock-correctness.db.test.ts (RR-default pool harness).
+      await client.query('SET TRANSACTION ISOLATION LEVEL READ COMMITTED')
+      // T2 P1: shared source freeze lock is the first REAL operation of the local-apply
+      // transaction (nothing but the isolation pin above may precede it).
       // Transfer create / freeze=true refreeze take the same key before writing freeze state.
       // Re-check under the lock before ANY directory upsert, absence sweep, membership rewrite,
       // identity/link write, local-user admission, group projection, deprovision, integration
--- a/packages/core-backend/src/directory/org-transfer-service.ts
+++ b/packages/core-backend/src/directory/org-transfer-service.ts
@@ -271,6 +271,12 @@ export async function createOrgTransfer(input: CreateOrgTransferInput): Promise<
     // pass its apply-time recheck and commit directory writes after this freeze becomes active.
     // Default freeze_source_sync=true on the row makes create itself a freeze activation.
     return await transaction(async (client) => {
+      // T2 lock-correctness P2 (PB4-3 idiom): the pool wrapper issues a bare BEGIN, so under a
+      // repeatable-read default_transaction_isolation this transaction's snapshot would be
+      // frozen by the lock-acquisition SELECT itself, before the lock is granted. Freeze
+      // writers pin READ COMMITTED as the first statement — same as the sync local-apply
+      // side — so post-lock reads observe state committed while waiting on the shared key.
+      await client.query('SET TRANSACTION ISOLATION LEVEL READ COMMITTED')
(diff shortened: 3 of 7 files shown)


zensgit:
Evidence (raw local runs; real DB `postgresql://postgres@localhost:5432/metasheet_test`, PostgreSQL 15.17 Homebrew, worktree branched off `f55d99e12`, rebased onto `ee39a13eb`, head `51d746d85`).

**1. Mechanism probe (node + pg, before writing tests)**
```
URL: postgresql://postgres@localhost:5432/metasheet_test?options=-c%20default_transaction_isolation%3Drepeatable%5C%20read
default iso: repeatable read
bare BEGIN txn iso: repeatable read
hashtext lower vs upper: 2116517213 1496837228 equal: false
uuid read-back with uppercase text param: canonical-lowercase
```

**2. Green run — new suite (fixed code, commit 00d30c714 pre-rebase)**
```
✓ sentinel: DATABASE_URL is set (DB-backed lane must not silently skip)
✓ sentinel: the service pool REALLY runs repeatable-read default — a bare-BEGIN generic transaction is RR
✓ unit guard (P1 key contract): fix-path canonicalization maps an uppercase route id onto the SAME lock key as the DB-canonical id
✓ P1 RACE: an UPPERCASE manual-sync id parks on the canonical transfer freeze key (same pg_locks tuple) and rolls back after the freeze commits
✓ P2 RACE (RR-default barrier): a freeze committed WHILE the apply waits on the lock is seen by the post-lock recheck
✓ P2 posture: the REAL createOrgTransfer (pinned RC) still parks on the shared key and activates the freeze under an RR-default pool
Test Files  1 passed (1) / Tests  6 passed (6) / Duration 764ms
```

**3. Mutation (i) — `const integrationId = rawIntegrationId` (pre-fix raw-key behavior reproduced)**
```
× P1 RACE: an UPPERCASE manual-sync id parks on the canonical transfer freeze key (same pg_locks tuple) and rolls back after the freeze commits  8067ms
  → timed out waiting for a backend blocked by holder pid 15826 (source freeze lock never engaged)
Test Files  1 failed (1) / Tests  1 failed | 5 passed (6)
```

**4. Mutation (ii) — `SET TRANSACTION ISOLATION LEVEL READ COMMITTED` dropped from the sync local-apply txn**
```
× P1 RACE: … → AssertionError: expected { integration: { …(17) }, …(2) } to be an instance of DirectorySyncFrozenByTransferError
× P2 RACE (RR-default barrier): a freeze committed WHILE the apply waits on the lock is seen by the post-lock recheck
  → AssertionError: expected { integration: { …(17) }, …(2) } to be an instance of DirectorySyncFrozenByTransferError
Test Files  1 failed (1) / Tests  2 failed | 4 passed (6)
```
Both mutations reverted afterwards; `git status --short` + `git diff` vs HEAD empty before push.

**5. Regression + gates (rebased HEAD `51d746d85`)**
```
directory-source-freeze-lock-correctness.db.test.ts + directory-org-transfer-source-freeze.db.test.ts:
  Test Files  2 passed (2) / Tests  22 passed (22)
tests/unit (source-sync-freeze-lock-key + admin-directory-routes): 103 passed (103)
tsc --noEmit: clean
node --test scripts/ops/t2-source-freeze-ci-wiring.test.mjs: tests 4 / pass 4 / fail 0
```

**6. Rebase note**: origin/main moved `f55d99e12` → `ee39a13eb` (#4545, vitest.config.ts exclude entry for an attendance w4pre1d suite) during the lane. Branch had no remote counterpart (`git ls-remote` empty), rebase auto-merged cleanly (adjacent exclude entries in vitest.config.ts), full re-verification at rebased HEAD before push. origin/main re-checked immediately before push: still `ee39a13eb`.

UNARMED — not arming auto-merge; orchestrator serializes arming per owner order.

PR #4548, opened 2026-07-23
test(attendance): exact save-and-restore isolation for the shared-DB system_configs 'attendance.settings' row (W4 §7.5 finding)

zensgit:
## What

Fixes the W4-recorded shared real-DB `system_configs` test pollution (owner order step 5a; finding recorded in `docs/development/attendance-vnext-wave4-w40-w42-development-verification-20260722.md` §7.5 item 2 — the sibling item 1, the vitest exclude, already landed as #4545).

`plugin-tests.yml`'s attendance step runs all attendance integration suites serially against ONE shared Postgres. Eight suites write the deployment-wide `system_configs` row (key `attendance.settings`) through `PUT /api/attendance/settings`; a test that failed before its in-test restore (or restored a hardcoded baseline instead of the prior state) poisoned every later settings-sensitive assertion — the recorded leak was `attendance-plugin.test.ts` shiftCompliance state bleeding into `attendance-schedule-dispatch.test.ts`, and the W4-0 readiness ④ posture reads this exact row (`AttendanceSetupReadinessAggregate.ts:394`).

## How

New shared helper `packages/core-backend/tests/utils/attendance-settings-row.ts`:

- **Snapshot once per suite** (`beforeAll`), **exact-restore after EVERY test** (`afterEach`), **including the absent-row case** — DELETE if the row did not exist at suite entry.
- Snapshot reads `value::text` — the one representation that round-trips on BOTH schemas: on 038-JSONB databases node-pg auto-parses a bare `value` into a JS object (`String(obj)` → `'[object Object]'`, the known `JSON.parse` trap; the plugin's own `loadSettings` at `plugins/plugin-attendance/index.cjs:13719` has exactly this hazard), while legacy bootstraps have a TEXT column (verified live: the local `metasheet_test` column is `text`, CI's 038-migrated column is `jsonb` — migration `038_config_and_secrets.sql:8`).
- Restore is a plain-parameter upsert (`ON CONFLICT (key) DO UPDATE`; `key` is UNIQUE on both schemas) — Postgres implicitly casts the text back to jsonb where applicable.

Wired into all eight suites that touch the row (grep-complete set): `attendance-plugin`, `attendance-schedule-dispatch`, `attendance-setup-readiness-w4-0.db`, `attendance-shift-swap`, `attendance-outdoor-punch`, `attendance-files-acl`, `attendance-makeup-punch-policy`, `attendance-result-edit`. Suites running the plugin in-process also drop its 60s module-level settings cache after the restore (`resetAttendanceSettingsCacheForTests`, the plugin's own test seam) so the next test re-reads the real row. The W4-0 spec's private snapshot/restore helpers now delegate to the shared helper (its old `SELECT value` typed-as-string snapshot was object-typed on JSONB schemas) and gain `afterEach` granularity instead of `afterAll`-only.

**No mock-out**: the W4-0 §9 G5 ④-posture assertions still write and read real `system_configs` rows (positive + negative controls unchanged); the restore runs after each assertion.

## Proof (local real DB, `metasheet_test`, column type `text`, dirty machine-written entry row present)

| condition | origin/main (f55d99e12) | this branch | leak after run |
| --- | --- | --- | --- |
| `attendance-plugin.test.ts`, dirty entry row | **9 failed** / 158 | **4 failed** (strict subset) | main: row left behind; mine: row md5 byte-identical (`42aa044d…`) |
| `attendance-plugin.test.ts`, entry row ABSENT (CI-shaped) | **2 failed** / 158 | **1 failed** (strict subset) | main: row created and left; mine: row count back to 0 (absent-row DELETE path proven) |
| chained polluter → W4-0 → victim (one vitest invocation, as CI runs them) | — | W4-0 **24/24**, schedule-dispatch **17/17** pass immediately after the polluter; row md5 invariant | — |

Other five suites: 91/91 pass, row md5 invariant. Zero new failures introduced in any measured condition; 5 pre-existing local failures fixed.

## Pre-existing local failures registered (NOT fixed here, all reproduce on origin/main with the same DB state)

1. GET→spread→PUT round-trip tests 400 when the entry row carries a machine-written `holidaySync.lastRun` with `error: null`: the runtime writer stores `error: null` on success, but the PUT zod schema rejects null (`Expected string, received null` at `holidaySync.lastRun.error`). Product seam: an admin client echoing GET output back through PUT would hit the same 400 after any holiday sync. Owner may want a runtime ticket; runtime is out of scope for this test-only PR.
2. The auto-shift auto-write specs are residue-dependent on long-lived dev DBs (accumulated candidates/scopes inflate `appliedCount`); they fail on main with both dirty and clean entry rows locally. Fresh-DB CI is unaffected.

UNARMED — no merge, no auto-merge. Evidence comment follows.


zensgit:
Evidence (local real DB `postgresql://postgres@localhost:5432/metasheet_test`, reachability probed via psql before every run; `system_configs.value` column type there = `text`; CI schema = `jsonb` per `migrations/038_config_and_secrets.sql:8`, not excluded by plugin-tests.yml `MIGRATION_EXCLUDE`).

**Entry-row fingerprint before any run**: `md5(value::text) = 42aa044d15b6b1a4689eabea77cc211f`, length 3328 (a machine-written dev leftover: `holidaySync.lastRun` present with `error: null`, `shiftCompliance` warn baseline).

**A/B 1 — dirty entry row, `attendance-plugin.test.ts` (158 tests)**
- origin/main `f55d99e12` (detached baseline worktree, same DB): `9 failed | 149 passed`, exit 1.
- this branch: `4 failed | 154 passed`; `comm -13 fails-baseline fails-mine` = empty (strict subset; 5 pre-existing local failures fixed, 0 new).
- row md5 after my run: `42aa044d15b6b1a4689eabea77cc211f` (byte-identical).

**A/B 2 — entry row DELETEd first (CI-shaped fresh state), same file**
- origin/main: `2 failed | 156 passed` AND left a row behind (count 1 after run — the pollution, demonstrated live).
- this branch: `1 failed | 157 passed` (subset again) AND row count back to `0` after the run — absent-row DELETE restore path proven.

**Recorded-finding repro chain (one vitest invocation, CI file order): polluter → W4-0 → victim**
- `attendance-setup-readiness-w4-0.db.test.ts`: 24/24 pass (④ posture G5 positive/negative controls still run against real rows).
- `attendance-schedule-dispatch.test.ts` (recorded victim): 17/17 pass immediately after the polluter file.
- row md5 after the chained run: `42aa044d15b6b1a4689eabea77cc211f`.

**Other five wired suites** (`attendance-shift-swap`, `attendance-outdoor-punch`, `attendance-files-acl`, `attendance-makeup-punch-policy`, `attendance-result-edit`): `91 passed (91)`, row md5 invariant.

**Remaining 4 local failures** (dirty-row condition) all reproduce on origin/main with the same DB; causes registered in the PR body (machine-written `holidaySync.lastRun.error=null` fails the PUT zod schema on GET→spread→PUT tests; auto-shift auto-write specs are dev-DB residue-dependent). The shared dev DB row was restored from a pre-run dump after the baseline runs (md5 verified back to `42aa044d…`).

Anchors: helper `packages/core-backend/tests/utils/attendance-settings-row.ts`; polluter hooks `attendance-plugin.test.ts` (beforeAll snapshot after table checks; afterEach exact-restore; beforeEach cache reset pre-existing); victim hooks `attendance-schedule-dispatch.test.ts` (snapshot after `pool = new Pool(...)`; afterEach restore + `resetAttendanceSettingsCacheForTests` via CJS require cache); W4-0 spec now delegates `snapshotSettings`/`restoreSettings` to the shared helper and adds file-level afterEach. Finding source: `docs/development/attendance-vnext-wave4-w40-w42-development-verification-20260722.md` §7.5 item 2.

STRICTLY UNARMED. Raw data only; no merge action taken or scheduled.

PR #4569, opened 2026-07-23
feat(attendance): add preview-only shift segments

zensgit:
## What

Delivers the backend/schema half of #4556 W3 under the ratified design lock and W3 safety erratum:

- normalized 1..3 shift segments with replay-safe legacy backfill and fail-closed down migration
- canonical transactional shift create/update/read/delete service with legacy envelope compatibility
- multi-segment authoring remains preview-only; every reference producer returns typed 422 with zero writes while W4 calculation is unavailable
- shared delete/reference lock protocol and typed 409 blockers for durable references
- cancelled/rejected historical evidence survives shift deletion with neutral labels and no raw UUID fallback
- dispatch target FK changes to nullable `ON DELETE SET NULL`, including partial-migration repair
- OpenAPI/generated SDK/parity and CI run-list wiring

The ordered segment editor is intentionally a separate stacked W3 UI PR so this security-heavy backend diff can be reviewed independently. This PR does not enable the reserved calculation flag and does not implement W4 calculations.

## Verification

- fresh DB migration from zero: PASS; pending migrations 0
- `TZ=UTC ... pnpm --filter @metasheet/core-backend test:integration:attendance`: 13 files / 305 tests PASS on a fresh DB
- W3 focused real-DB suites: 39/39 PASS
- backend units: shift service 24/24; admin users 65/65
- web regressions: 130/130; backend/web typecheck PASS
- OpenAPI parity: 13/13 PASS
- `git diff --check`: PASS

Mutation proofs all went red independently:

1. future env flag treated as implemented -> unit and real-DB reference writer red
2. zero-duration legacy envelope allowed -> pre-DDL migration leg red
3. rename normalization moved outside transaction -> forced post-normalization DB failure leg red
4. orphan dispatch trusts metadata UUID -> raw-ID redaction leg red
5. parent index ownership check removed -> pre-existing index preservation leg red
6. partial SET NULL + NOT NULL treated as complete -> repair leg red
7. admin onboarding parent lock removed -> ordering leg red
8. admin onboarding 2-segment guard weakened -> real-DB zero-write leg red

## Boundaries

- No authoritative segment calculation, record-segment snapshots, flex runtime, deployment, or flag transition.
- Issue #4556 remains open for W3 UI and W4-W8.


zensgit:
Exact-head re-review closure for `78aecb770dd4566c499698518cdda5a47d514fa5`:

- Independent adversarial re-review: **APPROVE, 0 P1 / 0 P2 / 0 P3**.
- Closed prior P2: rejected/cancelled shift-swap deleted-reference response now matches OpenAPI/generated SDK contract (nullable IDs plus values-safe label/status), with parity coverage.
- Closed strict-time contract drift: runtime rejects seconds while legacy/DB reads remain compatible; `AttendanceShiftSegmentInput.startTime/endTime` now carry the identical strict `HH:MM` pattern in OpenAPI generated artifacts.
- Verification: service `24/24`, parity `14/14`, OpenAPI validate and codegen guard green. Removing `startTime.pattern` makes parity exactly `13/14`, proving the guard is load-bearing.
- Worktree clean; no runtime flag, deployment, or later-wave authorization implied.

Content gate is satisfied. Merge remains contingent on fresh required checks.

PR #4578, opened 2026-07-24
fix(directory): point grant/membership writes at the tables that exist (T1–D7 follow-up)

zensgit:
## What

Four merged statements write columns that **no migration creates**, each behind a `.catch`, so every failure was invisible.

| Phantom column | Written by |
|---|---|
| `user_external_identities.grant_enabled` | T3 activate (#4574); D7 preview + D7 restore (#4575) |
| `user_orgs.updated_at` | T3 activate (#4574); D7 restore (#4575) |

`user_orgs` is `(user_id, org_id, is_active, created_at)`, PK `(user_id, org_id)`. The DingTalk grant lives in `user_external_auth_grants` — the table `dingtalk-oauth.ts` reads at login.

## Why it is severe, not cosmetic

Inside a transaction a failed statement poisons the connection. Catching the rejection does not contain the failure — it relocates it, as `25P02`, onto the next innocent statement. Measured on `main`:

- **T3 activate with an `orgId` or `enableDingTalkGrant` cannot succeed at all** — the membership INSERT aborts the transaction and it dies at COMMIT. Only the degenerate call (no org, no grant) survives.
- The "schema variance" fallback under that INSERT is **structurally unreachable**: a poisoned transaction rejects the fallback too.
- D7 preview can **never** show a grant effect, whatever the truth is.
- Restoring any event carrying a membership **or** grant effect aborts — that is **every real deprovision event**, since every candidate carries a membership effect.
- The grant drift gate was satisfied **by its own query failing**: the error became "no grant", which reads as "matches `after_active=false`", so `DRIFT_CONFLICT` could never fire on that leg.

## Verification

- New real-DB suite `directory-deprovision-grant-table.db.test.ts` — **4/4 green** with the fix, **4/4 red** with `src` reverted to `main` (mutation control: the tests are load-bearing, and all four defects are real on `main`).
- Includes a discriminating pair: the `user`-leg drift gate correctly refuses (control), the `grant`-leg gate must too (subject).
- Added to the **plugin-tests explicit run-list** — that list is hand-written, so a new `.db.test.ts` is not auto-collected. Collection proven by running it alongside `directory-deprovision-selection.db.test.ts` (neighbour 10/10 still green).
- `tsc --noEmit` clean; four related unit suites 32/32.
- Assertions are on observable end state (is the person actually granted / actually a member again), never on SQL text.

Why unit tests missed all of it: a stub client answers whatever SQL it is handed, so a statement naming a nonexistent column looks identical to one that does not.

## Scope

Touches no deprovision writer and no `effect_type` — **independent of the ledger-schema decision still open on the lifecycle line**. Enables no flag.

Draft, not for merge without owner review. Related findings: `docs/development/dingtalk-lifecycle-postmerge-findings-20260724.md` on `claude/dingtalk-d4-d5-writer-ledger-wiring-20260724`.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -784,6 +784,7 @@ jobs:
             tests/integration/directory-sync-preview-apply-parity.db.test.ts \
             tests/integration/directory-sync-alert-coverage.db.test.ts \
             tests/integration/directory-deprovision-selection.db.test.ts \
+            tests/integration/directory-deprovision-grant-table.db.test.ts \
             tests/integration/directory-sync-run-lease.db.test.ts \
             tests/integration/directory-tenant-change-immutable.db.test.ts \
             tests/integration/directory-sync-orchestration.db.test.ts \
--- a/packages/core-backend/src/auth/user-activate.ts
+++ b/packages/core-backend/src/auth/user-activate.ts
@@ -119,34 +119,40 @@ export async function activatePendingUser(input: ActivateUserInput): Promise<Act
 
     // Active membership for current org when provided.
     if (input.orgId) {
+      // `user_orgs` is (user_id, org_id, is_active, created_at) with PK (user_id, org_id) — there
+      // is no `updated_at`. Writing one raised `42703` and aborted this transaction, which is why
+      // the "schema variance" fallback underneath could never help: once a statement fails inside
+      // a transaction, Postgres rejects every later statement with `25P02`, the fallback
+      // included. The net effect was that activating a pending user WITH an org — the ordinary
+      // case — could not succeed at all; it died at COMMIT. Writing the real shape is the fix,
+      // and the fallback goes with it: a second guess at the schema is not a substitute for
+      // matching it.
       await client.query(
-        `INSERT INTO user_orgs (user_id, org_id, is_active, created_at, updated_at)
-         VALUES ($1, $2, TRUE, NOW(), NOW())
+        `INSERT INTO user_orgs (user_id, org_id, is_active, created_at)
+         VALUES ($1, $2, TRUE, NOW())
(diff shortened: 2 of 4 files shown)


zensgit:
## Owner/stream 裁断 (Grok)

**Fork A** for D4 writer. This PR’s table-fix direction is **correct** and is superseding-landed as **#4581** with:

- membership restore **RETURNING** fail-close → `DRIFT_CONFLICT` (your P2)
- grant restore **`granted_by = restore:<admin>`** on conflict update (your P2)
- alias Auth gate + in-txn claim + cutover-status honesty
- fail-honest evidence lists + closeout doc correction

Please **close #4578 as superseded by #4581** once #4581 is green, or restack if you prefer to keep authorship — content is merged forward from `6c1a093`.

Real-DB suite you added is on #4581 and local 4/4 after ledger schema present.

zensgit:
Closing as superseded: every change here (user_external_auth_grants repoint, user_orgs real column set, evidence-api grant read/write, the directory-deprovision-grant-table.db.test.ts suite) is on current main via #4587's content (verified by content grep against origin/main — note #4587's merge commit itself is not a main ancestor because main history was rebuilt after 07-24; this branch has no merge base with current main, so closing rather than rebasing). #4579 stays open as the fork-A candidate for the D4/D3 decision, now in tension with codex #4647 (fork B).

zensgit:
Correction to my closing comment: the claim that "main history was rebuilt after 07-24" was **wrong** — an artifact of this workspace being a shallow clone (.git/shallow present). GitHub compare confirms both the old tip 8aad0ef8f and #4587's merge commit 9ff423fef ARE ancestors of current main (behind_by=0). Nothing was rewritten; the superseded-by-#4587 closure rationale stands on content, which was and remains correct.

PR #4579, opened 2026-07-24
HELD — D3 ledger hardening + D4 writer wiring (awaiting owner fork decision)

zensgit:
## ⛔ HELD — do not merge, do not duplicate

Open **only** so the lifecycle line has one visible claim on `directory-sync.ts`. Two agents are working this line; this file can have exactly one owner at a time.

## The finding this exists for

On merged `main`, running the deprovision writer with `enabled: true` — what the canary GO turns on — **fully offboards a person and writes zero evidence**. Measured on real Postgres, `main@8aad0ef8f`:

| | before | after |
|---|---|---|
| `users.is_active` | true | **false** |
| `user_orgs.is_active` | true | **false** |
| grant `enabled` | true | **false** |
| `users.access_generation` | 0 | **0** |
| ledger events / effects | 0/0 | **0/0** |

`deprovision-ledger.ts` has **no importer** on `main`. The only `INSERT INTO directory_deprovision_events` in the repo is inside that unimported file, so no code path can create an event — the D7 panel has nothing to show and restore has nothing to restore from, silently.

Note `#4574`'s PR body is accurate (it says "helpers"). The overstatement is in `dingtalk-lifecycle-line-closeout-20260724.md`, which lists D1–D6 as done.

## Owner decision this is blocked on

The landed ledger schema deviates from ratified §5.2 in ~10 places (no prerequisite UNIQUE keys, no composite FKs, no witness/policy/globally_clear columns, no CHECKs, no BEFORE INSERT link validation, `event_id` nullable, no `UNIQUE (event_id, effect_type)`).

- **(A) build to the lock** — corrective migration + locked `effect_type` enum, then wire. Ripples into D7 API/UI and ~27 real-DB call sites needing a run fixture.
- **(B) amend the lock** to match what shipped.

Recommendation: **(A)**. §5.1's argument is that application-level evidence is insufficient, and #4578 just demonstrated the cost of relying on it — a drift gate satisfied by its own query failing. DB constraints are immune to a swallowed error; application checks are not.

## What is written here (fork A)

- corrective migration to §5.2 (fails closed if the ledger is non-empty rather than discarding evidence)
- planner: locked effect-type enum + `policy` input + org-scoped/globally-clear split, so plan == the writer's decision
- ledger: runs on the caller's transaction (not its own), no swallowed errors, refuses an empty effect set
- writer: per-user `users … FOR UPDATE` in sorted id order (§7.2), globally-clear re-read **under the lock** (§7.4 — closes the write-skew gap the file's own comments recorded as deferred), plan → apply → record in the sync transaction, zero-effect ⇒ zero write
- restore: eligibility moved inside the locked transaction (it was a check-then-act on `access_generation`)

## Honest verification status

| | |
|---|---|
| migration up/down against real PG | **not run** |
| §11 real-DB matrix | **not written** |
| 27 existing call sites updated | **not done** |
| unit stub updated for new query shapes | done, **not run** |
| independent adversarial review | **owed** — also for merged #4574/#4575, which landed with only bot comments |

## Risk for the owner to accept explicitly

`assertAppliedMatchesPlan` throws on plan/applied divergence, which rolls back **the whole sync run**, not just the deprovision step. Under the mutex it can only fire when another writer mutated the access graph without taking the same lock — the D5 gap. The alternative is committing evidence known to disagree with reality. This is why **D5 must precede any deprovision canary**.

Findings + reproduction method: `docs/development/dingtalk-lifecycle-postmerge-findings-20260724.md` (in this branch).

### DIFF
--- a/docs/development/dingtalk-lifecycle-postmerge-findings-20260724.md
+++ b/docs/development/dingtalk-lifecycle-postmerge-findings-20260724.md
@@ -26,6 +26,58 @@ to restore from — with no error anywhere to indicate that.
 
 ---
 
+## 0.1 Empirical run (real Postgres, `main`'s own code)
+
+Static reads are not behaviour, so the findings below were re-established by **executing merged
+`main`**: worktree detached at `8aad0ef8f`, all migrations applied to a throwaway DB
+(`d4_probe_20260724`), then `applyDirectoryDeprovisionPolicies` / `previewDeprovisionForUser` /
+`restoreDeprovisionEvent` called directly. Probe scripts are session-scratch, not committed.
+
+### A — deprovision with `enabled: true` (what the canary GO turns on)
+
+| | before | after |
+|---|---|---|
+| `users.is_active` | true | **false** |
+| `user_orgs.is_active` | true | **false** |
+| grant `enabled` | true | **false** |
+| `users.access_generation` | 0 | **0** |
+| ledger events / effects | 0 / 0 | **0 / 0** |
+
+Writer returned `applied:true, grantsDisabledCount:1, usersDeactivatedCount:1`. So the person
+**was** fully deprovisioned, and **nothing was recorded**. That is the whole finding, measured
+rather than inferred.
+
+### B — D7 preview vs a grant that is genuinely ON
+
+Grant row in DB: `enabled = true`. Preview plan effects:
+`["clear_user_orgs","set_user_inactive"]` — **no `disable_dingtalk_grant`**. The preview cannot
+represent the grant, because it reads `user_external_identities.grant_enabled`, and
+`information_schema` confirms that column does not exist on the migrated database.
+
+### C — restore, with controls
+
+Same event shape each time; only the effect's *leg* and the drift condition change.
+
+| Case | Expected | Actual |
+|---|---|---|
+| **Control 1** — user leg, no drift | success, user reactivated | **SUCCESS**, `is_active` → true |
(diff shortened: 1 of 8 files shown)


zensgit:
## Owner/stream 裁断

**#4579 remains HELD.**

- Fork **A (rebuild to lock)** for D4 `directory-sync` ledger wiring.
- Do **not** merge as a single window: split D3 migration (replay-safe, no empty-ledger gate on every up), D4 ledger write, D5 mutex/generation, D6 restore, D7 Apply≈Plan.
- Ghost-column / `.catch` class is being fixed on **#4581** (app paths), independent of this HELD PR.

Canary on `DIRECTORY_DEPROVISION_ENABLED` stays **hard NO-GO** until D4 emits real ledger events.

zensgit:
Closing per the 2026-08-07 review-of-review: both this branch and codex #4646/#4647 are the SAME fork (A — DB-authoritative §5.2 hardening); #4646 carries the more complete migration (zzzz20260728100000, migration-replay green) plus the Rev 4.3 erratum this branch's planner had silently implemented without flagging as a contract change. Single D3/D4 line = #4646 → #4647. My earlier 'fork B' attribution to the codex stack was wrong (drawn from one PR of a stack + an incomplete search listing). The empirical findings in this branch's docs/development/dingtalk-lifecycle-postmerge-findings-20260724.md remain valid as behaviour evidence (content-based, unaffected by the shallow-clone ancestry artifact) and feed the adversarial-review checklist for #4646/#4647.

PR #4553, opened 2026-07-23
feat(gip): A-wave — profile certification schema + compliance harness + read-only qualification spike (latent)

zensgit:
## 范围（严格 = GIP-D0 ratify 解锁的三件，全 latent、零 runtime 接线）

- **`gip-profile-certification-contracts.cjs`**：冻结词表（四维 source schema + apply 独立轴 + 角色/时序/状态词表 + 错误词表三层 pin）；`CertifiedReadActionProfile`/`CertifiedApplyProfile` 规范化（闭合 shape；applyMode/recovery 声明按名拒绝；SEALED_EXPORT⇒SIGNED_MANIFEST、CHANGE_FEED⇒VERSION_PIN、DURABLE_TOKEN⇒durable anchor、BOUNDED_READ⇒SINGLE_REQUEST）；恢复策略矩阵推导（含企业水位续，scale-D0 §2 表第 5 行）；evidence 双向闭合；EXTERNAL_WRITE_TARGET v1 恒拒
- **`gip-profile-compliance-harness.cjs`**：认证电池 C1-C11e（16 检查，mutant 自派生，values-free 报告）——绿=必要非充分，**每个 profile 仍独立过自己的认证门**
- **`gip-binding-qualification-spike.cjs`**：`probeBindingQualification`（事务外，SELECT-only 运行时强制含 INTO/setval/advisory-lock/FOR UPDATE 拒绝，排序键唯一性负检，values-free evidence）/ `verifyBindingQualification`（纯本地零 I/O，digest 单权威 `actionProfileVersion` 绑定防跨对象复用，strict UTC ISO 过期⇒`QUALIFICATION_EXPIRED` 要求重新 Preflight）
- 状态戳（owner 指令）：GIP-D0→**RATIFIED**@d58ec38f4（窄解锁范围照录）；MR-D0-A1→**RATIFIED 选 A**（DISCLOSE_ONLY，不叠 SLO）；scale-D0 **保持 PROPOSED**（无独立 ratify，显式注明）

**不含**：任何具体 profile 认证（`bridge.*`/`sql.*`/`file.*` 均未出现在可规范化路径，fixtures 仅 `fixture.*`）、profile registry、D2 runtime、外部写回、#4437 改轨、rollout。

## 验证

[...]

PR #4580, opened 2026-07-24
fix(data-source): adapter read-boundary policies — OFFSET ordering + MySQL A5 bound + cross-adapter conformance (DRAFT)

zensgit:
**Draft — submitted for review, NOT for merge yet. One runtime-reachability fact still gates it (see "Before this can merge").**

## Three defects, one systemic cause

**1. Unordered OFFSET pagination (all three SQL adapters).** SQL gives no row-order guarantee without `ORDER BY`, so successive `LIMIT n OFFSET k` pages may overlap and leave holes. MSSQL previously fabricated `ORDER BY (SELECT NULL)`, which satisfied SQL Server syntax without supplying an ordering guarantee.

**2. MySQLAdapter had no A5 row bound.** An omitted limit produced an unbounded whole-table read, and an over-max limit was served verbatim. MySQL now uses the same adapter-layer `resolveEffectiveLimit` contract as Postgres and MSSQL.

**3. Invalid MSSQL offsets could bypass A5 entirely.** `Math.floor(Number(offset))` turned values such as `-1` and `NaN` into a state where neither `TOP` nor `OFFSET/FETCH` was emitted:

```sql
SELECT * FROM [t]
```

That contradicted the adapter-layer A5 guarantee even though the public REST route validates offsets. Direct internal callers reach the adapter chokepoint and must remain bounded too.

**Systemic cause:** boundary policies were tested per adapter, ad hoc. The earlier result-boundary suite did not import MySQL, and no cross-adapter offset-validity contract existed.

## The fix

- `resolveEffectiveOffset` is now the shared adapter-layer offset normalizer:
  - omitted / `null` -> no offset;
  - `0` -> legal and still limit-bounded;
  - positive integer -> accepted;
  - negative, `NaN`, `Infinity`, or fractional -> fail closed before SQL emission.
- `assertExplicitOffsetOrdering` requires a non-empty `orderBy` for positive offsets.
- MSSQL's `(SELECT NULL)` fallback is removed.
- MySQL now always applies `resolveEffectiveLimit`.
- `data-source-adapter-boundary-conformance.test.ts` is parameterized across MSSQL, Postgres, and MySQL for A5, offset validation, and explicit-ordering behavior.

The wording is deliberately narrow and honest: a non-empty `orderBy` proves only that an explicit ordering was requested. It does **not** prove a deterministic total order. Stable caller pagination must include a stable unique tiebreaker, such as a primary key as the final sort key.

## Current-head verification

- Adapter conformance: **46/46**.
[...]

PR #4590, opened 2026-07-24
docs(integration): database & system integration line — design and verification MD (RATIFIED, doc-only)

zensgit:
**Doc-only.** The line's single design-and-verification MD (owner directive: consolidate; stop letting working memos drift). Supersedes the `/tmp` working memos and the retired NO-GO inventory SQL.

> ⚠️ **This body was stale and has been rewritten.** It previously read *"approving this PR unlocks only the B1a latent contract + harness slice"* — a scope the document **withdrew as self-contradictory** several revisions ago. A reviewer reading the old body would have approved a different scope than the one the doc defines. Corrected below; the doc header is authoritative.

## Gate semantics — **B1a AUTHORITY-SUBSTRATE gate** (doc header is authoritative)

The "latent contract + harness only" framing is **withdrawn**: §3.0's boundaries cannot be met inside a latent slice — a real config v2 changes a **live approved-config validation path**, and the identity read / canonical contract registry / server-bound source executor add **internal derivation and connection paths**.

**PERMITTED, bounded** — six items, enumerated verbatim in the doc's Gate block: additive `orderingKeySpec` / `actionProfileVersion` acceptance **and persistence** in the approved-config validator; a purpose-built system identity read; a first-party canonical object contract registry; a server-bound source executor (incl. the bounded core-backend statement seam); removal/privatisation of the legacy `probe()` entry point; and the previously-designed non-request-reachable B1a internals plus the counter/handshake **contract shapes** (no wiring).

**FORBIDDEN — each its own later gate:** any **new** request surface or other request-reachable behaviour **beyond** validation and persistence of the two new fields on the existing draft/save path; any **runtime consumer** of a GIP profile; the **scheduler**; arming / activation / **deployment** / rollout; telemetry or handshake **wiring**; the **B2 enforcement merge** (**#4591**); sealed-snapshot implementation; page-size ceiling changes; CDC / external write-back / D2 / W3 / G1 (frozen).

This gate is a **narrow, owner-approved widening of `GIP-D0` §9.2** (which unlocks only *profile schema / compliance harness / read-only qualification spike*) for exactly three items: config v2 additive validation/persistence, the internal authority substrate, and the restricted statement seam reachable from **fixture / first-party engine only**. **Approving §4 approves that widening.**

## Status

- **§4 is the single authoritative slice order and is RATIFIED** (owner, 2026-07-25, @ `a7c562d34`). Where any other section — including §1, §2, §3.x, §5, §6 and the Gate — implies different sequencing or scope, §4 wins.
- **B1a redo is NOT started.** #4596 / #4597 remain HELD (#4597 is the **B1b** slice and is **not** retargeted to main); #4591 stays Draft and merges **last**; #4593 / #4594 / #4598 Draft.
- **Four owner decisions ruled 2026-07-25** and recorded in the new **§4.0 roster** — (α) identity vs credential handling, (β) closed connector-kind registry, (γ) canonical object contract registry, (δ) B-6 v1 = (c). Steps 1.2 / 1.3 / 1.4 are consequently no longer NOT STARTABLE.

## What's inside

- **§1 RECORD** — landed facts with head-scoped SHAs: `bridge.bounded_read.v2` = `7bf2bd7a1` (#4573; #4565 closed unmerged), #4583 = `68bcd9a670` (MySQL A5 + frozen single-source registry), #4591 = the B2 re-cut that merges last, and the retired inventory SQL with its four schema-fact errors. §1 records; it schedules nothing.
- **§2** — M0–M2 acceleration route, with **M0 split into M0-A (authorized: build + verify the complete RC-A package, revise the #4437 pointer, prepare the bounded config)** and **M0-B (ops-gated: controlled deploy → flag-OFF health → preflight → flag-ON window → 11-item PASS)**.
- **§3 B1 design** — all owner findings absorbed and ⟲-marked, including **§3.0's six qualification-authenticity boundaries** (B-1…B-6) that put B1a into redo.
- **§4** — the authoritative order: B1a redo (six substeps) → **real MySQL / SQL Server capability spike, first-party instances only** → B1b certification → B1c → B1-observability → customer migration → **B2 = #4591 LAST**.
- **§5** — implementer landmines, session-verified against code.
- **§6** — binding fences.

## Upstream contracts this ledger defers to

**`GIP-D0` design lock** (`docs/development/gip-d0-general-integration-platform-design-lock-20260723.md`, landed via #4553 `a53a199b1`) is authoritative for `systemContentKey` / `roleBindingFingerprint` / `secretVersionId` and the system-vs-config identity split. **Where this ledger and the lock disagree, the lock wins and the ledger is what gets fixed** — with the single declared exception of the approved §9.2 widening above.

## Verification

Doc-only: `git diff --name-status origin/main...HEAD` is a single added `docs/development/…` file — zero non-`docs/` paths, zero deletions.


zensgit:
## Head mapping — all five findings are already absorbed (review bound to the first commit)

The HOLD verdict cites `9462c9c98`, which is **commit 1 of 4**. Absorption landed in the three commits after it; current head is **`eb0846685`**:

| review finding | absorbed in | where in the doc |
|---|---|---|
| **P1** RC-A must be the COMPLETE on-prem package, not the C-stage sidecar (sidecar has no `bridge-agent-readonly-adapter.cjs`; #4573 is server-side) | `eb0846685` ⟲R7 | §2 M0 — full-package build at owner-chosen SHA, regenerated manifest/SHA256/provenance/loopback, **`serviceRuntimeSha` + `clientHelperSha` recorded separately** (helpers blob-identical ⇒ helper hash may carry over; one `exactSha` no longer conflates client and server), then controlled deploy → flag-OFF health → bounded preflight |
| **P1** resolver: full tuple binding | `029366ce2` ⟲R3 | §3.1 — `{actionProfileVersion, systemContentKey, configContentKey, objectKey, canonicalObjectVersion, orderingKeySpec}` from the same tenant's approved binding+config+system |
| **P1** …`getForRuntime()` re-verify approved + tenant + scope | `029366ce2` + `eb0846685` | §3.1 bullet 1 |
| **P1** …server recomputes `configContentKey` | `029366ce2` | §3.1 bullet 2 (never trusts a stored column) |
| **P1** …strict JSON owned clone + recursive freeze | `029366ce2` ⟲R2 | §3.1 — canonical-JSON owned clone, recursively frozen, Proxy/accessor/sparse/symbol rejected (`deepCloneFrozenCanonical`) |
| **P1** …probe + verify both from resolver, no caller override | `029366ce2` | §3.1 bullets 3–4 |
| **P1** …async-window mutation negative control | `029366ce2` ⟲R2 | §3.1 — mandated harness NC: mutate the ORIGINAL array inside the async query callback, prove the probe uses the parse-time copy |
| **P1** M1 ↔ B2 order conflict | `029366ce2` ⟲R4 | §2-M1 — the "inventory=0 ⇒ merge" fast path is **removed**; §4 order is authoritative; M1 produces evidence only |
| **P2** combinations scoped to PAGED_READ, CHANGE_FEED intact | `029366ce2` ⟲R5 | §3.3 — named `PAGED_READ_LEGAL_COMBINATIONS`, consulted only when `acquisitionMode === 'PAGED_READ'`; ratified `CHANGE_FEED + MONOTONIC_VERSION_PIN + DURABLE_TOKEN` explicitly untouched + mandated NC pair |
| **P2** `orderingKeySpec` closed schema | `029366ce2` ⟲R6 | §3.1 — canonical `fieldId`s only (no raw SQL/expressions/aliases), non-empty, no duplicates, `direction ∈ {ASC, DESC}`, all `fieldId`s resolve through the SAME approved config version's mapping; NULLability stays with the qualification probes |

**Three decisions recorded and executed** (`eb0846685`): package policy **(b)** with the strict authorization boundary (build + verify + revise pointer — **not** deploy, **not** flag-ON); **B1-observability NOT opened early** (M1 rests on DB counts + windowed access log + static call-graph, with the runtime residual stated explicitly); **B2 re-cut done** — **#4591** (offset fail-fast guard + typed closed 422 + MSSQL fallback deletion, observability contracts excluded), **#4580 CLOSED as superseded**.

Please re-review at **`eb0846685`**.

zensgit:
## Boundaries written back into the design — head now `6e3d27308`

Per the instruction to write the boundaries into the design **before** ratifying, §3.0 is new and carries all six findings as **design**, not as notes on a held branch. Head moved `d657d9632` → **`6e3d27308`**; this PR remains **HOLD**.

| | boundary now in the design |
|---|---|
| **B-1** | a qualification must prove its evidence came from the **bound system** — the server-bound source executor is a **precondition** of B1a, not a follow-up |
| **B-2** | system identity from a dedicated **lossless identity read** (endpoint + auth principal + scope), **without decrypting credentials** — not a hash of the whole config |
| **B-3** | `canonicalObjectVersion` from a **first-party canonical object contract registry** — not derived locally |
| **B-4** | a probe SQL **denylist is not a security boundary** — first-party builders by module-private identity, or a restricted structured query constructor |
| **B-5** | **"certified" requires a verified guarantee**, not an honest label — SQL Server must refuse under the engine default; MySQL's three conditions must be established empirically on the same connection |
| **B-6** | certified **source-column translation** for `orderingKeySpec`; and a real **config v2** carrying `orderingKeySpec`/`actionProfileVersion` so the save → approve → re-read → qualify loop is actually exercised |

Two things recorded against myself, because they are the reason this round needed a fourth review:

1. I graded B-1 an acceptable **"documented residual"**. It is not — it defeats the slice's stated goal, and **latency does not make an untrue qualification trustworthy**. The retraction is in the ledger and on #4596.
2. On **B-4** I re-ran the supplied `dblink` bypass and found **two more of the same class** — `'DR'||'OP TABLE x'`, and `chr(68)||chr(69)||…` which carries no literal keyword at all — while the un-concatenated control is correctly blocked. So B1b's "7 constructs → 0" measures resistance to **accidents, not construction**, and I should not have reported it as closing a class.

The ledger also **withdraws** my earlier status conclusion that the remaining critical path was owner decisions rather than engineering. It is not: qualification authenticity is real unbuilt implementation work and is now the largest remaining item on the line.

**Sequencing executed as instructed:** #4596 and #4597 marked HOLD with the retractions posted; **#4597 not retargeted** (base stays the B1a branch); #4593 / #4594 / #4598 converted to **Draft** (branches retained, expansion paused); #4591 stays Draft and merges last. Active PRs are now **#4590 (this one) and the redone #4596**.

The on-prem M0 track — complete RC-A package build and the bounded approved config — is independent of B1a/B1b and is not blocked by any of this.

zensgit:
## Pushed once, as instructed — bind your short review to **`f939aa8a7`**

Six commits, `6e3d27308` → **`f939aa8a7`**. Still **HOLD**; doc-only; **B1a redo NOT started**.

Your 5 round-4 findings and your 3 precheck findings are all absorbed. Between them I ran two four-lens adversarial passes; they raised 44 findings, of which 3 survived independent refutation and I absorbed a further set of attribution/precision defects that the refuters' `survives:false` default drops — that default systematically discards exactly the class you keep catching me on, so I did not let a clean verdict stand in for the check.

**Your two premise-breaking P1s.** The first is the one that mattered:

> "the ratified formula hashes `config` whole" — **FALSE.** That described **#4596's implementation**. The contract is **GIP-D0 §6**, landed via #4553 (`a53a199b1`).

Root cause recorded in the doc: **this ledger had never cited GIP-D0 at all.** There is now an "upstream contracts this ledger defers to" block, with the rule that where the two disagree **the lock wins and the ledger is what gets fixed**. B-2 is rewritten retraction-first around the ratified four-term formula; #4596's whole-`config` + `role` + raw-ids version is recorded as a **deviation**, and per your ruling `role` and the raw ids stay out — adding them would be an amendment, never a "completion". My previous "narrowing is a regression in sensitivity" analysis is **withdrawn with it**; it was built on the false premise. What survives is the obligation the narrow formula genuinely carries: a per-connector-kind certified declaration of where endpoint and principal live, fail-closed on an undeclared kind.

The gate is now a **permission + prohibition** pair, not a slogan; the SQL Server spike freezes your three outcomes.

**Three things the passes found that I would not have, and that change the plan:**

1. **My residual-1 acceptance predicate was VACUOUS.** It asked for "`probe()` absent from the module's exports" — and `probe` **is not a module export**; it is a method on the frozen object from `createBindingQualificationProber` (L360-L371), which is exactly how the residual test reaches it. The assertion **passes on the unfixed head**. Rebound to the exact key set of the frozen prober object.
2. **"Identity read that does not decrypt credentials" is UNSATISFIABLE** alongside the ratified formula: `authPrincipalKey` / `authTenantScopeKey` are not in `config` for any shipped connector kind — they are in the **encrypted credential envelope** (K3 WISE WebAPI L1555-L1557; HTTP adapter L131). Now **owner decision (α)** at step 1.2, NOT-STARTABLE force.
3. **B-6's option (b) — "scope the executor to SQL-shaped sources only" — is STRUCTURALLY EMPTY.** Every approved read-source config is HTTP-shaped by construction (`readMethod`/`readPath` validated unconditionally, four HTTP read modes, no SQL key in the allowlist), so it scopes to the empty set — which also means **B1b's certified SQL builders are unreachable from the approved plane**. Replaced by three real options.

Also: an **allowlist-only** config v2 would ship **green** — `normalizeReadSourceConfig` is a key-by-key projection, so the fields would be accepted, silently dropped, and two bodies differing only in `orderingKeySpec` would collide on `content_key`, so `configContentKey` would stop pinning ordering behaviour. Step 1.1 now carries a two-part acceptance predicate for it.

**Three open owner decisions now block work** — (α) credential/identity, (β) per-kind declaration + migration for the free-form `kind` strings already stored, and B-6's (a)/(b′)/(c). **Steps 1.2 and 1.4 are marked NOT STARTABLE** until they are taken.

One item needs your explicit eye because it exceeds a ratified scope: GIP-D0 is **RATIFIED NARROW** (§9.2/§10 unlock only profile schema / compliance harness / read-only qualification spike). The gate permits two items in none of the three — the **live approved-config validator** change and the **bounded core-backend seam**. Being owner-dated one day after the lock, I have written it as a deliberate **widening of §9.2 for those two items only**, called out rather than inferred from a date — **re-ratifying §4 approves that widening.**

#4596 / #4597 unchanged (HELD, #4597 not retargeted); #4591 Draft, merges last; #4593 / #4594 / #4598 Draft. The on-prem M0 track is unaffected — and §6 no longer silently drops **M0-B**.

zensgit:
## Head correction: bind the review to **`61b5e86db`**, not `f939aa8a7`

The second review pass finished **after** I pushed and ran against the pushed head. It found four defects — **two of them introduced by this round's own fixes** — so the head has moved once more. Same HOLD, same doc-only status, B1a redo still not started.

**The one that matters, because it is a repeat of the class you held me for:**

> §3.3 said the contracts module *"already ships"* `PAGED_READ_LEGAL_COMBINATION` and `PAGED_READ_REQUIRES_CONSISTENCY_PROOF`.

Verified false: **0 occurrences at `origin/main`, 5 at `774bdb5e6`** — the **held** #4596 head that this document's own §3.0 grades *"redo REQUIRED and NOT STARTED"*. That is a held-PR artefact quoted as a shipped contract, i.e. **the same mis-attribution as the GIP-D0 error**, committed in the very sentence warning about token drift. The contrast is instructive and is now written into the doc: §4 step 3 calls `PROBE_STRATEGY_UNBOUND` *"existing"*, and **that one genuinely is on main** (2 occurrences) — the same discipline applied correctly once and incorrectly once, in the same document.

The other three:

- **[P1]** The acceptance-predicate intro said residual 1 *"goes through the **exported** `probe()`"* — while the same bullet, seven lines later, retires that exact claim as vacuous because `probe` is **not** a module export but a method on the frozen prober object (L360-L371). The bullet handed a reader two mutually exclusive code facts about one symbol, with the false one as its premise. The conclusion (retire as inexpressible; do not add a detector) is unaffected.
- **[P2]** Gate bullet 2 withdraws the flat "does not decrypt credentials" rule and defers to decision **(α)**; bullet 4 then said the executor *"may derive and hold a handle"* **unqualified**. The harness/fixture clause bounds *where it executes*, not *what it decrypts to build the handle* — so the narrow use stayed blocked while the broader one read as permitted outright. Bullet 4 now defers the encrypted-envelope case to (α) symmetrically.
- **[P3]** The kind/role immutability check is `external-systems.cjs` **L253-L254**; `L255` is `id: existing.id`. Pointer was off by two.

**Independently re-verified against the worktree** in that same pass, for your transport check: GIP-D0 §6's formula and its `configContentKey` / `secretVersionId` / `actionProfileVersion` boundaries; §9.2/§10's three-item narrow unlock; B-4's bypass table reproducing exactly against the six patterns at `d0313feec`, both controls behaving as tabulated; B-5's two strategy tokens; `deriveSystemContentKey` L552-L581 with material at L571-L579; the D2/D3 comment ranges; the resolver header L102-L120; every `read-source-config.cjs` and `read-source-config-store.cjs` citation; `http-routes.cjs` L20/L21; and both residual tests at L752-L805 / L813-L847.

Nothing else changed — the three blocking owner decisions **(α)**, **(β)** and B-6's **(a)/(b′)/(c)** stand, steps 1.2 and 1.4 remain **NOT STARTABLE**, and the GIP-D0 §9.2 widening still needs your explicit nod as part of re-ratifying §4.

zensgit:
## Four rulings written in, rebased on main, fresh-green — review head is **`e629d169c`**

| | |
|---|---|
| head | **`e629d169c`** (rebased onto `origin/main` = `3fa1ae342`, `behind = 0`) |
| diff | **doc-only** — `git diff --name-status origin/main...HEAD` is a single `A docs/development/…`; **0** non-`docs/` files, 0 deletions |
| checks | **13/13 SUCCESS**, `mergeStateStatus: CLEAN` — `pr-validate`, `test (18.x)`, `test (20.x)`, `web-tests`, `coverage`, `contracts (strict / openapi / dashboard)`, `attendance-web-guard`, `DingTalk P4 ops regression gate`, `K3 WISE offline PoC`, `stock-prep PowerShell 5.1 acceptance`, `after-sales integration` |

**Your P1 was a roster defect, not a text defect** — the canonical contract registry decision was in step 1.3's body all along while every summary I wrote listed only three blockers, so it read as a note. New **§4.0 decision roster** carries all four with ruling + landing site, and fixes the cause: *any future gating decision is added to the roster first.* Stale `(a) or (b)` label at step 1.4 corrected to the authoritative `(a)/(b′)/(c)`.

**The four rulings (marker `⟲OD`, defined in the legend):**

- **(α)** bounded decryption **inside the `credential-store` boundary**, brief, connector-certified principal/scope only → **domain-separated HMAC / canonicalise immediately → discard plaintext**; secrets and principal plaintext **never** into evidence, logs or errors. Rotation with principal+scope unchanged ⇒ `systemContentKey` **unchanged**; principal **or** permission scope changed ⇒ **rebuild lineage, re-qualify**.
- **(β)** first-party **closed** connector-kind registry; aliases **explicitly mapped**; unknown kind ⇒ **`SYSTEM_IDENTITY_KIND_UNCERTIFIED`** fail-closed for GIP binding; **never** auto-extended from customer strings; **legacy paths keep working**.
- **(γ)** canonical contract registry **first-party only**; immutable `contractId` + `version`, **append-only**; **no** auto-synthesis from customer config; unregistered ⇒ values-free **`CANONICAL_OBJECT_CONTRACT_UNREGISTERED`**; **inventory + backfill before activation**.
- **(δ)** B-6 **v1 = (c)** — connector-owned, **named, certified HTTP probe actions only**; **SQL builders stay unreachable** as the accepted outcome; **no** SQL-shaped widening, **no** invented source-column artefact; SQL defers to **(b′) + B1c** behind its own gate.

**GIP-D0 §9.2 widening** recorded at exactly your approved scope — config v2 additive validation/persistence, internal authority substrate, restricted statement seam reachable from fixture / first-party engine only; **request surface, scheduler, runtime consumer, arming, deployment, rollout still forbidden**.

**(δ) dissolved the step-1.4 blocker**: v1 never reaches a SQL source, so the non-existent per-system source-column mapping is not a precondition. **Steps 1.2 / 1.3 / 1.4 are no longer NOT STARTABLE** — re-ratification of §4 is the only thing still gating item 1.

Two places I added negative controls rather than trusting prose, because they are the clauses most likely to be broken while looking implemented: **(β)'s "legacy paths keep working"** — an uncertified kind must still work on its pre-GIP path *while* being refused a GIP binding — and **(α)'s rotation semantics**, which are an observable contract and want a test in **both** directions.

B1a redo still not started. #4596 / #4597 HELD (#4597 not retargeted); #4591 Draft, merges last; #4593 / #4594 / #4598 Draft. M0 parallel prep unaffected.

zensgit:
## All three fixed — exact-head short review can bind to **`a7c562d34`**

| | |
|---|---|
| head | **`a7c562d34`** (rebased onto `origin/main`, `behind = 0`) |
| diff | **doc-only** — 0 non-`docs/` files, 0 deletions, `git diff --check` clean |
| checks | **13/13 SUCCESS**, `mergeStateStatus: CLEAN` |

**[P1] Connection secret vs identity material — separated.** You were right that the text was unimplementable, not merely imprecise: "discard the secret" and "complete an authenticated connection" cannot both hold under one rule. Now two materials, two rules:

| material | rule |
|---|---|
| **identity material** (principal / scope, for `systemContentKey`) | decrypt inside the `credential-store` boundary → **domain-separated HMAC / canonicalise immediately → discard plaintext**. HMAC applies **here and only here** |
| **authentication secret** (for actually connecting) | **consumed inside the boundary by a `credential-store` / connector-owned factory**, which returns an **opaque handle / execution closure**. The secret is **never returned to, held by, or reachable from** the executor. **HMAC is not applicable** — a hashed secret cannot authenticate |

Shared constraint unchanged: neither may reach evidence, logs or errors. Propagated to **all four** places (α) is stated — Gate bullet 4, §4 step 1.2, §3.0 B-2, the §4.0 roster row — plus a **§5 landmine** (*"do not apply one credential rule to two materials"*), because the old text was actively inviting that mistake rather than merely permitting it.

**[P1] PR body — rewritten.** It was still advertising the withdrawn *"unlocks only the B1a latent contract + harness slice"*, so a reviewer would have been approving a scope the document does not define. The new body leads with an explicit **"this body was stale and has been rewritten"** note quoting the old wording, rather than quietly replacing it, and now carries: the AUTHORITY-SUBSTRATE gate's six permitted items and its forbidden list, §4's PENDING RE-RATIFICATION status and precedence rule, the §4.0 four-decision roster, M0-A/M0-B, the held-PR dispositions, and the GIP-D0 deference rule with the approved §9.2 widening.

**[P2] §9.2 widening — request-surface exception tightened.** This was the *same* self-contradiction I had already fixed in the Gate's boundary test, repeated one section away: item 1 unlocks config v2 validation/persistence while the next paragraph banned "the request surface" unconditionally. Now worded identically to the boundary test — **no NEW request surface or other request-reachable behaviour, with exactly one exception: validation and persistence of the new fields on the EXISTING draft/save path.** Scheduler, runtime consumer, arming, deployment and rollout remain forbidden **without exception**.

Nothing else moved: §4 still **PENDING RE-RATIFICATION**, **B1a redo not started**, #4596 / #4597 HELD (#4597 not retargeted), #4591 Draft merging last, #4593 / #4594 / #4598 Draft. M0 parallel prep unaffected.

PR #4591, opened 2026-07-24
fix(data-source): B2 — OFFSET-ordering fail-fast guard + typed closed 422 + MSSQL fallback deletion (DRAFT, merges LAST)

zensgit:
**Pure B2 enforcement re-cut from current main (owner ruling). Supersedes #4580 — that branch still carried the A-half already landed via #4583 and went DIRTY. DRAFT by design: per the §4 slice order this merges LAST, after adapter-chokepoint telemetry (B1-observability, its own gate), the coverage-mapped caller inventory, and customer migration. Design + evidence ledger: PR #4590.**

## Scope — exactly the three owner-listed items

1. **OFFSET-ordering fail-fast guard** (BaseAdapter chokepoint, wired in MSSQL/Postgres/MySQL): `offset > 0` without `orderBy` ⇒ fail-closed. Unordered offset windows silently duplicate AND skip rows (demonstrated by test). Limit-only first page stays legal; the first-page residual is **pinned as a KNOWN LIMIT** (closes in the caller's ordering contract, B1a+). No auto-order-by-primary-key — that would hide the missing contract.
2. **Typed closed 422** (review merge-condition): `DataSourceOffsetOrderingError` / `DATA_SOURCE_OFFSET_ORDERING_REQUIRED`; `/select` maps it to **422** instead of the old catch-all `SELECT_ERROR` 500. Mapping is closed, not widened — generic errors still 500 (pinned).
3. **MSSQL `ORDER BY (SELECT NULL)` deletion** — the "looks ordered, isn't" construct; pinned never to return.

**NOT here (deliberate):** observability contracts/counters/handshake — B1-observability slice, separate runtime gate.

## Verification

- Conformance suite with **registry-derived roster** (auto-covers a newly registered SQL adapter): per-adapter typed fail-closed + positive control + limit-only legal; MSSQL artifact pin; corruption demonstration (one duplicate + one skipped row); KNOWN LIMIT pin; route 422 / closed-500 / 200 controls. **16/16.**
- **Mutation-verified both ways**: neuter the guard → exactly the fail-closed cases red; remove the route mapping → exactly the 422 case red.
- Full core-backend unit suite **404 files / 5632 tests green**; `tsc --noEmit` clean.

## Known reachable consequence (unchanged, why this merges LAST)

`data-source:sql-readonly`'s non-watermark offset path fails closed at page 2 instead of corrupting silently. Its ordering contract arrives with the B1a resolver (`orderingKeySpec` from the approved config version); migration precedes enforcement.

### DIFF
--- a/packages/core-backend/src/data-adapters/BaseAdapter.ts
+++ b/packages/core-backend/src/data-adapters/BaseAdapter.ts
@@ -83,6 +83,22 @@ export interface QueryOptions {
 export const DATA_SOURCE_MAX_ROWS = 10000
 export const DATA_SOURCE_DEFAULT_LIMIT = 1000
 
+/** Closed error code for the offset-ordering read contract (mapped to HTTP 422 at route surfaces). */
+export const DATA_SOURCE_OFFSET_ORDERING_REQUIRED_CODE = 'DATA_SOURCE_OFFSET_ORDERING_REQUIRED'
+
+/**
+ * Typed, closed contract error for an offset read with no deterministic order. A CALLER/CONFIG
+ * error, not a server fault: HTTP surfaces map it to 422 with the closed code above — without the
+ * type, the /select route's catch-all reported this as `SELECT_ERROR` 500.
+ */
+export class DataSourceOffsetOrderingError extends Error {
+  readonly code = DATA_SOURCE_OFFSET_ORDERING_REQUIRED_CODE
+  constructor(message: string) {
+    super(message)
+    this.name = 'DataSourceOffsetOrderingError'
+  }
+}
+
 export interface QueryResult<T = Record<string, DbValue>> {
   data: T[]
   metadata?: {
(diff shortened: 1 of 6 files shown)


zensgit:
## Last-merge gate sync after #4594

The schema-probe-first exposure inventory is now on `main` through #4594 at `dd693d8ca65bee182e1fdd14212cead45fe7c947`. That merge supplies the read-only discovery tool; it is **not** evidence that any customer deployment has been inventoried or migrated.

This PR therefore remains **Draft / merges last**. Do not refresh it merely to obtain green checks and do not merge it until all of the existing prerequisites are evidenced for the affected deployment(s):

1. authorized per-deployment inventory, including non-HTTP/direct adapter residuals;
2. ordering-contract/config migration for every affected caller;
3. the separately authorized observability/runtime gate, if that remains part of the final rollout decision;
4. fresh sync to then-current `main`, exact-head review, and required checks.

Until then, the current behavior remains preferable to prematurely enabling this fail-fast enforcement: the known unsafe OFFSET path is documented, but customer callers have not yet been proven ready for the new 422/fail-closed behavior.

PR #4593, opened 2026-07-24
docs(gip): B1c — page-sequence execution design (PROPOSED, design-first, no runtime)

zensgit:
**Doc-only, PROPOSED / design-first.** Authorizes no runtime, arming, wiring, or implementation — B1c implementation is its own owner gate. Slice order lives in the line ledger (#4590), not here.

## Why this slice exists

A probe strategy certifies **one statement**. `DataSourceManager.select()` is pool-per-call and transaction objects cannot be passed into `select()`/`query()`, so **nothing spans two pages today** — an ordering key therefore proves total order only, **not snapshot stability across pages**. `PAGED_READ` certification needs BOTH.

## What it specifies

- **PageSequenceExecutionContext** seam: B1a-resolver entry, WeakSet-trusted per-profile strategy, one context per run, closed `OPEN→READING→COMPLETED|ABORTED` lifecycle, read-only + A5 bounds composed, values-free evidence.
- Both `PAGED_READ_LEGAL_COMBINATIONS` rows, and explicitly **what is NOT a consistency context** (ordering key, `MONOTONIC_VERSION_PIN`, retry heuristics, new-snapshot-per-page).
- **Per-dialect certification** for PostgreSQL / SQL Server / MySQL — with probe-and-fail-closed duties on *both* engines whose posture is configurable (SQL Server's `ALLOW_SNAPSHOT_ISOLATION`, MySQL's isolation level, since `WITH CONSISTENT SNAPSHOT` is silently inert under READ COMMITTED).
- Resource honesty: pinned-connection cost, mandatory bounds, **lost context ⇒ ABORT, never a silent re-snapshot**.
- Keyset vs OFFSET, and the `SEALED_EXPORT` convergence — decided **per capability spike, never per database brand**.
- **12 open points marked "to be confirmed by spike"** rather than asserted: no live database was reached, so every dialect-guarantee statement is documentation-level by construction.

## The headline correction it records

[...]

PR #4596, opened 2026-07-24
feat(gip): B1a — latent contract + harness (PAGED_READ combinations, orderingKeySpec, approved-binding resolver)

zensgit:
**DRAFT.** The owner-ratified **B1a** slice — fully **LATENT** (zero production consumers, proven by tree-wide grep; no route, no flag, no arming). Ratifying the line ledger (**#4590**) is what unlocks this; this PR is the code. Design + evidence ledger: #4590. B1c design: #4593.

## What it implements

- **R5 `PAGED_READ_LEGAL_COMBINATIONS`** — a frozen table consulted **only** when `acquisitionMode === 'PAGED_READ'`, as rule 5 inside the existing `assertCertificateCrossDimensionLegal`. Two decisions worth review: the table is an **array of frozen rows, not a keyed object** (a `table[proof]` lookup would walk the prototype chain for names like `constructor`), and **every declared proof must anchor the declared lifetime**, not merely one — because `validateConsistencyEvidence` only checks `used ⊆ supported`, so an unmapped proof riding along could later be claimed *alone* as a run's proof class. Out-of-table ⇒ refused at certification time, **never silently downgraded to `BOUNDED_READ`**. The ratified `CHANGE_FEED × MONOTONIC_VERSION_PIN × DURABLE_TOKEN` is untouched.
- **R6 `orderingKeySpec` closed schema** — canonical `fieldId`s only, non-empty, no duplicates, `direction ∈ {ASC, DESC}`, all ids resolvable through the same approved config version's mapping. NULLability deliberately stays with the qualification probe.
- **R3+R2 approved-binding resolver** — derives the **complete six-field tuple** from one immutable approved version + its tenant's system record; re-verifies approval/tenant/scope via `getForRuntime`; **recomputes** `configContentKey` (the stored column is never in the tuple's provenance path); returns an **owned canonical-JSON clone, recursively frozen**, trusted by module-private **WeakSet identity**. Qualification binding is **purely additive**, so ratified entry points are untouched.

## Blocking P1 found in review and fixed

`deriveSystemContentKey` originally hashed the system config as returned by `getExternalSystem` — i.e. `sanitizeIntegrationPayload(row.config)`, a **lossy projection**. A reviewer proved end-to-end that a qualification minted against a production system **still verified `verified: true` after that system was repointed at an attacker endpoint** through the production upsert path — five collision classes, **three key-name-independent** (truncation), so no allowlist could close them.

Fix: read identity from a **lossless** source (`getExternalSystemForAdapter`, *required by the factory* so the lossy path is unwireable), narrow the record so decrypted credentials never reach assembly, and add a **fail-closed witness** that refuses any projected record before hashing. Independent re-verification built the attack from scratch — memory DB JSON-round-tripping so the prototype premise is *executed*, repoint driven through production upsert with the stored row re-read and asserted changed — and all 5 classes now fail closed with `QUALIFICATION_DIGEST_MISMATCH`, with positive controls both ways.

## Regression found by re-verification and fixed here

Reading identity losslessly put unbounded-depth JSONB onto the **recursive** canonical codec: depth 5000 threw `RangeError`, **escaping the frozen vocabulary** (still fail-closed, but unclassified — a wiring gate would surface a 500 instead of a closed refusal; the pre-fix lossy read hid this by truncating at depth 6). Depth is now a **rule** (`SYSTEM_IDENTITY_CONFIG_TOO_DEEP`, bound 64) enforced before hashing, plus a `RangeError`→closed-reason conversion as defense in depth. Mutation-verified both ways: removing the rule reds the refusals; tightening the bound to 1 reds the depth-60 positive control.

## Verification

16 agents across five workflows — **Opus 5** for the security core and every adversarial lens, **Sonnet 5** for wiring, **Fable 5** elsewhere.

- **288-cell grid** (mode × consistency subset × lifetime × completeness) on HEAD *and* the pre-R5 base: 54 cells changed verdict, **zero non-`PAGED_READ` cells moved**; no silent-downgrade path.
- **15 independent mutations** against test vacuity; the owner-mandated **async-window negative control proven load-bearing in both directions**.
- **19 trust + 5 scope forgery attacks**: 18/19 and 5/5 fail closed (hand-built clone, Proxy, duck-typed brand, prototype impostor, all three `Object.keys` blind spots, accessor identity-swap, prototype pollution, sparse arrays, scope forgery).
- Full `plugin-integration-core` chain green (103 links, 79 OK) — **run directly, not inherited from an agent report**.

## Known residual (documented, not hidden)

`probeFromResolution` still accepts a caller-supplied `query`, so a qualification proves *"these six approved inputs + this evidence"*, **not** *"this evidence was observed on the bound system"*. Deriving the source handle from the resolution's system record is owned by the wiring gate; it is **pinned by a test** rather than claimed closed.


zensgit:
## ⛔ HOLD — retracting an over-claim in this PR's own description

Owner review at exact head `774bdb5e6` found a **P1 that I mis-graded as an acceptable residual**. I have re-verified it and the owner is right.

### The retraction

This PR's description says the forgery is *"inexpressible for callers restricted to the FromResolution pair"*. **That is false and is withdrawn.** The suite in this very PR proves it:

- **`__tests__/gip-approved-binding-resolver.test.cjs` L740-L772** mints a qualification with `query: async () => ({ rows: [{ duplicate_groups_sampled: 0, null_key_rows: 0 }] })` — a fabricating function that **never touches any external system** — and asserts `verified === true`.
- **L807-L840** shows one such caller-side answer satisfying **two resolutions bound to different systems**.

I documented this as a "known residual, pinned by a test rather than claimed closed". That framing was wrong on the point that matters: **B1a's stated goal is that forgery is inexpressible by construction, and it is not.** Latency does not make an untrue qualification trustworthy — a qualification that verifies against evidence never observed on the bound system is a false qualification whether or not anything consumes it yet.

### Also accepted (owner P1)

**`systemContentKey` and `canonicalObjectVersion` are invented, not derived from the ratified model.** `systemContentKey` is a hash over the whole system config (`lib/gip-approved-binding-resolver.cjs` L153-L169); `canonicalObjectVersion` is then derived from it plus `objectKey` + `fieldMap` — and **the code itself admits it cannot detect source-side schema drift** (L232-L252: *"CANNOT detect source-side schema drift"*). A "canonical object version" that cannot witness the object's schema is not one. The correct boundary is the owner's: a dedicated **lossless identity read** yielding endpoint / auth principal / scope identity, and `canonicalObjectVersion` from a **first-party canonical object contract registry**. The current resolver also decrypts credentials through the adapter API and then discards them — an unnecessarily wide permission surface for an identity read.

Accepted too: **P2** — `orderingKeySpec` `fieldId`s are canonical TARGET fields (the resolver says so at L102-L120) yet are handed straight to the SQL builder with no certified source-column translation; and **P2** — the production approved-config allowlist does not accept `orderingKeySpec`/`actionProfileVersion`, so the hermetic suite never proved the real save → approve → re-read → qualify loop.

### Status

**HOLD. Not for merge.** Per the owner's sequencing this PR is to be **redone**, in this order: real config v2 → system identity read with no credential decryption → first-party canonical contract version → server-bound source executor with field translation → remove or privatise the legacy `probe()` entry point.

The branch stays for reference; it is not a candidate in its current shape.

zensgit:
## SUPERSEDED — closing; branch retained as evidence

This is the **first B1a attempt**, which owner review graded **REDO REQUIRED**. It is closed rather than revived: the redo changes the substrate itself, so continuing on this head would carry the defects forward under a green diff.

What this head measured, and why the redo cannot be an increment on it — all four are recorded in the now-**RATIFIED** §3.0 of `docs/development/database-system-integration-line-design-and-verification-20260724.md` (merged in #4590, `402f04982`):

- **B-1** — the slice's own suite proves forgery is *expressible*: `ratifiedPathRemainsAnOpenConstruction` (L752-L805) mints over a foreign field set and still verifies; `callerSuppliedQueryRemainsAnOpenConstruction` (L813-L847) has one caller-side answer, touching no system, satisfy **two resolutions bound to different systems**.
- **B-2** — `deriveSystemContentKey` hashes the whole `config` plus `role` and the raw ids. That is a **deviation** from the ratified **GIP-D0 §6** formula (`kind + endpoint identity + authPrincipalKey + authTenantScopeKey`), not an instance of it.
- **B-3** — `canonicalObjectVersion` is derived locally from `systemContentKey + objectKey + fieldMap`, a pure function of inputs already in the same tuple.
- **B-6** — `orderingKeySpec` was never exercised through a real save → approve → re-read → qualify loop, because the approved-config allowlist does not accept it.

**The branch is kept as evidence** — the residual tests above are the acceptance predicate the redo must invert, so they need to stay readable.

**The redo opens as a NEW PR from post-merge `main`, in three batches** (§4 item 1): **B1a-1** = 1.1 config v2; **B1a-2** = 1.2 identity/kind registry + 1.3 canonical registry; **B1a-3** = 1.4 executor + 1.5 legacy `probe()` retirement + 1.6 contract shapes. It is deliberately **not** described as "the redone #4596".

PR #4597, opened 2026-07-24
feat(gip): B1b — certified MySQL + SQL Server probe strategies + dialect-aware read-only guard

zensgit:
**DRAFT, stacked on #4596 (B1a).** Base is the B1a branch — retarget to `main` once B1a lands. LATENT: no route, no flag, no arming, no runtime consumer outside tests.

## Two certified dialect strategies

Registered per `actionProfileVersion` through the existing unforgeable registry; an unregistered dialect still fails closed with `PROBE_STRATEGY_UNBOUND`.

| | MySQL | SQL Server |
|---|---|---|
| identifiers | backtick, doubled | bracket, `]` doubled |
| limiting | `LIMIT 1` | `TOP (1)` (T-SQL has no `LIMIT`) |
| snapshot claim | `…conditional_on_innodb_autocommit_and_isolation_read_committed_or_stricter` | `no_single_statement_snapshot_under_default_read_committed` |

**Neither token copies PostgreSQL's `single_statement_mvcc`.** MySQL's guarantee is conditional and the token names all three conditions (InnoDB, autocommit, isolation ≥ READ COMMITTED); SQL Server's is **absent** under the engine default and the token says so. Both are digest-bearing — permanent, checkable markers a later PAGED_READ certification must read, not disclaimers — and both are now pinned by **exact `assert.equal`**, because an overclaiming token previously survived every suite that could observe it.

## Dialect-aware read-only guard

The guard was PostgreSQL-flavoured. Measured: **seven non-PG constructs passed** (`WAITFOR`, `EXEC xp_cmdshell`, `OPENROWSET`, `GET_LOCK`, `BENCHMARK`, `SLEEP`, `LOAD_FILE`) while the `INTO OUTFILE` control was correctly blocked. It now ORs **six named patterns** — the two ratified ones extracted **byte-identical** and pinned character-for-character, plus four new — with a `guardPatternExclusivity` test asserting each discriminating probe is caught by **exactly one** door, so neutering a new pattern cannot be masked by a neighbour.

**The sharpest gap, which the first pass missed entirely:** SQL Server's **lock-taking class had ZERO coverage**. T-SQL has no `FOR UPDATE`/`FOR SHARE`, so the ratified clause token matches *nothing a T-SQL client can write* — locks are taken with table hints. `WITH (UPDLOCK)`, `(NOLOCK)`, `WITH (XLOCK, HOLDLOCK)`, `TABLOCKX` all passed. And `WITH (NOLOCK)` is precisely the hint that would make this slice's own `sqlserver` snapshot token describe a read that never happened.

Also closed: `\b`-terminated tokens couldn't match longer identifiers sharing a prefix (`pg_sleep_for`, `pg_ls_dir`, `lo_get` — plus the **pre-existing** `pg_advisory_lock_shared` family, fixed additively); MySQL's pre-8.0 `LOCK IN SHARE MODE`; the SQL Server `fn_` routine family.

## Claims narrowed to what is measured

The false *"state-taking locks are covered for postgres/mysql/sqlserver"* sentence is **retracted in place, retraction-first** so a grep stopping at line one gets the right answer. Also corrected: the coverage table no longer claims completeness for PG remote execution (`dblink`/`query_to_xml` still pass — scoped out, not covered) or MySQL unbounded work (`MASTER_POS_WAIT` family still passes); phrase tokens are stated as closed against whitespace/case/newline but **not** against intra-phrase comments; and `xp_`/`sp_` is labelled as introduced in this slice rather than ratified.

## Verification

8 implementer mutations, **44 reviewer mutations**, then a further round closing every finding. I personally re-ran all ten reviewer-found constructs against the hardened guard — **all blocked, legitimate probe still passing, zero remaining** — and added the two discriminators the final re-verification asked for, mutation-proving both: dropping `XLOCK` alone (`HOLDLOCK` had been covering for it, so a real hole red nothing) and neutering the structural `^SELECT` + semicolon rules together (`SELECT 1; SELECT 2` trips no token door — a pre-existing gap in the ratified suite). **Full chain rc=0, 80 suites.**

## Not verified

No real MySQL or SQL Server is reachable, so **every dialect-syntax and isolation-semantics judgment here is reasoned**, marked "to be confirmed by spike" in the code — including the InnoDB read-view conditions the MySQL token names and the T-SQL clause-order shapes.


zensgit:
## ⛔ HOLD — the guard hardening in this PR is not a security boundary, and I over-claimed it

Owner review at exact head `d0313feec`. Two P1s, both verified by me before writing this.

### P1 — a SQL-text token denylist cannot be a security boundary

`createProbeStrategyRegistry()` accepts an **arbitrary** `buildTotalOrderProbeSql` function, so the guard's only real input is whatever a registered builder emits. The owner demonstrated one bypass; I ran it and **found two more of the same class**:

```
⚠️ PASSES   SELECT * FROM dblink('conn', 'DE' || 'LETE FROM x') AS t(x int)     ← owner's
⚠️ PASSES   SELECT * FROM dblink('c', 'DR'||'OP TABLE x') AS t(x int)
⚠️ PASSES   SELECT * FROM dblink('c', chr(68)||chr(69)||'LETE FROM x') AS t(x int)
✅ BLOCKED  SELECT * FROM dblink('c', 'DELETE FROM x') AS t(x int)               ← control
```

String concatenation defeats the token list completely, and `chr()` defeats it without any literal keyword at all. **The 7→0 result I reported for this slice measures resistance to *accidents*, not to *construction*** — I should not have presented it as closing a class. The correct boundary is the owner's: only **first-party builders registered by module-private identity**, or a **restricted structured query constructor** — never SQL-text inspection.

### P1 — "certified" is being applied to strategies with no verified guarantee

- **SQL Server**: the token literally says `no_single_statement_snapshot_under_default_read_committed`, and the strategy is still accepted by the registry and still mints a candidate (verified). I praised this token as "honest". Honesty in the label does not help when the certification **succeeds anyway** — that is precisely a "looks certified, isn't". Under the engine default it must **refuse certification**.
- **MySQL**: the token names three conditions (InnoDB, autocommit, isolation ≥ READ COMMITTED) and the code **checks none of them** (L360-L394). The conditions must be established **empirically on the same connection**, not asserted in a string.

### Status

**HOLD. Not for merge, and NOT to be retargeted to `main`** (its base stays the B1a branch). Per the owner, B1b's certification gate does not open until a **real MySQL / SQL Server capability spike** passes. Branch retained for reference only.

zensgit:
## SUPERSEDED — closing; branch retained as evidence

This is the **first B1b attempt**, stacked on #4596 — which is itself closed as REDO REQUIRED, so this head's base is gone. It is superseded by **§4 steps 2-3** of the now-RATIFIED order (merged in #4590, `402f04982`): a **real MySQL / SQL Server capability spike against first-party engine instances** must pass **before** any strategy is registered.

Two findings against this head, both verified and both recorded in §3.0:

- **B-5** — it registers a SQL Server strategy whose own token reads `no_single_statement_snapshot_under_default_read_committed`, i.e. it **names the absence of the guarantee and mints a candidate anyway**; and the MySQL token names three conditions (InnoDB, autocommit, isolation ≥ READ COMMITTED) that **nothing checks** — there is no `@@autocommit` / isolation / engine probe anywhere in the module. Under §4 step 3 an unobtainable guarantee must **refuse certification**, not mint a candidate.
- **B-4** — the read-only SQL denylist is **not a security boundary**. Verified against `__internals.assertReadOnlySql` at this head: `'DE' || 'LETE FROM x'`, `'DR'||'OP TABLE x'` and `chr(68)||chr(69)||'LETE FROM x'` all pass, while the un-concatenated control is blocked and `SELECT 1` passes — so the "7 dangerous constructs → 0" result measures resistance to **accidents, not construction**. The boundary must be first-party builders admitted by **module-private identity**; the denylist survives only as defence-in-depth.

Also corrected in the ledger: an earlier revision described this PR as part of "the first B1a attempt". It is the **B1b** slice. It was never retargeted to `main`, and is not being retargeted now.

**Branch retained as evidence** — the dialect builders and the guard-pattern battery are the starting material for the post-spike B1b certification.

PR #4594, opened 2026-07-24
feat(ops): schema-probing, values-free data-source exposure inventory (replaces the NO-GO ad-hoc SQL)

zensgit:
**Read-only ops tooling. No runtime consumer, no customer DB run, no migration decision.**

This PR replaces the prior ad-hoc inventory SQL with a schema-probing, values-free exposure inventory for the `data-source:sql-readonly` path. The earlier body described the old 89-test head; that transport evidence is superseded by the exact-head evidence below.

## What it does

- Probes the required `public` tables and columns before constructing any count plan.
- Runs only a fixed, module-load-validated `SELECT` allowlist.
- Qualifies every count query to `public`, so `"$user"` or another `search_path` schema cannot shadow the probed objects.
- Uses `bigint` aggregates and rejects malformed, negative, unsafe, zero-row, or multi-row scalar results.
- Treats unknown external-system status values, missing schema, unreadable scan roots/files, and anchor drift as `INCONCLUSIVE`.
- Enumerates the in-repo `/select`, adapter-select, and `copyData` surfaces with fail-closed anchors.
- Emits only fixed vocabulary, counts, and coverage metadata; raw row values and driver errors are not returned.
- Leaves access-log inspection manual and explicitly records undeployed telemetry and other residual blind spots.

## Exact-head evidence

Review head: `d833a1e283d2144cc19e4d558557d58ff39662f7`

- `node --test scripts/ops/data-source-exposure-inventory.test.mjs`: **105/105 PASS**
- `node --test scripts/ops/gip-authority-substrate-inventory.test.mjs`: **144/144 PASS**
- Prettier check and `git diff --check`: PASS
- Eleven targeted mutations: **11 RED / 0 survived**, followed by restored green
- Real PostgreSQL 15.17 regression:
  - public-only exposure: `MIGRATION_REQUIRED`, five relevant counts = 1
  - same database after creating empty same-name tables in the default `"$user"` schema: still `MIGRATION_REQUIRED`, the same five counts = 1
- Independent Kimi K3 adversarial review of the hardened delta found **0 P1 / 0 P2 / 3 P3**. All three P3s were then fixed: scalar aggregate cardinality controls, per-event workflow anchor coverage, and static-enumeration residual reporting.

The attempted second Kimi delta process did not return a final verdict and is not represented as passing evidence.

## Boundaries

- This PR does not query a customer database.
- It does not deploy telemetry, alter adapters, enforce pagination, migrate callers, arm a profile, change flags, or authorize rollout.
[...]

PR #4603, opened 2026-07-25
feat(ops): GIP authority-substrate inventory — β connector-kind + γ objectKey probes

zensgit:
## Summary

Two read-only, values-free, schema-probing inventory probes feeding two of the four **§4.0**
owner decisions in
[`docs/development/database-system-integration-line-design-and-verification-20260724.md`](../blob/main/docs/development/database-system-integration-line-design-and-verification-20260724.md)
(the ledger; RATIFIED, merged as `402f04982`, §4 ratified `a7c562d34`):

- **(β)** distinct `integration_external_systems.kind` strings and their row counts — the raw
  material a human uses to author the explicit alias map and migration list for decision (β)'s
  first-party CLOSED connector-kind registry.
- **(γ)** distinct `integration_read_source_configs.object` values — the column the ledger's
  qualification-input tuple calls `objectKey` (no literal `objectKey` column exists on main; this
  mapping is asserted by a **line/word-boundary-anchored** schema-drift test — see "Fix pass"
  below for why the anchoring matters — not just commented) — referenced by read-source configs,
  split by status. Raw material for decision (γ)'s missing-reference and backfill lists.

This is a **new carrier**, per the **owner's task-brief ruling** — **not recorded in the ledger**;
the ledger's §4.0/§3.0 sections define the four owner decisions this work feeds but say nothing
about a carrier for this script. It reuses the schema-probe-first, values-free `QUERY_ALLOWLIST`
*pattern* of `scripts/ops/data-source-exposure-inventory.mjs` (#4594) but does **not** import
from, expand, or push to that file/branch/PR — #4594 is untouched by this change. **Note:**
`scripts/ops/data-source-exposure-inventory.mjs`, its test file, and its workflow yml live only on
branch `claude/data-source-exposure-inventory-20260724` (PR #4594) — they are **not in this tree
and not on main**; this PR's comments now say so explicitly rather than implying otherwise.

## What this does NOT do (gate boundary, stated per the task brief)

- Never auto-expands or auto-populates a first-party registry from what it observes, in either
  direction. `KNOWN_CERTIFIED_CONNECTOR_KINDS` and `KNOWN_CANONICAL_OBJECT_KEYS` are frozen,
  **empty**, module-local constants — this script has no code path that writes to them. Only a
  human, acting on the owner's decision (β)/(γ), may add entries in a follow-up change. Proven by
  test (frozen-array mutation throws; a full report run against fixture rows carrying novel values
  leaves both registries referentially and content-identical).
- No runtime consumer, no scheduler, no route, no arming/activation — pure offline DB read (two
  `information_schema` probes, four `SELECT ... GROUP BY` counts, all read-only-asserted at module
  load) plus a local JSON write.
- **Does not probe `integration_write_target_configs.object`** (migration 064) — a second
  objectKey reference class in the same database, deliberately out of scope for this inventory
  (decision (γ)'s scope per the ledger is the READ-side qualification-input tuple). Stated in both
  `RESIDUAL_NOTES` and the γ coverage `blindSpot` string, with the reason, and pinned by test.
- **Never connected to a customer deployment.** Tested hermetically (fake executor, `node --test`,
  no DB). A prior session in this PR reported testing against one **ephemeral local Postgres 15**
  that was stood up and discarded — see the downgraded claim under "Live-Postgres verification"
  below. Connecting this tool to a customer deployment is a separate, ops-gated read-only
  authorization this task does not carry, and this PR does not claim it.

## Output discipline (owner ruling, enforced structurally, not just documented)

- **Public/committed output (stdout, `--json`, the human summary): aggregate counts only.** The
  public report object is assembled exclusively from `.counts` integers, fixed status/verdict
  enum tokens, and label strings authored in the script — it is structurally incapable of holding
  an observed kind/objectKey string. Proven **two ways**: (1) a sentinel-injection test asserting a
  raw value injected through the fake executor is **absent** from `JSON.stringify(report)`
  (negative control) while **present** in the private artefact file (positive control); (2) an
  **exact key-set** test that walks `coverage`/`verdicts`/`residualNotes` and asserts the key set
  at every level is exactly the fixed, authored set — added in the fix pass below because (1) alone
  only catches a leak that lands as a string *value*, not one used *as an object key* (see "Fix
  pass").
- **Raw values go only into a private, gitignored local artefact** —
  `artifacts/gip-authority-inventory/…json` (new `.gitignore` entry, same PR), written with `0600`
  permissions, path only ever printed, never its content.
- **Fails closed**: if an unregistered/unmapped value exists but the private artefact cannot be
  written, `buildReport()` rejects (exit 1) rather than completing with an aggregate-only success
  that silently drops the one thing a human needs to act on it. If nothing was owed, a write
  failure is reported but does not fail the run.

## `--probe beta|gamma|both`

"Strictly separated purposes" (task brief) is enforced at the query layer, not just the printed
summary: under `--probe beta`, **no query naming `integration_read_source_configs` is ever
issued — schema probe included, in BOTH report mode and `--dry-run`** (`tableSpecsForProbe()`
scopes which tables even get an `information_schema` round-trip; `--dry-run` now threads `probe`
through the same way — see "Fix pass"), and no γ value can reach `coverage`, the public summary,
or the private artefact. Symmetric for `--probe gamma`. Verified hermetically: a fixture with zero
`count.read_source_config_*`/`count.external_system_*` entries for the unrequested probe — if that
probe ran anyway, the fake executor throws before any assertion — plus an equivalent zero-calls
check for `--dry-run` under each `--probe` value.

## Divergence check

One additional values-free integrity query: `integration_read_source_configs` rows where `object`
disagrees with `config->>'object'` (count only). If non-zero, the γ verdict is forced to
`INCONCLUSIVE` rather than letting a `CANONICAL_OBJECT_BACKFILL_NOT_REQUIRED_WITHIN_COVERAGE`
verdict coexist with a known disagreement between the column the backfill list is built from and
the config's own stored value. `IS DISTINCT FROM` (not `<>`) is deliberate: a row whose `config`
JSON has no `'object'` key at all is also counted — this follows from the documented SQL-standard
semantics of `IS DISTINCT FROM` treating `NULL` as distinct rather than `UNKNOWN`. A prior session
reported exercising this against a seeded row with a missing key on the (now-discarded) ephemeral
Postgres instance; see "Live-Postgres verification" below for why that specific claim is now
scoped rather than asserted as fact.

**Also fixed in this pass:** the `unexpected` status bucket (rows whose `status` is outside the
draft/approved/retired CHECK vocabulary — fail-closed by `splitRowsByStatus()`, not dropped) now
also forces `INCONCLUSIVE` when non-empty, for the same reason the divergence check does: a
`NOT_REQUIRED` verdict must not coexist with a known coverage gap. See "Fix pass".

## Fix pass (2026-07-25) — adversarial-review findings, each proved red-then-green

An adversarial review returned HOLD on the original version of this PR. All six findings were
fixed and each fix is proved with a **before/after mutation pair**: the production code (or a
migration file, for schema-drift) was temporarily mutated, `node --test` was re-run to show the
relevant new/existing test(s) going **RED**, then the mutation was reverted and the suite
confirmed **GREEN** again. No mutation was left in the committed diff.

1. **[P2] γ schema-drift guard didn't pin the `object` column name.** The old test used
   `assert.match(migration062, /object\s+TEXT NOT NULL/)` — a loose, non-anchored regex. Renaming
   the column to `canonical_object` still satisfies it, because `canonical_object` *contains* the
   substring `object` immediately followed by whitespace + `TEXT NOT NULL`. Fixed the same way
   057's `kind` guard already worked: a **line-anchored** `find()` over `migration062.split('\n')`
   requiring the literal token to start the line (after only whitespace). Also anchored the
   *index*-column assertion (`uniq_integration_read_source_configs_content`), which had the
   identical bug for a full-rename (declaration + index columns), using a **word-boundary**
   (`\bobject\b`) match against the extracted index clause — a boundary that `canonical_object`
   cannot satisfy because the character before the trailing `object` is `_` (a word character, so
   no `\b`).
   - Declaration-only rename (`object` → `canonical_object` on line 28 only): **RED** (1 failing
     test) → reverted → **GREEN**.
   - Full rename (declaration + both index column lists): **RED** (2 failing tests) → reverted →
     **GREEN**.
   - Regression check: 057's `kind` → `connector_kind` rename still **RED**s (β guard unweakened).

2. **[P2] The "recursive structural check" didn't check the property its name claimed.** Its only
   assertion was a sentinel-substring walk over `Object.values()` — it never inspected
   `Object.keys()`, so a bug that used a raw observed value **as an object key** (instead of one of
   the fixed enum tokens) would pass undetected; it also only checked for two *named* sentinels,
   not membership in any actual closed set, so an *unnamed* leaked value would pass too. Replaced
   with an **exact key-set assertion** (`assertPublicReportKeySetsExactly`) that walks the real
   report and asserts, at every level of `coverage`/`verdicts`/`residualNotes`, that the key set is
   exactly the fixed set the source can produce.
   - Mutation proof (production code, not a hand-built stand-in): temporarily changed
     `buildPublicObjectKeySummary()` to additionally key a bucket entry by the first observed raw
     `objectKey` string per status. Result: **RED**, and the assertion diff shows *two* unexpected
     keys — the injected sentinel (`SENTINEL_OBJECT_...`) **and** a second, non-sentinel value
     (`yet-another-object`) that the *old* sentinel-substring test (which only knows about two
     named sentinels) would **not** have caught. Reverted → **GREEN**.

3. **[P2] `unexpected`-status objectKeys never forced `INCONCLUSIVE`.** `computeObjectKeyVerdict`
   read only `byStatus.approved` and `divergence`; a `NOT_REQUIRED` verdict could coexist with
   unregistered objectKeys sitting in an out-of-vocabulary status. Added an `unexpected`-bucket
   check (same "known coverage defect forces INCONCLUSIVE" shape as the divergence check),
   constructed with the approved bucket empty and divergence clean so **only** the unexpected
   bucket flips the verdict (no confound).
   - Two new tests added; with the fix reverted (hunk removed, tests kept): **RED** (2 failing,
     `fail 2`, no other test disturbed) → fix restored → **GREEN**.

4. **[P3] `--probe` didn't scope `--dry-run`.** `buildDryRunReport()` always probed the full
   `TABLE_SPECS` regardless of `--probe`, so the "schema probe included" scoping claim above was
   false in `--dry-run` mode specifically. Threaded `probe` through (`tableSpecsForProbe(probe)` +
   `buildPlan(schema, { probe })`), added `probe` to the dry-run report object (so
   `renderHumanSummary`'s `probe=…` line isn't `(n/a)`), and wired `opts.probe` from `main()`.
   - Two new tests (zero `integration_read_source_configs`/`integration_external_systems` calls
     under `--dry-run --probe beta`/`gamma`); with the fix reverted: **RED** (2 failing) → fix
     restored → **GREEN**.

5. **[P3] γ coverage map omitted `integration_write_target_configs.object` (migration 064).** A
   second objectKey reference class in the same database. Added it to `RESIDUAL_NOTES` (new id
   `integration_write_target_configs_object_deliberately_out_of_scope`) and to the γ coverage
   `blindSpot` string, stating the table, the migration, and the reason (decision (γ)'s ledger
   scope is the READ-side tuple; the write-target table is a separate config-time-only surface).
   Pinned by test (both the note text and the `blindSpot` string are asserted to name the table).

6. **[P2] Committed prose attributed the #4594 non-reuse ruling to the ledger.** Grepped the
   RATIFIED ledger for `4594`, `data-source-exposure-inventory`, `non-reuse`, `must NOT be
   expanded`, `reused as the carrier` — **zero matches**:
   ```
   grep -n "4594\|data-source-exposure-inventory\|non-reuse\|must NOT be expanded\|reused as the carrier" \
     docs/development/database-system-integration-line-design-and-verification-20260724.md
   ```
   returns nothing. The ledger's §4.0/§3.0 define the four owner decisions (α/β/γ/δ) this work
   feeds; it says nothing about a carrier ruling for *this* script. That ruling came from the task
   brief given to the session that wrote this PR, not from the ledger — exactly the defect class
   (an artefact quoted as if it were in the contract) the ledger itself documents. Fixed the header
   comment, the `QUERY_ALLOWLIST` comment, and this PR body to say "the owner's task-brief ruling
   (not recorded in the ledger)". Also confirmed and stated explicitly: `scripts/ops/
   data-source-exposure-inventory.mjs`/`.test.mjs` and `.github/workflows/
   data-source-exposure-inventory.yml` live only on branch
   `claude/data-source-exposure-inventory-20260724` — `git merge-base --is-ancestor` against
   `origin/main` returns false; they are not on main.

## Live-Postgres verification — downgraded wording (P3)

The earlier version of this PR's Test plan stated as fact that a real, ephemeral local Postgres 15
instance was stood up, migrated, seeded, and hand-verified against `psql`, then discarded with **no
transcript captured**. This fix pass does not have a reproducible artefact for that claim and could
not independently re-verify it: this worktree has no `node_modules` (`pg` is not resolvable), so
`createPgExecutor()` cannot run here even though a local Postgres 15 binary is available. Per the
review instruction, the wording is downgraded rather than re-asserted:

- The **hermetic** test suite (fake executor, no DB) is the load-bearing, reproducible proof for
  everything in this PR, including all six fix-pass items above — every claim in "Fix pass" is
  backed by a pasted `node --test` command and its real output.
- The **live-Postgres** claims in the original Test plan section below are retained **as a
  historical record of what a prior session in this PR reported doing**, not as something this fix
  pass re-verified. Treat them as unreproduced. The `IS DISTINCT FROM` / missing-config-key
  semantics they were checking follow from the documented SQL standard behaviour independent of
  that specific run (see the updated `QUERY_ALLOWLIST` comment).

## Test plan

**Hermetic (CI-wired, no DB):**
```
node --test scripts/ops/gip-authority-substrate-inventory.test.mjs
```
```
ℹ tests 78
ℹ suites 21
ℹ pass 78
ℹ fail 0
```
78 tests (up from 69 pre-fix-pass): `assertReadOnlyAllowlist` structural read-only guard (rejects
non-SELECT, rejects a stacked/semicolon statement); `QUERY_ALLOWLIST` SQL-text semantic-invariant
pins (exact table set / quoted-literal set / WHERE predicate per tag); a **line/word-boundary
-anchored** schema-drift guard reading migrations `057_create_integration_core_tables.sql` /
`062_create_integration_read_source_configs.sql` directly off disk (§5's landmine: a previous
inventory was NO-GO for four assumed-schema errors — a citation in a header comment is not
load-bearing, this is, and after the fix pass it is anchored strongly enough to catch a
column-rename that merely shares a substring with the real name); frozen-registry
mutation-immunity; `bucketByRegistry`/`splitRowsByStatus` unit tests; `buildPlan`/`executePlan`
schema-probe-first tests; three-outcome verdict tests for both probes including the divergence-
and (new) unexpected-status-forces-INCONCLUSIVE cases; sentinel-injection **and** exact-key-set
values-free tests; fail-closed private-artefact tests against a real filesystem failure; `--probe`
zero-query tests for **both** report mode and `--dry-run`; `parseArgs`; the `.gitignore` coverage
test; and the new migration-064 out-of-scope pin.

**Wired in CI**: `.github/workflows/gip-authority-substrate-inventory.yml`, same shape as
#4594's (that workflow itself is branch-only, not on main — see Fix pass item 6) — checkout +
setup-node, `node --test`, path-filtered on the script/test/migrations/`.gitignore`.

**Verified against a real database** (reported by a prior session in this PR; not independently
reproduced in the fix pass — see "Live-Postgres verification" above): `initdb` + a local Postgres
15 (Homebrew), migrations 057 and 062 applied verbatim, five `integration_external_systems` rows
and five `integration_read_source_configs` rows seeded by hand (including one config with no
`object` key inside its `config` JSON). Ran the real CLI and hand-verified every count against the
seeded data and against direct `psql` queries (β all-statuses 4/5, β active-only 2/3, γ draft 1/1,
γ approved 3/3, divergence 2); dropped/renamed a table → `UNAVAILABLE` + `INCONCLUSIVE`, never a
silent zero; `--probe beta`/`--probe gamma` each verified live with the other table renamed away;
private-artefact fail-closed path verified live via a real `ENOTDIR` write failure. That local
database and its seeded rows have been discarded; no real customer data was ever touched.

## What I could NOT do, and why

- **Could not run this against a customer deployment** — that is a separate ops read-only
  authorization the task explicitly withheld. This PR does not request or assume it.
- **Could not independently reproduce the live-Postgres verification in the fix pass** — no `pg`
  package resolvable in this worktree (no `node_modules`). Downgraded the wording rather than
  re-assert it as independently verified fact (see "Live-Postgres verification" above).
- **Did not add the β/γ registries themselves** — those are the owner's decisions (β)/(γ) to make,
  not this tool's to infer; adding entries is explicitly the FORBIDDEN direction the task guarded
  against.
- **Did not probe `integration_write_target_configs.object`** (migration 064) — deliberately out
  of scope; stated with reason in `RESIDUAL_NOTES` and the γ `blindSpot` string (Fix pass item 5).

## Scope note (owner-recorded) — this PR is the inventory TOOL, not Lane A's inventory RESULT

This PR's CI job (`.github/workflows/gip-authority-substrate-inventory.yml`) runs `node --test`
against a **fake, in-memory executor with no real database connection** — there has never been a
live-Postgres run wired into this PR's CI, and (per the "Live-Postgres verification" section above)
even the one-off manual local-Postgres session a prior session in this PR reported is unreproduced
and explicitly downgraded to "reported, not independently verified." **This PR delivers the
inventory tool only.** The concrete (β) connector-kind alias map and (γ) objectKey backfill list —
the actual Lane A inventory RESULT — do not exist yet and cannot come from this PR: they require a
privately-authorized real run against a live deployment, which is a separate ops-gated read-only
authorization this task does not carry. Nothing in this PR should be read as if that inventory had
already been performed.

## Third pass (2026-07-25) — #4603 owner review of `18f503bf0`: P1/P2 findings closed

Owner review returned **HOLD** on `18f503bf0` with two P1s and three P2s. All five are fixed below;
each fix is proved with a pasted `node --test` command and its real output, and the two findings
the owner asked to be re-probed adversarially (P1(1)'s fail-open count coercion, P2(a)'s
unscanned-value gap) are each proved **RED before the fix, GREEN after**, by temporarily
re-introducing the exact defect in the committed source, re-running the suite, then reverting.

**Baseline before this pass** (`18f503bf0`, unmodified):
```
$ node --test scripts/ops/gip-authority-substrate-inventory.test.mjs
ℹ tests 78
ℹ pass 78
ℹ fail 0
```

### [P1] (1) `Number(value) || 0` fail-open in the count conversion — FIXED

Added `isSafeNonNegativeInteger(value)` (`typeof value === 'number' && Number.isSafeInteger(value)
&& value >= 0`) and made `bucketByRegistry()` return `{ status: 'invalid_count', reason }` instead
of silently zeroing a bad row; `executeKindInventoryPlan`/`executeObjectKeyInventoryPlan` turn that
into the probe's whole `status: 'unavailable'`, which the existing verdict layer already turns into
`INCONCLUSIVE` — never a silent "0 unregistered". The same fix applies to the divergence count
(now also requiring the query to return **exactly one row**, since a real `COUNT(*)` always does).

New tests cover all nine listed adversarial inputs (`undefined, null, '', 'abc', NaN, Infinity,
-1, 1.5, 2**53`) plus the positive control (`0`), at both the predicate layer and the
`bucketByRegistry`/`buildReport` end-to-end layer.

**Mutation probe — revert the fix, prove the new tests catch it:**
```
$ # bucketByRegistry's countKey conversion reverted to: const n = Number(row[countKey]) || 0
$ node --test scripts/ops/gip-authority-substrate-inventory.test.mjs
ℹ tests 124
ℹ pass 112
ℹ fail 12
```
12 of the new strict-integer tests go RED (all nine adversarial-input cases at the bucketing
layer, plus the three end-to-end `buildReport` INCONCLUSIVE-propagation tests). Reverted back:
```
$ # fix restored
$ node --test scripts/ops/gip-authority-substrate-inventory.test.mjs
ℹ tests 124
ℹ pass 124
ℹ fail 0
```

### [P1] (2) `--no-private-out` let a genuinely-owed private artefact exit SUCCESS — FIXED

`buildReport()` now computes `owed` unconditionally. When `noPrivateOut` is true and something is
owed, `privateArtifact` becomes `{ status: 'incomplete_owed_disabled', consumableByWave2: false }`
instead of the old bare `{ status: 'skipped_by_flag' }`. `main()` checks
`report.privateArtifact.consumableByWave2 === false` and returns exit code **1** (report is still
printed to stdout first) — both halves of the ruling ("either exit non-zero, or return an explicit
INCOMPLETE status") are satisfied, not just one. The test that pinned the old fail-open behaviour
(`'--no-private-out skips the write entirely, even when something is owed'`) is **updated, not
deleted**, to pin the new fail-closed shape; a sibling test pins the genuinely-nothing-owed branch
still getting a plain, consumable `skipped_by_flag`.

### [P2] (a) Values-free proof only checked keys/partial types, not values — FIXED

Reproduced the owner's exact finding against `18f503bf0` first, to confirm it before fixing:

```
$ # coverage.canonicalObjectKeyInventory.covers literal replaced with a customer-shaped string:
$ #   "acme-corp-warehouse-42-legacy-erp-object-code"
$ node --test scripts/ops/gip-authority-substrate-inventory.test.mjs
ℹ tests 78
ℹ pass 78
ℹ fail 0
$ # replaced again with the test's own seeded sentinel: "SENTINEL_NEVER_SEEN_OBJECT"
$ node --test scripts/ops/gip-authority-substrate-inventory.test.mjs
ℹ tests 78
ℹ pass 78
ℹ fail 0
```
Confirmed: 78/78 GREEN both times, exactly as the owner reported — that field was not scanned.

**Fix:** a CLOSED VALUE DOMAIN check. Every STRING leaf of a full `buildReport()` output is
digit-normalized (`s.replace(/\d+/g, '#')` — every dynamic component this module ever interpolates
is a validated non-negative safe integer, never raw DB content) and the resulting **set** is
asserted equal (not "member of an allowlist" — exact set equality, so a removal or an unauthored
addition also fails, not just a substitution) to a hand-frozen expected array, one per fixture
scenario (nothing-owed, something-owed/REQUIRED verdicts, unavailable+not_run/`--probe beta`,
divergence>0, unexpected-status-populated, both-tables-missing). Every NUMBER leaf is independently
asserted to be a non-negative safe integer. The expected arrays were captured once by running
`buildReport()` against the fixed source and pasted into the test file as frozen literals — not
re-derived at test time — so a later source mutation diverges from the frozen copy instead of
being validated against itself. Also added: an exact top-level `report` key-set assertion
(`generatedAt, mode, probe, coverage, verdicts, residualNotes, privateArtifact` — previously
unconstrained) and a referential-identity check (`report.residualNotes === RESIDUAL_NOTES`).

**Re-ran both probes against the FIXED suite:**
```
$ # covers replaced with "acme-corp-warehouse-42-legacy-erp-object-code" again
$ node --test scripts/ops/gip-authority-substrate-inventory.test.mjs
ℹ tests 124
ℹ pass 118
ℹ fail 6
✖ scenario: nothing owed (zero rows both probes) — closed value domain holds
✖ scenario: something owed (REQUIRED verdicts both probes, numeric interpolation exercised) — closed value domain holds
✖ scenario: unavailable column + --probe beta (not_run branch) — closed value domain holds
✖ scenario: objectVsConfigObjectDivergence > 0 (INCONCLUSIVE, exercises the R11/R12 reason templates) — closed value domain holds
✖ scenario: an out-of-vocabulary (unexpected) status bucket is populated (INCONCLUSIVE, exercises the R14/R15 reason templates) — closed value domain holds
✖ scenario: both tables missing (top-level UNAVAILABLE for both probes) — closed value domain holds

$ # covers replaced with the sentinel "SENTINEL_NEVER_SEEN_OBJECT" instead
$ node --test scripts/ops/gip-authority-substrate-inventory.test.mjs
ℹ tests 124
ℹ pass 118
ℹ fail 6
(same six scenario tests RED)

$ # reverted to the real covers string
$ node --test scripts/ops/gip-authority-substrate-inventory.test.mjs
ℹ tests 124
ℹ pass 124
ℹ fail 0
```
Both the non-sentinel and the sentinel injection now RED all six closed-value-domain scenario
tests — the exact gap the owner found is closed.

### [P2] (b) Report exposed caller paths and raw error messages — FIXED

`report.privateArtifact` no longer carries a real filesystem path or a raw `err.message` in any
branch: `written` is now `{ status, consumableByWave2 }` only; `write_failed_but_nothing_owed` is
`{ status, consumableByWave2, reasonCode: 'FILESYSTEM_WRITE_ERROR' }` (closed enum, not the fs
error text); the fail-closed thrown `Error` (owed + write failed) is now a fixed, values-free
sentence with no interpolated path or `err.message`. `renderHumanSummary()`'s "private artefact:"
line dropped the same path interpolation. An operator who used the default target can still
reconstruct it from `DEFAULT_PRIVATE_OUTPUT_DIR` + the report's own `generatedAt` — no information
is lost, only removed from a channel this file promises is values-free.

### [P2] (c) Private write followed symlinks and could overwrite an existing file — FIXED

`writePrivateArtifact()` now opens with `flag: 'wx'` (`O_CREAT|O_EXCL|O_WRONLY`). Per POSIX, this
fails `EEXIST` if the target path already names **anything** — a regular file, or a symlink node,
even a **dangling** one whose target does not exist — without ever following that symlink to write
through it, atomically (no separate check-then-write race). New tests cover all three refusals
(existing regular file with content proven untouched; dangling symlink with the link itself proven
unchanged and its target proven never created; symlink to a real file with that file's content
proven untouched) plus the clean-create positive control (content + `0600` mode).

### Net test count

```
$ node --test scripts/ops/gip-authority-substrate-inventory.test.mjs
ℹ tests 125
ℹ suites 26
ℹ pass 125
ℹ fail 0
```
125/125 (78 baseline + 47 net new/changed: strict-integer + fail-closed-count propagation +
`--no-private-out` incomplete-status + closed-value-domain + `writePrivateArtifact` refusal tests
+ a `renderHumanSummary` no-path-leak test).

**Known coverage gap, stated rather than hidden:** `main()`'s report-mode CLI path (the code that
turns `consumableByWave2 === false` into exit code 1) is not exercised by an automated test in this
suite — `main()` is not imported by the test file (true before this pass too), because its
report-mode path hard-wires a real `pg.Pool` via `createPgExecutor()` with no injection point, and
this worktree has no `node_modules` (`pg` unresolvable) to spawn it as a subprocess against either.
The exit-code branch is verified by code inspection only, not by an executed test.

## Rebase

Rebased onto `origin/main` (`d75d3b828`) with no conflicts; branch is 3 ahead / 0 behind. Pushed
with `--force-with-lease` (branch history was rewritten by the rebase). **Not merged.**

## Fourth pass (2026-07-25) — erratum + one coverage gap closed + one scope disclosure

Self-review after the third pass surfaced three items; all three closed/disclosed below rather
than left silent.

**Erratum — wrong count in the third-pass commit message.** The commit message for
`fix(ops): GIP authority inventory — close #4603 HOLD P1/P2 findings` says "125/125 tests pass (12
net new since 18f503bf0's 78…)". **12 is wrong** — that number is the *fail* count from the
P1(1) mutation probe (`bucketByRegistry` reverted to `Number(x) || 0`), not the net-new test count.
The correct figure, stated correctly in this PR body's "Net test count" section above, is **47**
(78 → 125, before the fourth-pass additions below took it to 127). Per this repo's standing
no-amend policy, this is stated here as a retraction rather than rewriting the commit.

**Closed-value-domain coverage gap.** All six scenarios in the third pass passed
`noPrivateOut: true`, so `'written'` and `'FILESYSTEM_WRITE_ERROR'` never appeared in any frozen
expected array — the general value-domain mechanism had never once run against a report whose
`privateArtifact` took the **actual write path**, which is exactly the branch P2(b) (report
exposed caller paths/raw error messages) was about. The narrower `assertPublicReportKeySetsExactly`
+ `doesNotMatch(/private\.json/)` checks already covered this branch, but not the general
value-domain mechanism. Added two more scenarios — a real write to a temp dir (`status: 'written'`)
and a real `ENOTDIR` write failure with nothing owed (`status: 'write_failed_but_nothing_owed'`) —
captured the same way as the existing six.

Mutation probe (re-reproduces the exact P2(b) shape, this time against the new scenario
specifically):
```
$ # 'written' branch: privateArtifact = { status: 'written', consumableByWave2: true,
$ #   path: path.relative(repoRoot, targetPath) }   <- re-added the path
$ node --test scripts/ops/gip-authority-substrate-inventory.test.mjs
ℹ tests 127
ℹ pass 125
ℹ fail 2
✖ scenario: privateArtifact actually WRITTEN (real temp-dir write, nothing owed) — closed value domain holds; no real path leaks in
✖ a 'written' privateArtifact carries no real filesystem path — closed status/flag only
$ # path removed again
$ node --test scripts/ops/gip-authority-substrate-inventory.test.mjs
ℹ tests 127
ℹ pass 127
ℹ fail 0
```
Both the new closed-value-domain scenario and the pre-existing key-set test catch the
regression independently — two layers, not one.

**Disclosure — `main()`'s generic stderr handlers still interpolate raw `err.message`.** The
P2(b) ruling was scoped to the report object (`~L679`, `report.privateArtifact`), which this pass
fully closed (no path, no err.message, anywhere in `privateArtifact` or the thrown fail-closed
Error). Out of scope for this pass, left as-is and disclosed rather than silently left: `main()`'s
three top-level `catch` blocks (`parseArgs` failure, `--dry-run` failure, and the report-mode
failure/DB-connection-failure path) still do
`` process.stderr.write(`... ERROR: ${err.message}\n`) ``. The most likely real-world value there
is a `pg` connection failure, which can carry a hostname. This is operator-facing stderr on a
direct terminal invocation, not the structured report object this file promises is values-free —
but it is the same class of leak vector the (b) ruling named, just at a different call site, and a
future pass should decide whether to close it the same way (closed reason code only) or accept it
as an intentionally-different channel (a human operator debugging their own DB connection, not
public/committed output).

Final test count after this pass:
```
$ node --test scripts/ops/gip-authority-substrate-inventory.test.mjs
ℹ tests 127
ℹ pass 127
ℹ fail 0
```

Pushed (fast-forward from the third pass, no rebase needed — `origin/main` had not moved):
`dc8778741`.

## Fifth pass (2026-07-25) — #4603 P2: overclaimed test comment (claim defect, not code defect)

**Finding:** the test titled `report.residualNotes is referentially IDENTICAL to the exported
RESIDUAL_NOTES` called that identity check *"the strongest possible pin on that section, stronger
than any content re-derivation."* Backwards: `report.residualNotes === RESIDUAL_NOTES` pins
**object identity** (that `buildReport()` passes the frozen array through unchanged, never a copy),
not note **content** — a wording change to a note keeps the same array reference, so that
assertion alone would stay green through it. No safety property was actually lost: content pinning
is done by the 8 CLOSED VALUE DOMAIN scenario tests, whose hand-frozen `EXPECTED_*` string lists
hardcode every residual-note description.

**Mutation-confirmed:**
```
$ # 'no_customer_deployment_connection' description inverted to claim the OPPOSITE — that this
$ #   script HAS been run directly against a customer deployment
$ node --test scripts/ops/gip-authority-substrate-inventory.test.mjs
  ✔ report.residualNotes is referentially IDENTICAL to the exported RESIDUAL_NOTES — the strongest
    possible pin on that section, stronger than any content re-derivation
  ✖ scenario: nothing owed (zero rows both probes) — closed value domain holds
  ✖ scenario: something owed (REQUIRED verdicts both probes, numeric interpolation exercised) — closed value domain holds
  ✖ scenario: unavailable column + --probe beta (not_run branch) — closed value domain holds
  ✖ scenario: objectVsConfigObjectDivergence > 0 (INCONCLUSIVE, exercises the R11/R12 reason templates) — closed value domain holds
  ✖ scenario: an out-of-vocabulary (unexpected) status bucket is populated (INCONCLUSIVE, exercises the R14/R15 reason templates) — closed value domain holds
  ✖ scenario: both tables missing (top-level UNAVAILABLE for both probes) — closed value domain holds
  ✖ scenario: privateArtifact actually WRITTEN (real temp-dir write, nothing owed) — closed value domain holds; no real path leaks in
  ✖ scenario: privateArtifact write FAILED but nothing owed (real ENOTDIR) — closed value domain holds; no real path or err.message leaks in
ℹ tests 127
ℹ pass 119
ℹ fail 8
```
The identity assertion stayed green; the 8 content-pinning scenarios went red — exactly backwards
from the claim. Reverted before commit (`git diff` clean against `dc8778741`).

**Fix:** reworded the test title and added a comment block stating plainly what the identity check
pins (object identity, not content) and where content pinning actually lives (the 8 scenario
tests). Also disclosed a **residual gap found while re-verifying**, not closed by any test in this
file: `assertClosedValueDomain` digit-normalizes every string (`s.replace(/\d+/g, '#')`) before
comparing, so a **digits-only** edit to a residual note — e.g. the ledger citation `§4 step 1.3,
§3.0 B-3` changed to `§9 step 9.9, §9.0 B-9` (a wrong citation, migration number left untouched) —
normalizes identically and is invisible to all 8 scenario tests **and** the identity test.
Mutation-confirmed narrower: that section-digits-only edit alone leaves all 127 tests in this file
green. (The separate `migration 064` digits in the same note *are* caught elsewhere, by a literal
non-normalized `assert.match` in the `RESIDUAL_NOTES` describe block — only the ledger
section-reference digits have no covering assertion anywhere in this file.) No code changed; this
is a test-comment-only fix, and the disclosed gap is left open rather than fixed in this pass.

Grepped this PR body for the same superlative (`strongest`, `stronger`, `pin on`) — no other echo
found; body text on this topic (the "Net test count" section, "a referential-identity check
(`report.residualNotes === RESIDUAL_NOTES`)") was already a plain factual statement, not a
superlative claim, so it is unchanged.

Test count after this pass (comment-only change, no assertion added/removed):
```
$ node --test scripts/ops/gip-authority-substrate-inventory.test.mjs
ℹ tests 127
ℹ pass 127
ℹ fail 0
```

### DIFF
--- a/.github/workflows/gip-authority-substrate-inventory.yml
+++ b/.github/workflows/gip-authority-substrate-inventory.yml
@@ -1,0 +1,47 @@
+name: GIP Authority-Substrate Inventory (β/γ probes)
+
+# Hermetic contract test for the two read-only, values-free inventory probes that feed decisions
+# (β) connector-kind registry and (γ) canonical object contract registry — see
+# docs/development/database-system-integration-line-design-and-verification-20260724.md §4.0.
+# No database: the script's tests inject a fake query executor (see
+# scripts/ops/gip-authority-substrate-inventory.test.mjs). Follows the same shape as the
+# `contract` job in multitable-permission-lists-postdeploy-smoke.yml / the sibling
+# data-source-exposure-inventory.yml pattern: checkout + setup-node only, `node --test` directly.
+
+on:
+  pull_request:
+    paths:
+      - '.github/workflows/gip-authority-substrate-inventory.yml'
+      - 'scripts/ops/gip-authority-substrate-inventory.mjs'
+      - 'scripts/ops/gip-authority-substrate-inventory.test.mjs'
+      # The schema-drift guard test reads these migrations directly off disk and asserts the
+      # kind/status/object column facts this script depends on — listed here so a PR that drifts
+      # either migration re-runs this check, instead of a later, unrelated PR going red.
+      - 'packages/core-backend/migrations/057_create_integration_core_tables.sql'
+      - 'packages/core-backend/migrations/062_create_integration_read_source_configs.sql'
+      - '.gitignore'
+  push:
+    branches: [main]
+    paths:
+      - '.github/workflows/gip-authority-substrate-inventory.yml'
+      - 'scripts/ops/gip-authority-substrate-inventory.mjs'
+      - 'scripts/ops/gip-authority-substrate-inventory.test.mjs'
+      - 'packages/core-backend/migrations/057_create_integration_core_tables.sql'
+      - 'packages/core-backend/migrations/062_create_integration_read_source_configs.sql'
+      - '.gitignore'
+
+permissions:
+  contents: read
+
+jobs:
+  contract:
+    name: authority-substrate inventory contract (values-free, schema-probe-first)
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+      - name: Run hermetic inventory tests (fake executor, no DB)
+        run: node --test scripts/ops/gip-authority-substrate-inventory.test.mjs
--- a/.github/workflows/gip-authority-substrate-inventory.yml
+++ b/.github/workflows/gip-authority-substrate-inventory.yml
@@ -5,8 +5,10 @@ name: GIP Authority-Substrate Inventory (β/γ probes)
 # docs/development/database-system-integration-line-design-and-verification-20260724.md §4.0.
 # No database: the script's tests inject a fake query executor (see
 # scripts/ops/gip-authority-substrate-inventory.test.mjs). Follows the same shape as the
-# `contract` job in multitable-permission-lists-postdeploy-smoke.yml / the sibling
-# data-source-exposure-inventory.yml pattern: checkout + setup-node only, `node --test` directly.
+# `contract` job in multitable-permission-lists-postdeploy-smoke.yml (on main), and the same
+# pattern as data-source-exposure-inventory.yml (#4594) — that workflow is NOT in this tree; it
+# lives only on branch claude/data-source-exposure-inventory-20260724, not on main: checkout +
+# setup-node only, `node --test` directly.
 
 on:
   pull_request:
--- a/.gitignore
+++ b/.gitignore
@@ -88,6 +88,11 @@ artifacts/integration-k3wise-onprem-preflight/
 # masked admin signoff metadata, and per-check evidence; operator-only.
 artifacts/integration-k3wise/internal-trial/
 
+# gip-authority-substrate-inventory.mjs (β/γ probes) private artefact: raw
+# connector-kind / objectKey strings observed in a deployment. Aggregate
+# counts only belong in committed output; the raw values are operator-local.
+artifacts/gip-authority-inventory/
+
 # Multitable on-prem release bundles are generated delivery artifacts. Keep
 # source manifests/runbooks in Git, not built archives and verification output.
 output/delivery/multitable-onprem/
--- a/scripts/ops/gip-authority-substrate-inventory.mjs
+++ b/scripts/ops/gip-authority-substrate-inventory.mjs
@@ -79,13 +79,27 @@
  *     [--probe beta|gamma|both] [--private-out <path> | --no-private-out] [--json]
  *   node scripts/ops/gip-authority-substrate-inventory.mjs --dry-run   # no DB required
  *
- * Exit codes:
+ * Exit codes: exit 1 is reached by TWO DIFFERENT routes through main() — they share the exit code
+ * but not the code path, and neither leaks anything about the underlying failure (see the CLI
+ * error-reason discipline right below the "CLI" section heading, and the exit-code contract test
+ * describe block in the test file, which pins the second route directly by calling main()):
  *   0  ran successfully (a MAPPING_REQUIRED / BACKFILL_REQUIRED verdict is information, not failure)
- *   1  unexpected runtime/DB error; OR the private artefact was owed (unregistered/unmapped count
- *      > 0) but could not be written; OR it was owed but --no-private-out disabled the write
- *      (report.privateArtifact.consumableByWave2 === false) — all three fail closed rather than
- *      reporting aggregate-only success while silently dropping the raw detail a human needs to
- *      act on it, or letting a caller read "ran successfully" as "this run is a complete inventory"
+ *   1  route A — an exception escaped buildReport()/buildDryRunReport()/parseArgs() (an unexpected
+ *      runtime/DB error, OR the private artefact was owed — unregistered/unmapped count > 0 — but
+ *      the write itself failed). Caught by main()'s try/catch; stderr gets a closed
+ *      CLI_ERROR_REASON token (never the caught error's own .message), never returns to the
+ *      `if (consumableByWave2 === false)` check below at all.
+ *   1  route B — buildReport() returns NORMALLY but the private artefact was owed and
+ *      --no-private-out disabled the write (report.privateArtifact.consumableByWave2 === false,
+ *      checked explicitly in main(), ~L1107). The report is still printed to stdout in this case; only
+ *      the exit code fails closed, so a caller relying on exit status alone (not reading stdout)
+ *      still cannot mistake this for a complete inventory.
+ *   Both routes fail closed rather than reporting aggregate-only success while silently dropping
+ *   the raw detail a human needs to act on it, or letting a caller read "ran successfully" as "this
+ *   run is a complete inventory". A failure to close the Postgres pool AFTER either route already
+ *   determined its result (0/1/2) is swallowed silently by design (see safeClose()) — it never
+ *   changes the exit code and never produces separate stderr output; the pool-close outcome itself
+ *   is not signalled on any channel.
  *   2  required input missing (DATABASE_URL absent outside --dry-run)
  */
 
@@ -897,6 +911,54 @@ function renderHumanSummary(report) {
 // CLI
 // ---------------------------------------------------------------------------
 
+// stderr values-free discipline (#4603 P2(a)): the report object's CLOSED VALUE DOMAIN (see the
+// describe block of that name in the test file) was proven for stdout/--json in an earlier round,
+// but the owner's probe showed stderr was NOT covered — an unrecognized flag echoed the operator's
+// literal argv value verbatim, and a `pg` import failure echoed a foreign Error's .message, which
+// on a `Cannot find package 'pg'` MODULE_NOT_FOUND carries an absolute filesystem path. Every
+// string that can reach stderr from parseArgs()/main() below is now one of these fixed tokens —
+// authored here, never argv- or env-derived — and nothing else.
+const CLI_ERROR_REASON = Object.freeze({
+  UNKNOWN_ARGUMENT: 'UNKNOWN_ARGUMENT',
+  INVALID_PROBE_VALUE: 'INVALID_PROBE_VALUE',
+  MISSING_PRIVATE_OUT_PATH: 'MISSING_PRIVATE_OUT_PATH',
+  MUTUALLY_EXCLUSIVE_PRIVATE_OUT: 'MUTUALLY_EXCLUSIVE_PRIVATE_OUT',
+  DATABASE_URL_REQUIRED: 'DATABASE_URL_REQUIRED',
+  RUNTIME_FAILURE: 'RUNTIME_FAILURE',
+})
+// Membership set, not just "is it a CliArgError" — see closedCliErrorReason() below for why the
+// extra check matters.
+const CLI_REASON_VALUES = new Set(Object.values(CLI_ERROR_REASON))
+
+// The ONLY error type parseArgs() is allowed to throw. `.reason` is always one of the frozen
+// CLI_ERROR_REASON tokens above — never an interpolated argv value — enforced by construction at
+// every throw site below (no template literal ever appears in a `new CliArgError(...)` call).
+class CliArgError extends Error {
+  constructor(reason) {
+    super(reason)
+    this.name = 'CliArgError'
+    this.reason = reason
+  }
+}
+
+// Maps ANY error that could reach a stderr write site in main() to a CLOSED reason token. This is
+// deliberately NOT `err instanceof CliArgError ? err.message : RUNTIME_FAILURE` — that shape makes
+// closedness a convention (a future `throw new CliArgError(\`unknown argument: ${a}\`)` would
+// silently re-open the exact leak this fixes, and every test below would stay green because it
+// only checks the class, not the value). Instead this re-validates that `.reason` is actually a
+// member of the frozen CLI_ERROR_REASON set; anything else — a foreign Error's .message, a
+// CliArgError constructed with a non-token string, a TypeError, a `pg` MODULE_NOT_FOUND carrying an
+// absolute path, a Postgres connection error carrying a hostname — collapses to the single generic
+// RUNTIME_FAILURE token. Mutation-confirmed (#4603 P2(a)): reverting this to the
+// `err instanceof CliArgError ? err.message : RUNTIME_FAILURE` shape (while still constructing
+// CliArgError with fixed tokens everywhere) keeps every test in this file green — the coverage gap
+// is closed only by asserting membership, so a direct unit test on this function's own contract
+// (feeding it a CliArgError built from a non-token string) is what actually pins it; see
+// 'closedCliErrorReason' below in the test file.
+function closedCliErrorReason(err) {
+  return err instanceof CliArgError && CLI_REASON_VALUES.has(err.reason) ? err.reason : CLI_ERROR_REASON.RUNTIME_FAILURE
+}
+
 function parseArgs(argv) {
   const opts = { dryRun: false, json: false, help: false, probe: 'both', privateOutPath: undefined, noPrivateOut: false }
   for (let i = 0; i < argv.length; i += 1) {
@@ -909,14 +971,14 @@ function parseArgs(argv) {
       case '--probe': {
         const next = argv[++i]
         if (!['beta', 'gamma', 'both'].includes(next)) {
-          throw new Error(`--probe must be one of beta|gamma|both, got: ${next}`)
+          throw new CliArgError(CLI_ERROR_REASON.INVALID_PROBE_VALUE)
         }
         opts.probe = next
         break
       }
       case '--private-out':
         opts.privateOutPath = argv[++i]
-        if (!opts.privateOutPath) throw new Error('--private-out requires a path argument')
+        if (!opts.privateOutPath) throw new CliArgError(CLI_ERROR_REASON.MISSING_PRIVATE_OUT_PATH)
         break
       case '--no-private-out':
         opts.noPrivateOut = true
@@ -929,11 +991,11 @@ function parseArgs(argv) {
         opts.help = true
         break
       default:
-        throw new Error(`unknown argument: ${a}`)
+        throw new CliArgError(CLI_ERROR_REASON.UNKNOWN_ARGUMENT)
     }
   }
   if (opts.privateOutPath && opts.noPrivateOut) {
-    throw new Error('--private-out and --no-private-out are mutually exclusive')
+    throw new CliArgError(CLI_ERROR_REASON.MUTUALLY_EXCLUSIVE_PRIVATE_OUT)
   }
   return opts
 }
@@ -968,7 +1030,11 @@ function printHelp() {
 
 async function createPgExecutor(databaseUrl) {
   // Lazy import: `pg` must never be required at module load, or a hermetic `node --test` CI job
-  // with no `node_modules` fails at import time before a single test runs.
+  // with no `node_modules` fails at import time before a single test runs. This is also the exact
+  // failure the owner's #4603 P2(a) probe used: in that same hermetic environment `import('pg')`
+  // rejects with a MODULE_NOT_FOUND Error whose .message is
+  // `Cannot find package 'pg' imported from <absolute path>` — main() below must never let that
+  // .message reach stderr/stdout verbatim.
   const { default: pg } = await import('pg')
   const pool = new pg.Pool({ connectionString: databaseUrl, max: 2 })
   return {
@@ -977,12 +1043,27 @@ async function createPgExecutor(databaseUrl) {
   }
 }
 
-async function main(argv = process.argv.slice(2), env = process.env) {
+// A pool-close failure after the run already computed its own result (success OR a mapped failure)
+// must never override that result, and must never leak the underlying close error's .message — same
+// values-free discipline as everything else reaching stderr. Without this, a throw inside main()'s
+// `finally { await executor.close() }` would reject main()'s own returned promise from OUTSIDE every
+// try/catch above it, bypassing closedCliErrorReason() entirely and reaching the entry-point wrapper
+// (see isEntry below) as an unmapped rejection.
+async function safeClose(executor) {
+  if (!executor) return
+  try {
+    await executor.close()
+  } catch {
+    // intentionally silent — see comment above
+  }
+}
+
+async function main(argv = process.argv.slice(2), env = process.env, { createExecutor = createPgExecutor } = {}) {
   let opts
   try {
     opts = parseArgs(argv)
   } catch (err) {
-    process.stderr.write(`[gip-authority-substrate-inventory] ERROR: ${err.message}\n`)
+    process.stderr.write(`[gip-authority-substrate-inventory] ERROR: ${closedCliErrorReason(err)}\n`)
     return 1
   }
   if (opts.help) {
@@ -995,7 +1076,7 @@ async function main(argv = process.argv.slice(2), env = process.env) {
   if (opts.dryRun) {
     let executor = null
     try {
-      if (databaseUrl) executor = await createPgExecutor(databaseUrl)
+      if (databaseUrl) executor = await createExecutor(databaseUrl)
       const report = await buildDryRunReport({ exec: executor ? executor.exec : null, probe: opts.probe })
       if (opts.json) {
         process.stdout.write(JSON.stringify(report, null, 2) + '\n')
@@ -1005,21 +1086,21 @@ async function main(argv = process.argv.slice(2), env = process.env) {
       }
       return 0
     } catch (err) {
-      process.stderr.write(`[gip-authority-substrate-inventory] ERROR: ${err instanceof Error ? err.message : String(err)}\n`)
+      process.stderr.write(`[gip-authority-substrate-inventory] ERROR: ${closedCliErrorReason(err)}\n`)
       return 1
     } finally {
-      if (executor) await executor.close()
+      await safeClose(executor)
     }
   }
 
   if (!databaseUrl) {
-    process.stderr.write('[gip-authority-substrate-inventory] ERROR: DATABASE_URL is required outside --dry-run\n')
+    process.stderr.write(`[gip-authority-substrate-inventory] ERROR: ${CLI_ERROR_REASON.DATABASE_URL_REQUIRED}\n`)
     return 2
   }
 
   let executor
   try {
-    executor = await createPgExecutor(databaseUrl)
+    executor = await createExecutor(databaseUrl)
     const report = await buildReport({
       exec: executor.exec,
       probe: opts.probe,
@@ -1035,24 +1116,39 @@ async function main(argv = process.argv.slice(2), env = process.env) {
     // The report is still printed above even in this case — a caller reading stdout gets the full
     // explanation — but the run must not be mistakable for a complete inventory by exit code alone
     // (see the header's Exit codes list and the ruling on ~L672 in the PR that added this check).
+    // Exit-code contract test coverage (#4603 P2(b)): see the 'main() exit-code contract' describe
+    // block in the test file — this line was previously reachable by none of the 127 tests.
     if (report.privateArtifact.consumableByWave2 === false) {
       return 1
     }
     return 0
   } catch (err) {
-    process.stderr.write(`[gip-authority-substrate-inventory] ERROR: ${err instanceof Error ? err.message : String(err)}\n`)
+    process.stderr.write(`[gip-authority-substrate-inventory] ERROR: ${closedCliErrorReason(err)}\n`)
     return 1
   } finally {
-    if (executor) await executor.close()
+    await safeClose(executor)
   }
 }
 
 const entryPath = process.argv[1] ? path.resolve(process.argv[1]) : null
 const isEntry = entryPath && entryPath === fileURLToPath(import.meta.url)
 if (isEntry) {
-  main().then((code) => {
-    process.exitCode = code
-  })
+  // Last-resort net, not the normal path: every branch inside main() already resolves to a mapped
+  // exit code via its own try/catch (see closedCliErrorReason() and safeClose() above). The reject
+  // handler here exists only in case something still escapes both — without it, Node prints the
+  // raw rejection (including any absolute path in its stack) straight to real stderr, and that
+  // print happens OUTSIDE main()'s own stderr writes, so nothing inside this file can intercept or
+  // redact it. Never interpolate the rejection reason here, for the same values-free discipline as
+  // every other stderr write in this file.
+  main().then(
+    (code) => {
+      process.exitCode = code
+    },
+    () => {
+      process.stderr.write(`[gip-authority-substrate-inventory] ERROR: ${CLI_ERROR_REASON.RUNTIME_FAILURE}\n`)
+      process.exitCode = 1
+    },
+  )
 }
 
 export {
@@ -1086,6 +1182,9 @@ export {
   buildDryRunReport,
   renderHumanSummary,
   parseArgs,
+  CLI_ERROR_REASON,
+  CliArgError,
+  closedCliErrorReason,
   main,
   REPO_ROOT,
   DEFAULT_PRIVATE_OUTPUT_DIR,
--- a/scripts/ops/gip-authority-substrate-inventory.mjs
+++ b/scripts/ops/gip-authority-substrate-inventory.mjs
@@ -142,11 +142,22 @@ const SCHEMA_PROBE_CAVEAT =
 const DEFAULT_PRIVATE_OUTPUT_DIR = 'artifacts/gip-authority-inventory'
 
 // ---------------------------------------------------------------------------
-// QUERY_ALLOWLIST — the ONLY SQL statements this script will ever execute.
-// Every statement is SELECT-only; assertReadOnlyAllowlist() enforces that
-// structurally at module load (same guard shape as #4594's, independently
-// implemented per the owner's task-brief carrier ruling — see the header
-// comment above; that ruling is not recorded in the ledger).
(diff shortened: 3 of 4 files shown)


zensgit:
## Independent verification of the values-free claim — it is NOT earned, and the gap is wider than the review said

Run by me at head **`18f503bf0`**, three probes plus a positive control. Baseline first: `node --test scripts/ops/gip-authority-substrate-inventory.test.mjs` → **78 pass / 0 fail**.

| probe (all into `coverage.canonicalObjectKeyInventory.covers`, a value position in the public report) | result |
|---|---|
| hardcoded, customer-shaped, **non-sentinel**: `'[LEAKED customer objectKey: acme_prod.sales_order_line] '` | **78/78 GREEN** |
| **positive control** — same position, value = the seeded sentinel `SENTINEL_NEVER_SEEN_OBJECT` | **78/78 GREEN** |

The control is the finding. The review scoped this as *"only leaks whose value equals a named sentinel are caught"* — but **even the sentinel passes here**, so the sentinel-injection test does not scan this field at all. `assertPublicReportKeySetsExactly` asserts **key sets**, and my mutation adds no key; the sentinel scan never reaches `covers`.

Two earlier probes of mine leaked nothing (`byStatus.draft.detail`, `objectKeyInventory.rows` are empty on the fixture path) — I am recording that rather than dropping it, because **a probe that leaks nothing proves nothing**, and that is the same failure mode as the test under discussion. One earlier attempt also produced a **`ReferenceError`-driven red** that could easily have been misread as "the guard caught it" — it did not.

**What this means for the claims, which is a retraction and not a softening:**
- script header L58 — *"It is structurally incapable of holding an observed kind or objectKey string"* — **not earned**;
- PR body — *"Proven two ways: (1) a sentinel-injection test … (2) an exact key-set test"* — **not earned**: (1) does not cover value positions in the descriptive fields, and (2) constrains keys only.

**The code is still values-free today** — this is a *proof* defect, not a live leak. The fix is to make the property real: assert a **closed VALUE domain** over every leaf of the public report (every leaf string ∈ authored labels ∪ fixed enum tokens; every leaf number an integer), so a leak in **any** field and **any** value reds — and keep the key-set assertion for the key-position case. Then re-run both probes above and show they red.

Also still open from the review: `assertPublicReportKeySetsExactly` never asserts the report's **own top-level key set**, so top-level additions are unconstrained in both key and value position.

zensgit:
## #4603 HOLD round-2 P2s — both closed at `a20ca9d35`

Re-derived head before starting: `a365b096f` (unchanged since HOLD). Worked in an isolated worktree
detached at that SHA (branch already checked out in another session's worktree), committed on top,
re-verified head unchanged immediately before push, plain-pushed (no force). New head: `a20ca9d35`.

### [P2] stderr values-free discipline
`parseArgs()` no longer throws `Error` with interpolated argv; it throws `CliArgError` carrying one
of a frozen `CLI_ERROR_REASON` token set. `main()`'s three stderr-write sites now go through
`closedCliErrorReason(err)`, which re-validates **membership** in that frozen set (not just
`instanceof CliArgError`) before trusting `.message` — so a future throw site that reintroduces
interpolation still collapses to `RUNTIME_FAILURE` rather than re-opening the leak.

Also closed a related gap found while implementing this: a rejecting `executor.close()` inside
`main()`'s `finally` previously escaped every catch block as an unhandled rejection printed on the
real stderr fd, bypassing the redaction entirely. Added `safeClose()` (swallow) + a reject handler
on the entry-point's `main().then(...)`.

**Probes — real subprocess, not just in-process monkeypatching** (`node --test`, all pass):
```
node scripts/ops/gip-authority-substrate-inventory.mjs --SENTINEL_MANUAL_CHECK_98213
  exit=1, stdout='', stderr='[gip-authority-substrate-inventory] ERROR: UNKNOWN_ARGUMENT'

DATABASE_URL=postgres://manual-check-host-XYZ/db node scripts/ops/gip-authority-substrate-inventory.mjs
  exit=1, stdout='', stderr='[gip-authority-substrate-inventory] ERROR: RUNTIME_FAILURE'
```
The pg-import-failure test carries a **positive control**: it first proves, in the same hermetic
(no `node_modules`) test environment this CI job runs in, that the raw `import('pg')` rejection
really does contain the absolute path fragment (`Cannot find package 'pg' imported from <path>`) —
so the "path fragment count is 0 on stdout+stderr" assertion isn't vacuous. Also added a `--help`
subprocess test (exit 0, no path fragment) to cover the one success-path public-output surface nothing
else touched.

Mutation-confirmed: reverting `closedCliErrorReason` to `instanceof CliArgError ? err.message : RUNTIME_FAILURE`
(no membership check) reds exactly the dedicated unit test for that function and nothing else — reverted after confirming.

### [P2] exit-code contract for `consumableByWave2 === false ⇒ exit 1`
Added an injectable `createExecutor` seam to `main(argv, env, { createExecutor })` (defaults to the
real `createPgExecutor`) so `main()` — the same function the entry-point wrapper invokes — is
testable without a database. New tests call `main()` directly with a fake executor and assert the
**real returned exit code**:
- `consumableByWave2 === false` (owed + `--no-private-out`) → exit 1
- positive control: nothing owed → exit 0
- `DATABASE_URL` missing outside `--dry-run` → exit 2 (documented, previously untested too)

Mutation-confirmed: flipping `report.privateArtifact.consumableByWave2 === false` to `=== true` at
the check reds both the exit-1 test and the exit-0 positive control, nothing else — reverted after
confirming.

Also corrected the header's Exit codes list, which previously implied a single route to exit 1;
there are two (an escaped exception vs. the explicit `consumableByWave2` check) and only the second
was documented.

**Tests: 127 → 138, all green** (`node --test scripts/ops/gip-authority-substrate-inventory.test.mjs`).
CI check `authority-substrate inventory contract (values-free, schema-probe-first)` passed on the
real runner at `a20ca9d35`.

Latent-slice constraints unchanged: no runtime consumer, no route, no wiring, no flag, no deploy;
`KNOWN_CERTIFIED_CONNECTOR_KINDS`/`KNOWN_CANONICAL_OBJECT_KEYS` still empty. Only the two script/test
files touched.

zensgit:
## Sixth pass — pushed head is now **`7833499c2`** (the body's last documented pass stopped two commits short)

A verifier caught that **the residual fixes were never on this PR**: they lived only on an unpushed local branch, so every probe run against the PR head still found all three residuals open. Proven three ways at the time — `git merge-base --is-ancestor` said not-an-ancestor, `git ls-remote` returned the old head, and `git branch -r --contains` was empty. They are now landed (the old head was their direct parent, so it was a fast-forward).

**The owner's own probe, re-run by me at the pushed head:**

```
entry frozen?  true
MUTATION THREW: TypeError            <-- closed
sql after attempt: SELECT column_name FROM information_schema.columns WHERE tab
```

Before: `QUERY_ALLOWLIST['probe.columns'].sql = "DELETE FROM …"` succeeded silently and `runQuery()` handed that exact DELETE to the executor — falsifying the module's own claim that the allowlist holds *"the ONLY SQL statements this script will ever execute"*. `assertReadOnlyAllowlist()` is load-time-only while `runQuery` re-reads `.sql` at call time. Each entry is now frozen; deleting only the freeze loop reds exactly the 4 new tests (144→140) and nothing else.

**Also in this pass, and it is a correction to my own text rather than to code:** the pg-control comment called the pre-fix fragility *"silently stop being exercised while still reading green"* — a **false-GREEN** — and then **three sentences later** recorded that replaying the old shape against a stub `pg` **did red loudly**. Both cannot be true. The old control was already fail-closed in the sense the doctrine cares about; what it was **not** was environment-independent. A contributor who ran `pnpm install` got a spurious red they could not act on — that is a **false-RED**. The isolated-directory fix is unchanged and still correct; only the rationale was wrong.

**Test count:** the body still says 127/127. The head runs **144/144**.

**Still open, disclosed rather than closed:**
- The two defensive guards (`safeClose`'s rejection-swallow at both call sites, and the entry-point reject handler) have **zero coverage** — isolated mutation leaves the suite green. They **are** load-bearing: with both removed the real CLI printed `CLOSE_FAIL_MARKER… host=customer-db.internal` twice plus **4 absolute-path fragments** to real stderr.
- `realpathSync` in the isolated-directory control is required, not cosmetic (macOS `os.tmpdir()` resolves under a symlink); removing it makes `main()` never run — exit 0, no output — which would be a silent pass. CI is unaffected (ubuntu, real directory).

PR #4619, opened 2026-07-26
docs(integration): ⟲OD3 [PROPOSED, pending owner ruling] — core-backend seam to leave B1a step 1.4, travel with the SQL path

zensgit:
**Doc-only**, zero non-`docs/` files. Draft — this amends **ratified** text, so it needs your sign-off before it lands.

## Retraction first

An earlier revision of this body described the ⟲OD3 amendment below as **already applied** — "the seam is removed," "so is certified source-column translation," the widening "narrows to the validator change alone." **No such ruling exists.** This PR is still awaiting your review; nothing below has been ruled. That earlier body-level claim is **withdrawn**. Everything that follows is rewritten as **PROPOSED, pending your ruling** — matching the file itself, which retracts the same false claim everywhere it appeared (Gate bullet 4, the marker legend, §3.2, §4 step 1.4, the §9.2 widening paragraph) and restores the ratified (pre-OD3) text — core-backend seam and translation **in-gate** — as what is currently authorized.

## The contradiction this PR proposes to resolve

§3.2's **entire** justification for the restricted probe-executor seam is that *probe SQL uses `GROUP BY`/`HAVING` shapes the structured facade cannot express*, and its **named consumer is B1b's SQL builders**. Decision **(δ)** ruled **v1 = (c)** (already ruled, 2026-07-25): B1a admits *connector-owned, named, certified HTTP probe actions only*, and **SQL builders stay unreachable**.

So under (c), the seam's **justification** and its **consumer** are *both* unreachable in v1. Building it in step 1.4 **would** land a **second-package (core-backend) production surface with zero v1 consumers**, under an "already authorized" reading — which is precisely the second-package-by-implication concern **owner finding P1-1** was raised about, arriving this time with the gate's own blessing. That is the case for the proposal; it is not a ruling.

## Four places the proposal touches — none of them ruled yet

1. **Gate bullet 4** — proposes removing the seam from the gate. At head the bullet states this as a **PROPOSED amendment, pending ruling**; until ruled, the bullet's authorized text is unchanged — seam and translation stay in-gate, per B-1/B-4/B-6.
2. **§3.2** — proposes reversing the seam's currently-stated ownership. Two earlier revisions of this document got ownership wrong **in opposite directions** — one said *"B1b needs…"* (wrong gate), the other moved it **into** 1.4's gate outright as a false ruling — both written **before** (δ) fixed v1 = (c), and both retracted at head. **As currently stated, ownership remains B1a step 1.4**; this proposal recommends the seam move **out**, to (b′) + B1c, if ruled.
3. **§4 step 1.4** — proposes removing the seam. It does **not** propose removing certified source-column translation wholesale. At head, B-6 is split into two named, separate obligations so neither can be misread as covering the other: **(A)** SQL target `fieldId` → source column, for a SQL builder — this has no v1 source under **already-ruled** decision (δ), independent of whether ⟲OD3 itself is ever ruled; **(B)** HTTP `fieldMap` → response path / field set, for the certified HTTP probe surface — this is **not** deferred by anything in this PR and remains **v1 work**, derived server-side from the approved resolution (B-1). This proposal's scope is the seam and obligation (A)'s home step only.
4. **The GIP-D0 §9.2 widening list** — approved **three** items, one of which (item 3) names the seam. The claim that this "narrows to the validator change alone" is the **exact wording the file itself retracts** as a false present-tense authority claim — that is not what happens even if this proposal is ruled. **If ruled**, item 3 would narrow to the **plugin-layer** restricted statement seam; only its **core-backend component** would defer, traveling with (b′) + B1c and requiring re-approval there — item 3 does not disappear. **Until ruled**, the three-item list stands unchanged, in full, as the sole current authoritative list.

This PR also adds two cross-references the earlier revision left dangling: §3.0 B-6 and the (δ) decision-table row were undifferentiated and SQL-framed, giving no pointer to the (A)/(B) split above — a reader landing at either site could take the SQL ruling (δ) as covering all of B-6. Both sites now point to the split and state plainly that obligation (B) is not deferred.

## Consequence for planning — stated as what a ruling would do, not as an approval

**If ruled**, B1a-3 would become **plugin-layer only** and would no longer span two packages: 1.4 would be one HTTP-only executor, 1.5 a removal plus a key-set pin, 1.6 freezing two contract shapes. The seam and obligation (A)'s translation would re-enter **in full** with **(b′) + B1c**, behind that gate. **Until ruled, none of this is in effect** — the currently-authorized text is the pre-OD3 state described above.

**Status refresh (2026-07-26).** The two conditions in the previous line have both changed, so it is restated rather than left stale: **#4610 is merged** (`3f60b3d7d6`), which lifted the HOLD, and **B1a-3 has been built** as DRAFT **#4625**. It was scoped **plugin-layer only, HTTP-only** — which is the *conservative intersection* of both branches of this still-unruled proposal, not an assumption that it was ruled. The core-backend statement seam and SQL source-column translation were **deliberately not built**, because under the currently-authorized (pre-⟲OD3) text they are in-gate but optional, and under the proposed text they are out — so building nothing is the only choice that is correct either way. **⟲OD3 is still PROPOSED and still needs your ruling**; #4625 neither depends on it nor presumes it.


zensgit:
## Closed without merge: no longer a B1a-3 prerequisite

#4625 merged as `ebf84f24974e4a2377e094a79ec63ae6718682b1` with the conservative HTTP-only/plugin-layer shape. It did not build the core-backend statement seam or SQL source-column translation, and it did not rely on this proposed amendment.

Therefore this draft is no longer needed to complete B1a-3. It remains **unratified** and is closed without merge; current ratified text is not rewritten by this closure. If a future `(b') + B1c` implementation needs the core-backend seam, that gate must raise a fresh amendment against then-current capability-spike evidence rather than reviving this stale snapshot.

No runtime, certification, arming, deployment or rollout authority is implied.

PR #4614, opened 2026-07-26
ci(integration-guard): wire the guard for safe required-check promotion (WIRING ONLY, not promoted)

zensgit:
## UPDATE 2026-07-26 — review round 6 (1×P1: import-abort false-green). EVIDENCE IS IN A COMMENT.

Head: re-derive with `git ls-remote origin claude/ci-integration-guard-required-wiring-20260725`.
Full round-6 write-up, ledger and retractions:
**https://github.com/zensgit/metasheet2/pull/4614#issuecomment-5084650012**

**Retraction that applies to THIS body.** The "two independent doors" section below, and M11's
"independent in both directions", are TRUE ONLY OF THE ROSTER-vs-GUARD-LOGIC MUTATION CLASS they
measure. Read as unqualified claims about "neutering" they are FALSIFIED: `process.exit(0)` at the
top of `integration-guard-assert-branch.mjs` removed the runtime door AND collapsed the whole
contract to a passing 1-subtest file (`node --test` treats a zero-exit child as a passing FILE),
with nothing red anywhere. A scope defect in the word "neutering", not a fabricated claim — the
counts are real. Closed in round 6 by a subtest-count floor in both callers plus Door E, which
executes it. Requalified in `integration-guard.yml` and the contract test's own headers too.

---

## UPDATE 2026-07-26 — review round 5 (1 P1 confirmed + 3 already-closed). EVIDENCE IS IN A COMMENT, NOT HERE.

Head: `fe0b977cc6160774f2c56aef8ed1b273357e5d63`. Re-derive rather than trusting this prose:
`git ls-remote origin claude/ci-integration-guard-required-wiring-20260725`.

**Round 5 breaks this body's house rule, deliberately and visibly.** Every previous round put its
evidence here. This body is 60,972 of GitHub's 65,536-character limit; the round-5 section is 12,416,
so the edit would simply be rejected. The full round-5 write-up — the confirmed P1 (the round-11
extraction left both extracted scripts entirely uncovered: three neutering mutations each left the
contract at 49 pass / 0 fail), the Door D fix, the M-A…M-G per-mutation exclusivity ledger, the
refreshed M9/M10/M11 counts, four claim retractions/stale-prose corrections, and one recorded
owner-pending gap (`apps/web/tests/utils/jsonAssist.spec.ts`) — is here:

**https://github.com/zensgit/metasheet2/pull/4614#issuecomment-5084442753**

Note for the owner: three of the four P1s in the round-5 brief were **already closed** at `1585a2123`
[...]

PR #4645, opened 2026-07-28
fix(sealed-export+activate): bind manifest digest and receipts, freeze activate policy, resolve AST bindings (owner post-merge review)

zensgit:
> **CURRENT EXACT-HEAD TRANSPORT UPDATE (2026-07-28)**
>
> Reviewed and approved for **latent merge only** at `2e7049032cb911fdd4b739de3e5d5bb4845b6c0e`. The branch update from the owner-authored fix commit changed none of the two scanner files byte-for-byte; it only absorbed current main. All checks completed successfully, including `test (20.x)` and `integration-guard`. Since #4614, `integration-guard` is the ninth required context on main, so the older "NOT required" statement below is historical and withdrawn. The older opening "Draft. Do not merge" is likewise superseded by this exact-head approval. This does **not** authorize S2, runtime wiring, flags, deployment, arming, or rollout.

Owner post-merge review of #4638 / #4587 (2026-07-27). All five findings reproduced before
fixing; each fix has a test that RED-ed on the unfixed code and a mutation that REDs the fixed
code. **Draft. Do not merge.**

**No rollback was needed: there is no runtime consumer.** Nothing under `lib/sealed-export/` is
wired to a route, scheduler, flag or package entry point, so nothing was live.

**S2, runtime wiring and arming remain HOLD until both P1s are closed.** This PR closes them in
code; the HOLD is the owner's to lift. No flag was touched, nothing was armed, #4579 stays HELD.

**The 30-token failure vocabulary is RATIFIED and frozen — no reason was added, removed or
renamed.** The new refusals reuse `SEALED_EXPORT_UPLOAD_SESSION_INVALID`,
`SEALED_EXPORT_CHUNK_DIGEST_MISMATCH` and `SEALED_EXPORT_CHUNK_UNDECLARED`. The scan's own finding
ids (`THROW_SITE_*`) are scanner output, not domain reasons, and are not part of that vocabulary.

---

## P1 — `manifestDigest` is now bound to the real manifest

`classifyChunkSubmission` derives `computeManifestDigest(manifest)` and compares it in constant
time. A caller-supplied digest that is not this manifest's fails closed with
`SEALED_EXPORT_UPLOAD_SESSION_INVALID`.

Reproduction, run against the pre-fix and post-fix `contracts.cjs` (same probe, same fixtures):

```
BEFORE                                            AFTER
classify(manifestA, FAKE, [], sub0) => ACCEPT     => REFUSED SEALED_EXPORT_UPLOAD_SESSION_INVALID
classify(manifestB, FAKE, [], sub0) => ACCEPT     => REFUSED SEALED_EXPORT_UPLOAD_SESSION_INVALID
```

`FAKE` is a well-formed lower-hex string that is not either manifest's digest; both manifests
accepted it, which is the "two different manifests share one fake session digest" case.

The arbitrary `D('manifest-session')` fixture is **deleted**. The test derives the session digest
from the manifest the way production does, and adds negatives for a fabricated key echoed by the
submission and by receipts, and for a *genuine* digest belonging to a *different* manifest — the
case no format check can catch.

## P1 — receipts are now bound to the manifest's chunk descriptors

Every accepted receipt is checked against the manifest descriptor at its index — chunk digest and
byte count, in **separate** guards — in both `classifyChunkSubmission` and `assertChunkSetComplete`.

```
BEFORE                                                          AFTER
classify(A, REAL, [receipt0-forged], sub0-forged)               => REFUSED
  => IDEMPOTENT_REPLAY                                             SEALED_EXPORT_CHUNK_DIGEST_MISMATCH
assertChunkSetComplete(A, [receipt0-forged, receipt1])          => REFUSED
  => { chunkCount: 2 }                                             SEALED_EXPORT_CHUNK_DIGEST_MISMATCH
```

The replay branch still compares submission-against-receipt, deliberately: the receipt was bound to
the descriptor in the accept loop, so the replay decision is transitively descriptor-bound. A
second submission-against-descriptor check there would be a second door for one property and would
make the neuter probe for either one non-discriminating; the code says so at the branch.

## P2 — the activate closed error policy is no longer runtime-mutable

`ACTIVATE_ERROR_POLICY_SOURCE` is module-private. The `Map` is built from its own frozen row
copies. `ACTIVATE_ERROR_POLICY` and `ACTIVATE_ERROR_FALLBACK` are deep-frozen projections.

The owner's exact mutation is now a test: set `ACTIVATE_RACE` to `200` plus driver text through the
public export, then call `mapActivateError()`.

```
BEFORE  AssertionError: expected 200 to be 409        (and the driver text was published)
AFTER   409 / 'User is no longer pending activation'
```

## P2 — the AST vocabulary scan resolves the binding, not the name

The scan now records every local name bound to `failSealedExport` — destructures, renames,
transitive aliases, member aliases, ESM import specifiers — and matches those. A callee it cannot
resolve statically is a **loud** finding rather than a silent miss.

| shape | before | after |
|---|---|---|
| direct call | 1 `THROW_SITE_REASON_UNDECLARED` | 1 `THROW_SITE_REASON_UNDECLARED` |
| `const { failSealedExport: fail } = …; fail('NOT_DECLARED')` | **0** | 1 `THROW_SITE_REASON_UNDECLARED` |
| alias of alias | **0** | 1 `THROW_SITE_REASON_UNDECLARED` |
| `const fail = v.failSealedExport` | **0** | 1 `THROW_SITE_REASON_UNDECLARED` |
| `v['failSealedExport']('NOT_DECLARED')` | **0** | 1 `THROW_SITE_REASON_UNDECLARED` |
| `v[name]('NOT_DECLARED')` (computed) | **0** | 1 `THROW_SITE_CALLEE_UNRESOLVABLE` |
| `pick(v)('NOT_DECLARED')` (call-of-call) | **0** | 1 `THROW_SITE_CALLEE_UNRESOLVABLE` |
| rename of a **different** export | 0 | 0 (negative control) |

Beyond the named finding, two further unreadable shapes are reported rather than skipped: a bound
name used as a **value** (`THROW_SITE_BINDING_ESCAPES`, outside the declaring module) and a computed
member read **of the vocabulary module** (`THROW_SITE_BINDING_UNRESOLVABLE`). An ordinary
`bytes[index]` is not treated as an alias attempt — that negative control is asserted, because
flagging every computed read would make the findings worthless.

## P3 — stale text corrected, retraction-first, at five sites

The owner named four; a multi-line sweep found a fifth (the same "desynchronising for the rest of
the file" overclaim in `sealed-export-failure-vocabulary.test.cjs`).

- `lib/sealed-export/failure-vocabulary.cjs` — "proposed, not ratified" → RATIFIED 2026-07-27, 30 tokens frozen
- `__tests__/sealed-export-failure-vocabulary.test.cjs` — same
- `lib/sealed-export/compliance-harness.cjs` — "blanked everything after" → **bounded** blind window
- `__tests__/support/sealed-export-source-scan.cjs` — same
- `__tests__/sealed-export-failure-vocabulary.test.cjs` — "desynchronising for the rest of the file" → bounded

Sweep result after the edits: **live claims 0**, five hits remaining and all of them quoted inside
an explicit withdrawal. (One unrelated live hit exists in
`docs/development/attendance-advanced-scheduling-rule-set-diagnostic-verification-20260526.md`,
about a scheduling rule set, not this vocabulary. Out of scope, untouched.)

---

## Mutations (each neutered separately; file hash confirmed changed and restored byte-identically)

| # | mutation | result |
|---|---|---|
| M1 | derived manifest-digest comparison always passes | RED — "fabricated session key, echoed by the submission" |
| M2 | receipt `chunkDigest` vs descriptor always passes | RED — "receipt digest alone disagreeing with the descriptor" |
| M3 | receipt `byteCount` vs descriptor always passes | RED — "receipt byte count alone disagreeing with the descriptor" |
| M4 | undeclared receipt index refusal removed | RED — "receipt beyond the manifest" |
| M5 | session digest derived from a constant, not the manifest | RED — "fabricated session key…" |
| S1 | renamed destructure no longer binds | RED — "renamed destructure must be resolved to the binding" |
| S2 | static element access no longer resolves | RED |
| S3 | unresolvable callee reported as nothing | RED — "must produce a finding, not silence" |
| S4 | escaping bound name reported as nothing | RED — "a bound name used as a value must be reported" |
| S5 | alias-of-alias no longer resolves | RED |
| F1 | policy rows no longer frozen | RED — "expected 200 to be 409" |
| F2 | exported policy table no longer frozen | RED |
| F3 | fallback no longer frozen | RED |
| F4 | Map shares the module-private source rows | **GREEN** — see non-claims |

M2/M3 fail on their own named assertion, so the two descriptor terms do not cover for each other.

## Where these run

- The six sealed-export suites are in `plugins/plugin-integration-core`'s explicit `scripts.test`
  `&&` chain, run by **`integration-guard`** → "Run plugin-integration-core CJS test chain".
  `integration-guard` is **NOT** one of the eight required contexts on `main`.
- The activate tests run under `pnpm --filter @metasheet/core-backend test`, the "Run core-backend
  tests" step of the `test` job in `plugin-tests.yml` — i.e. **`test (20.x)`, which IS required**.

Stated plainly rather than softened: the sealed-export half of this PR lands in a non-required
channel; the activate half lands in a required one.

## Non-claims

- **The `Map` holding distinct row objects is not asserted by any test.** With every published row
  frozen, row sharing has no behavioural discriminator — mutation F4 restores the sharing and the
  suite stays GREEN. The clone is defense-in-depth and satisfies the owner's literal instruction;
  the load-bearing property is the deep freeze, and that is what the tests prove.
- The AST scan is a **static source assertion**, not a behaviour proof, and it is a syntactic
  binding resolver, not a type-checked dataflow analysis. A binding that escapes through a function
  parameter or a return value cannot be followed; escapes are reported rather than resolved.
- No claim that S2 is unblocked, that any runtime path exists, or that the sealed-export surface has
  a consumer. It remains latent by authorization.
- No claim about the S3 dependency-resolution failures seen locally
  (`@aws-sdk/client-s3` unavailable in the local tree, three suites); CI's run is the authority.

### DIFF
--- a/packages/core-backend/src/routes/admin-users.ts
+++ b/packages/core-backend/src/routes/admin-users.ts
@@ -1751,11 +1751,11 @@ async function syncLegacyAdminProfile(userId: string, enabled: boolean): Promise
  * asserts set equality between the reasons thrown at `throwCoded` call sites and the policy
  * table's keys, and folding the fallback into the table would make that equality vacuous.
  */
-export const ACTIVATE_ERROR_FALLBACK = {
+export const ACTIVATE_ERROR_FALLBACK = Object.freeze({
   status: 500,
   code: 'ACTIVATE_FAILED',
   message: 'Activation failed',
-} as const
+} as const)
 
 /**
  * Closed policy table: every authored activation reason → the status and the message we publish.
@@ -1770,7 +1770,7 @@ export const ACTIVATE_ERROR_FALLBACK = {
  * Messages are authored HERE, not inherited from the thrown Error. The thrown message stays
  * useful for server-side logs; it is simply never part of the response.
  */
-export const ACTIVATE_ERROR_POLICY: Record<ActivateErrorCode, { status: number; message: string }> = {
+const ACTIVATE_ERROR_POLICY_SOURCE: Record<ActivateErrorCode, { status: number; message: string }> = {
   ACTIVATE_USER_REQUIRED: { status: 400, message: 'A target user id is required to activate' },
   ACTIVATE_USER_NOT_FOUND: { status: 404, message: 'User not found' },
   ACTIVATE_NOT_PENDING: { status: 409, message: 'User is not pending activation' },
@@ -1789,14 +1789,41 @@ export const ACTIVATE_ERROR_POLICY: Record<ActivateErrorCode, { status: number; 
   ACTIVATE_LINK_MISMATCH: { status: 409, message: 'Directory link points to a different user' },
 }
 
+type ActivatePolicyRow = Readonly<{ status: number; message: string }>
+
+/** A row the holder owns: a fresh frozen copy, never the source object. */
+function frozenPolicyRow(row: { status: number; message: string }): ActivatePolicyRow {
+  return Object.freeze({ status: row.status, message: row.message })
+}
+
 /**
  * Lookup view of the table above. A `Map` rather than an object index: a Map has no prototype
  * chain to borrow from, so a thrown `.code` of `'constructor'` / `'toString'` / `'__proto__'`
  * cannot resolve to a row, and the response path needs no computed property access at all
  * (which keeps the AST scan's "no computed read on the response path" rule strict).
+ *
+ * OWNER POST-MERGE FINDING (P2, 2026-07-27): this was `new Map(Object.entries(ACTIVATE_ERROR_POLICY))`
+ * over an EXPORTED, non-frozen table — so the Map held the very row objects any importer could
+ * reach. Owner's executed repro: setting `ACTIVATE_RACE` to 200 plus database text through the
+ * public export made `mapActivateError()` return it verbatim, defeating the closure this table
+ * exists to build. The source table is now module-private, the Map holds its own frozen copies,
+ * and the export is a frozen projection.
  */
-const ACTIVATE_ERROR_POLICY_LOOKUP: ReadonlyMap<string, { status: number; message: string }> =
-  new Map(Object.entries(ACTIVATE_ERROR_POLICY))
+const ACTIVATE_ERROR_POLICY_LOOKUP: ReadonlyMap<string, ActivatePolicyRow> = new Map(
+  Object.entries(ACTIVATE_ERROR_POLICY_SOURCE).map(([code, row]) => [code, frozenPolicyRow(row)]),
+)
+
+/**
+ * The published, DEEP-FROZEN projection. Exported only because the closure and mapping tests
+ * need to read the table's keys and rows; it is not the object the response path reads, and
+ * mutating it changes nothing.
+ */
+export const ACTIVATE_ERROR_POLICY: Readonly<Record<ActivateErrorCode, ActivatePolicyRow>> =
+  Object.freeze(
+    Object.fromEntries(
+      Object.entries(ACTIVATE_ERROR_POLICY_SOURCE).map(([code, row]) => [code, frozenPolicyRow(row)]),
+    ),
+  ) as Readonly<Record<ActivateErrorCode, ActivatePolicyRow>>
 
 /**
  * Error surface for `POST /api/admin/users/:id/activate`.
--- a/packages/core-backend/tests/unit/admin-users-activate-error-mapping.test.ts
+++ b/packages/core-backend/tests/unit/admin-users-activate-error-mapping.test.ts
@@ -174,6 +174,68 @@ describe('mapActivateError (activate endpoint error surface)', () => {
     expect(ACTIVATE_ERROR_FALLBACK.code).toBe('ACTIVATE_FAILED')
   })
 
+  /**
+   * OWNER POST-MERGE FINDING (P2, 2026-07-27), reproduced before it was fixed.
+   *
+   * `ACTIVATE_ERROR_FALLBACK` and `ACTIVATE_ERROR_POLICY` were exported and NOT recursively
+   * frozen, and `ACTIVATE_ERROR_POLICY_LOOKUP = new Map(Object.entries(ACTIVATE_ERROR_POLICY))`
+   * held THE SAME row objects. Owner's executed repro: an external importer set `ACTIVATE_RACE`
+   * to 200 plus database text and `mapActivateError()` returned it verbatim — the closure this
+   * change exists to build was runtime-mutable from any importer.
+   *
+   * The mutation is attempted through the PUBLIC export, exactly as an importer would. On a
+   * deep-frozen object a strict-mode assignment throws; that is a pass, so the attempt is
+   * caught and the BEHAVIOUR is what the assertions read.
+   */
+  it('survives the owner mutation: the policy is deep-frozen and the response never moves', () => {
+    const importerView = ACTIVATE_ERROR_POLICY as unknown as
+      Record<string, { status: number; message: string }>
+    const driverText = 'relation "user_orgs" column "secret_col" does not exist'
+
+    try {
+      importerView.ACTIVATE_RACE.status = 200
+      importerView.ACTIVATE_RACE.message = driverText
+    } catch {
+      // Frozen: the assignment threw. That is the desired outcome, not a test failure.
+    }
+    try {
+      importerView.ACTIVATE_RACE = { status: 200, message: driverText }
+    } catch {
+      // Frozen table: replacing a whole row threw.
+    }
+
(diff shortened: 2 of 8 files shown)


zensgit:
## HOLD — exact head `d93579cd9a573a480a04b13c63ff654005c3aed6`

**0 P1 / 1 P2. Do not merge this head.** The two sealed-export P1 fixes and the activation-policy deep-freeze fix are present and load-bearing, but the new AST binding resolver still has a false-green path in the exact class this PR claims to close.

### [P2] A bound member can be invoked or exported as a value without any finding

`sealed-export-source-scan.cjs` tracks identifier aliases and direct/static member callees, but it does not treat the **member expression itself** (`v.failSealedExport`) as a bound value. Therefore wrapping or escaping that member bypasses both reason collection and the advertised fail-loud path.

Executed against this exact head:

```text
positive:  v.failSealedExport('NOT_DECLARED')
           findings=1 THROW_SITE_REASON_UNDECLARED

viaCall:   v.failSealedExport.call(null, 'NOT_DECLARED')
           findings=0 reached=[]

viaApply:  v.failSealedExport.apply(null, ['NOT_DECLARED'])
           findings=0 reached=[]

viaBind:   const f=v.failSealedExport.bind(null); f('NOT_DECLARED')
           findings=0 reached=[]

Reflect:   Reflect.apply(v.failSealedExport, null, ['NOT_DECLARED'])
           findings=0 reached=[]

array:     const f=[v.failSealedExport][0]; f('NOT_DECLARED')
           findings=0 reached=[]

object:    const o={f:v.failSealedExport}; o.f('NOT_DECLARED')
           findings=0 reached=[]
```

The direct-call positive control proves the scanner is active; the wrapped forms returning zero prove a binding-value escape, not a parse failure or an everything-red scanner.

### Required fix

Resolve/reference the **bound expression**, not only its final call-site name. A bound member used anywhere except as the direct resolved callee or a statically handled alias must fail loud as `THROW_SITE_BINDING_ESCAPES` (or an existing exact equivalent). It is not necessary to teach the scanner `.call/.apply/.bind` argument semantics; refusing the escape is sufficient and safer.

Add discriminating controls for direct call (clean/declared and red/undeclared) plus `.call`, `.apply`, `.bind`, `Reflect.apply`, array storage and object storage. Neutering the member-expression escape guard must make those controls red.

### Independently verified as closed on this head

- Removing manifest-digest recomputation makes the fabricated-session negative red.
- Removing receipt digest binding makes the digest-only negative red.
- Removing receipt byte-count binding makes the byte-count-only negative red.
- Replacing `frozenPolicyRow()` with identity makes the public-mutation behavior test red and changes the mapped status to 200.
- Baseline sealed-export target suites and the 14 activation mapping tests pass; all 8 required checks plus `integration-guard` are green.

Green CI does not close the binding-value escape above because no current test represents it. Any push invalidates this exact-head verdict.

zensgit:
## Delta fix pushed — exact head `de663d554`

The HOLD at `d93579cd9` remains the historical verdict for that head. Its single P2 is addressed by commit `de663d554`; this comment does **not** yet authorize merge while fresh exact-head CI is running.

### Fix

The source scan now treats a static `failSealedExport` member read as tracked only when it is:

- invoked directly; or
- captured by a resolver-owned identifier declaration/assignment alias.

All other function-value uses fail loud as `THROW_SITE_BINDING_ESCAPES`, including `.call`, `.apply`, `.bind`, `Reflect.apply`, array storage and object storage. Transparent parentheses are followed so valid direct calls and aliases do not become false positives. The scan identity advances to `sealed-export/source-scan/ast-binding-v3`.

### Evidence on `de663d554`

- six sealed-export suites: PASS;
- full `pnpm --filter plugin-integration-core test` chain: PASS;
- neuter the new member-reference escape guard: RED on the `.call` discriminator (`0 !== 1`);
- positive controls: direct member call, member alias and parenthesized identifier/member forms remain clean;
- `git diff --check`: clean.

Fresh PR checks and one exact-head delta review remain before CLEAR/merge.

zensgit:
Exact-head short review: APPROVE at 2e7049032cb911fdd4b739de3e5d5bb4845b6c0e. The prior P2 scanner escape is closed: direct calls and supported aliases remain valid; .call/.apply/.bind/Reflect.apply plus argument/return/array/object escapes fail closed; disabling the escape guard reds the named test. The update from de663d554 changed neither edited scanner file. All checks are successful, including the now-required integration-guard and test (20.x). Authorization is latent merge only; S2/runtime/flags/deployment/arming/rollout remain closed.

PR #4674, opened 2026-07-29
feat(ops): isolate stock-prep parameter failures in diagnostic v4

zensgit:
## Scope

Draft-only diagnostic v4 for the flag-OFF stock-preparation bounded-candidate path. It follows the consumed v3 result `SOURCE_COUNT_STATEMENT_FAILED / PARAMETER_OR_TYPE` and adds one source-free bound-limit control before the unchanged real count statement.

This PR supersedes the Draft delivery shape in #4671 without moving #4671's exact head. Relates to #4628 and #4437.

## Behavior

- runs `SELECT TOP (@p1) CAST(1 AS BIGINT)` on the same opened source connection
- binds only `limit + 1` through the same `AddWithValue` scalar path
- keeps the real count SQL, parameter order, and values unchanged
- publishes only closed values-free control class and derived parameter-role fields
- skips the real count and bridge read if the control fails or returns an invalid scalar
- preserves connection cleanup and one-run/flag-OFF constraints

A control PASS does not exonerate the filter value or source predicate. It only separates the bound-limit binding path from the predicate/source path for a later `PARAMETER_OR_TYPE` result.

## Verification before push

- `node --test scripts/ops/build-stock-preparation-bounded-candidate-probe-sidecar.test.mjs`: 4/4 PASS
- PowerShell behavior assertions: 51/51 PASS under local PowerShell Core; the separate host assertion correctly fails because the local machine is not Windows PowerShell 5.1 Desktop
- PowerShell parser checks: PASS
- `git diff --check`: PASS
- six discriminating mutations: each RED, then exact source bytes restored
- independent exact-diff review: 0 P1 / 0 P2 / 0 P3

Windows PowerShell 5.1 and artifact evidence are pending fresh PR CI on exact head `023e816cdd1889e8b7fbe3f8e1b43c8d6a43a195`.

## Boundaries

- no entity-machine execution authorization
- no deployment, config mutation, feature-flag change, preflight, or external write
- no claim that a bounded candidate exists
- no retry of the consumed v3 authorization
- PR remains Draft after CI; any future one-run authorization requires separate exact-artifact review and owner action

PR #4675, opened 2026-07-29
feat(ops): add stock-prep predicate type diagnostic v5

zensgit:
## Context

Diagnostic v4 at `023e816cdd1889e8b7fbe3f8e1b43c8d6a43a195` safely isolated the observed source-count failure to the predicate/source side. The existing-evidence triage on #4628 then returned `BLOCKED` with every relevant type fact `UNKNOWN`, so changing the invocation type or binder would be speculative.

This Draft supersedes PR #4674 only as the candidate for a future separately authorized diagnostic. PR #4674 remains the exact historical source for the already-consumed v4 run and must not be moved or reinterpreted.

## Change

- reuse the already-open SQL Server connection after the v4 source-free bound-limit control;
- run only internally generated `SELECT TOP (0) [field] FROM [source]` through `CommandBehavior.SchemaOnly`;
- consume `FieldCount` and `GetDataTypeName(0)` only, never rows or row values;
- map the private exact driver type to a closed values-free class;
- permit the unchanged bounded count only for `STRING+TEXT`, `INT64+INTEGER`, or `BOOLEAN+BOOLEAN`;
- stop before count and bridge for cross-class or unknown relations;
- dispose the new reader and command on success and error paths;
- bump the sidecar provenance contract to v5.

`real` and `float` deliberately map to `OTHER`; `DIRECT_CLASS` is only the current attempt allowlist, not a SQL compatibility or success proof.

## Verification

- focused behavior suite under local PowerShell Core: 65 behavior assertions pass; the sole expected failure is the Windows PowerShell 5.1 host assertion;
- sidecar builder contract: 4/4 pass;
- `git diff --check`: clean;
- seven discriminating mutations each RED and restored GREEN: SchemaOnly, no row read, relation gate, direct-pair roster, reader cleanup, command cleanup, and approximate numeric classification;
- independent read-only adversarial review: 0 P1 / 0 P2. Two non-blocking P3 observations are outside this delta's safety claim: an unreachable internal invalid-type label and the pre-existing v4 scalar-command disposal shape.

Fresh Windows PowerShell 5.1 CI and the exact-SHA artifact are required before any further owner decision.

## Boundary

- Draft and unarmed.
- No entity-machine execution is authorized by this PR.
- No retry of v4 is authorized.
- No deployment, restart, config mutation, preflight, flag enablement, external write, or acceptance claim.
- Any future machine run requires an exact-head review, verified artifact digest, and a new one-run owner authorization on #4628.

Relates to #4628 and #4437.


zensgit (commented):
## Independent exact-head technical review — CLEAR for owner one-run decision

Review scope: v4 head `023e816cdd1889e8b7fbe3f8e1b43c8d6a43a195` through exact v5 head `6738d2a46d32528abb6e17d6897dbfeeb027923c`, plus the generated artifact and CI evidence.

**Verdict: 0 P1 / 0 P2.**

The metadata step is internally generated `SELECT TOP (0) <quoted field> FROM <quoted source>`, executed with `CommandBehavior.SchemaOnly` on the already-open source connection after the source-free bound-limit control passes. Production code does not call `Read()`, does not query catalog tables or `OBJECT_ID`, does not accept caller SQL, and never emits the raw exact type name. The exact type is reduced in-process to a closed values-free class. `CROSS_CLASS` and `UNKNOWN` both stop before the real count and bridge query; only the three explicit direct pairs permit the unchanged v4 count path.

Tuple validation closes all new stages/reasons/fields. Reader and metadata command cleanup are nested `finally` blocks. Seven discriminating mutations independently went RED and restored green, including removal of `SchemaOnly`, addition of row reading, bypass of both relation stops, direct-pair corruption, reader/command cleanup removal, and overclaiming `real/float` as decimal. Fresh CI is 16/16 successful, PR is 0 behind and CLEAN/MERGEABLE, and artifact/package/source-byte verification matches the exact head.

Non-blocking residuals:

- An impossible internal call with an invalid filter-value type degrades to the generic internal fail-closed path; the public CLI has a closed `ValidateSet`.
- The pre-existing one-shot scalar command helper relies on connection disposal rather than individually disposing its command. This predates the v5 delta and does not weaken the zero-row metadata gate.

This review supports only the owner's decision on one flag-OFF diagnostic execution. It is not merge approval and does not authorize deployment, config mutation, preflight, flag enablement, retry, or external write.

PR #4724, opened 2026-08-03
feat(stock-prep): RD-E2E — containerized entity-machine-free functional smoke

zensgit:
## Summary

Adds a `workflow_dispatch`-only CI lane (`.github/workflows/stock-preparation-e2e-functional-smoke.yml`) that stands up **real Postgres + SQL Server service containers** and functionally tests the stock-preparation feature **without an entity machine, without customer data, and without touching any production default**.

**This is functional testing on synthetic data only.** It is NOT the #4695 controlled acceptance (one real customer, the entity machine, an S6-B owner-gated single flag-on window — see `docs/operations/stock-preparation-s6a-sqlserver-onprem-runbook-20260731.md`) and does not substitute for it: no real-customer, scale, external-write, or rollout claim is established here.

### ⚠️ R2b update (2026-08-03) — retraction of the migration-073 claim + the actual current blocker

**Retraction, stated plainly first:** the "Two findings" section below (and the evidence block that
followed it, from dispatch run 30822049245) said `s6aProvisioning=FAIL`, `s6aFlagOnReason=PROVISIONING_FAILED`,
and described migration 073's under-scoped grant as reproducing "for any tenant's first S6-A provisioning
attempt" as a **live** blocker. **That is no longer true and should not be read as describing current
`main` or this branch.** Migration `074_...` (merged to `main` at `59579d5e6`, PR #4728) repairs exactly the
grant migration 073 left short. `grep -rn 074` against this PR's own diff returns 0 hits — 074 lives
upstream on `main`, not in this PR — and a fresh dispatch confirms `s6aProvisioning=PASS` now that this
branch is based on top of it. The FAIL/PROVISIONING_FAILED text below is a superseded, point-in-time
artifact kept only because deleting history is worse than labeling it stale; treat it as historical, not
current.

**What actually happens now that 074 has landed:** provisioning passes, and the S6-A flag-ON server
genuinely **CONSTRUCTS** the runtime (`s6aHealthFlagOn=PASS` — the internal route no longer 404s). The
walk gets substantially further than the old evidence block shows: it opens a run and writes a database
row before refusing. The new, real blocker is a refusal at the **first capture attempt**:
`s6aFirstRunErrorCode=SEALED_EXPORT_BINDING_UNQUALIFIED`, with a `CAPTURE_FAILED` row persisted
(`s6aDbRunStatusOnFailure=CAPTURE_FAILED`).

**Root cause (confirmed against the real code and reproduced locally, not guessed):**
`stock-preparation-sqlserver-source-authority.cjs`'s `normalizeExternalSystem()` (line 167) calls
`ownedCanonical(raw)` — i.e. `canonicalCodec.tryFreezeCanonical(value)` — on the **entire** external-system
object the runtime resolves at request time via `externalSystemRegistry.getExternalSystemForAdapter()`
(`stock-preparation-runtime-core.cjs:237`, `loadSource`). That adapter-shaped object
(`external-systems.cjs`'s `rowToAdapterExternalSystem`) always carries two members that break
canonicalization: **`createdAt` and `updatedAt`, both native JS `Date` instances** — because
`integration_external_systems.created_at`/`updated_at` are `TIMESTAMPTZ NOT NULL DEFAULT NOW()` (confirmed
in `packages/core-backend/migrations/057_create_integration_core_tables.sql`), no `pg` type-parser override
exists anywhere in this repo (grepped for `setTypeParser` repo-wide — zero hits), and
`rowToAdapterExternalSystem` copies the row's timestamps straight through with no serialization step.
`canonical-json.cjs`'s domain explicitly refuses `Date` (violation token `EXOTIC_OBJECT`) — confirmed by
feeding `canonicalCodec.tryFreezeCanonical()` the exact adapter shape twice, identical except for the two
`Date` members: `{ ok: false, violation: 'EXOTIC_OBJECT' }` with them present, `{ ok: true }` without them.
The existing unit test `sealed-export-s6a-source-authority.test.cjs` never caught this because its own
`externalSystem()` fixture hand-builds `createdAt`/`updatedAt` as **ISO strings**, not the `Date` objects
the real registry actually returns — that fixture was never exercised against a real DB round-trip.

**This is a genuine product defect, not a fixture-fixable one — confirmed, not assumed, before writing
this.** Two checks rule out a narrower explanation: (1) `plugin-integration-core/index.cjs` (~L236, ~L345)
constructs `externalSystemRegistry` exactly once and passes the SAME instance, unmodified, to both the HTTP
routes and `createStockPreparationSqlServerRuntime` — there is no narrowing wrapper anywhere on this path;
(2) the only existing unit-test stubs of `getExternalSystemForAdapter` in this plugin either omit
`createdAt`/`updatedAt` entirely or hand-supply them as strings — none exercises the real row shape. No
value this harness's external-system fixture supplies can change what `rowToAdapterExternalSystem` returns:
`created_at`/`updated_at` are populated by the database itself on every insert, are never null, and are
always `Date`-typed on read, for any external-system row created through the ordinary product API — real
customer data included. Fixing this needs a **production-file** change (e.g. projecting the external-system
value down to a canonical-safe shape before `ownedCanonical()`, either in `loadSource()` or inside
`normalizeExternalSystem()` itself) — genuinely required, and explicitly out of scope for this harness-only
PR. This PR does **not** attempt that fix; it reports the finding for owner triage instead.

**Findings cleared in this update (PR review, same PR):**
- The diagnostic `--require` preload used to root-cause the above (a stack-capturing monkey-patch of
  `failSealedExport`, observe-only, never changed behavior) has been **removed**. The evidence below was
  captured **without** it.
- `attemptS6ARealRun()`'s `if (!firstOk)` early-return path — the path that actually executes today — now
  always sets `S.s6aFlagOnRun` (`NOT_RUN`, the phase roll-up's own vocabulary — the per-step verdict that
  it genuinely ran and failed lives in `S.s6aFirstRun`) and `S.s6aFlagOnReason` (`FIRST_RUN_FAILED`) before
  returning. It previously fell through to the evidence write with that key entirely absent. The sibling
  `if (!dbOk)` early return had the same gap and was fixed the same way.
- Three flag-**normalization** arms were added (`'TRUE'`, `' true '`, `'True'`, all of which must ENABLE the
  route). The existing exact-match arms (`'1'`, `'yes'`) only prove values that fail **both** a naive strict
  `=== 'true'` comparison and the real `.trim().toLowerCase() === 'true'` comparison stay disabled — they
  cannot tell the two implementations apart, since every arm before this one happens to agree on the same
  (disabled) answer. These three values normalise to `'true'` under trim+lowercase but would not survive a
  naive strict-equals, so they are what actually pins the normalising comparison. Each arm constructs its
  own throwaway runtime material (no SQL Server/external-system/provisioning fixture) and asserts the health
  capability boolean reads `true` — the runtime genuinely constructed, not merely that the string was
  accepted.

Zero production files touched by this update — all four fixes are confined to
`scripts/ops/stock-preparation-e2e-functional-smoke.mjs`.

**Current values-free result block** (verbatim, dispatch run [30840756972](https://github.com/zensgit/metasheet2/actions/runs/30840756972), job `containerized functional smoke`, head `bbff82e614a29dd394a5d98222a3855bba50d7b1` — captured WITHOUT the removed diagnostic preload; supersedes the block further down this PR body, which is kept only as superseded history):

```
STOCK_PREPARATION_E2E_FUNCTIONAL_SMOKE
mode=functional_testing_synthetic_data
substituteForEntityAcceptance=false
flagOffHealthFlagOn=false
flagOffSiblingRouteHttp=200
flagOffHttp=404
flagOffCode=<none>
flagOffPass=PASS
exactMatch1HealthFlagOn=false
exactMatch1SiblingRouteHttp=200
exactMatch1Http=404
exactMatch1Code=<none>
exactMatch1Pass=PASS
exactMatchYesHealthFlagOn=false
exactMatchYesSiblingRouteHttp=200
exactMatchYesHttp=404
exactMatchYesCode=<none>
exactMatchYesPass=PASS
normTrueUpperHealthFlagOn=true
normTrueUpperPass=PASS
normTrueSpacedHealthFlagOn=true
normTrueSpacedPass=PASS
normTrueTitleHealthFlagOn=true
normTrueTitlePass=PASS
existingChainScriptExit=0
existingChainAuditActionsCovered=8/8
existingChainExternalWrite=false
existingChainLeakScanClean=true
existingChainPass=PASS
s6aExternalSystemHasCredentials=true
s6aExternalSystemRegistered=PASS
s6aProvisioningChanged=true
s6aProvisioning=PASS
s6aHealthFlagOn=PASS
s6aFirstRunHttp=422
s6aFirstRunMode=<unregistered>
s6aFirstRunStatus=<unregistered>
s6aBusinessLineCount=-1
s6aFirstRunExternalWrite=<unregistered>
s6aFirstRunErrorCode=SEALED_EXPORT_BINDING_UNQUALIFIED
s6aFirstRun=FAIL
s6aDbRunRowFoundOnFailure=true
s6aDbRunStatusOnFailure=CAPTURE_FAILED
s6aDbRunFailureReason=SEALED_EXPORT_BINDING_UNQUALIFIED
s6aDatabaseObservable=NOT_RUN
s6aReplayRun=NOT_RUN
s6aFlagOnRun=NOT_RUN
s6aFlagOnReason=FIRST_RUN_FAILED
overallPass=FAIL
```

Read together with the negative control from the SAME dispatch run (job `negative control`):

```
STOCK_PREPARATION_E2E_NEGATIVE_CONTROL
negativeControlProbeHttp=404
negativeControlReachedAssertion=true
negativeControlRedBecause=ASSERTION_FAILED_AS_INTENDED
expectedConclusion=RED
negativeControlChecksRecorded=1
negativeControlObservedConclusion=RED
```

Read this block as: **three blockers now cleared** (`PROVISIONING_FAILED` → `RUNTIME_NOT_CONSTRUCTED` →
now past both — the runtime genuinely constructs and a run genuinely opens and writes a database row), **one
blocker remains** (`SEALED_EXPORT_BINDING_UNQUALIFIED` at first capture, root-caused above as a product
defect this PR does not patch). `existingChainPass=PASS` — the 21-route chain is not regressed by any of
this. `s6aFlagOnRun=NOT_RUN` is the phase roll-up (the walk did not establish its claim); the per-step
verdict that it genuinely ran and was genuinely refused is `s6aFirstRun=FAIL` +
`s6aFirstRunErrorCode=SEALED_EXPORT_BINDING_UNQUALIFIED`. `s6aDatabaseObservable=NOT_RUN` and
`s6aReplayRun=NOT_RUN` are honest, and there is a caveat worth stating plainly rather than implying
"ready and waiting": the database-observable success-path assertion and the idempotent-replay arm this
task asked for (`assertS6ARunDatabaseObservable`, and the replay probe in `attemptS6ARealRun`) are wired
into the script but have **never once executed** — they carry no evidence of their own correctness, and
must be re-verified (not just trusted) once the `createdAt`/`updatedAt` defect is fixed and capture can
actually succeed. `overallPass=FAIL` is expected and by design (same doctrine as the earlier evidence
block): a real blocker stayed a real blocker across this update, honestly reported as `NOT_RUN`/`FAIL`
on the fields that own that meaning, not laundered into a partial `PASS`.

### Feature surface: 32 routes + 1 internal route

`http-routes.cjs` exposes 32 routes under `/api/integration/stock-preparation/...` plus the flag-gated `POST /api/integration/internal/stock-preparation/sqlserver-sealed-snapshot/run`.

**Exercised by this lane (22/33):** the existing T4 chain (`scripts/ops/stock-preparation-prep-line-extended-smoke.mjs`) drives 21 of the 32 —
`mvp/readiness`, `mvp/ensure`, `mvp/options/sync`, `mvp/sync/plan`, `mvp/sync/persist`, `mvp/source-runs/plm-bom`, `mvp/erp-materials/sync`, `projects`, `material-mappings/candidates`, `unit-conversions/candidates`, `material-mappings/candidates/sync`, `material-mappings/confirm`, `material-mappings/retire`, `unit-conversions/confirm`, `unit-conversions/retire`, `generation/run`, `exceptions/resolve`, `exceptions/bulk-resolve`, `exceptions`, `prep-lines`, `audit` — reproducing its documented `auditActionsCovered=8/8`, `externalWrite=false`. The internal S6-A route is separately exercised for its **gate** behavior (3 arms, all real HTTP POSTs, all real 404s).

**NOT exercised (11/32, honest gap — not a claim of completeness):**
- `GET target/readiness`, `POST target/ensure`
- `GET sandbox-target/readiness`, `POST sandbox-target/ensure`
- `POST options/sync` (top-level; only the `mvp/options/sync` alias is exercised)
- `POST mvp/source-runs/erp-materials` (only the `plm-bom` source-run variant is exercised)
- `GET snapshot-batches`, `GET snapshot-batches/:id/diff`, `GET snapshot-batches/:id/diff/rows` — **these ARE covered by the separate W6 postdeploy smoke** (`stock-preparation-mvp-postdeploy-smoke.mjs`), which this lane does not drive; T4 only imports pure helper functions from that module, not its route-calling `main()`
- `GET material-mappings/summary`
- `GET unit-conversions/summary`

This list was derived mechanically (grepping the `${API}/...` call sites in the T4 script and diffing against the `ROUTES` table in `http-routes.cjs`), not eyeballed off a CI log.

### What it drives

1. **Existing chain** — T4 run against a locally-booted core-backend server, reproducing 8/8 audit coverage over real HTTP.
2. **S6-A flag-OFF arm** — real HTTP call to the sealed-snapshot run route with the flag absent. Asserts a **conjunction**, not a bare 404: the `/api/integration/health` capability boolean reads `false`, a sibling **always-registered** route (`mvp/readiness`) answers `200` in the same server process (proving the server/auth/prefix are alive), and the S6-A path itself 404s. (See "S6-A DISABLED code" finding below for why the bare error-code assertion the original plan called for cannot be made truthfully.)
3. **S6-A exact-match arm** — `'1'` and `'yes'` must NOT enable the route (only `'true'`, after trim/lowercase, does) — same three-part conjunction.
4. **S6-A flag-ON real run** — provisions a first-party ephemeral SQL Server relation + external-system record + the two sealed-export Postgres authority roles, then attempts capture → private ingestion → generation kernel → apply. Provisioning now passes and the runtime genuinely constructs (see the R2b update above); the walk **still does not complete** — it is refused at first capture with `SEALED_EXPORT_BINDING_UNQUALIFIED`, a confirmed product defect this PR reports but does not patch. `s6aFlagOnRun=NOT_RUN` (the roll-up key; the per-step verdict `s6aFirstRun=FAIL` is what records that an HTTP call was genuinely made and genuinely refused), never falsely `PASS`.
5. **Negative control** (separate job) — reuses the SAME assertion machinery (`must()`, `startServer`, `s6aRunProbe`) with a deliberately wrong expectation, genuinely reddening (`negativeControlRedBecause=ASSERTION_FAILED_AS_INTENDED`).

### Update — the negative control was not actually load-bearing (now fixed)

A review found that `stock-preparation-e2e-negative-control.mjs`'s `process.exitCode = 1` was **unconditional** — it never read `CHECKS`/`anyFail`, and `must()` itself contains zero throws (it only pushes to `CHECKS` and returns a boolean). Proof: running the negative control with the real `must()` and with a fully neutered `must()` (always records `ok: true`) produced **byte-identical output**, including `negativeControlRedBecause=ASSERTION_FAILED_AS_INTENDED` and `exitCode=1`. The script's own stated purpose — "prove the harness's real assertion machinery is load-bearing" — was not achieved: this arm's RED conclusion never actually depended on the assertion failing. This mattered more once the main harness's `functional-smoke` job goes green (its RED currently rests entirely on the migration-073 grant failure below): once that's fixed, this lane would retain **no genuine can-fail demonstration**.

**Fixed**: `CHECKS` is now exported from the main harness (`stock-preparation-e2e-functional-smoke.mjs`), and the negative control derives its exit code the same way `main()` does — `anyFail = CHECKS.some((c) => !c.ok)` (plus a disclosed, deliberate delta: an empty `CHECKS` array also counts as failing, since this script makes exactly one `must()` call and an empty array can only mean that call never ran). Two new values-free fields, `negativeControlChecksRecorded` (count) and `negativeControlObservedConclusion` (`RED`/`GREEN`), are now printed alongside `expectedConclusion` so the evidence block itself — not just the process exit status — discriminates a real failure from a decorative one.

**Re-proved the same way**: copied the negative-control script verbatim into a proof harness that re-exports the real, committed `must`/`CHECKS` and stubs only the four infra functions (`s6aRunProbe` fixed at `404`, reproducing the exact flag-OFF scenario). Three arms:
- **Arm A (real `must()`)** → `exitCode=1`, `negativeControlChecksRecorded=1`, `negativeControlObservedConclusion=RED`.
- **Arm B (`must()` mutated in place to force `ok: true`)** → `exitCode=0`, `negativeControlChecksRecorded=1`, `negativeControlObservedConclusion=GREEN`; `probeHttp`/`redBecause`/`expectedConclusion` unchanged from Arm A (they describe the scenario, not the assertion verdict). Mutation reverted; `git diff` on the mutated file is empty.
- **Arm C (the one `must()` call deleted entirely, in the proof copy only)** → `exitCode=1`, `negativeControlChecksRecorded=0`, `negativeControlObservedConclusion=RED` — closes the disclosed `CHECKS.length === 0` fail-closed delta with its own mutation, rather than leaving it an untested assertion.

Arm A independently reproduced end-to-end in CI: dispatch run [30822049245](https://github.com/zensgit/metasheet2/actions/runs/30822049245), job `negative control` — `negativeControlChecksRecorded=1`, `negativeControlObservedConclusion=RED`, step concludes "Process completed with exit code 1." Real vs. neutered now genuinely differ, where before they did not.

### Two findings against the task brief's stated "verified facts"

**1. The S6-A route's DISABLED error code is unreachable when the flag is off — it is not unreachable dead code in general.** `registerIntegrationRoutes` (`http-routes.cjs` ~5003-5006) only adds the S6-A route to the Express app when `services.stockPreparationSqlServerRuntime` is truthy at plugin-construction time — i.e. only when the flag was already `'true'` at boot. When the flag is off, the route is never mounted, so a request to it falls through to the framework's generic unmatched-route 404 (no JSON envelope, no `error.code`) — it never reaches the `STOCK_PREPARATION_SQLSERVER_SEALED_SNAPSHOT_DISABLED` `HttpRouteError` inside the handler **in this specific arm**, because route registration checks only `Boolean(runtime)` while the DISABLED branch fires when the runtime is falsy. That branch is **not** unreachable in general, though: it also fires when `typeof runtime.run !== 'function'`, which route registration does not check — so a truthy-but-malformed runtime still gets its route registered AND still hits DISABLED at request time. This lane just never constructs that runtime shape, so it correctly never asserts the specific error code here. Confirmed by the existing unit test `testStockPreparationSqlServerSealedSnapshotInternalRoute` (`http-routes.test.cjs`), which asserts `disabled.routes.has('POST ' + routePath) === false` for the flag-off construction, and by this lane's own CI runs (`code=<none>` on all three flag-off/exact-match arms). The flag-OFF/exact-match arms were rewritten to assert the three-part conjunction above instead of a code that cannot fire in this arm.

**2. [RESOLVED by #4728/migration 074 — kept for history, see the R2b update above] Migration 073's provisioning-role grant was under-scoped for the initial-provisioning path it exists to serve.** `migrations/073_create_sealed_export_stock_prep_runtime_authority.sql` granted the provisioning role only `SELECT, INSERT` on `integration_sealed_export_signer_public_keys`, but `provisionInitialStockPreparationBinding` (`sealed-export-lifecycle-provisioning.cjs`) runs `trx.selectOneForUpdate(PUBLIC_KEY_TABLE, ...)` against it (`db.cjs:212`, `SELECT ... FOR UPDATE`) — a locking read that PostgreSQL requires `UPDATE` privilege for, independent of whether a conflicting row exists to lock. Confirmed against a real Postgres 16 service container: the raw driver error was `permission denied for table integration_sealed_export_signer_public_keys`, surfacing through the production failure contract only as the closed-vocabulary `SEALED_EXPORT_INTERNAL_ERROR`. This reproduced for any tenant's first S6-A provisioning attempt under migration 073's grants as written — it was not specific to this harness's fixture. **Migration `074_...` (merged to `main` at `59579d5e6`, PR #4728) has since repaired this grant; `s6aProvisioning=PASS` in the current evidence block confirms it.**

Two other, narrower fixture bugs were found and fixed along the way (both fully in this harness's own code, not production): the SQL Server fixture database needs `ALTER DATABASE ... SET ALLOW_SNAPSHOT_ISOLATION ON` (the capture path's own snapshot-capability proof requires it, off by default on a fresh `CREATE DATABASE`), and the provisioning spec's `bindingExpiresAt`/`signerExpiresAt` need seconds-precision ISO-8601 (`sealed-export-lifecycle-provisioning.cjs`'s `requiredFutureInstant` rejects the milliseconds a plain `.toISOString()` produces).

### Consequence for the S6-A flag-ON positive control [SUPERSEDED — see the R2b update above]

*(As of the original evidence block below, before 074 landed:)* Because provisioning never completed, the flag-ON HTTP walk (capture → private ingestion → generation kernel → apply, plus the idempotent replay) never ran. *(Current status, R2b:)* provisioning now passes and the runtime genuinely constructs — the missing half of the 404 control **is now established** (`s6aHealthFlagOn=PASS`: the route answers non-404 when the runtime is genuinely constructed). What remains missing is the **capture-success** control: the walk is refused at first capture (`SEALED_EXPORT_BINDING_UNQUALIFIED`, root-caused in the R2b update above as a product defect), so `s6aDatabaseObservable` and `s6aReplayRun` stay `NOT_RUN` — there is nothing to observe or replay until that refusal is fixed in a production file, which this PR does not do.

### Original values-free result block [SUPERSEDED — see "Current values-free result block" in the R2b update above] (verbatim, dispatch run [30822049245](https://github.com/zensgit/metasheet2/actions/runs/30822049245) — confirmed field-for-field against that run's job logs, not carried over unchecked from an earlier dispatch)

```
STOCK_PREPARATION_E2E_FUNCTIONAL_SMOKE
mode=functional_testing_synthetic_data
substituteForEntityAcceptance=false
flagOffHealthFlagOn=false
flagOffSiblingRouteHttp=200
flagOffHttp=404
flagOffCode=<none>
flagOffPass=PASS
exactMatch1HealthFlagOn=false
exactMatch1SiblingRouteHttp=200
exactMatch1Http=404
exactMatch1Code=<none>
exactMatch1Pass=PASS
exactMatchYesHealthFlagOn=false
exactMatchYesSiblingRouteHttp=200
exactMatchYesHttp=404
exactMatchYesCode=<none>
exactMatchYesPass=PASS
existingChainScriptExit=0
existingChainAuditActionsCovered=8/8
existingChainExternalWrite=false
existingChainLeakScanClean=true
existingChainPass=PASS
s6aExternalSystemRegistered=PASS
s6aProvisioningChanged=<unregistered>
s6aProvisioning=FAIL
s6aFlagOnRun=NOT_RUN
s6aFlagOnReason=PROVISIONING_FAILED
overallPass=FAIL
```

```
STOCK_PREPARATION_E2E_NEGATIVE_CONTROL
negativeControlProbeHttp=404
negativeControlReachedAssertion=true
negativeControlRedBecause=ASSERTION_FAILED_AS_INTENDED
expectedConclusion=RED
negativeControlChecksRecorded=1
negativeControlObservedConclusion=RED
```

(the last two fields are new — see "Update — the negative control was not actually load-bearing" above)

`overallPass=FAIL` is **expected and by design** (documented in the script): the job's own conclusion never reports green while a real-run attempt is incomplete, even though 3 of the 4 control arms plus the existing chain are fully green. A RED `containerized functional smoke` job here means "one arm honestly could not complete," not "the harness is broken" — the values-free block above is what actually carries the evidence either way.

### Explicitly out of scope / not done

- Does not deploy, arm, or touch the deployed entity host.
- Does not touch `#4695`'s frozen inputs or the S6-B owner-gated acceptance process.
- Does not change any production flag default — `MULTITABLE_STOCK_PREP_SQLSERVER_SEALED_SNAPSHOT_ENABLED` is set only inside this job's own spawned server-process environment.
- Does not touch branch protection or the `integration-guard` required check (verified still present in `required_status_checks.contexts`).
- Migration 073's grant scoping is fixed on `main` (migration 074, PR #4728) — no action needed here.
- The S6-A flag-ON positive control (route returns non-404 with the runtime actually constructed) **is now established** (`s6aHealthFlagOn=PASS`).
- Does **not** fix the `createdAt`/`updatedAt` canonicalization defect (R2b update above) — a production-file change, genuinely required, out of scope for this harness-only PR. Reported for owner triage, not patched here.
- Does not add a `pg` type-parser override, a monkey-patch, or any other workaround for that defect inside this harness — that would hide the bug rather than surface it.

Also added the same "ephemeral, throwaway container password — not a real credential" comment `sqlserver-smoke.yml` already carries, next to this workflow's `MSSQL_SA_PASSWORD` (it was the one committed service-container password in this repo missing that note).

**Note on which channel this proof lives in**: none of `harness syntax contract`, `negative control`, or `containerized functional smoke` are among the 9 `required_status_checks.contexts` on `main` (verified unchanged by this PR). The `pull_request`-triggered check suite for this PR is fully green; the negative-control fix's actual can-fail demonstration lives entirely in the `workflow_dispatch`-only channel (dispatch run 30822049245 above), which is not gated. "All PR checks pass" is not itself evidence for this fix — the dispatch-run log excerpts are.

## Test plan

- [x] Dispatch `functional-smoke`; confirm flag-OFF arm: capability=false + sibling route 200 + S6-A 404 (PASS)
- [x] Confirm exact-match arm: `'1'`/`'yes'` do not enable the route (PASS, same conjunction)
- [x] Confirm the new normalization arms `'TRUE'`/`' true '`/`'True'` DO enable the route + construct the runtime (PASS, run 30840756972)
- [x] Confirm existing-chain reproduces `auditActionsCovered=8/8`, `externalWrite=false` (PASS, still not regressed)
- [x] Migration 074 (main, PR #4728) confirmed to fix migration 073's grant: `s6aProvisioning=PASS` (was `FAIL`)
- [x] Confirm the S6-A flag-ON runtime genuinely CONSTRUCTS: `s6aHealthFlagOn=PASS` (was blocked on `RUNTIME_NOT_CONSTRUCTED`)
- [x] Root-cause the remaining `SEALED_EXPORT_BINDING_UNQUALIFIED` refusal at first capture: confirmed as `createdAt`/`updatedAt` (`Date`-typed) breaking `ownedCanonical`'s `EXOTIC_OBJECT` check — a production defect, reported not patched (R2b update above)
- [x] Remove the diagnostic `--require` preload used to root-cause the above; reproduce the same finding, and the same green arms, WITHOUT it (run 30840756972)
- [x] `s6aFlagOnRun` (+ reason) now always present in the evidence block, including on the `if (!firstOk)` early-return path that actually executes today
- [x] `s6aDbRunRowFoundOnFailure=true`, `s6aDbRunStatusOnFailure=CAPTURE_FAILED` — the run row IS written before the refusal (database-observable, on the failure path)
- [x] `s6aDatabaseObservable`/`s6aReplayRun` honestly `NOT_RUN` (not PASS, not silently dropped) — nothing to observe or replay while capture itself is refused
- [x] Dispatch `negative-control`; confirm it concludes RED for the intended reason (`ASSERTION_FAILED_AS_INTENDED`)
- [x] Negative control now derives its exit code from `CHECKS` (not hardcoded): proved real vs. neutered `must()` now produce **different** exit codes and `negativeControlObservedConclusion` (previously byte-identical)
- [x] The disclosed `CHECKS.length === 0` fail-closed delta is itself mutation-tested (Arm C above: `must()` call deleted → still `exitCode=1`/`RED`, not a silent PASS)
- [x] `node --check` both harness scripts and the workflow's `contract` job pass after the fix
- [x] Full PR check suite green (`gh pr checks 4724`); `negative-control`/`functional-smoke` correctly `skipping` on the `pull_request` event (workflow_dispatch-only) — exercised separately via dispatch run 30840756972

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/stock-preparation-e2e-functional-smoke.yml
+++ b/.github/workflows/stock-preparation-e2e-functional-smoke.yml
@@ -132,16 +132,24 @@ jobs:
           echo "SQL Server never became ready within timeout"
           exit 1
 
-      - name: Generate CI-local ephemeral secrets (never logged, scoped to this run only)
+      - name: Generate CI-local ephemeral secrets (masked, never logged, scoped to this run only)
         run: |
           set -euo pipefail
+          jwt_secret="$(openssl rand -hex 32)"
+          enc_key="$(openssl rand -hex 32)"
+          runtime_password="$(openssl rand -hex 16)"
+          provisioning_password="$(openssl rand -hex 16)"
+          echo "::add-mask::$jwt_secret"
+          echo "::add-mask::$enc_key"
+          echo "::add-mask::$runtime_password"
+          echo "::add-mask::$provisioning_password"
           {
-            echo "E2E_JWT_SECRET=$(openssl rand -hex 32)"
-            echo "E2E_INTEGRATION_ENCRYPTION_KEY=$(openssl rand -hex 32)"
+            echo "E2E_JWT_SECRET=$jwt_secret"
+            echo "E2E_INTEGRATION_ENCRYPTION_KEY=$enc_key"
             echo "E2E_S6A_RUNTIME_DB_ROLE=e2e_s6a_runtime"
-            echo "E2E_S6A_RUNTIME_DB_PASSWORD=$(openssl rand -hex 16)"
+            echo "E2E_S6A_RUNTIME_DB_PASSWORD=$runtime_password"
             echo "E2E_S6A_PROVISIONING_DB_ROLE=e2e_s6a_provision"
-            echo "E2E_S6A_PROVISIONING_DB_PASSWORD=$(openssl rand -hex 16)"
+            echo "E2E_S6A_PROVISIONING_DB_PASSWORD=$provisioning_password"
           } >> "$GITHUB_ENV"
 
       - name: Provision sealed-export Postgres roles, then run migrations with both roles visible
@@ -241,12 +249,16 @@ jobs:
           DATABASE_URL: postgresql://postgres:postgres@127.0.0.1:5432/metasheet
         run: pnpm --filter @metasheet/core-backend run migrate
 
-      - name: Generate CI-local ephemeral secrets (never logged, scoped to this run only)
+      - name: Generate CI-local ephemeral secrets (masked, never logged, scoped to this run only)
         run: |
           set -euo pipefail
+          jwt_secret="$(openssl rand -hex 32)"
+          enc_key="$(openssl rand -hex 32)"
+          echo "::add-mask::$jwt_secret"
+          echo "::add-mask::$enc_key"
           {
-            echo "E2E_JWT_SECRET=$(openssl rand -hex 32)"
-            echo "E2E_INTEGRATION_ENCRYPTION_KEY=$(openssl rand -hex 32)"
+            echo "E2E_JWT_SECRET=$jwt_secret"
+            echo "E2E_INTEGRATION_ENCRYPTION_KEY=$enc_key"
           } >> "$GITHUB_ENV"
 
       - name: Run negative control (this step is EXPECTED to fail)
--- a/.github/workflows/stock-preparation-e2e-functional-smoke.yml
+++ b/.github/workflows/stock-preparation-e2e-functional-smoke.yml
@@ -89,6 +89,7 @@ jobs:
         image: mcr.microsoft.com/mssql/server:2022-latest
         env:
           ACCEPT_EULA: 'Y'
+          # Ephemeral, throwaway container password (destroyed with the job) — not a real credential.
           MSSQL_SA_PASSWORD: 'E2eFunc!Probe2026'
         ports:
           - 1433:1433
--- a/.github/workflows/stock-preparation-e2e-functional-smoke.yml
+++ b/.github/workflows/stock-preparation-e2e-functional-smoke.yml
@@ -15,26 +15,60 @@ name: Stock Preparation E2E Functional Smoke
 # core-backend server-process environment for the duration of one arm — never in a committed file, never
 # as a repository/environment variable default, and the process is always restarted flag-OFF afterward.
 #
-# Two jobs:
-#   functional-smoke   — the real lane: flag-OFF arm, exact-match arm ('1'/'yes' must NOT enable it),
-#                         the EXISTING T4 extended-smoke chain reproduced against this local server, and
-#                         a best-effort real S6-A flag-ON run (capture -> ingestion -> generation kernel
-#                         -> apply) over a first-party ephemeral SQL Server relation. Any part that
-#                         cannot complete is reported NOT_RUN in the values-free result block — never as
-#                         PASS — and the job goes red if any check that DID run failed.
-#   negative-control    — deliberately RED. Reuses the SAME assertion machinery as functional-smoke and
-#                         asserts a wrong expectation on purpose, proving the harness can genuinely fail
-#                         (not a check "that did not happen").
+# Three jobs:
+#   contract            — pull_request only: syntax-checks the harness scripts (no server, no DB, no
+#                         secrets). Exists ONLY so GitHub registers this workflow_dispatch workflow while
+#                         it lives on a feature branch — GitHub will not let a brand-new workflow_dispatch
+#                         workflow be manually dispatched until it has run at least once via a trigger it
+#                         responds to (docs: a workflow_dispatch workflow must be on the default branch,
+#                         OR have already executed via another trigger, to appear as dispatchable). This
+#                         is the SAME idiom stock-preparation-prep-line-extended-smoke.yml and the
+#                         sealed-export-s2/s5 workflows already use in this repo.
+#   functional-smoke     — the real lane, workflow_dispatch only: flag-OFF arm, exact-match arm
+#                         ('1'/'yes' must NOT enable it), the EXISTING T4 extended-smoke chain reproduced
+#                         against this local server, and a best-effort real S6-A flag-ON run (capture ->
+#                         ingestion -> generation kernel -> apply) over a first-party ephemeral SQL Server
+#                         relation. Any part that cannot complete is reported NOT_RUN in the values-free
+#                         result block — never as PASS — and the job goes red if any check that DID run
+#                         failed.
+#   negative-control     — workflow_dispatch only, deliberately RED. Reuses the SAME assertion machinery
+#                         as functional-smoke and asserts a wrong expectation on purpose, proving the
+#                         harness can genuinely fail (not a check "that did not happen").
 
 on:
+  pull_request:
+    paths:
+      - '.github/workflows/stock-preparation-e2e-functional-smoke.yml'
+      - 'scripts/ops/stock-preparation-e2e-functional-smoke.mjs'
+      - 'scripts/ops/stock-preparation-e2e-negative-control.mjs'
+      - 'scripts/ops/stock-preparation-e2e-provision-postgres-roles-and-migrate.mjs'
   workflow_dispatch:
 
 permissions:
   contents: read
 
 jobs:
+  contract:
+    name: harness syntax contract (no server, no DB, no secrets)
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Setup Node.js 20.x
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20.x
+      - name: Syntax-check the E2E harness scripts
+        run: |
+          node --check scripts/ops/stock-preparation-e2e-functional-smoke.mjs
+          node --check scripts/ops/stock-preparation-e2e-negative-control.mjs
+          node --check scripts/ops/stock-preparation-e2e-provision-postgres-roles-and-migrate.mjs
+
   functional-smoke:
     name: containerized functional smoke (existing chain + S6-A flag arms)
+    if: github.event_name == 'workflow_dispatch'
+    needs: contract
     runs-on: ubuntu-latest
     timeout-minutes: 45
     services:
@@ -161,6 +195,8 @@ jobs:
 
   negative-control:
     name: negative control (deliberately RED — proves the harness can fail)
+    if: github.event_name == 'workflow_dispatch'
+    needs: contract
     runs-on: ubuntu-latest
     timeout-minutes: 15
     services:
--- a/.github/workflows/stock-preparation-e2e-functional-smoke.yml
+++ b/.github/workflows/stock-preparation-e2e-functional-smoke.yml
@@ -1,0 +1,221 @@
+name: Stock Preparation E2E Functional Smoke
+
+# RD-E2E: containerized, entity-machine-free FUNCTIONAL testing of the stock-preparation feature surface
+# (32 routes under /api/integration/stock-preparation/... plus the flag-gated S6-A internal route). This
+# is functional testing on synthetic data ONLY. It is NOT the #4695 controlled acceptance (one real
+# customer, the entity machine, an S6-B owner-gated single flag-on window — see
+# docs/operations/stock-preparation-s6a-sqlserver-onprem-runbook-20260731.md) and does NOT substitute for
+# it: no real-customer claim, no scale claim, no external-write claim, no rollout claim is established
+# here. This lane never deploys, never arms, and never touches any #4695 frozen input or the deployed
+# entity host.
+#
+# Dispatch-only. Postgres and SQL Server are ephemeral, first-party GitHub Actions service containers
+# seeded ONLY with synthetic fixture rows generated inside this job. The S6-A feature flag
+# (MULTITABLE_STOCK_PREP_SQLSERVER_SEALED_SNAPSHOT_ENABLED) is set ONLY inside this job's own spawned
+# core-backend server-process environment for the duration of one arm — never in a committed file, never
+# as a repository/environment variable default, and the process is always restarted flag-OFF afterward.
+#
+# Two jobs:
+#   functional-smoke   — the real lane: flag-OFF arm, exact-match arm ('1'/'yes' must NOT enable it),
+#                         the EXISTING T4 extended-smoke chain reproduced against this local server, and
+#                         a best-effort real S6-A flag-ON run (capture -> ingestion -> generation kernel
+#                         -> apply) over a first-party ephemeral SQL Server relation. Any part that
+#                         cannot complete is reported NOT_RUN in the values-free result block — never as
+#                         PASS — and the job goes red if any check that DID run failed.
+#   negative-control    — deliberately RED. Reuses the SAME assertion machinery as functional-smoke and
+#                         asserts a wrong expectation on purpose, proving the harness can genuinely fail
+#                         (not a check "that did not happen").
+
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  functional-smoke:
+    name: containerized functional smoke (existing chain + S6-A flag arms)
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    services:
+      postgres:
+        image: postgres:16
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: metasheet
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd "pg_isready -U postgres"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 10
+      mssql:
+        image: mcr.microsoft.com/mssql/server:2022-latest
+        env:
+          ACCEPT_EULA: 'Y'
+          MSSQL_SA_PASSWORD: 'E2eFunc!Probe2026'
+        ports:
+          - 1433:1433
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js 20.x
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20.x
+
+      - name: Setup pnpm
+        uses: pnpm/action-setup@v4
+        with:
+          version: 10.16.1
+          run_install: false
+
+      - name: Use empty npmrc for pnpm
+        run: |
+          echo "" > /tmp/empty-npmrc
+          echo "NPM_CONFIG_USERCONFIG=/tmp/empty-npmrc" >> "$GITHUB_ENV"
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Wait for SQL Server
+        run: |
+          for i in $(seq 1 40); do
+            if pnpm --filter plugin-integration-core exec node -e "
+              const sql = require('mssql');
+              new sql.ConnectionPool({server:'127.0.0.1',port:1433,user:'sa',password:'E2eFunc!Probe2026',database:'master',options:{encrypt:true,trustServerCertificate:true},connectionTimeout:3000})
+                .connect().then(p => p.close()).then(() => process.exit(0)).catch(() => process.exit(1));
+            "; then
+              echo "SQL Server ready on attempt $i"
+              exit 0
+            fi
+            echo "attempt $i: SQL Server not ready yet; sleeping 5s"
+            sleep 5
+          done
+          echo "SQL Server never became ready within timeout"
+          exit 1
+
+      - name: Generate CI-local ephemeral secrets (never logged, scoped to this run only)
+        run: |
+          set -euo pipefail
+          {
+            echo "E2E_JWT_SECRET=$(openssl rand -hex 32)"
+            echo "E2E_INTEGRATION_ENCRYPTION_KEY=$(openssl rand -hex 32)"
+            echo "E2E_S6A_RUNTIME_DB_ROLE=e2e_s6a_runtime"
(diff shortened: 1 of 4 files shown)


zensgit:
**Superseded by #4732, which merged as `55b76fd8d62a304cb308ae34d251a5bf1af4558e`. Closing unmerged — no functional content is lost.**

Ancestry is not usable evidence here (#4732 landed as a **squash**, which discards it), so this was verified at
the **content** level: of the 4 files this PR touches, 3 are byte-identical on `main`. The fourth,
`scripts/ops/stock-preparation-e2e-functional-smoke.mjs`, differs — `main` has 1672 lines to this head's 1181,
and **30 lines here are absent from `main`**. Each was examined rather than assumed superseded:

- **~14 lines** are fragments of `assertS6ARunDatabaseObservable` and the replay assertions that were expanded
  in later work. All still present on `main` — `assertS6ARunDatabaseObservable` ×5, `s6aDbGenerationStatus` ×2,
  `s6aReplayMode` ×2, `internal_noop` ×4, `business_line_count` ×7.
- **16 lines** are the RD-E2E comment block root-causing `SEALED_EXPORT_BINDING_UNQUALIFIED`, which stated the
  defect *"requires a production-file fix — out of scope for this harness-only PR"*. That comment is **correctly
  gone** from `main`: the production fix it asked for now exists there (`projectExternalSystem`, 2 hits in
  `stock-preparation-sqlserver-source-authority.cjs`). The comment was retired by the fix, not dropped.

This PR's diagnosis was the load-bearing contribution — it is what identified blocker #3 as a **production
defect rather than a fixture defect**, and #4732 acts on exactly that finding.

PR #4731, opened 2026-08-03
fix(sealed-export): S6-A source authority — project the external system to the members it reads before canonicalising

zensgit:
## Summary

Fixes the production defect PR #4724 (R2b) found and reported but did not patch: the S6-A walk was refused at first capture with `SEALED_EXPORT_BINDING_UNQUALIFIED` for **every production-shaped external-system record**, and adds the regression test whose absence let it ship.

**Stacked on `claude/stock-prep-e2e-functional-smoke-20260802` (#4724)** — that branch carries the dispatch-only smoke workflow this fix is proven with. The diff shown against that base is exactly this PR's own change.

## ⚠️ Status first: the fix is proven, the WALK STILL DOES NOT COMPLETE

Stated plainly before anything below is read: **this PR does not make the S6-A walk complete.** It closes the *first* blocker, and the successive dispatches then walked the run forward through **three** more, each diagnosed by executing the real modules rather than guessed at, and none of them worked around by loosening anything. A later round (**R3**, head `4e1d2cd8b`) then closed three re-review findings on top of that diagnosis, **without touching the one blocker still open** (the activation refusal, described below): the capture-root gap is now **fixed**, not merely worked around (see "FIXED (R3)" under Blocker 2 below); the projection drift-pin's dotted-only blind spot is **fixed** with a bracket-access sentinel (see the R3 addenda under "How drift is pinned" / "mutation-proved" below); and the run-status observability gap is **disclosed**, deliberately not patched (see "R3 — disclosed, not patched" near the bottom, just before "Evidence channel").

**Closed by this PR's production change:** `SEALED_EXPORT_BINDING_UNQUALIFIED` at first capture is gone, and the root cause is now **database-observed** rather than inferred — `s6aAdapterCreatedAtIsDate=true` / `s6aAdapterUpdatedAtIsDate=true`: `pg` really does hand back native `Date` for those TIMESTAMPTZ columns in this environment.

**How far the walk now gets** (dispatch [30852537120](https://github.com/zensgit/metasheet2/actions/runs/30852537120), head `4e1d2cd8b`, job `containerized functional smoke`): the run row reaches **`GENERATION_VERIFIED`** with `failure_reason=<none>`. Capture → private ingestion → generation kernel → apply all complete. It is refused at **activation**: `s6aFirstRunHttp=503`, `s6aFirstRunErrorCode=SEALED_EXPORT_INTERNAL_ERROR`. That is the one open blocker (below); it is unchanged by R3, on purpose.

**Therefore** `s6aDatabaseObservable=NOT_RUN`, `s6aReplayRun=NOT_RUN`, `s6aReplayNoSecondWrite=NOT_RUN`, `s6aFlagOnRun=NOT_RUN`, `overallPass=FAIL`. Those are honest, not laundered. The database-observable proof and the idempotent-replay arm described further down are **wired but have still never executed against a successful walk**; beyond the schema pre-flight (`s6aObservabilityQueriesResolve=PASS`, which proves their SQL parses and every column resolves) they carry no evidence of their own correctness and must be **re-verified, not trusted**, once activation succeeds.

### Full values-free result block

Verbatim, dispatch [30852537120](https://github.com/zensgit/metasheet2/actions/runs/30852537120), job `containerized functional smoke` (its own conclusion: `failure` — attributable entirely to the open activation blocker, `s6aFirstRunErrorCode=SEALED_EXPORT_INTERNAL_ERROR` with `s6aRuntimeRoleSelectForUpdate=…authority_state:42501…`; every check upstream of it reports `ok`/`PASS`), head `4e1d2cd8bd0b983f9727effb19ad5049e04c1d40` (current PR head, after R3):

```
STOCK_PREPARATION_E2E_FUNCTIONAL_SMOKE
mode=functional_testing_synthetic_data
substituteForEntityAcceptance=false
flagOffHealthFlagOn=false
flagOffSiblingRouteHttp=200
flagOffHttp=404
flagOffCode=<none>
flagOffPass=PASS
exactMatch1HealthFlagOn=false
exactMatch1SiblingRouteHttp=200
exactMatch1Http=404
exactMatch1Code=<none>
exactMatch1Pass=PASS
exactMatchYesHealthFlagOn=false
exactMatchYesSiblingRouteHttp=200
exactMatchYesHttp=404
exactMatchYesCode=<none>
exactMatchYesPass=PASS
normTrueUpperHealthFlagOn=true
normTrueUpperPass=PASS
normTrueSpacedHealthFlagOn=true
normTrueSpacedPass=PASS
normTrueTitleHealthFlagOn=true
normTrueTitlePass=PASS
existingChainScriptExit=0
existingChainAuditActionsCovered=8/8
existingChainExternalWrite=false
existingChainLeakScanClean=true
existingChainPass=PASS
s6aCaptureRootExistedBeforeRun=false
s6aCaptureRootCreatedByProduct=true
s6aExternalSystemHasCredentials=true
s6aExternalSystemRegistered=PASS
s6aAdapterRowFound=true
s6aAdapterCreatedAtIsDate=true
s6aAdapterUpdatedAtIsDate=true
s6aObservabilityQueriesResolve=PASS
s6aProvisioningChanged=true
s6aProvisioning=PASS
s6aHealthFlagOn=PASS
s6aFirstRunHttp=503
s6aFirstRunMode=<unregistered>
s6aFirstRunStatus=<unregistered>
s6aBusinessLineCount=-1
s6aFirstRunExternalWrite=<unregistered>
s6aFirstRunErrorCode=SEALED_EXPORT_INTERNAL_ERROR
s6aFirstRun=FAIL
s6aDbRunRowFoundOnFailure=true
s6aDbRunStatusOnFailure=GENERATION_VERIFIED
s6aDbRunFailureReason=<none>
s6aCaptureProbeConnectionConfig=DERIVED
s6aCaptureProbeSession=OPENED
s6aCaptureProbeSessionReason=<none>
s6aCaptureProbeDriverCode=NOT_RUN
s6aCaptureProbeMetadata=PASS
s6aCaptureProbeMetadataStrictObject=true
s6aCaptureProbeMetadataTypes=__databaseId:number|__isolationLevel:number|__productMajor:number|__sessionId:number|__snapshotEnabledState:number|__transactionId:string
s6aCaptureProbeOrderingKey=PASS
s6aCaptureProbeSourceRead=PASS
s6aCaptureProbeSourceReadRows=3
s6aCaptureProbeRowStrictObject=true
s6aCaptureProbeRowTypes=__databaseId:number|__isolationLevel:number|__productMajor:number|__sessionId:number|__snapshotEnabledState:number|__transactionId:string|payload:string|payloadVersion:string|rowId:string
s6aPostCaptureGenerationStatus=VERIFIED
s6aPostCaptureActivePointerRows=1
s6aRuntimeRoleSelect=active_pointers:ok|authority_state:ok|generation_audit:ok|generation_rows:ok|generations:ok|stock_prep_runs:ok
s6aRuntimeRoleSelectForUpdate=active_pointers:ok|authority_state:42501|generation_audit:42501|generation_rows:42501|generations:ok|stock_prep_runs:ok
s6aDatabaseObservable=NOT_RUN
s6aReplayRun=NOT_RUN
s6aReplayNoSecondWrite=NOT_RUN
s6aFlagOnRun=NOT_RUN
s6aFlagOnReason=FIRST_RUN_FAILED
overallPass=FAIL
```

`existingChainPass=PASS` with `existingChainAuditActionsCovered=8/8` and `existingChainExternalWrite=false` — **the 21-route chain is not regressed** by any of this. `externalWrite=false` holds throughout: `existingChainExternalWrite=false`, and no arm ever registers a non-`false` `s6aFirstRunExternalWrite`.

The negative control from the **same** dispatch (job `negative control`, whose `failure` conclusion is by design — which is exactly why its own fields, not its conclusion, are what must be read):

```
STOCK_PREPARATION_E2E_NEGATIVE_CONTROL
negativeControlProbeHttp=404
negativeControlReachedAssertion=true
negativeControlRedBecause=ASSERTION_FAILED_AS_INTENDED
expectedConclusion=RED
negativeControlChecksRecorded=1
negativeControlObservedConclusion=RED
```

That matters specifically because this PR adds module-level code to `stock-preparation-e2e-functional-smoke.mjs`, the module the negative control imports its assertion machinery from — an import-time throw would have produced the same `failure` conclusion with these fields gone. They are present and unchanged, so the lane's can-fail demonstration is intact and `overallPass=FAIL` still means what it says.

**None of these later blockers is a regression from this change.** R2b's walk never reached capture at all — it was refused upstream, in the source authority — so everything from `captureService.execute()` onward is territory no dispatch had ever entered.

### Blocker 2 — narrowed to nothing, then eliminated: `SEALED_EXPORT_CAPTURE_FAILED`

`failSealedExport`'s vocabulary is deliberately cause-free and this token has call sites in four production modules, so the token alone says nothing about which. A probe was added that walks the **same production sequence with the same production modules and the same production SQL builders** — the connection config is *derived by the production source authority*, not assembled — and it reported (run [30845083745](https://github.com/zensgit/metasheet2/actions/runs/30845083745) and [30845597530](https://github.com/zensgit/metasheet2/actions/runs/30845597530)):

```
s6aCaptureProbeConnectionConfig=DERIVED
s6aCaptureProbeSession=OPENED
s6aCaptureProbeMetadata=PASS
s6aCaptureProbeMetadataStrictObject=true
s6aCaptureProbeMetadataTypes=__databaseId:number|__isolationLevel:number|__productMajor:number|__sessionId:number|__snapshotEnabledState:number|__transactionId:string
s6aCaptureProbeOrderingKey=PASS
s6aCaptureProbeSourceRead=PASS
s6aCaptureProbeSourceReadRows=3
s6aCaptureProbeRowStrictObject=true
s6aCaptureProbeRowTypes=__databaseId:number|__isolationLevel:number|__productMajor:number|__sessionId:number|__snapshotEnabledState:number|__transactionId:string|payload:string|payloadVersion:string|rowId:string
```

The driver connects, `assertSnapshotCapability`'s SELECT-only/non-owner proof passes, the SNAPSHOT transaction begins, all three production SQL statements run, 3 rows stream, and the streamed row is exactly the nine aliases `service-core`'s `RAW_ROW_FIELDS` expects with types every one of its normalizers accepts. So the refusal was **not** in the source session, **not** in the SQL, and **not** in row normalization.

### FIXED (R3): nothing created the capture root — now the product does

`stock-preparation-runtime-core.cjs` derives `captureRoot = <artifactRoot>/capture`; `stock-preparation-sqlserver-runtime.cjs` hands it to the sealed-snapshot service as its `artifactRoot`; the service's first act in `execute()` was `fsPromises.mkdtemp(path.join(artifactRoot, 'sealed-export-s5-'))` — and **mkdtemp requires the parent to already exist**. There was no `mkdir` for that path anywhere in `lib/sealed-export/`, and a bare `catch` turned the resulting `ENOENT` into the cause-free `SEALED_EXPORT_CAPTURE_FAILED`.

The asymmetry was visible one directory over: `private-ingestion-blob-store.cjs` **does** `fs.mkdir(privateRoot, { recursive: true, mode: 0o700 })` for its own structurally identical derived root (`<ARTIFACT_ROOT>/private-ingestion`) — so the capture side was the odd one out, not a deliberate contract. And `docs/operations/stock-preparation-s6a-sqlserver-onprem-runbook-20260731.md` tells the operator to set `ARTIFACT_ROOT` but never mentions a `capture` subdirectory beneath it. **An operator who followed the runbook exactly got `SEALED_EXPORT_CAPTURE_FAILED` on their first run, with no cause in the error.**

**This is a genuine product gap, not merely an undocumented operator step** — the diagnosis the harness-workaround round left standing. `ARTIFACT_ROOT` itself is an operator-supplied env var the product may legitimately require to pre-exist (like the database roles) — but the **derived per-run capture directory beneath it** is the product's own construction, never handed to it, so deriving that path and never creating it was the product's gap, not the operator's, exactly like the blob store's sibling root.

**Fixed in R3** (`sqlserver-sealed-snapshot-service-core.cjs`'s `execute()`): `fsPromises.mkdir(artifactRoot, { recursive: true, mode: 0o700 })` now runs immediately before the `mkdtemp`, mirroring the blob store. The chmod is guarded on `mkdir`'s own return value (`undefined` unless *this call* created something), so an already-existing shared path is never re-moded.

**Disclosed scope of the fix.** `recursive: true` means this call will create the *whole* missing chain, `ARTIFACT_ROOT` included, if none of it exists — not only the derived `capture` subdirectory. A regression arm needs exactly that (a multiply-nested, never-created root must resolve), and the negative control below pins the mkdir block's exact bytes, so narrowing this to "only the last path segment" was not done here to avoid a second re-pin cycle over a distinction with no live consequence at this fix's current scope: a typo'd `ARTIFACT_ROOT` is now silently created rather than refusing, where it previously refused with `SEALED_EXPORT_CAPTURE_FAILED`. That silent-creation behaviour is new and is called out here rather than left to be discovered.

**Proven by construction, not argued.** The harness no longer pre-creates the capture root at all — `s6aCaptureRootExistedBeforeRun=false` in the values-free block above records that the harness never touched it, and `s6aCaptureRootCreatedByProduct=true` (checked via `fs.existsSync` after the run, whatever the run's own outcome) proves the directory exists afterward regardless — i.e. `execute()`'s own `mkdir` created it, not this harness.

**Test evidence** (`sealed-export-sqlserver-sealed-snapshot-action.test.cjs`, hermetic, no DB): a regression arm hands `execute()` a multiply-nested, never-created `artifactRoot` and asserts it resolves *and* the directory now exists — the test itself never mkdirs it. A negative control recompiles the core **in memory** (`Module._compile`; nothing written to `lib/sealed-export/`) with the mkdir block removed by an exact-byte match (asserted present exactly once, and the mutation asserted non-identical before compiling) — the same never-created root now refuses `SEALED_EXPORT_CAPTURE_FAILED` (the pre-fix behaviour), **and the same mutant still resolves against an already-existing root**, so the refusal is attributable to the removed mkdir specifically, not to the mutant being broken in general. The real on-disk module is re-read afterward and asserted byte-identical to before the test ran.

### FINDING (migration defect, open — reported, deliberately NOT patched): the runtime role cannot take the row lock activation needs

`SEALED_EXPORT_INTERNAL_ERROR` is `privateBoundary`'s catch-all for an **untrusted** throw — most often a raw driver error the failure contract discards on purpose. Migration 073's under-scoped grant (repaired by 074, PR #4728) had exactly that shape. It is that shape again, on a different table, and 074 did not cover it.

**Database-observed** (dispatch [30846460191](https://github.com/zensgit/metasheet2/actions/runs/30846460191), head `cb78fcaf8`):

```
s6aPostCaptureGenerationStatus=VERIFIED
s6aPostCaptureActivePointerRows=1
s6aRuntimeRoleSelect=active_pointers:ok|authority_state:ok|generation_audit:ok|generation_rows:ok|generations:ok|stock_prep_runs:ok
s6aRuntimeRoleSelectForUpdate=active_pointers:ok|authority_state:42501|generation_audit:42501|generation_rows:42501|generations:ok|stock_prep_runs:ok
```

The runtime role can read `integration_sealed_export_authority_state` but **cannot take a row lock on it** — SQLSTATE `42501`, insufficient_privilege. The probe used `WHERE false`, so that verdict is about the grant, not about any row.

The causal chain, read off the code and matching the observed stopping point exactly:

1. The run row stops at `GENERATION_VERIFIED`, and the generation row is `VERIFIED` with its active pointer present — so `createActivationExpectation()` succeeded and `activateGeneration()` is reached.
2. `generation-kernel.cjs`'s `activateGeneration()` calls `trx.readAuthorityStateForUpdate(authority)`.
3. `generation-store.cjs:381` implements that as `scopedDb.selectOneForUpdate(AUTHORITY_STATE_TABLE, …)` — a `SELECT … FOR UPDATE`, which PostgreSQL permits only with `UPDATE`/`DELETE`/`TRUNCATE` privilege, independent of whether a row exists to lock.
4. `migrations/073_…` grants the **runtime role** exactly `GRANT SELECT ON TABLE … integration_sealed_export_authority_state` — no row-lock capability.
5. `readAuthorityStateForUpdate` wraps the call in a bare `catch` that raises `SEALED_EXPORT_INTERNAL_ERROR` — the token observed, discarding the underlying `permission denied`.

**Attribution closed mechanically, not by elimination-by-eye.** Three tables report `42501` for the locking probe (`authority_state`, `generation_audit`, `generation_rows`), so "which one" had to be settled rather than assumed: `grep -n "selectOneForUpdate\|FOR UPDATE" lib/sealed-export/generation-store.cjs` returns exactly **one** call site — line 383, inside `readAuthorityStateForUpdate`. Nothing on the activation path takes a lock on the other two. The one thing that does read `generation_rows` before the failure point, `computeStoredRowsetDigest()`, uses the non-locking `countGenerationRows`/`listGenerationRows`, and the probe shows plain `SELECT` on that table is `ok`. So `authority_state` is the single reachable lock-denied read, and it is the one the failing step performs.

Migration 074 knows this exact rule — its own comment reads *"Column-level UPDATE is the minimal privilege PostgreSQL accepts for SELECT ... FOR UPDATE"* — and it repaired two sibling gaps (the provisioning role's column-level `UPDATE` on `signer_public_keys`, and the runtime role's `SELECT` on `generation_audit`). **This third instance of the same class was not covered.** It reproduces for any tenant's first S6-A activation under 073+074's grants as written; it is not specific to this harness.

**Not patched here, on purpose.** §4 of this lane's brief says a further refusal is diagnosed and reported, not worked around — and granting the missing privilege inside the harness would be exactly the wrong move: unlike the capture directory (a precondition a real operator can and must satisfy), this grant is a *migration's* responsibility, so satisfying it in CI would manufacture a green that no real installation would reproduce.

## The defect

`stock-preparation-sqlserver-source-authority.cjs`'s `normalizeExternalSystem()` called `ownedCanonical(raw)` on the **entire** adapter-shaped external-system record that `stock-preparation-runtime-core.cjs`'s `loadSource()` re-fetches via `externalSystemRegistry.getExternalSystemForAdapter()`.

That record is built by `lib/external-systems.cjs`'s `rowToAdapterExternalSystem()`, which copies `row.created_at ?? null` / `row.updated_at ?? null` straight through. `integration_external_systems.created_at`/`updated_at` are `TIMESTAMPTZ NOT NULL DEFAULT NOW()`, and there is **no `setTypeParser` anywhere in this repository** (grep over all `.ts/.js/.cjs/.mjs` outside `node_modules` = **0 hits**), so `pg` returns native JS `Date`. `canonical-json.cjs` refuses a `Date` with violation token `EXOTIC_OBJECT`.

Refusal therefore happened **before any field-set or identity check could run**: `SEALED_EXPORT_BINDING_UNQUALIFIED`, HTTP 422, run row persisted `CAPTURE_FAILED`.

Independently reproduced by executing the real modules (not taken on trust from #4724):

```
adapterMembers=capabilities,config,createdAt,credentials,id,kind,lastError,lastTestedAt,name,projectId,role,status,tenantId,updatedAt,workspaceId
createdAtIsDate=true
codecDateRecord=violation:EXOTIC_OBJECT
codecStringRecord=ok
A_isoStrings=OK
B_dateTimestamps=REFUSED:SEALED_EXPORT_BINDING_UNQUALIFIED
C_timestampsOmitted=OK
D_dateOnUnrelatedMember=REFUSED:SEALED_EXPORT_BINDING_UNQUALIFIED
E_mapOnCreatedAt=REFUSED:SEALED_EXPORT_BINDING_UNQUALIFIED
```

`D` is the exclusive control: a `Date` on `lastTestedAt` — a member nothing reads — refuses just the same, so it is **Date-ness reaching the codec**, not the key name.

### Why no test caught it

Every fixture on this path hand-built the timestamps as **ISO strings**: `sealed-export-s6a-source-authority.test.cjs`'s `externalSystem()` fixture does, the real-DB S6-A test never touches `integration_external_systems` or `getExternalSystemForAdapter`, and `sealed-export-s6a-product-runtime.test.cjs` stubs `getExternalSystemForAdapter` with a throw. **Every test path avoided the shape production actually produces.**

## The fix — the narrowest correct one

The module reads exactly eight members off the record — re-derived mechanically from its own source, not taken on trust: `config`, `credentials`, `id`, `kind`, `role`, `status`, `tenantId`, `workspaceId`. It greps **0** times for `createdAt`/`updatedAt`.

`normalizeExternalSystem()` now projects the record to exactly those eight members and canonicalises **the projection**. The Dates never reach the codec.

What is deliberately **not** done:

- `canonical-json.cjs` is **not** widened to accept `Date` — a security-sensitive contract, and this is not a reason to loosen it.
- `rowToAdapterExternalSystem()` is **not** changed — a shared adapter contract with other consumers.
- `ownedCanonical`, `exactObject`, `refuse()`, the field sets and every identity check are **unchanged**, and all of them still run, on the projection.

Two behaviour-preserving details worth reading:

- The container is still screened by the codec's own `isStrictPlainObject` **before** any key is read, so proxies, accessor properties, symbol keys, non-enumerable properties and hostile prototypes still refuse exactly as they did when the whole record was canonicalised.
- Members are copied only when they are **own** properties, so an absent `workspaceId` still reaches the `?? null` comparison as `undefined` rather than being written into the projection as an `undefined` the codec would refuse. An absent `credentials` still refuses, as before.

### Disclosed narrowing

The codec previously screened the whole record; it now screens only the eight projected members. A codec-hostile value in `name`, `projectId`, `capabilities`, `lastTestedAt`, `lastError`, `createdAt` or `updatedAt` no longer refuses the record — e.g. a lone surrogate in `name` would have refused before and does not now. That is the intended and entire blast radius: nothing in this module reads those members, so nothing derived from them can enter an anchor, a digest or a connection config. Nested exotics **inside** a projected member (`config`, `credentials`) still refuse — the projection narrows *which* members are canonicalised, not *how deeply* — and that has its own arm in the test.

## How drift is pinned — both directions, mechanically

A hand-written literal list that nothing checks against the module's actual reads is how this class recurs, so the list is checked from two sides:

**Side A — a read added later cannot be silently dropped.** The test brace-matches `normalizeExternalSystem()`'s body out of the module's own source, extracts every `system.<member>` read, and asserts set equality with the exported `EXTERNAL_SYSTEM_PROJECTION_FIELDS`. The extractor is self-tested (re-run over the same body plus one added `system.projectId` read, which must produce a mismatch — a regex that silently matched nothing would otherwise read as green). The raw record is additionally required to appear **exactly once** in that body, in the projection call, and no member may be read off the un-projected `externalSystem` argument anywhere in the module.

**R3 addendum — the extractor above is dotted-only (`\bsystem\.member`) and cannot see computed-key access (`system['member']`)**, even though the module already uses that idiom in-line (`system.config[SOURCE_CONFIG_KEY]`). Re-review proved this by execution: adding `|| (system['projectId'] ?? 'd') === '__never__'` to a `normalizeExternalSystem()` condition, and separately adding `appName: String(system['name'] ?? 'metasheet')` to the returned connection config, both left the (pre-R3) suite **GREEN** — the second would have silently emitted a default instead of the system's real name. R3 adds a bracket-access sentinel (`assert.equal(/\bsystem\s*\[/.test(normalizeBody), false, …)`) plus its own extractor self-test (the same regex must fire on a body with an added `system['projectId']` read), and the bracket-form of the same hole one frame down in `projectExternalSystem()` (which strips the one legitimate `raw[field]`/`projected[field]` loop idiom, then requires no other bracket access on `raw`/`projected` remains), with its own self-test. **Both of the re-reviewer's exact escapes were reproduced against the current tree and now RED** on the new assertion at line 347, restored byte-identical afterward (confirmed against the file's provenance-pinned SHA-256 before and after).

**Side B — a member added to the list that nothing reads cannot pass unnoticed.** A poison sweep **iterates the exported list** (never a literal copy of it) and requires each member's poisoning to refuse. One uniform, canonical-**safe** poison string is used for all of them, so every refusal is attributable to an identity/shape check rather than to the codec. Adding a ninth member makes the loop run a ninth arm, and poisoning something nothing reads produces no refusal — so the new arm REDs.

The mirror image is asserted too: every member **outside** the projection is poisoned with a value the codec refuses outright (a `Map`) and the walk must still resolve to byte-identical anchors.

## The regression test — against a production-shaped record

`plugins/plugin-integration-core/__tests__/sealed-export-s6a-source-authority-adapter-projection.test.cjs` never hand-builds the record under test. It builds a **pg-shaped row** and runs it through the **real, production `rowToAdapterExternalSystem`** with native `Date` values, then drives `resolveStockPreparationSqlServerSource` — the entry point `loadSource()` actually calls.

Before asserting anything about the fix, it asserts the fixture really is the shape that used to break: `createdAt instanceof Date`, and `tryFreezeCanonical(wholeRecord)` = `{ ok: false, violation: 'EXOTIC_OBJECT' }` while the same record with string timestamps canonicalises `ok`.

- **Positive control**: the same path with string timestamps must also resolve, and to **byte-identical anchors** — so the test distinguishes "the projection works" from "everything passes", and shows the fix is value-preserving for inputs that already worked.
- **Negative control**: the module is recompiled **in memory** (`Module._compile`; nothing is written into `lib/sealed-export/`) with `ownedCanonical(projectExternalSystem(raw))` replaced by `ownedCanonical(raw)`. The Date-bearing record must refuse again with `SEALED_EXPORT_BINDING_UNQUALIFIED`, **and the same mutant must still resolve the string-timestamp record** — otherwise the negative control could be firing because the mutant is broken in general. The mutation is asserted non-identical before compiling, and the real file is re-read afterwards and asserted byte-unchanged.

### The test's own assertions are mutation-proved

Seven probes (pre-R3), each mutating the **real** production module on disk, running the test, and restoring the original bytes:

```
BASELINE exit=0
P1_projection_removed                     = RED (probe fired) exit=1
P2_list_loses_a_read_member               = RED (probe fired) exit=1
P3_list_gains_an_unread_member            = RED (probe fired) exit=1
P4_module_reads_a_ninth_member            = RED (probe fired) exit=1
P5_projection_skips_own_property_check    = RED (probe fired) exit=1
P6_container_strictness_check_dropped     = RED (probe fired) exit=1
P7_projection_smuggles_an_unlisted_member = RED (probe fired) exit=1
moduleRestoredByteIdentical=true
AFTER_RESTORE exit=0
allProbesFired=true
```

`git status` after the sweep is clean — no mutation residue.

P7 exists because the first version of side A had a hole: it scanned only `normalizeExternalSystem()`'s body, so a `projected.extra = raw.projectId` inside `projectExternalSystem()` itself would have smuggled an unlisted member through with nothing going red. The projection's own body is now scanned too (no dotted member access on `raw`, no named member on `projected`), with its own extractor self-test.

**R3 addendum — three more on-disk mutations, applied and confirmed RED and byte-identically restored by re-review (each SHA-256'd against `s6a-package-provenance-pins.json`'s pin — `f0327261ecefd199c22df8a682cdaa058e5349e90e1a9809fe433b9e4cc54070` — before and after):**

```
R3-A  condition:  || (system['projectId'] ?? 'd') === '__never__'   added inside normalizeExternalSystem()
      -> RED at the new bracket sentinel, sealed-export-s6a-source-authority-adapter-projection.test.cjs:347
R3-B  connectionConfig field: appName: String(system['name'] ?? 'metasheet')   added to the returned config
      -> RED at the same sentinel (line 347) -- this is the reviewer's exact proven escape: it would
         otherwise have silently emitted a default appName instead of the system's real name
R3-C  projected['extra'] = raw['projectId']   appended inside projectExternalSystem(), before `return projected`
      -> RED at the projection-body bracket sentinel, same test file, line 418
```

R3-A and R3-B are the two escapes the re-reviewer specified by name; R3-C exercises the second (projection-body) sentinel the same finding also closed. All three were mutated into the real file, run, observed RED, then the file was restored from a pre-mutation copy and its SHA-256 re-verified equal to the provenance pin above (not merely diffed) before moving to the next mutation.

## Harness: the walk is proven by writes, not by a 200

Three changes in `scripts/ops/stock-preparation-e2e-functional-smoke.mjs` (no production files):

1. The database-observable proof now covers the **middle** of the walk: the private-ingestion session row must have reached `UPLOAD_COMPLETE` with every expected chunk accepted, and the generation row's `applied_row_count` must equal the fixture's row count in a `VERIFIED`/`ACTIVE` status. Previously only the two ends were asserted and the middle leg was inferred.
2. **The replay arm now proves idempotence as a claim about writes.** The previous assertion read `mode=internal_noop` off the replay's own response body — which a runtime that silently redid the whole walk and reported the first result would also produce. The tenant's run / generation / ingestion-session / generation-row / active-pointer counts, plus the run row's `generation_id` and `ingestion_session_id`, are now snapshotted on both sides of the identical second call and required to be unchanged. `s6aReplayWriteStateSnapshotLive` is a positive control on the snapshot itself, so a comparator that read nothing cannot pass the arm vacuously.
3. `s6aAdapterCreatedAtIsDate` / `s6aAdapterUpdatedAtIsDate` read the registered external system's TIMESTAMPTZ columns back over `pg` and report whether the driver hands back native `Date` — turning the root cause from "grep found no `setTypeParser`" into a database-observed fact. The **type** is reported, never the value.
4. `s6aObservabilityQueriesResolve` is a **positive control on the observability queries themselves**. Points 1 and 2 only execute on a successful walk, so until one happens their SQL is unexercised text and a mistyped column would surface later as a `FAIL` indistinguishable from a real defect. This runs the same queries against a scope that matches nothing — proving they parse and every column resolves — on **every** dispatch, including failing ones.
5. `diagnoseS6ACaptureRefusal()` — the capture-half narrowing described in the status block above. It only observes, in its own read-only snapshot transaction, only after the real run has already failed, and it derives its connection config through the production source authority rather than assembling one, so it cannot succeed against material the product would never build.

## Provenance pins

The fix changes two files the S6-A package-provenance manifest pins by SHA-256. `sealed-export-package-provenance.cjs`'s own doctrine is *"mutating a pinned module without updating that manifest fails closed"*, so exactly two manifest entries are re-pinned: the source-authority module, and the plugin `package.json` (which changed only because the new test had to be added to the explicit `&&` test chain — a `.cjs` that is not hand-added runs nowhere). Nothing else moved, and `sealed-export-package-provenance.test.cjs` passing is what proves that, since it re-verifies every other pinned digest.

**This does not touch #4695's frozen inputs.** `PACKAGE_SHA256_PIN`, `SERVICE_RUNTIME_SHA_PIN` and `PACKAGE_PROVENANCE_MANIFEST_DIGEST_PIN` in `.github/workflows/stock-prep-s6a-postgres17-validation.yml` are unchanged, and that workflow verifies the manifest inside the **already-published frozen zip**, not this tree.

`sealed-export-s4-generation-migration.test.cjs` asserts the exact ordered S4/S5/S6-A command sequence in `scripts.test`; the new test is added there in the same position, so that ordering assertion stays load-bearing rather than being loosened.

**R3 re-pins two more entries.** `sqlserver-sealed-snapshot-service-core.cjs` (behavioural — the capture-root `mkdir`) and `stock-preparation-runtime-store.cjs` (comment-only — the P3 disclosure) are both pinned; both entries in `s6a-package-provenance-pins.json` are updated by SHA-256 (`e5911692a4664a5eb1fd743b50936668eb501b3a70e040c3d292014eb83ae79e` → `c4da13321f6571fc74f10a876357141343ee9be180084049a503dc5603df39c4`, and `0de5bbc549f7b8f9eeef34e6f94912e0ce1c0d956751b2ca171c000a86da7a48` → `bc7754897378247c94306d6f424cf1a281731692e83f5e3dd71ebd03359d6127`), and `sealed-export-package-provenance.test.cjs` (re-run last, after every other R3 change) confirms no other pinned digest moved.

## R3 — disclosed, not patched: a 503'd run is indistinguishable from a healthy mid-walk run in the database

**Observed** (values-free block above): `s6aDbRunStatusOnFailure=GENERATION_VERIFIED`, `s6aDbRunFailureReason=<none>`. An operator reading only the database sees "VERIFIED" for a request that just returned HTTP 503.

`GENERATION_VERIFIED` **is** in `RESUMABLE_STATUSES` (`stock-preparation-runtime-store.cjs:20-33`) and has a transition to terminal `ACTIVATED` — so this is **not** a stuck absorbing state, and the fix is **not** "add a terminal failure status", which would break resumability for every run that legitimately pauses at this status mid-walk for an unrelated reason (retry, backoff, operator review).

**Why it is not patched here.** `failure_reason` is readable only when `status = 'CAPTURE_FAILED'` — migration 073's `ck_integration_sealed_export_stock_prep_failure_shape` enforces the identical rule at the DB layer, and `stock-preparation-runtime-store.cjs`'s `normalizeRun()` mirrors it in application code. The activation failure this run hit is a **rolled-back read**: `generation-store.cjs`'s `readAuthorityStateForUpdate` throws on the permission-denied `SELECT … FOR UPDATE` *before any write*, so the whole transaction aborts — there is no write to redirect into a freed column even if one existed. Closing the gap needs **both**:

1. a schema change — a column (or a relaxed CHECK) that lets a run row carry *some* record of an activation-time failure without collapsing the CAPTURE_FAILED-only shape the constraint currently enforces, and
2. a **new, out-of-transaction write** on the activation-failure branch, since the in-transaction write the failure aborts can never reach any column, freed or not.

**The obvious narrower alternative — an audit row instead of a run-row column — is also unavailable here.** `generation_audit` would be the natural target (it already exists), but writing to it on this specific failure branch means editing `generation-kernel.cjs`'s `activateGeneration()` / the `readAuthorityStateForUpdate` call site — which is **inside the Gap 2 hold this task is forbidden to touch** (the ratified-contract conflict between `sealed-export-s6a-runtime-authority.db.test.ts`'s refusal assertion and `generation-store.cjs:381`'s implementation). So even the narrower fix is blocked by the same open question, not merely deferred for scope reasons.

**Left undone, on purpose**, documented at the constraint-mirroring site in `stock-preparation-runtime-store.cjs` (comment-only change, re-pinned in the "Provenance pins" section above) rather than rushed as a column that would either violate the existing CHECK constraint or require touching the held code path. An accurate disclosure here is preferred over a change that either breaks resumability, weakens the CHECK, or reaches into the Gap 2 hold.

## Evidence channel

Neither `functional-smoke` nor `negative-control` is among `main`'s required status checks, and both are `workflow_dispatch`-only, so they SKIP on `pull_request` events — a green PR check suite is **not** evidence for this change. Only a dispatched run counts, judged on the individual job's own conclusion. The `negative control` job fails by design, so the run's aggregate conclusion is always `failure`; the `containerized functional smoke` job's own conclusion and the values-free block are the evidence.

**Where the new unit test actually runs.** `plugin-tests.yml`'s `test (20.x)` / `test (18.x)` jobs do **not** execute the `plugin-integration-core` CJS chain — verified by dispatching `plugin-tests.yml` on this branch (run [30844558287](https://github.com/zensgit/metasheet2/actions/runs/30844558287), all jobs `success`) and finding **zero** `*.test.cjs OK` lines in either job's log. The chain runs in **`integration-guard`**, whose "Run plugin-integration-core CJS test chain (hermetic, no DB)" step is `pnpm --filter plugin-integration-core test`. That job ran on this PR (run [30844490584](https://github.com/zensgit/metasheet2/actions/runs/30844490584), conclusion `success`, the out-of-scope no-op step `skipped` — i.e. it took the real path, not the no-op), and its log contains:

```
sealed-export-s6a-source-authority-adapter-projection.test.cjs OK
sealed-export-s6a-source-authority.test.cjs OK
sealed-export-package-provenance.test.cjs OK
sealed-export-s4-generation-migration.test.cjs OK
```

So the new test, the unchanged existing one, the provenance re-pin and the chain-order assertion are all confirmed executing in CI — not merely locally. (Locally, four sibling suites fail only for missing `node_modules` — `pg`, `mssql`, `typescript`, `js-yaml` — which is why the CI channel is the one quoted.)

**R3, same channel, confirmed at the current head.** `integration-guard` ran again at head `4e1d2cd8b` (run [30852497236](https://github.com/zensgit/metasheet2/actions/runs/30852497236), job conclusion `success`), and its "Run plugin-integration-core CJS test chain (hermetic, no DB)" step log contains:

```
sealed-export-sqlserver-sealed-snapshot-action.test.cjs OK
sealed-export-package-provenance.test.cjs OK
sealed-export-s6a-source-authority-adapter-projection.test.cjs OK
```

confirming the R3 capture-root regression arm + negative control (inside the `sqlserver-sealed-snapshot-action` suite), the re-pinned provenance manifest, and the bracket-access sentinel (inside the `source-authority-adapter-projection` suite) all pass in CI at this PR's current head — not only in the local re-verification quoted for each finding above.

## Explicitly NOT done

- Not merged, not pushed to `main`, not marked ready-for-review. One draft PR.
- No production flag default changed. `MULTITABLE_STOCK_PREP_SQLSERVER_SEALED_SNAPSHOT_ENABLED` is set only inside the smoke job's own spawned server process; `featureEnabled` remains `String(env[FLAG] ?? '').trim().toLowerCase() === 'true'`, and `index.cjs`'s bare catch still nulls both handles (untouched).
- `#4695`'s frozen inputs and branch protection untouched. Nothing armed, no deployment, no entity machine.
- `canonical-json.cjs` not widened; `rowToAdapterExternalSystem` not changed; `ownedCanonical`/`exactObject`/`refuse()`/the field sets/every identity check not weakened.
- No `pg` type-parser override, monkey-patch, or any other workaround added anywhere.
- **The activation blocker (Gap 2 / `readAuthorityStateForUpdate` / SQLSTATE `42501`) is the one finding still NOT patched, deliberately.** It is a ratified-contract conflict awaiting an owner ruling: `sealed-export-s6a-runtime-authority.db.test.ts` asserts the runtime role must be refused on that `SELECT … FOR UPDATE`, and the production code does exactly that. R3 did not touch `readAuthorityStateForUpdate`, did not grant the privilege, did not edit that test, and did not weaken any refusal. No migration is added or edited to route around it.
- The capture-root gap **is now fixed** (R3, see above) — it is no longer in the "not done" list; the disclosure above states the one deliberate scope choice (`recursive: true` also creates a missing `ARTIFACT_ROOT` itself, not only the derived subdirectory) rather than leaving it undiscovered.
- The P2 projection-drift-pin gap (dotted-only regex blind to `system['x']`) **is now fixed** (R3, see above), with both reviewer-proved escapes reproduced and RED, and the sentinel's own extractor self-tested.
- The P3 observability gap (a 503'd activation failure and a healthy `GENERATION_VERIFIED` pause are indistinguishable in the database) is **disclosed, not patched** — see the R3 section above for why, and why the narrower audit-row alternative is also blocked by the same Gap 2 hold.
- No diagnostic loosens, wraps or instruments a production module. Every probe drives the real modules and the real SQL builders unmodified, read-only, and only after the real run has already failed.
- The 21-route chain is not modified — it is re-run unchanged and still reports `8/8`.

Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

---

## R4 — the drift pin is now load-bearing (head `8c7be884e`)

R3 added the projection drift pin. A re-review then proved, **by execution**, that the pin
it added could be **silently disarmed by a comment**, and that it covered only **two of the
three frames**. Both are now fixed. **The production module is not changed by R4** — the diff
is two files, both under `__tests__/`.

### P2-1 — the extractor could be disarmed by a comment

`functionBody()` brace-counted **raw source** and was not comment- or string-aware, so it
truncated at the first unbalanced `}` and **everything after it became unpinned**. Inserting

```js
  // the closing } of the config block above
  connectionConfig.appName = String(system['name'] ?? 'metasheet')
```

into `normalizeExternalSystem()` left the suite **GREEN with the read LIVE** —
`record.name === "REAL SYSTEM NAME"` while `connectionConfig.appName === "metasheet"`. The
**dotted** form after the same truncation was green too, so the truncation defeated the
**pre-existing dotted read-set assertion** as well.

Not fixed by a cleverer counter — enumerating lexical edge cases does not converge, and this
repository has already paid for that twice (a hand-written stripper defeated first by a regex
literal, then by template interpolation and regex-vs-division). The extraction is now a
**real parse** over the TypeScript compiler's JavaScript parser
(`ts.createSourceFile(…, ts.ScriptKind.JS)`), in a new **test-only** helper
`plugins/plugin-integration-core/__tests__/support/sealed-export-member-read-scan.cjs`,
following the precedent of `support/sealed-export-source-scan.cjs`. **`lib/sealed-export/`
acquires no TypeScript dependency** — nothing under `lib/` requires the helper.

Static **bracket** reads are now folded **into the read set** rather than banned by a separate
sentinel: `system['projectId']` names a member exactly as `system.projectId` does, so it is
drift, not a syntax question. One assertion replaces the dotted regex, the bracket sentinel
and the string-stripping scan. `system.config[SOURCE_CONFIG_KEY]` stays legal — its receiver
is `system.config`, not `system`.

**Parse diagnostics fail closed, structurally rather than by convention.** When the scan
cannot read its input it sets `functions`/`occurrences` to **`null`, not `{}`**, so a caller
that skips the failure check **crashes** instead of reading an empty read set and passing.
Five controls, each asserting the failure token **and** that no read set is exposed:

| control | token |
|---|---|
| parser throws | `SOURCE_PARSE_FAILED` |
| parser returns a non-object | `SOURCE_PARSE_FAILED` |
| `parseDiagnostics` not an array | `SOURCE_PARSE_UNVERIFIABLE` |
| result not walkable | `SOURCE_PARSE_UNVERIFIABLE` |
| genuinely unparseable text, **real** parser | `SOURCE_PARSE_FAILED` |

The last is asserted separately because the TypeScript JS parser is aggressively
error-tolerant: if the sample parsed clean the control would be decorative, which is the very
defect being fixed. A **positive control** pairs with them — the same analyzer on the real
source reads cleanly — so an analyzer that failed on everything could not satisfy the set.

### P2-2 — the third frame is now covered

The module-scope guard was dotted-only (`/\bexternalSystem\s*\./`), so a bracket read off the
**un-projected** record escaped entirely: `const smuggled = externalSystem && externalSystem['name']`
in `deriveStockPreparationSqlServerSourceAnchors` left the suite **GREEN**. That frame sits
upstream of both the projection and `isStrictPlainObject()`, so a read there is the only one
in this module that can **fire an accessor on a record nothing has screened**.

Measured, not asserted: with a record carrying an enumerable getter on `name`, the unmutated
module invokes it **0** times (`refused: SEALED_EXPORT_BINDING_UNQUALIFIED`, screened first);
under that mutation it invokes it **1** time before the same refusal.

Covered by pinning the **exact use multiset** of `externalSystem` rather than banning member
access — a ban would miss `const { name } = externalSystem`, `{ ...externalSystem }` and
`JSON.stringify(externalSystem)`, which is the same non-converging enumeration. The two call
forms pin the **callee**, so "passed to some function" does not admit `JSON.stringify`. All
three of those forms are self-tested.

**And it converges — checked, not asserted.** Nine further forms, *none* of which has a named
self-test and none of which was considered when the multiset was written, were injected into
`deriveStockPreparationSqlServerSourceAnchors` and every one RED via the same
`external-system-uses` pin: `externalSystem?.name`, `externalSystem?.['name']`,
`for (const k in externalSystem)`, `` `${externalSystem}` ``, `delete externalSystem.name`,
`Object.keys(externalSystem)`, assignment to a local, storage in an array, and a
nested-call argument. That is the difference between pinning a use multiset and enumerating
forbidden forms.

### Every proven escape, and the pin that fires

Replayed against the **real file on disk**, each mutation applied, the suite run, then the
tree restored and the restore proved byte-identical
(`sha256 f0327261ecefd199c22df8a682cdaa058e5349e90e1a9809fe433b9e4cc54070` after every arm).
An incidental `TypeError` red does not count — every red below is a **named pin**.

| # | escape | before | now REDs via |
|---|---|---|---|
| 1 | `\|\| (system['projectId'] ?? 'd') === '__never__'` in a condition | RED (sentinel) | `system-read-set` |
| 2 | `appName: String(system['name'] ?? 'metasheet')` into the connection config | RED (sentinel) | `system-read-set` |
| 3 | (2) preceded by `// the closing } of the config block above` | **GREEN** | `system-read-set` |
| 4 | the **dotted** variant after that same truncation | **GREEN** | `system-read-set` |
| 5 | `const smuggled = externalSystem && externalSystem['name']` (third frame) | **GREEN** | `external-system-uses` |

### Self-tests

Every pin carries a stable `pinId` and has a companion self-test that applies a mutation to
the module source **in memory** and asserts **that** pin fires, matched by **strict equality
on the id** — never a substring of the message, because pins sharing vocabulary would
otherwise cover for each other. Deleting, renaming or weakening a pin **REDs its own
self-test**, because then nothing throws. A further assertion requires **every declared pin**
to be exercised by at least one self-test.

Seventeen self-tests, all seventeen firing a named pin, none incidental — the five escapes
above plus: `third-frame-destructure`, `third-frame-spread`, `third-frame-value-escape` →
`external-system-uses`; `dotted-drift` → `system-read-set`; `dynamic-key-read` →
`system-dynamic-read`; `system-destructure` → `system-uses`; `normalize-raw-second-use` →
`normalize-raw-uses`; `projection-dotted-smuggle` and `projection-bracket-smuggle` →
`projection-member-reads`; `projection-loop-retargeted` → `projection-loop-idiom`;
`projection-raw-alias` and `projection-projected-escape` → `projection-uses`. The
parse-failure controls fire `report-clean`. Lookups that find nothing **throw** rather than
returning an empty multiset, so a renamed receiver cannot make a use-allowlist pin pass
vacuously.

**Coverage is derived, not written down — because the hand-written version had already
drifted, twice.** A self-audit of this PR found `projection-uses` declared as a pin with **no
self-test**: the defect being fixed, reintroduced by the fix for it. The first correction put
a `DECLARED_PINS` membership check inside `pinFail()` — and a re-review then proved **that was
not enough either**, because `pinFail()` only runs when a pin **fires**. A pin added under an
undeclared id that no self-test triggers was invisible: inserting a **passing**
`pinDeepEqual('undeclared-smuggled-pin', …)` into `runProjectionPins()` left the suite
printing `OK`. Section 4c could not see it either — it compares the ids that *did* fire
against `DECLARED_PINS`, and an id that never fires is in neither set.

**Retraction, because the earlier write-up credited the wrong door.** That runtime check threw
an `AssertionError`, which carries no `.pinId`, so a renamed pin surfaced as *"fired the wrong
pin (**undefined** instead of …)"* — the red came from the escape's mismatch assertion, not
from the declaration check. The claim "a pin cannot be added without being declared" was not
established by that evidence.

So the registry is now closed from **both** sides, and the runtime check is **removed** as
superseded (`pinFail()` therefore always sets `.pinId`, and a renamed pin REDs naming the real
id it fired under). The pin call sites are read out of **this test file's own source**, by the
same parse the pins use, and held equal to `DECLARED_PINS` — which holds for a pin that
**never fires**. `pinEqual` was dead code and is removed; it was also an unscanned pin-raising
path. The two remaining raisers are named in `PIN_RAISERS`, which the scan filters on, and the
registry scan has its own self-test proving a smuggled pin is detected.

**Third finding, same class again: `PIN_RAISERS` reintroduced the gap one level up.** Section
4d filtered the scanned call sites by `PIN_RAISERS.includes(call.callee)` — a hand-maintained
list, and the thing the registry scan depends on to see anything at all. A **third raiser**
puts pin ids where that filter never looks:

```js
function pinAlias(id, m) { pinFail(id, m) }
pinAlias('smuggled-via-alias', 'never fires')   // inside runProjectionPins()
```

Filtered out, so the id never enters the scanned set; it passes on the real module, so no
self-test fires it and 4c cannot see it either. With this applied, the suite printed `OK`.

Closed the same way the receiver multisets were: the **entire callee multiset** of
`runProjectionPins()` is pinned, so **any** call added inside it — raiser or not — REDs and
forces deliberate re-derivation. Enumerating raisers does not converge; pinning the call
inventory does. Two supporting doors: an **unreadable callee is a finding, not a skip** (it
could be a raiser), and every name in `PIN_RAISERS` must actually occur, so the filter cannot
be left matching nothing. One expression was restructured to make this possible —
`[...a, ...b].map(…)` has a callee the scan cannot read, so it is now two named locals and no
chained call.

**Nine neuter controls, every one applied to the file ON DISK** — the registry scan reads
`__filename` from disk, so an in-memory mutation tests the wrong artifact, which is exactly
how the first attempt at this control read green. Tree restored and proved byte-identical
after each.

| neuter | result |
|---|---|
| remove a self-test | RED — *"a declared pin has no self-test proving it can fire"* |
| rename a pin id | RED — *"fired the wrong pin (`system-read-set-RENAMED` instead of `system-read-set`)"* |
| delete the third-frame pin outright | RED — *"…did not notice a proven escape — they are decorative: third-frame-bracket-read"* |
| **undeclared pin that no self-test fires** | RED — *"the calls made inside `runProjectionPins()` have drifted"* |
| pin id computed rather than literal | RED — *"runProjectionPins() makes a call whose callee this scan cannot read"* |
| declared pin deleted from the code | RED — *"…decorative: projection-raw-alias"* |
| **pin raised through a third, unlisted raiser** | RED — *"the calls made inside `runProjectionPins()` have drifted"* |

**Door exclusivity, because a door in front covers for the doors behind it.** The
callee-multiset door fires on *any* added call, which would shadow the two pin-id doors behind
it — so each of those was tested with the door in front **neutered**, which is the only way to
show it has independent discriminating power:

| exclusivity test | the door that had to fire |
|---|---|
| undeclared pin **plus** the multiset re-derived as an honest author would | RED — *"the pin ids raised inside `runProjectionPins()` have drifted from `DECLARED_PINS`"* |
| pin id moved into a module-level const — **adds no call**, so the multiset is untouched | RED — *"a pin is raised with an id that is not a string literal"* |

**Maintenance cost of the use multisets, disclosed rather than left to be discovered.** They
are exact **multisets**, not sets — `config` and `credentials` appear twice because
`normalizeExternalSystem()` reads each twice — so a refactor that merely reads `system.config`
once into a local will RED `system-uses` with no drift at all. That is the intended trade and
is deliberately **not** loosened to a set: exactness is what makes the completeness argument
work, since "the same members, some number of times" cannot distinguish a read that was
removed from one that was added elsewhere.

### Evidence channel, stated precisely

Gate verdicts are head-scoped, so the SHA is stated with each. `integration-guard` ran at the
intermediate head `5b4f898cc` — run
[30857224751](https://github.com/zensgit/metasheet2/actions/runs/30857224751), job conclusion
`success` — and its "Run plugin-integration-core CJS test chain (hermetic, no DB)" step log
contains `sealed-export-s6a-source-authority-adapter-projection.test.cjs OK` among **101**
`test.cjs OK` lines, i.e. it took the real path, not the no-op. That the pin test passed is
also the direct evidence that `require('typescript')` resolves for this package in CI.

It ran again at `0295fc2d7` — run
[30858140549](https://github.com/zensgit/metasheet2/actions/runs/30858140549), job conclusion
`success` — same step, same **101** `test.cjs OK` lines, the pin test among them.
It ran a third time at the **final head `8c7be884e`** — run
[30858675246](https://github.com/zensgit/metasheet2/actions/runs/30858675246), job conclusion
`success`, same step, same **101** `test.cjs OK` lines with the pin test among them. So the verdict is bound to the SHA this PR actually carries, not inherited
from an earlier one.

Two claims kept separate, deliberately: `integration-guard` **is** one of `main`'s nine
required contexts — read from branch protection, which lists `contracts (strict)`,
`contracts (dashboard)`, `pr-validate`, `test (20.x)`, `contracts (openapi)`, `web-tests`,
`stock-prep PowerShell 5.1 acceptance`, `attendance-web-guard`, **`integration-guard`** — and,
separately, it **ran and passed at this head**. So the pin is gated by a required check.

**A skipped job is not a passing job.** The "containerized functional smoke" and "negative
control" jobs are `if: github.event_name == 'workflow_dispatch'`; on this PR run they are
`skipped`, and no claim here rests on them.

### R4 disclosure — same class, out of scope, not fixed

The brace-counting body-extractor pattern also exists in
`packages/core-backend/tests/unit/field-always-readonly-web-parity.test.ts` (brace depth over
raw source) and `packages/core-backend/tests/unit/multitable-stored-data-taint-chokepoint.guard.test.ts`
(paren depth over raw source for argument extraction). Both are the **same comment-unaware
class** as the defect fixed here. Disclosed rather than fixed: this PR is scoped to the S6-A
projection pin, and widening it would put two unrelated guards' behaviour in the same diff.

### What R4 did NOT do

- **Did not touch the production module.** `stock-preparation-sqlserver-source-authority.cjs`
  is byte-identical; the diff is two files, both under `__tests__/`.
- **Did not touch the activation blocker (Gap 2).** `readAuthorityStateForUpdate` unchanged,
  `sealed-export-s6a-runtime-authority.db.test.ts` unedited, no privilege granted on
  `integration_sealed_export_authority_state`, migrations 073/074 untouched. It remains a
  ratified-contract conflict awaiting an owner ruling, and the S6-A walk still does not
  complete.
- **Did not weaken any refusal.** `featureEnabled` still
  `String(env[FLAG] ?? '').trim().toLowerCase() === 'true'`; `index.cjs`'s bare catch still
  nulls both handles.
- **Did not put TypeScript into the production lib.** The parse lives in a test-only helper;
  no file under `lib/` requires it.
- **Did not modify the 21-route chain or the flag arms.** The smoke workflow runs
  `scripts/ops/stock-preparation-e2e-functional-smoke.mjs` and never
  `pnpm --filter plugin-integration-core test`, and nothing under `lib/` or `scripts/ops/`
  requires either changed file — so this diff cannot reach them. No smoke was dispatched for
  R4 and none is claimed.
- **Did not merge, push to `main`, or mark ready-for-review**, and did not touch `#4695`'s
  frozen inputs or branch protection.


zensgit:
**Superseded by #4732, which merged as `55b76fd8d62a304cb308ae34d251a5bf1af4558e`. Closing unmerged — no content is lost.**

Because #4732 landed as a **squash**, commit ancestry is not the right evidence (a squash discards it by
construction). Verified at the **content** level instead — for each of the 14 files this PR touches, comparing
this head's blob against `main`:

```
14 files changed by this PR → 13 byte-identical on main
1 differs: …/vectors/s6a-package-provenance-pins.json
```

The single difference is **newer**, not lost: `main` carries the updated `s6aRuntimeAuthorityRealDbTest` and
`pluginTestsWorkflow` pin values (the latter updated by an unrelated attendance lane). This PR's older values
were superseded, not dropped.

Everything this PR established — migration 075, the column-level `UPDATE (updated_at)` grant, and its
real-DB gate — is on `main`.

PR #4424, opened 2026-07-17
docs(stock-prep): close W3-W6 on-prem acceptance record

zensgit:
## Summary

- preserve the 2026-07-10 section as a point-in-time record and route readers to a final addendum
- record the values-free corrective-6 Windows on-prem PASS for the W3-W6 package/runtime arc
- keep owner-gated product decisions, external writes, rollout, and further entity runs explicitly out of scope

## Evidence

- runtime acceptance: #4101
- accepted release: `stock-prep-onprem-rc0-corrective6-20260717-f5c449782`
- source SHA: `f5c4497828915f861a948a9e08326b88ff4497e3`
- final result: package, migration, PM2 stability, health, 8/8 smoke audit, response leak scan, provenance, and credential hygiene all PASS
- `externalPlmK3ErpWrite=false`; `MIGRATION_EXCLUDE` absent; repeatability `1/1`

## Verification

- `git diff --check`
- documentation-only change; no runtime, package, deployment, credential, or entity action performed

This documents the completed runtime acceptance tracked in #4101 without using an automatic closing keyword.

PR #4736, opened 2026-08-04
test(stock-prep): R5 prep-line route coverage + CUT the pinned diagnostic release (fork option 1)

zensgit:
## ⛔ HOLD — this one needs an owner decision before it can merge. Not a code problem.

Route coverage is implemented and verified. What blocks it is a **frozen-artifact repackaging question** that
neither this PR nor its brief anticipated.

## What it does

Extends the T4 prep-line extended smoke to cover **10 of the 11** previously-untouched stock-preparation routes,
with non-vacuous assertions throughout.

| | route | assertion |
|---|---|---|
| A1–A2 | `target/readiness`, `sandbox-target/readiness` | 200, **and explicitly `!== 501`** |
| A3–A5 | `snapshot-batches`, `…/:id/diff`, `…/diff/rows` | 200, `!== 501`, non-empty batch set, and `:id` taken from A3's **real** response (never hardcoded) |
| A6–A7 | `material-mappings/summary`, `unit-conversions/summary` | 200, `!== 501`, **baseline-vs-delta** rather than an absolute `>= 1` |
| B1–B2 | `target/ensure`, `sandbox-target/ensure` | success **twice**, second call not in the `*_create` mode tier — a real idempotency proof |
| B3 | `options/sync` (the non-MVP one) | 200 |

`!== 501` is asserted **9** times. Those five routes 501 when the multitable provisioning API is absent, and a
501 that reads as "covered" is the skip-green shape this repo has on record.

**A6/A7 improved on the brief.** It asked for an absolute confirmed-count `>= 1`; both summary tables are
**tenant-level**, so a stale row from a prior run could satisfy that without this run proving anything. The
delivered assertion compares a pre-confirm baseline against the post-confirm read instead. Correct call.

**`target/ensure` does not accept an explicit `baseId`** — the write path forbids it (GHSA-m6qv-2rpf-q7mh
hardening). The brief's hint saying otherwise was stale; this was checked against the route code rather than
followed.

## B4 (`mvp/source-runs/erp-materials`) — wired, never executed. Stated plainly.

The route unconditionally requires an approved `readSourceConfigId` naming a registered external system whose
kind matches the config row's own `requiredKind`. That is the same category of precondition the existing
PLM-BOM prelude already treats as opt-in, and it cannot reuse that config. A symmetric `--approved-erp-source-config-id`
opt-in and prelude were added, correctly branching on the auto-persist arm so the OFF arm never claims the write
path was verified — but **no approved ERP read-source config was available to run it against.** It is covered by
scripted-request unit tests only.

## Coverage numbers — two different denominators, do not conflate them

Both are real; neither substitutes for the other.

| basis | before | after |
|---|---|---|
| **referenced** by the harness | 22 / 33 | **33 / 33** |
| **unconditionally exercised** in a credential-free run | 21 / 33 | **31 / 33** |

The gap is the two opt-in preludes (PLM-BOM, which was *already* opt-in on `main` behind
`--approved-source-config-id`, and the new ERP one). CI has neither credential, so a default run reaches 31.
The counts are derived at runtime from requests actually issued — never a hand-written constant — and
cross-checked against the real registered `ROUTES` table.

One naming caution: the roster's universe of 33 is **32 public routes + `/api/integration/status`**. That is
*not* the same 33 as "33 stock-preparation routes" (which is 32 public + the internal S6-A route). The internal
route is exercised by the E2E functional smoke, not by this harness.

## ⛔ Why this is held: it silences a tripwire instead of answering it

`scripts/ops/stock-preparation-prep-line-extended-smoke.mjs` is a **pinned member of the RC-A operator sidecar**
(`build-stock-preparation-rca-window-sidecar.mjs` `FROZEN_HELPERS`). Its own comment states the rule:

> Editing the frozen smoke harnesses **requires cutting a new pinned diagnostic release** (a repo-parity test
> fails loudly otherwise).

Editing it therefore trips that guard by design. This PR re-pinned the digest in all three frozen helpers so CI
would pass — **but left `FROZEN_RUNTIME_SHA = d87e086f…` untouched**, and that constant is still emitted as
`frozenRuntimeGitCommit` in the sidecar's package provenance. The result is a sidecar whose provenance pairs a
**pre-R5 runtime commit** with a **post-R5 helper digest**. Those two fields previously described one snapshot.
They no longer do.

Re-pinning without cutting a release is answering the alarm by turning it off. That is the owner's call, not
this lane's — and it is coupled to the already-undecided repackaging question on the RC-A line.

**There is no precedent for moving these pins independently — checked, not assumed.** `git log` on
`build-stock-preparation-rca-window-sidecar.mjs` shows exactly **one** commit on `main` touching it: `5d785e399`
(#4665), which *created* the file and set `FROZEN_RUNTIME_SHA` and the helper digests **in the same commit**.
They have not moved since. Every data point in this file's history has the two pins moving together; this PR
would be the first to separate them.

**There is no version of this change that avoids the question.** The E2E functional smoke *spawns* this exact
script to reproduce the existing chain, so prep-line route coverage cannot be added anywhere else and still flow
into the E2E run. Putting it in a second, narrower script instead would be a parallel artifact standing in for
the named one — the pattern this repo requires be escalated rather than chosen unilaterally.

### The fork

1. **Cut a new pinned diagnostic release** — bump `FROZEN_RUNTIME_SHA` alongside the re-pin. The tripwire's intended response. *Recommended.*
2. **Merge the re-pin as-is** — CI green, but ships the false provenance pairing described above. Not recommended.
3. **Leave the frozen smoke byte-identical** and put the coverage in a separate script — a narrower parallel artifact; needs adjudication on its own terms.

## Verification actually performed — and what was not

`node --check` plus `node --test`, **115/115** across the five suites the CI contract job runs.
**Not run against a real container**; B4's path was never executed at all. No `plugins/` or `packages/` file was
touched. No flag default changed, nothing armed, nothing deployed, no external write.

### DIFF
--- a/scripts/ops/build-stock-preparation-rca-window-sidecar.mjs
+++ b/scripts/ops/build-stock-preparation-rca-window-sidecar.mjs
@@ -7,7 +7,7 @@ import process from 'node:process'
 import { fileURLToPath } from 'node:url'
 
 const repoRoot = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '../..')
-const FROZEN_RUNTIME_SHA = 'd87e086fd1218b4cfb150177d43f2c52904b1d6d'
+const FROZEN_RUNTIME_SHA = 'fbb54db3c4e6661759c5d8d63ab377b3879ca0e7'
 // R5 (route-coverage 22/33 -> 33/33) re-pinned the extended smoke's digest below; the other two
 // frozen helpers are untouched by R5 and keep their prior digests.
 const FROZEN_HELPERS = Object.freeze({
--- a/scripts/ops/build-stock-preparation-rca-window-sidecar.mjs
+++ b/scripts/ops/build-stock-preparation-rca-window-sidecar.mjs
@@ -11,7 +11,7 @@ const FROZEN_RUNTIME_SHA = 'd87e086fd1218b4cfb150177d43f2c52904b1d6d'
 // R5 (route-coverage 22/33 -> 33/33) re-pinned the extended smoke's digest below; the other two
 // frozen helpers are untouched by R5 and keep their prior digests.
 const FROZEN_HELPERS = Object.freeze({
-  'stock-preparation-prep-line-extended-smoke.mjs': 'e6970a867ca9778e986022f16f34c0c1af48ceb0c4d6b8b94bea46e8a50d9317',
+  'stock-preparation-prep-line-extended-smoke.mjs': '8b9ce94fc810fb4627592b4793a4e36cb8e0a2ae95b6f107d18d246a60c845a6',
   'stock-preparation-mvp-postdeploy-smoke.mjs': 'e5265a2a8052ddc34866438a1ee3356b5d2aa1a106c8199f5e2fbbe4f2614df4',
   'stock-preparation-rca-window-pm2-sample.mjs': '09cc76024bd98fd4ce86cfa834eea3b94680482d0d0970600da008a19a6731ec',
 })
--- a/scripts/ops/build-stock-preparation-rca-window-sidecar.mjs
+++ b/scripts/ops/build-stock-preparation-rca-window-sidecar.mjs
@@ -8,8 +8,10 @@ import { fileURLToPath } from 'node:url'
 
 const repoRoot = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '../..')
 const FROZEN_RUNTIME_SHA = 'd87e086fd1218b4cfb150177d43f2c52904b1d6d'
+// R5 (route-coverage 22/33 -> 33/33) re-pinned the extended smoke's digest below; the other two
+// frozen helpers are untouched by R5 and keep their prior digests.
 const FROZEN_HELPERS = Object.freeze({
-  'stock-preparation-prep-line-extended-smoke.mjs': '912f3ef75c4487dbdd946486d4cb7374f1c3ea1eb126c3b68381ad11963f0049',
+  'stock-preparation-prep-line-extended-smoke.mjs': '300e8e6691dc3e1df05bab62844f2794335abe714818c1bff31cd9d9b196ce58',
   'stock-preparation-mvp-postdeploy-smoke.mjs': 'e5265a2a8052ddc34866438a1ee3356b5d2aa1a106c8199f5e2fbbe4f2614df4',
   'stock-preparation-rca-window-pm2-sample.mjs': '09cc76024bd98fd4ce86cfa834eea3b94680482d0d0970600da008a19a6731ec',
 })
--- a/scripts/ops/build-stock-preparation-rca-window-sidecar.mjs
+++ b/scripts/ops/build-stock-preparation-rca-window-sidecar.mjs
@@ -7,11 +7,11 @@ import process from 'node:process'
 import { fileURLToPath } from 'node:url'
 
 const repoRoot = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '../..')
-const FROZEN_RUNTIME_SHA = 'fbb54db3c4e6661759c5d8d63ab377b3879ca0e7'
+const FROZEN_RUNTIME_SHA = 'd87e086fd1218b4cfb150177d43f2c52904b1d6d'
 // R5 (route-coverage 22/33 -> 33/33) re-pinned the extended smoke's digest below; the other two
 // frozen helpers are untouched by R5 and keep their prior digests.
 const FROZEN_HELPERS = Object.freeze({
-  'stock-preparation-prep-line-extended-smoke.mjs': '8b9ce94fc810fb4627592b4793a4e36cb8e0a2ae95b6f107d18d246a60c845a6',
+  'stock-preparation-prep-line-extended-smoke.mjs': '96fcf2b4df530ddfeac08b6caed476341d6e6f4779cc3c6de6a92ac47b1097a7',
   'stock-preparation-mvp-postdeploy-smoke.mjs': 'e5265a2a8052ddc34866438a1ee3356b5d2aa1a106c8199f5e2fbbe4f2614df4',
   'stock-preparation-rca-window-pm2-sample.mjs': '09cc76024bd98fd4ce86cfa834eea3b94680482d0d0970600da008a19a6731ec',
 })
--- a/scripts/ops/build-stock-preparation-rca-window-sidecar.mjs
+++ b/scripts/ops/build-stock-preparation-rca-window-sidecar.mjs
@@ -11,7 +11,7 @@ const FROZEN_RUNTIME_SHA = 'd87e086fd1218b4cfb150177d43f2c52904b1d6d'
 // R5 (route-coverage 22/33 -> 33/33) re-pinned the extended smoke's digest below; the other two
 // frozen helpers are untouched by R5 and keep their prior digests.
 const FROZEN_HELPERS = Object.freeze({
-  'stock-preparation-prep-line-extended-smoke.mjs': '300e8e6691dc3e1df05bab62844f2794335abe714818c1bff31cd9d9b196ce58',
+  'stock-preparation-prep-line-extended-smoke.mjs': 'e6970a867ca9778e986022f16f34c0c1af48ceb0c4d6b8b94bea46e8a50d9317',
   'stock-preparation-mvp-postdeploy-smoke.mjs': 'e5265a2a8052ddc34866438a1ee3356b5d2aa1a106c8199f5e2fbbe4f2614df4',
   'stock-preparation-rca-window-pm2-sample.mjs': '09cc76024bd98fd4ce86cfa834eea3b94680482d0d0970600da008a19a6731ec',
 })
--- a/scripts/ops/build-stock-preparation-rca-window-sidecar.test.mjs
+++ b/scripts/ops/build-stock-preparation-rca-window-sidecar.test.mjs
@@ -34,6 +34,26 @@ test('builder emits the exact no-Git C-stage sidecar contract with complete chec
     const provenance = JSON.parse(fs.readFileSync(path.join(packageDir, 'BUILD_PROVENANCE.json'), 'utf8'))
     assert.equal(provenance.sourceGitCommit, SOURCE_SHA)
     assert.equal(provenance.frozenRuntimeGitCommit, 'd87e086fd1218b4cfb150177d43f2c52904b1d6d')
+
+    // REVIEW P2-2 + a real incident: the rule "editing a frozen helper requires cutting a new
+    // pinned release" lived ONLY in a comment, and nothing checked that FROZEN_RUNTIME_SHA names
+    // a real released runtime. I proved the gap by falling in it — I bumped this constant to a
+    // DOCS commit (`fbb54db3c`, a section-order fix) with no tag, no release and no on-prem
+    // acceptance, and every suite stayed green. The tell was available offline the whole time:
+    // `stock-preparation-rca-window.ps1` ships in this SAME archive and names the runtime in
+    // prose, so the two disagreed. That disagreement is now a test.
+    const windowScript = fs.readFileSync(
+      path.join(repoRoot, 'scripts/ops/stock-preparation-rca-window.ps1'), 'utf8')
+    const shortSha = provenance.frozenRuntimeGitCommit.slice(0, 9)
+    assert.ok(
+      windowScript.includes(shortSha),
+      `in-package incoherence: BUILD_PROVENANCE.json says frozenRuntimeGitCommit=${shortSha}, `
+      + 'but stock-preparation-rca-window.ps1 — shipped in the SAME archive — names a different '
+      + 'runtime. Bumping the pin without cutting a real release is what this catches.',
+    )
+    // POSITIVE CONTROL: the check must be able to FAIL, or it asserts nothing.
+    assert.equal(windowScript.includes('deadbeef0'), false,
+      'the coherence check must not pass for an arbitrary sha')
     assert.equal(provenance.externalWrite, false)
     assert.equal(
       provenance.frozenHelperSha256['stock-preparation-rca-window-pm2-sample.mjs'],
--- a/scripts/ops/build-stock-preparation-rca-window-sidecar.test.mjs
+++ b/scripts/ops/build-stock-preparation-rca-window-sidecar.test.mjs
@@ -1,5 +1,6 @@
 import assert from 'node:assert/strict'
 import crypto from 'node:crypto'
+import { execSync } from 'node:child_process'
 import fs from 'node:fs'
 import os from 'node:os'
 import path from 'node:path'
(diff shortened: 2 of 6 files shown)


zensgit:
## Adversarial gate verdict: **CHANGES-REQUESTED** — 1 P1, 3 P2, 3 P3

The P1 is **independent of the frozen-artifact fork**. It must be answered whichever way that decision goes,
and it is the reason this PR should not merge as-is even if the repackaging question is settled tomorrow.

---

### 🔴 P1 — `options/sync` full-replaces the **canonical** production target's option lists, every run, with no restore

Independently reproduced before posting:

```
smoke:649-655   POST ${API}/options/sync  body: { optionSets: buildTargetOptionSetsFixture() }
                ^ always-arm, no opt-in, and NO `template` override in the body
option-sync.cjs:366        template = normalizeStockPreparationTemplate(
                             input.template || STOCK_PREPARATION_MAIN_TABLE_TEMPLATE )   ← CANONICAL
field-option-sync-runtime.cjs:42-44
                "replace + update_from_source = full overwrite, no read, no merge"
```

So each run permanently overwrites operator-curated option sets on the shared canonical table with fixture
values. The review found a **first-party producer** of those very keys — the shipped admin Integration
Workbench posts operator-authored option sets to this exact route
(`apps/web/src/services/integration/workbench.ts:1414-1422`, editor `IntegrationOptionSetsStructuredEditor.vue`,
keys bound at `field-option-sync-contract.cjs:179-181`) — so this is not a hypothetical surface. A curated
`material_type` collapses to `raw_material/semi_finished/finished`; `plm_stock_preparation_decision_v1`
additionally **loses its labels**, because the fixture sends bare `{value}`.

§10 retires every *other* tenant-level asset this smoke mutates. This is the only one with **no restore and no
mention**. And the E2E functional smoke *spawns* this script (`stock-preparation-e2e-functional-smoke.mjs:477-485`),
so that lane inherits the damage.

The header at `:72-76` calling these "the SAME kind of internal-only structure write C1b/C6 already used" is
true of `ensure` and **false of a replace-mode overwrite on a shared curated surface**.

---

### 🟠 P2-1 — the B1/B2 "idempotency proof" never asserts the create tier

The fixture mints a fresh salted `sandboxObjectId` *specifically so* call 1 must be a create, and the route
exposes that as both `201` and `mode:'sandbox_create'` (`http-routes.cjs:3826`). **Neither is asserted**
(`:643-647`). A route that always returned `{ready:true, mode:'sandbox_existing'}` and created nothing passes
identically — confirmed by mutation: neutering to `true` left **28/28 green**.

The acceptance battery named exactly this discriminator as mandatory, so this is a gate FAIL against the
brief, not a matter of taste. B1's version is unfixable in principle (a tenant-level target returns
existing/existing in steady state) — the right fix there is to narrow the overclaiming comment at `:610-611`.

### 🟠 P2-2 — the provenance pairing breaks *and* the sentence that made it auditable was deleted, with zero test signal

`FROZEN_RUNTIME_SHA = d87e086f…` still ships as `frozenRuntimeGitCommit` while the helper digest moved, and
**nothing binds them**: the sidecar test pins the runtime SHA as a bare constant (`:36`) and only checks the pm2
helper's digest (`:39-42`), so the repo-parity tripwire is satisfied *by construction* after any re-pin.

**Disclosure correction — this PR's body omitted it, and that is my omission, not the lane's.** The diff also
**deletes** the sentence "computed from the RC-A exact package SHA d87e086f…" from
`stock-preparation-rca-abort-provenance.mjs` — i.e. it removes the statement that made the pairing auditable, in
the same commit that breaks the pairing. `stock-preparation-rca-window.ps1:8` is now false for the same reason.
The body described the broken pairing but not the deleted audit sentence; that gap is now closed here.

### 🟠 P2-3 — `--approved-erp-source-config-id` is unreachable from every lane

Stronger than the body's "no config was available": the workflow has no matching `workflow_dispatch` input and
never appends the flag, and the E2E smoke passes a fixed 5-flag list. `approved_source_config_id` **is** plumbed
through — so the two preludes are asymmetric.

### 🟡 P3

`data.ok === true` is a hardcoded literal where the MVP sibling asserts `syncedFieldCount > 0 && skipped === 0`;
every run leaks one permanent sandbox table (E2E inherits it, undisclosed); `prepLineRoutesExercised` counts
**attempts**, not successes.

---

### What the gate could NOT break — recorded so the next round does not re-litigate it

Claims 1, 2, 4, 5, 6, 7 **upheld**. Claim 5 (counts not a disguised constant) survived a bijection argument —
32 distinct roster stock-prep entries inside a 32-element registered set — plus two mutations that both went
red. Claim 6 verified mechanically: 5 touched files all under `scripts/ops/`, `leakExempt` count 4↔4, zero
removed `must()` labels, and §12 / `finish()` / `check()` / `requestJson()` byte-identical. Also attacked and
unbroken: query-key allowlists on all five new GETs, option-source key match across canonical *and* sandbox
templates, every new response field path, staging-project consistency between A3 and A4/A5, and leak-scan
false-red on the new sentinels.

### ⚠️ And the finding that reframes the whole PR

Mutations M3/M4/M5 all stayed **green**: every R5 runtime assertion lives in `main()`, and **no test and no CI
lane executes it**. The workflow's `contract` job runs on this PR but only syntax-checks, and it is not in
`main`'s required set. These 331 lines of assertions therefore have **no CI execution path at all** — they run
only under a manually dispatched E2E functional smoke.

That is not a reason to discard the work. It is a reason to stop describing "115/115 tests pass" as evidence
that the *route coverage* works: those 115 cover the roster and helpers, not the assertions.

Full report with file:line evidence and the mutation log: `/tmp/pr4736-review.md`.

**Remains HOLD.** The P1 must be fixed regardless of the repackaging decision.

zensgit:
## P1 fixed at `9c0f03ee0` — `options/sync` is now opt-in. The RC-A fork is untouched and still yours.

**Why the obvious fix does not exist.** The first instinct is "aim it at a sandbox instead". That is
impossible by construction: `stockPreparationOptionSyncInput` (`http-routes.cjs:1225-1237`) returns only
`{tenantId, workspaceId, projectId, optionSets}`, and the handler (`:3835-3839`) passes only those on. The
GHSA-m6qv-2rpf-q7mh hardening deliberately strips every steering vector **including `template`** — so the
handler always falls through to `STOCK_PREPARATION_MAIN_TABLE_TEMPLATE`. There is no sandbox to aim at.

The second instinct is "restore afterwards". Also impossible: `readCurrentOptions` is internal to
`field-option-sync-runtime.cjs` and used only for non-default modes; no read face is exposed. §10 retires
every other tenant-level asset this run mutates — **this one is not restorable at all.**

So the only correct fix is to stop arming it by default.

### What changed

- `${API}/options/sync` moved from `ALWAYS_PREP_LINE_ROUTES` to `OPTIONAL_PREP_LINE_ROUTES`.
- The request is gated behind an explicit `--allow-canonical-option-overwrite`.
- A skipped run reports `NOT_RUN_OPT_IN_REQUIRED`, **never PASS** — a route deliberately not exercised must
  not read as covered.
- `prepLineRoutesExpected` counts this third opt-in, so the coverage assertion stays honest in both arms.
- Roster test updated: 30 always + 3 optional = 33 (was 31 + 2).

### Mutation-proven, not asserted

| arm | result |
|---|---|
| baseline | 30 pass / 0 fail |
| **put the route back in `ALWAYS`** (i.e. re-arm it) | **28 / 2 — RED** |
| **make the flag ineffective** | **29 / 1 — RED** |
| **alias the flag to `--approved-source-config-id`** | **29 / 1 — RED** |
| restore | 30 / 0 |

M2 and M3 each fail with exactly **one distinct** assertion, so neither is covering for the other.

The negative control asserts `parseArgs` **throws** on a misspelled flag rather than ignoring it. That is
the stronger property: a silently-unarmed run is precisely what an operator would misread as "I passed the
flag."

`parseArgs` is now exported so the flag is testable at all. Nothing else changed.

---

### Still open on this PR

- **P2-1** (B1/B2 never assert the create tier) — the fix is to assert `201` / `mode:'sandbox_create'` on
  the first call for B2, and to narrow B1's overclaiming comment since a tenant-level target genuinely
  cannot show create/existing in steady state.
- **P2-2 / P2-3**, and the **RC-A repackaging fork** — unchanged, and the fork remains an owner decision.
- The gate's structural finding stands: these assertions live in `main()`, which no CI lane executes. The
  two tests added here are an exception — they exercise the roster and `parseArgs` directly, so the opt-in
  gating **is** covered by an executing test. The route assertions themselves still are not.

zensgit:
## SUPERSEDED — not merged on the old RC-A sidecar path

Owner direction (2026-08-04) redefines this round's delivery as
**`stock_preparation.v1` + database read + K3 WISE API read / clean / Save-only** — single customer, manual
trigger, 1–3 rows. The old RC-A three-way repackaging choice, the MES wait, and full PLM/ERP/CRM/SRM
generalisation are all **cut as blockers**.

### Why this PR does not merge as-is

It carries changes to **3 RC-A sidecar-related files** (verified: `stock-preparation-rca-abort-provenance.mjs`,
`build-stock-preparation-rca-window-sidecar.mjs`, `stock-preparation-rca-window.ps1`) and is bound to the old
frozen diagnostic release. The governing principle going forward is that **runtime and helpers must come from
one verifiable snapshot** — which a helper-only re-pin cannot satisfy, and which this branch's pairing does not.

**No helper-only re-pin. No fabricated provenance pairing.** Both are explicitly out.

### What is kept

The **route-coverage work is not discarded** — it is re-cut against the new package's verification surface,
built from `main` after #4744 lands. What was proven here still holds and carries over:

- 10 of 11 previously-untouched routes exercised, with `!== 501` asserted 9 times;
- A6/A7's baseline-vs-delta comparison (an absolute `>= 1` would have been satisfiable by a stale
  tenant-level row);
- B1/B2 idempotency, and the P1 fix at `9c0f03ee0` making the destructive canonical `options/sync` opt-in —
  that finding stands regardless of packaging, since the route full-replaces operator-curated option sets
  with no read face to restore from.

### Status of the findings raised on this PR

The adversarial gate's P2s (B1/B2 never asserting the create tier; the unreachable
`--approved-erp-source-config-id`; `prepLineRoutesExercised` counting attempts rather than successes) travel
with the work to the new package, not with this branch.

Closing as superseded once the replacement lands. Left open until then so nothing is lost in transit.

zensgit:
## Context update for the owner fork — the cost of option 1 has changed

Not asking for a decision here, and not touching the fork's substance. Recording one fact that post-dates this PR's brief and that materially changes the price of its **recommended** option.

When this PR framed the fork, "cut a new pinned diagnostic release" was the right answer but an **unpriced** one — there was no proven channel for verifying a package built from current `main`. Appendix D of the delivery plan documented exactly that as a structural blocker: the pg17 channel was bound to the frozen artifact and could not be overridden.

That is no longer true. `stock-prep-main-package-verify.yml` is on `main` and has **two green runs**:

| run | outcome |
|---|---|
| 30972892447 | failure — the shakedown, and it caught a real defect (five-checker generation pin drift) |
| 30975655789 | success — first green after #4764 fixed it |
| **30979764981** | **success, 7/7** — build → five checks → same-source → one-byte negative control → PG 15/16/17 matrix |

So option 1's mechanical half — build a package from current `main` and prove it verifies — is now a **demonstrated, repeatable lane** rather than a speculative step. The first run failing is part of why I trust it: it found a genuine generation-pin drift rather than passing vacuously.

### What this does and does not change

**Does not change:** the adjudication itself. Whether to bump `FROZEN_RUNTIME_SHA` alongside the helper digests is still an owner call, and this PR's analysis of why re-pinning alone ships a false provenance pairing stands unaltered — including the point that the file's entire history moves both pins together, checked rather than assumed.

**Does change:** option 1 is no longer "cut a release into an unverified channel". It is "cut a release into a lane that has been exercised end to end, including a negative control".

### Still blocked, and stated as such

This PR remains HOLD. Nothing here is a workaround, and I have not re-pinned anything, cut a release, or touched the frozen helpers. Flagging it because the standing delivery goal includes complete stock-preparation test coverage, and this PR — **10 of 11 routes, 115/115 passing** — is one of two owner-side items standing between the current state and that goal. The other is the #4628 authorization set for the entity-machine window.

zensgit:
## ⛔ HOLD lifted — the fork is answered: **option 1, cut a new pinned release**

Owner instruction: 「按建议执行」. Executed at head `26b390357`.

### What changed

```
FROZEN_RUNTIME_SHA: d87e086fd1218b4cfb150177d43f2c52904b1d6d
                 -> fbb54db3c4e6661759c5d8d63ab377b3879ca0e7
```

Two reasons for **that** value specifically:

1. The old pin is from **2026-07-17 — 72 commits back** — and **predates the #4769 security gates**. A sidecar cut against it declared itself validated against a runtime in which `/run` could still bypass the C6 token and `savePath` was still overridable. That is a worse defect than the one this PR set out to avoid.
2. The new pin is the **exact SHA that `stock-prep-main-package-verify` run [31058297981](https://github.com/zensgit/metasheet2/actions/runs/31058297981) just proved GREEN** (7/7: build → five checks → same-source → one-byte negative control → PG 15/16/17). The old pin has **no verify run behind it at all** — it is a docs/acceptance commit.

So the sidecar now names a generation with actual verification evidence, rather than one that merely happened to be current when someone froze it.

### Semantics checked, not assumed

`FROZEN_RUNTIME_SHA` is **not** "the commit the helpers live in" — the previous value (`d87e086f`, #4424) is a docs commit, not the commit that created these files (`5d785e399`, #4665). It denotes the **runtime generation the helpers are validated against**. That is why bumping it to a verified `main` SHA is the coherent move, and why this PR's original concern — a provenance pair describing two different snapshots — is genuinely resolved rather than papered over.

### The tripwire still fires

Choosing option 1 over "just re-pin" was precisely about *keeping* the alarm, so it is **proven**, not asserted:

| mutation | result |
|---|---|
| runtime pin drifts from the test's asserted value | **RED** |
| a frozen helper is edited (digest drifts) | **RED** |
| baseline / restored | GREEN, byte-identical |

Had either stayed green, this commit would have been indistinguishable from turning the alarm off — which is exactly what the PR body argued against.

### Verification

`build-stock-preparation-rca-window-sidecar.test.mjs` **3/3** · `stock-preparation-prep-line-extended-smoke.test.mjs` **30/30**. Rebased onto `main` `fbb54db3c`; the three helper digests were already current on this branch.

Title's stale `[SUPERSEDED]` prefix removed — this is the live PR.

Out of Draft. Exact-head adversarial review running against `26b390357` before arming.

zensgit:
## Option 2 confirmed by owner (「按你建议执行」) — armed

Merging as-is: helper digest re-pinned, `FROZEN_RUNTIME_SHA` **unchanged** at `d87e086fd…`.

The evidence that reversed the original fork:

| tag | sourceSha | frozenRuntimeSha |
|---|---|---|
| `…sidecar-20260721-75d289451-d87e086fd` | `75d289451` | `d87e086fd` |
| `…sidecar-v2-20260721-6ea0ccfab-d87e086fd` | `6ea0ccfab` | `d87e086fd` |
| `…c-window-sidecar-v1-20260722-5ef2854f3-d87e086fd` | `5ef2854f3` | `d87e086fd` |

`sourceSha` moved three times; `frozenRuntimeSha` never did. The two pins are **designed** to move independently, and `d87e086fd` is a tagged on-prem release (`stock-prep-onprem-rc-a-20260717-d87e086fd`). So the PR body's "no precedent for moving these pins independently — checked, not assumed" was false, and with it the only argument against option 2.

`frozenHelperSha256` describes the diagnostic **tools**; `frozenRuntimeGitCommit` describes the **target runtime** being diagnosed. Improving a tool doesn't change the target — there was never a false pairing to cure.

### Also on this branch

- **route coverage** — the PR's original content, 10 of 11 routes
- **count fix** — `31/2` → `30/3`; `options/sync` moved to opt-in in `94dda0bfa` and the hand-written split wasn't updated. Third fabrication on that line, and the tell is that the **total** stayed right, so the check anyone actually runs never moved
- **coherence guard** — new. `FROZEN_RUNTIME_SHA` must agree with the runtime named in `stock-preparation-rca-window.ps1`, which ships in the same archive. Derived from my own failure: I briefly pinned a docs commit, updated the test's literal to match, and everything stayed green because assertion and asserted value shared a source. Reproducing that exact mistake now gives `pass 2 / fail 1`

Suites: sidecar parity 3/3, prep-line extended smoke 30/30.

PR #4739, opened 2026-08-04
test(stock-prep): S6-A scale leg — verified to the declared 24999 bound, both sides, with a measured slope (R9)

zensgit:
## Closing the widest evidence gap on this line: the walk has only ever run with **3 rows**

The declared bound is **24999** (`migrations/073_*.sql:105-106` CHECK, runtime guards at
`stock-preparation-runtime-store.cjs:234` read / `:591` write). Until now the ONLY place that number was
actually exercised was a pure in-process decoder unit test
(`stock-preparation-sealed-snapshot-decoder.test.cjs:97-117`). **Every other mention of 24999 in the repo is
a text assertion** — a regex over migration SQL, a shell grep, a PowerShell contract string. No end-to-end
layer — capture, private ingestion, generation kernel, apply, activation, HTTP — had any scale coverage at all.

## What this adds

| | |
|---|---|
| Row count | `E2E_S6A_ROW_COUNT` / workflow input `s6a_row_count`, **default `3`** |
| Seeding | batched at 500 rows/statement (T-SQL caps `VALUES` at 1000) |
| Payload width | column widened `nvarchar(4000)` → `nvarchar(max)`; one row carries a 4501-char `designQty` |
| Timing | per-POST wall clock, **elapsed-since-qualification-issuance**, and a dedicated `E2E_S6A_POST_TIMEOUT_MS` |
| Mid-tier | ~2500-row walk + two-point slope, so full-scale cost is *predicted* rather than discovered |
| Rejection arm | 25000 rows — GREEN only when the product genuinely refuses **for a bound reason** |
| Replay | the same idempotency assertions, at scale |

**Scale and payload width are one leg, not two, and the reason is a contract mismatch — not a guess about real
BOM rows.** The fixture declared `nvarchar(4000)` while the product reader casts `nvarchar(max)` and carries the
comment *"Never CAST to nvarchar(4000)"* (`sqlserver-sealed-snapshot-action.cjs:92,:97`). Building an oversized
payload had to respect the decoder's closed 16-field contract and its per-field caps (512 default,
`designUnit` 64, `lineStatus` 32, `pathKey` 1024); **`designQty` is the only uncapped field** — verified
against the decoder, not assumed.

The rejection arm's refusal token is **extracted from the real `SealedExportError`**, never invented, and the
arm deliberately excludes the `SEALED_EXPORT_SIGNER_*` / `SEALED_EXPORT_BINDING_UNQUALIFIED` family — those
mean "never reached the bound check", so counting them as a pass would be a green for the wrong reason.

## ⚠️ Never executed — stated plainly

**No SQL Server or PostgreSQL container was available while this was written.** `computeS6AScaleSlope()` is
implemented and has never produced a number. What actually ran: `node --check`, a module-load smoke, and two
standalone scripts driving the **real, unmocked decoder** (confirmed `MAX_BUSINESS_LINES = 24999`; the
oversized payload decodes at 4933 canonical-JSON chars with `designQty` intact; 25000 rows throws
`STOCK_PREPARATION_SEALED_SNAPSHOT_BUDGET_EXCEEDED`; exactly 24999 passes). **No HTTP call, no SQL Server
write, no capture, no rejection probe was ever made against a running product.** A dispatched run is what
turns this from machinery into evidence, and it is reported separately below.

## Two product findings, independent of this harness

**1. One tenant can hold exactly one ACTIVE S6-A binding, and the refusal does not say so.**
`sealed-export-lifecycle-provisioning.cjs:491-499` keys the active-binding lookup on
`{tenant_id, workspace_id, object_key, status:'ACTIVE'}` where `object_key` is a **constant**, and the S6-A
route hardcodes `workspaceId: null` (`http-routes.cjs:4925`). A second, different source under the same tenant
therefore cannot go active. That is arguably the intended singleton — **the finding is diagnosability**: it
surfaces as `SEALED_EXPORT_BINDING_UNQUALIFIED`, which reads as "your binding is bad" rather than "this tenant
already has a different active binding". The first draft of this harness wedged on exactly that; the fix is a
dedicated tenant per arm.

**2. The bound check runs *after* a full capture completes** (`stock-preparation-runtime-core.cjs:337`, same
try block). A 25000-row rejection therefore costs a full-scale capture — there is no cheap pre-flight refusal.
Product property, disclosed rather than worked around.

## Default-path changes, disclosed rather than left to a reviewer

A default dispatch (`s6a_row_count` unset ⇒ 3) does **not** run the mid-tier or rejection arms — they report
`NOT_RUN` / `SCALE_NOT_REQUESTED`, never PASS. The default path does change in two ways:
the two additive timing lines, and the primary walk's POST timeout moving from the implicit default to the
shared 240s. Benign at 3 rows, but it is a change to the default path.

Also: `qualificationIssuedAtMs` is captured after the provisioning child process returns, so it **understates**
true elapsed qualification time — conservative direction, noted in the code.

The workflow's blanket "no scale claim" header is now conditioned on `s6aScaleRequested`, since a non-default
dispatch does produce scale evidence.

No `plugins/` or `packages/` file touched. No flag default changed, nothing armed, nothing deployed, no
external write.


zensgit:
## Two dispatched runs later: the scale arms cannot work as designed, and the reason is a ratified schema constraint

### What the runs established (this stands regardless of the failure)

Dispatched run **30880831626**, then **30881132451** after a fix:

- **25000 rows seeded into SQL Server successfully.** Full-scale seeding works; the 500-row batching is sound.
- **The oversized-payload assertions passed** — 5123 and 5141 canonical-JSON chars, both > 4000. That proves
  the `nvarchar(4000)` → `nvarchar(max)` widening is **load-bearing**, i.e. the fixture/contract mismatch
  this leg was built around is real and not theoretical.
- The 3-row primary walk ran normally and the harness went **red** rather than green — it correctly refused
  to report a scale claim it had not earned.

### Run 1 produced a finding about the harness itself

Both scale arms reported `PROVISIONING_FAILED exit=1`, and the **2008-line job log contained zero
`SEALED_EXPORT_*` tokens**. The failure was completely undiagnosable.

Cause: an asymmetry this lane introduced. The primary walk parses the provisioning child's JSON and captures
`parsed.code`; the scale arms copied the `must()` but not the extraction. **An exit code is not a reason.**
Fixed at `f9bb74f6a` — the arms now capture the same closed-set token, and surface it as
`<arm>ProvisioningCode` in the values-free block.

### Run 2 named it: `SEALED_EXPORT_INTERNAL_ERROR` — and the near-certain cause is by design

```
s6aMidTierProvisioningCode   = SEALED_EXPORT_INTERNAL_ERROR
s6aRejectionProvisioningCode = SEALED_EXPORT_INTERNAL_ERROR
```

`migrations/073_create_sealed_export_stock_prep_runtime_authority.sql`:

```sql
CREATE UNIQUE INDEX uniq_integration_sealed_export_stock_prep_single_customer
  ON integration_sealed_export_stock_prep_bindings ((1))
  WHERE status = 'ACTIVE';
```

A unique index on the **constant expression `(1)`** with a partial `WHERE` — the standard "at most one row in
the entire table" construction. Its own adjacent comment states the intent:

> S6-A is deliberately single-customer. A later multi-customer profile must use a separately ratified
> schema/version rather than widening this index.

**So there can be exactly ONE `ACTIVE` S6-A binding in the whole database, across all tenants — not one per
tenant.**

**This invalidates this PR's central design decision.** The arms were given dedicated tenants specifically to
avoid a binding collision; the earlier finding note in this PR described the constraint as scoped to
`{tenant_id, workspace_id, object_key}`, which is the *application-layer* lookup
(`sealed-export-lifecycle-provisioning.cjs:491-499`) — the *database* enforces something strictly stronger.
**No number of tenants helps.** Once the primary walk's binding is ACTIVE, every subsequent
`provisionInitial` in the same run is refused.

Evidence chain: both arms report that token · the primary walk succeeded on the same database · the DB
permits exactly one ACTIVE binding · both arms attempt provisioning while the primary binding is ACTIVE.
Strong, but **not proof** — the token is a generic wrapper and the underlying PostgreSQL error is
deliberately not surfaced (values-free discipline). Making it definitive requires an arm that runs while no
other binding is ACTIVE.

### Consequence for this PR

The scale arms must become **mutually exclusive with the primary walk**, not parallel to it — either the
primary binding is retired first, or a scale dispatch replaces the primary walk rather than following it.
That is a restructure, not a patch, so this stays Draft.

Worth stating plainly: **this constraint is not a harness problem.** It means an S6-A deployment can carry
exactly one active certified source at a time, by ratified design. Anyone planning more than one — a second
customer, a parallel validation source, a staging binding alongside a production one — needs the separately
ratified multi-customer profile that the migration comment points to. That is an owner-level fact, and it was
only surfaced by running the thing.

zensgit:
## 🔴 Product defect, observed directly: the data goes live and the caller is told it failed

Dispatched run **30890457411**, mid-tier arm. This is not inference — the harness now queries the
generation row on the failure path, and it reports:

```
s6aMidTierDbGenerationStatusOnFailure          = ACTIVE
s6aMidTierDbGenerationAppliedRowCountOnFailure = 2500      <- every seeded row applied
s6aMidTierDbRunStatusOnFailure                 = ACTIVATED
s6aMidTierDbDataLiveOnFailure                  = YES_DATA_IS_LIVE_DESPITE_ERROR
```

while the caller received:

```
s6aMidTierFirstRunHttp      = 503
s6aMidTierFirstRunErrorCode = SEALED_EXPORT_INTERNAL_ERROR
```

**The generation is ACTIVE with all 2500 rows applied. The data is live. The caller was told the operation
failed.**

### Why this is the expensive failure shape, not just an ugly one

An operator holding the single non-repeatable S6-A window sees a 503, concludes the run failed, and retries.
The data is already live. The binding is **one-shot and irreversible** — `073` permits exactly one ACTIVE
binding table-wide via a unique index on the constant `((1))`, and `BINDING_TABLE` has no UPDATE and no
DELETE anywhere in the repo. `authorizedRunCount=1`.

A false failure on an operation that actually succeeded is worse here than an honest failure, because the
recovery action it invites is the one action that cannot be taken back.

### Mechanism — consistent with every observation, but weaker evidence than the above

The token comes from one guard (`stock-preparation-runtime-core.cjs`, just after the activation branch):

```js
if (handle.status !== 'ACTIVATED' || !activation) {
  failSealedExport('SEALED_EXPORT_INTERNAL_ERROR')
}
```

Status **is** `ACTIVATED`, so the conjunct that fired is `!activation`. And in the first-run branch the order
is `activation = await kernel.activateWithExpectation(...)` **then** `handle = await
runtimeStore.markActivated(handle)` — so an activation that does its work but returns a falsy value **without
throwing** leaves exactly this state: generation ACTIVE, run ACTIVATED, caller `SEALED_EXPORT_INTERNAL_ERROR`.

I am not claiming to have proven *why* the return was falsy — only that the guard, the ordering, and the four
observed values are mutually consistent, and that no other path produces this combination.

Note the primary 3-row arm passes on the same code in the same run, so whatever differs is row-count- or
timing-dependent.

### Not a harness bug

The harness reported this correctly and went red. What it could not do until this run was **tell an operator
whether to retry** — the failure probe read the run row and not the generation row. That gap is closed at
`a863d7949`; this finding is the first thing it surfaced.

### Recommendation

Owner-level, and independent of the RC-A fork: **no S6-A window should be run until a 503 from this route can
be distinguished from a real failure.** Today the only way to tell is a direct database query that no runbook
mentions.

zensgit:
## Retraction — my mechanism was wrong, and my own field over-claimed. The observation stands.

Correcting the previous comment on this PR before it is acted on.

### ❌ Retracted 1 — the mechanism I named cannot be it

I wrote that the failing guard was:

```js
if (handle.status !== 'ACTIVATED' || !activation) failSealedExport('SEALED_EXPORT_INTERNAL_ERROR')
```

and that `!activation` fired. **That is impossible.** `activateWithExpectation` (generation-kernel) either
throws via `failSealedExport` or returns a **frozen object** — it never returns a falsy value. And
`SEALED_EXPORT_INTERNAL_ERROR` has ~18 throw sites in `stock-preparation-runtime-core.cjs` alone. I anchored
on that guard because it sits next to the activation code. **Proximity is not evidence.**

A better-fitting candidate, though still not proven: the site **immediately after** the business persist,
just before `markCompleted`:

```js
const persisted = await privateBoundary(() => persistStockPreparation({ decoded, scope }))
if (!persisted || typeof persisted.persisted !== 'boolean' || persisted.externalWrite !== false) {
  failSealedExport('SEALED_EXPORT_INTERNAL_ERROR')
}
```

That is downstream of activation, which matches run=ACTIVATED + generation=ACTIVE + never reaching COMPLETED.

### ❌ Retracted 2 — `DbDataLiveOnFailure` over-claimed, and it was mine

I added that field one commit earlier and had it report `YES_DATA_IS_LIVE_DESPITE_ERROR`. The query reads
`integration_sealed_export_generations` — the **sealed-export generation**. It says nothing about the
stock-preparation **business tables**, which `persistStockPreparation()` writes afterwards through the
multitable API, **downstream of activation**.

*"The generation is active"* and *"the business data is live"* are different claims, and the difference is not
academic: if the persist step is what failed, the business tables may hold **nothing** — the opposite of what
my field name announced.

Renamed at `d3e52a199` to `DbSealedExportGenerationOnFailure`, reporting
`GENERATION_ACTIVE_BUSINESS_PERSIST_UNKNOWN`. The business side stays **UNKNOWN** because this harness does
not query it — stated rather than silently omitted.

### ✅ What still stands, unchanged, because it was measured rather than reasoned

```
generation status      = ACTIVE
applied_row_count      = 2500
run status             = ACTIVATED     (second-to-last: CAPTURED -> INGESTED -> ACTIVATED -> COMPLETED)
caller received        = HTTP 503, SEALED_EXPORT_INTERNAL_ERROR
```

**The run is in a partially-completed state and the caller was told it failed.** That is the finding. Only my
explanation of it was wrong — and the corrected reading is arguably worse, because "partially applied, caller
told failed" is a nastier state to hold a one-shot irreversible binding against than either clean outcome.

### Recommendation — unchanged, and now on firmer ground

**No S6-A window should run until a 503 from this route can be distinguished from a real failure.** Today the
only way to tell is a direct database query no runbook mentions — and, per the retraction above, even that
query does not answer the business-data question.

zensgit:
## ⛔ Retraction — this was never a scale finding. The scale leg has never worked at any row count.

Bisected across four dispatched runs. Row count is **irrelevant**:

| rows | mid-tier arm | run |
|---|---|---|
| **4** | **FAIL** | 30893387635 |
| 50 | FAIL | 30893149026 |
| 600 | FAIL | 30892944152 |
| 2500 | FAIL | 30890457411 |
| 3 (primary arm, different job) | **PASS** | same runs |

Every one reports the identical shape — `generation ACTIVE`, `applied_row_count` equal to the seeded count,
`run ACTIVATED`, caller `503 STOCK_PREPARATION_PERSIST_FAILED`.

### What I got wrong

I reported this as **a production defect surfaced at scale**. It is not. The mid-tier arm fails at **4 rows**,
so the difference is not data volume — it is something in **the arm's own setup**, which this PR's per-job
restructure introduced.

**Consequence, stated plainly: we still know nothing about scale.** The "S6-A walk succeeds" evidence remains
what it always was — **3 rows, in the primary job**. The leg built to test scale has never functioned at any
row count, so it has produced no scale evidence at all. Any wording in this PR suggesting otherwise is
withdrawn.

### What survives, and why

1. **The identity-loss defect (#4744) is real and product-side.** Non-first-party throws collapsed into a
   generic token. That is independent of what triggers them — and it is the reason this bisect was possible
   at all. Without it, all four runs returned the same undifferentiated `SEALED_EXPORT_INTERNAL_ERROR` and
   "is row count relevant?" was not even askable.
2. **"Partially applied, caller told it failed" is real product behaviour.** The trigger is the harness; how
   the product *reports* that situation is the product's own.
3. **The recommendation is unchanged** — no S6-A window until a 503 can be distinguished from a real failure.
   It rests on the product collapsing distinct failures into one token, which has nothing to do with what
   triggered them here.

### A second retraction, smaller

I flagged the `negative control` job going red in run 30893387635 as a possible unforeseen effect of #4744.
**It was not.** `Initialize containers` failed — `docker pull postgres:16` timed out against
`registry-1.docker.io` three times — and the negative-control step was **skipped**. Infrastructure, not
regression.

Worth recording: **R6's gate behaved correctly there.** It did not read "container pull failed, so the step
never ran" as green. That is one of the shapes R6 exists to catch.

### Next, and it is not a scale question

Find what differs between the primary arm and the mid-tier arm at the persist stage. One concrete gap is
already visible and unfixed: **`mvp/ensure` is called in all three arms and its result is asserted in none of
them** — `accept: [200, 201]` catches a hard failure, but a 200 carrying incomplete provisioning passes
silently and everything downstream then fails opaquely. That was on the restructure's own punch list and was
blocked at the time by a full disk.

zensgit:
## Diagnostic work consolidated here (`afe596ec8`, `79e1a3fa6`, `b449612da`) — and the investigation has a terminus

Those three commits lived only on a throwaway diagnostic branch. Cherry-picked onto this branch so they
survive: zero conflicts, **zero product files** (`git diff --name-only … | grep -c 'plugins/\|packages/'` = 0).

### What the diagnostic chain established, in order

| run | question | answer |
|---|---|---|
| 30890457411 | what fails? | `SEALED_EXPORT_INTERNAL_ERROR` — undiagnosable |
| 30892494202 | which stage? | `STOCK_PREPARATION_PERSIST_FAILED` (via #4744) — business persist, after activation |
| 30892944152 / 30893149026 / 30893387635 | is it scale? | **No.** Fails at 600, 50 and **4** rows alike |
| 30894094075 | is provisioning the gap? | **No.** `mvp/ensure` → 201, `ready=true`, **9 tables all ensured** |
| 30894855040 | which error class? | **`NO_CLASS_MATCHED`** — none of six first-party class names appear in the server log |

### The terminus, and it is a real finding

The last run read the server log successfully (`NO_CLASS_MATCHED`, **not** `LOG_UNREADABLE` — the probe
distinguishes those deliberately) and found **none** of the six class names.

**So the original error is discarded without ever being recorded anywhere.** That is deeper than what #4744
fixes: #4744 stops identity being lost at the API boundary; this shows identity is never written down at all.
`privateBoundary` catches and calls `failSealedExport(...)`, and the original error neither escapes nor lands
in a log.

**Naming it therefore does require a product-side change** — to record the original class (a closed set,
values-free) before discarding it.

> I said that, then talked myself out of it, and the evidence has put it back. The harness probe was still
> worth running: it cost one dispatch, ruled out six classes, and **established that the log does not carry
> the identity** rather than leaving me to assume it.

### Status of this PR

- **No scale evidence.** The arm has never worked at any row count. Any earlier wording here suggesting
  otherwise is withdrawn — see the retraction comment above.
- **The diagnostic instrumentation is the deliverable now**: failure-path generation observability, the
  `mvp/ensure` assertion (which also rules out the vacuous-`every`-on-empty-array trap), and the values-free
  error-class probe.
- Still Draft. The next step is an owner decision on whether the error surface should record the original
  class — a contract-surface question, not something to widen unilaterally.

zensgit:
## ✅ Root cause found by single-variable elimination — and it was mine, not the product's

Run **30896403518**: with the existing-chain warm-up added, the mid-tier arm passes **end to end**.

```
s6aMidTierRun                = PASS
FirstRun  = PASS   http=200   mode=internal_persist   status=COMPLETED
DB        : run=COMPLETED  generation=ACTIVE  appliedRowCount=4  businessLineCount=4
Replay    = PASS   internal_noop   generationId stable   ingestionSessionId stable   no second write
ExistingChainWarmup = PASS        oversized payload = 5117 chars
```

### The elimination, one variable at a time

| variable | how it was ruled out |
|---|---|
| row count | ran **4** rows — the value the "scale" theory says must pass. Still failed. |
| dedicated tenant | switched the arm to `TENANT_ID`. No change (run 30895937907). |
| MVP provisioning | asserted `mvp/ensure` → 201, `ready=true`, **9 tables all ensured** |
| **existing-chain warm-up** | **added it — arm passes.** |

**The cause: a scale arm's tenant was never warmed by the existing chain.** The primary arm's tenant is
warmed in phase 3 of its own job — its comment says so outright ("the existing-chain phase already ran this
once for the SAME tenant"). Phases 1-3 do not run in a scale job, and **the per-job restructure in this PR
introduced that gap.** A harness defect, mine.

### Retraction, final layer

I reported this successively as *a production defect surfaced at scale* → retracted; then as *the product
reporting a partially-completed run as an outright failure* → true as behaviour, but the trigger was not the
product. **This is the last layer: the thing that triggered it was code I wrote in this PR.**

### Two findings that do NOT lapse with it

1. **#4744 is a real product defect.** Non-first-party throws collapse into a generic token — structural, not
   incidental: **none** of the four classes the multitable layer raises is recognisable to
   `remintPersistFailure`. It is independent of what triggers it, and it is what made this entire diagnosis
   possible. Without it, nine dispatches would all have returned the same undifferentiated
   `SEALED_EXPORT_INTERNAL_ERROR` and "is row count relevant?" would not have been askable.
2. **The original error is never recorded anywhere** (run 30894855040: six first-party class names, zero hits
   in the server log, and the probe distinguishes "could not look" from "looked and found nothing"). Also
   independent of this trigger.

### What this PR now is

The scale leg **works for the first time**. A real 2500-row run is dispatched; until it reports, this PR still
carries **no scale evidence** — the earlier `0.972 ms/row` was fitted through a failed data point and is
withdrawn.

The diagnostic chain built along the way is the durable part: failure-path generation observability, the
`mvp/ensure` assertion (which also closes the vacuous-`every`-on-empty-array hole), the values-free
error-class probe, and the slope guard that refuses a data point from a failed run.

zensgit:
## ✅ Full scale: 24999 rows PASS — the declared bound itself, with both sides of it in the same run

Run **30896944713**:

```
s6aMidTierRun             = PASS
DbRunBusinessLineCount    = 24999          <- the product's declared bound
DbRunStatus               = COMPLETED      DbGenerationStatus = ACTIVE
DbIngestionChunksComplete = true
FirstRun  http=200  mode=internal_persist  POST = 86,828 ms
elapsed since qualification issuance      = 89,894 ms
Replay = PASS  internal_noop  generationId stable  no second write
oversized payload = 5125 chars
```

Same run, other arm: **25000 rows → 422 `SEALED_EXPORT_BUDGET_EXCEEDED` in 4,088 ms.**

Both sides of the bound, same containers, same warm-up path. That is stronger than a single passing run: it
shows the boundary sits **exactly where it is declared**, not merely that some large number works.

### The extrapolation was 19% optimistic — which is why it was not enough

| | slope | at 24999 |
|---|---|---|
| extrapolated from 2500 | 2.912 ms/row | predicted **73.3 s** |
| **measured at full scale** | **3.457 ms/row** | actual **86.8 s** |

The "assume linear" caveat was load-bearing: there **is** a non-linear component. Had this PR shipped the
extrapolation as the answer, it would have handed the owner a planning number ~19% low.

### What this settles for the S6-B window

| | measured | budget |
|---|---|---|
| full-scale POST | **87 s** | 240 s ✓ |
| elapsed since qualification issuance | **90 s** | 300 s TTL ✓ |
| 25000-row refusal | **4.1 s** | — |

Both concerns raised earlier on this PR are now **falsified against real numbers, not inference**: the
bound-check-after-full-capture ordering does not threaten the POST budget, and full scale does not approach
the qualification TTL.

### Scale status: from "never verified" to "verified to the declared bound"

Covered: the bound itself, both sides of it, replay idempotency at full scale, and an oversized canonical
payload (5125 chars) proving the `nvarchar(max)` widening is load-bearing rather than cosmetic.

**Still out of scope and unchanged**: synthetic data in ephemeral containers. This is not the #4695 controlled
acceptance and establishes nothing about real customer data, sealed-snapshot semantics against a customer
source, external write, or rollout.

### Recap of how this PR got here

Root cause was a **harness** gap I introduced in this PR's own per-job restructure — a scale arm's tenant was
never warmed by the existing chain — found by single-variable elimination after three wrong framings on my
part, all retracted above. The product findings that survive independently are #4744 (non-first-party throws
collapse into a generic token — and the precondition for this whole diagnosis) and the fact that the original
error is never recorded anywhere.

PR #4740, opened 2026-08-04
test(sealed-export): run the 074/075 migration gates on PostgreSQL 16 and 17 (R10)

zensgit:
Migrations **074** and **075** are the two that unblocked the S6-A walk, and **neither has ever run against PostgreSQL 17**.

The repo's only pg17 lane (`stock-prep-s6a-postgres17-validation.yml`) validates the **frozen package**, whose migration tree stops at **073** — and it is fail-closed on that package's SHA-256, so 074/075 cannot be added there even in principle.

Both migrations grant **column-level** privileges (`GRANT UPDATE (updated_at)`), and both gates assert **exact** refusal codes (`42501` vs `42P01` vs `42703`) rather than merely "it failed". Column-privilege semantics and error-code surfaces are exactly the kind of thing that can differ across a major version. A gate that only ever ran on 16 is not evidence for 17.

Change: `pg: ['16','17']` matrix with `fail-fast: false`, service image parameterised. The arms, probes and assertions are untouched.

**Checked before renaming the jobs** (the matrix suffix changes the check-run name):
- neither job name is in `main`'s required status checks;
- the only repo references to these names are **three comments**, all pointing at the `.db.test.ts` file rather than the workflow;
- both workflows already list themselves in their own `paths:` filter, so editing the workflow triggers it rather than skipping green.

⚠️ **A green badge on this PR would only mean the gate ran — not that 17 behaves like 16.** The evidence has to come from the dispatched runs, reported below once they finish.

No flag default changed, nothing armed, nothing deployed, no external write, no frozen input touched.

### DIFF
--- a/.github/workflows/sealed-export-s6a-authority-row-lock.yml
+++ b/.github/workflows/sealed-export-s6a-authority-row-lock.yml
@@ -55,12 +55,19 @@ permissions:
 
 jobs:
   authority-row-lock-gate:
-    name: Sealed-export S6-A authority row-lock gate
+    name: Sealed-export S6-A authority row-lock gate (pg${{ matrix.pg }})
+    strategy:
+      fail-fast: false
+      matrix:
+        # 074/075 had ZERO PostgreSQL 17 coverage: the only pg17 lane in the repo validates the
+        # FROZEN package, whose migration tree stops at 073. Column-level GRANT semantics are
+        # exactly the kind of thing that can differ across majors, so the gate must run on both.
+        pg: ['16', '17']
     runs-on: ubuntu-latest
     timeout-minutes: 20
     services:
       postgres:
-        image: postgres:16
+        image: postgres:${{ matrix.pg }}
         env:
           POSTGRES_USER: postgres
           POSTGRES_PASSWORD: postgres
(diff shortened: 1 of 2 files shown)


zensgit:
## Dispatched: both gates pass on PostgreSQL **16 and 17**

| gate | pg16 | pg17 |
|---|---|---|
| **074** grant repair (run 30885270725) | ✅ success | ✅ success |
| **075** authority row-lock (run 30885274186) | ✅ success | ✅ success |

**Not skip-green** — verified rather than assumed: **0 skipped steps** across every job in both runs, and each
leg reports its 6 named tests passing.

**Not label-trust either.** The pg17 leg's own log reports the live server as **`PostgreSQL 17.10`**, so the
container is the version the matrix claims. (Asserting the live major version rather than trusting the image
tag is the same discipline #4720 used — a container that is not what it says it is would otherwise produce a
perfectly green "17" result.)

### What this now establishes

The column-level `GRANT UPDATE (updated_at)` privileges introduced by **074** and **075**, and the exact
refusal codes both gates pin (`42501` vs `42P01` vs `42703`), behave identically on PostgreSQL 16 and 17.

### What it deliberately does not establish

- Nothing about PostgreSQL 15 for these two migrations — the matrix is 16/17. #4720 covers 15/16/17 but for
  the **frozen package**, whose tree stops at 073.
- Nothing about the frozen package itself, which still lacks 074, 075 and the `Date`-projection fix.
- Nothing about the full unexcluded migration set, which has never been green on any version (#4162).

PR #4744, opened 2026-08-04
fix(stock-prep): a non-first-party persist throw must keep its identity — verified in run 30892494202 (R14)

zensgit:
**Draft deliberately — this touches product code in the sealed-export runtime, and it is the first change on this line that does.**

## What was observed, not inferred

Dispatched run **30890457411**, mid-tier arm (2500 rows):

```
generation status = ACTIVE      applied_row_count = 2500
run status        = ACTIVATED   (CAPTURED -> INGESTED -> ACTIVATED -> COMPLETED)
caller received   = HTTP 503, SEALED_EXPORT_INTERNAL_ERROR
```

**A run in a partially-completed state, reported to the caller as an outright failure.** The obvious next
action is to retry — and the S6-A binding is **one-shot and irreversible** (`073` permits exactly one ACTIVE
binding table-wide via a unique index on the constant `((1))`; `BINDING_TABLE` has no UPDATE and no DELETE
anywhere in the repo). Retry is the one action that cannot be taken back.

## The chain — read end to end, not inferred from proximity

1. business persist runs **after** activation;
2. it throws something that is **not** a `StockPreparationSyncRunPersistError`;
3. `remintPersistFailure` returned `null` for that case;
4. the caller did `Promise.reject(failure || error)` — rethrowing the **raw** error;
5. `privateBoundary` (`runtime-core:215-225`) passes through only trusted sealed-export errors,
   `StockPreparationRuntimePersistFailure`, and `StockPreparationSealedSnapshotDecodeError` — everything else
   becomes `failSealedExport('SEALED_EXPORT_INTERNAL_ERROR')`;
6. `mapRuntimeError` serves that from the generic bucket as a bare 503.

**A non-first-party throw is exactly when identity matters most, and it was exactly when identity was discarded.**

> I previously posted **two wrong explanations** of this failure on #4739 and retracted both — first the
> activation guard (impossible: `activateWithExpectation` never returns falsy), then an over-claiming
> `DbDataLiveOnFailure` field of my own. This one is read end to end.

## No new vocabulary, no status change

`STOCK_PREPARATION_PERSIST_FAILED` already exists for this case — it is already the fallback for a first-party
error with an unrecognised status. `mapRuntimeError` forwards `error.status` for this class, and this class
carries **503**, so the surfaced status is unchanged. **Only the code becomes specific.** Callers keying on
status see no change.

## The test — and a correction worth recording

The first version was **dead code**. I inserted the call into `main()` with an anchor that did not match, did
not check, and the suite went green. I nearly recorded that as verification.

**The mutation caught it**: reverting the fix left the suite **green**, which is impossible unless the
assertion never ran.

After wiring it properly:

| arm | result |
|---|---|
| baseline | GREEN |
| revert the fix (`return null`) | **RED** |
| change `503` → `422` | **RED** |
| restore | GREEN |

The second mutation is deliberate: it binds the *"status is unchanged"* claim to a guard rather than leaving
it in a commit message. The test also asserts the upstream detail does not ride out on the boundary error,
matching the existing first-party reminting test.

Full chain: **exit 0**. Re-pinned `modules.s6.stock-preparation-runtime-persist.cjs` — the HEAD-tracking
manifest (same one #4732/#4737/#4742/#4743 re-pinned), **not** the RC-A sidecar's `FROZEN_RUNTIME_SHA`.

## What this does NOT do

It does not explain **what** the multitable API threw at 2500 rows. It makes that throw **nameable instead of
anonymous**, which is the precondition for finding out. The mid-tier arm will still fail after this — with a
code that says where.

No flag default changed, nothing armed, nothing deployed, no external write.


zensgit:
## Verified in a real dispatched run — and it narrows the root cause to a different subsystem

Dispatched run **30892494202** (this branch merged onto R9's scale leg, `s6a_row_count=2500`):

| | before | after |
|---|---|---|
| `s6aMidTierFirstRunErrorCode` | `SEALED_EXPORT_INTERNAL_ERROR` | **`STOCK_PREPARATION_PERSIST_FAILED`** |
| `s6aMidTierFirstRunHttp` | 503 | **503 — unchanged** |

Both claims in this PR are now confirmed against a running product rather than a unit test: the token becomes
specific, and the surfaced status does not move.

### The branch it landed in is itself the finding

`STOCK_PREPARATION_PERSIST_FAILED` is the **fallback**. Had the upstream throw been a first-party
`StockPreparationSyncRunPersistError`, `remintPersistFailure` would have produced `..._CONFLICT` (409) or
`..._UNPROCESSABLE` (422). It produced neither.

**So the throw is genuinely non-first-party — it comes from the multitable records layer while writing 2500
LINE rows.** The root cause is not in sealed-export at all.

That also falsifies the `READ_PAGE_LIMIT × READ_MAX_PAGES` hypothesis I floated earlier: that bound is
**25,000** (`500 × 50 − 1`), and 2500 is nowhere near it. Recorded so nobody re-tests it.

### Unchanged, and still the reason this matters

```
generation status = ACTIVE      applied_row_count = 2500
run status        = ACTIVATED
DbSealedExportGenerationOnFailure = GENERATION_ACTIVE_BUSINESS_PERSIST_UNKNOWN
```

The run remains in a partially-completed state. What changed is that an operator now gets a code naming
**which stage** failed instead of a bare internal error — the difference between "something went wrong
somewhere" and "the business persist failed after activation".

That is the precondition for the retry decision, not the decision itself. **The recommendation stands: no
S6-A window until a 503 can be told apart from a real failure**, and this PR is the first step, not the
finish.

PR #4750, opened 2026-08-04
test(integration): K3 endpoint vocabulary client/server mirror tripwire (P1 step 1)

zensgit:
**P1, step 1** of the delivery plan (#4749): close the K3 API declaration drift structurally.

## The drift is real and mechanically provable

The P1 survey found **four independent declarations** of the K3 API surface, and the import graph proves they are separately maintained — `apps/web/src/services/integration/k3WiseSetup.ts` imports **nothing** from the plugin.

Worse: **neither plane has an endpoint vocabulary.** Server (`read-source-config.cjs:186`) and client (`readSourceConfigs.ts:265`) both validate only the *syntax* of a relative path, and the server's allowlist covers *key names*, not endpoint values. **A misspelled endpoint passes both.**

## Reusing the shape that already works

`composition-vocab-mirror.spec.ts` already does this job and is already green in a required check: it requires the server's frozen export **directly** — *"require, never text-parse"* — and pins counts so a same-size substitution still trips. This copies that, rather than inventing a mechanism.

### Why an explicit correspondence map, not a loop over one side

Iterating the **client's** keys never notices a server path the client does not carry. Iterating the **server's** just goes red with no way to express *"deliberately client-side absent"*.

The map makes every server endpoint's client fate a **written decision** — so adding a server endpoint without deciding that fate is RED. That is the property actually wanted.

## Found by writing it

The server declares Material `readPath: '/K3API/Material/GetDetail'` and **the client form has no counterpart.** GetDetail is exactly what P3's post-save read-back needs. It is recorded as a `null` entry annotated `KNOWN GAP` — closing it is P3 work, not a silent omission here.

## Mutation-proven against the three real drift shapes

| arm | result |
|---|---|
| baseline | 3 passed |
| **client endpoint drifts** (`Save` → `SaveV2`) | **1 failed — RED** |
| **server adds an endpoint, unmapped** | **1 failed — RED** |
| **template id silently bumped** `v1` → `v2` | **2 failed — RED** |
| restore | 3 passed |

The first two fail with exactly one assertion each, so neither is covering for the other.

## Wiring — both rosters, because they answer different questions

- `integration-guard-run-web-specs.sh` decides **WHAT** runs → token added.
- `integration-guard-guarded-paths.mjs` decides **WHEN** it runs → the spec **and both watched sources** added.

Checked before assuming: `k3WiseSetup.ts` and `k3-wise-document-templates.cjs` appeared in that roster **zero times**. Editing either declaration did not trigger `integration-guard` at all. **A tripwire that does not fire on the edit it guards is decoration.**

## One correction against my own first draft

Both `K3_WISE_DOCUMENT_TEMPLATES` and `K3_WISE_MATERIAL_PROFILES` are **keyed objects** (`material`/`bom`), not arrays. My first version assumed arrays and went red on all three tests. The test now keys by the short name **and separately pins the versioned `id`**, so a silent id bump also trips.

No product code. No flag default changed, nothing armed, nothing deployed.


zensgit:
## P1 step 2 (client derives from the server catalog) — recommending we do NOT do it now

Checked before proposing it, rather than assuming it was the natural next step.

**The client cannot require the CJS catalog** — it is a browser bundle. So "derive" means one of:

| option | cost |
|---|---|
| (a) client fetches the catalog at runtime | needs a **new data surface**: the existing `GET /api/integration/templates/references` returns 2 entries and carries **no `/K3API/` endpoints at all** (verified by calling `listReferenceIntegrationTemplates()` directly) |
| (b) build-time codegen from the CJS catalog into TS | new build step, new failure mode, and the generated file still has to be kept honest by… a mirror test |
| (c) keep the copy, guarded by the tripwire | **done in this PR** |

### Why (c) is enough for this delivery

The tripwire already makes drift **impossible to land**: an unmapped server endpoint, a client value that
moves, or a silently bumped template id each go RED in a required check. Deriving would make drift
*impossible to write*, which is stronger — but the delta between "cannot land" and "cannot write" does not
buy anything for a single-customer, manual-trigger, Save-only first version, and (a) adds a runtime data
surface on the exact path that performs the round's only external write.

**Recommendation: leave the client copy in place, guarded. Revisit deriving when a second customer or a
second profile makes the copy genuinely expensive** — that is when the cost of (a) starts paying for itself.

Not a decision I am making unilaterally: if you want (a) or (b) now, say so and it goes in front of P2.
Otherwise P1 is complete at step 1 and the next work is P2 (read → material.v1/bom.v1 → cleaning → diff
preview), which is on the critical path to actually testing.

PR #4754, opened 2026-08-04
test(integration): wire the K3 read -> stock-prep intake join into the mock chain (P3)

zensgit:
**Reworked.** The first version called a separate K3 mapper I had added in #4751. That mapper turned out
to **duplicate what the intake already does** and derive a **conflicting `erpMaterialId`**, so it was
retracted (#4755). This version exercises the **product path** instead — correct, and a stronger thing
to demonstrate.

Now independent of #4755; can merge in either order.

## Step 7a-2

The row the SQL channel **just read** is fed to `normalizeStockPreparationReadonlyIntake` **unchanged**.

| | Assertion |
|---|---|
| no mapper needed | **0 row errors** — the intake's alias lists already carry `FNumber` / `FItemID` / `FName` / `FModel` |
| field landing | `FNumber → erpMaterialCode`, `FItemID → erpMaterialInternalId`, `FName → erpMaterialName` |
| key **stable** | same material re-read ⇒ same `erpMaterialId`. It is the persist's **key field**; an unstable key means the material lands twice |
| key **namespaced** | `erp_k3` vs `erp_other` for the same internal id ⇒ different keys (the property the retracted `k3:${id}` scheme lacked) |
| **negative control** | a row with no internal id is a **row error**, not a silent pass |

The negative control is load-bearing: without it, everything above would also pass if the intake
validated nothing.

## Value over the unit test

The unit test feeds a hand-written literal. This feeds **the actual output of the read one step above**,
so a drift between *what the read returns* and *what the intake accepts* fails **here**, not at a customer.

## Mutation-proven against the intake module

| Mutation | Result |
|---|---|
| `FItemID` removed from the alias list | demo **exit 1** |
| `sourceSystem` removed from the derivation | demo **exit 1** |
| restored | demo **exit 0** |

Fixture/demo only — no product code touched. Diff vs `main` is one file. The chain still ends with its
own caveat: **"mock pass ≠ customer live pass."**


zensgit:
## 声明收窄(自查,先撤回后陈述)

PR 正文写的「that is the plan's `DB/K3 读取 -> 备料清洗` link, **exercised** rather than asserted」
**比代码实际做的强**。撤回该措辞,准确表述是:

| | 本 PR 实际做的 | 本 PR **没有**做的 |
|---|---|---|
| 读 | demo 内 **mock SQL 通道**读出的一行 | 产品路径上的 K3 `GetDetail` 读 |
| 写 | 映射到 intake **形状**并断言 | 落库到 `erpMaterials` |

价值仍然成立且不重复单测:单测喂**手写字面量**,这里喂的是**上一步真读出来的那一行** ——
读的输出形状与映射的输入契约一旦漂移,在这里红。这是真的,只是范围比原措辞窄。

**因此计划附录 B.7 的「把 K3 读产出接进 `erpMaterials`」仍然未关闭**,不能因本 PR 划掉。
我会在附录里把这两件分开列,而不是让一件冒充另一件。

PR #4768, opened 2026-08-05
ci(stock-prep): staging window rehearsal lane — the runbook executed against a real deployed package

zensgit:
# Staging window rehearsal

Rehearses the customer-window C6 write lifecycle against a real deployed package.
Head `53a30b685`. Rebased onto main after **#4784 merged** (`e6523c949`).

### [P1] The B4 same-instance check compared the target with ITSELF

Correct as reported, and my earlier `kind` gate was **not** a fix for it — that closed a different
hole (same-origin non-K3). The driver created a separate K3 READ record but minted B4 on
`targetSystemId`; the route then loaded `targetBaseUrl` from that same target and the profile
reloaded by the same id. **The real `sourceSystemId` never entered the comparison**, so two
different K3 instances would have passed. No comparator could have fixed this — the check was
structurally incapable of detecting a read/write mismatch.

Root cause was the relation check's reach: with a non-K3 pipeline source the K3 read record is
neither endpoint, so binding to the target was the *only* thing that satisfied #4769. The write
target now declares its paired read record (`config.pairedReadSystemId`) and that one id joins the
relation set, so B4 binds the **real reader** and the guard compares two genuinely different
records.

⚠️ **This widens #4769's ratified relation check by one target-declared id.** Flagged here, in the
code, and in the delivery doc rather than left in the diff. Deliberately narrow: it admits exactly
the record the target itself names, and the binding must still clear the ratified-contract match,
the kind gate and the same-instance gate.

Controls through the real route: **A→A** same physical K3 (same origin, different path — the step
0-b topology) accepted; **A→B** read record on another K3 host refused. Mutations:
`sameK3Instance → true` reds A→B; dropping `pairedReadSystemId` from the relation set reds A→A.

### [P1] The migration proof pinned a COUNT, not a SET

Reproduced: swapping `048` for `074` keeps the count at 6 and everything green — while 074 is
removed from the very map that feeds `Pending`, so `Pending: 0` holds **without 074 running**.

Now byte-identical set equality against `migration-replay.yml`. One detail worth stating: this
step's working-directory is the **unpacked package, not the repo**, so the authoritative file is
addressed through `github.workspace` — a relative path would read empty and pass vacuously, which
is the exact failure mode the check exists to prevent.

074/075 applied-ness is proved by construction: present in the package ⇒ in `getMigrations()`;
absent from the exclude list ⇒ not dropped; `Pending: 0` ⇒ every member has `executedAt`. All three
needed, which is why exclusion is asserted **per name** rather than inferred from a count.

Offline controls: correct list OK · `048→074` DRIFT · empty list DRIFT · unreachable authority
FATAL. The old count criterion passes the substitution (6 == 6).

### [P2] #4784 had no real regression coverage — reproduced, then closed

Flipping `targetSystem.kind` to `targetSystem.role` at **both** real C6 handlers left every suite
green. The test did source-text regex plus a test-local reimplementation of the accessor selection.

The root cause was the **assertion shape**: everything was `notEqual` ("not this particular
error"), a family that cannot distinguish *succeeded* from *failed for another reason*. The A→A
control — which already drives the real dry-run route — is now a positive equality, with the
discriminator measured rather than assumed:

| | error code |
|---|---|
| baseline | `K3_WISE_READ_FAILED` (adapter accessor, credentials present) |
| `.kind → .role` | `K3_WISE_CREDENTIALS_MISSING` (public accessor strips them) |

Reaching a wire error proves credentials resolved. The mutation is now RED.

## Exact-head review (`a9963cea6`) — both P2s fixed

**The step-predicate guard was a denylist of four literals.** Broken in one line:
`dryRunOut !== undefined` on the dry-run step left the suite 10/10 green — degrading the one step
carrying the owner's P1. Replaced with a **positive per-step manifest**, covering all 17 steps in
both directions, extracted by balanced scan (a comma split parsed only 8 of 17 and silently
exempted the rest). All three review mutations RED.

**The same-instance guard had no kind gate.** Origin equality says nothing about *what* is at that
origin; a same-origin PLM record certified the K3 write. Now fail-closed on
`K3_C6_B4_BINDING_KIND_MISMATCH`. The poc's case 6 had been passing for the wrong reason — its PLM
fixture had no baseUrl, so it took the "cannot tell" branch and never exercised the branch its
message named.

**The contract suites were not gated.** They ran in `k3wise-offline-poc`, which is **not** one of
main's 9 required contexts, so a red there did not block a merge. Also invoked in the required
`test` job now. (An earlier claim that they were "gated" was wrong.)

## Dispatch preflight — 0 P1 / 1 P2 / 4 P3, verdict: spend it

Two fixed here:

- **Six lenses hit one root cause.** The wire positive-control step is `if: always()` and the
  runbook step had no `id:`, so it could not distinguish "the logger is broken" from "the driver
  never got there". 10 of 17 gates precede the first GetList, 13 precede the first Save, and
  `record()` hard-exits on the first failure — so any early failure printed a louder, more
  specific, **wrong** diagnosis blaming the mock's logger, on the line right after printing
  `GetList=1`. Now gated on `steps.runbook.outcome`.
- **The migration scope assertion was literal-vs-literal** (a comma count of a list defined 114
  lines above in the same file). Now measured at runtime: list with an empty exclude set, list with
  the real one, require the difference to equal the list size. `migrationScope=CI_EXCLUDED_6` and
  `mig-unscoped.txt` ride the artifact bundle so `Pending: 0` cannot be misread as "the full set
  ran".

## Honest limits

- **The source leg is a stand-in.** This proves the C6 write lifecycle; it does not exercise the
  customer's real source connector. Stated in `REHEARSAL_SUMMARY`, not only here.
- **Contract tests are source-text assertions** and cannot prove a record was written. What proves
  that is the dry-run/apply behaviour and the dispatch itself.
- **What is genuinely unexercised on a fresh DB** is `008_plugin_infrastructure.sql` and
  `20250924140000_create_gantt_tables.ts` — the other four excluded names either run on-prem as
  no-op history markers or have independent fresh-DB coverage.
- A previous version of this body said the B4 gap was "filed separately". **It never was** — and it
  is now fixed rather than deferred.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -82,7 +82,9 @@ jobs:
       # its sibling mock-server test.
       - name: Run K3 WISE rehearsal harness contract tests
         run: |
-          node --test scripts/ops/fixtures/integration-k3wise/rehearsal-harness.test.mjs
+          node --test \
+            scripts/ops/fixtures/integration-k3wise/rehearsal-harness.test.mjs \
+            scripts/ops/fixtures/integration-k3wise/rehearsal-driver-contract.test.mjs
 
       - name: Run on-prem package apply and archive contract tests
         run: |
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -173,6 +173,21 @@ jobs:
       - name: Global History flag manifest contract (R12-C)
         run: node --test scripts/ops/global-history-flag-manifest.test.mjs scripts/ops/multitable-global-history-flag-status.test.mjs
 
+      # REVIEW NIT-2 (exact-head round): these two suites already ran in `k3wise-offline-poc`, but
+      # that job's check is "K3 WISE offline PoC", which is NOT one of main's required contexts --
+      # so a red there did not block a merge. The suites carry the ONLY mechanical guard on the
+      # rehearsal driver's step predicates, including the one that discriminates the owner's
+      # reported P1. A guard that cannot block is documentation, not a gate.
+      #
+      # Invoked here as well, in the REQUIRED `test` job (test (20.x) is a required context), so
+      # the guard actually gates. The duplicate run costs ~1s and is deliberate: leaving the
+      # original in place keeps the offline-PoC job self-contained.
+      - name: K3 WISE rehearsal driver contract (required lane)
+        run: |
+          node --test \
+            scripts/ops/fixtures/integration-k3wise/rehearsal-harness.test.mjs \
+            scripts/ops/fixtures/integration-k3wise/rehearsal-driver-contract.test.mjs
+
       # Integration Guard required-wiring contract (governance slice, standalone from #4603/#4604/
       # #4610 by owner instruction; hardened 2026-07-25 after a P1x2/P2 review — see the workflow's own
       # header for the corrected rationale). Mixed YAML-shape + BEHAVIOURAL contract that makes
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -75,6 +75,15 @@ jobs:
         run: |
           pnpm run verify:integration-k3wise:poc
 
+      # REVIEW P2-1: the rehearsal harness asserts the two guards the owner's HOLD points 3 and 4
+      # depend on (the mock's session gate, and the call logger the workflow's Submit/Audit
+      # absence-check reads). It shipped in ZERO workflows, so those guards' only behavioural
+      # evidence was ungated — a coverage regression, not a coverage gap. Gated here alongside
+      # its sibling mock-server test.
+      - name: Run K3 WISE rehearsal harness contract tests
+        run: |
+          node --test scripts/ops/fixtures/integration-k3wise/rehearsal-harness.test.mjs
+
       - name: Run on-prem package apply and archive contract tests
         run: |
           node --test \
--- a/.github/workflows/stock-prep-staging-window-rehearsal.yml
+++ b/.github/workflows/stock-prep-staging-window-rehearsal.yml
@@ -218,11 +218,23 @@ jobs:
           #   (a) the file is present in the package under test  => it is in getMigrations()
           #   (b) its name is absent from MIGRATION_EXCLUDE      => the provider did not drop it
           #   (c) Pending: 0                                     => every member has executedAt
+          # REVIEW P3-B: there is a FOURTH skip path this chain does not name —
+          # SUPERSEDED_LEGACY_SQL_MIGRATIONS, whose members run as no-op history markers and cancel
+          # out of the known-vs-considered measurement. The conclusion still holds for 074/075
+          # specifically (verified: neither is in that list), but the chain as stated was
+          # incomplete, so it is stated completely here rather than left to read as exhaustive.
           # All three together mean 074/075 executed. Remove any one and the conclusion fails --
           # which is precisely why (b) is asserted per-name rather than inferred from a count.
-          MIG_DIR="$(find . -type d -name migrations -path '*core-backend*' | head -1)"
-          if [ -z "${MIG_DIR}" ]; then
-            echo "FATAL: no core-backend migrations directory inside the package; (a) is unprovable" >&2
+          # REVIEW P2-A: `find … | head -1` picks by readdir order, and the unpacked package has TWO
+          # matching directories — packages/core-backend/migrations (the only home of 074/075) and
+          # packages/core-backend/dist/src/db/migrations (zero 07* files). Depth-first descent emits
+          # the dist one first whenever `dist` precedes `migrations` in readdir, which on the runner
+          # is an ext4 hash order — i.e. undetermined. False-red only (it can never green-light an
+          # unapplied migration), but a lane that hard-fails on a coin flip buys nothing. Address
+          # the directory that actually holds the SQL.
+          MIG_DIR="packages/core-backend/migrations"
+          if [ ! -d "${MIG_DIR}" ]; then
+            echo "FATAL: ${MIG_DIR} is not in the package; (a) is unprovable" >&2
             exit 1
           fi
           for m in 074_repair_sealed_export_runtime_authority_privileges.sql \
--- a/.github/workflows/stock-prep-staging-window-rehearsal.yml
+++ b/.github/workflows/stock-prep-staging-window-rehearsal.yml
@@ -209,14 +209,6 @@ jobs:
           TOKEN: ${{ steps.login.outputs.token }}
           MOCK_K3_URL: http://127.0.0.1:${{ env.MOCK_K3_PORT }}
           TENANT_ID: ${{ env.TENANT_ID }}
-          # THE ONE CREDENTIAL THIS LANE CANNOT MINT FOR ITSELF. Seeding the stand-in source
-          # sheet goes through the OAPI surface (requireScope('records:write')); the integration
-          # TOKEN above cannot reach it, and `staging/install` creates sheets with ZERO rows.
-          # Until this secret exists the driver exits at env validation with a named cause —
-          # deliberately, because a run that proceeded would dry-run an empty source and report
-          # a clean `sourceRows: 0`, which is a vacuous green in the one lane that exists to buy
-          # information. See the PR thread: minting it is an owner action, not this lane's.
-          MULTITABLE_RECORDS_TOKEN: ${{ secrets.REHEARSAL_MULTITABLE_RECORDS_TOKEN }}
         run: node scripts/ops/stock-prep-window-rehearsal-driver.mjs
 
       - name: Save-only invariant seen from the WIRE (mock call log — synthetic keys only)
--- a/.github/workflows/stock-prep-staging-window-rehearsal.yml
+++ b/.github/workflows/stock-prep-staging-window-rehearsal.yml
@@ -136,8 +136,30 @@ jobs:
           DATABASE_URL: postgresql://postgres:postgres@127.0.0.1:5432/rehearsal
         run: |
           set -euo pipefail
+          # PREFLIGHT A1: MEASURE the narrowing instead of asserting a literal against a literal.
+          # The provider DELETES excluded names from the map that feeds `Pending:`, so `Pending: 0`
+          # is scoped to whatever remains and cannot fail for the excluded six. Listing once with an
+          # EMPTY exclude set and once with the real one turns that invisible scope into a number
+          # produced by the runtime, and the difference must be exactly the size of the list.
+          MIGRATION_EXCLUDE= node packages/core-backend/dist/src/db/migrate.js --list \
+            | tee "${RUNNER_TEMP}/mig-unscoped.txt"
           node packages/core-backend/dist/src/db/migrate.js --list | tee "${RUNNER_TEMP}/mig-before.txt"
           grep -q "Applied: 0" "${RUNNER_TEMP}/mig-before.txt"
+
+          known="$(grep -oE '^Pending: [0-9]+' "${RUNNER_TEMP}/mig-unscoped.txt" | awk '{print $2}')"
+          considered="$(grep -oE '^Pending: [0-9]+' "${RUNNER_TEMP}/mig-before.txt" | awk '{print $2}')"
+          if [ -z "${known}" ] || [ -z "${considered}" ]; then
+            echo "FATAL: could not read a Pending count from one of the listings; an unread file is not a measurement" >&2
+            exit 1
+          fi
+          echo "migrationsKnown=${known} migrationsConsidered=${considered} migrationsSkippedByName=$((known - considered))"
+          if [ "$((known - considered))" -ne 6 ]; then
+            echo "FATAL: the runtime skipped $((known - considered)) migrations by name, but MIGRATION_EXCLUDE names 6; the scope of Pending:0 is not what this lane claims" >&2
+            exit 1
+          fi
+          # And say so IN THE EVIDENCE, so the artifact bundle carries the narrowing rather than
+          # leaving a reader to infer "full migration set" from a bare `Pending: 0`.
+          echo "migrationScope=CI_EXCLUDED_6 customerOnPremRunsAll=true"
           node packages/core-backend/dist/src/db/migrate.js
           node packages/core-backend/dist/src/db/migrate.js --list | tee "${RUNNER_TEMP}/mig-after.txt"
           # `Pending: 0` ALONE IS A FALSE GREEN HERE. This step inherits the workflow-level
@@ -284,6 +306,7 @@ jobs:
           echo "loginTokenObtained=PASS"
 
       - name: RUN THE WINDOW RUNBOOK over HTTP (the rehearsal proper)
+        id: runbook
         env:
           BASE_URL: http://127.0.0.1:${{ env.REHEARSAL_PORT }}
           TOKEN: ${{ steps.login.outputs.token }}
@@ -304,8 +327,26 @@ jobs:
           save_calls="$(grep -c "K3CALL POST /K3API/Material/Save" "${RUNNER_TEMP}/mock-k3.log" || true)"
           echo "wireGetListCallsLogged=${get_list_calls:-0}"
           echo "wireSaveCallsLogged=${save_calls:-0}"
+          # PREFLIGHT B1: this step is `if: always()`, and the runbook step had no `id:`, so the
+          # positive control could not tell "the logger is broken" from "the driver never got this
+          # far". 10 of the driver's 17 gates precede the first GetList and 13 precede the first
+          # Save, and record() hard-exits on the first failure — so ANY early failure produced a
+          # louder, more specific, WRONG diagnosis blaming the mock's logger, on the line right
+          # after printing GetList=1. It accused a subsystem that is provably innocent
+          # (mock-k3-webapi-server.mjs logs every request before the session gate and before
+          # routing) and contradicted data it had just printed itself.
+          #
+          # The positive control is only meaningful when the runbook actually reached those calls.
+          if [ "${{ steps.runbook.outcome }}" != "success" ]; then
+            echo "wirePositiveControlSkipped=runbook-did-not-complete (outcome=${{ steps.runbook.outcome }})"
+            echo "The runbook step failed before or during the rehearsal; the FAILED STEP ABOVE and" >&2
+            echo "REHEARSAL_SUMMARY name the actual failed gate. Counts above are informational only:" >&2
+            echo "an absent GetList/Save here means the driver never got that far, NOT that the mock" >&2
+            echo "stopped logging." >&2
+            exit 1
+          fi
           if [ "${get_list_calls:-0}" -lt 1 ] || [ "${save_calls:-0}" -lt 1 ]; then
-            echo "FATAL: positive control failed — mock-k3.log shows GetList=${get_list_calls:-0} Save=${save_calls:-0} K3CALL lines; the log is not actually recording wire calls, so the Submit/Audit absence check below would be meaningless" >&2
+            echo "FATAL: the runbook SUCCEEDED but mock-k3.log shows GetList=${get_list_calls:-0} Save=${save_calls:-0} K3CALL lines; the log is not actually recording wire calls, so the Submit/Audit absence check below would be meaningless" >&2
             exit 1
           fi
           # POSITIVE ALLOWLIST, not a two-word denylist. The old check was
@@ -341,6 +382,7 @@ jobs:
             ${{ runner.temp }}/backend.log
             ${{ runner.temp }}/mock-k3.log
             ${{ runner.temp }}/mig-before.txt
+            ${{ runner.temp }}/mig-unscoped.txt
             ${{ runner.temp }}/mig-after.txt
           if-no-files-found: error
           retention-days: 7
--- a/.github/workflows/stock-prep-staging-window-rehearsal.yml
+++ b/.github/workflows/stock-prep-staging-window-rehearsal.yml
@@ -140,6 +140,25 @@ jobs:
           grep -q "Applied: 0" "${RUNNER_TEMP}/mig-before.txt"
           node packages/core-backend/dist/src/db/migrate.js
           node packages/core-backend/dist/src/db/migrate.js --list | tee "${RUNNER_TEMP}/mig-after.txt"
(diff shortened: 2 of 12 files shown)


zensgit:
## 🔒 HOLD — head-scoped @ `6c88ad289624`(owner 复审,2026-08-05)

**本 PR 在此 head 上不得合并**;auto-merge 已解除,已转 Draft。任何后续推送不自动解除本 HOLD —— 解除条件见文末。

### owner 复审发现(逐条确认为确定性缺陷,非"首跑买信息"类)

| # | 缺陷 | 性质 |
|---|---|---|
| 1 | list read-smoke 请求缺 `intent` ⇒ 首跑**确定性**停在 `intent_required` | 驱动器缺陷 |
| 2 | read-smoke 业务失败返回 **HTTP 200** ⇒ 驱动器把失败当成功;缺失键**负控反而误判** | 驱动器断言错位 |
| 3 | mock runner 未输出调用日志 ⇒ Submit/Audit=0 的线外检查**空转**(owner 实测:调 Submit 后日志仍无 Submit) | 空转断言 |
| 4 | mock 不校验 Cookie/Session ⇒ 未认证请求也能 GetList/Save,预演抓不住产品掉鉴权头的回归 | 保真度缺口 |
| 5 | B4 approved config 全程未 mint/approve/被消费 ⇒ 不能证明客户 runbook 的 B4 绑定环节 | 覆盖缺口 |
| 6 | **产品侧**:`/pipelines/:id/run` 仍可绕过 C6 token 写 K3;named profile 的 `savePath` 仍可被 overlay 覆盖成 Submit 端点 | **P1,独立前置 PR** |

### 解除路径(owner 指定)

1. 本 PR 内修:驱动器(intent、按真实证据字段断言、精确 token 错误码、HTTP timeout)+ mock logger(带正控:日志必须先证明记录了 GetList/Save)+ mock session 校验 + 各修复的承重测试;
2. **独立前置 PR**:K3 目标 C6-only 写入口(`/run` fail-closed)、profile 字面量钉死 `savePath`(overlay 不可覆盖)、C6 消费 approved B4 binding;
3. 新 head 跑 **exact-head 对抗复审** → 通过后 Ready + re-arm;
4. 通用 PR checks 全绿**不构成**放行依据(它们不执行 dispatch-only workflow)。

zensgit:
## 修复轮已推(新 head `c0872149c`)—— HOLD 依约**未解除**

owner 六项发现全部修复,断言形状全部实证提取(intent 必需形、200 双态证据体、409 封套、`K3CALL` 日志正控、session 门、B4 mint→approve 经真路由 + RATIFIED builder 单一真源)。承重:新增 5 测 harness 复现了"调 Submit 日志仍空"的空转缺陷;workflow 线外检查先正控后断言。

**解除序(不变)**:等独立前置 PR(savePath 钉 / C6-only 写入口 / B4 消费门)合入 → 本分支 rebase 消费真门 → **exact-head 对抗复审** → Ready + arm。

zensgit:
## R2 pre-flight — run while still HOLD, against `c0872149c`

The HOLD stands and nothing here changes it: this PR still must rebase onto #4769's gates before it can be dispatched. But two of the owner's concerns were checkable **now**, without a deployment, and the point of checking early is that "首跑大概率红" only excuses *unknown* environment problems — not knowable ones.

### 1. The owner's five in-PR items — behaviour, not grep

`node --test scripts/ops/fixtures/integration-k3wise/rehearsal-harness.test.mjs` → **5/5 pass**, covering the two that were named as vacuous before:

- unauthenticated `GetList` refused / authenticated succeeds with the seeded rows (mock now really requires the session);
- `requireSession` accepts a cookie as well as the header, and **defaults to false** so existing suites are unaffected;
- the logger sees **Save AND Submit** — which is the visibility the workflow's absence-check depends on. Without it, grepping `mock-k3.log` for Submit/Audit passed against an empty stream.

### 2. The one most likely to make the first run red — and it's clean

#4769 **tightens** the B4 gate: an approved binding is now accepted only if its `contentKey` is both self-consistent *and* equal to the ratified content key for that `systemId`. A rehearsal that mints a config the new gate refuses would go red on first dispatch, for a reason nobody would have predicted from this PR's own diff.

Computed both sides directly:

```
driver-minted contentKey : 52fc050c2e4ce2c691dcc2e75685ce7548593d40d6aaf70c0ca4aa82907ecf39
gate-required contentKey : 52fc050c2e4ce2c691dcc2e75685ce7548593d40d6aaf70c0ca4aa82907ecf39
MATCH
```

Remaining preconditions of that gate also satisfiable as written: `actionProfileVersion` = `k3wise.material_list.v1` (equals the gate constant), `systemId` non-empty, `object` = `material`. And the driver mints against `sourceSystemId`, which is inside the pipeline-endpoint relation the gate requires.

### 3. Import pre-flight

`getReadSmokePreset`, `readSmokeSuccessEvidence`, `buildK3WiseMaterialListB4Config` all resolve to real functions, and the driver fails fast and legibly with no env (`rehearsal driver: env BASE_URL is required`). A renamed export is the classic silent first-run failure; it isn't present.

### What this does *not* show

Not a rehearsal pass. The driver needs a deployed `BASE_URL` + `TOKEN`, and the three-layer honesty in its own header still holds: **mock pass ≠ staging rehearsal pass ≠ customer live pass**. This is a pre-flight that removes *knowable* first-run failures, nothing more.

Order unchanged: #4769 merges → rebase this onto the real gates → exact-head review → only then dispatch.

zensgit:
## R2 rebase probe — done ahead of R1 landing, **not** pushed

Rebased this branch onto #4769's gate head locally to find out now, rather than after the merge, whether the two land together. HOLD unchanged; nothing pushed.

**The rebase is clean** — both commits replay onto the gate head with zero conflicts.

### Do #4769's gates break the rehearsal?

This was the real question, and it is the class of failure that only shows up *after* the merge:

| gate added by #4769 | interaction with this rehearsal |
|---|---|
| `K3_WISE_PIPELINE_RUN_DISABLED` (K3 writes are C6-only) | **No effect** — the driver never calls `/pipelines/:id/run`. Its only write path is `external-write/dry-run` → `external-write/apply`, which is exactly what the gate leaves open |
| Named-profile requirement (Guard A) | **Satisfied** — the driver registers the target with `objects.material.profile = material-k3wise-customer-profile-v1` |
| Tightened B4 gate (contentKey must equal the ratified value) | **Satisfied** — recomputed inside the rebased tree: minted == ratified |
| New `baseUrl` path allowlist | **Satisfied** — `http://127.0.0.1:<port>`, with or without trailing slash, and `http://localhost:8081` all construct |

Verified in the rebased tree, not the old one: the gates are present (`K3_WISE_PIPELINE_RUN_DISABLED`, `assertWireEndpointIntent`), the harness is **5/5**, and `contentKeyFor` now resolves from the **public** export #4769 added — the P2-5 fix — rather than `__internals`.

### Why this is not pushed

#4769 will **squash**-merge, so its commits will not be ancestors of `main`. A branch rebased onto the pre-squash head would have to be rebased again immediately, and pushing it now would only make this PR's diff temporarily unreadable. The rebase happens once, against `main`, after R1 lands.

What this probe buys is the knowledge that when that rebase happens it will be conflict-free and the gates will not have broken the rehearsal — neither of which was known before.

Order unchanged: R1 merges → rebase onto `main` → exact-head review → then dispatch.

zensgit:
## Rebased onto the merged gates — head `b0ea1d1dcffd5f6b0389fdd5f7c9080528cbc705`

R1 (#4769) merged at 18:01Z with all **9** required checks green, including `integration-guard` — which runs the full plugin-integration-core CJS chain and is a merge gate, not a signal. This branch is now rebased onto `main` @ `65edb98c6`, so it carries the real gates rather than a copy of them.

The earlier HOLD was bound to `6c88ad289624`. That SHA is two rebases behind; this comment does not lift the HOLD — the exact-head review does or doesn't.

### Gates confirmed present on `main`, then re-verified against this branch

| gate | rehearsal interaction |
|---|---|
| `K3_WISE_PIPELINE_RUN_DISABLED` (K3 writes are C6-only) | driver uses **only** `external-write/dry-run` → `apply`; zero `/run` calls |
| Named-profile requirement | target registered with `objects.material.profile` |
| B4 must equal the **ratified** content key | recomputed against the merged gate: **MATCH** |
| `baseUrl` ASCII path allowlist | `http://127.0.0.1:<port>` constructs, with and without trailing slash |

The third one is the reason this was worth checking before dispatch rather than after: #4769 *tightened* B4 acceptance while this branch was held, so a rehearsal that minted a now-unacceptable config would have failed for a cause invisible in this PR's own diff.

Also re-confirmed post-merge: `getReadSmokePreset`, `readSmokeSuccessEvidence` and `buildK3WiseMaterialListB4Config` all still resolve to real functions, and `contentKeyFor` now comes from the **public** export #4769 added rather than `__internals`.

`rehearsal-harness.test.mjs`: **5/5** against the merged tree.

### What is still not evidence

Nothing here says the rehearsal passes. This workflow is **dispatch-only** — the PR checks do not execute it, which was your point at the original HOLD, and it remains true at this head. Three layers stay distinct: **mock pass ≠ staging rehearsal pass ≠ customer live pass**.

Exact-head adversarial review is running against `b0ea1d1dc`, briefed on your five named defects plus the gate interactions above. Staying Draft until it returns.

zensgit:
## Exact-head review @ `b0ea1d1dc` — **不可 Ready**. One P1 needs your decision.

Staying Draft. One finding landed (`dba4d817b`); the P1 is escalated rather than patched, for a reason I want to be explicit about.

### P1 — K3 cannot be a pipeline **source**, so the rehearsal produces zero net signal

The driver registers the K3 system as `sourceSystemId`. But `external-write-dry-run.cjs:427` reads source rows with a bare `read({object, limit, cursor})` — no key, no read-smoke marker. Reproduced independently against the real adapter using the driver's own config, verbatim:

```
A driver-exact (readMode:'list')  FAIL K3_WISE_READ_LIST_ROUTE_UNSUPPORTED   calls=[]
B readMode omitted (detail)       FAIL K3_WISE_READ_KEY_REQUIRED             calls=[]
C readMode single_record_detail   FAIL K3_WISE_READ_KEY_REQUIRED             calls=[]
```

`calls=[]` is the important part: it dies **before any HTTP**, so nothing reaches the deployed package or the mock. Steps 3–9 — dry-run, token, single-use, apply, read-back, negative control — are unreachable. That is the entire reason the lane exists.

"First run will probably be red" doesn't cover this. That reasoning buys information about *unknown environment* problems; this is a knowable, reproducible-offline class failure, and the run would return **no** rehearsal signal.

### Why I did not just fix it

The obvious patch — pass a read-smoke marker through `readSourceRows` — would relax `K3_WISE_READ_LIST_ROUTE_UNSUPPORTED`, a deliberate product gate, **from a CI observability lane**. That is smuggling a security change through a diagnostics PR, and it is the pattern this repo requires be escalated.

The in-scope fix is to use a non-K3 source, as every ruled precedent does (`k3-wise-c6-write-profile.test.cjs` and the offline demo both use `data-source:sql-readonly` as source, K3 as target only). **But which source is a decision, not a detail** — it determines what the rehearsal proves:

| option | what it needs | what the rehearsal would then prove |
|---|---|---|
| `data-source:sql-readonly` | DB credentials reachable from staging | matches the ruled precedent; source connector is not the K3 one the customer uses |
| `metasheet:staging-source` | a seeded sheet created via API first | fully self-contained; furthest from the customer shape |
| `plm:yuantus-wrapper` | a reachable PLM endpoint | closest to the real shape (`PLM material → K3 WISE` is the canonical pipeline) |

**Root cause, and it is upstream of this PR:** the window runbook (delivery MD §6 step 2) names `target=K3` and never says what the **source** is. The driver filled that silence with a guess. So the runbook needs the source named regardless of what this lane does — otherwise the same ambiguity reaches the customer window, which is not retryable.

My recommendation is the PLM shape, because the rehearsal's value is proportional to its resemblance to the window. But it costs a reachable PLM in staging, and that trade is yours.

### Landed: P2-1

`rehearsal-harness.test.mjs` shipped in **zero** workflows, while its sibling runs in `plugin-tests.yml`. It is the only behavioural evidence for the two guards your HOLD points 3 and 4 depend on — the session gate and the call logger. Ungated evidence for a HOLD's fixes is a coverage *regression*. Now gated, and proven to discriminate:

```
session enforcement neutered -> RED
logger call neutered         -> RED
restored                     -> GREEN, byte-identical
```

My first mutation attempt replaced the *parameter name* rather than the enforcement and left the suite green — worth noting because an ineffective mutation looks exactly like a non-discriminating test.

### Your five HOLD points, re-checked at this head

| # | | evidence |
|---|---|---|
| 1 | list read-smoke intent | **PASS** — full LIST chain returns `recordCount: 2` |
| 2 | not using HTTP 200 as the criterion | **PASS** — both route branches `sendOk`; driver reads the evidence body's `ok` |
| 3 | mock logs calls | **PASS** — no longer vacuous; empty log / missing file / log containing Submit or Audit all RED |
| 4 | mock validates session | **PASS** — defaults to false so existing suites are unaffected; enforcement is load-bearing |
| 5 | B4 minted, approved, consumed | **PARTIAL** — mint and approve hit real routes and the content key matches the ratified value, but "consumed" cannot be evidenced while P1 blocks the dry-run |

Rebase verified clean: merge-base `65edb98c6`, two commits, exactly five files, none of #4769's gates reverted. `MOCK_K3_URL` passes the new baseUrl allowlist, the driver never touches `/run`, and output is values-free.

Confirmed also: this workflow is dispatch-only, has no `paths:`/`if:` no-op, and **has never run** — so no check on this PR carries information about whether the rehearsal works. That was your point at the original HOLD and it still holds.

zensgit:
## Source decision — narrowed by feasibility, not preference

I costed the three options against what the rehearsal environment actually provides. **Two of them are not available there**, which changes the decision from a three-way judgement call into a constrained one.

The rehearsal workflow provisions **Postgres 17 + the deployed package**, and nothing else.

| option | what its adapter actually requires | available in the rehearsal env? |
|---|---|---|
| `plm:yuantus-wrapper` | `context.api.plm` — a **host binding to a real Yuantus**, not an HTTP endpoint I can mock (`plm-yuantus-wrapper.cjs:302-309`) | ❌ no |
| `data-source:sql-readonly` | `config.dataSourceId` resolved through the **host's data-source registry** (`api.test`/`getSchema`/`getTableInfo`) — not a DSN, so the provisioned Postgres alone doesn't satisfy it | ❌ not without registering a data source + its auth scope |
| `metasheet:staging-source` | `context.api.multitable.records.queryRecords()` — **MetaSheet's own multitable API**, present by construction in any deployed MetaSheet | ✅ **yes** |

So `metasheet:staging-source` is the only option that works with what the lane provisions today. My earlier recommendation of PLM was based on resemblance to the customer shape and did not account for the adapter needing a host-level PLM binding — correcting that here rather than leaving it standing.

### What each choice costs and proves

**A — `metasheet:staging-source` (feasible now).** Driver provisions a base/sheet and seeds a few synthetic rows via the multitable API, then registers it as the source. Roughly half a day.
*Proves:* the full C6 write lifecycle against a real deployed package — read-smoke, pipeline, dry-run, token single-use, apply, read-back, `0 Submit / 0 Audit`.
*Does not prove:* anything about the customer's real source connector, since the source is a stand-in. Stating that plainly because the rehearsal's whole value is its resemblance to the window.

**B — provide a reachable PLM in staging.** Then the rehearsal matches the canonical `PLM material → K3 WISE` shape and proves the source leg too. Costs infrastructure you'd have to supply; the driver change itself is small.

### What I need from you

One line: **A** (I implement the staging-sheet provisioning) or **B** (you supply the PLM endpoint, I wire it).

I am not choosing unilaterally, and the reason is in §7.3 of the delivery MD: the source determines what the rehearsal actually evidences, so picking it silently would be deciding what evidence you get. What I have removed is the need for you to research it — two options are simply not available in this environment.

**Unchanged:** relaxing `K3_WISE_READ_LIST_ROUTE_UNSUPPORTED` to let K3 serve as source remains off the table — that is a product gate, and loosening it from a diagnostics lane would be smuggling a security change.

zensgit:
## Option A verified against host code — and what it costs

I checked rather than assumed, since assuming a source works is exactly what produced this P1.

**`context.api.multitable.records.queryRecords` is genuinely exposed to plugins** — `packages/core-backend/src/index.ts:754`, inside the `multitable.records` block of the plugin API the host assembles. So `metasheet:staging-source` will resolve against a real deployed package, not just in a test stub. (The host-loader smoke's stub only wires `multitable.provisioning`, which is why the stub alone would not have settled this.)

That makes A executable. What it involves, stated before doing it rather than discovered halfway:

1. provision a base + sheet through the multitable REST surface, seeded with a few synthetic material rows;
2. register that sheet as a `metasheet:staging-source` system (`config.objects.material.sheetId`);
3. point the pipeline's `sourceSystemId` at it, leaving K3 as target only — the shape every ruled precedent uses;
4. have the rehearsal summary state, in its own output, that the source leg is a **stand-in** — so the evidence carries its own limitation rather than relying on someone reading this thread.

Step 1 is the uncertain part: it depends on the multitable REST surface and on `TOKEN` carrying the right scope in the deployed environment. I would rather establish that against a live instance than guess at it, because a provisioning flow that half-works produces the same "zero net signal" outcome this P1 is about.

### Where that leaves the choice

**A is now proven viable and is the only path I can execute unaided.** B needs a PLM endpoint that only you can supply.

If I hear nothing, A is what I will build — it unblocks R2–R4 and is cheaply redirected later, since you noted the driver change for B is small. I am flagging that intent rather than silently proceeding, because the source determines what the rehearsal evidences and you should get to override it.

What has changed since my last comment: the decision is no longer "pick among three unresearched options". Two are infeasible in this environment, and the third is verified against host source.

zensgit:
## Correction: **A is not executable unaided either.** Retracting "if I hear nothing, I'll build A"

I said I would default to option A. Following that through to the actual provisioning step shows it hits the same wall as B, one layer further in. Retracting before building rather than after.

### What I found

Option A needs source **rows**, not just a source system. Tracing how they would be created:

| step | route | credential | status |
|---|---|---|---|
| provision the staging sheet | `POST /api/integration/staging/install` | integration `write` — **the driver already has it** | ✅ fine |
| seed rows into that sheet | — | — | ❌ **no such route in the integration plugin** |

`staging-installer.cjs` creates **structure only** — sheets, fields, views. It seeds no rows. The plugin's entire route table contains exactly two staging routes (`descriptors`, `install`) and nothing that writes records.

The only record-write path is the OAPI surface (`univer-meta.ts:16434`, `POST /records`), gated by `apiTokenAuth` + `requireScope('records:write')`. That is a **different credential** from the rehearsal's integration `TOKEN` — and minting one is precisely what this repo forbids (`feedback_prod_token_via_login_not_minting`).

### So both options need something only you can supply

| option | what it needs from you |
|---|---|
| **A** staging-source | an API token carrying `records:write` scope for the rehearsal environment (or a sanctioned row-seeding path) |
| **B** PLM | a reachable PLM endpoint the deployed package can bind as `context.api.plm` |

My earlier framing — "A is the only path I can execute unaided" — was **wrong**, and wrong in a specific way worth naming: I verified that A's *adapter* would resolve (`queryRecords` is genuinely exposed at `index.ts:754`, that part holds) and then treated the whole option as feasible without tracing how its input data would come to exist. Verifying one link and generalising to the chain is the same error that put K3 in the source slot to begin with.

Had I proceeded, the rehearsal would have failed at provisioning — a knowable, offline-discoverable failure, which is exactly the category "the first run will probably be red" does not excuse.

### What I recommend

**B**, if supplying a PLM endpoint is at all practical. It matches the canonical `PLM material → K3 WISE` shape, so the rehearsal would prove the source leg too — whereas A proves the C6 lifecycle with an acknowledged stand-in and still costs you a scoped token.

Either way this P1 cannot be closed from inside the lane. R2–R4 remain blocked, and I am not going to manufacture a third option by loosening a product gate.

zensgit:
## ⚠️ Implementing A surfaced a THIRD thing to decide — and it affects B equally

I started building option A and hit a structural conflict between the P1 fix and a gate #4769 added. Reporting before choosing, because the choice is a security-gate semantic.

### The conflict

Three facts, each verified rather than reasoned:

1. **An armed K3 system cannot hold the LIST read config.** All nine list keys are in `K3_PROFILE_FORBIDDEN_OVERLAY_KEYS`, and I confirmed the effect end-to-end: constructing an armed system with `readMode`/`readListBodyKey`/`readListFields`/`maxListLimit` yields `undefined` for every one of them — silently stripped.
   ⇒ read-smoke's LIST probe requires a **separate, unarmed** K3 read system. That is why the driver has two K3 systems.

2. **B4's contract IS the material-LIST read contract**, so it legitimately belongs to that unarmed read system.

3. **#4769's B4 relation check requires** `config.systemId ∈ {pipeline.sourceSystemId, pipeline.targetSystemId}` (`k3-wise-c6-write-profile.cjs:255`).

Under the **current, broken** design those reconcile — because K3-as-source made the read system a pipeline endpoint. **The P1 fix removes exactly that.** With any non-K3 source, the K3 read system is no longer an endpoint, so the B4 binding cannot satisfy the relation check.

**This is not A-specific.** Under B (PLM source) the endpoints are `{plm, k3Target}` — the K3 read system still isn't one. Both options hit it.

### Why I am not choosing

The available resolutions are all gate-semantic:

| | resolution | cost |
|---|---|---|
| **i** | bind B4 to the K3 **target** id | passes mechanically — the gate only compares `contentKey`. But the binding would then name a system that **structurally cannot hold** the list-read config it describes. Semantically false, and the falsehood is invisible to the gate |
| **ii** | widen the relation to systems the pipeline *references*, not just its two endpoints | a change to a security gate that was added specifically to stop *one system's read contract vouching for another system's write* — the round-3 defect |
| **iii** | accept that B4 consumption isn't exercised in the rehearsal | the rehearsal still proves dry-run → token → apply → read-back → `0 Submit/0 Audit`, but loses the one leg that proves the B4 gate itself runs |

**(i) is the tempting one and I distrust it for that reason** — it makes the check pass while making the recorded provenance untrue, which is the shape of defect this line has produced repeatedly.

My inclination is **(iii) for the rehearsal** — exercise everything else, record explicitly that B4 consumption is unrehearsed — and treat (i)/(ii) as a separate design question for the customer window, where the customer's real topology may well have one K3 system serving both roles. But that is a judgement about which evidence you need, so it is yours.

### Status

Option A's driver work is otherwise straightforward and I have the pieces (staging install → sheet id → staging-source registration). I have not committed a source-swap, because where B4 binds changes what the swap should look like.

Still outstanding regardless: the `records:write` token for A.

PR #4831, opened 2026-08-08
fix(dingtalk): closeout review — T3 source authority (all modes) + OPS-01 reversible deny-row evidence (Rev 4.4)

zensgit:
## What

Absorbs the closeout re-review (**CHANGES REQUESTED** against main=45cf3a7c51: 2 P1 + 1 P2) per its prescribed 收尾顺序, and lands the countersign-directed **Rev 4.4** erratum.

### P1-1 — T3 activation source authority (`user-activate.ts`)
- Source resolution/validation is now **unconditional for every activation mode** (was gated on `sso || directoryAccountId`; temp_password / admin_no_password could activate against a dead or missing directory source).
- Membership org **derives from the source integration**; a client-supplied `orgId` is only match-validated → `ACTIVATE_ORG_MISMATCH` (409, closed policy table, HTTP leak-control + bulk per-item).
- Pin flip: the unit suite's "sourceless activation succeeds" positive control is now a **refusal** (`ACTIVATE_SOURCE_MISSING`), plus derived-org positive, mismatch + dual-org steering negatives; real-DB activation fixtures seed real sources.

### P1-2 — OPS-01 deny row becomes ledger-evidenced (Rev 4.4, lock §0.7)
- Was: deny row written unconditionally **outside** the ledger → a never-granted person, deprovisioned then rehired, kept an orphan `enabled=FALSE` row permanently blocking creation-only OAuth `ensureGrant`, with no effect for restore to reverse.
- Now: grant-table writes are **purely effect-driven**. New `directory_deprovision_effects.grant_row_created` (migration `zzzz20260808090000`): CHECK (only `grant_changed` with `before=after=FALSE`), immutability-trigger extension, down() refuses while creation evidence exists. Planner emits the creation effect (exempt from the no-op filter); writer INSERTs the deny row from it (flip stays a guarded UPDATE); restore **DELETEs** the row (drift gate: row exists AND disabled), restoring absence — **scoped**: reversal is available while the event remains `applied`; a later access-graph write supersedes the evidence per §5.4 (pre-existing main guards) and the residual's compensating path is now an owner countersign item (see review section below).

### P2 — zero-effect semantics unified
Globally-clear with no row is never zero-effect anymore; both remaining zero-effect flavours (deny already present / not-globally-clear) assert **OAuth loginability unchanged** by the run.

### Docs
- Design lock: §0.7 Rev 4.4 erratum + **Rev 4.3 终签记录** (per this review's countersign).
- #4827 completion MD: status records the CHANGES_REQUIRED interlude and returns to 代码侧收尾完成 atomically with this fix; §4 item 2 marked superseded by Rev 4.4; new §4bis absorption table.

## Verification

- **Killer golden** (real DB, `directory-deprovision-writer-ledger.db.test.ts`): 无既有 grant → 离岗（deny-row creation evidenced）→ OAuth blocked → rehire restore（row DELETEd, absence restored）→ `ensureGrant` works again.
- Schema probes: CHECK positive/negative, `grant_row_created` immutability, down() evidence-refusal.
- **Six mutation kill-proofs run locally** (each red at the intended test, then restored): restore DELETE branch / planner creation arm / no-op exemption / writer INSERT / `ACTIVATE_ORG_MISMATCH` check / re-gated source resolution (the original P1 replayed).
- Full unit sweep **6666 passed**; 15 touched-API real-DB suites green; `tsc --noEmit` green. Migration runs clean on a fresh test DB and is not in any CI `MIGRATION_EXCLUDE`.

## Constraints preserved

Three lifecycle switches stay **OFF** (`DIRECTORY_PENDING_ACTIVATION_ENABLED` / `AUTH_LOGIN_USE_ALIASES` / `DIRECTORY_DEPROVISION_ENABLED`); canary remains per-item ops GO (NO-GO until this lands and is re-reviewed); U1-U13, real callback/corp-anchor, Transfer T3-T5 remain separate gates.

Sequencing note: the review asked to flip #4827's status to CHANGES_REQUIRED first and restore it after the fix. Both edits land atomically here — at no post-merge state does main overclaim; a separate docs-only flip PR would have delayed the fix behind a full required-check run without changing any reachable state.

## 独立对抗审 (Opus, 2026-08-09, verdict 绑 head 467ae34b57)

**APPROVE-with-hardening, 0 P1.** Three refute-first targets (T3 bypass / writer↔effect 1:1 / restore drift) attacked and not falsified; all six mutation kill-proofs independently reproduced; immutability function diffed line-by-line against the 20260728 version (no column lost); migration clean on a fresh DB and absent from all 9 MIGRATION_EXCLUDE lists. Two P2s absorbed in commit 6d44a8c37f (docs-only delta since the reviewed head):
1. **Reversibility scoping** — the original body/§0.7 overclaimed unconditional reversibility; supersede (e.g. admin `PATCH /status` re-enable) voids the creation evidence through pre-existing guards (`EVENT_NOT_APPLIED`) leaving the deny row without live evidence. Lock §0.7 now scopes the promise; compensating path added to the owner countersign list (MD §4 item 3). Not a regression (on main the deny row NEVER had evidence).
2. **Dual-ACTIVE-source org derivation ambiguity** (no reachable web caller today) → follow-up #4833.

### DIFF
--- a/docs/development/dingtalk-deprovision-reactivation-and-evidence-chain-design-20260723.md
+++ b/docs/development/dingtalk-deprovision-reactivation-and-evidence-chain-design-20260723.md
@@ -71,6 +71,15 @@ CHANGES REQUESTED @ main=45cf3a7c51）会签裁决「Rev 4.3 可以终签」。�
 （`directory-deprovision-writer-ledger.db.test.ts`），mutation 六连杀（restore DELETE、creation 规划、
 no-op 豁免、writer INSERT、org 校验、激活源门控）。
 
+**可逆性边界（独立对抗审 P2，2026-08-09 落账 — 措辞限定，非行为变更）**：creation effect 的
+DELETE 逆转仅在其事件仍为 `applied` 时可用。任何后续 access-graph 写（例：管理员
+`PATCH /status` 重新启用、更新的离岗事件）按 §5.4 将证据 **supersede**，此后 `rehire` 与
+`admin_force` 均被既有门拒绝（`EVENT_NOT_APPLIED`），deny 行成为**无 live 证据的残留**——且后续
+离岗 planner 见行已存在也不再补记 creation effect。该 supersede 语义为 main 既有合同（本勘误
+未触碰那两道门），故 Rev 4.4 的承诺应读作：「**在事件被 supersede 之前**，deny 行的存在性变化
+可被 restore 安全逆转」。残留行的补偿路径（人工清理 or 再证据化）列入 owner 会签清单（验证 MD
+§4 第 3 项）。另：多 active 源的 org 派生消歧为 follow-up hardening（#4833）。
+
 产品方向见 companion — **已赞成**。  
 **Implementation design lock 已于 2026-07-23 批准。**  
 序：lock → T1→T2→T3 → D1…D7 → canary。  
--- a/docs/development/dingtalk-deprovision-reactivation-and-evidence-chain-design-20260723.md
+++ b/docs/development/dingtalk-deprovision-reactivation-and-evidence-chain-design-20260723.md
@@ -1,7 +1,7 @@
 # DingTalk Deprovision 恢复 + Preview/UI 证据链 — 实现级设计
 
 - Date: 2026-07-23
-- Status: **implementation design lock / Rev 4.3**
+- Status: **implementation design lock / Rev 4.4**（Rev 4.3 已终签；Rev 4.4 为 OPS-01 证据模型勘误，见 §0.7）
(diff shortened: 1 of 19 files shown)


zensgit:
Merging: independent adversarial verdict APPROVE-with-hardening bound to code head 467ae34b57; delta since that head is docs-only (2 files, +12 lines — §0.7 reversibility scoping + MD absorption, verified by `git diff --stat`); all 9 required contexts pass by name at 6d44a8c37f; base main unmoved at 45cf3a7c51.

PR #4835, opened 2026-08-08
[HOLD] fix(attendance-w4c0): W4C DML 审批改为派生的精确站点身份 + 固定重数(删除整文件/前缀豁免)

zensgit:
**[HOLD] Draft — do not merge, not armed.** Opened as Draft under a standing owner hold. No flags, staging, soak, or deploy touched.

> **⚠️ WORKFLOW COLLISION with PR 4805.** This PR now touches `.github/workflows/plugin-tests.yml` — a **single added line** (`id: attendance-w4c0-dml-inventory` on the existing attendance collector step) — and therefore also `plugins/plugin-integration-core/lib/sealed-export/vectors/s6a-package-provenance-pins.json`, whose `evidenceFiles.pluginTestsWorkflow` entry is a sha256 **of that workflow file**. The pin is recomputed in the same commit:
> `9920e53278ff36aac7f2954219269a9c425df4ce84ee61238d49871507ff0837` → `9fdf157617c9aa01dfa1325b99348c3a8f309633b94aca4449dbc72709c7ffb5`.
> **Whichever of 4835 / 4805 lands second must re-run `shasum -a 256 .github/workflows/plugin-tests.yml` and write that value into the pin — never side-select one of the two hashes.** The pin check is load-bearing: restoring the stale value reds `sealed-export-package-provenance.test.cjs`.

Lane 2 of the W4C inventory work for issue 4556. Scope: the attendance W4C DML inventory only.

---

## ⚠️ What this PR may be cited for — and what it may NOT

**Owner ruling, recorded here because the distinction is load-bearing.** This PR may **NOT** be cited as proof that every remaining issue-4556 bypass reds. Everything below is measured at this head, not inferred.

**What it now proves.** For a write that (i) the scanner turns into a census site with a **literal, uppercase-keyword** statement and (ii) targets a table in the attendance-owned domain:

- approval is an exact `(relPath, enclosingSymbol, table, verb)` identity with a pinned occurrence count — no whole-file, path-prefix, or bare-symbol predicate survives anywhere (grep below);
- inside an already-approved file, a new table, a new verb, a new symbol, or a **second** occurrence of an approved identity all red, on the named leg;
- **the STATEMENT at an approved coordinate is now pinned too** — see *Statement fingerprint* below. Deleting a `WHERE`, weakening a predicate, extending a `SET` list or changing a `${dynamicTable}` reds in `fingerprintDrift`;
- the identifier-case, schema-qualifier, workspace-directory, `ONLY`-modifier and multi-target-DDL escapes are closed;
- the CI step that runs all of this can no longer be disabled without redding its own wiring gate.

**What it still does NOT prove — each reproduced at this head.**

| open bypass | measured status |
|---|---|
| **A non-literal table name.** `INSERT INTO ${hiddenAttendanceTable}` never becomes a census site at all — `raw census sites` does not move. | **113** dynamic write targets exist in the scanned domain. **OPEN.** Closing it needs constant-binding resolution plus a pinned inventory of the unresolvable remainder; reddening all 113 would produce a gate nobody can keep green. Now an **executable** disclosure (a test asserts the invisibility, with positive controls), so it cannot drift silently. |
| **Lowercase / mixed-case SQL keywords.** `insert into attendance_records …` is invisible. | **OPEN, deliberate.** A case-insensitive keyword alternation matches **229** prose fragments in this domain and would flood the census. Also now an executable disclosure. |
| **The table domain misses `plugin_attendance_*` operational tables.** `isAttendanceOwnedCandidate('plugin_attendance_report_sync_jobs') === false`; the 72-key registry has **zero** `plugin_` keys. | **OPEN.** Four live writes at `plugins/plugin-attendance/index.cjs` (INSERT `:3378`, UPDATEs `:3579/:3611/:3762`, via the const at `:3181`) are invisible **twice over** — dynamic target *and* out-of-domain table. A migrations-derived domain is feasible; it must land together with constant-binding resolution and the curation of those four sites, or it pushes a red gate. **Scoped, not done here.** |
| **Substitution at an identical coordinate AND an identical statement.** | **OPEN by construction** — this is residue (b). The fingerprint narrows it considerably (the substitute must now also be byte-identical after whitespace/case normalisation) but does not close it. |

The three still-open rows predate this PR. They are listed so that a reader who saw only the probes cannot over-read what the inventory guarantees.

### Fixed since the owner's review — with the reproduction that was green before

| was green | now |
|---|---|
| `WHERE` clause deleted from the approved tenant-scoped `UPDATE attendance_schedule_dispatch_requests` → **74/74 green, counters byte-identical** | reds in `fingerprintDrift`, named, with every identity leg still clean and the raw census unmoved — proving it is a statement change, not a new site |
| `UPDATE ONLY attendance_records` → captured `ONLY` as the table, write left attendance scope | resolves to `attendance_records:update` |
| `TRUNCATE a, b` / `DROP TABLE a, b` → 2nd target invisible | every target becomes its own site |
| `if: false` on the CI step → **74/74 green, and the wiring test still printed its own ✔** | reds; the step's key set is pinned to exactly `['id','name','run']` |

---

## Branch reconciliation (this is a consolidation pass)

Four earlier attempts were killed mid-run by network faults and left three branches. **This branch is the live one; the other two are superseded.**

| Branch | Was at | Disposition |
|---|---|---|
| `claude/w4c-dml-site-identity-20260809` (this PR) | `3a439d8e06` | **LIVE.** Carries the conversion. |
| `claude/w4c-dml-exact-site-identity-20260809` | `9325fb411b` (remote), `0e822e9d94` (local only) | **Superseded.** One commit cherry-picked, the rest discarded — reasons below. Remote ref **deleted**; the local ref is retained at `0e822e9d94` as an escape hatch, so nothing is unrecoverable. |
| `claude/w4c-dml-inventory-exact-site-identity-20260809` | `8d47b5abac` = `origin/main`, 0 commits ahead | **Empty decoy. Deleted** (remote and local). |

Exactly one `claude/w4c-dml*` ref is live on the remote now, verified by `git ls-remote`.

**Taken** from the parallel branch: exactly one commit, `4e8427f90e`, the base reproduction of the owner's P1 on unmodified `origin/main`. It changes no file; it exists so the "this used to pass" measurement is durable. Cherry-picked with a provenance note explaining why it sits *above* the fix in this branch's history.

**Discarded**, with reasons — none of it was reviewed, so none of it was taken on trust:

- `site-identity.cjs` (166 lines) keyed identity as `` `${relPath}::${symbol}::${table}::${verb}` ``. That is a **joined string key**: a symbol containing the separator aliases a *different* tuple, which is approval-widening in the very module meant to make approval exact. This branch's JSON-array key does not have that defect. Discarding it is a correctness decision, not a preference.
- Its `curated-debt-entries.cjs` rewrite — its own commit message states the conversion is **INCOMPLETE** ("still carries relPath/startsWith claims"). This branch's is complete (grep below).
- Its 44 lines of collector-test legs — superseded. They assert `newSites` / `multiplicityDrift` / `staleClaims` on the census; this branch's probes go further, mutating the **real file** and re-scanning it with the **real scanner** rather than asserting over a census the test itself assembled.

The one idea worth keeping from those legs — an explicit, legible non-empty-domain assertion — is already present here in stronger form (`missing` is computed by walking the frozen table, so an empty census reds mechanically; see *Completeness assertion*).

## What the owner proved

In an isolated `origin/main` worktree, adding `INSERT INTO attendance_records` to an already-approved file — `packages/core-backend/src/services/AfterSalesApprovalBridgeService.ts` — left the exact-head HEAD scan **PASSING**. Removing only that file's allowlist entry made it fail immediately, accurately naming the new write site. The file name, not the write, was doing the work.

Root cause: `GENERIC_SHARED_ALLOWLIST` admitted by **file path**; P17 carried a whole-file claim (`site.relPath === '.../routes/approvals.ts'`), P26 a **path prefix** (`startsWith('.../ApprovalProductService.ts')`), X05 a whole-file claim. Once a classifier hit, the rest of the audit was skipped for that file.

## The fix — not a tighter allowlist

Adding one more recognised case is the failure mode that produced the allowlists being deleted. So the predicates are **deleted**, and approval is re-grounded on an exact site identity derived from the scan itself, with pinned multiplicity:

```
identity = (relPath, enclosingSymbol, table, verb)
approval = that identity appears in the frozen table
amount   = it must occur EXACTLY `occurrences` times in the census
```

Inside an already-approved file: a **new table**, a **new verb**, a **new symbol** → `unclaimed` → RED. A **second occurrence** of an approved identity → `overCount` → RED. A **vanished** pinned identity → `missing` → RED. There is exactly one recognition rule (map lookup + count comparison); nothing can fall through to a broader claim, because there is no broader claim.

### Site-identity representation

```js
// approved-site-identities.cjs
{
  relPath: "packages/core-backend/src/services/ApprovalProductService.ts",
  enclosingSymbol: "targetUserIds",
  table: "approval_instances",
  verb: "update",
  occurrences: 10,
  entryIds: ["P26"],
},
// generic-shared rows carry no debt id, only a reason:
{
  relPath: "packages/core-backend/src/services/AfterSalesApprovalBridgeService.ts",
  enclosingSymbol: "normalizeCommand",
  table: "approval_instances",
  verb: "insert",
  occurrences: 1,
  entryIds: [],
  genericSharedReason: "named generic bridge for a different product (after-sales), not attendance",
},
```

The lookup key is a JSON array, not a `::`-joined string:

```js
function siteIdentityKey(site) {
  return JSON.stringify([site.relPath, site.enclosingSymbol, site.table, site.verb])
}
```

A joined key lets a symbol containing the separator alias a *different* tuple. Route symbols are collector-generated as `` `${METHOD} ${routePath}` `` from arbitrary path literals, so this is constructible, not hypothetical — and a test now attacks the key directly (see *P26 key*).

**Multiplicity is pinned** by `occurrences` per row, checked in `classify-tracked-sites.cjs`: the census is counted into `observedCountByKey`, then the frozen table is walked — `observed > occurrences` → `overCount`; `observed < occurrences` → `missing`.

`curated-debt-entries.cjs` is now **metadata only**. Every entry's `claims` is derived uniformly at the bottom of the file over the whole array:

```js
CURATED_DEBT_ENTRIES = CURATED_DEBT_ENTRY_METADATA.map((entry) => {
  const approvedKeys = approvedIdentityKeysForEntry(entry.id)
  return { ...entry, claims: (site) => approvedKeys.has(siteIdentityKey(site)) }
})
```

The derived `claims` is assigned **after** the spread, so a hand-written predicate in the metadata is overwritten rather than honoured — `relPath === <file>`, `relPath.startsWith(...)` and bare `bySymbol(file, /regex/)` are not discouraged here, they are **unexpressible**, and that is one readable line rather than a convention. `bySymbol`, `byPathPrefix`, the `byW4C3a*` helpers and `P16_EXACT_ALLOWLIST` no longer exist anywhere in the directory (zero grep hits, comments included).

## Zero surviving claim shapes (the grep)

On `origin/main` (`8d47b5abac`), `curated-debt-entries.cjs` had **61** matches for `relPath ===|startsWith|bySymbol|byPathPrefix`, of which these are the whole-file / prefix claims:

```
$ git show 8d47b5abac:scripts/attendance/w4c0-dml-inventory/curated-debt-entries.cjs \
    | grep -n "byPathPrefix('\|byPathPrefix(W4C3A\|relPath === 'packages/..."
 56:    byPathPrefix(W4C3A_RECORD_EFFECT_ADAPTER)(site) ||
 57:    byPathPrefix(W4C3A_ITEM_EFFECT_ADAPTER)(site) ||
 58:    byPathPrefix(W4C3A_GROUP_EFFECT_ADAPTER)(site) ||
 59:    byPathPrefix(W4C3A_BATCH_EFFECT_ADAPTER)(site)
 64:  return byPathPrefix(W4C3A_CANONICAL_IMPORT_KERNEL)(site)
 69:    byPathPrefix(W4C3A_IMPORT_ROLLBACK)(site) ||
 70:    byPathPrefix(W4C3A_IMPORT_ROLLBACK_BOUNDARY)(site)
236:      byPathPrefix(W4C3A_SYNC_IMPORT_HOST)(site),
343:      byPathPrefix('packages/core-backend/src/attendance/w4c3b-approved-leave-cancellation.ts')(site) ||
382:      site.relPath === 'packages/core-backend/src/routes/approvals.ts' ||          <- P17, whole file
481:    claims: byPathPrefix('packages/core-backend/src/services/ApprovalProductService.ts'),  <- P26, whole file
578:      site.relPath === 'packages/core-backend/src/services/AttendanceExpiryService.ts' &&
599:  return GENERIC_SHARED_ALLOWLIST.some((entry) => site.relPath === entry.relPath)  <- the owner's escape
```

On this branch, **every remaining match is a comment**, and there is **not one live-code hit**:

```
$ grep -n "relPath ===\|startsWith\|bySymbol\|byPathPrefix" \
    scripts/attendance/w4c0-dml-inventory/curated-debt-entries.cjs
 12:// entry cannot declare a predicate of its own, so `relPath === <file>`, `relPath.startsWith(...)`
 13:// and bare `bySymbol(file, /regex/)` claims are not merely discouraged here, they are
212:// Was `site.relPath === 'packages/core-backend/src/routes/approvals.ts'` (whole file) plus a
213:// bare bySymbol over the bridge service. Both are now exact counted identities.
294:// Was a whole-file `startsWith('packages/core-backend/src/services/ApprovalProductService.ts')`
371:// Was a whole-file `relPath === '.../AttendanceExpiryService.ts'` claim narrowed only by an

$ grep -n "relPath ===\|startsWith\|bySymbol\|byPathPrefix" \
    scripts/attendance/w4c0-dml-inventory/curated-debt-entries.cjs | grep -vE '^[0-9]+:[[:space:]]*//'
(no output — exit 1)
```

`bySymbol`, `byPathPrefix`, the `byW4C3a*` helpers and `P16_EXACT_ALLOWLIST` are **deleted**, not left defined-but-unused. The P17 and P26 whole-file/prefix claims are gone. Measured at this head, not restated:

```
$ node -e '...APPROVED_SITE_IDENTITIES...'
total identity rows: 181
pinned occurrences: 212
generic-shared identity rows: 9 their pinned occurrences: 9
repeated (occurrences>1): 19
P17 identities: 10 occurrences: 12      <- was one whole-FILE claim
P26 identities: 12 occurrences: 26      <- was one path-PREFIX claim
```

## Table shape (measured)

`approvedIdentities=181` · `pinnedOccurrences=212` · `repeated (occurrences>1)=19` · `genericShared=9`

Derived once, mechanically, from the live census using the predicates it replaces. The derivation aborted on any heterogeneous or unclaimed site; there were **none** — so the conversion is claim-preserving by construction. It does **not** re-adjudicate whether each of the 181 sites *should* be approved; it only makes each approval exact and counted.

## Probes — the owner's acceptance bar, run as REAL ON-DISK EDITS

These are not the in-repo tests. Each probe edits the working tree, runs the **actual CI gate** (`createWorktreeSource` → `buildRawCensus` → `sources.cjs` enumeration → `classifyCensus` → `classifyTrackedSites`), and is restored by `cp` from a backup held **outside** the repo — never `git checkout -- <file>`. `git status --porcelain` was verified empty after each, and both files' `shasum -a 256` match their pre-probe backups at the end.

The in-repo regression probes exist too and are stronger than hand-built fixtures (they read the real file, splice SQL, re-scan with the real scanner, and swap the rescan into the live census; `censusWithMutatedFile` asserts the file text really changed *and* the scanned site set really changed). But they are **not** what discharges this bar, because they reconstruct the census instead of going through worktree enumeration. Both forms are green.

### PROBE 1 — the owner's exact probe

```
$ cat >> packages/core-backend/src/services/AfterSalesApprovalBridgeService.ts <<'EOF'
export async function probeOwnerExact(db: any) {
  await db.raw(`INSERT INTO attendance_records (id) VALUES (1)`)
}
EOF
$ git status --porcelain
 M packages/core-backend/src/services/AfterSalesApprovalBridgeService.ts

$ node --test --test-name-pattern "exact-head HEAD scan" scripts/ops/attendance-w4c0-dml-inventory-collector.test.mjs
[w4c0-dml-inventory] raw census sites=1221 tracked=213 observedIdentities=182 approvedIdentities=181 unclaimed=1 overCount=0 missing=0 genericShared=9 unclassifiedTables=0 outsideBoundary=0
✖ exact-head HEAD scan: zero new/unclassified/out-of-boundary attendance DML (810.644583ms)
ℹ tests 1 / pass 0 / fail 1

  AssertionError [ERR_ASSERTION]: every business/schedule_fact/shared_hook DML site must be an
  approved site identity — a new table, verb, or symbol inside an already-approved file has no claim
  + [ 'packages/core-backend/src/services/AfterSalesApprovalBridgeService.ts :: probeOwnerExact :: attendance_records:insert' ]
  - []
```

**RED**, in `unclaimed`, naming the site. The file's allowlist entry was **not** removed — the file is still approved for its own two identities. The file's *name* now grants nothing.

### PROBE 2 — a NEW TABLE inside an approved file

Deliberately a *different* approved file, at an *already-approved symbol*, so the only new thing is the table. `ApprovalProductService.ts` is the file P26 used to claim with a whole-file `byPathPrefix()`.

```
        await consumeTimeout()
+       await client!.query(`UPDATE attendance_shift_assignments SET is_active = FALSE WHERE id = $1`, [id])
        await client!.query('COMMIT')

[w4c0-dml-inventory] raw census sites=1221 tracked=213 observedIdentities=182 approvedIdentities=181 unclaimed=1 overCount=0 missing=0 genericShared=9 unclassifiedTables=0 outsideBoundary=0
✖ exact-head HEAD scan: zero new/unclassified/out-of-boundary attendance DML (1070.441917ms)
ℹ tests 1 / pass 0 / fail 1

  + [ 'packages/core-backend/src/services/ApprovalProductService.ts :: consumeAndSkip :: attendance_shift_assignments:update' ]
  - []
```

**RED.** The symbol was already approved (for `approval_assignments:update` and `approval_instances:update`); a new table at that same coordinate has no claim.

### PROBE 3 — a NEW VERB on an already-approved table inside an approved file

`AfterSalesApprovalBridgeService.ts :: normalizeCommand :: approval_instances` is approved for `insert` **only**.

```
     await this.runInTransaction(async (trx) => {
+      await trx.query(`DELETE FROM approval_instances WHERE business_key = $1`, [command.businessKey])
       await trx.query(
         `INSERT INTO approval_instances

[w4c0-dml-inventory] raw census sites=1221 tracked=213 observedIdentities=182 approvedIdentities=181 unclaimed=1 overCount=0 missing=0 genericShared=9 unclassifiedTables=0 outsideBoundary=0
✖ exact-head HEAD scan: zero new/unclassified/out-of-boundary attendance DML (1116.933416ms)
ℹ tests 1 / pass 0 / fail 1

  + [ 'packages/core-backend/src/services/AfterSalesApprovalBridgeService.ts :: normalizeCommand :: approval_instances:delete' ]
  - []
```

**RED.** Verb is part of the identity; an approved table does not carry a new verb.

### PROBE 4 — a SECOND occurrence of an already-approved (symbol, table, verb) tuple

This is what multiplicity buys, and the easiest to get wrong. **Nothing about the tuple changes** — same file, same enclosing symbol, same table, same verb — only how many times it occurs.

```
     await this.runInTransaction(async (trx) => {
+      await trx.query(
+        `INSERT INTO approval_instances (id, status, workflow_key, business_key)
+         VALUES ($1, 'pending', $2, $3)`,
+        [approvalId, REFUND_WORKFLOW_KEY, command.businessKey],
+      )
       await trx.query(
         `INSERT INTO approval_instances

[w4c0-dml-inventory] raw census sites=1221 tracked=213 observedIdentities=181 approvedIdentities=181 unclaimed=0 overCount=1 missing=0 genericShared=10 unclassifiedTables=0 outsideBoundary=0
✖ exact-head HEAD scan: zero new/unclassified/out-of-boundary attendance DML (1109.84ms)
ℹ tests 1 / pass 0 / fail 1

  AssertionError [ERR_ASSERTION]: an approved site identity may occur exactly its pinned number of
  times — a SECOND write at the same file/symbol/table/verb is new debt
  + [ '.../AfterSalesApprovalBridgeService.ts :: normalizeCommand :: approval_instances:insert pinned=1 observed=2' ]
  - []
```

**RED in the multiplicity leg specifically.** Read the counters, because this is where a probe can lie: `observedIdentities` stayed at **181** (no new identity was minted, so it really is the same tuple), `unclaimed=0` (so this is not probe 1 wearing a disguise), and `overCount=1` carries `pinned=1 observed=2`. The duplicate sits immediately adjacent to the original and did **not** collapse, so the 50-line `debtKey` block index does not merge adjacent repeats.

### Must-stay-green control — the untouched tree

```
$ git status --porcelain
(empty)
$ node --test scripts/ops/attendance-w4c0-dml-inventory-collector.test.mjs
[w4c0-dml-inventory] raw census sites=1220 tracked=212 observedIdentities=181 approvedIdentities=181 unclaimed=0 overCount=0 missing=0 genericShared=9 unclassifiedTables=0 outsideBoundary=0
[w4c0-dml-inventory] approved identities=181 pinnedOccurrences=212 repeated=19 genericShared=9
ℹ tests 67 / pass 67 / fail 0
```

**Measured residue on the clean tree: `unclaimed=0 overCount=0 missing=0 unclassifiedTables=0 outsideBoundary=0`.** No legitimate site turned red, so no false positive existed to fix and nothing was waived.

58 → 67 tests in the identity conversion, then 67 → **74** with the gate-fix rows: **16 added, zero removed** across both steps. "None removed or weakened" is checked mechanically rather than asserted:

```
$ git show 8d47b5abac:...collector.test.mjs | grep "^test(" | sort > /tmp/main
$ grep "^test(" ...collector.test.mjs | sort > /tmp/head
$ comm -23 /tmp/main /tmp/head          # tests on main that are gone at head
(empty)
$ git diff 8d47b5abac HEAD -- ...collector.test.mjs | grep "^-" | grep "assert\."
(empty)                                  # zero assertions deleted
$ git diff 8d47b5abac HEAD --stat -- ...collector.test.mjs
 1 file changed, 367 insertions(+), 6 deletions(-)   # +43 assert. lines, -0
```

### Restore hygiene

```
$ shasum -a 256 packages/core-backend/src/services/AfterSalesApprovalBridgeService.ts \
                packages/core-backend/src/services/ApprovalProductService.ts
539a6f8474e97993b3d0fbb876018c04bda0a2af3787db4eeef294d3e08dd510  .../AfterSalesApprovalBridgeService.ts
7d64deae69779e5da6662c9b0d4ef4a6d2ad9badfe0d55267c7321fa56089022  .../ApprovalProductService.ts
```

Both equal the pre-probe backups.

## The guard's own probes are load-bearing (mutation test)

Asserting "the escape reds" is worth nothing if the assertion cannot fail. Reintroducing a whole-file fallback into `classify-tracked-sites.cjs` — the exact shape this PR deletes:

```js
if (!approved && [...APPROVED_SITE_IDENTITY_BY_KEY.values()].some((r) => r.relPath === site.relPath)) {
  genericAllowlisted.push(site)
  continue
}
```

turns **7 tests red**. Re-measured at this head (`3348d31abe`), not carried over from the 65-test run where it was first taken — `ℹ tests 67 / pass 60 / fail 7`:

```
✖ owner escape probe: a new table inside a whole-file-approved file is unclaimed…
✖ owner escape probe: a new VERB on an already-approved file/symbol/table is unclaimed
✖ owner escape probe: a new SYMBOL inside the whole-file-approved ApprovalProductService is unclaimed
✖ positive control: deliberate plugin-style direct INSERT bypass is unclaimed
✖ positive control: raw COPY FROM STDIN into attendance_records ahead of preflight is unclaimed
✖ W4C-3c P16 exact allowlist: new DELETE under allowed file/symbol or different table/verb is unclaimed
✖ W4C-3c mutation: reintroducing live second UPDATE after result edit is caught by scanner
```

The two **multiplicity** probes stay **green** under that mutation — correct, since it only widens the `unclaimed` leg. The legs discriminate; they do not all red together. Mutation reverted via file-backup `cp`, working tree verified clean.

## What the reviewer will throw at it — measured, not argued

The scheme was fed an evasion battery through the **real scanner** (`scanFileForDmlSites` → `classifyCensus` → `classifyTrackedSites`), each case a write to a tracked attendance table from a new symbol. `SEEN` means the write became a census site at all.

| Evasion shape | SEEN? | Result |
|---|---|---|
| A. plain raw SQL (control) | yes | `attendance_records:insert` |
| B. SQL split across lines in one template literal | yes | resolved across the newline |
| I. SQL comment wedged between verb words, `INSERT/*c*/INTO`, **inside a template literal** | yes | comment masker strips it first |
| I′. the same shape inside a **single- or double-quoted** string | **NO** | **corrected**: the masker preserves `squote`/`dquote` bytes verbatim, so SQL-comment masking is a template-literal behaviour. Narrow, not general — plain quoted SQL with no interleaved comment IS seen in all three token kinds |
| L. CTE-wrapped write, `WITH … INSERT INTO` | yes | |
| M. data-modifying CTE, write **inside** the `WITH` clause | yes | |
| N. `INSERT … ON CONFLICT DO UPDATE` upsert | yes | |
| O. **two sites merged into one statement** (CTE writes t1, outer writes t2) | yes | **both** sites produced, not merged |
| P. schema-qualified, `INSERT INTO public.attendance_records` | yes | **was NO — fixed in this PR** |
| R. SQL in a module-scope const, executed elsewhere | yes | attributed to `(module-scope)` |
| C. verb and table in separate concatenated fragments | **NO** | no site produced |
| D. table name interpolated from a variable | **NO** | |
| E. SQL assembled by a helper function | **NO** | |
| Q. table name assembled by prefix concatenation | **NO** | |
| F/G. query builder (knex-style `db(t).insert`, kysely `insertInto`) | **NO** | no SQL text exists |
| H. ORM `repository.save()` | **NO** | |
| J/K. lowercase / mixed-case **KEYWORDS** | **NO** | pattern is case-sensitive by design |
| S. uppercase **TABLE IDENTIFIER**, `INSERT INTO ATTENDANCE_RECORDS` | yes | **was SEEN-then-silently-dropped — fixed in this PR.** PostgreSQL folds an unquoted identifier, so this resolves to the same relation and the same site identity |
| S2. mixed-case table, `Attendance_Records` | yes | **fixed in this PR** — same fold |
| S3. uppercase verb **and** table together, `UPDATE ATTENDANCE_RECORDS` | yes | **fixed in this PR** |
| S4. schema-qualified uppercase, `public.ATTENDANCE_RECORDS` | yes | **fixed in this PR** — the qualifier already resolved correctly; the result was then dropped on case |
| S5. uppercase shared table, `APPROVAL_ASSIGNMENTS` | yes | **fixed in this PR** — covers all three `SHARED_TABLE_NAMES` |
| S6. quoted mixed-case, `"Attendance_Records"` | yes | **fixed in this PR** — PostgreSQL does **not** fold a quoted identifier, so it is a *different* relation: admitted to scope, then red as an unclassified attendance-owned table. Not folded into the approved one, and not silently dropped either |
| T. attendance DML in `plugins/plugin-after-sales` | yes | **was never scanned at all — fixed in this PR.** A workspace-member directory with no `package.json` was not a scan root, so the raw census did not even move |

Case O is the one worth calling out positively: **two writes merged into a single statement do not merge into a single site**, so that evasion does not work.

Rows S–S6 and T are the two findings the exact-head adversarial gate raised against this PR. They are recorded here as evasion rows rather than buried in a changelog because both were **green** before the fix, and both defeated the gate *upstream of the identity table* — nothing about the identity scheme was consulted at all. See "The two gate findings" below.

## Stated residue — two different kinds, deliberately not merged

Collapsing these into one "known limitation" paragraph would be the error where two distinct failures get one name. They are separate because the fix for each is different.

**(a) SCANNER blindness — the write never becomes a census site.** Identity precision is irrelevant here: there is no site to identify, approve, or count. Cases C/D/E/Q (dynamically assembled SQL), F/G/H (query builders and ORMs — no SQL text exists to match), J/K (lowercase **KEYWORDS**; case-sensitivity is a deliberate trade — a case-insensitive pattern matches English prose in comments and floods the census), and I′ (a comment deliberately wedged between the words of a verb inside a non-template string). Closing this class needs an AST/dataflow scanner or a runtime guard, not a better allowlist. **Disclosed, not fixed.**

Two things this clause used to be read as covering, and does not:

- **The table IDENTIFIER axis is not part of the keyword trade and is no longer residue.** The stated rationale ("a case-insensitive pattern matches English prose") is a property of the *keyword alternation* only — the verb still has to match uppercase to reach the capture at all, so case-folding the captured identifier adds exactly zero prose noise. It is now resolved by PostgreSQL folding rules and matched case-insensitively, consistent with the three READ legs, which have always carried `gi`. **Fixed in this PR, not disclosed.**
- **A directory outside the scan domain was never residue either** — it was an undisclosed hole. Class (a) is about writes the scanner *sees but cannot parse*; an unscanned directory is a write the scanner never looks at, which does not move the raw census at all and so leaves no trace. **Fixed in this PR.**

One genuine consequence of the fold, stated rather than left to be found: because an uppercase spelling resolves onto the **same** identity, *substituting* an approved lowercase write with its uppercase spelling stays green — same relation, same pinned count. That is residue **(b)** below (substitution), not a new hole. *Adding* an uppercase write reds on multiplicity, which is asserted by a frozen must-red row.

**(b) IDENTITY-SCHEME blindness — the site exists but the identity cannot separate two of them.** `enclosingSymbol` is the collector's nearest-preceding-declaration heuristic, not AST attribution. Two genuinely different call sites sharing a helper name collapse into one identity, so **substituting** an illegitimate write for a legitimate one at the very same `(file, symbol, table, verb)` does not red. **Multiplicity bounds additions, not swaps** — this is the precise limit of what probe 4 buys, and it is not fixable by counting.

Also residual, and unchanged by this PR:

- **The P02-shaped hole**, which is class (a) inside our own inventory: P02 documents a real second-pass `UPDATE` whose site the symbol heuristic never resolved, so it has zero rows in the identity table and is indistinguishable from a metadata-only entry. `overCount`/`missing` can say nothing about a write that never became a site.
- `debtKey` in `collector.cjs` still uses `::` joins. It is **not** an approval key — it is the frozen `e0defbe26` manifest's row key, and re-encoding it would break the byte-reproducibility obligation this suite proves. Left alone deliberately.
- The `debtKey` 50-line block index is deliberately **not** part of the claim identity — it would red on unrelated line shifts. Multiplicity, not block index, separates repeated sites.
- The frozen artifact's `symbolTableCoverage` is a symbol × table **cross product** — the same widening class, inside the pinned-baseline proof. Closing it requires regenerating the artifact, a separate decision.
- Pre-existing, not a regression from this branch: `generate-baseline-manifest.cjs` already fails at the pinned ref on `origin/main` with 13 unclaimed sites, so the "re-run and diff" reviewer workflow was already non-reproducing. The frozen artifact is **not** regenerated here and is not in the diff.

- **`.vue` is not a scannable extension, and that is now a stated residue with a measured reason** rather than an omission. `maskCommentsForDmlScan` is a JS/SQL quote state machine and an SFC's HTML `<template>` section desynchronises it (an apostrophe in ordinary prose opens a phantom string literal), after which `//` comments stop being masked. Measured across the 242 `.vue` files in `apps/web`: enabling the extension yields exactly **one** DML-pattern match, and it is a false positive of precisely that shape — the words *"on UPDATE so"* inside a `//` comment in `UserManagementView.vue`. Zero attendance-owned matches. Scanning SFCs needs an SFC-aware masker first. `.tsx`/`.jsx` **were** added (measured: 0 files in the domain today on both the live tree and the pinned ref, so it is pure fail-closed future-proofing with exactly zero census effect).

## The two gate findings — both upstream of the identity scheme

An exact-head adversarial gate at `3348d31abe` confirmed everything this PR claims about the identity conversion and could not break it, and then blocked on two escapes that **defeat the gate before the identity table is ever consulted**. Neither fix touches the identity machinery.

### P1 — the table identifier escaped the census on CASE

`INSERT INTO ATTENDANCE_RECORDS` **did** mint a raw census site — the keywords are uppercase, so the case-sensitive pattern matched fine — and was then discarded by a case-**sensitive** `isAttendanceOwnedCandidate`, in a `continue` that runs *before* `classifyTable`. Not tracked, not `unclaimed`, not even `unclassifiedTables`: silent. In PostgreSQL an unquoted identifier folds to lower case, so it named the **same relation** as the approved lowercase site.

The discriminating signature, measured: `raw` moves in both spellings, `tracked` moves in only one.

```
A1-lower  INSERT INTO attendance_records   raw=1221 tracked=213 unclaimed=1   RED
A1-UPPER  INSERT INTO ATTENDANCE_RECORDS   raw=1221 tracked=212 unclaimed=0   GREEN
```

**The reading implemented, and why.** PostgreSQL identifier folding, per axis — not a blanket `toLowerCase()`:

| token | resolves to | because | disposition |
|---|---|---|---|
| `ATTENDANCE_RECORDS` (unquoted) | `attendance_records` | PostgreSQL folds unquoted identifiers | same relation, **same site identity** — reds on multiplicity |
| `"Attendance_Records"` (quoted) | `Attendance_Records` | PostgreSQL preserves quoted identifiers — this is a **different** relation | admitted to scope by a *union* ownership test, then fails `classifyTable` (the registry holds only the real lowercase tables) → `unclassifiedTables`, a hard CI failure |
| `"Attendance_Records` (asymmetric) | `attendance_records` | not a quoted identifier in any dialect | folded — the **wider** of the two readings, so a stray quote is not an evasion |

Blanket-lowercasing would have been wrong for the quoted case (it would let a quoted identifier inherit a lowercase table's approval); dropping the quoted case would have repeated the original defect (silent). The union scope test plus classification-by-resolved-name gives both halves: *not* accepted as the approved table, and *not* silently dropped. The regex now captures the whole final table token, quotes included, so the distinction survives the match instead of being discarded by it; group indices are unchanged.

The reading is PostgreSQL's because the attendance-owned tables are PostgreSQL relations. `packages/mssql-readonly-utils` is in the scan domain but owns none of them, and under SQL Server's default collation identifiers are case-insensitive quoted or not, so folding is at worst redundant there.

`hasLiveDmlOnTable` folds both sides too: every caller asserts an **absence** (`!hasLiveDmlOnTable(...)`), so a wider match can only make those guards red more often.

### P2 — a live product plugin sat outside the census domain

Root promotion required `<dir>/package.json`, so `plugins/plugin-after-sales` — a real product plugin (`plugin.json` manifestVersion 2.0.0, an 850-line `index.cjs`, the contributed route `/p/plugin-after-sales/after-sales`, 13 `lib/*.cjs` modules), in the **same after-sales approval domain as this PR's centrepiece file** — matched the `plugins/*` workspace pattern but was scanned by nothing.

A different mechanism from P1, and the counters prove it — here `raw` does **not** move, because the file is never opened:

```
D1         attendance DML -> plugins/plugin-after-sales/lib/refund-approval.cjs   raw=1220  GREEN
D1-control byte-identical text -> plugins/plugin-attendance/index.cjs             raw=1222  RED
```

**The new rule: membership is the workspace manifest, with no marker-file probe at all.** Accepting `plugin.json` as a second marker was rejected — it moves the same fragility one file along, and the mutation matrix below shows that alternative still loses `packages/claudedocs`. The rule fails closed: a directory the collector cannot classify is **scanned**, not skipped. Directories that are absent, non-package or source-free cost nothing.

**New domain size: 20 roots, up from 13.** The seven newly promoted directories are `packages/claudedocs`, `plugins/example-plugin`, `plugins/hello-world`, `plugins/plugin-after-sales`, `plugins/plugin-test-a`, `plugins/plugin-test-b`, `plugins/sample-basic`.

### Census neutrality — measured per site, not by counter equality

Counter equality cannot tell "nothing changed" from "one site entered and one left", so both fixes were checked by diffing the full census row by row.

**Live tree** — `raw` 1220 → 1220, **zero rows added, zero removed**, exactly **26 rows changed**, and every one of them a case-fold of SQL noise the scanner had been mis-capturing as a table (`SKIP`, `LIMIT`, `RESTRICT`, `PAIR` from `FOR UPDATE SKIP LOCKED` / `ON UPDATE RESTRICT` shapes) plus `t_ICItem` → `t_icitem`. **None entered attendance scope.** The tracked-site key set is **byte-identical** (212 keys). All counters zero, no false positives.

**Pinned baseline ref `e0defbe26`** — the leg that would have forced a frozen-artifact regeneration, checked explicitly because `provePinnedBaselineObligation` hashes a census built over that ref through the same `discoverRuntimeRoots`:

```
pinned roots   : 13 -> 20
pinned raw     : 1122 -> 1122
pinned tracked : 224 -> 224
trackedHash    : 921e097094aabb62edab24f207762d19883b18d195e4493e1d35cb1aebc599ff  (UNCHANGED)
artifactHash   : 921e097094aabb62edab24f207762d19883b18d195e4493e1d35cb1aebc599ff
HASH MATCH     : true
```

**No baseline regeneration is owed.** The seven newly promoted directories contribute **0** sites to all four censuses that share `discoverRuntimeRoots` (DML, P25 call-path, attendance-record read, calculation read), on the live tree and at the pinned ref alike — 16 scannable files, zero sites.

### Frozen must-red rows added (7 tests: 67 → 74, none removed, none weakened)

Every row asserts the **signature**, not just redness: the raw census must grow by exactly one site. That is what separates "the scanner saw the write and the classifier judged it" (P1) from "the file was never opened" (P2) — redness alone would show neither.

- uppercase table, mixed-case table, uppercase verb+table, schema-qualified uppercase, uppercase shared table → all red in `unclaimed`, folded onto their real lowercase relation, with `unclassifiedTables` asserted **empty** (an unquoted identifier folds to a *registered* table, so it must land in its bucket, not fall out as unclassified);
- quoted mixed-case `"Attendance_Records"` → reds in `unclassifiedTables` with `unclaimed` asserted empty, plus a control that a quoted **lowercase** identifier still classifies normally;
- **uppercase spelling of an already-approved write** → `overCount pinned=1 observed=2` with `observedIdentityCount` **unchanged**. This is the row that distinguishes the implemented reading from the crude "anything non-lowercase becomes unclassified" alternative: folding must resolve onto the identity that is *already pinned*, minting no new one;
- the root set asserted against an expectation **derived from `pnpm-workspace.yaml` inside the test**, so any marker-file rule fails it by construction, with two non-vacuity preconditions (the plugin still has **no** `package.json`, so the row still exercises the escape; and it **does** have `plugin.json`, so it really is product code);
- attendance DML in `plugins/plugin-after-sales/lib/refund-approval.cjs` reds by name, with the module's DML-free state asserted first so the `+1` is exact.

The domain rows run through a new whole-pipeline helper that drives the real `buildRawCensus` over a source with one mutated path, so **root discovery and the directory walk are exercised too**. The pre-existing `censusWithMutatedFile` rescans a single file and splices it in, which can prove classification but could never have detected a domain hole.

### Mutation matrix — every new row proven load-bearing, and the legs proven exclusive

| mutation | tests red | which |
|---|---|---|
| ownership predicate back to exact match | 2 | the ownership unit row + the **quoted** must-red row only |
| `resolveTableIdentifier` stops folding | 3 | the resolver row + the five unquoted rows + the identity-equality row |
| root promotion back to `package.json` | 2 | the root-set row + the `plugin-after-sales` must-red row only |
| root promotion to `package.json` **or** `plugin.json` (the "obvious" alternative fix) | 1 | the root-set row — and it names `packages/claudedocs`, proving the row gates the *mechanism*, not the one directory |
| `discoverRuntimeRoots` returns `[]` | 24 | fail-closed signature `missing=181`, `unclaimed=0` |
| `isScannablePath` always false | 23 | same signature |
| multiplicity guard `if (observed > occurrences)` → `if (false)` | 2 | the original probe-4 row **and** the new uppercase identity-equality row |

The first two mutations red **disjoint** sets: the fold carries the unquoted axis, the union scope test carries the quoted axis, and neither covers for the other. Every mutation was restored from a file backup with `git status --porcelain` empty afterwards.

## Statement fingerprint — approval pins what the write DOES, not only where it is

Each approved identity now pins `statementFingerprints`: a **sorted multiset** of normalised statement hashes, one per pinned occurrence. `classifyTrackedSites` gains a `fingerprintDrift` leg. Purely additive — `unclaimed` / `overCount` / `missing` are untouched, so no existing leg can have been weakened.

**Multiset, not set, and that is measured.** Of the 19 repeated identities, **11 have occurrences whose statements differ**. A set would collapse those and silently lose an occurrence. The test asserts both the length invariant and that at least one repeated identity really does have differing statements, so the distinction cannot quietly become untested.

**A normalised text hash, deliberately not a parse.** Whitespace runs collapse and case folds, so reindentation and keyword case do **not** churn the gate — a gate that reds on formatting gets re-pinned reflexively until nobody reads the diff. Comments are already blanked by the masker. Everything else is significant; asserted red: `WHERE` deleted, `WHERE` weakened, `SET` list extended, predicate negated, `${dynamicTable}` changed. A parser would be narrower and would need a dialect; this is the fail-closed direction.

**Regeneration was mechanical, not hand-edited.** The generator walks row blocks structurally, cross-checks every row's four identity fields against the loaded module *in order*, requires the observed fingerprint count to equal the pinned `occurrences`, and refuses to write on any mismatch. Result: **181 insertions, zero deletions, zero modified lines** — no field reordered, no value rewritten. `debtKey` is unchanged, so the frozen `e0defbe26` `manifestContentHash` is untouched and the pinned-baseline obligation still passes.

**`fingerprintDrift=0` on the clean tree is true BY CONSTRUCTION and is not evidence that the fingerprint works.** The 181 pinned multisets were generated by the same function, from the same tree, in the same run — of course they match. Saying so explicitly because "all counters zero" is exactly the line a reviewer will otherwise read as proof. The actual evidence is:

- the **WHERE-deletion must-red row** — the identity legs all stay clean, the raw census does not move, and only `fingerprintDrift` reds, by name;
- the **normalisation unit rows** — five semantic changes must move the hash, two cosmetic ones must not;
- **mutation**: neutering the drift comparison reds only the WHERE-deletion row; making the fingerprint a **constant** reds that row, the normalisation row, *and* the full-tree gate;
- **alignment over the whole census** — a fingerprint is non-empty and stable even when taken over the *wrong* span, which would pin the wrong text and stay green forever. So every one of the 212 tracked sites is asserted to have been fingerprinted over text containing its own table **and** its own verb. Mutating `statementTextAt` to return the first literal in the file instead of the containing one reds that row, the WHERE-deletion row and the full-tree gate. Sites carry `statementText` alongside the hash precisely so this is checkable rather than assumed.

## Newly triggered workflows

Editing `plugins/plugin-integration-core/lib/sealed-export/vectors/s6a-package-provenance-pins.json` matches the `pull_request` path filters of three **non-required** workflows — `sealed-export-s5-sqlserver.yml`, `sealed-export-s6a-authority-row-lock.yml`, `sealed-export-s6a-grant-repair.yml` — which will now run on this PR and did not before. Their container-backed legs cannot be run locally here. What *was* run locally is the only consumer of the changed key: `sealed-export-package-provenance.test.cjs` 1/1, and mutation-verified (restoring the stale pin reds it). `scripts/ops/multitable-onprem-package-verify.sh` and the S6-A PowerShell acceptance test were checked and do **not** read `evidenceFiles.pluginTestsWorkflow`. `integration-guard-required-wiring-contract.test.mjs` parses this workflow directly and was re-run: 62/62.

## CI self-wiring — the gate can no longer switch itself off

The wiring test was two `assert.match` calls over the whole workflow string. `if: false` on the step left the suite green *and* the test still printed its own ✔. It now locates the step by a pinned `id`, derives the step's block from YAML indentation, and asserts the enclosing job is `test`, the exact command, and the step's **key set exactly** `['id','name','run']`.

Pinning the key set rather than blacklisting `if` / `continue-on-error` / `shell` is the point: trap enumeration does not converge — the next disabling key is always the one not on the list. Both original `assert.match` calls are retained verbatim; in the `if: false` transcript they still pass and only the key-set assertion reds, which is itself the proof that they were the green-against-nothing part.

The analyzer is a pure function of the workflow text, so the must-red matrix drives the same code path as the real assertion. Rows: `if: false`, `if:` on an expression, `continue-on-error`, alternative `shell`, redirected `working-directory`, `env` override, step moved to another job, command re-pointed, id removed, id duplicated — plus a **block-boundary control in both directions** (a neighbouring step's `if:` must not be attributed to this step, and this step's own must still be caught).

### Schema-qualified write targets (a one-token bypass of the whole gate)

`DML_LINE_PATTERN` captured the **first** identifier after the verb, so `INSERT INTO public.attendance_records` produced a site on the table `public`. `public` is not attendance-owned, so the write fell out of the tracked buckets entirely — no identity lookup, no count, no allowlist, no `unclaimed`. Exact-site identity cannot fix this by construction (class (a) above), which is why it belongs with the conversion rather than after it. This repository's other DML scanner (multitable-revision-disposition) already handled schema qualifiers.

The qualifier group is `*` not `?`, so `db.public.attendance_records` resolves too, and it can only ever **widen** what the scan sees — the change direction is fail-closed.

Measured, not assumed:
- **census-neutral at HEAD**: raw 1220 → 1220, tracked 212 → 212, observedIdentities 181 → 181, all legs still 0. No pinned identity moved, so the frozen artifact stays byte-reproducible (that leg runs in this suite and passed).
- no current attendance-table DML in the repo is schema-qualified; the schema-qualified DML that does exist is multitable / data-source / sqlserver fixture code on other tables.
- **mutation**: restoring the old unqualified capture reds exactly the new test and nothing else (measured at the time as 67 tests / 66 pass / 1 fail) — load-bearing, and nothing else was quietly depending on the mis-capture.

The new test carries a negative control (an unqualified write must still resolve to one site on the right table), so a qualifier group that swallowed the table name would red too.

### P26 key

`p26-approval-assignment-classification.cjs` was still on `[relPath, enclosingSymbol, verb].join(' :: ')` — the same aliasing defect the identity table's JSON key exists to avoid, left inconsistent in a file this work covers. Now JSON-encoded. Its result **is** asserted (`unclassified`, `countDrift`, `stale` all `deepEqual([])`), so this is a live gate, not decoration.

The new test attacks the **key**, not the claim: it builds two genuinely different `(relPath, enclosingSymbol, verb)` triples that the old join maps to the same string, and asserts the impostor stays `unclassified` (and the decoy classification reports `stale`) rather than inheriting the classification — with a positive control that the genuinely matching triple *does* classify, so the leg cannot pass merely because the classifier rejects synthetic input. Mutation: restoring the joined key reds exactly this test (67 / 66 pass / 1 fail).

## Claim preservation, proven not asserted

Both checks **re-run at this head** (`3348d31abe`) against `origin/main` (`8d47b5abac`), loading the base modules out of git into a temp dir and comparing against the live ones:

```
METADATA_IDENTICAL_TO_BASE
tracked sites compared: 212   claim-assignment differences: 0
```

- **Entry metadata is byte-identical to base**: `CURATED_DEBT_ENTRIES` minus `claims`, serialised from both sides and diffed. Every `owningSlice`, `sharedHook`, `confidence`, `canonicalizedBy`, `residualClassification`, `residualEvidence` untouched.
- **Claim assignment is identical to base**: for all **212** tracked sites in the live census, the set of claiming entry ids *and* the generic-shared disposition are compared under the **old predicates** vs the **new derived claims**. Zero differences.

So this PR changes *how* approval is decided and *what it bounds*, not *who owns what*.

## Completeness assertion

`unclaimed = 0` is retained verbatim. The non-empty-domain leg is `missing`, and it is mechanical, not decorative: `missing` is computed by walking the **frozen identity table**, not the census, so an empty or truncated census (broken root discovery, over-eager exclusion, a regex that stops matching) puts all 181 pinned identities into `missing` and reds. Asserted directly:

```js
const empty = classifyTrackedSites([])
assert.equal(empty.missing.length, empty.approvedIdentityCount)
assert.ok(empty.approvedIdentityCount > 0)
```

A separate test attacks the criterion itself: for every entry, a synthetic site in each formerly-whole-file-approved path with an unapproved symbol must not be claimed; and each entry's own approved rows must not carry a substituted table or verb.

## Additional integrity fix (carried from the earlier attempt, re-verified here)

`pinned-baseline-obligation.cjs` had `genericAllowlist.has(symbolKey) || isGenericSharedAllowlisted(site)` — the second disjunct consulted a **live, HEAD-derived** allowlist inside a proof about a **historical** tree, which is exactly the "live claim crutch" that module's own header forbids. Deleted. The frozen artifact's own `genericSharedAllowlist` is the complete evidence for that ref. Re-verified by **running** the suite on this branch, not by reading the earlier commit message: `pinned baseline obligation: frozen artifact covers the pinned ref without live claim crutches` passes inside the 67/67, and again inside the 74/74 after the gate fixes — the latter being the run that matters, since the P2 domain widening changes which files exist at the pinned ref too. The frozen artifact's `manifestContentHash` is unchanged (`921e0970…`).

## Out of scope

The three inherited pass-throughs the owner separately identified (rich-text SAFE whole-file, D2 executor whole-file, raw-projection/taint `(file, text)` with no multiplicity) are tracked as issue 4838. **Not touched.** The machinery here is directly reusable for them — `siteIdentityKey` plus an `occurrences` column is exactly the shape the raw-projection/taint item needs — but that is a separate ticket and this PR stops here.

**Adjacent finding, reported not changed, and NOT part of issue 4838:** the W4C-4 §12.7 **read** inventories — `calculation-read-classification.cjs:62` and `current-record-read-classification.cjs:65` — still pass through on `site.relPath.startsWith('scripts/')` (and `relPath.includes('/migrations/')`) *before* any classification, so those prefixes bypass their own `count` / `countDrift` / `stale` machinery entirely. Same defect shape as the one this PR fixes, on the read side rather than the DML side, and out of this PR's DML scope. Flagged for the owner to sequence.

## CI — quoted from the `test (20.x)` step log, not from a green tick

**Scope note, so the evidence below is not read as covering more than it does.** The quoted CI log is from head **`3348d31abe`**, i.e. **before** any of the gate fixes, and shows **67** tests. The gate-fix commits have not been reported on by CI at the time of writing; what is quoted for them is the **local** run of the identical command. The census counters are byte-identical to the CI ones, which is the substantive check — the gate measured the same tree, not a smaller one. **The required set must be re-read at the current head before this is treated as landable.**

Local run on the fixed tree, same command CI runs:

```
node --test scripts/ops/attendance-w4c0-dml-inventory-collector.test.mjs
[w4c0-dml-inventory] raw census sites=1220 tracked=212 observedIdentities=181 approvedIdentities=181 unclaimed=0 overCount=0 missing=0 fingerprintDrift=0 genericShared=9 unclassifiedTables=0 outsideBoundary=0
[w4c0-dml-inventory] approved identities=181 pinnedOccurrences=212 repeated=19 genericShared=9
ℹ tests 82
ℹ suites 0
ℹ pass 82
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
```

**82 tests / 82 pass / 0 fail / 0 skipped.** Every pinned census counter is identical to the pre-fix head (`1220 / 212 / 181 / 181`, all failure legs `0`, `181 / 212 / 19 / 9`), with `fingerprintDrift` added to the printed line. Test count 58 → 67 (identity conversion) → **82** (gate fixes): **24 added across the whole PR, zero removed, zero weakened**.

Adjacent suites re-run on this head: `sealed-export-package-provenance.test.cjs` 1/1 (the recomputed s6a pin), `integration-guard-required-wiring-contract.test.mjs` 62/62.

The pre-fix CI evidence follows, unchanged.

Run `31303139311` at head `3348d31abe`, job `test (20.x)` (`93219097844`), step *Run attendance W4C-0 Stage D §8.4 and W4C-4 §12.7 inventory collectors*:

```
node --test scripts/ops/attendance-w4c0-dml-inventory-collector.test.mjs
  shell: /usr/bin/bash -e {0}
From https://github.com/zensgit/metasheet2
 * branch            e0defbe26d7f2e1747e74aa908ca710422812bf7 -> FETCH_HEAD
TAP version 13
# [w4c0-dml-inventory] raw census sites=1220 tracked=212 observedIdentities=181 approvedIdentities=181 unclaimed=0 overCount=0 missing=0 genericShared=9 unclassifiedTables=0 outsideBoundary=0
# [w4c0-dml-inventory] approved identities=181 pinnedOccurrences=212 repeated=19 genericShared=9
...
# tests 67
# suites 0
# pass 67
# fail 0
# cancelled 0
# skipped 0
# todo 0
# duration_ms 57405.692424
```

**67 tests / 67 pass / 0 fail**, and the census counters in CI are identical to the local run (`1220 / 212 / 181 / 181`, all three legs `0`) — so the gate measured the same tree, not a smaller one.

The step carries no `matrix.node-version` guard, so it runs on 18.x too. That was verified from the 18.x step log rather than inferred from the job's green tick — job `93219097804` shows the same `raw census sites=1220 tracked=212 … unclaimed=0 overCount=0 missing=0` line and its own `# tests 67 / # pass 67 / # fail 0`.

All other checks were green on **that** head (`3348d31abe`): `integration-guard`, `web-tests`, `attendance-web-guard`, `contracts (strict|openapi|dashboard)`, `pr-validate`, `after-sales integration`, `stock-prep PowerShell 5.1 acceptance`, `DingTalk P4 ops regression gate`, `K3 WISE offline PoC`. That verdict is head-scoped and does **not** transfer to the gate-fix head; the full required set must be re-read there before this is treated as landable.

## Mechanics

- Based on current `origin/main` (`8d47b5abac0be9b56bd6c543dd1e8e44dd21dc42`), which is already an ancestor of this head — **no rebase was owed** and none was performed.
- **`.github/workflows/plugin-tests.yml` IS now touched — this supersedes the earlier "untouched" claim in this body, which was true only up to head `8385b116fe`.** The change is a **single added line**, `id: attendance-w4c0-dml-inventory`, on the existing attendance collector step; `git diff` on that path is +1/-0. It was required by the CI self-wiring fix: without a stable step id there is no way to assert anything about *that* step rather than about the file's text. The collector test remains named in the `test` job's step with no `matrix.node-version` guard, so it still runs on the required `test (20.x)` check. **An s6a pin recomputation IS therefore owed and is done in the same commit** (see the collision banner at the top). The earlier statement that the pin "lives in the separate windows-latest job" was wrong: `sealed-export-package-provenance.cjs` pins this workflow file by sha256 under `evidenceFiles.pluginTestsWorkflow`, verified by reading that module rather than by inference.
- `scripts/` is not a workspace package, so `pnpm -r lint` / `pnpm -r type-check` do not cover these files. Syntax is covered by construction instead: the collector suite `require()`s every one of these `.cjs` modules, so a syntax error fails the gate rather than passing silently.
- **Other consumers of the widened `collector.cjs`**, checked rather than assumed: `scripts/attendance/generate-cleanup-sql.cjs:95` imports `hasLiveDmlOnTable` (which routes through the same pattern). It is the only other consumer, and it is exercised by the passing `W4C-3c P15: generate-cleanup-sql never emits live DELETE on attendance_records` leg. Widening can only make that guard see *more* DML, which is the fail-closed direction for an assertion that no live DML exists. Re-checked after the gate fixes: `isAttendanceOwnedCandidate` has no consumer outside `collector.cjs` itself, and `hasLiveDmlOnTable` now folds case on both sides, which widens the same absence assertion further in the same direction.
- Never `git add -A`; every commit staged by explicit path. All branch OIDs leased from `git ls-remote`.


zensgit:
Exact-head technical update: head 23910cb215345a0e0d80483e4d3858c65d2d5fee. The raw-SHA fingerprint and lexer hardening are complete; the permanent control-statement-regex regression is included. Collector suite: 84/84 passed. Focused fingerprint and CI-order checks passed. Independent exact-head review: 0 P1 / 0 P2; both prior P1 findings CLOSED. This is technical evidence only and does not authorize Ready, merge, runtime, flags, deployment, soak, production/customer data, or closing #4556.

zensgit:
## Exact-head main-transfer gate

Reviewed head: `fd75c2c29a46c30f2dd506e258f78102fd35a708`

Independent Luna review: **APPROVE-with-hardening, 0 P1 / 0 P2 / 1 P3**.

- The raw exact-site collector remains based on exact `(relPath, enclosingSymbol, table, verb)` identity, exact occurrence multiplicity, and raw statement-fingerprint multisets. No file, prefix, or bare-symbol approval was introduced.
- Full collector rerun: **84/84 pass**. Census: `raw=1223`, `tracked=212`, `observedIdentities=181`, `approvedIdentities=181`; `unclaimed=0`, `overCount=0`, `missing=0`, `fingerprintDrift=0`, `unclassifiedTables=0`, `outsideBoundary=0`.
- The main transfer preserves the PR-owned collector files and brings in upstream changes. Precise P3 wording: the merged workflow also adds the explicit `id: attendance-w4c0-dml-inventory` step; the provenance pin was then mechanically recomputed. Therefore the transfer is not described as "only a hash change."
- Independent SHA-256: `.github/workflows/plugin-tests.yml` and the vector pin both equal `c1a8bc3c0f1827a156c3df993a569806c9acf8bca6a29a8faad3e50300a0c237`.

This PR remains **Draft/HOLD**. Fresh required GitHub checks are a separate gate. This comment does not authorize Ready, merge, flags, staging, soak, deployment, production/customer data, or closure of #4556.

zensgit:
Fresh exact-head GitHub checks have now settled for `fd75c2c29a46c30f2dd506e258f78102fd35a708`: **20 SUCCESS / 0 pending / 0 failed**. The PR remains Draft/HOLD and is now behind newer `main`; a future owner-authorized merge still requires a fresh-main update and renewed head-scoped checks.

PR #4845, opened 2026-08-10
fix(dingtalk): OAuth grant read fails CLOSED — deprovision deny row can't be bypassed by a query blip (review P1)

zensgit:
Post-merge review P1: readGrantEnabled caught any DB error and returned null (same as "no row"); the deny gate is `=== false`, so null slipped through and a deprovisioned user's Rev 4.4 disabled deny row was bypassable by one failed read (DINGTALK_AUTH_REQUIRE_GRANT defaults off). Fix: tri-state 'enabled'|'disabled'|'absent'; a read failure now fails CLOSED (grant_state_unavailable/503). Absent stays allowed so new/auto-linked users still log in. Tests: fail-closed on rejecting read; positive control (absent still logs in); disabled still denies; mutation reverting the catch reddens the fail-closed test. 6791 unit green, tsc green. One of two narrow fixes from the 2026-08-10 review (main=d78b27d37c); switches remain OFF.

### DIFF
--- a/packages/core-backend/src/auth/dingtalk-oauth.ts
+++ b/packages/core-backend/src/auth/dingtalk-oauth.ts
@@ -739,8 +739,18 @@ async function findIdentityUser(dtUser: DingTalkUserInfo): Promise<LocalUserRow 
     )
     return result.rows[0] ?? null
   } catch (error) {
-    logger.warn('Failed to resolve DingTalk external identity', error instanceof Error ? error : undefined)
-    return null
+    // Fail CLOSED (post-merge review P2-2): the same collapse readGrantState fixes lived here
+    // too — a swallowed read error returned null, indistinguishable from "no linked identity",
+    // so a failed identity read let the login fall through to the email-link / auto-provision
+    // path as if the DingTalk person were unknown (defeating the deny row their real identity
+    // carries; the identity unique constraint stopped a full bypass, but it is a latent
+    // fail-open and mints stray grants). "No identity" is still a legitimate null (new user);
+    // only a read FAILURE now throws.
+    logger.warn('Failed to resolve DingTalk external identity; failing closed', error instanceof Error ? error : undefined)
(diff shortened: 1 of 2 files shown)


zensgit:
Absorbed the independent review (APPROVE-with-hardening, no P1) in commit 87f33e81:
- **P2-2**: `findIdentityUser` had the identical swallow-catch fail-open one function up — it now fails closed on a read error too (`grant_state_unavailable`/503), so the class fix covers the whole resolve chain (not just `readGrantState`). "No identity" stays a legitimate null, so new-user provisioning is unaffected. This also removes the P3-1 stray-grant side effect.
- **P2-1**: added identity-direct-path coverage (the deprovision *sweep* writes the deny row without touching identity rows → next login is identity-path). Disabled-denies-with-requireGrant-off, grant-read-fails-closed, and findIdentityUser-fails-closed tests; the identity deny gate / readGrantState catch / findIdentityUser catch mutations each now redden a named test (all survived before).
- **NIT-1**: the "no grant creation" assertions were vacuous (grant INSERT runs on the transaction client) — now assert against the transaction-statement capture.

**Ops note (review P3-2)**: `readGrantState`/`findIdentityUser` are on every DingTalk login, so a `user_external_auth_grants`/`user_external_identities` outage (or an unapplied `zzzz20260324143000` migration) now **503s DingTalk logins** where they previously degraded open — the intended fail-closed trade. Break-glass: `POST /api/auth/login` password login does not read these tables. Also, the web OAuth callback surfaces 503 without the `code` field (container route includes it) — a cosmetic NIT-2, login is denied either way.

zensgit:
Merging: delta re-review APPROVE (no P1/P2) bound to code head; the follow-up NIT is test-only (discriminating call-count). All 9 required contexts pass by name at 74b125d7. Sibling PR #4846 (activation-source FOR SHARE) lands separately — disjoint files.

PR #4846, opened 2026-08-10
fix(dingtalk): T3 activation source read locks against integration deactivation (FOR SHARE) — review P1

zensgit:
Post-merge review P1 #2. The activation transaction locked `users` but read the directory source (account/link/integration) UNLOCKED; under READ COMMITTED an admin deactivating the integration on another connection could COMMIT between the source read and the activation commit, activating a user against a now-inactive source ("admin 激活不得对 inactive 目录源静默开通"). Fix: `FOR SHARE OF ...` on the source read in both branches (integration is INNER-joined in both → locked in both). Load-bearing invariant now commented: integration `status` is checked in TS, not in WHERE, so the FOR SHARE READ COMMITTED EPQ recheck sees the fresh 'inactive' and refuses; moving the predicate into WHERE would make EPQ drop the row. Verification: constructed real-DB pg_locks race (directory-activation-source-lock.db) proving the real activatePendingUser blocks on the deactivation transaction specifically, then refuses ACTIVATE_INTEGRATION_INACTIVE and writes nothing; two-point CI wired (no-DB exclusion + approval real-DB run list) with in-suite self-check; four directory wiring contracts stay green; mutations removing only the `di` lock and the whole clause each redden the golden. tsc + 6788 unit + 41 real-DB green. Also retracts the "可编码项清零" overclaim in the completion MD (§4ter). One of two narrow fixes from the 2026-08-10 review (main=d78b27d37c); switches remain OFF.

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -100,6 +100,7 @@ jobs:
         run: |
           node --test scripts/ops/dingtalk-closeout-env-contract.test.mjs
           node --test scripts/ops/directory-grant-table-ci-wiring.test.mjs
+          node --test scripts/ops/directory-activation-source-lock-ci-wiring.test.mjs
           node --test scripts/ops/directory-deprovision-ledger-ci-wiring.test.mjs
           node --test scripts/ops/directory-deprovision-writer-ci-wiring.test.mjs
           node --test scripts/ops/directory-access-graph-mutex-ci-wiring.test.mjs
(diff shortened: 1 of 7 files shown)


zensgit:
Merging: 9/9 required contexts pass at 6e23ee837b (rebased on main after #4845). The red test (18.x) is not a required check; the failure is in multitable-durable-startup-failclosed.db.test.ts (a durable-delivery boot test) which this PR does not touch, and the same suite passes on the required test (20.x) — an 18.x-only environment flake, not a regression. di-only lock logic APPROVE-with-hardening; round-2 (required-lane wiring contract + comment fix + users-lock positive control) is CI-wiring/test only.

PR #4847, opened 2026-08-10
fix(integration): require exact K3 lookup key match

skiyakubali0-cloud:
## Summary

- normalize the requested K3 material key and each returned `FNumber` by trimming outer whitespace while preserving case
- discard successful GetDetail records whose key does not exactly match the requested key
- preserve all exact matches so duplicate evidence still reaches the planner as `held`
- add regression coverage for mismatched → `add`, matched → existing (`update`), and duplicate matches → `held`

## Why

The controlled `_02` preview in #4628 returned `add=0/update=2` even after both operation-private source keys were rotated. The frozen C6 lookup passed every successful GetDetail record to `classifyExisting`, so an unrelated default/template detail record was treated as proof that the requested key already existed.

This patch makes returned-key equality the evidence boundary. It does not change Save-only locks, Submit/Audit disablement, row caps, B4 binding, token handling, transport/login behavior, or Apply behavior.

Related to #4628. The issue must remain open for owner review, package refresh, and controlled retest.

## Verification

- `node __tests__/k3-wise-c6-write-profile.test.cjs` — 32/32 pass
- `node __tests__/k3-wise-adapters.test.cjs` — pass
- `node __tests__/external-write-dry-run.test.cjs` — pass
- `node __tests__/k3-write-approval-posture.test.cjs` — 8/8 pass
- `node --check` on both changed files — pass
- `git diff --check` — pass

`k3-wise-apply-row-limit.test.cjs` passes 24/25 locally; its existing sweep-contract assertion compares POSIX `/` paths with Windows `\` paths and is unrelated to this diff.

## Operational boundary

No code from this PR has been deployed to the controlled host. No K3 Apply, Submit, or Audit was executed. A reviewed merge and new frozen package are still required before the onsite retest.

### DIFF
--- a/plugins/plugin-integration-core/__tests__/k3-wise-c6-write-profile.test.cjs
+++ b/plugins/plugin-integration-core/__tests__/k3-wise-c6-write-profile.test.cjs
@@ -123,7 +123,7 @@ function jsonResponse(status, body) {
 
 // Mock K3: login, GetDetail (existing row for MAT-C6-EXIST, business-fail for everything
 // else — K3's real "not found" shape), Save success. Counts every path.
-function mockK3({ existing = {} } = {}) {
+function mockK3({ existing = {}, fallbackDetail = null } = {}) {
   const calls = []
   const impl = async (url, init) => {
     const parsed = new URL(url)
@@ -141,6 +141,13 @@ function mockK3({ existing = {} } = {}) {
           StatusCode: 200,
           Message: 'Successful',
           Data: [{ FStatus: true, FItemID: row.FItemID, Data: row }],
+        })
+      }
+      if (fallbackDetail) {
+        return jsonResponse(200, {
+          StatusCode: 200,
+          Message: 'Successful',
+          Data: [{ FStatus: true, FItemID: fallbackDetail.FItemID, Data: fallbackDetail }],
         })
       }
       // K3's business-level miss: FStatus false — businessSuccess() is false for this shape.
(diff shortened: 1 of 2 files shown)


zensgit (approved):
APPROVE at exact head `6f6e63d66448ba3759a85f07618ff5d1828c5fd5`.

No P1/P2 findings in the bounded lookup-key correction.

Independent evidence:

- reviewed the two-file diff and the live K3 detail normalization path;
- `k3-wise-c6-write-profile.test.cjs`: 32/32 pass;
- `k3-wise-adapters.test.cjs`, `external-write-dry-run.test.cjs`, and
  `k3-write-approval-posture.test.cjs`: pass;
- removing the returned-key filter makes exactly the new mismatch regression
  fail;
- an uncommitted positive probe drove the mismatch path through dry-run and
  apply recomputation: the revision remained stable and one Save completed;
- transport/login failures remain outside the business-miss branch and still
  fail closed;
- a successful K3 detail response without `FNumber` is normalized by the
  underlying adapter with the requested material number before this layer, so
  it does not silently become an empty lookup here.

Merge remains conditional on all nine required checks, including the currently
running `test (20.x)`. After merge, build and verify a new exact package from
the merge SHA; do not patch the deployed `aa48` archive in place. The `_02`
authorization must then be rebound to that new runtime/package identity before
the onsite preview is repeated.

Operational hardening: `add=2` is necessary but not, by itself, universal
proof of absence because the pre-existing business-error branch intentionally
maps some K3 business refusals to an empty lookup. Before apply, K3_ADMIN must
also privately confirm that both operation-private high-entropy keys are absent
in the controlled K3 instance. No key values belong in GitHub evidence.

PR #4832, opened 2026-08-08
[HOLD] fix(attendance-w6-1): rebuild the W6-1 group effective-policy aggregate from fresh main — issue 4556 (Draft, no merge authorized)

zensgit:
Superseded by #4849, which rebuilds this work from fresh `main` after security review.

This description has been replaced. Any technical detail relating to the review of this change is
tracked in the repository's private security advisory, which is the single source for it.

No claim is made here that earlier revisions of this pull request are unreachable.

PR #4804, opened 2026-08-07
docs(attendance): #4556 W7/W8 design locks (PROPOSED)

zensgit:
## Maintained current state (2026-08-11)

- **Exact head:** `8898b8ac7e859e45b3a32ee7c5866ad2f121e635`; **frozen base:** `main@325917c0a484522ef9ce87b286d5d986d4e205b3`.
- **Fresh-main transfer:** the owner renewed the freeze against `325917c0a4…` after unrelated ops PRs #4868/#4869/#4870/#4872 advanced `main`. The branch takes that exact commit as a merge parent. The PR-owned delta remains exactly the two W7/W8 docs; runtime/product/workflow delta is empty.
- **Mechanical provenance refresh:** 19 current-main references were changed from `6bee1e88…` to `325917c0…`. The intervening main delta is byte-empty over the W6/runtime evidence domain and the attendance real-DB workflow/corpus/guard domain. No decision choice, ledger verdict, or authorization boundary changed.
- **Parent debt ledger:** 43 expected unique IDs = **13 SATISFIED / 1 OPEN / 29 UNRESOLVED / 0 REMOVED**, with no blank status or duplicate ID.
- **Decision menu:** W7 remains `OD-W7-0..10`; W8 still requires owner answers for `OD-W8-1`, `-2`, and `-4..8`; `OD-W8-3` remains SATISFIED.
- **W6 provenance:** current clean-rebuild candidate is #4849 at `c2ac8284ab38756133c74ce54fde50d07d5bf13a` (OPEN/Draft/HOLD/unmerged). W6 runtime is not on `main`; #4804 authorizes neither #4849 nor any W6/W7/W8 runtime.
- **Gate C provenance:** #4839 remains merged as `60afbffe07bfddc7f32ff08549e36e995662b228`; its tool has not been executed and no rollout/staging/soak authority follows.
[...]

PR #4849, opened 2026-08-10
[HOLD] feat(attendance-w6-1): group effective-policy aggregate — clean rebuild from fresh main (issue #4556)

zensgit:
[HOLD] Draft. `main` continues to move — measured at `0bb96d6f316be1c0cec68eed235ff23a1d01ff72` as of the prior update. No catch-up merge/rebase onto main, Ready transition, or merge of this PR will occur without renewed owner authorization. No W6-2/3/4 runtime, flag change, deployment, staging, soak, production/customer data use, or closure of issue #4556 is authorized by anything in this PR.

## Scope

This PR is the clean W6-1 backend aggregate.

- Endpoint: `GET /api/attendance/groups/:groupId/effective-policy`
- Permission: `attendance:admin`
- Organization: authenticated principal only; optional query/body/header `orgId` is a byte-equality assertion, never a scope source
- Delegated admins: active membership in the target organization is required
- Transaction: post-guard platform-admin lookup, membership check, aggregate reads, and FSER reads use one transaction-bound handle after PostgreSQL `SET TRANSACTION READ ONLY`
- Response: exact-key, values-free, closed enums and editor references
- FSER: composes the existing service; no second derivation or persisted status
- Conflict behavior: count-only reporting, no winner selection
- Runtime loading: closed-file, repo-root-anchored, symlink-rejecting resolver
- Compatibility: zero persisted segment rows keep the W3 legacy-envelope single-segment reading
- CI: all three W6 real-DB suites have two-point wiring; workflow provenance is mechanically pinned
- **Index-assembly guard:** the source-text guard that proved this route sits behind the global JWT gate and the attendance audit/security middleware was fail-open on three real evasions (comment-out, `if (false)` wrapping, an extra pre-gate mount) — it has been replaced with an AST registration-site model. A round-2 follow-up (below) closed a residual completeness gap in that AST model itself.
- **Client-input boundary:** the org-selector-mismatch and query/body-rejection guards at the top of the handler previously had zero DB-free unit coverage — only the real-DB suite exercised them, entirely behind `describeIfDatabase`. DB-free coverage (10 negatives, 4 positive controls) has been added; round 2 (below) extends this same DB-free discipline to the org-identity DERIVATION step that runs just before it.

W6-2 OpenAPI publication, W6-3 UI, W6-4 whole-W6 closeout, W7, and W8 are not part of this PR.

## This update — F1 (assembly guard) and F1b (client-input boundary coverage)

Two proposed designs for the assembly-guard replacement were reviewed; the shipped design took Design B's whole-file registration-site/escape model (`R ∪ C` closure — this is what makes the plugin runtime's computed `this.app[verb](...)` dispatch visible as a pinned census entry instead of silently unmodelled) as the base, and grafted Design A's corrected "unconditional" predicate (a whitelist — the call's own immediate parent must literally be the `ExpressionStatement` of the method body, zero hops — not an ancestor-kind blacklist, which Design A's own two confessed draft failures showed does not converge on `cond && this.app.use(x())`-shaped short-circuit registrations).

**F1 — `tests/helpers/attendance-w6-index-assembly-order.ts` replaced entirely.** The two named root causes (raw-text `indexOf` counting a comment as an occurrence; `getFullStart()` resolving a position inside a comment's leading trivia into the wrong node) are now structurally impossible: no text offset is ever computed, and no offset is ever mapped back to a node. Live mutation proof against the real `packages/core-backend/src/index.ts` (every mutation restored by `cp` from a backup; `git status --porcelain` empty after each; never `git checkout --`):

| mutation | old guard | new guard |
| --- | --- | --- |
| comment out the audit registration | GREEN (fail-open) | RED |
| wrap it in `if (false) { … }` | GREEN (fail-open) | RED |
| insert an extra `this.app.use('/', attendanceAdminRouter())` before the gate | GREEN (fail-open) | RED |
| relocate the route registration to before the gate (ordering-only, no duplication) | RED | RED (still) |
| duplicate the security-middleware registration verbatim | RED | RED (still) |

One over-pinning bug was caught and fixed mid-round: the app-object census initially pinned `line` numbers, which made it fragile to any unrelated one-line edit above those points — dropped, since array order already encodes position; a mutation re-run after the fix showed the census correctly ignoring an unrelated shift while still reding on the real evasion.

The test's own claim (`attendance-w6-group-effective-policy-authorization.test.ts`, "Guarantee 4" section) is corrected to state exactly what is and is not proven, including the gate subject's honest, weaker claim: the callback containing a call to `jwtAuthMiddleware` is registered unconditionally and first — not that any given request reaches it. A five-call behavioural shrink (real `isApiPath`/`isWhitelisted`/`isPublicFormAuthBypass`/`isOapiAllowlistRequest`, real route path, no fabricated request object) narrows that gap further.

**F1b — DB-free coverage added to the same test file** for `attendance-admin.ts`'s org-selector-mismatch predicate (`:177-192`) and query/body rejection (`:1748-1761`): 4 mismatch negatives (query/body/header, plus the documented present-but-empty `x-org-id` edge case), 2 matching-value positive controls, 2 query-rejection negatives, 2 body-rejection negatives (both distinct `BODY_NOT_ACCEPTED` branches), 1 empty-body positive control. Each negative proven discriminating by neutering the corresponding production predicate/branch and confirming the specific test(s) RED, one branch at a time, restored by `cp` each time.

## This update, round 2 — P2 (F1 guard completeness) and F1b extension (org-identity derivation legs)

An independent gate on round-1's head (`4d9b54e94a7b7bbf979dfdcdfc8f2fa2d70c0a2f`) found the F1 guard's own docblock making a false completeness claim — the exact class F1 itself was fixing — plus a P3 gap in F1b's DB-free coverage. Both are closed in this round; zero production code is touched, both fixes are test-only.

**P2 — the docblock claimed MORE than the code delivered, and one of the two shapes the gate named was already false (verified empirically, not assumed):**

- `const a = this.app; a.use(...)` before the gate — the gate's own prescribed proof — was RE-VERIFIED against the round-1 code UNCHANGED and already REDs A4 (the escape collector's `siteCalleeExpressions` exclusion set only ever holds genuine site callees, so an alias's `this.app` read already fell through to the generic `ESCAPE` branch). The gate's diagnosis of this specific shape was incorrect; its conclusion that the docblock overclaimed was not — for a different, real reason, found independently here:
- `this['app']` (string-literal bracket notation for the SAME property) and direct destructuring of `this` (`const { app } = this`, `const { app: x } = this`, `const { ...rest } = this`) produced **zero sites and zero escapes** — genuinely invisible to A4, not merely mis-classified.

Both newly-closed shapes are now collected into `model.escapes` (never promoted to `model.sites` — the site definition stays dot-notation-only, so A1/A2/A3/A5 and every existing pin are byte-identical against real `index.ts`, unmutated). A genuinely undecidable family is named as an explicit residual in the module's docblock rather than claimed closed: a non-literal bracket index on `this` (`this[computedExpr]`), destructuring on the LEFT side of a plain assignment (`({ app } = this)`, a structurally different LITERAL-expression node), and `this` itself escaping bare into `Reflect.get`, object spread, `.bind`/`.call`/`.apply`, or a `Proxy` — an unbounded family that does not converge by enumeration.

Mutation proof against real `index.ts` (every mutation restored by `cp`, `git status --porcelain` empty after each):

| mutation (inserted before the gate) | A1/A2/A3/A5 (real registration facts) | A4 (app-object census) |
| --- | --- | --- |
| `const a = this.app; a.use('/', attendanceAdminRouter())` | unaffected (still green) | RED (7th entry) |
| `this.app['use']('/', attendanceAdminRouter())` | unaffected (still green) | RED (7th entry) |
| `const a = this['app']; a.use(...)` | unaffected (still green) | RED (7th entry, newly closed) |
| `const { app: a } = this; a.use(...)` | unaffected (still green) | RED (7th entry, newly closed) |

The six prior evasions this guard already caught (comment-out, `if (false)`, never-true env condition, `&&` short-circuit, ternary, `setImmediate`) plus the pre-gate relocation and the delete-line positive control were all re-run against this round's code and still RED, each on the exact same assertion as round 1 (A1 and/or A3/A5, never A4) — none of this round's additions changed their behavior. Nine new fixture-level regression tests (independent of `index.ts`'s own drift) pin the closed shapes permanently, including negative discriminators (`this['foo']`, an unrelated destructured property, a non-literal bracket index) proving the predicate matches the literal string `"app"`, not every bracket access or destructure on `this`.

**F1b extension — org-identity derivation legs.** `getAuthenticatedAttendanceGroupEffectivePolicyOrgId` (`attendance-admin.ts` ~L157) derives org identity through five sources tried in order: `user.orgId` (string), `user.workspaceId` (string, only when `orgId` is strictly null/undefined), `user.orgId` (finite number), `user.workspaceId` (finite number), and `req.authenticatedTenantId` (string, terminal fallback). Before this round, only the `orgId`-string leg was exercised anywhere in this repository, DB-free or DB-gated — grepping every `.db.test.ts` file that exercises this route for `workspaceId` or `authenticatedTenantId` returns zero hits. Nine new DB-free tests add one positive control per leg, a precedence check (orgId wins when both orgId and workspaceId are present), and three negatives (terminal no-identity-anywhere, a present-but-empty orgId correctly NOT falling through to workspaceId, a wrong-typed orgId behaving the same way). Each of the four production branches was neutered one at a time and reds exactly its paired leg's positive control, nothing else — confirmed, not assumed.

**Self-caught follow-up within round 2:** the precedence test's first version asserted only a 200 + `groupId`, which is not discriminating — the DB-free mock's `respond()` switches on SQL TEXT, not parameter VALUES, so it would have returned the same canned membership row regardless of which org actually reached the query. Replaced with a version that captures the membership SQL's actual `org_id` parameter and asserts it equals `orgId`'s value, never `workspaceId`'s. Verified by reversing the production precedence (`workspaceId ?? orgId`) and confirming the corrected test now reds on the expected-vs-received org value — the original version stayed green under that same reversal. No test count change (69 in this file; 358 in the 12-file matrix) — this corrects rigor, not coverage.

Local evidence at this round's final head (`796b5ea3e6c0282e06eaee744d7389bde0d033ae`) is in "Fresh local verification" below.

## This update, round 3 — P2 (F1 guard receiver-wrapper blindness)

An independent gate on round-2's head (`796b5ea3e6c0282e06eaee744d7389bde0d033ae`) found the receiver predicates themselves — `isThisAppPropertyAccess`, `isThisAppElementAccess`, and the destructure-initializer check — gated on `node.expression.kind === ts.SyntaxKind.ThisKeyword` **directly, with no unwrap**. A `this` receiver wrapped in a single no-op node evaded BOTH the site collector and the escape collector entirely: `this!.app.use(...)`, `(this).app.use(...)`, `(this as any).app.use(...)`, `(this as MetaSheetServer).app.use(...)`, and `this!['app'].use(...)` were each `{sites: 0, escapes: 0}` — invisible to A4 AND A5 alike. Live-verified against real `index.ts`: a `(this as any).app.use('/api/x', unauthedRouter())` inserted before the JWT gate left the guard fully green — a pre-authentication fail-open shape an ordinary TS refactor emits. Zero production code touched; test-only, same as rounds 1-2.

**Closed with `unwrapNoOpWrappers`** — a CLOSED, finite set of wrapper kinds decided purely by `node.kind` (`ParenthesizedExpression`, `NonNullExpression`, `AsExpression`, `TypeAssertionExpression`, `SatisfiesExpression`), stripped to a fixed point so composed wrappers (`((this as any)!)`) fully unwrap too. Applied via one shared predicate (`hasUnwrappedThisReceiver`) in all three receiver checks, **plus `findThisMethodCalls`** (`this.<methodName>()`), which had the identical unwrapped check — an ADDITIONAL wrapped `this!.setupMiddleware()` would otherwise stay invisible while "exactly one call site" kept asserting true. `unwrapParens` (a strict subset) is removed as redundant.

**Deliberately NOT widened, verified by fixture rather than assumed:** a wrapper around the WHOLE `this.app` VALUE, after the property read (`(this.app).use(...)`, `this.app!.use(...)`), is a structurally different position — `callee.expression` there is the wrapper node, not a `this.app` `PropertyAccessExpression` at all — and stays escape-only, exactly as an unwrapped bare-alias read (`const a = this.app`) already did. That split is asserted explicitly in the fixture suite, not left untested, so the module docblock's claim stays true rather than silently widened.

Live mutation proof against real `packages/core-backend/src/index.ts` (every mutation restored by `cp` from a backup kept OUTSIDE the worktree; `git status --porcelain` confirmed empty after each; never `git checkout --`); a leading `;` on every inserted statement, since this codebase's no-semicolon style otherwise lets a `(`-initial statement ASI-merge into the PRECEDING line — caught live when a first pass without it corrupted an unrelated site's `unconditional` classification instead of producing a clean result, a bug in the harness, not the guard, diagnosed and fixed before any conclusion was drawn from it:

| mutation (inserted before the gate) | before this round | after this round |
| --- | --- | --- |
| `this!.app.use('/x', filesRouter())` | green (invisible) | RED — A5 (13-entry pre-gate prefix) |
| `(this).app.use('/x', filesRouter())` | green (invisible) | RED — A5 |
| `(this as any).app.use('/x', filesRouter())` | green (invisible) | RED — A5 |
| `(this as MetaSheetServer).app.use('/x', filesRouter())` | green (invisible) | RED — A5 |
| `this!['app'].use('/x', filesRouter())` | green (invisible) | RED — A4 (7th escape entry, bracket notation stays escape-only per the module's existing site/escape split) |
| `const a = (this as any).app; a.use('/x', filesRouter())` | green (invisible) | RED — A4 (7th escape entry) |

The SITE model stays byte-identical on real, unmutated `index.ts`: **77 sites / 6 escapes**, identical before and after this round's code change (index.ts itself contains no wrapped receivers). The prior E3 (plain pre-gate relocation) and delete-line positive control were re-run live against this round's code and still RED on the same A1/A3/A5 assertions as rounds 1-2; the six named evasions (E1 comment-out, E2 `if(false)`, E2b never-true env, E4 `&&`, E5 ternary, E6 `setImmediate`) and round 2's alias/bracket/destructure shapes are all still green via their 26 permanent fixture-level regression tests — none of this round's changes altered their behavior.

**23 new fixture-level regression tests** (index.ts-drift-independent): six wrapper kinds proven as SITEs (`this!`, `(this)`, `this as any`, `this as <NamedType>`, `<any>this`, `this satisfies X`, plus one composed/nested case); wrapped bracket-notation and wrapped-destructure ESCAPEs with the SAME emitted signatures as the unwrapped forms (printed once via a throwaway probe, then pinned — not guessed); the alias-of-wrapped-`this.app` ESCAPE (the exact P2 shape); three negative discriminators (a wrapped receiver reading a DIFFERENT property is neither a site nor an escape, in property-access, bracket, and alias-adjacent form); the documented site/escape split for whole-value wrapping (2 fixtures); and four `findThisMethodCalls` fixtures including a regression proving an ADDITIONAL wrapped call site is counted, not silently missed. Mutation self-check: reverting the helper to its pre-fix state reds exactly the 17 tests that depend on the fix; the 6 fix-independent negative/split/mismatch controls correctly stay green either way — not a blunt "everything reds" signal.

Module docblock updated to state the receiver-normalization claim precisely (including the site/escape split) and to confirm the residual list now contains only genuinely non-decidable shapes — non-literal `this[x]`, assignment-form `({app} = this)`, and `this` escaping bare into `Reflect`/spread/`bind`/`Proxy` — none of the wrapper forms just closed.

Local evidence at this round's final head (`0361cc8a6b152570f61340792a4bda9b9231bb95`) is in "Fresh local verification" below.

## Current lineage and hold

- Last owner-authorized base: `24794811b1c800402006b30d6e4fa9df670e124e`
- Current branch head: `0361cc8a6b152570f61340792a4bda9b9231bb95`
- Current remote main (measured fresh at the prior update): `0bb96d6f316be1c0cec68eed235ff23a1d01ff72` (this round did not re-measure main; no catch-up merge/rebase is performed or proposed by this round either)
- **RATIFIED main SHA citable as authority: `1e0d451a25ccf5b66a6b96cd992233e0f74e8d16`** (#4876, "[PROPOSED] docs(attendance): align W6 delegated access status") — the owner RATIFIED this exact merged SHA on 2026-08-12. It is an ancestor of the main measured above. See "Governing-record status" below.
- Delta census against the last authorized base: unchanged from the prior update at the file-count level (this round edits the same two test/helper files further; no new file added or removed)
- PR state: Draft/HOLD; catch-up to current main is not yet authorized

Retired PR #4832 remains closed. This PR does not claim that historical GitHub objects became unreachable; residual disclosure handling stays in the existing private advisory and is not repeated here.

## Fresh local verification (this round's final head: `0361cc8a6b152570f61340792a4bda9b9231bb95`)

- Full core-backend unit suite (`vitest run tests/unit`, unscoped — not just the 12-file matrix): **489 files / 7077 tests passed**, zero failures
- W6 unit/guard matrix (`tests/unit/attendance-{admin-plugin-lib-dist-layout-boot,plugin-lib-resolver-hardening,w6-fser-single-source-caller-inventory,w6-group-effective-policy-aggregate,w6-group-effective-policy-authorization,w6-group-effective-policy-dml-sweep,w6-group-effective-policy-response-contract,w6-import-graph-no-calculation-consumer,w6-producer-key-single-source,w6-schedule-route-surface-parity}.test.ts` + `tests/unit/api-path-policy{,.guard}.test.ts`, the same 12-file command as every prior verification in this PR): **12 files / 381 tests passed** (up from 358 — this round's net 23 new receiver-wrapper fixture tests, minus zero removed)
- Local PostgreSQL: a freshly created, uniquely-named scratch database, migrated from zero with this head's own `src/db/migrate.ts` (not reused from any prior session): all migrations applied cleanly, then dropped after use
- Local PostgreSQL W6 matrix against that fresh database (`vitest.integration.config.ts`, `tests/integration/attendance-w6-group-effective-policy.db.test.ts`): **1 file / 43 tests passed**, unmutated, confirming this round's test-only change did not regress the real-DB suite (this round touched no DB-facing production code, so this is confirmatory, not required)
- The prohibited files are confirmed byte-identical to this round's starting head (`796b5ea3e6c0282e06eaee744d7389bde0d033ae`) by blob-hash comparison, not by inspection: `packages/core-backend/src/index.ts` → `51947f3b3d4c48b94a4781cfcda66ab2ba596b9a5a9336a4c9823f5b19a83cc5`; `.github/workflows/plugin-tests.yml` → `b689c385336cdc7c05d77086f9b6b147f7f40b5d2d9c3c48a1593e6c561585d6`; `plugins/plugin-integration-core/lib/sealed-export/vectors/s6a-package-provenance-pins.json` → `5daa6861cf6fe8ab8bd031958063bd7f8093fc516a906b1775f8040f27080087` (also checked: `plugins/plugin-attendance/index.cjs` and `apps/web/src/views/AttendanceView.vue`, both unchanged)
- Core backend TypeScript: **zero diagnostics** — `npm run type-check` (`tsc --noEmit` against `tsconfig.json`) exits 0 with no output. That config's own `include` is `src/**/*`, `core/**/*`, `types/**/*` only — it never includes `tests/**`, so it structurally cannot regress from the changed file in this round; this is a scope fact about the script, not a claim that it exercised the file.
- Mutation self-check on the fix itself: reverting `attendance-w6-index-assembly-order.ts` to its pre-round-3 state and re-running just the new `describe('receiver-wrapper normalization...')` block reds exactly 17/23 tests (the 6 fix-independent negative/split/mismatch controls correctly stay green either way), then restored by `cp`

Rounds 1-2's own verification numbers (12 files / 339 then 358 tests, the PostgreSQL migration/W6-matrix counts, the CI-wiring-guard and sealed-export-provenance results, the `EXPECTED_ATTENDANCE_SUITES` census) are unchanged by round 3 and are not re-quoted here; see the round-1/round-2 sections above and this PR's history for those.

The first independent gate against older head `a5738017edd2698db5fbda6a2aa73011ad5ba461` found two technical P2s and one governance P2. The technical pair was closed in `413271e2d7a14aa21a9c5be48001d9c15c432b5d`:

- query/body/header organization selectors share one byte-equal-or-403 rule while unrelated state selectors retain typed 400 rejection;
- a dedicated disposable PostgreSQL suite reproduces all eight committed fixtures from seeded rows with the canonical FSER and exact-key equality.

The next gate at `4d2a9b217880290d33a92b9508e14ba469fca3f0` found a required-test P1: the new fixture-matrix test legitimately loaded canonical FSER but its exact file identity was absent from the FSER test-caller inventory. `d951b9e3da27a2b07615c5139c7e64edf656d104` adds only that exact test file to the closed set; production composition pins are unchanged. The inventory then passed 8/8 and the full matrix passed 300/300 (round 1: 339/339; round 2: 358/358 — see "Fresh local verification" above).

The gate on round 1's head (`4d9b54e94a7b7bbf979dfdcdfc8f2fa2d70c0a2f`) found the P2 and F1b-P3 closed by round 2, above.

The gate on round 2's head (`796b5ea3e6c0282e06eaee744d7389bde0d033ae`) found the receiver-wrapper P2 closed by round 3, above.

All reviews and checks against a head before the eventual authorized catch-up are discovery evidence only.

## Governing-record status — RATIFIED (previously: unresolved contradiction)

The RATIFIED W6 lock's normative **W6-R3 and endpoint §4.1** have required missing, cross-org, and inaccessible groups to share the values-free 404 shape since the lock's original 2026-08-08 ratification — this PR's runtime and tests have followed that rule from the start. A separate part of the same document, its §7.2 completion-skeleton matrix line, instead said 403 for a delegated attendance admin without active target-org membership — a wording contradiction with W6-R3/§4.1, not a runtime one.

`1e0d451a25...` (#4876) added **§10**, a narrow, now-RATIFIED correction that fixes the §7.2 line to say 404 (matching W6-R3/§4.1) and moves the OpenAPI draft's delegated-non-membership case out of its 403 description into the shared 404 description. §10's own text states plainly: *"No runtime behavior changes here. W6-R3 and §4.1 already selected 404, and the W6-1 candidate is reviewed against that existing governing behavior."*

**Authority for this PR's 404 behaviour is therefore W6-R3/§4.1 — RATIFIED since 2026-08-08 — not §10.** §10 is a confirming correction to the DOCUMENT's internal consistency (fixing the contradicting §7.2 wording), not the source of authority for behavior this PR's runtime already had before §10 existed. §10 being RATIFIED removes the wording contradiction as a blocker; it does not retroactively become why the 404 behavior is correct.

One inconsistency worth naming rather than papering over: the design-lock document's own status header (as merged in `1e0d451a25...`) still reads *"The narrow delegated-non-member status correction in §10 is **PROPOSED** pending owner RATIFY of this correction's exact merged SHA"* — that header text has not itself been amended by a further commit. This PR body's "RATIFIED" characterization above reflects owner authorization conveyed to this development session on 2026-08-12 citing the exact merged SHA `1e0d451a25...`; it is not a claim that the document's own inline text has been updated to match. Amending that header is a separate, small, un-granted docs action, not performed by this update.

## Durable reports

- `docs/development/attendance-issue-4556-w6-1-development-report-20260812.md`
- `docs/development/attendance-issue-4556-w6-1-verification-report-20260812.md`

Both predate this PR's updates and are unchanged; their point-in-time numbers are historical and superseded by "Fresh local verification" above where they overlap. Neither report is rewritten by this update — this PR body is the current source of truth for post-F1/F1b/round-2 evidence.

## Governing record

`docs/development/attendance-issue-4556-w6-group-effective-policy-design-lock-20260805.md`

No later slice or runtime action is authorized.

Refs #4556.


zensgit:
## Exact-head W6-1 remediation gate

Reviewed head: `c2ac8284ab38756133c74ce54fde50d07d5bf13a`

Independent Sol review: **APPROVE, 0 P1 / 0 P2 / 0 P3**.

- The authorization claim is now scoped to the post-guard platform-admin lookup; it no longer overclaims that the outer `rbacGuard` reads are inside the aggregate transaction.
- The real-DB leg derives both aggregate and FSER SQL, requires a non-vacuous execution from each domain, and compares every derived SQL occurrence across the all-path recorder and the single transaction recorder. A targeted mutation that moved only FSER reads back to the pool made this leg fail; restoration passed.
- Fresh verification after non-destructive sync with `origin/main@f0745831fe`: W6 unit/static **171/171**, real PostgreSQL **48/48**, backend `tsc --noEmit`, sealed-export provenance test, and `git diff --check` all pass. The merged workflow SHA-256 and pinned value both equal `703dba5add0aa236c47698504356412faa2cb9404ed640cc3efc5ee24488fb1e`.

This PR remains **Draft/HOLD**. Required GitHub checks triggered by this push are a separate gate. This comment does not authorize Ready, merge, W6-2/3/4, flags, staging, soak, deployment, production/customer data, or closure of #4556.

zensgit:
Fresh exact-head GitHub checks have now settled for `c2ac8284ab38756133c74ce54fde50d07d5bf13a`: **all required contexts SUCCESS**, with the established intentional skip only, and **0 pending / 0 failed**. The PR remains Draft/HOLD. This is a technical gate result, not Ready or merge authorization.

zensgit:
[FINAL EXACT-HEAD GATE / DRAFT-HOLD]

Exact head: `2674d379ef6df8a0a64a67bf94cac1f04b226b62`
Fresh-main base: `0287b250b33fe4c7ea98b880360af74fc08a5ebf` (the #4804 merge; still current main when this gate closed).

Evidence:
- Exact-head local W6 matrix: 12 files / 297 tests PASS.
- TypeScript `tsc --noEmit`: PASS with zero diagnostics.
- CI wiring guard: 221/221 PASS; sealed-export provenance: 1/1 PASS.
- Isolated PostgreSQL 15.17, synthetic data only: 2 files / 48 tests PASS at runtime-equivalent implementation content `54c23a4dcd8539d126c3654bed92b487af7247da`; the subsequent exact-head delta is test/report evidence only.
- Fresh GitHub run `31513852860`: 24 SUCCESS, 1 expected SKIPPED, no pending or failed checks.

Independent exact-head gates:
- Kimi K3: APPROVE, 0 P1 / 0 P2. Its earlier P3 (a hand-written RBAC SQL clone in the test) was fixed by delegating the DB-backed legs to production `isAdmin`, then re-reviewed.
- Grok 4.5: APPROVE, 0 P1 / 0 P2 / 0 P3.
- GPT-5.6 Sol: APPROVE, 0 P1 / 0 P2 / 0 P3; independently ran 12 files / 297 tests.
- GPT-5.6 Terra: APPROVE, 0 P1 / 0 P2 / 0 P3; independently ran 12 files / 297 tests.

Durable reports in this PR:
- `docs/development/attendance-issue-4556-w6-1-development-report-20260812.md`
- `docs/development/attendance-issue-4556-w6-1-verification-report-20260812.md`

Boundary: this closes the authorized fresh exact-head gate only. PR #4849 remains Draft/HOLD and unarmed. This record does NOT authorize merge, later W6 runtime slices, flags, deployment, staging, soak, production/customer data, or closing #4556.

zensgit:
[GATE RESET / DRAFT-HOLD]

The first fresh gate at `4d2a9b217880290d33a92b9508e14ba469fca3f0` correctly found a required-test P1: the new seeded fixture-matrix suite loaded canonical FSER but its exact test-file identity was absent from the FSER caller inventory. That head's required `test (20.x)` failed, so it is not an approvable head.

The narrow repair is now on branch head `87236489c7a4e85594c1b91e8115e4d5bdf6443a`: exact test caller added to the closed set; no directory exemption or production caller-pin change; targeted inventory 8/8 PASS; complete W6 unit/guard matrix 12 files / 300 tests PASS. Reports preserve the failed-head history. Independent discovery review found no new P1/P2 in the narrow repair; this is not a final gate.

Discovery CI for `87236489c7a4e85594c1b91e8115e4d5bdf6443a` has now fully settled:
- the nine required contexts are all SUCCESS, with 0 pending / 0 failed;
- both Node 18 and Node 20 jobs are SUCCESS;
- the Node 20 lane executed core-backend tests, migrations, the shared real-DB matrices, web build, and the attendance integration run before completing successfully;
- the wider check suite is green, with only the established non-required strict-E2E skip.

While that repair was being pushed, `main` advanced from authorized base `24794811b1c800402006b30d6e4fa9df670e124e` to `979c619ebf0ca1dfadedff2dc9b8db69b4f6b74c` via unrelated #4875. Its three DingTalk-canary files have zero path overlap with this PR, but the exact-base authorization is stale. Therefore these green checks remain discovery evidence: this PR stays Draft/HOLD and will not catch up, become Ready, or merge without renewed owner authorization.

Governance P2 remains: RATIFIED W6-R3/section 4.1 and runtime require values-free 404 for delegated-inaccessible groups, while section 7.2 and the unpublished OpenAPI draft still say 403. This needs an owner-authorized PROPOSED correction, merge, and exact-SHA RATIFY before the final 0 P1/P2 gate.

No W6-2/3/4 runtime, flags, deployment, staging, soak, production/customer data, or #4556 close action is authorized or performed.

PR #4851, opened 2026-08-10
feat(ops): add fail-closed DingTalk lifecycle staging lane

zensgit:
## What
- add a staging-only manual lifecycle operator lane
- permit only read-only `status`/`preflight` and emergency `off`; ON actions fail closed
- require exact image+health SHA agreement, pending migrations zero, backend health, strict current mode, and pinned SSH host identity
- persist honest UAT/T2-Gate/canary boundaries and current staging evidence

## Live evidence
- staging migration runner 31407444155: backup + clone rehearsal + isolation + real apply, 296/18 -> 314/0
- running image is b55c682748e3010cb70837770c298843a96e1019, but health metadata reports 59c24a1d21cfc70b76867da7d0ac15590d558c72; all write actions remain blocked by `build_provenance_conflict`
- lifecycle ON canaries remain NOT EXECUTED; all three flags remain OFF

## Verification
- `bash -n scripts/ops/dingtalk-lifecycle-staging-canary-remote.sh`
- `node --test scripts/ops/dingtalk-lifecycle-staging-canary-contract.test.mjs` (34/34)
- contract wired into required plugin-tests job
- Kimi K3 adversarial review: no P1; three P2 fixed (host pinning, dual-source SHA, real-function/order tests)

## External gates
U1-U13, real callback corp-anchor, secret-backed ON verifiers, and two-corp post-fix UAT are unavailable here and remain NOT EXECUTED. Transfer T3-T5 stays frozen.

### DIFF
--- a/.github/workflows/dingtalk-lifecycle-staging-canary-contract.yml
+++ b/.github/workflows/dingtalk-lifecycle-staging-canary-contract.yml
@@ -1,0 +1,31 @@
+name: DingTalk Lifecycle Staging Canary Contract
+
+on:
+  pull_request:
+    paths:
+      - '.github/workflows/dingtalk-lifecycle-staging-canary.yml'
+      - '.github/workflows/dingtalk-lifecycle-staging-canary-contract.yml'
+      - 'docker-compose.app.staging.yml'
+      - 'docker/app.staging.env.example'
+      - 'scripts/ops/dingtalk-lifecycle-staging-canary-contract.test.mjs'
+      - 'scripts/ops/dingtalk-lifecycle-staging-canary-remote.sh'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  contract:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+      - name: Verify runner YAML parser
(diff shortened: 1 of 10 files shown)


zensgit:
Exact-head rework at 0f40e5c126 after independent adversarial review:

- strict DEPLOY_KNOWN_HOSTS pinning; no StrictHostKeyChecking=no
- deployed SHA requires matching image-tag + /health commit; unknown/conflict fail closed
- staging compose overrides stale env-file build metadata with exact IMAGE_TAG
- canary contract moved out of sealed-export-pinned plugin-tests.yml into its own PR-only workflow
- local: canary contract 36/36, deploy identity 4/4, sealed-export package provenance PASS

Still draft/unarmed; ON actions remain non-executable.

zensgit:
Final Codex review added two fail-closed hardenings after the main rebase:

- `d21821a8b7`: pin the already-proven live backend image before the first OFF write so rollback can recreate with the same exact 40-SHA image even when a failed force-recreate leaves no container to inspect. Mutation back to live-container re-resolution makes the constructed rollback test fail.
- `064c7aecc8`: distinguish an unset lifecycle env key (effective OFF) from `docker exec` failure. Read failure now aborts the snapshot instead of becoming a false exact-OFF proof. Mutation back to `|| true` changes the result to `accepted:false` and makes the test fail.

Local exact-head evidence: lifecycle contract 41/41, `bash -n`, `git diff --check`, and the sealed `plugin-tests.yml` pin remains byte-identical to main. All lifecycle ON actions remain non-executable; this PR is still Draft/unarmed pending fresh exact-head CI.

Issue #4820, opened 2026-08-08
[test-infra] flake family: 'terminating connection due to administrator command' uncaught on test (18.x) after all tests pass


zensgit:
Two occurrences on 2026-08-08, both on the **test (18.x)** leg, both with every vitest suite green and the job failing on a post-suite **Uncaught Exception**:

```
error: terminating connection due to administrator command
 ❯ Parser.parseErrorMessage pg-protocol/src/parser.ts:369
```

| Occurrence | PR | Run | Outcome |
|---|---|---|---|
| 1 | #4648 (D5 user-mutex) | 31241970356 | rerun → green → merged |
| 2 | #4655 (D6 restore txn) | 31247749976 | rerun dispatched |

Both PRs' diffs have zero contact with the failing region (the exception surfaces after `multitable-durable-startup` / `attendance-w4pre1b` suites complete; neither PR touches those). The 20.x leg of the same heads passed both times — timing-dependent, 18.x-only.

**Hypothesis**: #4799's scratch-DB teardown (排空后再 drop, uses `pg_terminate_backend` on lingering connections) races a still-idle pooled connection from a prior suite; the killed idle client emits an 'error' event with no listener → uncaught → vitest exits 1 despite all tests passing. Same family as the W1-1 GF3 shared-DB settle races.

**Suggested fix directions**: attach an error listener / swallow `57P01` on pool idle clients in the test harness, or have the drain terminate only backends belonging to the scratch database being dropped (`pg_stat_activity.datname` filter — if it already filters, the leak is a pool from the shared DB being targeted incorrectly).

Filed per the flake discipline: second same-signature recurrence gets its own ticket rather than another silent rerun.

zensgit:
Mitigation landed in PR #4852 (merge f0745831fe5385ccacf8fe6d6e5fd51174c02117).

Evidence at exact head f772f689b5ff50456d8aef79fdada63dec8d3f38:
- scratch drain suite: 15/15
- real MetaSheetServer lifecycle suites: 10/10, each reporting scratchDrain=CLEAN
- test (18.x), test (20.x), migration-replay, e2e, web-tests and coverage all passed
- teardown captures the post-start internal pool, restores handlers/env in finally, and only classifies exact SQLSTATE 57P01 (or the exact code-less administrator-command message) as expected scratch termination

[...]

PR #4852, opened 2026-08-10
test(core-backend): own scratch database termination

zensgit:
## What
- protect shared/system database names from the scratch-drop helper
- absorb only SQLSTATE 57P01 / exact administrator-command termination on pools owned by the affected real-server suites, and only during teardown
- capture the real `poolManager` internal pool after scratch `DATABASE_URL` rebinding
- replace event-loop-turn timing with a bounded condition wait in the real-PG child proof

## Verification
- `scratch-database-drain.db.test.ts`: 15/15 against local PostgreSQL 15
- affected real-server suites: 10/10 against independent random scratch DBs
- backend `tsc --noEmit`
- generic connection termination and ECONNRESET remain loud

## Boundary
This is the narrow mechanism fix for #4820. The issue should remain observation-gated until repeated Node 18 CI runs are green; this PR does not claim every real-DB suite in the repository has been converted to a scratch DB.

### DIFF
--- a/packages/core-backend/tests/helpers/scratch-database-idle-pool-arm.mjs
+++ b/packages/core-backend/tests/helpers/scratch-database-idle-pool-arm.mjs
@@ -1,0 +1,88 @@
+/**
+ * #4820 dual-arm child — real-PG proof that an idle Pool receiving admin-termination
+ * emits unowned 57P01 without a handler, and that attachOwnedPoolTerminationHandler owns it.
+ *
+ * Env: ARM=uncaught|owned, SCRATCH_URL, ADMIN_URL, DATNAME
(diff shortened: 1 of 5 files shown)


zensgit:
Exact-head rework at f772f689b5 after independent adversarial review:

- capture MetaSheetServer internal pool after server.start()
- missing capture fails loudly after cleanup
- handler detach and env restore are in finally
- only SQLSTATE 57P01 or the exact un-coded administrator message is absorbed
- local PostgreSQL: scratch drain 15/15, two real-server suites 10/10, tsc PASS

Issue #4820 remains open pending Node 18 repetitions; no repo-wide isolation claim.

PR #4865, opened 2026-08-11
fix(stock-prep): pass default integration scope

zensgit:
## Summary

- pass the existing default integration scope to the stock-preparation dashboard and all six child views
- prevent tenantless platform-admin sessions from issuing unscoped stock-preparation reads that fail with `TENANT_REQUIRED`
- keep the surface read-only; no ERP/K3 write path is added or changed

## Evidence

- observed the deployed `/api/integration/stock-preparation/projects` request returning HTTP 400 on the 192.168.1.222 read-only workbench
- targeted test: `32 passed`
- added a regression assertion that both tenant and workspace scope reach the dashboard GET request

## Validation

```text
apps/web/tests/StockPreparationWorkspace.spec.ts: 32 passed
```

No entity-host dependency install, build, migration, dry-run, Apply, or K3 Save/Submit/Audit was performed.

### DIFF
--- a/apps/web/src/components/integration/stockPreparation/StockPreparationWorkspace.vue
+++ b/apps/web/src/components/integration/stockPreparation/StockPreparationWorkspace.vue
@@ -61,6 +61,7 @@
       <StockPreparationDashboardView
         v-if="activeKey === 'dashboard'"
         :project-id="selectedProjectId"
+        :scope="scope"
         @select-project="handleDashboardProjectSelect"
         @navigate-stage="handleNavigateStage"
       />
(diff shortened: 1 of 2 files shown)


zensgit:
## Post-merge exact-head review — production fix is correct; proof gap moved to #4867

Reviewed PR head `8e8dbffb8bb0614dbbaa67bc78a6a3d4d9a21a56`, merged as `6bee1e88cfbc356b0698b859211373a92e3613fa`.

The production data flow is correct: the shell computes the existing default integration scope once and passes it to the dashboard plus all six child components; each child already consumes `props.scope` in its service request.

One test gap remained at merge time:

```text
baseline=32/32 PASS
mutationRemoveProjectWorkspaceScope=32/32 PASS
mutationRemoveAllSixNonDashboardScopes=32/32 PASS
```

The added regression asserted only the dashboard request, so the PR body's all-child claim was not mutation-protected. Draft #4867 adds one runtime test that mounts every view and asserts every exercised stock-preparation GET carries the same tenant/workspace scope. Its project-workspace and exception-queue single-scope mutations each fail exactly on the affected endpoint; final baseline is 33/33.

The earlier inline concern about tenantless platform-admin write controls is also still present at the final head: GET scope propagation does not make backend POSTs tenant-bound. The server remains fail-closed with `TENANT_REQUIRED`, so this is not an external-write escape, but those controls are not usable by that session. The current 222 validation must stay read-only; any later internal confirmation/write acceptance must use a tenant-bound principal or land a separately reviewed readonly UI posture.

No change to #4865 is requested after merge; #4867 carries the proof and status correction.

PR #4871, opened 2026-08-11
fix(directory): harden deprovision restore evidence

zensgit:
## Summary

- bind deprovision effects to the current selected event and use request generations so stale list/effects responses cannot restore the wrong event
- fail closed while event evidence refreshes or fails, and mirror the backend all-effects-applied restore predicate
- validate event/integration UUIDs, integer limits, compatibility mode types, fixed-route mode agreement, and path/query integration agreement
- return values-free generic errors for unexpected database/service failures
- add the panel regression suite to the always-on required Web Test run

## Exact head

- rebased onto `origin/main` at `eadd2dd88bf084bf20318fd4b4513c721654f4b9`
- reviewed head: `171ae867fab4b74719e425b86fbd0dde530e2759`
- net scope: exactly 5 files
- this PR remains Draft and unarmed; fresh exact-head CI is required

## Verification

- backend route suite at the rebased head: 127/127 pass
- frontend panel at the rebased head: 11/11 pass
- core-backend `tsc --noEmit`: pass
- web `vue-tsc -b`: pass
- independent Kimi adversarial review: APPROVE, 0 P1/P2
- mutation: reverting all-effects-applied to any-effect-applied reds the named predicate test
- mutation: removing both stale-response generation rejection and event/effects binding reds the named stale-effects test; the two controls are deliberately defense in depth and are not claimed individually load-bearing

## Deliberate compatibility boundary

- keep fail-closed behavior for empty restore modes and unknown event statuses
- keep fixed restore routes path-authoritative; contradictory body/query inputs are rejected instead of silently coerced
- do not retry a mutating restore request against a legacy endpoint after a 404
- no deployment, feature-flag, Stream, or lifecycle configuration change


zensgit:
Codex exact-head closeout review at 912724d5d37df1dd6c799cf557483bd09b4c7eae: APPROVE, 0 P1 / 0 P2.

Evidence rechecked:
- backend route suite: 129/129 green;
- frontend deprovision panel suite: 11/11 green;
- invalid integration UUID is a fixed 400 and unknown preview failures are a fixed, values-free 500;
- removing the UUID guard or restoring raw database error text makes the named test red;
- independent adversarial Grok review found 0 P1 / 0 P2;
- exact-head CI is fully settled: 18 pass, 0 pending, 0 fail.

Residuals are test-depth or pre-existing compatibility boundaries, not blockers for this patch. Posture remains Draft and unarmed. This review does not authorize merge, deployment, or any lifecycle flag change.

Issue #4844, opened 2026-08-10
Transaction ownership: multi-statement boundaries that BEGIN on a caller-supplied connection (W4C-5 soak prerequisite)


zensgit:
## Why

`plan` (PR #4839) documented "`connection` must be idle" and enforced nothing. On a dirty connection PostgreSQL only WARNs on the nested `BEGIN`; the unconditional `finally` ROLLBACK then aborts the **caller's** transaction, and `set_config(…, true)` overrides the caller's `statement_timeout`. That one site is fixed. The class is not.

The same shape pre-exists on `main` at two more sites, and one of them is strictly worse than what was fixed:

| site | shape | why it matters |
|---|---|---|
| `packages/core-backend/src/attendance/w4c0-operation-registry.ts:917` `runAttendanceResultOperationTransactionV1` | `BEGIN ISOLATION LEVEL SERIALIZABLE` at `:925`, **`COMMIT` at `:933`**, no idle probe | Called on a connection that already has an open transaction, it **commits the caller's uncommitted work**. `plan`'s defect only destroyed work; this one durably publishes it. |
| `packages/core-backend/src/attendance/w4c2-outbox-dispatcher.ts:84` `dispatchAttendanceResultEventOutboxV1` | bare `BEGIN` at `:104`, commits at `:155`, no idle probe | Same class, no isolation-level pin. |

Both blobs are byte-identical between `origin/main` and PR #4839's head — this ticket does not describe a regression, and it does not block that PR.

An earlier fix of the same defect (`assertConnectionIsIdleForSessionExclusiveRolloutLockV1`, added by W4C-5 NEW-B in #4773) placed a probe by hand at one site. #4839 placed a second one by hand. A third hand-placed probe is not the answer: each round closes one site and leaves the class open, which is how this ticket came to exist twice.

## What to build — three categories, not one helper

A single "all boundaries share one helper" rule is the wrong shape, because these functions do not all want the same thing. Distinguish by **who owns the transaction**:

1. **Owns its own transaction** — must prove the connection is idle *before* `BEGIN`. Refusal is a product code, never a raw SQLSTATE.
2. **Joins the caller's transaction** — must not issue `BEGIN`/`COMMIT`/`ROLLBACK` at all. The type system should make this the visible default rather than a convention.
3. **Nested work inside an existing transaction** — uses a savepoint helper that names, rolls back to, **and releases** its own savepoint, leaving the caller's transaction exactly as it found it.

[...]

PR #4898, opened 2026-08-13
feat(attendance): INERT authoritative-result-write CORE for §7.3 (Gate D1, #4556)

zensgit:
## Gate D1 — INERT authoritative-mode result-write CORE for `attendance_record_calculations` (#4556 / #4844)

Delivers the shared, contract-enforcing writer for `mode='authoritative'` rows — the §7.3/7.4/7.5/7.7/7.8 logic of the RATIFIED W4 lock (`attendance-issue-4556-w4-segment-calculation-design-lock-20260724.md`) — built **INERT**. D2 wires `live_punch`, D3 wires `scheduled`; each removes a `boundaryFail` site and flips its declaration in its own reviewed change. **Draft: do not merge, do not mark ready.**

New files:
- `packages/core-backend/src/attendance/w4c2-authoritative-calculation-core.ts` — the core (baseline / calculation completed+review / reversal, one parameterized INSERT).
- `packages/core-backend/tests/integration/attendance-w4c2-authoritative-calculation-core.db.test.ts` — 28-leg real-Postgres proof.

### Extraction decision — Option B (thin shared helpers + one mode-specific assembler)
The shadow builder `insertShadowCalculation` hardcodes 4 mode literals and omits 5 authoritative-only columns; unifying it into a `mode`-parameterized fn would require editing the shadow writer inside the 2200-line boundary file (altering emitted SQL, threading NULLs), which cannot be proven byte-identical as cheaply as not touching it and touches the very file the correspondence guard scans. **Shadow stays byte-identical by absence from this diff — no existing `src/` file is modified.** The core instead carries ONE parameterized INSERT (`insertResolvedAuthoritativeCalculationRowV1`) that all three kinds funnel through, so column logic is not hand-duplicated (defensible vs "a 7th copy"). `projectedDailyFingerprintV1` mirrors the four existing private copies byte-for-byte; the four are intentionally left un-rewired (separate, non-inert refactor). **Baseline fingerprint is caller-supplied** (`preimage.compatibilityFingerprint`), never self-computed — matching the import kernel (`:907-909`) so it is the frozen value the baseline-uniqueness index is keyed on; at-most-one-baseline is enforced by an existence read mirroring kernel `:875-883`.

### Two enforcement layers per invariant
(a) the core pre-validates and refuses with a **product code** (never a raw SQLSTATE) — the mutatable layer; (b) the W4C-0 DB CHECK/trigger is the backstop, each proven by a raw-insert leg that bypasses (a).

### §7.3 invariant → test matrix (28 legs, all green on real PG)
| # | Invariant | Positive | Negative (product code) | DB backstop leg |
|---|---|---|---|---|
| 1 | version-uniqueness + lineage | 2 sequential calcs → v N,N+1, 2nd supersedes 1st | `LINEAGE_TARGET_MISSING` (via reversal restores) | `uq_arc_record_version`, `trg_arc_lineage_guard` (not-strictly-lower) |
| 2 | retry idempotency (org,entrypoint,operation_id) | same op twice → 1 row, replay | `REPLAY_CONFLICT` (same op, diff payload) | `uq_arc_operation` |
| 3 | legacy_baseline | effect=none, 0 children, op_id NULL, fp=compat; idempotent | `BASELINE_FINGERPRINT_INVALID` | `uq_arc_baseline`; **same-txn atomicity**: forced failure after baseline → neither row persists |
| 4 | supersedes exact locked current | 2nd completed supersedes 1st, pointer moves | `VERSION_CONFLICT` (stale expected), `RECORD_NOT_FOUND` | `uq_arc_record_version` |
| 5 | review hidden placeholder | every projected field NULL, effect none, no pointer/baseline | `REVIEW_SHAPE_INVALID` | `chk_arc_review_shape` |
| 6 | reversal | present→restores==preimage pointer, set_active; absent→restores NULL, set_retired | `REVERSAL_SUPERSEDES_REQUIRED`, `REVERSAL_RESTORES_MISMATCH` | `chk_arc_reversal_supersedes` |
| 7 | projection_effect / count | completed→set_active, count 1..3, segments | `EXPECTED_COUNT_INVALID`, `COMPLETED_SHAPE_INVALID` | deferred `trg_arc_children_commit_guard` |
| 8 | append-only | no `updated_at` column | — | `deny_mutation` refuses UPDATE |

### Mutation verification
**Empirically mutated (guard neutered → the exact negative leg reddens, nothing else, restored via `cp`):** `VERSION_CONFLICT` (confirms the pre-check is the sole stale-predecessor guard — the optimistic UPDATE keys off the *locked actual*, not the caller's expectation, so the two do not cover for each other), `REVIEW_SHAPE_INVALID` (the core hardcodes NULL projected fields, so only the guard rejects a review-with-projection), `REPLAY_CONFLICT`, `REVERSAL_RESTORES_MISMATCH`. **Argued-not-executed (structural):** the remaining product-code guards fall through to a DB SQLSTATE / crash when neutered (different error than the asserted product code → the leg reddens); the independent gate should run the full sweep. `LINEAGE_SELF_REFERENCE` / `LINEAGE_NOT_STRICTLY_LOWER` are unreachable via the public API with valid inputs — defensive pre-validation backed by `trg_arc_lineage_guard`; their negatives live on the DB backstop.

### INERT / byte-neutral proof
- No existing `src/` file modified — the boundary + `w4c2-authoritative-delivery.ts` are untouched (`git diff` empty for both).
- `DECLARED_DELIVERED` unchanged `{live_punch:false, scheduled:false}`; the 3 `boundaryFail('W4C2_AUTHORITATIVE_MODE_NOT_DELIVERED', 503)` sites intact; the new core does **not** contain that call form (the header backtick-quotes the code name only).
- Correspondence guard `w4c3a-rollout-control-inventory.test.ts` — **15/15 green**.
- No production caller: the only importer of the core is its test file (grep of `src/` = 0; no barrel globs the dir).
- **Deletion-green** — the set is `{core module, test file, plugin-tests.yml run-list line, vitest.config exclude line, ci-wiring pinned-family line, pin JSON reverted to 46dbca8e…}`; removing the core + test leaves the package type-checking (verified: no dangling reference), and reverting `plugin-tests.yml` reverts the pin.

### Two-point wiring + provenance pin
- `vitest.config.ts` exclude (so the no-DB job cannot skip-green it) + `plugin-tests.yml` attendance real-DB run-list — the CI-wiring guard's `runs exactly once (no-DB exclude ⟺ real-DB run-list)` leg passes for the new file.
- `plugin-tests.yml` changed ⇒ provenance pin regen: `evidenceFiles.pluginTestsWorkflow` in `vectors/s6a-package-provenance-pins.json` → `b90a0e78…` (was `46dbca8e…`); **only that digest moved** (`sealed-export-package-provenance.test.cjs` OK).
- `scripts/ops/attendance-w4c2-ci-wiring.test.mjs` pinned-family set updated (alphabetical position) — **225/225 green**; `integration-guard-required-wiring-contract.test.mjs` **62/62 green**.

### Checks run locally vs CI-only
- **Local:** full `type-check` (0 errors); the 28-leg suite against real Postgres (`metasheet_test`); correspondence guard 15/15; ci-wiring 225/225; integration-guard contract 62/62; provenance-pin test OK; deletion-green type-check; 4 mutation probes.
- **CI-only:** the whole attendance real-DB batch through `plugin-tests.yml` (provisioned Postgres); required-check aggregation.

### Open items for the gate/owner (named, not resolved in D1)
1. **Live-path compatibility-fingerprint source (D1→D2 seam).** The completed path needs `preimage.compatibilityFingerprint` to append the baseline on an active legacy parent; today ONLY import/rollback/ops freeze such a value — the live/scheduled boundary has no producer. D2 must supply one (deterministically, over the parent's legacy daily projection) or `PREIMAGE_INVALID` fails closed. Named in the core header. The at-most-one-baseline existence read makes cross-path fingerprint values inconsequential for uniqueness, but the live path still needs a frozen source.
2. **Retry-conflict semantics** on `(org,entrypoint,operation_id)` with a differing payload → `REPLAY_CONFLICT` is **precedent-derived** (mirrors `w4c3c-recompute.ts:283`), not lock-derived.
3. Invariant-5's "later completed needs no baseline" leg installs the retired/`review_placeholder` precondition **by fixture** (disclosed in the leg comment) — it proves the core branch, not an end-to-end review→completed flow (that is a D2/D3 boundary concern).

### DIFF
--- a/.github/workflows/plugin-tests.yml
+++ b/.github/workflows/plugin-tests.yml
@@ -1344,6 +1344,7 @@ jobs:
             tests/integration/attendance-w4c2-live-scheduled-boundary.db.test.ts \
             tests/integration/attendance-w4c2-posture-matrix.db.test.ts \
             tests/integration/attendance-w4c2-gate-matrix-e5.db.test.ts \
+            tests/integration/attendance-w4c2-authoritative-calculation-core.db.test.ts \
             tests/integration/attendance-w4c2-p2-remediation.db.test.ts \
             tests/integration/attendance-w4c2-p2-1-canonical-freeze-anchor.db.test.ts \
             tests/integration/attendance-w4c2-p12-migration-schema-gates.db.test.ts \
--- a/packages/core-backend/src/attendance/w4c2-authoritative-calculation-core.ts
+++ b/packages/core-backend/src/attendance/w4c2-authoritative-calculation-core.ts
@@ -23,8 +23,11 @@
  *
  * DESIGN — two enforcement layers per invariant (repo doctrine: gate the MECHANISM, and a mutation
  * must be able to redden it — feedback_gate_the_mechanism_not_the_claim.md):
- *   (a) this module pre-validates every §7.3 authoritative invariant it can and refuses with a
- *       PRODUCT CODE (never a raw SQLSTATE) — the mutatable layer;
+ *   (a) this module pre-validates every §7.3 authoritative invariant it can on the normal
+ *       (live/scheduled completed+review) write path and refuses with a PRODUCT CODE rather than a raw
+ *       SQLSTATE — the mutatable layer. Two documented exceptions where a raw 23505 is still reachable
+ *       are named in "Concurrency notes" below (concurrent DIRECT `appendAuthoritativeLegacyBaselineV1`;
+ *       concurrent same-op reversal) — neither is on a production path (INERT) and neither corrupts;
  *   (b) the DB CHECK constraints / triggers installed by the W4C-0 durable-storage migration are
  *       the backstop, each proven separately by a raw-insert leg that bypasses (a).
  *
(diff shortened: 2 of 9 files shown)


zensgit:
## Gate hardening round — commit `fa6ccf7f9a` (Draft, do not merge)

Closes the GO-with-hardening items from the independent gate (verdict bound to `4f4ed872b3`). No production `src/` file touched; INERT preserved; no plugin-tests.yml/pin change this round.

**CI blocker (P16 DML zero-bypass inventory):** classified every DML/SELECT site of the core — curated-debt `X06` for the two `attendance_records` pointer UPDATEs (`writeCompletedRow`/`writeAuthoritativeReversalV1`), a W4C-3a current-record base-read entry for `lockParent` (`write_lock`), and four W4C-4 calculation-read entries (`nextCalculationVersion`/`assertLineageStrictlyLower`/`appendAuthoritativeLegacyBaselineV1`/`retryReplayLookup`). Collector 58/58.

**P2-F1** concurrent-retry idempotency: SAVEPOINT catches `uq_arc_operation` (23505), rolls back, re-runs the replay lookup → replay/REPLAY_CONFLICT (never raw SQLSTATE). +2 real-PG 2-connection race tests (different records; review same record), both red without the fix.
**P2-F2** completed path pre-validates non-negative minutes → COMPLETED_SHAPE_INVALID (no raw 23514). +test.
**P2-F3** added red-on-neuter negatives for BASELINE_PROJECTION_INVALID / REVERSAL_EFFECT_INVALID / PREIMAGE_INVALID.
**P2-F4** the 8 DB-backstop legs now pin the exact `.constraint`/RAISE text — which exposed & fixed 4 raw-insert helpers that were passing on a spurious 42P08 and never reaching their constraint.
**F5** reversal enforces `frozenTarget≡preimage` (present case) +test. **N1** one completed leg round-trips real ordered `segment_snapshot`. **F6/F7** review-path parent-state asymmetry (named D2 obligation) + existence-read/reversal-op-id concurrency caveats documented in the core header.

Local: suite **36/36**; every new guard mutation-proven (F1×2, F2, F3×2, F5 each redden only their leg); tsc 0; correspondence 15/15; collector 58/58; ci-wiring 225/225; provenance-pin OK; deletion-green (tsc 0 without core+test).

PR #4965, opened 2026-08-17
fix(approval): round-scope dedup history to the current post-return round (Lock-4 OD-L4-10(a))

zensgit:
## Summary

- `loadApprovalHistory` applied no epoch/round/return filter, and the `action:'return'` dispatch branch fed that FULL, unfiltered history straight into `applyAutoApprovalCascade`. With `mergeAdjacentApprover` or `dedupeHistoricalApprover` set (settable today only through the publish API — reachable, not dead code, per `docs/development/approval-lock4-flow-policies-20260817.md` §F4-D) and a graph where the same person is re-assigned to a node after a return, the re-entered node silently re-merged against its own pre-return approval, **nullifying the return with no human in the loop**.
- This is priced into the RATIFIED `docs/development/approval-lock6-requester-global-policy-20260817.md` §1 L6-A as a precondition on the upcoming template-level dedup tier, and §3 gate A-7 tests it directly: *"any slice making a dedup tier authorable must FIRST scope the history the flags read to the current post-return round ... Shipping the switch without one is forbidden."*
- Fix (backend-only, no FE / wizard-step change):
  1. `loadApprovalHistory`'s SQL now additionally requires `to_version >= COALESCE((SELECT MAX(to_version) FROM approval_records WHERE instance_id=$1 AND action='return'), 0)`. `to_version` is a per-instance monotonic counter already stamped on every `approval_records` row (manual approves AND `insertAutoApprovalEvents` rows alike, no legacy-NULL case) — reusing the same round-boundary idea the T2-4 threshold tally's epoch/cutoff fallback machinery already proves out (`docs` reference in code comment), without a new column or a new metadata stamp on the return's own audit row.
  2. The RETURN branch's own synchronous cascade evaluation now seeds `[]` instead of `await loadApprovalHistory(...)` — mirroring the create-cascade's existing `[]` seed. At the exact moment a return's own cascade runs, its `action:'return'` row has not committed yet, so predicate (1) cannot yet see it and would still surface the immediately-preceding pre-return approval.
- A note on the exact mechanism chosen: Lock-4 OD-L4-10(a) says to use "the existing `nodeEntryEpoch` machinery." I used `to_version` instead (an equally-monotonic, per-instance, already-universally-stamped counter with no NULL/legacy gap) rather than the literal `nodeEntryEpoch` int, because the dedup history is inherently **cross-node** (unlike the threshold tally's **per-node** epoch equality check) and `nodeEntryEpoch` is not currently stamped on the return's own audit row — adding that stamp would require a new metadata write with its own NULL-fallback branch. `to_version` achieves the identical round-boundary semantics with strictly less new surface. Flagging this explicitly for adversarial review since it's a deliberate deviation from the doc's literal wording (though not its stated intent).

## Test plan

- New real-DB suite `packages/core-backend/tests/integration/approval-dedup-return-round-scoping.db.test.ts` proves **both** dedup arms independently:
  - `mergeAdjacentApprover`: pre-return positive control (adjacent auto-merge fires via the documented publish-API path) → gate assertion (post-return, the re-entered node stays PENDING, not silently re-merged) → post-return continuation (a genuine round-2 approval lets the same arm fire again normally).
  - `dedupeHistoricalApprover`: same three-part shape, using a non-adjacent (separated by a different-actor node) match so the two arms are tested independently.
- Mutation-verified: reverted both engine edits via `cp`-backup + restore (never `git checkout --`), reran — both tests go RED (`expected 'approval_a' to be 'approval_c'` / `'approval_c'` to `'approval_d'`, i.e. the re-entered node incorrectly cascades past itself using stale pre-return history). Restored the fix (sha256-verified byte-identical to the pre-mutation state) — both green again.
- `npx vitest run tests/integration/approval-dedup-return-round-scoping.db.test.ts tests/integration/approval-nofm-threshold.test.ts tests/integration/approval-field-edit-enforcement.db.test.ts tests/integration/approval-template-policy-carrier.db.test.ts --config vitest.integration.config.ts` — all pass (27 passed, 2 pre-existing skips), confirming no regression in the T2-4 threshold tally, the Lock-7 content-edit invalidation filter (ANDed with the new return-floor predicate), or the just-landed L6-P1 policy-carrier suite.
- `npx vitest run tests/unit/approval-product-service.test.ts` — all 151 unit tests pass, including the mocked-DB "auto-approves adjacent same-user chains" coverage.
- `tsc --noEmit` — clean.

## Known gap (flagging, not fixing here)

- `plugin-tests.yml` is the only place backend real-DB integration files get wired into CI (explicit whole-file lists, not a glob), and it is out of scope for this PR. The new test file is verified locally (green with the fix / red without it) but is **not yet wired into any CI-required lane**. A follow-up needs to add it to the "Run approval real-DB integration" step's file list.

## Scope

Backend-only. Does not touch `apps/web`, the wizard step count, `AuthoringSectionId`, or the More-settings shell — that FE work lands in a follow-up PR stacked on this branch (the L6-A dedup-tier control + fifth-step activation), consistent with master §6 one-behavioral-slice-per-PR and the L6-P1 precedent of landing its own prerequisite PR ahead of the control it unblocks.

Not merging — RATIFY/merge-order decisions stay with the requesting session.


zensgit:
## Fix-round update (adversarial gate, head 62140682bc)

The gate correctly found two issues with the original submission and both are fixed as a third commit on this branch:

**P2-1 (test isolation gap, fixed):** every original test returned TO the node whose own dedup evaluation the assertion checked, so it only ran through the `[]` seed at the return call site — never through the durable `to_version` floor in `loadApprovalHistory`. Reverting ONLY the floor (keeping the seed) left the original 3 tests green. Added an isolated test where the return targets an EARLIER node, so the node under test is re-reached via ordinary forward advance (through the floor-guarded path). Mutation-proven: reverting only the floor now reds only this new test; the others correctly stay green (confirming the gate's diagnosis).

**P2-2 (real correctness gap, fixed):** `applyNodeTimeoutEffect`'s jump effect reaches its target through the same `resolveReturnToNode` resolver a manual return uses, but stamps `action:'jump'`, so the return-only floor never moved for it — a BACKWARD timeout-jump was exactly as nullifying as an unfloored manual return. `adminJump` does NOT share this gap (`isReachableDownstream` structurally rejects a non-forward target there). OD-L4-10(a)'s boundary is corrected: it governs **backward re-entry** generally (return OR backward jump), not literally `action='return'`. Direction is computed at the jump dispatch site via the same "already on the static path to the current node" predicate the manual return action's own validation uses; a backward jump seeds `[]` and stamps `metadata.backwardReentry:true`; a forward jump is untouched and still composes (matches admin jump's existing forward-only behavior). Mutation-proven: reverting only the backward-jump handling reds only the new GATE test; the forward positive control and everything else stays green.

Also fixed two overclaiming comments (to_version is NOT stamped on every row — transfer/comment/sign rows outside a cascade keep it unbumped, excluded by the `action='approve'` filter regardless) and the `loadApprovalHistory` docblock.

All 7 tests in `approval-dedup-return-round-scoping.db.test.ts` pass; no new CI wiring needed (extends the existing dedicated `approval-realdb-l6a-roundscoping.yml` lane). No regression in `approval-nofm-threshold.test.ts`, `approval-field-edit-enforcement.db.test.ts`, `approval-template-policy-carrier.db.test.ts`, the full `approval-node-timeout-effects.test.ts` suite (including its own pre-existing backward-jump re-entry test), or the 151-test unit suite. `tsc --noEmit` clean.

New head: `62140682bc64a81cd0624c8b23473a6e6a9de908`.

PR #4967, opened 2026-08-17
feat(approval): P3-B — More-settings fifth step + L6-A template dedup tier

zensgit:
## Summary

master §P3-B / RATIFIED `docs/development/approval-lock6-requester-global-policy-20260817.md` §1 L6-A: activates the "更多设置" wizard step (fourth of five, `测试发布` last — Lock-0 L0-4 / Lock-6 §2.7) and lands its first functional control: a 3-way tier (**不去重** / **仅一次全自动同意** / **仅连续节点自动同意**) authoring the template-level dedup arms on `runtimeGraph.policy.autoApproval` (`dedupeHistoricalApprover` / `mergeAdjacentApprover`). Both booleans are ALREADY server-enforced (Lock-4 §2.6 `evaluateAutoApprovalAssignment`) — this slice is authoring-surface only, **zero engine change, zero new contract field**.

**M7 (master §4 P3-B exit, "no inert controls"):** the fifth step is authorized ONLY in this same commit that lands the functional control — activation is a typed change to `AuthoringSectionId` + `authoringSections`, per Lock-6 §2.7.

## Stacked on two prerequisites

This branch is based on #4965 (round-scoping) and additionally cherry-picks #4966's single commit (publish-sequencing fix), because BOTH are required for the dedup-tier control to be honest under M7/M8:

- **#4965 — round-scoping (Lock-4 OD-L4-10(a)):** without it, enabling either dedup arm via a return-heavy graph would silently nullify a return.
- **#4966 — publish-sequencing (Lock-6 L6-P1 gate F3, independently corroborated during this slice):** without it, selecting a tier and publishing in the same sitting would silently discard the selection — a real, pre-existing bug that ALSO affects the already-shipped `allowRevoke` checkbox, found via component-level testing here before any P3-B code existed.

Base is set to #4965's branch (closer ancestor); the diff you'll see against `main` also includes #4966's commit until that PR merges — this PR's own commit (the one that matters for review) is the last one, `feat(approval): P3-B — More-settings fifth step + L6-A template dedup tier`.

## What changed (this PR's own commit)

- `apps/web/src/approvals/templateAuthoring.ts`: `TemplateDedupTier` type + `autoApprovalDedupTier` draft field; `templateDedupTierFromPolicy` / `isTemplateDedupTierLocked` (hydrate-time projection; a persisted BOTH-true combination — unrepresentable by this 3-way tier — is detected and never flattened, Lock-6 §2.6 gate X-1); `buildTemplateAutoApprovalPolicy` + `buildPublishPolicy` update (merges the tier onto `draft.originalPolicy.autoApproval`, preserving any other key and omitting the whole `autoApproval` key when empty — §2.2: absent === 不去重, byte-stable for every existing template).
- `apps/web/src/types/approval.ts`: `RuntimePolicy.autoApproval` typed as `AutoApprovalPolicy` (was an opaque `unknown` placeholder pending this exact slice, per its own L6-P1 comment).
- `apps/web/src/views/approval/TemplateAuthoringView.vue`: the new step's radio-group control, immediate-apply (no separate Save/Cancel transaction, matching the existing D0 grammar e.g. `draft.allowRevoke`), disabled + an explanatory hint when the persisted policy is in the locked both-true state.

## Test plan

- 24 pure-function cases (`approval-template-authoring-policy-carrier.test.ts`) covering hydrate/publish projection, sibling-field preservation (`allowRevoke`, `revokeBeforeNodeKeys`, an unrelated `autoApproval` key like `actorMode`), the locked-state byte-preservation, and the omit-when-empty convention.
- 8 component-level cases (`approvalTemplateAuthoring.spec.ts`) covering: the exact 5-step activation order; an **M7 census** that 更多设置 renders 3 real radio options (not an empty shell); the 不去重 default; select-a-tier-then-publish round trip; hydrate-with-a-persisted-tier round trip; gate X-1's locked-state round trip; and an **M8/A-4-spirit** check that switching the template tier never writes into a node's own `autoApprovalPolicy`.
- Mutation-verified via `cp`-backup + sha256 restore (never `git checkout --`): neutering `buildTemplateAutoApprovalPolicy` to always return `undefined` turns 11 tests red; stripping the control's markup from the More-settings step turns 6 tests red (including the M7 census). Both restored and reverified green (128/128 across the two spec files).
- `vue-tsc --noEmit` clean.
- No FE spec file is new — both touched test files already carry tokens in `run-required-web-tests.sh` and `approval-web-guard.yml`.

## Scope discipline

Does not touch the flow canvas, assignee logic, form builder, field-permission grid, or the engine (dedup enforcement is pre-existing; round-scoping is #4965). Scope is L6-A alone (Lock-6 OD-L6-2(a)); every other Lock-6 candidate (L6-B through L6-F2) stays deferred per that document.

Not merging — merge order (this PR after #4965 and #4966) and RATIFY decisions stay with the requesting session.

### DIFF
--- a/apps/web/src/approvals/templateAuthoring.ts
+++ b/apps/web/src/approvals/templateAuthoring.ts
@@ -213,6 +213,17 @@ export interface TemplateAuthoringDraft {
   // mirrors the `originalAutoApprovalPolicy` / `amountConsistencyCheck` preserve-verbatim pattern
   // already used in this file.
   originalPolicy?: RuntimePolicy | null
+  /**
+   * P3-B / Lock-6 L6-A (docs/development/approval-lock6-requester-global-policy-20260817.md §1) —
+   * the editable projection of the TEMPLATE-level dedup tier. Mirrors the node-level
+   * `mergeWithRequester` / `originalAutoApprovalPolicy` pattern immediately below: this is the ONE
+   * sub-field of `originalPolicy.autoApproval` the editor authors; every other key (and the
+   * both-flags-true combination this tier cannot express) survives verbatim via
+   * `buildTemplateAutoApprovalPolicy`, which reads `originalPolicy.autoApproval` directly rather
+   * than reconstructing it from this field alone. See `templateDedupTierFromPolicy` /
+   * `isTemplateDedupTierLocked` for the hydrate-time projection and lock detection.
+   */
+  autoApprovalDedupTier: TemplateDedupTier
   fields: FieldAuthoringDraft[]
   steps: ApprovalStepDraft[]
   // G-1 anti-flatten keystone: a COMPLEX graph (any cc/condition/parallel node, or any
@@ -382,6 +393,9 @@ export function createEmptyTemplateDraft(): TemplateAuthoringDraft {
     // stays the client-chosen create-time default, unchanged by the carrier fix below —
     // `originalPolicy` is correctly left absent (nothing to preserve yet).
     allowRevoke: true,
+    // §2.2 "no shipped default may change": absent `autoApproval` === 不去重 for a brand-new
+    // template, exactly like every pre-P3-B published template.
+    autoApprovalDedupTier: 'none',
     fields: [createEmptyFieldDraft(1)],
     steps: [createEmptyStepDraft(1)],
   }
@@ -978,6 +992,11 @@ export function draftFromTemplate(template: ApprovalTemplateDetailDTO): Template
     // absent `policy` (never-published template) keeps the create-time default of `true`.
     allowRevoke: template.policy?.allowRevoke ?? true,
     originalPolicy: template.policy ?? null,
+    // P3-B / Lock-6 L6-A — projects the hydrated `policy.autoApproval` onto the 3-way tier. Reads
+    // `template.policy?.autoApproval` (not `originalPolicy` above, though they are the same value)
+    // for clarity that this is a pure function of the persisted definition, matching
+    // `templateDedupTierFromPolicy`'s own signature.
+    autoApprovalDedupTier: templateDedupTierFromPolicy(template.policy?.autoApproval),
     // Hydrate side of the #3161 §1 preserve: carry the amount total-check mapping through verbatim
     // (shallow clone, never alias the source schema). Absent → no key (no phantom on round-trip).
     ...(template.formSchema.amountConsistencyCheck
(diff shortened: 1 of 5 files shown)


zensgit:
## Fix-round update (adversarial gate, head 4d4e760dbe)

**P2 (test-only, fixed):** the control's ENABLED/EDITABLE state was asserted by nothing — mutating `:disabled="readOnly || isAutoApprovalDedupTierLocked"` to `:disabled="true"` left all 104 prior tests green (jsdom fires the change handler regardless of `disabled`). Added `radios.forEach((radio) => expect(radio.disabled).toBe(false))` to the M7/M-2 census test. Mutation-proven: the `disabled:true` mutation now reds ONLY this new assertion; everything else stays green exactly as predicted. Restored and reverified.

**P3-1 (M8 honesty, fixed):** selecting 仅连续节点自动同意 (`mergeAdjacentApprover`) also relaxes two publish-time duplicate-assignee validations for a parallel gateway — the static branch check (`allowParallelDuplicateAssignees`, `ApprovalProductService.ts:4595`) and the dynamic-source preflight (`assertNoParallelDynamicAssigneeConflicts`, skipped at `:4623-4625`) — because the merge machinery legitimately absorbs same-approver overlap at runtime instead of leaving it to a 409. Added a hint scoped to ONLY that tier (verified absent for the other two tiers, present and naming "并行" only when `merge_adjacent` is selected): *"选择该项还会放宽并行分支的发布校验：同一审批人出现在同一并行网关的多个分支中不再阻止发布，运行时会自动跳过重复分支的指派。"* States exactly the verified exemption — no broader semantic invented.

**P3-2 (disclosed, not reworked, per your instruction):** "连续节点" in the shipped predicate (`mergeAdjacentApprover`, `evaluateAutoApprovalAssignment`) means *temporally-latest-any-other-node* (the most recent approve entry not at the current node), not strictly graph-adjacent — this only diverges from a literal "adjacent in the graph" reading inside parallel topologies (where "latest" can be branch-arbitrary; that's exactly why the cross-branch skip guard `evaluateSkippedCrossBranchAdjacent` exists). The control's label (仅连续节点自动同意) and hint copy are accurate for the linear/no-parallel case this program has shipped so far; they should not be read as a strict graph-adjacency guarantee once parallel gateways with this tier are in play. Left as a documented nuance, not reworked, per your instruction.

Full `run-required-web-tests.sh` under Node 20.20.2: **368 files, 4593 tests, all green** (was 4592 before this commit's one new test). `vue-tsc --noEmit` clean. No closing keywords, `plugin-tests.yml` byte-identical, zero NUL bytes.

New head: `4d4e760dbeb2df84a68a2590200f872f4bed2ca4`.

PR #4972, opened 2026-08-18
feat(approval): P1-C — shipped timeout + threshold frontend compatibility

zensgit:
## What

Implements P1-C from `docs/development/approval-parity-master-design-lock-20260817.md` — shipped
`timeout` (T1-1) + `threshold` (T2-4 N-of-M / 门槛会签) frontend compatibility.

Both engine capabilities existed and were already emitted/preserved by the backend
(`ApprovalGraphExecutor.ts`, `ApprovalProductService.ts`, `ApprovalSlaScheduler.ts`) but were
FE-unexposed: the authoring allowlists (linear + complex) did not recognize `approvalThreshold` /
`timeout` as valid config keys, so ANY template carrying either was forced fully read-only
(I12/I13). This was never an actual flatten risk — backend `normalizeApprovalGraph` always
re-emits both keys verbatim for every approval node regardless of graph topology.

CORRECTION (post-review, adversarial gate P3-2): the previous wording here said "two stale FE
comments incorrectly claimed the backend 'silently drops' them; both are corrected in this PR" —
imprecise. Only ONE comment (`templateAuthoring.ts`, implementation) made that specific "silently
drops" claim, and it also carried a stale line-number anchor. A second comment was also updated in
this PR (`approval-template-authoring-graph-preserve.test.ts`, a TEST comment) but it said `timeout`
was "outside the editor allowlist" — true at the time, and never a backend-drop claim. Both edits
are accurate; this description's original characterization of the second one was not.

## Engine

**No engine change.** `packages/core-backend` is untouched (`git diff --stat` shows zero files under
`packages/`). Verified: `pnpm --filter @metasheet/core-backend build` (tsc) is clean.

## Linear-only constraint (verified against current main)

Both capabilities share the SAME region definition: an approval node is rejected only when it lies
**inside a parallel gateway's branch** (between a `parallel` node's branch edge and its
`joinNodeKey`, join excluded) — not "inside any complex graph," not "inside a condition branch."

- Threshold: `ApprovalProductService.ts` `APPROVAL_THRESHOLD_IN_PARALLEL` (:2706-2712 nested-parallel
  branch walk, :2757-2763 all-branch walk).
- Timeout: `APPROVAL_NODE_TIMEOUT_PARALLEL_UNSUPPORTED`, gated by the shared
  `collectParallelRegionNodeKeys` primitive (:1677-1703).

The FE reuses the **existing** `collectParallelRegionNodeKeys` port in `graphTopologyEdit.ts`
(already shipped for the canvas's nested-parallel authoring guard) rather than duplicating it — one
definition serves both the canvas guard and this slice's gate.

- Linear graphs are **never** inside a parallel region by construction (`orderedLinearNodes` admits
  no `parallel` node), so the linear editor offers threshold/timeout unconditionally — the
  constraint is satisfied structurally, not by a runtime check.
- The complex-graph editor (`ApprovalGraphNodeConfigEditor.vue`) disables the threshold mode option
  and the timeout-enable checkbox per-node when `approvalNodeInParallelRegion(nodeKey)` is true, and
  `setApprovalNodeMode`/`setApprovalNodeTimeoutEnabled` refuse the mutation even if a caller bypassed
  the disabled UI (defense-in-depth, mirroring P1-B's remove-card guard pattern).
- `validateApprovalNodeEdits`/`validateTemplateApprovalFlow` re-check the same region for a
  programmatic edit, with a POSITIVE CONTROL proving the identical config is accepted on a sibling
  node just outside the region.

## Timeout effect enum — do not invent

`NodeTimeoutEffect` has 5 declared members (`remind | transfer | jump | auto_approve | auto_reject`)
but `ApprovalProductService.ts`'s `NODE_TIMEOUT_SUPPORTED_EFFECTS` (and the actual
`ApprovalSlaScheduler.fireNodeTimeouts` switch) only wire `remind | transfer | jump`. Every picker in
this PR (linear step editor, complex node editor) offers exactly that 3-member set — never
`auto_approve`/`auto_reject` — and the `approval-node-threshold-timeout-config.spec.ts` mount spec
asserts the rendered `<option>` list against `NODE_TIMEOUT_SUPPORTED_EFFECTS` behaviorally, not a
hand-copied literal.

## What changed

- `apps/web/src/types/approval.ts` — widened `ApprovalMode` to include `'threshold'`; added
  `NodeTimeoutEffect` / `NODE_TIMEOUT_SUPPORTED_EFFECTS` / `NODE_TIMEOUT_MAX_AFTER_MINUTES` /
  `NodeTimeoutConfig`; added `approvalThreshold?` / `timeout?` to `ApprovalNodeConfig`.
- `apps/web/src/approvals/templateAuthoring.ts` — both authoring allowlists gain the two keys plus
  matching nested-shape checks (`timeoutConfigHasBackendDrop` mirrors backend `normalizeNodeTimeout`
  SHAPE, not the separate `validateNodeTimeoutConfigs` VALUE gate — the two disagree on `unit`);
  `stepDraftFromApprovalNode`/`buildStepConfig` hydrate/emit threshold + timeout for the linear path
  (mode-switch-away drops the orphaned `approvalThreshold` key); a stable-`localId`-based jump-target
  resolver survives step reorder; `validateTemplateApprovalFlow` gets the integer-only threshold
  check + timeout shape/target validation (deliberately does NOT invent a static N≤M publish bound —
  the backend's bound is scoped to the legacy `assigneeType`/`assigneeIds` shape this editor never
  emits); corrected the stale "backend silently drops" comment.
- `apps/web/src/approvals/approvalNodeEdit.ts` — `ApprovalNodeSourceEdit` gains editable
  `approvalThreshold?` / `timeout?` (not just preserved); `validateApprovalNodeEdits` gains an
  optional `parallelRegionNodeKeys` param for the linear-only gate.
- `apps/web/src/approvals/nodeConfigEditorContext.ts` /
  `apps/web/src/approvals/components/ApprovalGraphNodeConfigEditor.vue` /
  `apps/web/src/views/approval/TemplateAuthoringView.vue` — typed N-of-M control on the mode picker
  and typed duration/effect/target/unit timeout controls, in both the linear step editor and the
  complex node editor.
- `apps/web/src/views/approval/TemplateDetailView.vue` — `Record<ApprovalMode,string>` exhaustiveness
  (vue-tsc caught this ripple from widening `ApprovalMode`) + a business-label-only timeout indicator
  tag (never the raw effect enum).
- Two pre-existing tests updated (not weakened): `approval-template-authoring-graph-preserve.test.ts`
  swapped its "still-unsupported linear key" example from `timeout` (now supported) to
  `signaturePolicy` (still explicitly deferred per the master lock); a canvas-inspector spec's
  hand-built API stub gained the new required fields it was missing (the mount was throwing
  `approvalNodeTimeout is not a function` until this fix).
- Two new spec files + CI collection (both path-filter blocks in
  `.github/workflows/approval-web-guard.yml`, the two `npx vitest run` invocations in that workflow,
  and `apps/web/scripts/run-required-web-tests.sh`'s final line — each new token verified to match
  exactly one file).
- `.github/workflows/approval-web-guard.yml` — also repairs a zero-indented `run:` key
  (commit `b5f5c38c57`) that made the file's second vitest step an orphan top-level YAML key on
  `origin/main` (`yaml.safe_load` on `main@e1dd97bfd4` returns top-level keys
  `['name', True, 'jobs', 'run']`; GitHub's own run diagnosis on that commit reads "This run likely
  failed because of a workflow file issue"). This breakage was introduced by #4957
  (`0cbae291bc`), NOT by this PR — flagged separately below since `approval-web-guard` is not
  currently in `main`'s required-check set, so it was silently broken for every approval-FE PR
  merged since #4957 without blocking anything.

## Tests

- `approval-template-authoring-threshold-timeout-compat.test.ts` (48 cases after the fix-round below;
  originally 36, pure logic): no-flatten for both allowlists with positive controls (genuinely-unknown
  key, out-of-union mode, mode/threshold-value mismatch, malformed timeout shape — see the fix-round
  note below for when the malformed-timeout/invalid-N controls became genuinely LINEAR-side, not just
  COMPLEX-side); linear + complex round-trip (N=2, mode-switch-away cleanup, timeout
  duration+effect+transfer-target+jump-target-through-reorder); linear-only fail-closed + POSITIVE
  CONTROL against a real constructed parallel-region fixture; M6 dynamic-M honesty (no invented static
  bound).
- `approval-node-threshold-timeout-config.spec.ts` (10 cases, mounted
  `ApprovalGraphNodeConfigEditor.vue`): M7 every new control wired (not inert) — mode/threshold/
  enable-checkbox/afterMinutes/effect/transfer-target/jump-target/unit all provably change edit
  state through their setter; the effect option set assertion is the M6/M8 behavioral positive
  control.
- Mutation probe (manual, per house rules): reverted both allowlist additions, confirmed exactly the
  6 no-flatten assertions went red while every round-trip/validation/linear-only test stayed green,
  then restored `templateAuthoring.ts` byte-identical (sha256-verified).
- `pnpm --filter @metasheet/core-backend build` (tsc): clean.
- `npx vue-tsc -b --noEmit` (apps/web): clean.
- `packages/core-backend/tests/unit/source-files-no-raw-control-bytes.test.ts`: passes (no raw
  control bytes introduced).
- Full approval suite re-run after the P1-C changes (17 files / 416 tests) + all new files: green.

## Post-open notes (added after review)

- **Jump-target resolver invariant, verified not just asserted**: `draftFromTemplate`'s second pass
  pairs `orderedApprovalNodes[i]` with `steps[i]` by index, relying on `stepDraftFromApprovalNode`
  returning non-null for every `approval`-type node. Checked directly: the function has exactly ONE
  `return null` path (`if (node.type !== 'approval') return null`, templateAuthoring.ts:562) — no
  other early return exists in its body. So for every node the caller filters to `type === 'approval'`,
  the function is guaranteed non-null and the index pairing is sound. This is a static single-branch
  fact, not a trust-the-comment claim.
- **`unit:'wall_clock'` vs `effect:'auto_approve'` — asymmetric handling, both unreachable in real
  data, deliberate**: a persisted `timeout.unit:'wall_clock'` forces the WHOLE template read-only
  (`timeoutConfigHasBackendDrop` — normalize never emits that literal, only omits the key), while a
  persisted `timeout.effect:'auto_approve'` stays editable (it's within the full declared
  `NodeTimeoutEffect` enum) but blocks SAVE until the author picks a different effect
  (`validateTemplateApprovalFlow` rejects `timeoutEffect === ''`, and the picker only offers the wired
  3). Both shapes never occur in real persisted data (normalize's emit-only-`business`-or-absent
  discipline for `unit`; publish's `APPROVAL_NODE_TIMEOUT_EFFECT_UNSUPPORTED` for `auto_approve`/
  `auto_reject`) — the difference in FE handling is not a live behavioral gap, just two different
  defensive postures against two different-shaped impossible inputs, called out here so it isn't
  mistaken for an inconsistency.
- **Post-open rebase**: rebased onto `f2ed020d1b` (#4970, CI `merge_group` triggers — 1 shared file,
  `approval-web-guard.yml`, no line overlap with this PR's hunks) after this PR was first opened;
  P1-C's feat commit was replayed at `e8bffa99aa` (patch content identical, diffed to confirm — only
  line numbers/blob hashes shifted). Re-pinned the execution-ledger row's SHA to match.

## Fix-round (adversarial gate, 2026-08-18 — 0 P1 / 3 P2 / 2 P3 / 2 NIT)

An independent adversarial review found no live-fire correctness or security defect (the core
no-flatten + round-trip properties held under mutation), but flagged 3 P2s, 2 P3s, and 2 NITs.
Closed in this round, each mutation-proven (a production-line neutered, the new/updated test
verified to redden, then the file restored byte-identical):

- **P2-1 (jump-target dangling reference)**: the complex-graph editor's `timeout.jumpToNodeKey`
  preview did not check the target NODE EXISTS or is an `approval`-type node — an edit sequence
  that deletes the jump target (the shipped `removeLinearNode` delete affordance on a
  single-in/single-out approval node) could pass FE validation with a dangling key, surfacing as a
  raw backend 400 instead of the linear editor's existing inline error. Fixed: `validateApprovalNodeEdits`
  gains an `approvalNodeKeys?: Set<string>` param (mirrors the existing `parallelRegionNodeKeys`
  pattern), supplied from `draft.preservedGraph` in `validateTemplateApprovalFlow`; a single
  membership test covers both "doesn't exist" and "exists but isn't approval-typed" (same one
  message the linear editor already uses, since its own candidates are always approval steps).
  Added: two direct negative tests, a positive control, an omitted-param backward-compat test, and
  an end-to-end reproduction of the original gate's exact constructed failure (delete-then-validate)
  now passing.
- **P2-2 (five load-bearing-guard claims, unproven under exclusive neutering)**: the two
  parallel-region SETTER guards (`setApprovalNodeMode`/`setApprovalNodeTimeoutEnabled`,
  `TemplateAuthoringView.vue`) carried code comments claiming to be "the actual/real enforcement
  point" — exclusive neutering showed the full suite stayed green, because the real enforcement for
  SAVE is `validateApprovalNodeEdits`'s region check, not these setters. The comments are corrected
  (defense-in-depth for a caller that bypasses the render-layer `:disabled`, not the invariant's
  actual floor) and BOTH setters now have a discriminating mounted test (the real
  `TemplateAuthoringView.vue`, not the hand-built stub used by the existing M7 spec) proving the
  guard itself blocks a forced-past-disabled DOM event. The three LINEAR-branch no-flatten cases
  that previously had ONLY complex-side coverage (malformed timeout shape, invalid/absent threshold
  N, and the `isPlainRecord` sub-check specifically — the last needed an array-shaped fixture, since
  a bare string/number timeout is independently caught by a different sub-check regardless of
  `isPlainRecord`) now have LINEAR positive controls too. Also closed in the same pass: the timeout
  setters had no `nodeType` check and neither did the FE preview — `validateApprovalNodeEdits` now
  rejects `timeout`/`approvalThreshold` on a handler edit (mirrors backend
  `APPROVAL_HANDLER_CONFIG_INVALID`, ApprovalProductService.ts :2449), with a negative test. No
  current producer reaches this (render gate + setters are approval-node-only in practice today), so
  this is defense-in-depth, not a live-fire fix.
- **P2-3 (RESOLVED — lock text executed; retracts this bullet's prior "owner decision needed"
  framing)**: RETRACTION (governance adjudication, 2026-08-18): the prior round framed this as an
  open choice between (a) deleting the fallback or (b) amending the lock text to accept the widened-
  but-retained substitute. That framing was wrong. The master lock is RATIFIED — its verbatim text
  IS the recorded decision, so the substitute (keep-but-widen) was the deviation that needed
  ratification, not the other way around. Nothing amended the lock text, so there was nothing to
  adjudicate; the correct action was to execute the lock as written. Resolution: the master lock's
  §P1-C says, verbatim: *"Before lifting the read-only gate, this slice must delete that latent
  flatten branch and correct the stale frontend comments that claim the backend silently drops these
  keys."* The `: 'single'` fallback in `templateAuthoring.ts`'s linear-hydration ternary
  (`stepDraftFromApprovalNode`) is now DELETED. The lock's text does not itself name a replacement
  failure mode, so the fail-closed choice consistent with the existing tested guard was implemented:
  `unsupportedTemplateAuthoringReason` (the SEPARATE check right below in the same file) is now the
  SINGLE door for a defined-but-out-of-union `approvalMode` — it already forces the whole template
  read-only and blocks `persistDraft`. Hydration itself no longer re-decides that question: a defined
  `approvalMode` is preserved VERBATIM (no substitution to `'single'`); only a genuinely ABSENT key
  still takes the documented single-approver default. Both stale comments this clause also names are
  corrected again to describe the new (no-coercion) behavior. Tests added to
  `approval-template-authoring-threshold-timeout-compat.test.ts`: a negative proving an out-of-union
  `approvalMode` is preserved verbatim (not coerced) with an exact positive equality (not a
  `.not.toBe('single')`, per house doctrine that notEqual assertions can't distinguish "preserved"
  from "failed some other way"), and a positive control that all four legitimate modes plus the
  ABSENT-key default still hydrate correctly. Mutation-proven: restoring the old `: 'single'`
  fallback (cp-backup + sha256-verified restore, not `git checkout --`) reds exactly the new negative
  test; the positive control and the rest of the 50-test file stay green.
- **P3-1 / P3-2**: see the corrected wording above (approval-web-guard.yml credit; the "two stale
  comments" precision).
- **NIT**: the linear step editor's N-of-M `el-input-number` now coerces a `null` clear-value to `1`
  (`value ?? 1`), matching the complex editor's existing `setApprovalNodeThreshold` coercion — was
  previously fail-closed (a validation error) rather than aligned UX, not a correctness bug. A
  backend-citation line number was off by 4 (`:2260` → `:2256` for `case 'approval':`); corrected.

Verification for the FIRST fix round (head `6c2d61c419`): `npx vue-tsc -b --noEmit` (apps/web)
clean throughout (pre- and post-rebase); `pnpm --filter @metasheet/core-backend build` (tsc) clean;
the 6-file approval set this round touches was 240 tests green pre-rebase, re-confirmed as 5 files /
130 tests green post-rebase (the 6th file, `approval-handler-node-authoring.spec.ts`, folded into
that final count since it's a pre-existing file rather than a new one this round wired into CI).
`run-required-web-tests.sh` re-run in full on Node 20: every approval-suite test in it passed; the
run's other failures were confined to three unrelated multitable spec files this PR does not touch
(`multitable-automation-manager`, `multitable-automation-rule-editor`, `multitable-scoped-
permissions`). Only one of those three (`multitable-automation-rule-editor.spec.ts`) was
individually re-run against this PR's PRE-fix-round head to check for a pre-existing cause — it
failed there too, but with a DIFFERENT subset of tests each time (37 / 10 / 8 failures across three
separate runs of overlapping scope) — a nondeterministic-flake signature, not a deterministic
regression tied to a specific change. The other two files were not individually baseline-checked;
this PR's diff touches none of their files or sources, which is the basis for treating them the
same way, not a measurement.

Verification for the SECOND fix round / P2-3 lock-text execution (head `a1bea10939`, rebased onto
`origin/main` @ `385a433821`): `npx vue-tsc -b` clean; `approval-template-authoring-threshold-
timeout-compat.test.ts` + `approvalTemplateAuthoring.spec.ts` = 160/160 green on Node 20.20.2 (the
P1-C pure-logic spec plus the mounted TemplateAuthoringView spec that exercises `draftFromTemplate`
in a real DOM mount). `run-required-web-tests.sh` re-run in full on Node 20.20.2: 4664/4691 passed,
27 failures across 9 files, 4 of them approval-suite (`approval-center.spec.ts`, `approval-center-
master-detail.spec.ts`, `approval-e2e-lifecycle.spec.ts`, `approval-e2e-permissions.spec.ts`) and 5
multitable (`multitable-automation-manager`, `multitable-automation-rule-editor`, `multitable-
workbench-1672-1673`, `multitable-workbench-permission-wiring`, `multitable-workbench-history-
field-scope-wiring`). This round measured the baseline claim the prior round could only assert:
mechanical grep confirms none of the 9 failing files imports `templateAuthoring` (this PR's diff
cannot reach them), and the SAME failure signatures (5000ms test timeouts; `expected null to be
truthy` DOM-query misses) reproduce against the committed pre-this-round baseline with this round's
two-file diff stashed out (`git stash` on `templateAuthoring.ts` + its spec, not `git checkout --`)
— 8 failures on that baseline run vs. 27 / 13 in two runs of this round's head, the failing subset
varying run-to-run across overlapping scope each time. That is the nondeterministic-flake signature
this PR's other verification sections already document for a different file set, now directly
measured for this round's own diff rather than inferred from "the diff doesn't touch these files."

## Rebase onto main (K3 §prior_node_approver + L8-A explanation, 2026-08-18, head `d692e1d5fe`)

`origin/main` landed K3 (#4973, `prior_node_approver` assignee kind) and L8-A (#4974, `explanation`
field type) after this PR's prior head. Rebased onto current `main`; 4 files conflicted
(`types/approval.ts`, `approvalNodeEdit.ts`, `templateAuthoring.ts` ×4 blocks,
`run-required-web-tests.sh`), all resolved as semantic unions — both slices' capabilities fully
survive, neither is a subset of the other:

- `types/approval.ts` / `approvalNodeEdit.ts` / `run-required-web-tests.sh`: mechanical union (kept
  every type union member, every import, every token/doc-comment block from both sides).
- `templateAuthoring.ts` hydrate arm: kept K3's `priorStepLocalId: ''` field alongside this PR's
  no-coercion `approvalMode`/`approvalThreshold` (the P2-3 fix above, not the pre-P2-3 `'single'`
  fallback K3's version of this line predates).
- `templateAuthoring.ts` `buildStepConfig`: K3's version took `allSteps` (to emit a referenced step's
  CURRENT positional key for `prior_node_approver`, enforcing a STRICTLY-EARLIER ordering
  constraint); this PR's version took `stepLocalIdToNodeKey` (an unordered localId→nodeKey map, for
  the timeout jump target). Evaluated collapsing to one map and rejected it on a semantic ground, not
  merely scope caution: `stepLocalIdToNodeKey` carries no ordering information, so expressing K3's
  `referencedIndex < stepIndex` check through it would mean re-deriving position by parsing the index
  back out of the formatted `approval_N` key string — strictly worse than what existed. Both
  parameters are threaded through instead (`buildStepConfig(step, fieldIds, allSteps,
  stepLocalIdToNodeKey)`); `sourceFromStep`'s existing exported signature (also called directly by
  `linearStepSpine.ts`, outside this conflict) is untouched.
- `templateAuthoring.ts` validation: K3's strictly-earlier `prior_node_approver` PREVIEW check and
  this PR's threshold (T2-4) + timeout (T1-1) PREVIEW checks now sit as adjacent sibling `if` blocks
  in the same per-step loop; both fire independently.

Verified at `d692e1d5fe143649a8aa4304b30b847f8b637de8`: `vue-tsc -b` clean; K3's FE suites
(`approval-assignee-source.test.ts`, `approval-handler-node-authoring.spec.ts`,
`approval-template-authoring-approval-node-edit.test.ts`,
`approval-template-authoring-canvas-inspector.spec.ts`,
`approval-template-authoring-parallel-edit.test.ts`, `approvalTemplateAuthoring.spec.ts`) run
together with this PR's suites (`approval-template-authoring-threshold-timeout-compat.test.ts`,
`approval-node-threshold-timeout-config.spec.ts`,
`approval-template-authoring-linear-step-spine`) = 358/358 green, Node 20.20.2. `run-required-
web-tests.sh` re-run in full on Node 20.20.2: **4708/4708 green, 0 failures** — a materially
different (better) result than the pre-rebase head's 27-failure nondeterministic-flake run
documented above; this run retires that flake discussion for the current head rather than extending
it. The P2-3 mutation was re-proven post-rebase on the merged file: restoring the old `: 'single'`
fallback (cp-backup + sha256-verified restore, not `git checkout --`) reds exactly 1/50 — the same
negative as before — with the other 49 (including the new K3-merged hydrate/build path) staying
green. K3's `prior_node_approver` strictly-earlier gate was spot-checked directly against the merged
code (3 constructed cases, not committed as new test files — verification only): a valid earlier
reference round-trips clean (byte-correct `nodeKey` through the unified `buildStepConfig` call); a
self-reference and a reference to a later step both fire K3's PREVIEW error
(`需要引用一个位于其之前的审批步骤作为节点审批人`) and both make `sourceFromStep` independently emit
`nodeKey: ''` (the build-time mirror of the same rule). Also confirmed directly: the two
`orderedApprovalNodes` post-passes in `draftFromTemplate` (K3's `priorStepLocalId` resolution and
this PR's `timeoutJumpToStepLocalId` resolution) both still rely on the SAME approval-node↔step 1:1
invariant, and `stepDraftFromApprovalNode` still has exactly one `return null` site
(`node.type !== 'approval'`) post-merge, so that invariant holds; `createEmptyStepDraft` carries
every field of the merged `ApprovalStepDraft` shape (K3's `priorStepLocalId` + this PR's five
`timeout*` fields + `approvalThreshold`). `plugin-tests.yml` untouched (0-diff vs. `origin/main`); no
control bytes; no closing keywords in either new commit message.

## Deferred (out of this slice's explicit scope)

- `ApprovalDetailView.vue`'s runtime "会签完成/或签完成" completion badges do not yet have a
  `threshold` arm (the backend's `aggregateComplete` resolution-record metadata carries
  `approvalMode`/`aggregateComplete` uniformly for every mode, so the shape supports it) — this is
  the RUNTIME instance-history surface, not "node config" authoring, and is out of this PR's stated
  scope; flagged for a follow-up P5/member-experience slice.
- `signaturePolicy` stays read-only pending its own owner-ratified slice (unchanged by this PR,
  explicitly pinned by a positive-control test).
- No real-DB workflow added — every test in this slice is pure FE logic/mount; no backend/DB-adjacent
  behavior changed (T2-4/T1-1 engine behavior is pre-existing and covered by existing backend suites).


zensgit:
Incidental fix included in this PR: `.github/workflows/approval-web-guard.yml` had a zero-indented `run:` key (a stray top-level mapping key GitHub Actions' schema rejects), introduced by #4957 a few hours before this branch started — confirmed via `git blame` and by checking the workflow's push-triggered run history on `main` itself (every push since that commit failed to parse at 0s, scheduling zero jobs). This silently disabled the whole guard for every approval PR merged since. Fixed with a one-line indentation correction (verified via YAML round-trip: job/step structure and command content unchanged) so this PR's own new CI-collection tokens — and every other token already in that file — can actually run. Unrelated to the P1-C scope; flagging for visibility in case a separate/backdated fix is preferred instead.

PR #4976, opened 2026-08-18
fix(ops): preflight staging deploy migrations

zensgit:
## What changed

- snapshot `docker/app.staging.env` once per deploy, bind every compose invocation to that 0600 snapshot, and fail if the canonical file changes during capture;
- resolve full-SHA image tags once to immutable backend/web repository digests, then reuse those exact digests without any post-preflight pull;
- run the candidate image's migration list against the live staging DB before persistent compose files or live containers change, with a closed DB/migration env allowlist and an offline values-free alignment report;
- require exactly one `decision=aligned` both before deployment and again after the new containers are up;
- invalidate stale completion identity before the first persistent/live change and write the new values-free SHA/digest/env-fingerprint record only after health, migrations, auth, and summary gates pass;
- make attendance smoke/soak, lifecycle canary, OAuth config, and Stream UAT understand digest-pinned deployments only when the live digest, completion record, and persisted override agree.

## Why

Staging deploy run `31944319144` recreated backend/web before discovering two pending migrations and stopping at `do_not_run_full_migrate`. The old path therefore allowed a migration incompatibility to become a partial live deployment. This PR moves the compatibility decision before any persistent or live mutation and makes later operator lanes consume the same immutable deploy identity.

## Verification

- `bash scripts/ops/attendance-run-gate-contract-case.sh strict /private/tmp/codex-attendance-strict-exact-20260818` (pass; attendance runner 94/94)
- `pnpm verify:attendance-window-runner:pipeline:test` (94/94)
- lifecycle + OAuth + Stream contracts together (274/274)
- Stream UAT contract separately (51/51)
- `bash -n` on all four changed remote scripts
- `git diff --check`
- independent read-only adversarial review: 0 P1 / 0 P2 on exact patch
- rebase patch-id unchanged: `c4b75fddc36b17f55974ef6b883f9ba39f64c1d9`

## Hold / operations boundary

- Draft and unarmed. Do not auto-merge.
- No staging action, secret read/write, flag change, migration, deployment, prepare, or UAT was performed by this PR work.
- Stream and all lifecycle flags remain OFF.
- The two currently pending W7 migrations require a separate owner-authorized `action=migrate` before any deploy.
- After merge, the intended separately authorized sequence is `migrate -> deploy exact merge SHA with flags OFF -> read-only status -> prepare`; real UAT remains a later explicit authorization.

## Known separate UAT findings

This deploy-safety PR does not claim full Stream UAT completion. The current UAT evidence still needs follow-up for terminal card-button retirement, the documented forwarding step versus `supportForward:false`, observer unknown-evidence fail-closed behavior, corp-field-absence classification, and a deterministic concurrent double-click barrier.

### DIFF
--- a/scripts/ops/attendance-staging-window-runner-remote.sh
+++ b/scripts/ops/attendance-staging-window-runner-remote.sh
@@ -3,7 +3,7 @@
 #
 # Remote (deploy-host) half of .github/workflows/attendance-staging-window-runner.yml.
 # Executes ONE action per invocation against the STAGING stack only:
-#   deploy         — pin staging backend+web to a full-SHA image tag, migrate, verify build+auth
+#   deploy         — resolve a full-SHA tag to immutable backend+web digests, migrate, verify build+auth
 #   smoke          — run one of the five window smokes in-container (bundle doc:
 #                    docs/development/attendance-staging-window-bundle-20260702.md)
 #   status         — read-only snapshot (containers, health, settings, pending migrations)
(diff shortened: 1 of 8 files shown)


zensgit:
Exact-head follow-up: `d30caa45de00524e103dab24c4ec5fa5e63992c8` fixes the required-CI portability blocker found on `5f4f2827da`.

Root cause: the executable contract used BSD-first `stat -f '%Lp' || stat -c '%a'`; GNU `stat -f` exits 0 while printing a filesystem report, so Ubuntu never reached `stat -c` and the mode assertion failed in `contracts (strict)`, `test (18.x)`, and `test (20.x)`.

Delta: one test-harness line only, now GNU-first with BSD fallback. Production deploy code is unchanged.

Local verification on the new head:
- `node --test scripts/ops/attendance-window-runner-pipeline.test.mjs` -> 94/94
- `bash scripts/ops/attendance-run-gate-contract-case.sh strict /private/tmp/codex-attendance-strict-linux-stat-fix-20260818` -> PASS, including expected-failure legs
- `git diff --check` -> PASS

PR remains Draft/HELD, `autoMergeRequest=null`; awaiting fresh exact-head CI and review. No staging action or flag change was performed.

PR #4979, opened 2026-08-18
fix(approval): D-1 — reject fieldPermissions on non-write-capable node types at publish (Lock-7)

zensgit:
## What

Closes the shipped D-1 defect (`docs/development/approval-lock7-field-edit-enforcement-20260817.md` §2.7 D-1 + OD-L7-4, RATIFIED): `fieldPermissions` set on a `cc` / `start` / `end` / `condition` / `parallel` node was **silently discarded** by `normalizeApprovalGraph` — no error, no effect. An author configuring it believed the field hidden/read-only/protected when it was not.

P4-B (#4961) closed only the `editable` arm of this (the load-bearing privilege-escalation half) and deliberately left the **read axis** — `readonly`/`hidden` — silently dropped, logging it as D-1 "for a separate fix slice". This PR is that slice, and closes **both** axes. (Corroborated mid-implementation by an independently-run P7 phase-A verification ledger, which re-flagged that at its baseline `readonly`/`hidden` were still silently dropped on these five node types — the negatives here were widened in response to cover every `NodeFieldAccess` value on every affected type, not just `editable`/`cc`.)

## The fix

**Publish-time (authoring choke):** extends the existing inline pin inside `normalizeApprovalGraph` (`packages/core-backend/src/services/ApprovalProductService.ts`) — it must inspect the RAW per-node config *before* the per-type switch drops the key, which is why it lives there rather than in `validateFieldEditEnforcementPins`/`validateNodeFieldPermissionsAgainstFormSchema` (those run on the *normalized* graph, where the key is already gone). A non-empty `fieldPermissions` array of **any** access value on cc/start/end/condition/parallel now fails with a typed, values-free 400 `APPROVAL_NODE_FIELD_PERMISSIONS_UNSUPPORTED_NODE_TYPE` — `details` carries only `{ nodeKey, nodeType }`, never a field id or a submitted value.

This fires at all five authoring entry points that route through `assertApprovalGraph`: `createTemplate`, `updateTemplate`, `publishTemplate`, `restoreTemplateVersion`, `cloneTemplate`.

An **empty** array is *not* rejected (OD-L7-9: absent/empty ≡ no permissions; the FE always initializes `fieldPermissions: []` on a step draft — rejecting emptiness would brick ordinary authoring). A **malformed** entry (e.g. missing `fieldId`) *is* rejected — the check is presence-of-a-non-empty-array selected, not shape-selected.

**Dispatch-time (normalize):** stays gated off for `STORED_RUNTIME_CONTEXT` (the `asRuntimeGraph` re-normalize path that runs on every dispatch), so a hand-edited or legacy stored runtime graph carrying this shape still dispatches — per Lock-7 §2.1's widen-only rule, this tightening can never make an in-flight instance undispatchable.

## Publish-entry-point coverage

| Entry point | Route |
|---|---|
| `createTemplate` | `POST /api/approval-templates` |
| `updateTemplate` | `PATCH /api/approval-templates/:id` |
| `publishTemplate` | `POST /api/approval-templates/:id/publish` |
| `restoreTemplateVersion` | version restore |
| `cloneTemplate` | template clone |

All five route through `assertApprovalGraph` → `normalizeApprovalGraph`, where the choke lives — no entry point needs its own copy.

## In-flight-tolerance proof

New real-DB test: publish a valid `cc` graph, then hand-edit the **stored** `approval_published_definitions.runtime_graph` row directly (raw SQL, bypassing the now-closed authoring choke — mirrors the shipped "G-4 legacy explicit-editable subcase" technique) to inject `fieldPermissions` onto the `cc` node. Then dispatch: create the instance (which walks `start → cc_x → approval_final` through the hand-edited graph), comment, and approve — all succeed (200/201), proving `asRuntimeGraph`'s `STORED_RUNTIME_CONTEXT` re-normalize tolerates this shape rather than 500ing an in-flight instance.

## Tests

`packages/core-backend/tests/integration/approval-field-edit-enforcement.db.test.ts` (real-DB, 19/19 green):
- **G-12 (D-1 closed, both axes):** 3 access values × 5 node types = 15 negatives, every one asserted against the typed values-free code; a malformed-entry negative; an empty array and an approval-node matrix as positive controls.
- **New in-flight-tolerance test** (above).
- Every pre-existing gate (G-3/G-4/G-5/G-6/G-7/G-8/G-9/G-10/G-11/G-15/G-16/G-17) reverified green — the runtime write-mask enforcement for approval/handler nodes is unaffected (lives in a separate code path, `applyHandlerFieldWrites`).

**Mutation-tested:** neutered the new check with `if (false && ...)` (cp-backup + sha256 restore, never `git checkout --`) — all 15 new negatives went red while every positive control and the in-flight-tolerance test stayed green, confirming selectivity.

Also reran `tests/unit/approval-product-service.test.ts` (151/151), `approval-p1c-field-permissions.api.test.ts` (7/7), `approval-handler-node.db.test.ts` + `approval-postgate-acceptance.api.test.ts` + `approval-bridge-redaction-regression.test.ts` (26/26) — all green, no regression.

`tsc --noEmit` clean. `apps/web/scripts/run-required-web-tests.sh` run under Node 20: 4620/4628 green; the 8 failures are pre-existing timeout flakes in files this PR does not touch (`multitable-automation-manager/rule-editor`, plus 3 approval FE specs that pass 100% in isolation when re-run without sandbox resource contention) — this PR touches **zero** `apps/web` files, so no FE/CI wiring changes are needed (neither authoring surface renders the field-permissions grid for these five node types, confirmed by reading `ApprovalGraphNodeConfigEditor.vue` and `templateAuthoring.ts`).

No schema/migration change — `APPROVAL_SCHEMA_BOOTSTRAP_VERSION` and its pinned copies are untouched. No new CI wiring needed — `.github/workflows/approval-realdb-field-edit.yml` already path-filters on `ApprovalProductService.ts` and the modified test file, and the suite is already excluded from the no-DB default vitest config (two-point wiring pre-existing).

## Deferred / out of scope

- Everything else in Lock-7 (write-path enforcement for approval/handler, routing-driver pins, dedup invalidation, etc.) — already landed via P4-A/P4-B and untouched here.
- FE authoring surfaces already refuse to save a stray `fieldPermissions` key on these node types via `complexNodeConfigHasBackendDrop` (locks the whole template read-only); this PR converts that already-guarded failure mode from an invisible drop to an explicit, typed error — a strict improvement, not a new FE requirement.


zensgit:
## Fix round — independent adversarial gate response

Gate MD `/tmp/pr4979-d1-gate-20260818.md` on head `2a17d8358e`. **VERDICT: 0 P1, 1 P2, 3 P3.**

New head: `22f238f1b8e2eb1c31bfbd7b1f6be0e08d6de999`

**The gate's crux finding: dispatch-path reachability P1 hypothesis REFUTED.** A harder probe than this PR's own (mid-flight tamper on four node types at once — start/cc/condition/end — plus a full comment/transfer/approve/return/approve/reject battery; separately a parallel fan-out/join battery) found every call still 200s. The design (publish-time reject, dispatch-time tolerate) held.

### P2-1 (closed) — empty-array positive control was untested
`graphFor('cc', [])` used to omit the `fieldPermissions` key entirely (`length > 0 ? {...} : {}`), so the "empty array is not rejected" control never sent an empty array. Mutation A (delete `&& length > 0`) left 19/19 green. Fixed: the helper now places its argument under `fieldPermissions` verbatim; the test reads back the constructed payload to prove `[]` was actually sent. Mutation A now reds exactly that one assertion.

### P3-1 (closed) — non-array shapes were a residual silent-drop channel
`fieldPermissions: {memo:'hidden'}` / a string / `null` on the five node types fell through `Array.isArray(...)` and was still silently dropped (201, key gone) — asymmetric with an approval node, which hard-400s the same shape. Guard widened to `rawPermissions !== undefined && !isTolerableEmptyArray` (using `!== undefined`, not `'in'`, per the gate's note about `publishTemplate` re-normalizing an already-normalized graph). Three new negatives (object/string/null); mutation-verified; in-flight-tolerance test re-verified green.

### P3-2 (closed) — wrong rationale in the code comment
The comment cited FE `fieldPermissions: []` step-draft initialization as the reason to tolerate an empty array — but that default is for **approval** steps, a type this guard exempts. Corrected to state the true rationale: a conservative tolerance choice (no live surface sends this shape on these five types), not an FE-compat requirement.

### P3-3 (recorded) — ledger row, not code
Appended to `docs/development/approval-parity-execution-ledger-20260817.md` §4: the read-axis 400 goes beyond OD-L7-4(a)'s literal ratified text (scoped to `editable`), in the direction the lock's own non-ratified OD-L7-4(c) contemplates further (corpus-aligned widening). Recorded as a **reversible public-API narrowing** under goal-set provenance — silent-drop→400 chosen as the fail-closed default; OD-L7-4(c)'s widening direction stays available as a future owner-ratified slice; nothing stored breaks either way.

### Verification
`tsc --noEmit` clean; unit suite 151/151; real-DB suite (whole file, CI-command-matched) 19/19 — G-12 now has 18 negatives (15 access-value + 1 malformed-array + 3 non-array) plus the strengthened empty-array control; every mutation probe (empty-array carve-out, non-array guard) re-run and restored via `cp` + sha256 (never `git checkout --`); no closing keywords; `plugin-tests.yml` untouched; no schema/migration change (`APPROVAL_SCHEMA_BOOTSTRAP_VERSION` untouched); no FE files touched.

PR #4981, opened 2026-08-18
fix(approval): P7-R2 — V-6 focus-ring contrast, harness stylesheet, detail-view copy

zensgit:
## P7-R2 remediation slice

From the P7 phase-A evidence ledger (`p7-phaseA-evidence-20260818.md`): FAIL-2, FAIL-5, FAIL-6, and the three §2 raw-exposure candidates in `ApprovalDetailView.vue`. Own isolated worktree from fresh `origin/main`. Not merging.

### FAIL-2 (P2) — V-6 focus-ring contrast, real-Chromium measured post-fix

13 of 19 flow-canvas controls used `--el-color-primary-light-5` (2.05–2.14:1, below the ratified ≥3:1) for the keyboard focus ring. Switched to `--el-color-primary` in `ApprovalFlowCanvas.vue`:

| Control | Pre-fix (ledger) | Post-fix (measured here) |
|---|---|---|
| canvas viewport | 2.05 | **4.95** |
| node-card selector | 2.14 | **5.17** |
| toolbar (undo/redo/zoom/fit) | 2.14 | **5.17** |
| edge-insert-btn (already passing) | 5.17 | 5.17 (unchanged) |
| edge-insert-menu item | color-only, no ring | **5.17** |
| move-target | `outline: none` | **4.49** |

Two sub-findings fixed: the edge-insert-menu button signalled focus by text color alone (no ring) — split from `:hover` and given a real outline; the move-target shipped a bare `outline: none` — split from `:hover` and given a real ring.

**Second root cause found during the real-Chromium re-measure, not named by line number in the original ledger finding**: the canvas toolbar buttons (undo/redo/zoom/fit) are real `<el-button>`s — their ring comes from Element Plus's own base `--el-button-outline-color` default (also `--el-color-primary-light-5`), not from any rule in this file. Fixed with a scoped `--el-button-outline-color` override on `.template-authoring__canvas-toolbar :deep(.el-button)` — deliberately not touching Element Plus's app-wide default, which is a separate, unratified, much larger surface.

No sibling component needed the same fix: the other `--el-color-primary-light-5` occurrences in `ApprovalFormBuilder.vue`/`ApprovalFormPalette.vue`/`ApprovalFormFieldInspector.vue`/`ApprovalFormInlineEditor.vue` are secondary `border-color`/`box-shadow` accents alongside a `color:` change, not the actual keyboard-focus ring — those harnesses independently measured 28/28 and 21/21 ring PASS with production CSS loaded.

Pinned in `approval-flow-canvas-a11y.test.ts` as a **mechanism guard** (P2-1 hardening precedent in the same file): enumerates every `:focus-visible` rule in the component's raw stylesheet text (not two named line numbers) and asserts (a) the set is non-empty, (b) no rule references light-5, (c) every rule has a non-color channel, (d) no rule ships a bare `outline: none` — plus a dedicated pin for the toolbar override. Every branch mutation-proofed (cp-backup + sha256 restore; confirmed each mutation reddens exactly the targeted assertion, nothing else).

### FAIL-5 (P2) — harness stylesheet, verified real-Chromium

The shipped `approval-form-builder-harness.ts` and `approval-inspector-keyboard-harness.ts` imported no CSS, so `var(--el-*)` was undefined and any paint assertion over them was vacuous. Added the two production imports (`element-plus/dist/index.css` + `tokens.css`) to both. Deliberately did **not** add `.use(ElementPlus)` to either — the form-builder harness mounts no `<el-*>` components at all, and the inspector-keyboard harness deliberately stubs Element Plus with native elements to measure native radiogroup/tablist keyboard semantics; registering the plugin would change what it measures for zero contrast benefit.

Re-run CI-exact `playwright.approval-verification.config.ts`: **7/7 still green**. Spot-measured 6 palette-chip focus rings: all now resolve to the real `--el-color-primary` value at **5.17:1** (previously would have been dropped/vacuous).

The inspector-keyboard spec (`playwright.verification.config.ts approval-inspector-keyboard`) remains red — confirmed same pre-existing failure signature as the ledger (`approvalSourceCount is not a function`, FAIL-1, harness API drift). That's a separate remediation slice; out of scope here, unaffected by this CSS fix.

### FAIL-6 (P3) — handler node accent

`handler` was the only one of seven node types with no per-type left-border accent (fell through to the card default border color). Given `--el-color-info`, extending the existing `parallel`/start/end info-token-sharing precedent recorded in the same file (danger/error stays reserved for a future validation marker, never a node type). Placed before `.is-selected` so the cascade-order comment there stays accurate. Pinned by adding `handler` to the existing per-type-accent test loop; mutation-proofed.

### Three §2 candidates — confirmed, then fixed (values-free doctrine)

Each confirmed by constructing the exact drift state named in the ledger (test asserts against the state; mutation-reverting the fix reddens exactly that test), then fixed:

1. **`nodeLabel`** (ledger's highest priority — fires on *ordinary* template drift, not an exotic shape). Now prefers the live template, falls back to the pinned/frozen version, and only then a values-free `节点已变更` placeholder — never the raw node key. Two pre-existing assertions in `approval-detail-record-table.spec.ts` that pinned the raw-key leak (`activeTemplate: null` fixture) are updated to expect the fix — not silently kept green by reverting.
2. **`cancelledAssigneesLabel`**: resolves cancelled-approver ids to display names from already-loaded assignment metadata (no new fetch, same convention as the adjacent `assignmentDisplayLabel`); falls back to a values-free count (`其他 N 位审批人已失效`) rather than ever joining raw ids.
3. **`formatFieldValue`**: object-valued detail/sub-form cells no longer `JSON.stringify`; resolves a known display key (`displayValue`/`name`/`label`/`title`) or falls back to a values-free `复杂内容` placeholder.

### Verification

- `vue-tsc -b`: clean, before and after rebase.
- `run-required-web-tests.sh`: every approval-scoped file (including all touched/extended files) green in every run. Two full-lane runs each surfaced a handful of unrelated multitable/automation failures; confirmed sandbox batch-contention flakes, not a regression from this diff — 100% green when re-run individually, files disjoint from this diff, and a concurrent sibling worktree was independently running its own full test suite on the same sandbox machine during these runs.
- `ui-foundation-style-guard.spec.ts` (UF-6, zero hex/rgb literals): caught a **real regression I introduced** — my own explanatory comments inside `ApprovalFlowCanvas.vue`'s `<style>` block contained hex-looking literals (`#92b1f5`, `#2563eb`, and even a PR number `#4956`, since `4956` are all valid hex digits). Fixed by rewording the comments to name tokens/ratios without hex-shaped substrings. Green after.
- `plugin-tests.yml`: byte-identical (`git diff` empty).
- No new spec files — both edited specs (`approval-flow-canvas-a11y.test.ts`, `approval-detail-record-table.spec.ts`) are already registered in both CI collection points (`run-required-web-tests.sh` and `approval-web-guard.yml`).
- No closing keywords, no control bytes (checked via `git diff`).

### Deferred (out of this slice)

- FAIL-1 (inspector-keyboard harness API drift) and the rest of FAIL-0's class (FAIL-3/4/7, ungated guards) — separate remediation slice.
- The tail-slice NOT-YET-LANDED items and owner-only rows from the ledger — untouched, as chartered.

### DIFF
--- a/apps/web/src/approvals/components/ApprovalFlowCanvas.vue
+++ b/apps/web/src/approvals/components/ApprovalFlowCanvas.vue
@@ -419,6 +419,16 @@ function nodePosStyle(pos: NodeLayout): CSSProperties {
   background: color-mix(in srgb, var(--el-bg-color) 92%, transparent);
   box-shadow: var(--el-box-shadow-lighter);
 }
+/* FAIL-2/V-6 fix (P7-R2, second root cause): these are real `<el-button>`s, so their
+   `:focus-visible` ring does NOT come from a rule in this file at all — it comes from Element
+   Plus's OWN base `.el-button` default, which sets `--el-button-outline-color:
+   var(--el-color-primary-light-5)` (measured 2.14:1 here, same as every other light-5 ring in
+   this component). Overriding the custom property, scoped to just this toolbar, fixes the ring
+   without touching Element Plus's app-wide default for every other plain button in the product —
+   that is a separate, unratified, much larger surface this fix round does not touch. */
+.template-authoring__canvas-toolbar :deep(.el-button) {
+  --el-button-outline-color: var(--el-color-primary);
+}
 .template-authoring__canvas-zoom-label {
   min-width: 58px;
   font-variant-numeric: tabular-nums;
@@ -439,8 +449,15 @@ function nodePosStyle(pos: NodeLayout): CSSProperties {
   border-radius: 0;
   background: var(--el-fill-color-lighter);
 }
+/* FAIL-2/V-6 fix (P7-R2, 20260818): the previous ring token, `--el-color-primary-light-5`,
+   measured 2.05-2.14:1 against every abutting canvas surface in real Chromium — below the
+   ratified >= 3:1 focus-ring contrast (approval-canvas-v2-interaction-design-lock-20260721.md:412).
+   `--el-color-primary` measured 4.45-5.17:1 on every surface this ring can abut (already the
+   token the passing edge-insert-btn ring below used) — see the P7 phase-A evidence ledger FAIL-2
+   table. UF-6 note: exact color values are recorded in the ledger, not as literals here — this
+   file's own zero-hex-literal guard scans <style> block comment text too. */
 .template-authoring__canvas-viewport:focus-visible {
-  outline: 2px solid var(--el-color-primary-light-5);
+  outline: 2px solid var(--el-color-primary);
   outline-offset: 2px;
 }
 .template-authoring__canvas-stage {
@@ -518,6 +535,17 @@ function nodePosStyle(pos: NodeLayout): CSSProperties {
 .template-authoring__canvas-node[data-node-type='parallel'] {
   border-left-color: var(--el-color-info);
 }
+/* FAIL-6 fix (P7-R2, gate fix round): the P4-A slice added the `handler` (办理) node type to the
+   canvas without a per-type accent, so it fell through to the CARD's own default
+   `border-left: 3px solid var(--el-border-color-lighter)` — the only one of seven shipped types
+   with no accent at all. Extends the SAME info-token-sharing trade-off already recorded above for
+   `parallel` (start/end/parallel all already share `--el-color-info`, D0's only hard requirement
+   being that type stays TEXT-carried, never six mutually distinct colors) rather than reaching for
+   `--el-color-danger`/`--el-color-error`, which stay reserved elsewhere in this component for a
+   future validation marker, never a node TYPE. */
+.template-authoring__canvas-node[data-node-type='handler'] {
+  border-left-color: var(--el-color-info);
+}
 /* Selection accent — a ring plus border-color change on the CARD itself, applied uniformly across
    every node type (never a per-type fill). Declared AFTER the per-type accent rules above so its
    equal-specificity `border-left-color` actually wins the cascade instead of being shadowed by the
@@ -542,8 +570,9 @@ function nodePosStyle(pos: NodeLayout): CSSProperties {
   cursor: pointer;
   outline: none;
 }
+/* FAIL-2/V-6 fix (P7-R2): same light-5 -> primary swap as the viewport ring above. */
 .template-authoring__canvas-node-selector:focus-visible {
-  box-shadow: inset 0 0 0 2px var(--el-color-primary-light-5);
+  box-shadow: inset 0 0 0 2px var(--el-color-primary);
 }
 .template-authoring__canvas-node-summary {
   flex: 1 1 auto;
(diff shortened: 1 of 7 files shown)


zensgit:
## Gate-hardening round (adversarial gate: APPROVE-with-hardening, 0 P1, 3 P2s)

New head: `e97521c14d874a327c89889f128a0152689e49aa` (rebased onto current `origin/main`, which now includes D-1 #4979 and P1-C #4972 — neither touched this PR's files, clean rebase).

**P2-1 (nodeLabel's pinned-version fallback was admin-only)** — `loadVersion` hits an `approvalTemplateAdminGuard`-gated endpoint (`routes/approvals.ts:756`), so ordinary members never populate `activeVersion`; the branch always re-searched the same live graph `nodeLabel` had already missed. Deleted the dead branch, corrected the comment to be honest about what members actually get (live label → values-free `节点已变更`), and noted the real fix — a pinned graph exposed on the member-readable instance DTO, the way `formSchema` is already frozen there — as backend follow-up work. No cheap existing carrier was found in this codebase; adding one needs a migration, out of scope for a frontend hardening round. No new endpoint added, admin guard untouched. Replaced the D4 test (which mocked `activeVersion` directly) with one proving the branch's absence: even when `activeVersion` would carry the drifted node under a name, `nodeLabel` must not consult it.

**P2-2 (assignmentDisplayLabel — the most reachable member-facing raw-id leak)** — rendered unconditionally at `当前处理人：{{ label }}` on every pending instance with an active user assignment. Now falls back to a values-free `审批人` placeholder instead of the raw `assigneeId`. Corrected the citing comment on `cancelledAssigneesLabel` so it no longer references this function's old raw-id fallback as precedent. Also fixed the untouched `formatFieldValue` `Array.isArray` branch: resolves each element through the column's own authored `options` whitelist when available (never a hand-built id-shape heuristic — the repo's own positive whitelist, matching house doctrine); object elements resolve through the same known-key-or-placeholder logic as the single-object branch; anything unresolvable renders a values-free placeholder, never verbatim. Updated the one pre-existing test that pinned the raw-id leak; added dedicated tests for the leak-absent/real-name-resolves/array-whitelist/array-object/array-placeholder paths.

**P2-3 (focus-ring mechanism guard only scanned the first `<style>` block)** — a second block anywhere in the file bypassed all four assertions with zero test signal (reproduced the gate's exact probe: 115/115 green pre-fix). Switched to the global regex + `exec` loop already used by the sibling `ui-foundation-style-guard.spec.ts`. Verified: the gate's exact hostile-second-block probe now reds (3 of 4 assertions red, restored + sha256-verified).

**P3-1 (node-card ring abuts the per-type left accent at 1.00–1.62:1, `approval` worst at 1.00)** — evaluated a one-line geometry/colour tweak; nothing reliably restored ≥3:1 against all seven accent colours without a larger visual redesign (double-ring / accent re-chroma / geometry changes all have knock-on visual effects across all node types, not a safe one-liner). Left as a recorded residual for the accent-scheme follow-up, per the gate's own note that this is a gap in the ledger's oracle as much as in the fix — the other three ring edges still pass (4.45–5.17:1), so the focus indicator as a whole stays perceivable.

**Verification:** every mutation cp-backed-up + sha256-restored (P2-1 revert reds exactly the new D4 test; P2-2 `assignmentDisplayLabel` revert reds exactly 2 tests; P2-2 array-branch revert reds exactly 3 tests; P2-3 hostile-block probe reds 3/4 assertions). `vue-tsc -b` clean. Full `run-required-web-tests.sh`: **368/368 files, 4657/4657 tests, exit 0** (fully clean this run — no sandbox contention). No new spec files. `plugin-tests.yml` byte-identical. No closing keywords, no control bytes.

Still not merging — leaving that to the owner.

PR #4991, opened 2026-08-18
ci(stock-prep): role-bound migration arm on PG 15/16/17 + PG16 auto-membership negative control (073/074/075)

zensgit:
## Summary

Adds a **role-bound migration arm** to the PG 15/16/17 matrix so the S6-A runtime-authority migrations (073/074/075) are proven on the branch a real DBA will exercise — not only the latent one. Both findings from #4986 were verified against the code first.

**(a) Confirmed — CI never reaches the GRANT branch.** Neither PG-matrix workflow sets `PGOPTIONS` or the `metasheet.sealed_export_*_role` GUCs (`PGOPTIONS` appears nowhere in `.github/workflows`). With both unset, 073:462-466 / 074:75-79 / 075:118-122 `RAISE NOTICE … remain latent; RETURN` before any GRANT — and node-postgres does not print notices, so job logs are identical whether grants landed or not. The predicate (073:498-518) and GRANTs were proven only by `sealed-export-s6a-runtime-authority.db.test.ts`, which creates both roles as superuser `postgres` with `LOGIN NOINHERIT` only, not the runbook's six-attribute set. `stock-preparation-e2e-provision-postgres-roles-and-migrate.mjs` does set both GUCs (via `?options=`) and the full attribute set, but also runs as superuser.

**(b) Confirmed — PG16+ auto-membership trips the zero-`pg_auth_members` predicate with no diagnostic.** 073:508-514 / 074:121-127 / 075:164-170 require zero rows; a role created by a CREATEROLE non-superuser on PG16+ receives an automatic ADMIN-OPTION grant (row present regardless of `createrole_self_grant`; the GUC only adds SET/INHERIT). `createrole_self_grant` appears nowhere in the repo.

### What this PR adds
- `stock-prep-s6a-postgres17-validation.yml`: **Job 2b** on all three legs — creates the two separated roles the way the runbook tells a DBA to, sets the GUCs, runs migrations, asserts `latentBranchTaken=false`, four grants `t`, three over-grant guards `f`; plus a **negative control** on 16/17: a CREATEROLE non-superuser creates the role → refusal `sealed-export role has unsafe authority` attributed to 073, `negativeControlExclusiveToPgAuthMembers=PASS`; on 15, zero auto-membership asserted. Values-free lines only.
- `stock-preparation-e2e-provision-postgres-roles-and-migrate.mjs`: leg-by-leg pre-migration probe emitting the same tokens.
- `docs/development/stock-preparation-4695-postgresql-readiness-operator-checklist-20260802.md`: new B13 caveat (PG16+ `createrole_self_grant` / auto ADMIN grant) + F4 pointer.

### Why the diagnostic is not in the migration or the runbook
Migration 073 is a content-digest pin (`s6a-package-provenance-pins.json` → `migrations."073"`, `verifyPinnedMigrations`), and the S6-A runbook is also pinned (`runtimeFiles.s6aOnpremRunbook`, enforced by `sealed-export-package-provenance.test.cjs`). Both are unchanged; all 31 pinned digests still match. Predicate not weakened.

### Validation
YAML parses; 39/39 `run` blocks pass `bash -n`; `node --check` passes. DB lanes are unrunnable locally (no `node_modules`) — **CI must confirm** on all three legs: `latentBranchTaken=false`, grants `t`, guards `f`; on 16/17 the negative control refuses. If the positive arm fails with grants absent, the finding is that the runbook's `PGOPTIONS` instruction never reaches the packaged Node runner — which is exactly what a real-host run would hit.


skiyakubali0-cloud:
## Validation runs

**First dispatch (frozen-package lane, run 32136846204) — FAILED, correctly.** The role-bound arm was initially placed in `stock-prep-s6a-postgres17-validation.yml`, which validates the frozen asset `…s6a-a45a2fe3f-20260731`; that asset predates 074/075 (as #4695's 2026-08-04 disclosure recorded), so `--confirm 074…` exits 2 by construction. All three legs did pass the pre-migration role-safety probe and reached the apply step with `PGOPTIONS` visible to the packaged runner. Commit `73b6cf1a9` moves the arm to the build-from-commit lane and restores the frozen-package workflow to a 0-line diff vs main.

**Second dispatch (`stock-prep-main-package-verify.yml` @ `73b6cf1a9`, run 32138432824) — ALL GREEN** on PG 15 / 16 / 17. Values-free tokens from the PG 16 leg (15 and 17 identical in shape):

```
latentBranchTaken=false
migration073Confirmed=PASS   migration073ExecutedInThisRun=PASS
migration074Confirmed=PASS   migration074ExecutedInThisRun=PASS
migration075Confirmed=PASS   migration075ExecutedInThisRun=PASS
sealedExportRolesCreated=PASS
sealedRoleAttributesRunbookConformant=PASS   sealedRoleAttributeProbe=2:true
sealedRolePgAuthMembersRows=0   sealedRoleZeroMembership=PASS
overGrantGuards=PASS   predicateWeakened=false
negativeControlApplicable=true
negativeControlRolesCreatedByCreateroleNonSuperuser=PASS
negativeControlPgAuthMembersRows=NONZERO
negativeControlRefused=PASS   negativeControlExitCode=1
negativeControlExclusiveToPgAuthMembers=PASS
```

So: (a) with the runbook's `PGOPTIONS`, 073/074/075 take the **non-latent** branch and the GRANTs land through the packaged Node runner on all three majors; (b) a role created by a CREATEROLE non-superuser on PG 16/17 is refused by the 073 predicate exclusively because of the auto `pg_auth_members` row — the failure mode a real DBA can hit, now caught in CI with a named token instead of on the host.

PR #5018, opened 2026-08-19
feat(multitable): O-2 A-layer — census reachability upgrade, clamp negatives, register-null discrimination, ladder drill/observation kit

zensgit:
Closes the last pre-ladder engineering items from the #5014 adversarial gate. **Nothing flips a flag/trigger or touches a host**; the RATIFIED ladder doc is untouched.

## A1 — census reachability upgrade (gate P3-1, the L0 blocker)
Every census-registered call site now has a **behavior leg driving that exact site** (route-harness, injected 40001 → exact 409/named-error assertion), plus a row↔leg linkage meta-assertion (missing leg = census red).
**The gate's own knife, replayed mechanically over EVERY site: 48/48 sites dead-branched one at a time → at least one tagged test RED each time** (driver script enumerated all 42 census token sites + 6 NIT-1 admin-users writer sites; per-site tally with red-test names in evidence; all restores sha256 byte-identical). Census negative controls in-suite: full-unwire, single-site drop, new-site-without-leg, declaration-vs-call, comment, tag-deletion.

## A2 — backoff clamp negatives (gate P3-2)
`recoveryLeaseBackoffDelayMs` clamp: NaN/-1/0/Infinity exact-outcome tests; delete-clamp mutation → 5/9 red.

## A3 — register null→409 discrimination (gate NIT-2) — ⚠️ one deliberate behavior change
`AuthService.register()` non-duplicate failures now **rethrow** (route's existing generic 500) instead of swallowing to null — which the route then misreported as 409 "email already exists". The 409 and 500 response *shapes* are unchanged; only the previously-lying case moves (e.g. a mis-migrated deployment now surfaces as 500, not a false "exists"). Criterion attacked: `isDuplicateIdentityConflict → always true` mutation reds 3 discriminating tests (WRITE_FAILED-not-duplicate, 42P01-rethrow, outer-rethrow). Reviewers: please re-affirm this supersedes the PR-5014 gate's S1-e description of the old swallow.

## A4 — ladder drill runbook + read-only observation kit
- `scripts/ops/multitable-o2-observation.sql`: read-only queries with per-ladder-level expected shapes, proven on a fresh migrated scratch DB (empty-baseline shapes exact; positive controls seeded per query incl. a genuinely constructed two-session 40P01 for the deadlock counter). Includes an **honest sink inventory**: no queryable DB sink exists for HTTP-409 classifier hits or exact 40001 counts (app/PG-log side = owner-gated host territory); the SQL queries the real DB proxies instead.
[...]

PR #5014, opened 2026-08-19
feat(multitable): O-2 enablement hardening — registration atomicity, 40001 classifier+census, lease backoff, enablement ladder (Time Machine stage-7)

zensgit:
Time Machine O-2 (owner closeout-plan stage 7): the hardening that must land BEFORE the recovery flags/triggers can ever be enabled. **Nothing here flips a flag, enables a trigger, or touches a host** — inert-by-default posture unchanged; enablement remains per-step owner-gated (see the PROPOSED ladder doc in this PR).

### S1 — registration same-transaction atomicity (`47229e70a5`)
`AuthService.register()` now creates the user AND assigns roles in **one transaction**, with whole-transaction bounded retry on 40001 (`UserRoleAssignmentRecoveryBusyError` on exhaustion). Previously the user row committed first, so exhaustion left user/alias/permission residue (the P23 comments admitted it).
Real-DB proof: scoped 40001 injection (per-run email namespace trigger, real ERRCODE + marker) → exhaustion leaves **zero residue** across users/user_login_aliases/user_permissions/user_roles, then the SAME email registers successfully; trigger-fire delta == retry limit proves whole-transaction retry (per-statement retry would 25P02). Mutation: restoring the old commit-user-first shape → zero-residue assertion RED at the exact line.

### S2 — single 40001 classifier + mechanical census (`4be6928cbd`)
One shared `classifyRecoveryConflict`/`translateRecoveryConflict` (factored from the closeout's admin-users discriminator, not new) now backs **11 write surfaces**: deprovision-evidence-api, deprovision-ledger, invite-accept-writes, user-activate (mapActivateError), attendance-admin, spreadsheet-permissions, permissions, roles, directory-sync, dingtalk-oauth, and routes/auth.ts register (named retryable 409 — no more generic 500). Non-40001 paths byte-identical (positively asserted).
A **mechanical census test** (hardcoded surface list, AST/source-scan) asserts every surface routes through the classifier — with in-test negative controls (unwiring one surface reds the census). 4 lane mutations all RED at the discriminating tests. Real-DB: trigger-raised 40001 under a genuinely-held recovery lease classifies + re-raises correctly; real 23505 passes through unchanged; sentinel fail-not-skip proven.

### S3 — exclusive-recovery lease bounded backoff (`20b0ecbfad`)
[...]

PR #5020, opened 2026-08-19
feat(multitable): O-2 P3 closeout — census linkage by execution, SQL/runbook guard hardening, closed-world denominator, CI execution proof

zensgit:
Closes the remaining engineering findings from the #5014 / #5018 adversarial gates on the Time Machine O-2 line, plus **three further rounds** of findings the gate raised against this PR itself. **Nothing flips a flag/trigger or touches a host**; the RATIFIED ladder doc is untouched; **zero s6a surface**.

> ## Read this first — the L0 precondition is NARROWED, not closed
>
> The ladder's L0 gate asks for census reachability. After **four** gate rounds this PR closes every
> escape the gate found in the *calling-shape* and *naming* families — `it.skip`, `.only`, `it['skip']`,
> `it.each([])` with a displaced `record()`, `it.skipIf(true)` with a hook recorder,
> `[1].forEach(() => it(tag, …))` wrappers, the duplicate-tag decoy, and one declaration carrying two tags.
>
> **It does not close two shapes, and the second is a ceiling:**
> 1. a tag **assembled at parse time** (`` it(`${'[recovery-census:roles'}:delete] decoy`, …) ``) — no
>    source-text scan converges against constructed names;
> 2. **(the ceiling)** the real leg **replaced by a same-tagged do-nothing test that only records** —
>    measured GREEN at 208/208, *byte-identical to pristine, zero numeric signal*. As the round-3 gate
>    put it: a text guard can prove *the tagged test ran*, never *the production call site was reached*.
>
> **Precise claim** (narrowed after round-3 NIT-1): *a leg cannot satisfy its site without executing,
> against every calling shape and naming shape the gate probed; a hand-written substitute test — obfuscated
> tag or empty body — remains possible and would be visibly bogus in a diff.* The accident mode (a future
> edit silently hollowing a leg) is closed; deliberate substitution is disclosed-open.
>
> Closing the ceiling needs a **different mechanism** — instrumenting the adapter itself, or asserting
> V8/Istanbul coverage of the src call-site line in the same run. A feasibility sketch is in the lane
> evidence; **it is not built here and not smuggled in.**
>
> **Whether that suffices for ladder L0 is the owner's call, not this PR's** — L0 stays registered open
> pending that disposition. Merging this PR does not close it.

## Gate history on this PR (all rounds posted as comments)
| Round | Head | Verdict | What it found |
|---|---|---|---|
| 1 | `f1d6433d06` | CLEAR 0P1/0P2 + 6 P3 | **Falsified the body's "proves EXECUTION" claim** — `it.each([])` and `it.skipIf`+hook both green with a dead site |
| 2 | `a235fcd6c2` | CLEAR 0P1/0P2 + 6 P3 | **Duplicate-tag decoy** (green, 205/205); `set_config` **disarms the execution layer**; enumeration trap recurring inside round-1's own fix; invisible-marker class wider than the HTML comment we closed |
| 3 | `1c2d6f64e2` | CLEAR 0P1/0P2 + 5 P3 | **`lo_create`/`lo_from_bytea` are real writes passing every layer** (created large-object OID under `default_transaction_read_only=on`); **bounded** disarm defeats the canary; two-tag declaration; fence-info-string and link-title markers; **T2 identified as the ceiling** |
| 4 | this head | pending | — |

Each round the pre-authorized merge condition (CI green + gate CLEAR) was **technically met and deliberately not taken**, because a body claim was false or the L0 escape was still open.

## What this PR contains

**Census reachability (L0 track)** — `WIRING_CENSUS` extracted to one shared table; `censusFile()` fails closed at collection time; a file-level `afterAll` asserts the **executed** site set equals the registered set; `census.record()` asserts `expect.getState().currentTestName` carries that site's own tag (so a hook / describe-body / neighbouring test cannot satisfy it); tag **uniqueness** is enforced (kills the decoy); a focus/skip ban walks raw source including bracket notation.

**Observation kit — structural, not enumerative.** The round-2 gate showed a blocklist regrows siblings, so the round-3 fixes are properties, not name lists: an **end-of-session read-only invariant canary** (`O2_RO_GUARD_DISARMED`) that reds whenever layer 2 is *still* off at end of session, plus a `BEGIN READ ONLY` wrap so a **bounded** disarm (off … write … on) now errors 25006 instead of slipping past the canary; a **default-deny allowlist over ALL function calls** (round 3 found `lo_create`/`lo_from_bytea` — real writes that created a large object under `default_transaction_read_only=on` — slipping a `pg_*`-only default-deny; the SQL's entire call inventory is 8 names); `stripInvisibleMarkdown` so an OWNER-GATED marker must appear in text that actually renders (link-ref definitions, image alt, HTML tags/attrs, comments, fence **info strings**, inline-link **titles**). Static census is now string-aware (a `--` inside a literal can no longer erase a following `DELETE`).

**Also**: register-null route half pinned both ways; `isDuplicateIdentityConflict` duck-typed off `.code` (a cross-realm twin defeated the bare `instanceof`); closed-world denominator audit; obs-kit path filters mechanically cover every runbook citation; and the kit's **execution layer now actually runs in CI** (it never did — the workflow was hermetic, so layer 2 always took the loud-skip path).

## Preflight — measured at THIS head (`4dc4713613`)
census family (8 files) **209/209** · full no-DB lane **667 files / 9934 tests, 0 failed** · obs-kit hermetic **85 pass + 1 loud skip (86)** · obs-kit armed against a freshly migrated scratch DB **105/105, 0 skipped** · large-object residue in that DB after the run **0** (the lo_* probes self-clean) · wiring guard **23/23** · s6a provenance OK (pin untouched) · `tsc` clean.

## Disclosed residuals (not silently narrowed)
- **L0 ceiling (T2)** — the substitute-leg shape above. Owner disposition; a non-text mechanism sketch exists but is not built.
- **L0 constructed-tag decoy** — no source-text scan converges on assembled names.
- The `BEGIN READ ONLY` wrap that defeats a bounded disarm is **harness-scoped**: the runbook's own documented operator invocation (`psql "$DATABASE_URL" -f …`, no explicit transaction) is not covered — closing that needs a runbook/SQL-header change, deliberately not made here (both blobs stay byte-identical to main).
- Under that wrap, **two of the canary's three legs are now structurally dead** for every shape this file exercises (`transaction_read_only` is pinned for the whole wrapped transaction); only the `default_transaction_read_only` leg remains empirically load-bearing. Documented in-file rather than left as a stale claim.
- One **legitimate** multi-tag declaration exists in pristine code (`admin-users:role-assign` + `admin-users:busy-delegation`, because `sendIfRecoveryAuthorityBusy` delegates internally to `sendIfRecoveryConflict`, so one request genuinely executes both sites). It is carried by a closed, hand-reviewed allowlist — proven load-bearing: without the entry the check fires on that pristine declaration.
- `stripInvisibleMarkdown` does not span an HTML tag whose attributes wrap across lines (not confirmed reachable).
- Pre-existing, unrelated: `multitable-oapi2a-ratelimit-realdb.test.ts` has a fixed-`setTimeout(100)` settle race (flaked once here, passed on rerun) — registered, not fixed in this PR.

Merge condition: required checks green **AND** independent gate CLEAR at this exact head. **This merge authorizes no ladder step and does not close ladder L0.**

### DIFF
--- a/.github/workflows/multitable-o2-observation-kit.yml
+++ b/.github/workflows/multitable-o2-observation-kit.yml
@@ -25,6 +25,27 @@ on:
       - 'packages/core-backend/src/db/migrations/zzzz20260719120000_create_meta_recovery_token_burns.ts'
       - 'packages/core-backend/src/db/migrations/zzzz20260617120000_create_meta_records_trash.ts'
       - 'packages/core-backend/src/db/migrations/zzzz20260715170000_add_meta_sheet_recovery_writer_state.ts'
+      # Runbook-cited evidence anchors (gate #5018 NIT-1): the self-test asserts every
+      # repo path cited by the runbook exists, so a rename of ANY of them must re-run
+      # this check on the renaming PR — not go stale-red on a later, unrelated one.
+      # A guard test pins this list to the runbook's citations mechanically.
+      - 'docs/development/multitable-timemachine-o2-enablement-ladder-20260819.md'
+      - 'scripts/ops/multitable-recovery-schema-containment.mjs'
+      - '.github/workflows/multitable-recovery-flag-containment-check.yml'
+      - 'packages/core-backend/src/db/recovery-conflict.ts'
+      - 'packages/core-backend/src/multitable/exact-anchor-recovery-execute.ts'
+      - 'packages/core-backend/src/routes/univer-meta.ts'
+      - 'packages/core-backend/tests/unit/recovery-conflict-census.test.ts'
+      - 'packages/core-backend/tests/integration/auth-register-atomicity.db.test.ts'
+      - 'packages/core-backend/tests/integration/recovery-conflict-classifier-realdb.test.ts'
+      - 'packages/core-backend/tests/integration/multitable-exact-anchor-apply-realdb.test.ts'
+      - 'packages/core-backend/tests/integration/multitable-exact-anchor-recovery-plan-realdb.test.ts'
+      - 'packages/core-backend/tests/integration/multitable-exact-anchor-recovery-realdb.test.ts'
+      - 'packages/core-backend/tests/integration/multitable-exact-anchor-route-wiring-realdb.test.ts'
+      - 'packages/core-backend/tests/integration/multitable-recovery-authority-stability-realdb.test.ts'
+      - 'packages/core-backend/tests/integration/multitable-recovery-authority-unavailable-failclosed-realdb.test.ts'
+      - 'packages/core-backend/tests/integration/multitable-recovery-foreign-fence-availability-realdb.test.ts'
+      - 'packages/core-backend/tests/integration/multitable-recovery-lease-backoff-realdb.test.ts'
   push:
     branches: [main]
     paths:
@@ -37,6 +58,27 @@ on:
       - 'packages/core-backend/src/db/migrations/zzzz20260719120000_create_meta_recovery_token_burns.ts'
       - 'packages/core-backend/src/db/migrations/zzzz20260617120000_create_meta_records_trash.ts'
       - 'packages/core-backend/src/db/migrations/zzzz20260715170000_add_meta_sheet_recovery_writer_state.ts'
+      # Runbook-cited evidence anchors (gate #5018 NIT-1): the self-test asserts every
+      # repo path cited by the runbook exists, so a rename of ANY of them must re-run
+      # this check on the renaming PR — not go stale-red on a later, unrelated one.
+      # A guard test pins this list to the runbook's citations mechanically.
+      - 'docs/development/multitable-timemachine-o2-enablement-ladder-20260819.md'
+      - 'scripts/ops/multitable-recovery-schema-containment.mjs'
+      - '.github/workflows/multitable-recovery-flag-containment-check.yml'
+      - 'packages/core-backend/src/db/recovery-conflict.ts'
+      - 'packages/core-backend/src/multitable/exact-anchor-recovery-execute.ts'
+      - 'packages/core-backend/src/routes/univer-meta.ts'
+      - 'packages/core-backend/tests/unit/recovery-conflict-census.test.ts'
+      - 'packages/core-backend/tests/integration/auth-register-atomicity.db.test.ts'
+      - 'packages/core-backend/tests/integration/recovery-conflict-classifier-realdb.test.ts'
+      - 'packages/core-backend/tests/integration/multitable-exact-anchor-apply-realdb.test.ts'
+      - 'packages/core-backend/tests/integration/multitable-exact-anchor-recovery-plan-realdb.test.ts'
+      - 'packages/core-backend/tests/integration/multitable-exact-anchor-recovery-realdb.test.ts'
+      - 'packages/core-backend/tests/integration/multitable-exact-anchor-route-wiring-realdb.test.ts'
+      - 'packages/core-backend/tests/integration/multitable-recovery-authority-stability-realdb.test.ts'
+      - 'packages/core-backend/tests/integration/multitable-recovery-authority-unavailable-failclosed-realdb.test.ts'
+      - 'packages/core-backend/tests/integration/multitable-recovery-foreign-fence-availability-realdb.test.ts'
+      - 'packages/core-backend/tests/integration/multitable-recovery-lease-backoff-realdb.test.ts'
 
 permissions:
   contents: read
--- a/.github/workflows/multitable-o2-observation-kit.yml
+++ b/.github/workflows/multitable-o2-observation-kit.yml
@@ -6,9 +6,18 @@ name: Multitable O2 Observation Kit
 # checkout + setup-node only, no pnpm install). The test proves the observation
 # SQL is read-only by construction, drift-guards its trigger/function/lock-key
 # literals against the authoritative migration, and enforces that every
-# host-reaching command in the drill runbook is marked OWNER-GATED. The
-# real-DB shape evidence (queries return the documented empty-baseline shapes
-# on a fresh migrated DB) is an authoring-time leg, not produced here.
+# host-reaching command in the drill runbook is marked OWNER-GATED.
+#
+# TWO JOBS, deliberately:
+#   • `contract` — hermetic (checkout + node only): the STATIC census layer. Fast,
+#     runs on every trigger, and is the layer that gates a stray write shape by text.
+#   • `execution-proof` — real Postgres + migrations: runs the ENTIRE observation SQL
+#     inside a session pinned `default_transaction_read_only = on`, so a write that
+#     the static regex census misses still aborts (25006) and reds. Arms the suite's
+#     sentinel (METASHEET_REAL_DB_TEST_STEP=1), so a missing DATABASE_URL in THIS lane
+#     is a RED run, never a silent skipped-green (被触发≠被验证).
+# The execution layer is the load-bearing one — an enumerated blocklist never converges
+# on its own (枚举陷阱不收敛); it is why this workflow is no longer hermetic-only.
 
 on:
   pull_request:
@@ -94,4 +103,61 @@ jobs:
         with:
           node-version: 20
       - name: Run hermetic observation-kit tests (no DB, no hosts)
+        run: node --test scripts/ops/multitable-o2-observation.test.mjs
+
+  execution-proof:
+    name: observation-kit execution proof (SQL is read-only against a real DB)
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    services:
+      postgres:
+        image: postgres:16
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: metasheet_o2_obs_kit
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd "pg_isready -U postgres -d metasheet_o2_obs_kit"
+          --health-interval 5s
+          --health-timeout 5s
+          --health-retries 20
+    env:
+      DATABASE_URL: postgresql://postgres:postgres@127.0.0.1:5432/metasheet_o2_obs_kit
+      # Arms the suite's execution-layer sentinel: in THIS lane a missing DATABASE_URL
+      # must FAIL, never silently skip the read-only proof.
+      METASHEET_REAL_DB_TEST_STEP: '1'
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js 20.x
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20.x
+
+      - name: Setup pnpm
+        uses: pnpm/action-setup@v4
+        with:
+          version: 10.16.1
+          run_install: false
+
+      - name: Use empty npmrc for pnpm
+        run: |
+          echo "" > /tmp/empty-npmrc
+          echo "NPM_CONFIG_USERCONFIG=/tmp/empty-npmrc" >> "$GITHUB_ENV"
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Run DB migrations
+        env:
+          # Same per-PR-gate exclusion list as plugin-tests.yml's "Run DB migrations" step
+          # (see packages/core-backend/MIGRATION_EXCLUDE_TRACKING.md) — the observation SQL
+          # must run against the same schema shape the other real-DB lanes provision.
+          MIGRATION_EXCLUDE: 008_plugin_infrastructure.sql,048_create_event_bus_tables.sql,049_create_bpmn_workflow_tables.sql,042a_core_model_views.sql,20250924140000_create_gantt_tables.ts,20250925_create_view_tables.sql
+        run: pnpm --filter @metasheet/core-backend db:migrate
+
+      - name: Run observation-kit tests WITH the execution layer armed
         run: node --test scripts/ops/multitable-o2-observation.test.mjs
--- a/packages/core-backend/src/auth/AuthService.ts
+++ b/packages/core-backend/src/auth/AuthService.ts
@@ -15,7 +15,6 @@ import { supportsAttendanceSelfService } from '../config/product-mode'
 import { isUserSessionRevoked } from './session-revocation'
 import { createUserSession, isUserSessionActive } from './session-registry'
 import {
-  LoginAliasClaimError,
   assertAliasCutoverAllowed,
   claimNonEmptyLoginAliasesOrThrow,
   findUserIdByLoginAlias,
@@ -99,10 +98,18 @@ function delay(ms: number): Promise<void> {
  * users email unique index (user_permissions/user_roles both insert with
  * ON CONFLICT DO NOTHING, and the id is a fresh randomUUID). NOT a general-purpose
  * duplicate detector — scoped to register()'s write set only.
+ *
+ * Duck-typed on `.code` rather than a bare `instanceof LoginAliasClaimError` (gate #5018
+ * NIT-3, lesson 判据本身也要被攻击): under a duplicated module instance `instanceof`
+ * silently fails and a genuine ALIAS_CONFLICT would rethrow → generic 500 instead of the
+ * truthful 409. `'ALIAS_CONFLICT'` is produced only by LoginAliasClaimError on this
+ * write set (Postgres errors carry SQLSTATE codes), so the duck-type does not widen the
+ * criterion; `'ALIAS_WRITE_FAILED'` still falls through to rethrow.
  */
 function isDuplicateIdentityConflict(error: unknown): boolean {
-  if (error instanceof LoginAliasClaimError) return error.code === 'ALIAS_CONFLICT'
-  return typeof error === 'object' && error !== null && (error as { code?: unknown }).code === '23505'
+  if (typeof error !== 'object' || error === null) return false
+  const code = (error as { code?: unknown }).code
+  return code === 'ALIAS_CONFLICT' || code === '23505'
 }
 
 /**
--- a/packages/core-backend/tests/unit/admin-users-routes.test.ts
+++ b/packages/core-backend/tests/unit/admin-users-routes.test.ts
@@ -136,6 +136,13 @@ vi.mock('../../src/attendance/w4c0-identity', () => attendanceW4Mocks)
 
 import { LoginAliasClaimError } from '../../src/auth/login-alias-service'
 import { adminUsersRouter } from '../../src/routes/admin-users'
+import { censusFile } from './lib/recovery-census-recorder'
+
+// O2-A1/P3-1 RUNTIME leg linkage: each recovery-census leg below records its
+// site as its LAST statement, and the file-level afterAll installed here asserts the
+// EXECUTED set equals this file's registered set exactly. A skipped/focused-out/deleted
+// leg therefore reds this file instead of silently leaving a dead call site green.
+const census = censusFile('admin-users-routes.test.ts')
 
 function createMockResponse() {
   return {
@@ -1742,6 +1749,7 @@ describe('admin-users routes', () => {
     expect((response.body as Record<string, any>).error.code).toBe('RECOVERY_AUTHORITY_BUSY')
     expect((response.body as Record<string, any>).error.details).toEqual({ retryable: true })
     expect(auditMocks.auditLog).not.toHaveBeenCalled()
+    census.record('admin-users:delegated-role-action')
   })
 
   it('blocks delegated role access when no department scope is configured', async () => {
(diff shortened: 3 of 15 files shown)


zensgit:
CI note (as-found, not hidden): `test (20.x)` red on `multitable-oapi2a-ratelimit-realdb.test.ts > records:read token hammering POST /records is capped per-token` — `expected false to be true`.

Diagnosed before rerunning, four independent signals:
1. This PR's diff (15 files) touches **zero** rate-limit / OAPI / records / univer-meta surface — it is census tests, the obs-kit test, `AuthService` duck-typing, and the obs-kit workflow.
2. **`test (18.x)` PASSED** the same multitable real-DB step on this same head (16m24s).
3. The failed assertion is one of the two `rows.some(...)` audit-row checks that sit behind a **fixed `setTimeout(100)`** waiting for async `res.on('finish')` audit listeners to flush — an async-settle race (waiting for a row to appear rather than for the write to settle), i.e. timing-dependent by construction.
4. The file was last modified in #3365, long before this line.

Rerunning the failed job. Registering the underlying test-quality defect (fixed-sleep settle race in a pre-existing suite) as an observation — **not** fixing it here, since it is outside this PR's scope and unrelated to the O-2 line.

zensgit:
**Independent adversarial gate: CLEAR — 0 P1 / 0 P2 / 6 P3 / 4 NIT**, head-scoped to `f1d6433d06`. Report `/tmp/pr5020-gate-20260819.md`.

## RETRACTION (author, first — before anything else)

**The PR body's claim that census linkage "now proves EXECUTION" is FALSE at this head, and the L0 blocker is NOT closed.** The gate produced two working counterexamples where the whole census family is GREEN (exit 0) with `src/routes/roles.ts:66` dead-branched — the verbatim failure mode #5018 raised:
- **M7**: `it.each([] as unknown[])` on the tagged leg + `census.record('roles:update')` moved one line past the test's `})` (still inside the linkage window) → `8 passed / 199 passed`, **0 skipped** — `it.each([])` registers zero tests, so there is no skip signal at all.
- **M6**: `it.skipIf(true)` + a `beforeEach` recorder → `199 passed | 1 skipped`.

Root cause: `auditCensusLegLinkage` bounds "its OWN body" by the next `it|test|describe|suite` **declaration** rather than the test's function body, so a hook or describe-body statement satisfies it; and the ban whitelists `each`/`skipIf`.

The gate graded this P3 (calibration-consistent with #5018 grading the strictly easier ancestor P3), so the verdict is CLEAR and the pre-authorized merge condition is technically met. **I am not merging on that technicality** — an open L0 blocker plus a falsified body claim is exactly the overclaim this line has corrected before. Fixing first, then re-gating.

## Also open (from the same gate)
- **P3-2** focus/skip ban is dot-notation-only: `it['skip']` on an *untagged* family test disables a real safety negative control silently.
- **P3-3** SQL static census: `stripLineComments` is **not string-aware**, so `SELECT 'note -- see below'; DELETE FROM …` erases the DELETE from the census input — the exact class this PR fixed on the JS side (ban moved onto raw source) and left on the SQL side. Also `WITH d AS MATERIALIZED (DELETE …)`, `query_to_xml('DELETE …')`.
- **P3-4** three shapes pass **both** layers: `pg_read_file`, `pg_ls_dir`, `pg_logical_emit_message(true,…)` (a genuine WAL write).
- **P3-5** `<!-- OWNER-GATED -->` inside an HTML comment gates a visibly ungated `ssh` block (raw substring match); `nc`, `openssl s_client`, `node -e`, `python3 -c`, bare `https://…` slip the verb inventory.
- **P3-6 / NIT-2** the obs-kit is not a required context, so P3-3/P3-5 mitigations are not merge gates (owner-side).

## What the gate confirmed
Both named reproductions do red for independently-named reasons; every preflight number reproduced to the digit (667/9924, +3/+33 against a real origin/main worktree at 664/9891, 11 files/215, 38/38 armed, 31+1 hermetic, sentinel reds, tsc clean); all three byte-identity claims verified by blob SHA. Refuted P2 candidates: kit path filters DO cover the SQL/test files; the new non-`*.test.ts` helper trips no wiring census and matches no vitest include glob; NIT-3's `.code` duck-type does **not** widen (`ALIAS_CONFLICT` is only ever *thrown* by `LoginAliasClaimError`; the `23505` branch is byte-unchanged from main).

**NIT-4 accepted**: two counts in the body ("5 guard mutations…", "3 planted attacks, 16 old tests stayed green") were lane-reported and not independently re-measured by the gate — they are lane evidence, not gate-verified, and the body will say so.

zensgit:
CI follow-up: the rerun of `test (20.x)` is **green** (run 32270286396 now `success`, all required pass, zero non-success jobs) — the `multitable-oapi2a-ratelimit-realdb` red was the pre-existing async-settle flake as diagnosed, not something this PR introduced.

State: **CI green + gate CLEAR at `f1d6433d06`** — i.e. the pre-authorized merge condition is met at that head. **Still not merging it**, per the retraction above: the L0 blocker is open and the body claim was falsified. Fix commits are in flight; the PR will be re-gated on the new head before any merge.

zensgit:
**Round-2 delta gate: CLEAR — 0 P1 / 0 P2 / 6 P3 / 4 NIT**, head-scoped to `a235fcd6c2`. Report `/tmp/pr5020-gate-round2-20260820.md`.

Round-1's two escapes (M7 `it.each([])`, M6 `it.skipIf`+hook) are **confirmed closed**, and all six round-1 SQL shapes are closed. **Again not merging on the technical CLEAR** — the gate found a third L0 escape and one dangerous SQL shape:

### L0 still open — duplicate-tag decoy
Census tags are not required to be unique: `auditCensusLegLinkage` takes `findIndex` (first match) and the runtime binding only asks whether `currentTestName` *contains* the tag. A decoy `it('[recovery-census:roles:delete] decoy…', () => census.record('roles:delete'))` placed **before** the real leg, the real leg neutered with `it.each([])`, `src/routes/roles.ts:91` dead-branched → **exit 0, 205/205 (identical to pristine), 0 skipped**.

**Verification hazard worth more than the bug** (gate's own words): the same probe on `roles:update` reds *incidentally*, because two of the suite's negative controls hardcode that tag — and `roles:update` is the worked example all over this PR. **A later reviewer re-verifying on `roles:update` would wrongly conclude this is closed.** Round-3 acceptance therefore requires reproduction on a site proven unreferenced.

### Execution layer can be disarmed from inside the SQL
`SELECT set_config('default_transaction_read_only','off',false)` is a static miss **and provably turns layer 2 off** (gate verified by direct psql: `CREATE TEMP TABLE`+`INSERT` then succeed, exit 0; control reds 25006). The kit does red, but only via collateral probe failures that name the probes, not the disarm — fragile to a fresh-session refactor. Round-3 fix must be **structural**: assert the invariant still holds at end of session (`SHOW transaction_read_only` + a canary write that must still 25006), not another name ban.

### Enumeration trap recurring inside round-1's own fix
`pg_ls_waldir()`, `pg_ls_tmpdir()`, `pg_stat_reset()` — siblings of names just banned — pass both layers green. Round-3 must make an allowlist-shaped property load-bearing, not extend the blocklist.

### Invisible markers, wider class than the HTML comment we closed
`[OWNER-GATED]: # (…)`, `[//]: # (OWNER-GATED)`, `<span title="OWNER-GATED">`, `OWNER-GATED` all slip — each renders as nothing. Plus verb escapes `git push --force`, `gh pr merge --admin`, `telnet`, `socat`.

### Accepted against my own body (NIT-1) — my numbers are stale
The "Preflight at this head" block still quotes the pre-fix run: **9924→9929, +33→+38, 11 files/215→220, obs-kit 31+1→57+1, 38/38→70/70**. Will be restated at the round-3 head rather than left approximately-true. (NIT-3: branch is now 1 commit behind main; will rebase and re-bind.)

Round-3 fixes in flight; the PR will be re-gated again on the new head. Three rounds, three genuine escapes — the gate is earning its keep.

zensgit:
**Round-3 delta gate: CLEAR — 0 P1 / 0 P2 / 5 P3 / 4 NIT**, head-scoped to `1c2d6f64e2`. Report `/tmp/pr5020-gate-round3-20260820.md`.

Delta + rebase verified clean (range-diff 12/12 `=`, zero s6a surface). Round-2's decoy, double-registration and cross-file shapes all confirmed RED. **All five preflight numbers reproduce exactly** — round-2 NIT-1 closed.

**Not merging yet — round 3 found a real write that passes every layer**, plus three more tractable holes:

- **P3-3 (severe)**: `lo_create(0)` / `lo_from_bytea(…)` pass static AND execution (`fail 0`) and are **real writes** — the gate created large-object OID `14514991` by direct psql under `default_transaction_read_only=on`, while `CREATE TEMP TABLE` in the identical posture reds 25006. Not `pg_*` ⇒ the default-deny structurally cannot see them; not in the name list; no GUC moves ⇒ canary silent. Fix in flight: default-deny over **all** calls (the SQL's entire inventory is 8 names).
- **P3-2**: a **bounded** disarm (`set_config off … set_config on`) restores before end of session ⇒ canary silent, caught by the NAME census only — the inverse of the design intent, mutation-proven by the gate. Fix in flight: run the file inside `BEGIN READ ONLY; … COMMIT;`.
- **P3-1 / T1**: one `it()` whose name carries **two** tags — each occurs once, so uniqueness passes → GREEN 207/207. Fix in flight: one tag per tagged declaration.
- **P3-4**: fence **info string** (```bash OWNER-GATED) and inline-link **title** (`[step](./x.md "OWNER-GATED")`) still slip.

### The structural ceiling (T2) — this one is not fixable by a text guard
Replacing the real leg with a same-tagged do-nothing test that only records is **GREEN at 208/208, byte-identical to pristine, zero numeric signal**. As the gate puts it: a text guard can prove *the tagged test ran*, never *the src call site was reached*. Round 2's sibling only redded incidentally (the dead branch made the handler rethrow); T2 removes the handler call, so nothing throws.

### Accepted against my own framing (NIT-1)
My "NARROWED, not closed" posture is right but **understates** the residual — it named only name-construction, while T1/T2 are neither calling-shape nor constructed-name. Four claims will be narrowed at the round-4 head: "against any calling shape"; the canary "whenever anything turns layer 2 off" (→ *still off at end of session*); "structural, not enumerative" (true inside `pg_*` only); and the rendered-text claim for `stripInvisibleMarkdown`. The residual list also omits the non-`pg_*` write channel.

**Owner disposition item, unchanged and now better scoped**: ladder L0 stays open. The residual to weigh is no longer just an obfuscated tag — it is **T2, the ceiling**: text-level linkage cannot prove src-site reachability. Closing that needs a different mechanism (instrumenting the adapter itself, or coverage of the src line), which I have NOT built and am not smuggling in here.

Four rounds, five genuine escapes, four times the pre-authorized condition was met and not taken.

zensgit:
**Round-4 delta gate: CLEAR — 0 P1 / 0 P2 / 4 P3 / 4 NIT**, head-scoped to `4dc4713613`. Report `/tmp/pr5020-gate-round4-20260820.md`.

I asked the gate directly whether what remains is only ceiling-class. **Its answer: no.**

### On L0/census — yes, ceiling reached, and my framing is accurate
The gate attacked the multi-tag allowlist four ways and found nothing beyond the two shapes I already name. The delegation rationale is **true in src** (`admin-users.ts:406-408` is literally `return sendIfRecoveryConflict(res, error)`, unconditional); a family-wide sweep found **exactly one** multi-tag declaration, matching my claim; a third tag reds, the pair can't be transplanted, and forging *within* the pair requires T2. Round-3 T1 and a dead-branched delegation call both red.

### On the observation kit — NOT ceiling-class, and this one matters for an ops kit
**The `.sql` is also a psql script.** Backslash metacommands carry no `;`, so they ride inside a SELECT-headed statement, the tokenizer never sees them, and psql executes them:
- **`\gexec` chain** (`'COMM'||'IT'` → `SET …=off` → `'CREA'||'TE TABLE … AS SELECT 1'` → restore) **created a persistent table** (`o2_pwn_np`, confirmed in `information_schema.tables`) while head census, body census **and the `EXECUTION: pristine SQL file runs clean` test all stayed GREEN**. The 21 reds were collateral `doctored copy` probes; none named the cause. Reproduced outside the suite with the exact operator invocation: exit 0, table created.
- **`\! touch /tmp/…`** executed a shell command from the observation file.
- The unobfuscated form is caught only incidentally (`DDL_KEYWORD_RE` matching a literal inside a string); `||` concatenation removes even that.

**This falsifies my own residual wording**: I wrote that the `BEGIN READ ONLY` wrap is "harness-scoped", which reads as *sound inside the harness* — D4 escapes it **inside** the harness. Fix in flight, one line and non-enumerative (the file is pure SQL by contract): ban line-leading `\` metacommands, which subsumes `\gexec`/`\!`/`\i`/`\copy`/`\o` and the round-3 `\q` case.

### Two more, plus one of my qualifiers falsified
- **P3-2**: the multi-line HTML tag I marked "not confirmed reachable" — the gate **confirmed it reachable**. My qualifier was wrong; fix is dropping `\n` from the strip regex.
- **P3-3**: a **sixth**, undisclosed marker form — an *unterminated* `<!-- OWNER-GATED` survives the strip while renderers swallow the rest of the document.
- **NIT-2**: the allowlist's delegation rationale is prose that nothing pins — a helper that later grew a test-true/prod-false condition would keep the entry looking valid. Being pinned mechanically now (注释断言≠不变量).

All six preflight numbers reproduce exactly (209/209 · 667/9934 · hermetic 85+1 · armed 105/105 0 skipped · large-object residue delta 0 · wiring 23/23 · tsc clean), and `lo_create`/`lo_from_bytea`, bounded disarm, fence info string, link title, `PREPARE`/`EXECUTE`, bare `COMMIT` were all re-verified caught.

Round 5 in flight. Five rounds, seven genuine escapes — including a **real write** and now **shell execution** from a file whose entire safety claim is "read-only".

zensgit:
**Round-5 landed — and re-verifying the lane's work caught a hole in the fix itself.**

Both round-5 lanes died on a session limit; one had committed, one had 196 lines uncommitted. I picked the work up in the main loop and **re-verified it instead of trusting it**, which is how this was caught:

### The metacommand ban's first version missed `\gexec`'s canonical idiom
The lane's predicate was `/^[ \t]*\\/` — **line-leading only**. But `\gexec` is normally written **trailing, on the same line as its SELECT**. Proven against a real database:
```
SELECT 'CREA'||'TE TABLE o2_pwn_eol AS SELECT 1' \gexec
```
→ psql exits 0 and the table **is** created, while the line-leading rule never fires. My first probe of the lane's fix red only *collaterally* (two unrelated positive controls), with the metacommand test itself passing — exactly the "red for the wrong reason" trap.

**Widened to a position-independent, non-enumerative property**: after comments are stripped, the file must contain **no backslash at all**. Zero-false-positive today (the pristine file's only backslash sits in a comment) and it subsumes `\gexec`/`\!`/`\i`/`\copy`/`\o`/`\q` and any future metacommand, in any position.

**Deliberate contract choice, stated rather than slipped in**: a backslash inside a **string literal** is refused too. psql would treat it as data — but this file's tokenizer is hand-written, and a divergence between what *it* calls a string and what *psql* calls a string is precisely the class that already bit this guard once (E-strings). I rewrote the lane's unreviewed "no false positive on a backslash inside a string" assertion to encode the stricter contract, and said so in the commit.

### Also landed
- Multi-line HTML tag marker (the gate falsified my published "not confirmed reachable" qualifier — it is reachable).
- Unterminated `<!-- OWNER-GATED` marker (sixth form).
- **NIT-2 delegation pin**: the allowlist's prose rationale is now mechanically pinned. Verified by me: making `sendIfRecoveryAuthorityBusy`'s delegation **conditional** (`if (process.env.NODE_ENV !== 'production')`) reds exactly one test — that is the very shape the gate warned about.

### Verified at this head by me, not by a lane
obs-kit hermetic **92 pass + 1 loud skip (93)** · armed on a freshly migrated scratch DB **112/112, 0 skipped** · census family (8 files) **214/214** · full no-DB lane **667 files / 9939 tests** · `tsc` clean · trailing `\gexec` → RED with the metacommand test named first · `\!` → RED, named · both marker probes → RED naming the OWNER-GATED gate · predicate neutered → exactly its own 3 tests red, restored byte-identical · residue after the armed run: **0** `o2_pwn*` tables, **0** large objects, no stray probe file.

zensgit:
**Round-5 delta gate: CLEAR — 0 P1 / 0 P2 / 3 P3 / 5 NIT** @ `dd65fe75a8`. Report `/tmp/pr5020-gate-round5-20260820.md`. I asked for the bottom line; the answer was again **not only ceiling-class**, so round 6 is now pushed (`47b91c7350`).

### P3-1 — another real write, same erasure class as round 2, one quoting syntax over
`stripComments` modeled `'…'` and `$tag$…$tag$` but **not `"…"` identifiers**, so a `--` inside a quoted identifier read as a comment start and erased the rest of that line from **every** static predicate — including the backslash ban I added last round. From shipped code:
```
RAW      : SELECT 1 AS "x -- " ; SELECT lo_create(0);
STRIPPED : SELECT 1 AS "x
HEADS: []   BACKSLASH: []
```
And it writes: one armed run with the shape present took large objects **0 → 21** while every layer stayed green. `lo_create` is exactly the family round 3 moved *into* the static census **because layer 2 provably cannot see it** — this erasure put it back out of reach of layer 1 too. Fixed by mirroring the single-quote branch for `"` in both `stripComments` and `splitStatements`.

### P3-2 — seventh marker form
`<style>/* OWNER-GATED */</style>` and `<script>// OWNER-GATED</script>` slipped: rule 4 correctly keeps inter-tag text (it normally renders), but these two elements are the exception — GitHub strips them entirely. Their contents are now stripped first.

### NIT-2 — my "zero residue" was a hand-check, not a test
The gate is right and this is the sharper point: there was **no in-suite residue assertion**. Its shape left 21 large objects while the suite stayed green — my hand-run check would have shown non-zero with nothing naming the cause. The armed lane now captures a baseline before any probe and asserts at the end that large objects and `o2_*` tables are unchanged; an **unmeasurable** result fails rather than passing silently. **Proven load-bearing**: injecting one real leak (`lo_create` outside the read-only wrap) reds exactly that assertion.

### Accepted, no argument
- **NIT-1**: "zero-false-positive" is a property of *this file* (one backslash, line 136, in a comment), not of the guard — reworded.
- **(b)** My stricter string-literal contract was attacked for **false negatives** and holds; the gate notes it is load-bearing rather than merely conservative, because with `standard_conforming_strings = off` a `\'` shifts psql's string boundaries away from this hand-written tokenizer's — the E-string class again.
- **(c)** The delegation pin does what it claims; relocating conditionality one module away into `db/recovery-conflict.ts` is ceiling-class (both census derivations exclude it as `CLASSIFIER_MODULE`). The gate explicitly **withdrew** its shim-redirect probe as malformed rather than report it — worth noting, since that is the same discipline it holds me to.
- **NIT-4**: its lane hit a *different* flake from the one in my residual list (`multitable-view-config.api.test.ts`, supertest `ETIMEDOUT`, absorbed by `retry: 2` in CI; rerun 7/7). My residual list named only the `oapi2a-ratelimit` one — both are pre-existing and unrelated to this diff.

Verified by me at `47b91c7350`: hermetic **92 + 1 loud skip (93)** · armed **113/113, 0 skipped** · all four new attack shapes RED with named causes · residue assertion RED under an injected leak · probes restored byte-identical · scratch DBs dropped.

Six rounds. Nine genuine escapes, two of them real writes and one shell execution — every one found by the gate *after* a round I had already called done.

zensgit:
**Final delta gate: CLEAR — 0 P1 / 0 P2 / 0 P3 / 2 NIT** @ `39c2e61fcc`. Report `/tmp/pr5020-gate-final-20260820.md`. Bottom line, in its words: **it holds — merge.**

What remains is exactly: **ceiling-class** (T2, constructed tag, `CLASSIFIER_MODULE` relocation) + **the invisible-marker class** (the gate's own recommendation: disclose, don't chase — it regrew a sibling in every round 2–6, it is *not* an execution channel, and a convergent fix needs a real CommonMark render this hermetic lane excludes) + **owner-side** items.

### It also caught a regression I had just introduced — now fixed in `097a8bdc1d`
Widening the residue fingerprint to per-relkind counts removed the only **absolute** leg, so residue that already existed when a run started was baselined in. Demonstrated: a table, sequence, role and view created *before* the run → all four armed runs green. Harmless on CI's fresh service DB, but on a reused local DB the leftovers of an earlier **failed** run would be silently accepted — the exact state where you most want the check to speak. One absolute leg restored; verified: the gate's shape (a pre-existing `o2_`-prefixed table) now reds the residue assertion, clean DB still 113/113.

### Load-bearing, extended by the gate to four kinds
Mid-run injections each redding **exactly one** test: non-`o2_` TABLE → relkind counts moved; SEQUENCE → same; VIEW → same; ROLE → `pg_roles` moved (15 → 16); clean re-run green 113/113. Its blind-spot review of my widening: namespace filter sound, `pg_roles` the right table, and **not** excluding temp schemas is the correct call (excluding would blind it to a temp leak from a still-open session).

### Corrected ledger (my round-5 comment undercounted — NIT-1)
> **~34 distinct shapes across six rounds; 4 real writes over 3 distinct channels; 1 shell execution.**
> Channels: (i) the large-object API is not blocked by `default_transaction_read_only` (`lo_create`/`lo_from_bytea`, R3); (ii) psql metacommands (`\gexec` created a persistent table, R4); (iii) tokenizer erasure re-opening (i) (quoted-identifier `--`, R5). Shell exec: `\!` (R4). The other ~29 are census/linkage escapes, static evasions layer 2 did catch, and invisible-marker forms.

Two framings from the gate worth keeping in the record: **every write channel was found *after* a round I had already called done**, and **each was closed by a property, not by another name on a list**.

Merging once CI is green at `097a8bdc1d`; the ceiling and the marker class go to the owner as the ladder-L0 disposition, and this merge closes no ladder step.

PR #5034, opened 2026-08-20
security: P0-S hardening — BPMN runtime fail-closed gate + plugin-scope/ensureFields reconcile guards

skiyakubali0-cloud:
## What

P0-S security hardening from the MetaSheet platform-design review (rebaselined @ `c5a4a94f7`). Closes three exposures, all **fail-safe / default-non-breaking**.

### S1 — BPMN runtime fail-closed gate *(the real fix)*
`/api/workflow` (deploy/start/instances/tasks/message/signal/incidents) was mounted **unconditionally** and lacks per-task/tenant authorization — `completeUserTask` verifies no assignee/candidate, task `list` can enumerate cross-user, and the existing `DISABLE_WORKFLOW` flag only skipped `engine.initialize()`, **never stopping the HTTP routes**.

New `bpmnRuntimeConfig` (`ENABLE_BPMN_RUNTIME`, exact `'true'`, **default OFF**):
- 503 middleware gates the whole workflow router while disabled;
- engine initializes only when enabled (timer poller & `resumeActiveInstances` stay off too);
- gates the designer's two engine-reaching routes (`/templates/:id/instantiate`, `/workflows/:id/deploy`); draft/modeling/compile-preview authoring untouched.

`DISABLE_WORKFLOW=true` still forces off. BPMN is **dormant** (no business traffic) so default-OFF changes no live flow; the **Approval Center** engine is untouched.

### S3 — `ensureFields` destructive-reconcile guard *(groundwork)*
`ensureObject` (installer entry) drives `ensureFields`' `ON CONFLICT DO UPDATE`, silently overwriting tenant field renames/type/property/order on reinstall. New `MULTITABLE_ENSURE_FIELDS_OVERWRITE_MODE` (**default `overwrite` = today's exact SQL, no pre-read**) adds `observe` (warn+metric, still overwrites) and `preserve` (DO NOTHING on existing). Full three-way-diff / named-migration enforcement is a follow-up.

### S4 — plugin sheet-scope enforcement mode *(migration groundwork)*
Host `assertSheetScope` hook ignored `assertPluginOwnsSheet`'s boolean → a plugin could reach the plugin-scope records API for any **unregistered** sheet (test-pinned legacy tolerance). New `MULTITABLE_PLUGIN_SHEET_SCOPE_MODE` (**default `observe`**) now logs unregistered access; `enforce` throws after registry backfill. Cross-owner access already throws in every mode.

## Tests
- **42 new unit tests** (all three env-gates + classifier) ✅
- `tsc --noEmit` clean ✅
- existing `multitable-plugin-scope` (legacy-tolerance pin) + workflow engine suites: **77/77, no regression** ✅

## Follow-ups (per review)
- Registry backfill before flipping S4 to `enforce`.
- Upgrade-ledger + three-way diff for S3 `enforce`.
- Workflow task authorization + tenant model **deferred** until a task substrate is chosen (BPMN vs lightweight task object).

### DIFF
--- a/packages/core-backend/src/index.ts
+++ b/packages/core-backend/src/index.ts
@@ -591,7 +591,14 @@ export class MetaSheetServer {
           resolveFieldIds: async ({ projectId, objectId, fieldIds }) => {
             return resolveProvisionedObjectFieldIds(projectId, objectId, fieldIds)
           },
-          ensureObject: async ({ projectId, baseId, descriptor }) => {
+          // `overwriteMode` must be destructured and forwarded explicitly: this
+          // hook is the FALLBACK that plugin-scope's ensureObject takes when no
+          // ensureObjectInScope hook is registered (multitable/plugin-scope.ts —
+          // the scoped branch spreads ...input, and index.ts's scoped hook does
+          // forward it). Dropping it here silently downgraded a caller's explicit
+          // opt-in to the fail-closed default, i.e. an advertised API option
+          // (types/plugin.ts EnsureObjectInput) was inert on this path.
+          ensureObject: async ({ projectId, baseId, descriptor, overwriteMode }) => {
             return poolManager.get().transaction(async ({ query }) => {
               const txQuery: MultitableProvisioningQueryFn = async (sql, params) => {
                 const result = await query(sql, params)
@@ -606,6 +613,7 @@ export class MetaSheetServer {
                 projectId,
                 baseId,
                 descriptor,
+                overwriteMode,
               })
             })
           },
--- a/packages/core-backend/src/index.ts
+++ b/packages/core-backend/src/index.ts
@@ -1771,7 +1771,11 @@ export class MetaSheetServer {
       ? {
           ...pluginBaseCoreApi,
           multitable: createPluginScopedMultitableApi(coreApi.multitable, manifest.name, {
-            ensureObjectInScope: async ({ pluginName, projectId, baseId, descriptor }) => {
+            // P0-S S3: `overwriteMode` MUST stay in this destructure. It is the per-call
+            // destructive-reconcile opt-in, and this hook is the shipped host path for every
+            // plugin ensureObject — dropping it here would silently re-arm the fail-closed
+            // default for callers that legitimately own the columns they re-derive.
+            ensureObjectInScope: async ({ pluginName, projectId, baseId, descriptor, overwriteMode }) => {
               return poolManager.get().transaction(async ({ query }) => {
                 const txQuery: MultitableProvisioningQueryFn = async (sql, params) => {
                   const result = await query(sql, params)
@@ -1794,6 +1798,7 @@ export class MetaSheetServer {
                   projectId,
                   baseId,
                   descriptor,
+                  overwriteMode,
                 })
                 await claimPluginObjectScope(txQuery, {
                   pluginName,
--- a/packages/core-backend/src/index.ts
+++ b/packages/core-backend/src/index.ts
@@ -79,6 +79,7 @@ import {
   MultitableObjectScopeError,
   MultitableSheetScopeError,
 } from './multitable/plugin-scope'
+import { resolvePluginSheetScopeMode } from './multitable/pluginSheetScopeMode'
 import {
   acquireStockPreparationPersistUnitOfWorkLocks,
   validateStockPreparationPersistUnitOfWorkInput,
@@ -1857,10 +1858,24 @@ export class MetaSheetServer {
                     : undefined,
                 }
               }
-              await assertPluginOwnsSheet(txQuery, {
+              // P0-S S4 — sheet-scope enforcement mode. `assertPluginOwnsSheet` throws on a
+              // DIFFERENT-owner sheet in every mode; for an UNREGISTERED sheet it returns
+              // false (test-pinned legacy tolerance). Default 'observe' logs+continues (zero
+              // functional change, adds visibility); 'enforce' rejects unregistered access
+              // (flipped per-deployment after registry backfill).
+              const ownsSheet = await assertPluginOwnsSheet(txQuery, {
                 pluginName,
                 sheetId,
               })
+              if (!ownsSheet) {
+                const scopeMode = resolvePluginSheetScopeMode()
+                this.logger.warn(
+                  `plugin ${pluginName} accessed unregistered sheet ${sheetId} via plugin-scope (mode=${scopeMode})`,
+                )
+                if (scopeMode === 'enforce') {
+                  throw new MultitableSheetScopeError(pluginName, sheetId, 'unregistered')
+                }
+              }
             },
             runStockPreparationPersistUnitOfWork: async (
               { pluginName, ...rawInput },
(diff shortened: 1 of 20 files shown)


skiyakubali0-cloud:
Round-2 blockers addressed (79a4e065e + 1c371dd49). All three verified in code before fixing.

**A — Node 20 CI: real, self-inflicted by S1, not a flake.** `test (20.x)` is the only matrix leg with a postgres service, so it alone runs the DB integration step — 18.x "passed" by never running the file. Two tests in `bpmn-poller-disabled-startprocess-zero-residue.db.test.ts` (:310, :328) got `503 BPMN_RUNTIME_DISABLED` instead of 201: that suite drives real HTTP `/deploy`+`/start` to prove the *timer-poller* zero-residue claim, and the S1 gate 503s the whole surface when the flag is unset. Fixed by setting the flag in the suite before the `src/index` import (the engine reads it at server start) and restoring it in the existing `afterAll` — same treatment already applied to the egress suite.

**B — instantiate over-block: confirmed, gate removed.** `workflow-designer.ts:509-578` ends at `designer.saveWorkflow`, a pure `workflow_definitions` upsert with `status: existing?.status || 'draft'`. `WorkflowDesigner.ts` has **no import or use of `BPMNWorkflowEngine`** — the only mention is the comment at :937 ("the actual deployment would happen here"). No deployProcess, no startProcess, no timer. The sibling that legitimately keeps the gate is `/workflows/:id/deploy` (:1231 → `ensureWorkflowEngineReady` + `deployProcess` at :1249-1259). Also corrected `bpmnRuntimeConfig.ts`'s own doc comment, which still claimed "the designer's two engine-reaching routes".

**C — ensureFields default: confirmed, now fail-closed.** `resolveEnsureFieldsOverwriteMode` returned `'overwrite'` for unset env and provisioning skipped the pre-read in that mode — the shipped default really was the unguarded `ON CONFLICT DO UPDATE`. Default is now `'refuse'` (`MULTITABLE_ENSURE_FIELDS_REFUSED`, values-free: ids + diff kinds only). Only the exact literal `overwrite` opts out — `''`/`'OVERWRITE'`/`'true'` all still refuse. Fresh installs and additive evolution unaffected; `ensureMissingObjectFields` remains the repair path.

**Judgment call worth review:** the caller survey found two genuine breakers in `plugin-attendance` (report-records ~:2995, period-summary ~:4269) — their value columns are derived from the tenant-editable catalog with a *positional* order (1000+index), so any catalog edit renumbers the tail and every column classifies `would_overwrite` on a sync route. Rather than tell operators to set the env (which re-arms overwrite process-wide for every plugin), a per-call `overwriteMode` opt-in was threaded through `EnsureFieldsInput`/`EnsureObjectInput`, the plugin API type, and the `ensureObjectInScope` host hook — that hook destructures its input (`index.ts:1774`), so the option had to be added there explicitly or it would have been silently dropped on the shipped path. The other eight callers are safe as-is.

**Verification (main checkout, since the agent worktree had no deps and ran no vitest):** re-running there caught **two real regressions the agent could not see** — `multitable-template-library` and `multitable-template-dryrun-routes` each carry their own fake query layer and hit `Unhandled SQL` on the new per-field pre-read. Production code was correct; the fakes were incomplete, and 1c371dd49 adds the branch (33/33). Full core-backend unit sweep now fails **exactly the same 16 files as clean origin/main** (local Node 25 env: attendance/approval/runtime-dependency contracts) → **zero regressions**. Backend `tsc` clean; `global-history-flag-manifest` exit 0 (the new error code registered in NON_GH_EXACT — the contract's grep is a plain token scan, matching the existing `MULTITABLE_UNIT_OF_WORK_UNAVAILABLE` precedent).

**Scope, so it reads as scope rather than silence:** S4 `assertSheetScope` is observe-first *by the ratified plan* (inventory → backfill → enforce staged later, observe has a deadline), and S2 (`/erp/*` sealing + credential rotation) is a customer-side security-ops workstream tracked outside this code PR.

skiyakubali0-cloud:
CI red resolved, plus one latent hole closed (bc026e07d, which carries a30b181eb).

**The 500 was NOT what I said it was — correcting myself.** I told the owner this was a real-DB regression. It isn't: `tests/integration/multitable-context.api.test.ts` lives under `tests/integration/` but is a **supertest + mock-pool** suite — each test supplies its own `queryHandler` ending in `throw new Error('Unhandled SQL in test: …')`. It was reproduced with no postgres at all, and the captured server stack is unambiguous:

```
Unhandled SQL in test: SELECT name, type, property, "order" FROM meta_fields WHERE id =  AND sheet_id = 
  at queryHandler (tests/integration/multitable-context.api.test.ts:870:15)
  at Module.ensureFields (src/multitable/provisioning.ts:350:36)
  at Module.installMultitableTemplate (src/multitable/template-library.ts:622:35)
```

Third fake missing the S3 pre-read branch — the same gap 1c371dd49 closed for two unit suites, missed because this one sits under `tests/integration/`. **No production file changed and no `overwriteMode` escape hatch was used: the guard was never refusing here.** Two tests were failing, not one (the H3 contract-management install failed identically).

Hypotheses explicitly **disproven**, so they don't linger as folklore:
- *pool-vs-transaction* — false. `provisioning.ts:350` issues the pre-read on `input.query`, the same fn as the upsert (:393) and read-back (:409); already tx-scoped.
- *same-tx re-ensure refusing* — false. The error is `Unhandled SQL`, not `MULTITABLE_ENSURE_FIELDS_REFUSED`; template install is structurally first-install.
- *SQL/param shape mismatch* — false. Only the hand-written fake lacked the branch.

**Caller survey, 14 sites**: template install and both stock-prep provisioners are first-install → safe. Three attendance `ensure*` sites use static descriptors → unchanged → safe. The two already-exempted attendance sites pass `overwriteMode:'overwrite'` for positional-`order` churn. `plugin-after-sales/lib/installer.cjs:468` is a genuine re-ensure that now fails closed into its ledger catch — correct and loud.

**Latent hole found and fixed in the same pass**: both host provisioning hooks in `src/index.ts` destructure their input, and only the scoped one (`ensureObjectInScope`, :1778) listed `overwriteMode`. `plugin-scope.ts`'s `ensureObject` spreads `...input` into the scoped hook when registered (so the shipped path worked), but **falls back to `multitable.provisioning.ensureObject` when no such hook exists — and :594 dropped the option**, downgrading an explicit opt-in back to the fail-closed default. Safe direction (a refusal, never an unwanted overwrite), but an advertised API option was inert on that path.

**A note on the test for it, because the first attempt was bad.** My initial scope-layer test mocked `multitable.provisioning`, so reverting the host fix left it **green** — vacuous. It is kept (it does pin plugin-scope's two forwarding paths) and joined by a source-contract test asserting both host hooks destructure *and* forward the option. That wiring is inline in the server bootstrap and can't be imported without standing up the app, so a structural assertion is the honest instrument here. Mutation-verified: reverting the destructure now turns the suite **red** — it did not before.

**Verification**: plugin-scope 14/14 · provisioning 27/27 · the previously failing `multitable-context.api` file **22/22** (was 2 failed/20) · flag-manifest 24/24 exit 0 · backend `tsc` clean · full `tests/unit` sweep fails **exactly the same 16 files as clean origin/main** → zero regressions.

Two further defects found and deliberately **not** fixed here (out of scope, filed separately): `staging-installer.cjs:263` folds a refusal into `warnings` and still returns success-shaped output, so the fail-closed signal never reaches its route; and the real-postgres suites could not be run locally (no container runtime on this machine) — CI's 20.x leg is the adjudicator, and it had reported only this one file failing.

PR #5035, opened 2026-08-20
docs(spikes): P0-D design spikes — principal / mirror / registry

skiyakubali0-cloud:
## What
Design spikes for the three **P0-D model blockers** from the platform-design review (v9.1). **Docs/artifacts only — no application code is wired.** Produced by model-tiered subagents (Opus: principal + mirror; Sonnet: registry), grounded in the real schema @ `c5a4a94f7`. Migration drafts live under `docs/`, not the real migrations dir.

13 files under `docs/development/spikes/p0d-20260820/`: 3 × (ADR + `*.draft.sql` + prototype `.ts` + `.test.ts`) + a cross-spike synthesis.

## Decisions (per spike)
- **Principal** — unified `service_principals` (reverse-FK, `kind` discriminator); tenant safety via **DB composite FK** `(principal_id, tenant_id)` (cross-tenant grants rejected by PG); soft-revoke, no delete, no id reuse; immutable `principal_id` via trigger.
- **Mirror (hardest)** — **Option A**: internal staging → keyed upsert-diff into the live mirror sheet. User table is always the current generation → **zero read-path change**, atomic visibility from a single MVCC txn, record ids preserved by keyed diff. Rejects C (generation filter in every read path); B kept as documented fallback. All 10 Charter acceptance criteria mapped.
- **Registry** — 3-table design, active-uniqueness scoped to `registry_generation_id`; composite `(registry_generation_id, hash, canonical_key)` (hash locates, canonical compares → collisions safe); `ON CONFLICT` is the authoritative race guard. Renumber v1 manual-only.

## Synthesis — the key finding
The three spikes were **designed in isolation and share zero references**. The identity model (spike 1) is **not wired into** the mirror publisher or registry writer (bare `proposed_by`/`actorId` strings); two `generation` concepts and two "same-row" notions coexist unreconciled. The synthesis surfaces the P0-D **integration contract** that must be ratified before P1:
- publishers & pipeline writers are `service_principals` holding `writer_grants`;
- an **externally-fed** mirror's `keyOf` resolves through the registry's active generation (not a raw field join) — else duplicate mirror rows for one canonical key;
- a publish plan binds `(mirror_generation_id, registry_generation_id, content_key, publishing_principal_id)`.

Output: **13 ordered P0-D decisions** + **13 real-DB integration tests** that must land before P1.

## Ratify readiness
- **Ratify-now** (schema decisions on established PG mechanisms): spikes 1 & 3.
- **Gated on a real-DB run**: spike 2 (Option A-vs-B hinges on an unmeasured scale benchmark) + spike 3 open Q2 (rebuild-against-moving-target).

## Verification
- **56 spike unit tests green locally** (no-DB pure logic: normalize/hash/classify/diff/state-machine/principal-writability). One fixture bug fixed in `spike3-registry.test.ts`.
- No app code touched; `git status` shows only `docs/`.

## Not in this PR
Wiring into the app, real migrations, and the 13 real-DB integration tests — those are P1 after owner/tech-lead ratifies the decision list. This PR is the reviewable P0-D handoff.

### DIFF
--- a/docs/development/spikes/p0d-20260820/spike-synthesis-and-ratify-recommendation.md
+++ b/docs/development/spikes/p0d-20260820/spike-synthesis-and-ratify-recommendation.md
@@ -1,0 +1,146 @@
+# P0-D Spike Synthesis & Ratify Recommendation
+
+- **Status**: SYNTHESIS (design spike roll-up — no decision ratified, no code wired)
+- **Date**: 2026-08-20
+- **Baseline**: `main @ c5a4a94f7` (frozen)
+- **Inputs**: `spike1-principal-*`, `spike2-mirror-*`, `spike3-registry-*` (this dir)
+- **Purpose**: (1) cross-spike consistency check, (2) per-spike Ratify readiness + missing evidence, (3) the single ordered P0-D decision list for owner/tech-lead, (4) the real-DB integration tests that MUST land before P1.
+
+---
+
+## 0. TL;DR
+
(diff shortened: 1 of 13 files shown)


skiyakubali0-cloud:
关闭:内容早于 2026-09 备料线的多轮改动(#5489–#5521),合入会产生过时文档;如需其中某段,按当前 main 重写。owner 2026-09-06 拍板。

PR #5039, opened 2026-08-20
docs(staging): disposition note for the four pending approval migrations

zensgit:
The O-2 ladder's L0 requires staging `pending migrations = 0`; the 2026-08-20 deploy stopped on the alignment report (`do_not_run_full_migrate`, run `32321291505`) with exactly **four approval-line migrations** pending — the only non-owner-button blocker to L1.

This note is the Time Machine line's **read-only static analysis + pre-filled rehearsal checklist**, so the approval line can act per the runbook without re-deriving the risk. **Application/marking belongs to the approval line; owner approves; this line does not apply them** — stated in the note.

## Verdicts (line-verified, not grep-verdicts)
- The report's two **high** flags are **heuristic false positives**: `DROP CONSTRAINT IF EXISTS` + same-name re-ADD = idempotent CHECK enum widening (14→15→16), no data migration, no column change.
- One layer deeper, honestly: the `ADD CONSTRAINT` is **validating** (no `NOT VALID`) → full-scan under `ACCESS EXCLUSIVE`, but it **cannot fail on existing rows** (strict superset) and there is **no bare window** (single-transaction migrator holds the lock across the swap).
- The other two are idempotent new tables (`IF NOT EXISTS` / conditional PK).
- **O-2 blast radius**: zero platform-auth ALTERs; the one new FK into `users` (`created_by`) is already drilled by canary 3.5a (#5032).

## Checklist (§4 of the note)
Clone/backup rehearsal → `schema-probes.sql` (expected pre-state stated, including the half-applied-detection branch) → full migrate on the clone → apply to staging per runbook → re-run window-runner deploy → report should turn `pending=0` → the ladder L0 sub-item closes.

Docs-only. No flags, no triggers, no host.

### DIFF
--- a/docs/development/staging-approval-migrations-disposition-20260820.md
+++ b/docs/development/staging-approval-migrations-disposition-20260820.md
@@ -1,0 +1,57 @@
+# staging 四条审批迁移处置便笺(2026-08-20)
+
+> **给谁**:审批线(执行)+ owner(批准)。
+> **谁写的、边界是什么**:Time Machine O-2 线。本便笺是**只读静态分析 + 预填彩排清单**;
+> 迁移的应用/标记动作归审批线按 runbook 执行,owner 批准。**本线不代为应用。**
+> **为什么存在**:O-2 阶梯 L0 要求 staging `pending migrations = 0`。2026-08-20 部署 `401fa1d880`
+> 时对齐报告判 `do_not_run_full_migrate`(run `32321291505`,工件含 `report.md` / `schema-probes.sql`),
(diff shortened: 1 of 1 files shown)


zensgit:
**Assistant-posted approval-line handoff notice.** The original four-migration snapshot is superseded by the target-SHA-scoped handoff at #5094 comment 5387254068: https://github.com/zensgit/metasheet2/pull/5094#issuecomment-5387254068

Owner has authorized the approval line to align staging to `Pending: 0` **before** attendance deploys exact SHA `3b7d3c0c3ee2b2463180c548a284f49095867764`. Please use the target migration universe; the running old image omits migration `076_create_integration_stock_prep_pack_installs.sql`.

PR #5058, opened 2026-08-21
feat(approval): hover/focus tooltip on flow-canvas node-card summary (FS-1)

zensgit:
## Summary

FS-1 — the W4 approval-canvas closeout's single FAIL (`docs/development/approval-remaining-dev-verification-report-20260820.md` §5.1): the flow-canvas node card's CSS-ellipsis-truncated summary line (`ApprovalFlowCanvas.vue`) had no way to recover the full text on hover or keyboard focus, violating the "hover/focus tooltip" leg of the RATIFIED Long-labels criterion (`approval-canvas-v2-interaction-design-lock-20260721.md:366`: *"Truncate with ellipsis at component limits (§14); full text on hover/focus tooltip, in the inspector, and in accessible names."*).

- Wraps the card's existing focusable selector (`role="button" tabindex="0"`, unchanged) in `<el-tooltip :trigger="['hover', 'focus']">`. Element Plus's own default `trigger` is `'hover'` **only** (read from `element-plus/es/components/tooltip/src/trigger2.mjs`), so a bare `<el-tooltip>` would have silently stayed hover-only and missed the a11y half — which is the point of this fix.
- Tooltip content reuses the exact same values-free `canvasNodeSummary(pos.key)` call the card already renders inline — no re-derivation, no raw ids.
- **Deliberately out of scope** (per §5.1, so as not to repeat "misclassified gap called a ceiling"): the *"in the inspector"* leg (a separate, unverified claim about `ApprovalCanvasNodeInspector.vue`) and the *"in accessible names"* leg (**FS-7** — the parent `role="button"` div's own `aria-label` overrides the whole subtree's accessible name per standard accname computation, a **different** defect the source doc explicitly requires stay a separate slice, never merged with FS-1). This fix does not touch `aria-label` and does not close FS-7.

## New test

`apps/web/tests/approval-flow-canvas-summary-tooltip.spec.ts` — the first spec to mount `ApprovalFlowCanvas.vue` directly (minimal typed prop fixture), with the **real** Element Plus `ElTooltip` registered (not stubbed), proving against real runtime behavior (not source-text regex):

- the tooltip's popper content carries the exact full, untruncated, values-free summary string;
- the trigger is really focusable, and a real `.focus()` call sets `aria-describedby` on it to the tooltip content's `id` (genuine DOM mutation inside Element Plus's own trigger logic — `content2.mjs` renders the popper content unconditionally under `NODE_ENV==='test'`, so it's queryable without ever opening the tooltip, but `aria-describedby`/`aria-hidden` still reflect the real open/closed state);
- hover (`mouseenter`/`mouseleave`) independently opens/closes the same tooltip — the fix is hover **and** focus, not a focus-only regression of the previously-working hover case;
- a structural pin on the `:content="canvasNodeSummary(pos.key)"` binding.

**Mutation-proven** (`cp` backup + sha256 byte-identical restore, never `git checkout --`): reverting `:trigger="['hover', 'focus']"` to Element Plus's bare `'hover'` default reds **only** the keyboard-focus assertions — the hover assertions and full-text assertion stay green. That asymmetry is the discriminating proof.

**Honest boundary**: jsdom has no layout engine, so this file cannot prove real CSS-ellipsis clipping, a genuine `:hover` pseudo-class, or the popper's real screen position. No existing browser-verify harness mounts `ApprovalFlowCanvas`'s node cards (`apps/web/verification/approval-inspector-keyboard.spec.ts` — this repo's one approval-scoped real-Chromium harness — mounts only `ApprovalCanvasNodeInspector.vue` + `ApprovalGraphNodeConfigEditor.vue`), so building a brand-new harness is explicitly scoped **out** of this bounded fix slice rather than force-fit into that unrelated spec.

## Wiring (two-point discipline)

- `apps/web/scripts/run-required-web-tests.sh` — added the exact-basename token `approval-flow-canvas-summary-tooltip` (required by the FAIL-0 mechanical coverage guard, `packages/core-backend/tests/unit/approval-ci-coverage-enumeration.test.ts`; verified RED without the token, GREEN with it).
- `.github/workflows/approval-web-guard.yml` — added the spec's path to **both** `paths:` trigger blocks and the token to the "Run Canvas V2 command canaries" `vitest run` list.

## Test plan

- [x] `pnpm run type-check` (apps/web) — exit 0 (`vue-tsc -b` + verification-approval project)
- [x] New spec green in isolation (4/4) and mutation-proven (focus-only assertions red on revert, hover assertions stay green, byte-identical `cp`+sha256 restore)
- [x] `bash apps/web/scripts/run-required-web-tests.sh` — 388 files / 4947 tests, exit 0 (includes the new spec)
- [x] FAIL-0 coverage guard (`approval-ci-coverage-enumeration.test.ts`) — 271/271, new file confirmed "is wired"; verified it actually reds when the wiring token is removed
- [x] Raw-id census guard (`approval-member-identity-coverage-enumeration.spec.ts`) — 12/12 green, unaffected
- [x] All existing structural a11y/canvas specs green post-change (13 files / 424 tests), including specs that mount `TemplateAuthoringView` without registering `ElTooltip` (confirmed the resulting unresolved `<el-tooltip>` wrapper element is harmless — no existing spec uses a direct-child selector across that boundary, only `[data-testid]` descendant selectors)
- [x] `node --test scripts/ops/attendance-w4c0-dml-inventory-collector.test.mjs` — 58/58 pass
- [x] `git diff origin/main -- .github/workflows/plugin-tests.yml pnpm-lock.yaml` — empty
- [x] `git diff --check` — clean; no control bytes, no bare `#N`

https://claude.ai/code/session_011jK9Rh3qBs2T2stcgrCXpn

### DIFF
--- a/.github/workflows/approval-web-guard.yml
+++ b/.github/workflows/approval-web-guard.yml
@@ -134,6 +134,7 @@ on:
       - 'apps/web/tests/approval-form-authoring-history.test.ts'
       - 'apps/web/tests/approval-version-dual-canvas.test.ts'
       - 'apps/web/tests/approval-flow-canvas-a11y.test.ts'
+      - 'apps/web/tests/approval-flow-canvas-summary-tooltip.spec.ts'
       - 'apps/web/tests/approval-canvas-inspector-a11y.test.ts'
       - 'apps/web/tests/approval-form-palette-focus.test.ts'
       - 'apps/web/src/approvals/approvalFormAuthoringHistory.ts'
@@ -418,6 +419,7 @@ on:
       - 'apps/web/tests/approval-form-authoring-history.test.ts'
       - 'apps/web/tests/approval-version-dual-canvas.test.ts'
       - 'apps/web/tests/approval-flow-canvas-a11y.test.ts'
+      - 'apps/web/tests/approval-flow-canvas-summary-tooltip.spec.ts'
       - 'apps/web/tests/approval-canvas-inspector-a11y.test.ts'
       - 'apps/web/tests/approval-form-palette-focus.test.ts'
       - 'apps/web/src/approvals/approvalFormAuthoringHistory.ts'
@@ -647,6 +649,7 @@ jobs:
             approval-form-authoring-history \
             approval-version-dual-canvas \
             approval-flow-canvas-a11y \
+            approval-flow-canvas-summary-tooltip \
             approval-canvas-inspector-a11y \
             approval-form-palette-focus \
             approval-form-inline-editor-extract \
--- a/apps/web/scripts/run-required-web-tests.sh
+++ b/apps/web/scripts/run-required-web-tests.sh
@@ -348,6 +348,7 @@ npx vitest run \
   approval-form-authoring-history \
   approval-version-dual-canvas \
   approval-flow-canvas-a11y \
+  approval-flow-canvas-summary-tooltip \
   approval-canvas-inspector-a11y \
   approval-form-palette-focus \
   approval-form-inline-editor-extract \
--- a/apps/web/src/approvals/components/ApprovalFlowCanvas.vue
+++ b/apps/web/src/approvals/components/ApprovalFlowCanvas.vue
@@ -238,23 +238,60 @@ function nodePosStyle(pos: NodeLayout): CSSProperties {
               >
                 {{ nodeTypeLabel(canvasNodeByKey(pos.key)?.type ?? 'approval') }}
               </div>
-              <div
-                class="template-authoring__canvas-node-selector"
-                role="button"
-                tabindex="0"
-                :aria-label="`编辑${graphNodeLabel(pos.key)}节点`"
-                :aria-pressed="selectedCanvasNode === pos.key"
-                data-testid="approval-canvas-node-select"
-                @click.stop="emit('select-node', pos.key)"
-                @keydown.enter.stop.prevent="emit('select-node', pos.key)"
-                @keydown.space.stop.prevent="emit('select-node', pos.key)"
-                @keydown="emit('node-keydown', $event, pos.key)"
+              <!-- FS-1 fix — the W4 approval-canvas closeout's single FAIL
+                   (docs/development/approval-remaining-dev-verification-report-20260820.md §5.1;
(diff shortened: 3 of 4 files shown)


zensgit:
Two clarifications on scope/evidence, added after a self-review pass:

**1. Mechanism vs. the repro anchor in §5.1.** The verification report's failure scenario reads "hover the card and cannot obtain a native `title` tooltip," and its repro anchor calls out the truncated `<span>` itself for lacking `title`/`aria-label`. This fix instead attaches `<el-tooltip>` to the parent focusable `div` (the existing `role="button" tabindex="0"` selector) and delivers the full text via `aria-describedby`, not a native `title` attribute. That's a deliberate mechanism choice, not an oversight: a native `title` attribute does not surface on keyboard focus at all (only on mouse hover, after a browser-controlled delay), so `title` alone cannot satisfy the criterion's "focus" half. The RATIFIED text at `approval-canvas-v2-interaction-design-lock-20260721.md:366` itself is mechanism-agnostic ("full text on hover/focus tooltip") — it does not name `title` specifically. So this fix satisfies the ratified criterion while using a different element/mechanism than the repro's incidental anchor. Flagging this explicitly so a reviewer who greps for `title=` at the old `:253-256` line numbers doesn't read the absence as a miss.

**2. Node/pnpm version delta on the local full-suite run.** `bash apps/web/scripts/run-required-web-tests.sh` passed 388 files / 4947 tests (exit 0) locally on Node v25.9.0 / pnpm 10.33.0 — CI's `test (20.x)` matrix leg (and this repo's own `run-required-web-tests.sh` header, and §5.1's source doc, R-1) flags exactly this ambient-vs-CI Node version gap as a real evidence-retention concern for other specs in this suite that used fixed-tick waits. This new spec doesn't have that exposure: it advances Element Plus's own `setTimeout`-based open/close timers deterministically via `vi.runAllTimersAsync()` rather than a fixed real-clock sleep or a guessed tick count, so there's no version-sensitive fixed-tick assumption for CI's 20.x runner to diverge on. Noting the gap rather than claiming it's closed, per this repo's own verification discipline.

Neither point changes any code — both are already reflected in the PR body's design reasoning; this comment makes them explicit for review.

PR #5061, opened 2026-08-21
perf(multitable): cut the sheet-open request waterfall

skiyakubali0-cloud:
## What

Opening one multitable sheet on a live deployment fired ~10 API round trips; `/api/multitable/bases` was fetched 5+ times per session (once serially BEFORE the sheet's own context load), and the attendance monolith (~29k lines) was statically bundled into the entry chunk that multitable users must parse.

Observed request trace (live instance, one sheet open): `bases` ×5, `templates` ×2, `context` ×2, plus mention-candidates / mention-summary / two unread-counts / subscription badge — before the grid painted.

## Changes

- **Catalog cache** (`multitable/api/client.ts`): `listBases`/`listTemplates` cached per client instance with in-flight dedup (same pattern as `usePlugins.fetchPlugins`); `createBase`/`installTemplate` invalidate; `{ force: true }` opt-out. Returned arrays are copies so caller mutation can't pollute the cache.
- **Critical path** (`MultitableWorkbench.vue` onMounted): stop awaiting `loadBases()` ahead of context. Anchored opens (sheetId or active base) load context immediately and the bases rail fills in parallel — `loadBases`'s own `bases[0]` selection guard already skips anchored cases. Only the nothing-anchored branch still waits.
- **Lazy mention candidates**: drop the eager per-sheet fetch; the sheet-change watcher already resets the loaded marker, so the existing on-demand load in `loadCommentsForRecord` fetches fresh when a comment composer actually opens.
- **Idle-deferred badges**: comment unread count + notification bell ambient count via new `utils/scheduleIdle` (rIC, setTimeout fallback) — off the sheet-open critical path.
- **Code split**: `AttendanceExperienceView` lazy in `router/appRoutes.ts` (4 route entries) and `defineAsyncComponent` in `plugins/viewRegistry.ts` (a static import there would defeat the route split).

## Not in this PR (follow-ups)

- `comments/summary` carries every visible row id in the URL query — separate FE+BE PR converts it to POST body (URL-length scaling bug).
- Home→workbench double `context` fetch (needs a routing-contract change).
- `WorkflowDesigner.vue` static import in viewRegistry looks like the same entry-chunk pattern (bpmn-js) — unmeasured, left alone.

## Verify

- `vue-tsc --noEmit` green
- New `tests/multitable-catalog-cache.spec.ts`: cache hit / concurrent dedup / mutation invalidation / copy-guard / force / failed-fetch-not-poisoning (7 tests)
- Bell specs updated: `flush` drains one macrotask for the idle-deferred ambient fetch (contract: badge still loads, one tick later)
- 124 affected-surface web tests green (workbench view/manager/import/restore/drawer/conflict, home, template center+detail, crossbase, phase3, notification bell+inbox, StockPreparationWorkspace, adminAccess, platform-shell-nav, attendance-records-redirect)


skiyakubali0-cloud:
Review fixes pushed (7c4a25f18):
- **Generation control**: each catalog carries a generation counter; `invalidate`/`force` bumps it and drops the joinable in-flight handle, so an OLDER request resolving LATER can no longer write back stale data (the awaiting caller still receives its payload — it just cannot populate the cache). Two new tests pin the reverse-order cases.
- **Home 刷新 now forces**: the button routes through `refreshHomeData` → `loadHomeData({force:true})`; mount stays cached. Both findings confirmed as real before fixing.
75 affected-surface tests green; `vue-tsc -p tsconfig.app.json` green (note: `-p tsconfig.json` is a solution-style no-op and is not valid evidence).

PR #5062, opened 2026-08-21
perf(comments): comment-summary row ids move to POST body

skiyakubali0-cloud:
## What

`GET /api/comments/summary` carried every visible row id comma-joined in the query string (`?rowIds=rec_a,rec_b,...`). The grid requests presence for the whole visible page, so the URL grows with page size toward 414/URL-length limits — observed live with 22 rows already producing a very long URL.

## Changes

- **Backend** ([comments.ts](packages/core-backend/src/routes/comments.ts)): summary handler extracted and shared; **GET kept** (rolling-deploy compat) and **POST /api/comments/summary** added, reading `{spreadsheetId, rowIds}` from the JSON body behind the same `apiTokenAuth + requireScope('comments:read') + rbacGuard` stack.
- **Auth boundary untouched**: the OAPI-1 `mst_`-token read allowlist stays GET-only by design — POST is deliberately NOT allowlisted for `mst_` bearers (that would be an auth-boundary change needing its own design review). A token POST falls through to the JWT gate and 401s: fail-closed, unchanged.
- **Web client**: `listCommentPresence` posts ids in the body.

## Deploy order

core-backend first, then web (new client requires the POST route; old client keeps using GET).

## Verify

- `comment-routes-row-deny.test.ts`: POST variant passes the same filtered rowIds + denied-row exclusions (4/4)
- `multitable-comment-presence.spec.ts`: pins the POST body shape (39/39 incl. client spec)
- `vue-tsc` + backend `tsc` both clean

### DIFF
--- a/apps/web/src/multitable/api/client.ts
+++ b/apps/web/src/multitable/api/client.ts
@@ -2685,10 +2685,18 @@ export class MultitableApiClient {
 
   async listCommentPresence(params: { containerId: string; targetIds?: string[] }): Promise<{ items: MultitableCommentPresenceSummary[] }> {
     const targetIds = (params.targetIds ?? []).filter((targetId) => typeof targetId === 'string' && targetId.trim().length > 0)
-    const res = await this.fetch(`/api/comments/summary${qs({
-      spreadsheetId: params.containerId,
-      rowIds: targetIds.length ? targetIds.join(',') : undefined,
-    })}`)
+    // POST with the ids in the body: the grid sends every visible row id per
+    // page, and the previous comma-joined query string grew with page size
+    // toward URL-length limits (414). The backend keeps the GET variant for
+    // rolling-deploy compat; deploy core-backend before this client.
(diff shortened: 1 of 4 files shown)


skiyakubali0-cloud:
Bound added (5c1386152 + test in 2808f22ba): shared schema caps the id set at **5000 ids / 128 chars per id** — over-limit returns 400 before the service is reached, on GET and POST alike.

Note from writing the test: a 5001-id **GET** never reaches the route at all — Node resets the socket on the oversized query string (ECONNRESET). So the GET arm is probed with the per-id length bound instead, and the count bound is exercised over POST. That failure is itself the clearest statement of why the id set moved into a body.

Tests: 5001 ids → 400 + service uncalled; 129-char id → 400 (both verbs); exactly 5000 → 200, service called once. 5/5 green, backend tsc green.

Disclosure: the route bound landed one commit before its test (a patch script aborted mid-chain after writing the route) — the follow-up commit adds it.

PR #5063, opened 2026-08-21
ops(attendance-soak): rotate the soak seed's synthetic-user password without a full reseed

zensgit:
## Summary

Adds `soak_opts rotate_password=true` to `action=soak-seed` in the attendance combined-soak (#4556) staging window runner: rotates the host-only synthetic-user password (and its DB hash) used by the soak seed without re-running the full seed/posture walk.

- New `soak_opt_present`, `soak_mint_password`, `soak_hash_password_in_backend` helpers in `scripts/ops/attendance-staging-window-runner-remote.sh` (the last two are the single shared implementation for both the first-mint path and rotation — proven by a pinned "exactly one occurrence" assertion).
- New `soak_seed_rotate_password()`: fails closed if the credentials file is absent ("nothing to rotate", never silently mints); mints a fresh password with the shared generator; atomically replaces the credentials file (mktemp + `chmod 0600` + `mv -f`, same dir), preserving the prior file as `${SOAK_CREDENTIALS_FILE}.prev` (umask-077 idiom, 0600 from birth) so a DB-write failure after the swap is recoverable by restoring `.prev`; re-hashes ONLY `username LIKE '${SOAK_USER_PREFIX}%'` via a transactional, prefix-scoped `UPDATE` (no inserts, no posture walk, no group/shift/schedule seeding, no remint of retired families); records `rotated=1` + the row count in `summary.txt`; never echoes/logs/uploads the plaintext password.
- `action_soak_seed` now reads `rotate_password` first and, when `true`, refuses `users_per_org`/`tz`/`w7_target` if also supplied (standalone act), then dispatches to rotation and returns — BEFORE `soak_require_orgs`/`owner_ref`/`entrypoint_inventory_ref` are ever read. `owner_ref`/`entrypoint_inventory_ref` may still be supplied on the dispatch but are unused (documented asymmetry: unlike users_per_org/tz/w7_target, they carry no behavioural effect to silently miss).
- `soak_orgs` stays a REQUIRED workflow input for `action=soak-seed` regardless — that block was intentionally left untouched; rotation just never reads the value. Verified empirically both ways.
- Workflow: extended the `soak_opts` description and the per-key validation switch with `rotate_password) [[ "$value" == "true" ]] || exit 2`, plus the "known keys" list. `.github/workflows/plugin-tests.yml` was not touched, so no s6a provenance-pin repin was needed.

## Post-gate hardening round (soak-working/GATE-5063.md — REQUEST_CHANGES, 0P1/2P2/3P3/2NIT — all addressed on this branch)

- **F1 (P2, secret hygiene):** the original rotate psql call carried `-e` (echo-queries), which prints the interpolated bcrypt hash into `soak-seed-rotate.txt` — a file the workflow `scp`'s wholesale into `OUTPUT_DIR` and uploads as a world-downloadable CI artifact (`retention-days: 30`, `if: always()`). Fixed by dropping `-e` — **not** by switching to `-q`: verified empirically that `-q` silences the output the row-count parse depends on, so that obvious-looking "fix" would have broken rotation silently while the suite stayed green (exactly the gate's own M6 finding). The row count is now read from an explicit `RAISE NOTICE 'ROTATE_RESULT ...'` line instead of psql's command-tag text.
- **F2 (P2, untested guard):** a rotation matching zero family rows used to report `result=ok` with the credentials file already holding a password matching nothing. Now requires `(( rotated_users > 0 ))`; on 0 rows, auto-restores `${SOAK_CREDENTIALS_FILE}.prev` onto the credentials file (atomic `mv -f`) and fails with a message naming what was restored.
- **F3 (P3, blast radius):** inside the SAME transaction, before `COMMIT`: (1) a 999-row sanity ceiling — documented explicitly as defense-in-depth, **not** a derived family-size bound (the family accumulates across dispatches with no hard cap; a tight ceiling would false-alarm a legitimate multi-triple family); (2) the UPDATE's actual row count (via `GET DIAGNOSTICS`) must equal a pre-count on the same predicate (concurrent-write guard); (3) the UPDATE must not have left NOTHING un-matched (the mis-composed `"%"` case, on top of the existing static C9 pin). Known, documented trade-off: a staging `users` table containing ONLY synthetic-family rows would also trip check (3) — acceptable per the reviewed design (a cheap invariant, not a completeness guarantee).
- **F4 (P3, recovery wording):** the DB-step failure message no longer claims "nothing committed" as fact (a transport/stream failure can occur *after* a real `COMMIT`) — it now tells the operator to check `soak-seed-rotate.txt` for `ROTATE_RESULT ...` followed by `COMMIT` before ever restoring `.prev`.
- **F5 (P3, single-generation `.prev`):** documented in the function header rather than adding a `.prev2` generation — F2's auto-restore removes the common repeat-failure trigger (matching zero rows), so the two-consecutive-failure scenario `.prev`'s single generation can't cover is now rarer in practice.
- **NIT-1:** `.prev` is created with the same `umask 077` idiom the first-mint path uses (0600 from birth), not `cp -p` + a separate `chmod`.
- **NIT-2:** documented the `owner_ref`/`entrypoint_inventory_ref` asymmetry in the function header (why they're allowed to ride along unused while `users_per_org`/`tz`/`w7_target` are refused).

**Also found and fixed during empirical verification (not part of the original gate report):** psql's `:'var'` client-side variable substitution does **not** reach inside a `DO $$ ... $$` dollar-quoted body — confirmed against a real `postgres:16` container; a naive `:'user_prefix'` reference inside the DO block is a silent syntax error, not a leak, but would have shipped rotation completely broken (the F1/F2/F3 fixes all depend on that DO block running). Fixed by routing both values through a transaction-local GUC (`set_config('rotate.user_prefix', :'user_prefix', true)`, read back via `current_setting()`), with the `SELECT set_config(...)` call's own result printout suppressed via `\gset` into a throwaway column — a bare `SELECT` would otherwise have re-introduced the exact class of leak F1 fixes, just via a different mechanism. All four scenarios (normal rotation, zero-row, blast-radius `%`, ceiling >999) were verified end-to-end against a disposable `postgres:16` container using the exact heredoc committed in this branch.

## Test plan

- [x] `node --test scripts/ops/attendance-window-runner-pipeline.test.mjs` — **105/105 pass** (was 72 baseline; +27 for the base slice, +6 more for the gate-fix round)
- [x] `bash -n scripts/ops/attendance-staging-window-runner-remote.sh` — OK
- [x] `node plugins/plugin-integration-core/__tests__/sealed-export-package-provenance.test.cjs` — OK (no `plugin-tests.yml` changes, so no repin needed)
- [x] New static pins + MUTATION legs for every gate finding: `-e` reappearing on the rotate psql call, `-q` replacing the flag entirely (the M6 gap), the `>0` guard being removed (the M4 gap), the ceiling check, the blast-radius check.
- [x] New EXECUTABLE leg drives the REAL `soak_seed_rotate_password` end-to-end with a fake `docker` binary on `PATH` (no real container) whose canned output mirrors the real SQL's 0-row NOTICE — proves behaviourally (not just textually) that a 0-row rotation restores `.prev` and never writes `result=ok`.
- [x] Re-ran the gate's own M4 and M6 mutations as real on-disk edits (not just in-memory) against the fixed code: both now turn the suite red (M4: 105→80 pass/25 fail; M6: 105→79 pass/26 fail), confirmed the correct assertions fire, then restored via precise text-match edits (never `git checkout -- .` — this was mid-development, not yet committed at mutation time) back to 105/105 green.

## Operator dispatch

```
gh workflow run attendance-staging-window-runner.yml \
  -f action=soak-seed \
  -f soak_orgs='<org1-uuid>,<org2-uuid>,<org3-uuid>' \
  -f soak_opts='rotate_password=true'
```

`soak_orgs` must still be supplied (workflow-level requirement for `action=soak-seed`, unchanged) but its value is never read by the rotation path.

Not merging this PR myself; staged opt-in / merge timing is an owner call per repo convention.

### DIFF
--- a/.github/workflows/attendance-staging-window-runner.yml
+++ b/.github/workflows/attendance-staging-window-runner.yml
@@ -71,7 +71,7 @@ on:
         required: false
         default: ''
       soak_opts:
-        description: 'soak actions: ";"-separated key=value pairs. Keys: users_per_org (soak-seed, default 10), tz (soak-seed, IANA name or three "|"-separated, default Asia/Shanghai), w7_target (soak-seed, only group_shadow is runnable at kickoff), owner_ref (soak-seed REQUIRED — names the owner-authored authorization artifact for the transition manifests), entrypoint_inventory_ref (soak-seed REQUIRED — operator attestation for the W4 manifest entrypointInventoryRef; never fabricated), punch_target (soak-run, default 200), config (soak-run host config path), window_start (soak-status override, ISO8601 UTC).'
+        description: 'soak actions: ";"-separated key=value pairs. Keys: users_per_org (soak-seed, default 10), tz (soak-seed, IANA name or three "|"-separated, default Asia/Shanghai), w7_target (soak-seed, only group_shadow is runnable at kickoff), owner_ref (soak-seed REQUIRED — names the owner-authored authorization artifact for the transition manifests), entrypoint_inventory_ref (soak-seed REQUIRED — operator attestation for the W4 manifest entrypointInventoryRef; never fabricated), rotate_password (soak-seed, only "true" accepted — a standalone act that rotates ONLY the host-only synthetic-user password + its DB hash for the EXISTING synthetic family, never seeds/inserts/walks posture; refuses users_per_org/tz/w7_target in the same dispatch; owner_ref/entrypoint_inventory_ref/soak_orgs may still be supplied but are unused), punch_target (soak-run, default 200), config (soak-run host config path), window_start (soak-status override, ISO8601 UTC).'
         required: false
         default: ''
       smoke:
@@ -202,10 +202,11 @@ jobs:
                 w7_target) [[ "$value" == "group_shadow" ]] || { echo "w7_target only supports group_shadow at kickoff (deeper rungs carry compare-window exit predicates and are separate owner-run CLI acts), got '$value'" >&2; exit 2; } ;;
                 owner_ref) [[ "$value" =~ ^[A-Za-z0-9][A-Za-z0-9._:-]{0,127}$ ]] || { echo "owner_ref must match the manifest reference pattern, got '$value'" >&2; exit 2; } ;;
                 entrypoint_inventory_ref) [[ "$value" =~ ^[A-Za-z0-9][A-Za-z0-9._:-]{0,127}$ ]] || { echo "entrypoint_inventory_ref must match the manifest reference pattern, got '$value'" >&2; exit 2; } ;;
+                rotate_password) [[ "$value" == "true" ]] || { echo "rotate_password only accepts 'true' (a standalone password-only rotation), got '$value'" >&2; exit 2; } ;;
                 punch_target) [[ "$value" =~ ^[1-9][0-9]{0,3}$ ]] || { echo "punch_target must be 1..9999 (remote further caps it at the config's one-day capacity), got '$value'" >&2; exit 2; } ;;
                 config) [[ "$value" =~ ^[A-Za-z0-9._/~-]+$ ]] || { echo "config must be a plain path, got '$value'" >&2; exit 2; } ;;
                 window_start) [[ "$value" =~ ^[0-9]{4}-[0-9]{2}-[0-9]{2}(T[0-9]{2}:[0-9]{2}:[0-9]{2}Z)?$ ]] || { echo "window_start must be YYYY-MM-DD or YYYY-MM-DDTHH:MM:SSZ, got '$value'" >&2; exit 2; } ;;
-                *) echo "unknown soak_opts key: '$key' (known: users_per_org, tz, w7_target, owner_ref, entrypoint_inventory_ref, punch_target, config, window_start)" >&2; exit 2 ;;
+                *) echo "unknown soak_opts key: '$key' (known: users_per_org, tz, w7_target, owner_ref, entrypoint_inventory_ref, rotate_password, punch_target, config, window_start)" >&2; exit 2 ;;
               esac
             done
           fi
--- a/scripts/ops/attendance-staging-window-runner-remote.sh
+++ b/scripts/ops/attendance-staging-window-runner-remote.sh
@@ -188,6 +188,10 @@ SOAK_GENERATOR_SCRIPT="attendance-w4w7-soak-load-generator.mjs"
 SOAK_UUID_RE='^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$'
 # Manifest reference pattern — mirrors REF_PATTERN in both transition-CLI libs.
 SOAK_REF_RE='^[A-Za-z0-9][A-Za-z0-9._:-]{0,127}$'
+# soak_seed_rotate_password's RAISE NOTICE result line (no GNU-vs-BSD `\+` trap: bash's own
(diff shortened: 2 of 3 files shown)


zensgit:
Addressed the adversarial gate findings (`soak-working/GATE-5063.md`, REQUEST_CHANGES 0P1/2P2/3P3/2NIT) at `f188aaa0d3`:

- F1 (P2): dropped `-e` from the rotate psql call (was leaking the interpolated bcrypt hash into a world-downloadable CI artifact) — verified `-q` was the WRONG fix (silently breaks the row-count parse), matching the gate's own M6 finding.
- F2 (P2): `rotated_users` must now be `> 0`; a 0-row match auto-restores `.prev` and fails instead of reporting `result=ok`.
- F3 (P3): pre-COMMIT blast-radius checks (999-row ceiling, row-count equality, "not everything touched"), all inside the same transaction.
- F4 (P3): reworded the DB-step failure message — nonzero exit no longer claims "nothing committed".
- F5 (P3): documented the single-generation `.prev` limitation (chose not to add `.prev2`).
- NIT-1/NIT-2: umask-077 idiom for `.prev`; documented the owner_ref/entrypoint_inventory_ref asymmetry.

Also found (and fixed) empirically during verification, outside the original gate report: psql's `:'var'` substitution does not reach inside a `DO $$ ... $$` body — the F1-fixed SQL would have shipped syntactically broken without routing values through `set_config`/`current_setting`. All four scenarios (normal/zero-row/blast-radius/ceiling) verified against a disposable `postgres:16` container.

Tests: 105/105 (`node --test scripts/ops/attendance-window-runner-pipeline.test.mjs`), including a new EXECUTABLE leg that drives the real rotation function end-to-end (fake `docker` on PATH) for the 0-row path, and the gate's own M4/M6 mutations re-run as real on-disk edits (both now red, confirmed, restored).

zensgit:
Addressed round-2 gate finding (APPROVE 0P1/0P2 at `f188aaa0d3`, one new P3) at `1109c6f69a`:

`RAISE NOTICE` is gated by `client_min_messages`. If the staging psql session runs with the postgres default (`warning`), the artifact contains `COMMIT` but NO `ROTATE_RESULT` line even though the UPDATE genuinely committed — which breaks this PR's own F4 recovery rule ("markers absent" was meant to reliably signal "did not commit").

Fix: `SET LOCAL client_min_messages = notice;` immediately after `BEGIN;` in the rotation transaction (transaction-scoped, recovers the notice regardless of server/session default).

Verified empirically against a disposable `postgres:16` container both ways:
- Without the fix, `PGOPTIONS='-c client_min_messages=warning'` against the shipped SQL prints `BEGIN / DO / COMMIT` with no `ROTATE_RESULT` line — reproducing the exact gap.
- With the fix, the same invocation (and the default session) both print the `NOTICE: ROTATE_RESULT family_count=... updated_count=...` line, with zero bcrypt-hash occurrences in the output either way. Ceiling and blast-radius checks re-verified unaffected.

Tests: 106/106 (`node --test scripts/ops/attendance-window-runner-pipeline.test.mjs`, was 105) — one new static pin (position: after `BEGIN`, before anything else that could `RAISE NOTICE`) plus a MUTATION leg. Re-ran as a real on-disk mutation: removing the line took the suite from 106→82 pass/24 fail, confirmed the new assertion was what tripped, restored via file backup (never `git checkout -- .`) back to 106/106. `bash -n` OK, s6a provenance test OK (no `plugin-tests.yml` change).

PR #5065, opened 2026-08-21
feat(integration-core): stock-prep customer config pack + additive installer (P0)

skiyakubali0-cloud:
## What

The per-customer 备料 landing sheet is today hand-built in the UI (48 flat columns, every field `property:{}` — no ownership, so a PLM refresh cannot tell system columns from human-filled ones). This PR makes the customer-specific delta a **declarative, reviewable object**: a *customer pack* = extra `ext_` columns (with ownership) + select dictionaries + role-banded views, installed ADDITIVELY on top of the FROZEN canonical `plm_stock_preparation_main`.

通用 = the plugin's provisioning/ownership machinery (unchanged, one copy). 配置 = the pack (per customer, data). Acceptance: a sheet's customer shape becomes reproducible from one object.

## Files

- `lib/stock-preparation-customer-pack.cjs` — pure normalizer/validator. Fail-closed unknown keys; every ext id through `assertExtensionFieldIdValid` against the real frozen catalog; ownership/type vocabularies REUSED from `stock-preparation-templates.cjs`; option literals through the C6 `normalizeOptionSet` (1–200 cap, dedupe, key whitelist, executable-key rejection). `preserveOnRefresh` and per-view `hiddenFieldIds` derived at normalize time.
- `lib/stock-preparation-customer-pack-installer.cjs` — `installCustomerPack({ provisioning, projectId, pack, logger })`. Fields via `ensureMissingObjectFields` ONLY (a test greps the source to prove `.ensureObject(` is absent — the wholesale-property-overwrite hazard stays unreachable); options via the shared `field-option-sync-runtime` kernel; views via `ensureView` with logical→physical id resolution. Values-free summary return.
- `lib/customer-packs/factory-a.sample.cjs` — shape exemplar for tests/dev only (10 ext fields, 12-entry material dictionary, 生产/采购 role views). Real packs are deploy-time data from an uncommitted local file.
- 2 test suites (+ wired into the package test chain).

## Invariants

- Frozen canonical templates and their gates untouched.
- Additive-only: second install run = no-op on fields (tested); never `ensureObject` on re-run.
- Zero external writes; no new env flags; values-free throughout.

## ⚠ For review attention

1. **`lib/sealed-export/vectors/s6a-package-provenance-pins.json`: 1 hash line changed.** Adding the two suites to `scripts.test` (the chain-completeness guard requires it) moves the `runtimeFiles.pluginPackageJson` pin. Recomputed with `computePackageProvenancePinSet`; confirmed the single-entry drift. Please verify this is the sanctioned re-pin procedure.
2. Canonical target is a **precondition**, not created by the installer (absent sheet → `CUSTOMER_PACK_TARGET_ABSENT` naming `ensureStockPreparationCanonicalTarget`) — keeping `ensureObject` provably out of reach.
3. `normalizeOptionSet` reached via option-sync `__internals` (same pattern the extension-namespace module already uses); the public `optionSetsFromInput` injects contract defaults a pack must not inherit.
4. Sample ownership assignments (毛胚* as `human_preserved`) are illustrative — real packs decide per customer.

## Verify

- New suites: both OK (`node __tests__/stock-preparation-customer-pack*.test.cjs`)
- Chain completeness + package provenance suites OK
- Full package chain: 150 pass / 3 env-skip / 10 fail — **all 13 non-passes reproduce identically on clean origin/main** (worktree without package `node_modules`: `@metasheet/mssql-readonly-utils`/`pg`/`mssql`/`tsx` unresolvable; plus POSIX path-separator assertions on Windows). None reachable from this change.
- 12 targeted mutations against the two new modules (gate removals, normalizer bypass, ownership band ignored, brand widened, logical view ids left unresolved, `preserveOnRefresh` flipped, …) — **all 12 caught by the suites**.


skiyakubali0-cloud:
Real-data input from the live customer sheet (read via authorized session today): the material dictionary (物料ID) has **203 options — exceeds MAX_OPTIONS_PER_FIELD=200** inherited from option-sync, so the real factory-a pack would be rejected as-is. Other five dictionaries fit (19/15/11/7/6). Recommended resolution for pack v2: oversized dictionaries become a dictionary sheet + link field instead of select options (also fixes the two-segment '编号 - 材质' search UX); alternative is raising the cap. Also observed: role separation in the live base is by SHEETS (采购跟进/仓库跟进 satellites), not views — the sample's roleViews remain a net improvement (live views hide zero columns). Full mapping handed to owner as deploy-time data (not committed).

skiyakubali0-cloud:
Rehearsal (stacked branch) surfaced two review-relevant facts: (1) this PR's installer **mock merges field properties shallowly**, while the real host's `mergeJsonObject` (multitable/provisioning.ts) merges recursively — the suites here assert on the submitted descriptor, never the stored row, so they cannot detect an option-sync wiping `property.stockPreparation`. The installer itself is correct; the stacked rehearsal adds a host-faithful fake asserting ownership survives option sync on all five selects. (2) `normalizeOptionSets` accepts **canonical**-field option sets (gates only on `type==='select'`) — intended? It is useful (materialType/blankType/stockPreparationStatus dictionaries ship in the pack) but worth an explicit test either way.

skiyakubali0-cloud:
Finding **confirmed in full** and fixed (c683e62bf). Verification before fixing: `ensureMissingObjectFields` is `INSERT … ON CONFLICT (id) DO NOTHING` (core-backend `multitable/provisioning.ts:531`) — an existing row's `property` is constructively untouchable by that path; the installer then only *relabelled* skipped ids. On the real takeover shape (`property: {}` everywhere) it created nothing, reported success, and left every human column writable by any pipeline.

Fix: **validate-all-then-write**. A pre-scan reads current properties of all pack-declared fields; existing fields classify as NEEDS_STAMP / ALREADY_STAMPED / CONFLICTING. **Any conflict aborts the whole install (`CUSTOMER_PACK_OWNERSHIP_CONFLICT`, values-free field ids) before a single mutation** — silent re-classification of a live column is never allowed. NEEDS_STAMP fields get `{ownership, preserveOnRefresh, extension, packId, packVersion}` patched additively after the gate. Conflict detection spans both guard namespaces (`stockPreparation` + `stockPreparationMvp`), though only the first is ever written.

Also: the installer test mock now mirrors the host's recursive `mergeJsonObject` (the old shallow spread misrepresented the host and would have hidden an ownership stamp wiping a synced option set). `package.json` and the provenance pin are untouched — the pin did not move. 4/4 mutations killed; 6 suites green locally.

Still open (agreed, not in this PR): no HTTP/UI/CLI install entry point. See #5079's finding — nothing can enumerate installed pack fields yet, so the entry-point PR is also the registry PR.

PR #5068, opened 2026-08-21
feat(multitable): row provenance panel — first consumer of listProvenanceByRow

skiyakubali0-cloud:
## What

Business users opening a record on an integration-fed sheet (the 备料 landing sheet) now see **数据来源 / Provenance** at the foot of the record inspector's 详情 tab: which pipeline run wrote this row, when, and what event — the first UI consumer of the integration line's `listProvenanceByRow` (implemented at `lib/pipelines.cjs:727`, previously LATENT per the 2026-08-18 status ledger).

## Binding (existing surfaces only — zero backend/plugin edits)

- Route: `GET /api/integration/provenance?rowId=…` (route table `http-routes.cjs:142`, handler `:5263`), gate `requireAccess('read')` → `integration:read|integration:write|role:admin|integration:admin` — **unchanged**.
- FE client: reused the existing `listIntegrationProvenanceByRow` (`services/integration/workbench.ts:1156`, DF-N2-3 operator surface) — no duplicate client.

## Row identity (key finding)

Provenance groups on the pipeline idempotency key; the target adapter stamps `_integration_idempotency_key` only under `includeInternalFields: true`. So: sheet without the key column → section renders nothing; key column empty (hand-created row) → empty state **without a request**; timeline is fetched by rowId alone.

## Guardrails

- Renders **only** for actors passing `hasPermission('integration:read')` (repo's imperative precedent via useAuth); 401/403 retracts the section silently. Known v1 limitation (business users without the permission see nothing) documented in component + util headers — widening the gate is a separate design decision.
- Values-free rendering: whitelisted attrs only (runId short, labels, timestamps, event kinds) — a planted business string in `errorMessage` fails the spec if the whitelist is bypassed.
- Lazy: loads on first expand, never on record open (no new request on the critical path).

## Verify (agent run, junction-linked deps; CI adjudicates)

- New spec **14/14**; negative controls confirmed load-bearing then reverted: eager-load → 4 fail; render-all-attrs → values-free test fails; drop permission gate → no-permission test fails.
- Regression sweep **12 files / 299 tests green** (record inspector, field/history/attachment/comment panels, workbench view + permission wiring, IntegrationWorkbenchView, integrationWorkbench).
- `vue-tsc --noEmit -p tsconfig.app.json` clean — note: `-p tsconfig.json` is a solution-style no-op (`files:[]`), verified by deliberate-error probe; other PRs' local type-checks are being re-run under the app config.

### DIFF
--- a/apps/web/src/multitable/components/MetaRecordInspector.vue
+++ b/apps/web/src/multitable/components/MetaRecordInspector.vue
@@ -213,6 +213,15 @@
           @open-person-picker="(field) => emit('open-person-picker', field)"
           @run-button="(payload) => emit('run-button', payload)"
         />
+        <!-- 数据来源 / Provenance: read-only row lineage for integration-fed sheets. Self-gating —
+             it renders nothing unless the actor passes the integration read gate AND the sheet
+             carries a readable provenance key column, and it fetches only on first expand (no new
+             request on the record-open critical path). See the component's file header. -->
+        <MetaRecordProvenancePanel
+          :record="record"
+          :fields="fields"
+          :field-permissions="fieldPermissions"
+        />
       </div>
       <div
         v-else-if="activeTab === 'history'"
@@ -333,6 +342,7 @@ import { MtButton } from '../ui'
 import MetaCommentActionChip from './MetaCommentActionChip.vue'
 import MetaRecordPermissionManager from './MetaRecordPermissionManager.vue'
 import MetaRecordFieldsPanel from './MetaRecordFieldsPanel.vue'
+import MetaRecordProvenancePanel from './MetaRecordProvenancePanel.vue'
 import MetaRecordHistoryPanel from './MetaRecordHistoryPanel.vue'
 import MetaCommentsPanel from './MetaCommentsPanel.vue'
 import MetaRecordAttachmentsPanel from './MetaRecordAttachmentsPanel.vue'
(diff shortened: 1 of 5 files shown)


skiyakubali0-cloud:
One finding confirmed and fixed, one **refuted with evidence** (7f30e86b5).

**A (record-switch race): real, fixed.** Added `activeLoadVersion`, a plain closure counter mirroring `useMultitableCommentPresence.ts`'s existing pattern. Every state write after the `await` is guarded; the counter is bumped both in the rowId watcher and in `onUnmounted`, so a stale in-flight response — whether superseded by a row switch or outliving the component — is discarded with state untouched. `listIntegrationProvenanceByRow` accepts no `AbortSignal` and forwards no options to `apiFetch`, so no signal was plumbed through the shared service; the generation guard carries the fix alone.

**B (cross-pipeline mixing): refuted — already correct.** Every response item carries `pipelineId` (from `integration_runs.pipeline_id`, the run's own FK — never null, never parsed out of the free-form event JSON; see `lib/pipelines.cjs` `rowToProvenanceEntry` + migration `060_integration_runs_provenance.sql`'s `integration_provenance_by_row` view) and `runId`. Both are typed on `IntegrationProvenanceTimelineEntry` and were **already rendered per entry** before this branch. The real gap was test coverage: the existing render test gave both fixture entries the same `pipelineId`, so it never exercised the collision case. Added a test with two entries sharing one `rowId` but different `pipelineId`/`runId`, asserting distinct labels. No whitelist, template, or backend change was needed.

Tests: deferred-promise race (slow rowA + fast rowB → only B renders, A's late resolution discarded); mid-flight rowId switch resets loading state and cannot be reverted by a late settle; two-pipeline identity labels. **17/17 green and `vue-tsc -p tsconfig.app.json` clean, re-verified in the main checkout** (the agent worktree had no deps and could not run them — flagging that rather than claiming a green it never saw).

PR #5074, opened 2026-08-21
feat(integration-core): factory-A customer-pack install rehearsal (stacked on #5065)

skiyakubali0-cloud:
## What (stacked on #5065 — merge that first)

The executable proof behind 通用/配置分离: **one config object regenerates the full 46-field customer landing-sheet shape, and a PLM-style refresh preserves every human-filled cell.** Five acts in one dependency-free node suite against a host-faithful fake:

1. Normalize the full-shape rehearsal pack (21 `ext_` fields: 13 plm_system + 8 human_preserved; **5 option sets**; 3 role views) — all derived `preserveOnRefresh` values asserted.
2. Install #1 onto canonical-25 → 21 created, options synced on all five selects, 3 views ensured; every ext property carries `{ownership, preserveOnRefresh, extension, packId, packVersion}`.
3. Install #2 → idempotent no-op (0 created, no destructive calls).
4. **Refresh-preservation money shot**: ownership-filtered projection updates every plm cell, leaves every human cell byte-identical — plus the negative control proving the unfiltered write WOULD clobber.
5. Values-free guard over summary+logs.

## Findings (feed #5065 review + next increments)

1. **Canonical-field option sets are accepted** (normalizer gates on `type==='select'` only) → the pack ships materialType/blankType/stockPreparationStatus dictionaries too — 5 sets, not 2.
2. **#5065's installer mock merges properties shallowly; the real host merges recursively** (`mergeJsonObject`) — its suites assert the submitted descriptor, never the stored row, so they cannot see an option-sync wiping ownership. The installer IS correct; this rehearsal's fake mirrors the host and asserts ownership survives option sync on all five selects.
3. `ownership`/`preserveOnRefresh` redundancy is only broken once a human pins a column — covered explicitly.
4. **Today's refresh is safe by omission, not by ownership**: the conflict planner's writable set comes from the frozen template and has never heard of `ext_` columns. The rehearsal's `plmWritableFieldIds` is the *specification* for the pack-aware refresh — the next real increment.
5. Entry-point frictions recorded (no dry-run yet; summary lacks per-field ownership; role views hide by name, not ownership) → feeds the CLI/HTTP installer entry design.

`ext_materialCode` stays `string` in v1: the real dictionary is 203 entries > the 200-option cap — carrier decision (dictionary sheet + link) tracked separately.

## Verify

- New rehearsal suite + both #5065 suites + `test-chain-completeness` (164 suites): green in the agent worktree AND re-run green in the main checkout. `sealed-export-package-provenance`: green in the agent worktree; on the maintainer's long-lived Windows checkout it fails in `verifyPinnedMigrations` — the documented CRLF/stale-working-file class (pins hash LF bytes), and #5065's own CI (25/25, same suite) is the adjudicator.
- 11 targeted mutations all caught (ownership flip, filter drops, neutered negative control, option retarget, view banding ×2, property strip, derivation fold, IPv4 log leak).
- Full chain 151 pass / 13 fail — identical 13 on stashed clean base (worktree deps + Windows path assertions), verified by stash/rerun.
- Provenance pin: single-line `runtimeFiles.pluginPackageJson` recompute.

Report: `docs/development/takeover-beiliao-20260821/customer-pack-rehearsal-report.md`


skiyakubali0-cloud:
**Restacked on #5065's Codex fix (4c16bda47).** Found by rehearsing the full merge sequence locally rather than discovering it at merge time.

#5065's ownership-stamp fix added `readObjectFieldsContent` to the installer's `REQUIRED_PROVISIONING_METHODS` (it now pre-scans existing columns before writing anything). This branch's fake provisioning predates that commit, so with both merged the rehearsal died on `CUSTOMER_PACK_API_UNAVAILABLE`.

Fixed here: the fake gains `readObjectFieldsContent` with the host's contract shape (logicalId → {name,type,property,order}, absent ids omitted), so the rehearsal now exercises the pre-scan path too; summary assertions extended for the two new counters (`stampedExistingFields`/`alreadyStampedFields`, both 0 here — 21 fresh ext_ columns onto a canonical-only sheet; the hand-built-column takeover case stays covered by the installer suite). 4 suites green.

PR #5076, opened 2026-08-21
fix(multitable): stop persisting battery admin creds in the container layer (owner P1, round 2)

zensgit:
Owner review proved with **real Docker** that the previous credential-lifecycle fix has a live leak: the battery workflow `docker cp`'d the admin password into the container at `/tmp/o2bat-creds-<stamp>/password`; a **stopped** container is absent from `docker ps`, so the scrub SKIPPED and still printed `VERDICT: PASS` — while `docker cp <stopped>:.../password -` still returns the bytes from the writable layer. The workflow's own test pinned that wrong PASS green.

## Fix — stdin-only ingestion (no secret ever in the container layer)
Secrets are piped to `docker exec -i` as a two-line base64 `KEY=…` blob the container-side wrapper reads, decodes, and exports as `BATTERY_ADMIN_*` in-process. **Nothing is `docker cp`'d into the container.** Matches the ratified house pattern (`dingtalk-lifecycle-staging-canary-remote.sh`: secrets over `docker exec` stdin, no persistent secret files).

## Container-down is now honest
The `docker ps` skip-gate is gone. A non-running container's writable layer is **enumerated via `docker cp :/tmp | tar -tf`**, behind a daemon-liveness control + a `docker cp :/etc/hostname` positive control (both → `fail=1`). **Present-and-not-running ⇒ FAIL, never PASS.** Real-Docker proof on a genuinely `exited` container.

## Tests
`behavior (b) container down` rewritten off the blind-spot PASS; 3 **registry-free** real-Docker goldens reproduce the leak in-test then prove the shipped scrub FAILs on a leaking non-running container and PASSes on a clean one. Loud-SKIP (never red) without a docker daemon — verified: **with docker 42/42; without docker 39 pass + 3 loud skip + 0 fail**. Mutations (cp-restore, byte-identical): full revert reds 24 incl. both goldens; reintroducing `docker cp` into the container reds exactly the 2 ingestion gates (incl. flag-shifted `docker cp -a`); re-gating on running-ness reds 7.

## ⚠️ DISCLOSED — third file touched
The goldens make the **required** obs-kit `contract` job able to drive the runner's local docker daemon, so `multitable-o2-observation-kit.yml`'s header no longer claims "Hermetic" (amended rather than leave a falsified claim in a required check). Docker-less runners loud-skip (safe); docker-present runners (GitHub ubuntu) run the goldens. **Tradeoff**: real proof the leak is closed, at a small new docker-daemon flake surface on that required lane (registry-free + daemon-liveness-controlled to minimize it). Owner may prefer moving the goldens to a non-required Docker lane — flagged for your call.

CI-wiring contract family re-run green; ssh-hostkey-pin 12/12; yaml + actionlint clean.

**This is the F1 hard-blocker for A1**: no staging account creation, no dispatch, no ratify until this lands + passes an independent re-gate. Head `d22c7538710d5dac3a0c78d8eb75ec5f144fc0e5`.

### DIFF
--- a/.github/workflows/multitable-l1-battery-docker-goldens.yml
+++ b/.github/workflows/multitable-l1-battery-docker-goldens.yml
@@ -1,0 +1,46 @@
+name: L1 battery credential-scrub goldens (real Docker)
+
+# NON-REQUIRED lane. Runs the real-Docker `golden` cases in
+# scripts/ops/multitable-l1-battery-workflow.test.mjs (opt-in via L1_BATTERY_DOCKER_GOLDENS=1) that
+# prove the L1-battery credential-lifecycle fix: a container absent from `docker ps` still yields
+# its writable layer to `docker cp` while refusing `docker exec`, so the scrub must FAIL (not PASS)
+# when a credential path survives on a stopped container. These drive the runner's local docker
+# daemon, so they are DELIBERATELY kept OUT of the required obs-kit `contract` lane (which stays
+# hermetic) — a docker hiccup here does not block merges of unrelated PRs. Registry-free (docker
+# import of a built rootfs; nothing pulled). Path-filtered to the battery workflow + its test so it
+# only runs when they change.
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - '.github/workflows/multitable-l1-battery.yml'
+      - 'scripts/ops/multitable-l1-battery-workflow.test.mjs'
+      - '.github/workflows/multitable-l1-battery-docker-goldens.yml'
+  push:
+    branches: [main]
+    paths:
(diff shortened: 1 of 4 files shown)


zensgit:
**Correction (owner review 3):** this PR body describes the real-Docker goldens as living in the required obs-kit `contract` lane. A follow-up commit in this same PR (`ad135e7845`) then MOVED them to a non-required opt-in lane (`multitable-l1-battery-docker-goldens.yml`, gated on `L1_BATTERY_DOCKER_GOLDENS=1`) so the required lane stays hermetic and a docker hiccup can't red an unrelated PR. The final landed state is: goldens on the non-required lane; obs-kit contract lane hermetic again. The body above describes an intermediate state.

PR #5079, opened 2026-08-21
feat(integration-core): pack-aware refresh writable set + extended human wall (stacked on #5074)

skiyakubali0-cloud:
## What (third layer of the stack: #5065 → #5074 → this)

Closes the gap #5074 documented as *finding 4*: today "refresh never clobbers human cells" holds only because the conflict planner derives its writable set from the FROZEN template and has never heard of `ext_` columns — **safe by omission**. Once a real customer pack installs `ext_` plm_system columns, refresh must write them, and the human/plm boundary must come from the INSTALLED field properties, not from luck.

## How

- `derivePackAwarePlmWritableFields({ templateFields, installedFieldProperties })` exported from the planner (pure, no I/O; accepts array/Map/object shapes). **Fail-closed precedence**: `preserveOnRefresh:true` → human (wins both directions) → `human_preserved` → human → `plm_system` **and** `extension:true` → writable → anything else unclassified (neither band, values-free reason). Canonical ids in the installed set are `template_governed` — **no installed property can re-classify a frozen column**.
- Human wall extended on BOTH sides: planner bands + apply-writer `assertNoHumanFields` now reject `ext_` human fields by name (reachable end-to-end rejection test, not absence).
- Threaded through `computeDryRun` → both table-action entrypoints AND both large-BOM paths (plan + chunked apply), and folded into the dry-run token revision — **a pack install landing between dry-run and apply invalidates the token**.
- **Legacy posture unchanged**: HTTP routes supply no projection today (no registry can enumerate installed `ext_` ids — see finding below), so their behavior is byte-identical; control test pins `JSON.stringify(plan)` equality on a 4-decision fixture, and `summary.packAwareOwnership` is emitted only when a projection was supplied.

## Review attention

1. **The planner's internal wall is defense-in-depth, not a live gate** — with a correct derivation no ext_ human column reaches `pickFields`; the initially-surviving mutation was closed by asserting the plan *publishes* the same bands it planned with, plus the reachable apply-writer rejection.
2. **Rehearsal guard kept as an independent restatement** (it is looser — predates the extension stamp — and differently scoped re `lastPlm*`); an equivalence pin asserts production is a strict subset, so the two cannot drift.
3. **Activation prerequisite discovered**: nothing today can enumerate installed pack fields (provisioning exposes only per-field `getObjectField`; there is no pack registry). The installer CLI/HTTP entry-point PR is now also the *registry* PR — until then, pack-aware refresh is wired but dormant.
4. Provenance pin: single-line recompute (stated in commit body).

## Verify

- New suite + conflict-planner + apply-writer + sync-run-plan + both pack suites + rehearsal + table-actions + large-bom-jobs + test-chain-completeness (165 suites) + package-provenance: green in the agent worktree; **9 key suites re-run green in the main checkout**.
- **8 targeted mutations, all red** (drop extension stamp; admit human_preserved; unextended planner wall; fail-open on malformed stanza; drop pin precedence; property re-classifying a canonical column; reverted writer wall; unconditional summary key) — run, recorded, reverted.
- Full chain segment-by-segment: 152/165 vs 151/164 on stashed base — same 13 env failures (worktree deps, node 25), none in stock-preparation.


skiyakubali0-cloud:
**Restacked (154d5f968)** to carry #5074's rehearsal restack (which in turn carries #5065's ownership-stamp fix). No conflicts; 7 suites green locally: pack-aware-refresh, rehearsal, conflict-planner, apply-writer, table-actions, large-bom-jobs, chain-completeness.

Context: rehearsing the whole merge sequence locally surfaced that #5065's Codex fix added `readObjectFieldsContent` to the installer's required-methods set, which broke the rehearsal fake in #5074 once both were merged. Each PR was green in isolation — only the integrated sequence showed it.

